PR #402, opened 2016-01-24
Add deprecation notices to WDL specific command line functionality

geoffjentry:


geoffjentry:
@cjllanwarne @scottfrazer (also see https://github.com/broadinstitute/wdltool/pull/1 which is the corresponding change)

src/main/scala/cromwell/Main.scala
```diff
@@ -66,6 +67,10 @@ object Main extends App {
     argCapitalized = arg.capitalize
     action <- Actions.values find (_.toString == argCapitalized)
   } yield action
+
+  def deprecationNotice(): Unit = {
+    Console.err.println("This functionality is deprecated. Please use wdltool: https://github.com/broadinstitute/wdltool")
```
cjllanwarne: This is fine as long as github is the main way people will download cromwell.

As soon as people start getting Cromwell in binary form (e.g. homebrew, website, ...?) this should point to our official release place.

geoffjentry: That's exactly the plan

cjllanwarne: Also, Cromwell and wdltool appear to have different versioning structures.

Since it's a sine qua non that the versions match up - this message should point to a specific wdltool version. And wdltool should probably have a switch to say which cromwells it supports?

geoffjentry: Cromwell should not know about wdl tool and vice versa

cjllanwarne: Good point. And good latin rebuttal.

geoffjentry: It was unintentional but I smiled when I saw it

geoffjentry: Eventually the right thing to do is for both to state whichx version of the spec they support

geoffjentry:
@cjllanwarne  @scottfrazer  Note that I removed the deprecation notice from validate. While the current form wouldn't be long for this world we know that it'll reappear w/ slightly different functionality. Instead of removing it and re-adding the endpoint IMO we might as well leave it here (e.g. in the future it wouldn't be a wdl validation but just a workflow validation - whether that be wdl, cwl, etc - in the context of that cromwell jar)

cjllanwarne:
:+1: do we have a timescale for the full erasure? It might be nice to say "deprecated... and will disappear in Cromwell 0.18", for example

geoffjentry:
@cjllanwarne My plan was following the next release I'd make the change in develop, so 2 releases from now

src/main/scala/cromwell/Main.scala
```diff
@@ -66,6 +67,10 @@ object Main extends App {
     argCapitalized = arg.capitalize
     action <- Actions.values find (_.toString == argCapitalized)
   } yield action
+
+  def deprecationNotice(): Unit = {
+    Console.err.println("This functionality is deprecated and will be removed in 0.18. Please use wdltool: https://github.com/broadinstitute/wdltool")
```
scottfrazer: This might be a good place for red text instead of default colored text.  I tried it locally and it stands out a lot more for me

```
Console.err.println(TerminalUtil.highlight(1, "This functionality is deprecated and will be removed in 0.18. Please use wdltool: https://github.com/broadinstitute/wdltool"))
```

scottfrazer:
:+1: Besides the one comment

src/main/scala/cromwell/Main.scala
geoffjentry: You and your highlighting ...

PR #405, opened 2016-01-26
Update build sbt to 0.18

Horneth:
RC branch for 0.17 is here https://github.com/broadinstitute/cromwell/tree/release-0.17

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -3,7 +3,7 @@ import sbtrelease.ReleasePlugin._
 
 name := "cromwell"
 
-version := "0.17"
+version := "0.18"
 
 organization := "org.broadinstitute"
 


mcovarr:
Is there a particular motivation for this change?

Horneth:
https://broadinstitute.atlassian.net/wiki/display/DSDEEPB/Cromwell+Release

cjllanwarne:
Are there any other changes which will have to be wrapped up in the RELEASE 0.17 umbrella?

cjllanwarne:
Also, this is going to be 0.17 so the original version was correct.. maybe we should update the doc to update the SBT version on the first Develop commit AFTER branching?
:-1:

geoffjentry:
@cjllanwarne :us:

kshakir:
:+1:

PR #411, opened 2016-01-29
Fix bug discovered in load testing

scottfrazer:


src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -50,9 +50,14 @@ object WorkflowActor {
   final case class CachesCreated(startMode: StartMode) extends WorkflowActorMessage
   final case class AsyncFailure(t: Throwable) extends WorkflowActorMessage
   final case class PerformTransition(toState: WorkflowState) extends WorkflowActorMessage
+  sealed trait PersistenceMessage extends WorkflowActorMessage {
+    def callKey: ExecutionStoreKey
+    def executionStatus: ExecutionStatus
+  }
   final case class PersistenceCompleted(callKey: ExecutionStoreKey,
```
mcovarr: Yay ADT!  This makes me think what is currently the `PersistenceCompleted` message should probably be renamed to `PersistenceSucceeded` to be more clear.  There is also a multi-persistence `PersistencesCompleted` (note the plural) that might be renamed and brought under the base type as well.

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -569,10 +581,8 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
       // Running and Aborting have explicit handlers for Running and no pending writes, so either we are not in those
       // states or the specified key is no longer Running and/or has pending writes.
       stay()
-    case Event(PersistenceCompleted(callKey, status, _), data) =>
-      logger.debug(s"Got whenUnhandled PersistenceCompleted(${callKey.tag}): $status")
-      val updatedData = data.removePersisting(callKey, status)
-      stay using updatedData
+    case Event(m: PersistenceCompleted, data) => removePendingCallKeyPersistence(data, m)
+    case Event(m: PersistenceFailed, data) => removePendingCallKeyPersistence(data, m)
```
mcovarr: if everything is under the base trait then these two and the `PersistencesCompleted` handler could be replaced with

```scala
case Event(m: PersistenceCompleted, data) => removePendingCallKeyPersistence(data, m)
```

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -552,6 +558,12 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
     context.system.scheduler.scheduleOnce(ResendInterval, self, message)
   }
 
+  private def removePendingCallKeyPersistence(data: WorkflowData, message: PersistenceMessage) = {
+    logger.debug(s"Got whenUnhandled message: $message")
```
cjllanwarne: This line should go in the message handling block rather than this private function

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -552,6 +558,12 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
...
+    val updatedData = data.removePersisting(message.callKey, message.executionStatus)
+    stay using updatedData
```
cjllanwarne: Hmm, I'm not sure I really like this return type either. But now I see this is very specifically a message handling method so maybe neither are quite so terrible a I first though

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -50,9 +50,14 @@ object WorkflowActor {
   final case class CachesCreated(startMode: StartMode) extends WorkflowActorMessage
   final case class AsyncFailure(t: Throwable) extends WorkflowActorMessage
   final case class PerformTransition(toState: WorkflowState) extends WorkflowActorMessage
-  final case class PersistenceCompleted(callKey: ExecutionStoreKey,
+  sealed trait PersistenceMessage extends WorkflowActorMessage {
+    def callKey: ExecutionStoreKey
+    def executionStatus: ExecutionStatus
+  }
```
geoffjentry: \n after brace

cjllanwarne:
Nominations are in for "Best Reviewer": @mcovarr and @cjllanwarne

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
mcovarr: It needs to return the next state + state data, though it would be nice to see an explicit return type on the method.

mcovarr:
:+1: minus the one nit

scottfrazer:
Thanks, @cjllanwarne I sort of forgot about this ticket, I'll address comments now

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -569,10 +581,9 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
       // Running and Aborting have explicit handlers for Running and no pending writes, so either we are not in those
       // states or the specified key is no longer Running and/or has pending writes.
       stay()
-    case Event(PersistenceCompleted(callKey, status, _), data) =>
-      logger.debug(s"Got whenUnhandled PersistenceCompleted(${callKey.tag}): $status")
-      val updatedData = data.removePersisting(callKey, status)
-      stay using updatedData
+    case Event(m: PersistenceMessage, data) =>
+      logger.debug(s"Got whenUnhandled message: $m")
```
cjllanwarne: Is this debug necessary? It looks like this is really the expected path for Persistence message handling?

cjllanwarne:
:+1:

PR #412, opened 2016-01-29
Preemptible VMs

Horneth:


src/test/scala/cromwell/engine/backend/BackendCallSpec.scala
```diff
@@ -21,7 +21,7 @@ class BackendCallSpec extends CromwellTestkitSpec with ScalaFutures {
   "BackendCall hash function" should {
     "not change very often - if it changes, make sure it is for a good reason" in {
       val actual = backendCall.hash.futureValue.overallHash
-      val expected = "fe71298bb881a586178dca7c92fa945f"
+      val expected = "1ecd959c6081990491b214f5ae58850b"
```
Horneth: This changed because the runtime attribute went from a `Boolean` to an `Option[Int]`

cjllanwarne:
Review wheel says: @cjllanwarne and @scottfrazer

src/main/scala/cromwell/engine/ExecutionStatus.scala
```diff
@@ -2,11 +2,11 @@ package cromwell.engine
 
 object ExecutionStatus extends Enumeration {
   type ExecutionStatus = Value
-  val NotStarted, Starting, Running, Failed, Done, Aborted = Value
+  val NotStarted, Starting, Running, Failed, Preempted, Done, Aborted = Value
 
   implicit class EnhancedExecutionStatus(val status: ExecutionStatus) extends AnyVal {
     def isTerminal: Boolean = {
-      Seq(Failed, Done, Aborted) contains status
+      Seq(Failed, Done, Aborted, Preempted) contains status
```
geoffjentry: I realize this is preexisting but we should start getting in the habit of not using `Seq()` and using `List()` instead. A call to `Seq()` *is* a call to `List()`, but it's less obvious.

Note to our resident OOP lovers (Looking at you @mcovarr and @kshakir ) - I'm not suggesting (yet) removing `Seq` from signatures

src/main/scala/cromwell/engine/WorkflowDescriptor.scala
```diff
@@ -104,10 +104,10 @@ case class WorkflowDescriptor(id: WorkflowId,
     val logger = backend.workflowLogger(this)
 
     def copyFile(file: WdlFile): Try[Unit] = {
-      val src = file.valueString.toPath(workflowLogger, gcsFilesystem)
-      val wfPath = wfContext.root.toPath(workflowLogger, gcsFilesystem).toAbsolutePath
+      val src = file.valueString.toPath(gcsFilesystem)
+      val wfPath = wfContext.root.toPath(gcsFilesystem).toAbsolutePath
       val relativeFilePath = src.subpath(wfPath.getNameCount, src.getNameCount)
-      val dest = destDirectory.toPath(workflowLogger, gcsOutputsFilesystem).resolve(relativeWorkflowRootPath).resolve(relativeFilePath)
+      val dest = destDirectory.toPath(gcsOutputsFilesystem).resolve(relativeWorkflowRootPath).resolve(relativeFilePath)
```
geoffjentry: Also pre-existing but I wonder if the double-resolve is necessary or can it somehow be done in one pass

src/main/scala/cromwell/engine/WorkflowDescriptor.scala
```diff
@@ -171,6 +171,9 @@ case class WorkflowDescriptor(id: WorkflowId,
 }
 
 object WorkflowDescriptor {
+
+  def buildWorkflowRootPath(rootPath: String, name: String, workflowId: WorkflowId) = s"$rootPath/$name/${workflowId.toString}"
```
geoffjentry: Do you need the explicit toString here?

src/main/scala/cromwell/engine/backend/ExecutionResult.scala
```diff
@@ -21,8 +21,13 @@ final case class SuccessfulBackendCallExecution(outputs: CallOutputs, executionE
...
+sealed trait FailedExecution extends ExecutionResult {
+  def e: Throwable
+  def returnCode: Option[Int]
+}
 
+/**
+  * Failed execution, possibly having a return code.
+  */
+final case class NonRetryableExecution(e: Throwable, returnCode: Option[Int] = None, events: Seq[ExecutionEventEntry] = Seq.empty) extends FailedExecution
+final case class RetryableExecution(e: Throwable, returnCode: Option[Int] = None, events: Seq[ExecutionEventEntry] = Seq.empty) extends FailedExecution
```
geoffjentry: Add a \n at the end of the file to keep @kshakir from coming over and insulting the honor of :fr:

src/main/scala/cromwell/engine/backend/jes/JesBackend.scala
```diff
@@ -177,11 +167,9 @@ case class JesPendingExecutionHandle(backendCall: JesBackendCall,
                                      run: Run,
                                      previousStatus: Option[RunStatus]) extends ExecutionHandle {
   override val isDone = false
-
-  override def result = FailedExecution(new IllegalStateException)
+  override def result = NonRetryableExecution(new IllegalStateException)
```
geoffjentry: This looks like it could be a val - it's always the same, right?

src/main/scala/cromwell/engine/backend/ExecutionResult.scala
```diff
@@ -21,8 +21,13 @@ final case class SuccessfulBackendCallExecution(outputs: CallOutputs, executionE
  */
 case object AbortedExecution extends ExecutionResult
 
-/**
- * Failed execution, possibly having a return code.
- */
-final case class FailedExecution(e: Throwable, returnCode: Option[Int] = None) extends ExecutionResult
+sealed trait FailedExecution extends ExecutionResult {
+  def e: Throwable
```
geoffjentry: Is this really an exception or are we just after the error message? i.e. is the actual exception type meaningful elsewhere?

src/main/scala/cromwell/engine/backend/jes/JesBackend.scala
```diff
@@ -625,6 +602,38 @@ case class JesBackend(actorSystem: ActorSystem)
...
+
+  private def handleFailure(backendCall: BackendCall, errorCode: Int, errorMessage: Option[String], events: Seq[ExecutionEventEntry]) = {
+    import cromwell.util.IntUtil._
+
+    val taskName = s"${backendCall.workflowDescriptor.id}:${backendCall.call.unqualifiedName}"
+    val attempt = backendCall.key.attempt
+
+    if (errorMessage contains "Operation canceled at") {
+      AbortedExecutionHandle.future
+    } else if(preemtped(errorCode, backendCall)) {
```
geoffjentry: space after if - and you've got a typo: `preemtped`

src/main/scala/cromwell/engine/backend/jes/JesBackend.scala
```diff
@@ -625,6 +602,38 @@ case class JesBackend(actorSystem: ActorSystem)
...
+      val preemptedMsg = s"Task $taskName was preempted for the ${attempt.toOrdinal} time."
+      val max = backendCall.maxPreemption
+
+      if (backendCall.key.attempt < max) {
```
geoffjentry: You assigned `backendCall.key.attempt` to `attempt` above. Either use `attempt` everywhere or don't create it

src/main/scala/cromwell/engine/backend/jes/JesBackend.scala
```diff
@@ -625,6 +602,38 @@ case class JesBackend(actorSystem: ActorSystem)
...
+        val e = new PreemptedException(
```
geoffjentry: Similar comment to what I made above. Are these exceptions merely vessels to carry error messages around the system or is the actual `PreemptedException` being used somewhere?

src/main/scala/cromwell/engine/backend/jes/JesBackend.scala
geoffjentry: This whole if block I think could be made smaller/broken out into functions

src/main/scala/cromwell/engine/backend/jes/JesBackendCall.scala
```diff
@@ -72,6 +72,23 @@ class JesBackendCall(val backend: JesBackend,
   def downloadRcFile = authenticateAsUser(workflowDescriptor) { storage => Try(storage.readFile(returnCodeGcsPath)) }
 
   /**
+    * Determines the maximum number of times a call can be started with a Preemptible VM.
+    * Runtime attributes > workflow options > Jes configuration
+    */
+  lazy val maxPreemption = {
```
geoffjentry: This bothers me but I can't quite put my finger on why at the moment

src/main/scala/cromwell/engine/backend/jes/JesRuntimeInfo.scala
```diff
@@ -6,7 +6,7 @@ import cromwell.engine.backend.runtimeattributes.CromwellRuntimeAttributes
 import scala.collection.JavaConverters._
 
 object JesRuntimeInfo {
-  def apply(commandLine: String, runtimeAttributes: CromwellRuntimeAttributes): JesRuntimeInfo = {
+  def apply(commandLine: String, runtimeAttributes: CromwellRuntimeAttributes, preemptible: Boolean): JesRuntimeInfo = {
```
geoffjentry: While I understand why it is the way it is, it seems a little odd that preemptible is being passed in considering the other runtime attribute derived things aren't. Perhaps a comment that that particular value is being modified by other logic upstream would be in order.

Also while not at all strictly necessary this would be an opportunity to have a JesRuntimeInfo and PreemptibleJesRuntimeInfo ADT where the latter gloms on the .setPreemptible. While that'd make me smile don't feel pressured to do so here

src/main/scala/cromwell/engine/io/gcs/GcsFileSystemProvider.scala
```diff
@@ -47,11 +48,14 @@ object ExecutionContextExecutorServiceBridge {
   * Implements java.nio.FileSystemProvider for GoogleCloudStorage
   * This implementation is not complete and mostly a proof of concept that it's possible to *copy* around files from/to local/gcs.
   * Copying is the only functionality that has been successfully tested (same and cross filesystems).
-  * @param googleCloudStorage must be properly set up (credentials) according to the context. Might be absorbed by this class eventually.
+  * @param gcs Interface to perform operations with GCS. Type is a try so this class can be instantiated even with a failed interface
+  *            to allow Path manipulations via GcsFileSystem. The interface might be absorbed by this class eventually.
   * @param executionContext executionContext, will be used to perform async writes to GCS after being converted to a Java execution service
   */
-class GcsFileSystemProvider private (googleCloudStorage: GoogleCloudStorage, executionContext: ExecutionContext) extends FileSystemProvider {
+class GcsFileSystemProvider private (gcs: Try[GoogleCloudStorage], executionContext: ExecutionContext) extends FileSystemProvider {
```
geoffjentry: Do we actually use the exception from the Try elsewhere or just the error message? If the latter, consider `ErrorOr[GoogleCloudStorage]`

src/main/scala/cromwell/engine/io/gcs/GcsFileSystemProvider.scala
```diff
@@ -47,11 +48,14 @@ object ExecutionContextExecutorServiceBridge {
...
 
+  // We want to throw an exception here if we try to use this class with a failed gcs interface
+  private lazy val googleCloudStorage = gcs.get
```
geoffjentry: nm, I think i just answered my own question

src/main/scala/cromwell/engine/workflow/ExecutionStoreKey.scala
```diff
@@ -14,17 +14,33 @@ sealed trait ExecutionStoreKey {
...
 
-trait OutputKey extends ExecutionStoreKey
+sealed trait OutputKey extends ExecutionStoreKey
 
-trait CallKey extends OutputKey
-case class BackendCallKey(scope: Call, index: Option[Int]) extends CallKey
-case class CollectorKey(scope: Call) extends OutputKey {
+sealed trait CallKey extends OutputKey
+case class BackendCallKey(scope: Call, index: Option[Int], attempt: Int = 1) extends CallKey {
+  private val CallPrefix = "call"
```
geoffjentry: These constants should be in a companion object

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -379,6 +383,32 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
       stay() using data.copy(startMode = Option(startMode))
   }
 
+  private def handleCallPreempted(callKey: OutputKey,
```
geoffjentry: Preempted might not be the right terminology outside of the jes backend. Other similar things might call it something else.

geoffjentry: and/or "preempted" might not be the reason we get here in other contexts

geoffjentry:
@Horneth Make sure the readme gets updated as appropriate

src/main/scala/cromwell/engine/ExecutionStatus.scala
Horneth: For this particular case I would change it to `Set` then because there is no ordering on the statuses

src/main/scala/cromwell/engine/WorkflowDescriptor.scala
Horneth: `Path.resolve` only takes 1 path argument unfortunately,
we can do
```scala
val relativeFilePath = Paths.get(relativeWorkflowRootPath).resolve(src.subpath(wfPath.getNameCount, src.getNameCount))
val dest = destDirectory.toPath(gcsOutputsFilesystem).resolve(relativeFilePath)
```

but there are still 2 `resolve`s

src/main/scala/cromwell/engine/backend/ExecutionResult.scala
Horneth: :fr: honor will not be insuled by a missing \n .. not today

src/main/scala/cromwell/engine/backend/ExecutionResult.scala
Horneth: Currently I don't think the type is actually checked anywhere.
However we can imagine that there would be different behaviors in the WA depending on the type of failure.
For retryable errors for instance we might want to do something differently if it's been `Preempted` or `WhateverOtherResasonTriggeringARetry` ?

src/main/scala/cromwell/engine/backend/jes/JesBackend.scala
Horneth: See comment above - I'm not opposed to changing it to a String - I just thought it would be nice to keep the information that it's been preempted (when this bubbles up) as a type rather than just an error message.

src/main/scala/cromwell/engine/ExecutionStatus.scala
geoffjentry: Which is a prime example of why it's bad to use `Seq()`. This whole time it *was* a `List` but you're saying it shouldn't be.

(I don't really care what it is, just that we're not creating Seqs willy nilly)

src/main/scala/cromwell/engine/WorkflowDescriptor.scala
geoffjentry: I had a feeling the cure would be worse than the disease. Disregard.

src/main/scala/cromwell/engine/backend/ExecutionResult.scala
geoffjentry: I'll admit that when I wrote this I was imagining using `ErrorOr[A]` but there's not actually an `Or` here :) 

The real culprit that I'm complaining about is Try because then you have to unbox the Try and the exception. Considering how often I moan about stringly typed systems and how all strings should be wrapped by something I suppose an exception is good enough for things like this..

Disregard me.

src/main/scala/cromwell/engine/backend/jes/JesBackend.scala
geoffjentry: Disregard me here as well

src/main/scala/cromwell/engine/WorkflowDescriptor.scala
Horneth: I remember having an issue without it - `toString` is overriden in `WorkflowId` to return the `UUID.toString`

src/main/scala/cromwell/engine/backend/jes/JesBackendCall.scala
Horneth: The fact that it's lazy or that it's "big" for a val ? Or something else ?

src/main/scala/cromwell/engine/WorkflowDescriptor.scala
geoffjentry: Discussed offlist

cjllanwarne:
For record-keeping, should we add a `preemptableVM` booloean to JES_JOB which is true if the attempt was preemptable, and false otherwise?

src/main/migrations/changesets/add_attempt_in_execution.xml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.3.xsd">
+
+    <changeSet author="tjeandet" id="add-attempt-in-execution">
+        <addColumn
+                tableName="EXECUTION">
+            <column name="ATTEMPT" type="INT" />
```
cjllanwarne: Add to metadata endpoint

src/main/scala/cromwell/engine/WorkflowDescriptor.scala
```diff
@@ -61,16 +62,12 @@ case class WorkflowDescriptor(id: WorkflowId,
   lazy val fileHasher: FileHasher = { wdlFile: WdlFile => SymbolHash(ioManager.hash(wdlFile.value)) }
 
   // GCS FS with the workflow working directory as root
-  val gcsFilesystem = for {
-    interface <- gcsInterface
-    fs <- Try(GcsFileSystem.instance(interface, wfContext.root))
-  } yield fs
+  lazy val gcsFilesystem = Try(GcsFileSystem.instance(gcsInterface, wfContext.root))
```
cjllanwarne: I don't understand - why does `WorkflowDescriptor` have a `GcsInterface`? Is `WorkflowDescriptor` never used on `LocalBackend`?

src/main/scala/cromwell/engine/backend/jes/JesAttributes.scala
```diff
@@ -17,7 +18,8 @@ object JesAttributes {
     "project",
     "baseExecutionBucket",
     "endpointUrl",
-    "maximumPollingInterval"
+    "maximumPollingInterval",
+    "preemptible"
```
cjllanwarne: table

src/main/scala/cromwell/engine/backend/jes/JesAttributes.scala
```diff
@@ -30,9 +32,10 @@ object JesAttributes {
     val project: ValidationNel[String, String] = jesConf.validateString("project")
     val executionBucket: ValidationNel[String, String] = jesConf.validateString("baseExecutionBucket")
     val endpointUrl: ValidationNel[String, URL] = jesConf.validateURL("endpointUrl")
+    val maximumPreemptionAttempts: Int = jesConf.getIntOption("preemptible").getOrElse(0)
 
     (project |@| executionBucket |@| endpointUrl) {
-      JesAttributes(_, _, _)
+      JesAttributes(_, _, _, maximumPreemptionAttempts)
```
cjllanwarne: Never used

src/main/scala/cromwell/engine/backend/jes/JesBackend.scala
```diff
@@ -58,9 +58,10 @@ object JesBackend {
   val AuthFilePathOptionKey = "auth_bucket"
   val WriteToCacheOptionKey = "write_to_cache"
   val ReadFromCacheOptionKey = "read_from_cache"
+  val MaxPreemptionTriesOptionKey = "preemptible"
```
cjllanwarne: table

mcovarr: This is the correct spelling per `gcloud`

src/main/scala/cromwell/engine/backend/jes/JesBackend.scala
```diff
@@ -177,11 +167,9 @@ case class JesPendingExecutionHandle(backendCall: JesBackendCall,
                                      run: Run,
                                      previousStatus: Option[RunStatus]) extends ExecutionHandle {
   override val isDone = false
-
-  override def result = FailedExecution(new IllegalStateException)
+  override val result = NonRetryableExecution(new IllegalStateException)
```
cjllanwarne: I imagine this would be better as `override val result = throw new NotImplementedException("BAD")` since it should be a programmer error to reach here.

As it is, if we ever reach here, the execution handle reports that it failed in a non-retryable way which isn't really accurate

src/main/scala/cromwell/engine/WorkflowDescriptor.scala
Horneth: GCS is a separate thing from JES. You can use GCS on local backend.

src/main/scala/cromwell/engine/backend/jes/JesAttributes.scala
Horneth: Hmm I thought we said we were going to use ible since that's what Google uses in their API ?

src/main/scala/cromwell/engine/backend/BackendCall.scala
```diff
@@ -63,9 +63,14 @@ final case class SuccessfulExecutionHandle(outputs: CallOutputs, events: Seq[Exe
   override val result = SuccessfulBackendCallExecution(outputs, events, returnCode, hash, resultsClonedFrom)
 }
 
-final case class FailedExecutionHandle(throwable: Throwable, returnCode: Option[Int] = None) extends ExecutionHandle {
+final case class FailedExecutionHandle(throwable: Throwable, returnCode: Option[Int] = None, events: Seq[ExecutionEventEntry] = Seq.empty) extends ExecutionHandle {
```
cjllanwarne: it'd be really cool if we added a `Preempted Run 1 Started` and `Preempted Run 2 Preempted` to the events list when we get preempted. So the preemption attempts also show up on the timing diagrams...

Maybe a separate ticket?

src/main/scala/cromwell/engine/backend/jes/JesBackendCall.scala
```diff
@@ -48,22 +47,23 @@ class JesBackendCall(val backend: JesBackend,
 
   def jesCommandLine = s"/bin/bash ${cmdInput.local}"
 
-  val callGcsPath = JesBackend.callGcsPath(workflowDescriptor, key)
-  val callDir = GcsPath(callGcsPath)
-  val gcsExecPath = GcsPath(callGcsPath + "/" + JesExecScript)
-  val defaultMonitoringOutputPath = callGcsPath + "/" + JesMonitoringLogFile
+  lazy val callGcsPath = key.callRootPathWithBaseRoot(workflowDescriptor, backend.rootPath(workflowDescriptor.workflowOptions))
+  //TODO: Switch to NioGcsPath
```
cjllanwarne: TODONE?

src/main/scala/cromwell/engine/backend/jes/JesBackendCall.scala
cjllanwarne: As a heads up, the default runtime attributes in workflow options (coming soon, I promise!) will make the preemption option in the workflow options redundant.

src/main/scala/cromwell/engine/backend/jes/JesBackendCall.scala
```diff
@@ -72,6 +72,23 @@ class JesBackendCall(val backend: JesBackend,
   def downloadRcFile = authenticateAsUser(workflowDescriptor) { storage => Try(storage.readFile(returnCodeGcsPath)) }
 
   /**
+    * Determines the maximum number of times a call can be started with a Preemptible VM.
+    * Runtime attributes > workflow options > Jes configuration
+    */
+  lazy val maxPreemption = {
+    runtimeAttributes.preemptible getOrElse {
```
cjllanwarne: table

src/main/scala/cromwell/engine/backend/jes/JesRuntimeInfo.scala
```diff
@@ -27,10 +20,33 @@ object JesRuntimeInfo {
...
       .setRamGb(runtimeAttributes.memoryGB)
       .setCpu(runtimeAttributes.cpu)
-      .setPreemptible(runtimeAttributes.preemptible)
       .setZones(runtimeAttributes.defaultZones.asJava)
       .setDisks(runtimeAttributes.defaultDisks.asJava)
   }
 }
 
-case class JesRuntimeInfo(resources: Resources, docker: DockerExecutor)
+object NonPreemptibleJesRuntimeInfo extends JesRuntimeInfoBuilder {
```
cjllanwarne: table

src/main/scala/cromwell/engine/backend/jes/JesBackend.scala
Horneth: I kind of like the `IllegalStateException` because it describes accurately what happens if we get there. What about just `override val result = throw new IllegalStateException("JesPendingExecutionHandle cannot yield a result")` ?

src/main/scala/cromwell/engine/backend/jes/JesRuntimeInfo.scala
```diff
@@ -27,10 +20,33 @@ object JesRuntimeInfo {
...
+  def apply(commandLine: String, runtimeAttributes: CromwellRuntimeAttributes): JesRuntimeInfo = {
+    // See comment above
+    val dockerImage = runtimeAttributes.docker.get
+    val resources = buildResources(runtimeAttributes).setPreemptible(true)
+    new PreemptibleJesRuntimeInfo(resources, buildDockerExecutor(commandLine, dockerImage))
+  }
+}
+
+case class NonPreemptibleJesRuntimeInfo(resources: Resources, docker: DockerExecutor) extends JesRuntimeInfo
+case class PreemptibleJesRuntimeInfo(resources: Resources, docker: DockerExecutor) extends JesRuntimeInfo
```
cjllanwarne: I'm curious - why not just one class `JesRuntimeInfo(resources: Resources, docker: DockerExecutor, preemptable: Int)`?

src/main/scala/cromwell/engine/backend/jes/JesAttributes.scala
Horneth: `maximumPreemptionAttempts` ? It is used in `JesBackendCall`

src/main/migrations/changesets/add_attempt_in_execution.xml
Horneth: Ok

src/main/scala/cromwell/engine/backend/jes/Pipeline.scala
```diff
@@ -18,6 +17,7 @@ object Pipeline {
             key: BackendCallKey,
             runtimeAttributes: CromwellRuntimeAttributes,
             jesParameters: Seq[JesParameter],
+            preemptible: Boolean,
```
cjllanwarne: table

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
```diff
@@ -17,7 +16,7 @@ case class CromwellRuntimeAttributes(docker: Option[String],
                                      failOnStderr: Boolean,
                                      continueOnReturnCode: ContinueOnReturnCode,
                                      cpu: Long,
-                                     preemptible: Boolean,
+                                     preemptible: Option[Int],
```
cjllanwarne: a

src/main/scala/cromwell/engine/backend/BackendCall.scala
mcovarr: separate ticket would be good

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
```diff
@@ -56,7 +55,7 @@ object CromwellRuntimeAttributes {
     val FailOnStderr = false
     val ContinueOnReturnCode = ContinueOnReturnCodeSet(Set(0))
     val Memory = 2.0
-    val Preemptible = false
+    val Preemptible = 0
```
cjllanwarne: In fact... just search/replace them all :)

src/main/scala/cromwell/engine/backend/jes/JesRuntimeInfo.scala
Horneth: `JesRuntimeInfo` doesn't really care about the number of times we want to try preemptible VMs. At this point it's a Boolean (preempt or don't).
It was `JesRuntimeInfo(resources: Resources, docker: DockerExecutor, preemptable: Boolean)` before but @jg asked a case classification.

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -440,7 +470,10 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
     case Event(message @ CallCompleted(collectorKey: CollectorKey, _, _, _, _, _), _) =>
       resendDueToPendingExecutionWrites(message)
       stay()
-    case Event(message @ CallFailed(callKey, returnCode, failure), data) if data.isPersistedRunning(callKey) =>
+    case Event(message @ CallFailedRetryable(callKey, events, returnCode, failure), data) if data.isPersistedRunning(callKey) =>
+      // Note: Currently Retyrable == Preempted. If another reason than preemption arises that requires retrying at this level, this will need to be updated
```
mcovarr: Retyrable => Retryable

src/main/scala/cromwell/util/IntUtil.scala
```diff
@@ -0,0 +1,18 @@
+package cromwell.util
+
+object IntUtil {
+
+  val OrdinalSuffixes = List("th", "st", "nd", "rd", "th", "th", "th", "th", "th", "th")
```
mcovarr: hey this looks familiar :smile:

src/main/scala/cromwell/engine/backend/jes/JesBackendCall.scala
Horneth: `callGcsPath` is a NioGcsPath but all the other attributes below are not
```scala
lazy val callDir = GcsPath(callGcsPath.toString)
lazy val gcsExecPath = GcsPath(callGcsPath + "/" + JesExecScript)
...
```

src/main/scala/cromwell/engine/backend/BackendCall.scala
Horneth: I can look into it as part of this ticket if it's not too hard to add

src/main/scala/cromwell/util/IntUtil.scala
Horneth: Yes, I was waiting for the lenthall PR to be merged and switch to the newest snapshot

src/main/scala/cromwell/engine/io/gcs/GcsFileSystem.scala
```diff
@@ -9,12 +9,13 @@ import java.util.{Collections, Set => JSet}
 import cromwell.engine.PathString
 
 import scala.concurrent.ExecutionContext
+import scala.util.Try
 
 object GcsFileSystem {
   import PathString._
-  def instance(interface: GoogleCloudStorage, root: String)(implicit executionContext: ExecutionContext) = {
+  def instance(interface: Try[GoogleCloudStorage], root: String)(implicit executionContext: ExecutionContext) = {
```
cjllanwarne: IMO, `instance` implies this returns a singleton instance whereas this is creating a whole new one each time

Horneth: Yes this should probably be an `apply`

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -694,7 +732,7 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
         case _ => Nil
       }
       val dependencies = upstream.flatten ++ downstream
-      val dependenciesResolved = dependencies.isEmpty || dependencies.forall(isDone)
+      val dependenciesResolved = dependencies.isEmpty || dependencies.forall(d => isDone(d) || isPreempted(d))
```
cjllanwarne: This doesn't look right to me - if a call gets preempted then any calls which depend on it can now start?

cjllanwarne: Also, as a sidenote on this line, I'm pretty sure `forall` covers the `.isEmpty` case already

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -960,7 +998,7 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
     }
   }
 
-  private def isWorkflowDone: Boolean = executionStore forall isDone
+  private def isWorkflowDone: Boolean = executionStore forall { d => isDone(d) || isPreempted(d) }
```
cjllanwarne: Does preempted not count as Done?

src/main/scala/cromwell/util/IntUtil.scala
```diff
@@ -0,0 +1,18 @@
...
+
+  implicit class IntEnhanced(val value: Int) extends AnyVal {
+    def toOrdinal: String = value match {
+      case v if v.isBetweenInclusive(10, 20) => s"${v}th"
```
cjllanwarne: Hmm, I think `if (v % 100).isBetweenInclusive(10,20)` to cover 111th, 100011th, etc

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
Horneth: Yes and no, when a call gets preempted we immediately re-insert a clone of this call with the `NotStarted` Status (and attempt incremented) before calling `tryStartingRunnableCalls` again. Because the copy has the same `Call` object, calls that were depending on the preempted call will be depending on the new clone so they won't start yet.
A comment could be useful though to make this clearer.

src/main/scala/cromwell/util/google/GoogleCredentialFactory.scala
```diff
@@ -96,11 +96,15 @@ abstract class GoogleCredentialFactory {
   }
 
   private def forServiceAccount(serviceAccountMode: ServiceAccountMode): Credential = {
+    val pemFile = new File(serviceAccountMode.pemPath)
+    if (!pemFile.exists()) {
+      throw new FileNotFoundException(s"Pem file ${Paths.get(serviceAccountMode.pemPath).toAbsolutePath} does not exist")
+    }
```
cjllanwarne: Is this related to preemption?

src/test/scala/cromwell/engine/backend/BackendCallSpec.scala
cjllanwarne: I can't remember whether we got this far on customising the Call Caching algorithm but if we did, preemptable should be a runtime option which is NOT included as part of the call caching algorithm

src/main/scala/cromwell/util/google/GoogleCredentialFactory.scala
Horneth: No when I was trying to write a test for preemption on JES I ran into some not very clear exceptions so I added some more context to some of them.

src/main/scala/cromwell/engine/WorkflowDescriptor.scala
geoffjentry: On a similar note, what happens when there is S3, azure blob storage, etc?

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
Horneth: I had it like you say originally but I found it confusing having `isDone` return true for something else than `CallStatus.Done`. I don't really mind though.

src/main/scala/cromwell/engine/backend/jes/JesRuntimeInfo.scala
geoffjentry: IIRC there was some value which could then be hardcoded in one of the cases but not the other but that doesn't seem to be true here

src/main/scala/cromwell/engine/backend/BackendCall.scala
cjllanwarne: I think as-is it should work already, it'll just be on a separate row... as long as all of the failed attempts still show up in metadata, we should be fine.

src/main/scala/cromwell/util/IntUtil.scala
Horneth: This has been moved to lenthall but yes nice catch I'll make a new PR

src/test/scala/cromwell/engine/backend/BackendCallSpec.scala
Horneth: As I see it in the code all runtime attributes are included in the hash :
```scala
val orderedRuntime = Seq(
      ("docker", dockerHash getOrElse ""),
      ("defaultZones", runtimeAttributes.defaultZones.sorted.mkString(",")),
      ("failOnStderr", runtimeAttributes.failOnStderr.toString),
      ("continueOnReturnCode", runtimeAttributes.continueOnReturnCode match {
        case ContinueOnReturnCodeFlag(bool) => bool.toString
        case ContinueOnReturnCodeSet(codes) => codes.toList.sorted.mkString(",")
      }),
      ("cpu", runtimeAttributes.cpu.toString),
      ("preemptible", runtimeAttributes.preemptible.toString),
      ("defaultDisks", runtimeAttributes.defaultDisks.sortWith((l, r) => l.getName > r.getName).map(d => s"${d.getName} ${d.size} ${d.getType}").mkString(",")),
      ("memoryGB", runtimeAttributes.memoryGB.toString)
    )
```
I think we said we were going to tune this later but never got to that point yet it seems

src/main/scala/cromwell/engine/WorkflowDescriptor.scala
Horneth: I know I'm not saying this is perfect, it was a first step towards differentiating between `Backend`s and `FileSystem`s. We can definitely codify this better, however because we allow workflow options to be passed containing refresh tokens and so on, each workflow potentially has different ways enabled to communicate with different FileSystems, which is why this ended up in the WorkflowDescriptor.

geoffjentry: :+1:

src/main/scala/cromwell/engine/backend/BackendCall.scala
Horneth: Will do. I noticed that events are not written in the database for a Call that fails the workflow though (NonRetryable failure)

src/main/scala/cromwell/util/IntUtil.scala
cjllanwarne: Will it be removed from Cromwell in that case? It'd be nice to do that before merging a duplicate

cjllanwarne:
:+1:

README.md
```diff
@@ -976,6 +979,7 @@ Valid keys and their meanings:
 * **refresh_token** - (JES backend only) Only used if `localizeWithRefreshToken` is specified in the [configuration file](#configuring-cromwell).  See the [Data Localization](#data-localization) section below for more details.
 * **auth_bucket** - (JES backend only) defaults to the the value in **jes_gcs_root**.  This should represent a GCS URL that only Cromwell can write to.  The Cromwell account is determined by the `google.authScheme` (and the corresponding `google.userAuth` and `google.serviceAuth`)
 * **monitoring_script** - (JES backend only) Specifies a GCS URL to a script that will be invoked prior to the WDL command being run.  For example, if the value for monitoring_script is "gs://bucket/script.sh", it will be invoked as `./script.sh > monitoring.log &`.  The value `monitoring.log` file will be automatically de-localized.
+* **preemptible** - (JES backend only) Specifies the maximum number of times a call should be executed with a preemptible VM. This option can be overridden by [runtime attributes](#preemptible).
```
scottfrazer: Maybe mention what the default value is?

src/main/scala/cromwell/engine/backend/jes/JesRuntimeInfo.scala
```diff
@@ -27,10 +20,33 @@ object JesRuntimeInfo {
...
+}
+
+object PreemptibleJesRuntimeInfo extends JesRuntimeInfoBuilder {
+  def apply(commandLine: String, runtimeAttributes: CromwellRuntimeAttributes): JesRuntimeInfo = {
+    // See comment above
+    val dockerImage = runtimeAttributes.docker.get
+    val resources = buildResources(runtimeAttributes).setPreemptible(true)
+    new PreemptibleJesRuntimeInfo(resources, buildDockerExecutor(commandLine, dockerImage))
+  }
+}
```
scottfrazer: It does kind of feel like there's some duplicated code here where the only actual difference is `setPreemptible(false)`

cjllanwarne:
Please mention the change (and the update to metadata format) in the CHANGELOG.MD

src/main/scala/cromwell/engine/backend/jes/JesRuntimeInfo.scala
scottfrazer: Also I'm really not a fan of the the two case classes + two companion objects + duplicated code for expressing the fact that `JesRuntimeInfo` can have a `preemptible: Boolean`.  But at the risk of engaging in fisticuffs with @geoffjentry, I'm  not going to fight that fight

scottfrazer:
:+1:

src/main/scala/cromwell/engine/backend/jes/JesRuntimeInfo.scala
geoffjentry: No, I totally agree w/ you - although we likely disagree on the resolution.

IMO the way to go is to make the apply on the companion object of the base trait which then creates and returns the correct one.

geoffjentry:
:+1:

src/main/scala/cromwell/engine/backend/jes/JesRuntimeInfo.scala
Horneth: But then you'll need a `preemptible: Boolean` in the apply signature to know which one to create so at this point why not just have the boolean be an attribute of `JesRuntimeInfo` ?

geoffjentry: Sure, but the object it creates you can compute on based on type. It's not really a big deal either way

PR #413, opened 2016-01-30
WorkflowManagerActor is finally FSM'd.

mcovarr:
Basic structure should be okay but I haven't finished testing it.


mcovarr:
The wheel has decided it's @geoffjentry's and @scottfrazer's lucky day

src/main/scala/cromwell/engine/workflow/SingleWorkflowRunnerActor.scala
```diff
@@ -79,18 +85,19 @@ case class SingleWorkflowRunnerActor(source: WorkflowSourceFiles,
...
+          workflowManager ! WorkflowAbort(id)
+          log.info(s"$tag: Waiting for workflow $id to abort...")
+          Await.ready(donePromise.future, Duration.Inf)
+        }
       }
       stay using data.copy(id = Option(id))
+    case Event(Transition(_, _, WorkflowAborted), data) =>
+      // The .get is safe because the shutdown hook is only registered when the workflow ID
+      // has been set in the state data.
+      log.info(s"$tag: Workflow ${data.id.get} aborted.")
```
mcovarr: that might benefit from an ADTing

src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -147,32 +147,94 @@ class WorkflowManagerActor(backend: Backend) extends Actor with CromwellActor {
...
+      workflowStore.toMap.values.foreach { _ ! WorkflowActor.AbortWorkflow }
+      goto(Aborting)
+  }
+
+  when (Aborting) {
+    case Event(Transition(workflowActor, _, toState: WorkflowState), _) if toState.isTerminal =>
+      // Remove this terminal actor from the workflowStore and log a progress message.
+      removeTerminalWorkflow(workflowActor)
+      logger.info(s"$tag: Waiting for all workflows to abort (${workflowStore.size} remaining).")
+      // If there are no more workflows to abort we're done, otherwise just stay in the current state.
```
mcovarr: The comment is misleading as written, abort messages were already sent to all running workflows.

src/main/scala/cromwell/engine/workflow/SingleWorkflowRunnerActor.scala
```diff
@@ -79,18 +85,19 @@ case class SingleWorkflowRunnerActor(source: WorkflowSourceFiles,
...
+      stay
```
mcovarr: parens

geoffjentry: Why are you nitpicking yourself? :)

mcovarr: Think of these as post its so I don't forget :smile:

src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -47,12 +47,21 @@ object WorkflowManagerActor {
...
   // restart 500 workflows at exactly the same time crushes the database connection pool.
   lazy val RestartDelay = 200 milliseconds
+
+  sealed trait WorkflowManagerState
+  case object Running extends WorkflowManagerState
+  case object Aborting extends WorkflowManagerState
+  case object Done extends WorkflowManagerState
+
+  sealed trait WorkflowManagerData
+  case object NoData extends WorkflowManagerData
```
mcovarr: The state data should include the workflow store, at which time the write-once class can be deleted.

src/main/scala/cromwell/engine/workflow/SingleWorkflowRunnerActor.scala
geoffjentry: How were you picturing it?

src/main/scala/cromwell/engine/workflow/SingleWorkflowRunnerActor.scala
geoffjentry: Consider it done.

src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -62,32 +71,23 @@ object WorkflowManagerActor {
...
-          numRemaining = Await.result(result,Duration.Inf).size
-          log.info(s"$tag: Waiting for all workflows to abort ($numRemaining remaining).")
-        }
-        log.info(s"$tag: All workflows aborted.")
-      }
-    })
+  private val donePromise = Promise[Unit]()
+
+  if (abortJobsOnTerminate) {
+    sys.addShutdownHook {
```
geoffjentry: why is this here and on the single workflow actor?

mcovarr: These are the shutdown hooks our friends added to support abort on SIGINT.  I did a bit of cleanup related to that in SWRA.  Unfortunately WMA was too crufty to take a similarly minimal change, which motivated the FSM'ing.  This and the CallExecutionActor were the only major actors not to be FSM'd, though @Horneth is already contemplating FSM'ing CAE for reasons unrelated to shutdown hooks.  :smile:

src/main/scala/cromwell/engine/workflow/SingleWorkflowRunnerActor.scala
mcovarr: With more state data types and without so many `Option`s :smile:  Haven't thought it through any more than that yet.

src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
geoffjentry: What I meant was why do you need both hooks? The one in the single actor sends WMA a signal to abort its WF and the one here aborts all workflows. If this is just for run (i forget) wouldn't this be redundant and if this is for both run/server wouldn't the other be redundant?

mcovarr: Ah gotcha.  I hadn't considered that, I think you're right we may be able to get away with only one hook in WMA.  SWRA's `WorkflowAborted` handler would need to be modified to allow for that, but I think that should be done anyway as the current SWRA abort handler looks wrong (it notes that the only workflow it cares about has been aborted and then tells no other actors and hangs out forever doing nothing).

src/main/scala/cromwell/util/WriteOnceStore.scala
```diff
@@ -21,4 +21,21 @@ class WriteOnceStore[K, V] {
...
   def toMap: Map[K, V] = writeOnceStore.readOnlySnapshot().toMap
-}
\ No newline at end of file
+
+  /**
+    * Remove an entry in the store as specified by key.
+    * @param key Key of the entry to remove.
+    * @return `Some(value)` if the entry is present in the store, otherwise `None`.
+    */
+  def remove(key: K): Option[V] = writeOnceStore.remove(key)
```
cjllanwarne: Something seems strange about making a WriteOnce class which lets you update its values...? Is it more intellectually honest to call this a NoDuplicatesStore?

Come to think of it... isn't this just a mutable Set?

geoffjentry: I forgot about this (and honestly, I forgot this class was here). You're right, this really bends the spirit of what it is supposed to be, it was supposed to literally be write once per value and that's it

src/main/scala/cromwell/engine/workflow/SingleWorkflowRunnerActor.scala
```diff
@@ -79,18 +85,19 @@ case class SingleWorkflowRunnerActor(source: WorkflowSourceFiles,
...
-            while(stateName != Done)
-              Thread.sleep(1000)
-            log.info(s"$tag: Workflow $id aborted.")
-          }
-        })
+      if (abortJobsOnTerminate) {
+        sys.addShutdownHook {
+          workflowManager ! WorkflowAbort(id)
+          log.info(s"$tag: Waiting for workflow $id to abort...")
+          Await.ready(donePromise.future, Duration.Inf)
```
cjllanwarne: This donePromise seems incredibly funky to me. 

It's basically used like a boolean var, but hidden in a Promise to avoid offending FP sensibilities?

src/main/scala/cromwell/engine/workflow/SingleWorkflowRunnerActor.scala
```diff
@@ -79,18 +85,19 @@ case class SingleWorkflowRunnerActor(source: WorkflowSourceFiles,
...
+        }
```
cjllanwarne: Why not move the `if` block up into the `CromwellActor` (right below the configuration reading) and just override a `customShutdownHook` method here?

src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -11,8 +11,8 @@ import cromwell.engine.db.DataAccess._
 import cromwell.engine.db.ExecutionDatabaseKey
 import cromwell.engine.db.slick._
 import cromwell.engine.workflow.WorkflowActor.{Restart, Start}
+import cromwell.engine.workflow.WorkflowManagerActor.{WorkflowManagerData, WorkflowManagerState}
```
cjllanwarne: I guess because the Object methods are used in the Class? 

Has anyone ever opined on whether its better to have these all the way up here, or as a line in the class definition?

src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -62,32 +71,23 @@ object WorkflowManagerActor {
  * WorkflowOutputs: Returns a `Future[Option[binding.WorkflowOutputs]]` aka `Future[Option[Map[String, WdlValue]]`
  *
  */
-class WorkflowManagerActor(backend: Backend) extends Actor with CromwellActor {
+class WorkflowManagerActor(backend: Backend) extends LoggingFSM[WorkflowManagerState, WorkflowManagerData] with CromwellActor {
   import WorkflowManagerActor._
```
cjllanwarne: Oh wait, you ARE importing it here anyway!

src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
geoffjentry: I personally strongly prefer having just about everything up at the top unless there's a good reason not to (e.g. pulling a specific implicit into scope)

src/main/scala/cromwell/util/WriteOnceStore.scala
cjllanwarne: I actually just saw a TODO higher up this PR saying "... and this lets us remove the WriteOnceStore!" so hopefully this is about to disappear?

src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
cjllanwarne: woohoo!

src/main/scala/cromwell/engine/workflow/SingleWorkflowRunnerActor.scala
scottfrazer: I think it is structured this way because the Promise will be fulfilled asynchronously and there's no way to say "wait for a boolean var to become `true`"?

cjllanwarne: I guess there's a reason why `Await.ready(..., Duration.Inf)` is better than `while(!boolVar) sleep(1)` - I'd be happy enough to hear "behind-the-scenes-FP-magic, don't worry about it"

scottfrazer: If I understand correctly, `sleep(1)` doesn't actually yield control of the thread in Akka-land, whereas the `Await.ready()` would.  I do agree with you that it is funky looking, though.

geoffjentry: @cjllanwarne I'm probably missing a variety of other bits but one thing is that `Await.ready` calls `blocking {}` under the hood, which gives a heads up to your execution context that danger is incoming so it can try to manage things a bit better. You also might get better granularity than in your polling loop.

Another angle which seems trivial but I'd argue it's very much not is that it makes it more clear that blocking is happening. `Await` pops out like a giant red alarm bell, at least to me.

Iff I haven't missed anything key the differences would be relatively minor.

@scottfrazer I believe both `Awaits` really do block the thread, but they should both have that `blocking` keyword in there (nothing prevents someone from doing that themselves) which helps make things more sane. That's part of why they're also viewed with suspicion.

scottfrazer: Hmm I thought something was up... but it's very unlike @mcovarr to write blocking code so I thought I wasn't really understanding something

mcovarr: As far as I understand the contract (and consistent with what I've observed), the JVM tries to execute all shutdown hooks and then exit.  So this actually needs to block, otherwise the JVM would shut down as soon as the code here finished without waiting for workflow(s) to abort.  I'll add a comment to this effect.

src/main/scala/cromwell/util/WriteOnceStore.scala
mcovarr: yup

src/main/scala/cromwell/engine/workflow/SingleWorkflowRunnerActor.scala
scottfrazer: Ah that makes a lot more sense

src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
mcovarr: These two are actually used as type variables in the *definition* of the class.  But I can `_` this and get rid of the other import below.

mcovarr:
@dgtester just a heads up that I did some refactoring in the area of your recent PR that might at first look like I deleted your work, but actually that logic has been consolidated to `WorkflowManagerActor`.  I've tested that SIGINT still works in both "run" and "server" use cases, but please let me know if you have any questions or concerns.

src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -47,12 +48,33 @@ object WorkflowManagerActor {
...
   // How long to delay between restarting each workflow that needs to be restarted.  Attempting to
   // restart 500 workflows at exactly the same time crushes the database connection pool.
   lazy val RestartDelay = 200 milliseconds
+
+  sealed trait WorkflowManagerState
+  case object Running extends WorkflowManagerState
+  case object Aborting extends WorkflowManagerState
+  case object Done extends WorkflowManagerState
+
+  type WorkflowActorRef = ActorRef
```
geoffjentry: What is this buying you?

src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -320,28 +392,38 @@ class WorkflowManagerActor(backend: Backend) extends Actor with CromwellActor {
...
+    */
+  private def submitWorkflow(source: WorkflowSourceFiles, replyTo: Option[ActorRef], id: Option[WorkflowId]): WorkflowManagerData = {
     val workflowId: WorkflowId = id.getOrElse(WorkflowId.randomId())
-    log.info(s"$tag submitWorkflow input id = $id, effective id = $workflowId")
+    logger.info(s"$tag submitWorkflow input id = $id, effective id = $workflowId")
     val isRestart = id.isDefined
 
-    val workflowActor = for {
-      descriptor <- Future.fromTry(Try(WorkflowDescriptor(workflowId, source)))
+    case class ActorAndStateData(actor: WorkflowActorRef, data: WorkflowManagerData)
```
geoffjentry: w00t class and not tuple!

geoffjentry:
:+1:

src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
mcovarr: Preexisting, I just moved it up a ways.  Arguably more beauteous.

geoffjentry: I don't really care BTW, just so long as you think it's providing value

mcovarr: I don't think it adds a ton of value, but it was already here so (shrug)

geoffjentry: It's not hard to remove :)

scottfrazer:
:+1:

PR #420, opened 2016-02-03
Runtime attributes are interpreted as WDL expressions

scottfrazer:
See corresponding wdl4s changes: https://github.com/broadinstitute/wdl4s/pull/9

We can now do stuff like this

```
task test {
  String tag
  Int disk_gb
  Int memory_gb

  command { ps }
  runtime {
    docker: "ubuntu:" + tag
    defaultDisks: "Disk1 " + disk_gb + " SSD"
    memory: memory_gb + " GB"
  }
}

workflow w {
  call test
}
```


cjllanwarne:
The wheel has spoken: @geoffjentry @Horneth

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
```diff
@@ -24,156 +27,159 @@ case class CromwellRuntimeAttributes(docker: Option[String],
 object CromwellRuntimeAttributes {
   private val log = LoggerFactory.getLogger("RuntimeAttributes")
 
-  def apply(wdlRuntimeAttributes: RuntimeAttributes, backendType: BackendType): CromwellRuntimeAttributes = {
-    val attributeMap = AttributeMap(wdlRuntimeAttributes.attrs)
+  def apply(wdlRuntimeAttributes: RuntimeAttributes, backendCall: BackendCall): Try[CromwellRuntimeAttributes] = {
```
geoffjentry: While one can define them to return anything one wants I think it's odd for an `apply` not to have the return type be the actual type

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
```diff
@@ -24,156 +27,159 @@ case class CromwellRuntimeAttributes(docker: Option[String],
...
+    for {
+      keys <- validateKeys(wdlRuntimeAttributes, backendCall.backend.backendType)
+      attributes <- TryUtil.sequenceMap(wdlRuntimeAttributes.evaluate(backendCall.lookupFunction(backendCall.locallyQualifiedInputs), backendCall.engineFunctions))
+      validatedAttributes <- validateRuntimeAttributes(attributes)
+    } yield validatedAttributes
+  }
+
+  def validateKeys(wdlRuntimeAttributes: RuntimeAttributes, backendType: BackendType): Try[RuntimeAttributes] = {
```
geoffjentry: Why convert the ErrorOr to Try?

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
```diff
@@ -24,156 +27,159 @@ case class CromwellRuntimeAttributes(docker: Option[String],
...
+      case scalaz.Success(x) => scala.util.Success(wdlRuntimeAttributes)
+      case scalaz.Failure(e) => scala.util.Failure(new IllegalArgumentException(e.list.mkString("\n")))
     }
   }
 
+  def unsupportedKeys(keys: Iterable[String], backendType: BackendType): Seq[String] = {
+    val supportedKeys = backendType.supportedKeys map { _.key }
+    val unsupportedKeys = keys.toSet -- supportedKeys
+
+    if (unsupportedKeys.isEmpty) Vector.empty
```
geoffjentry: Might as well use List. I was the only jackass pushing Vector and for what turned out to be incorrect reasons

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
```diff
@@ -24,156 +27,159 @@ case class CromwellRuntimeAttributes(docker: Option[String],
...
+      "defaultDisks" -> WdlString(LocalizationDisk.toString),
+      "defaultZones" -> WdlString("us-central1-a"),
+      "continueOnReturnCode" -> WdlInteger(0),
+      "failOnStderr" -> WdlBoolean.False,
+      "preemptible" -> WdlBoolean.False,
+      "memory" -> WdlString("2GB")
+    )).get
+  }
+
+  private def validateRuntimeAttributes(attributes: Map[String, WdlValue]): Try[CromwellRuntimeAttributes] = {
```
geoffjentry: ErrorOr. This isn't a public facing function so debate on whether or not a library API should throw or Try doesn't apply here

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
```diff
@@ -24,156 +27,159 @@ case class CromwellRuntimeAttributes(docker: Option[String],
...
+      new CromwellRuntimeAttributes(_, _, _, _, _, _, _, _)
+    } match {
+      case Success(x) => scala.util.Success(x)
+      case Failure(nel) => scala.util.Failure(new IllegalArgumentException(nel.list.mkString("\n")))
     }
   }
 
-  private def validateCpu(value: Option[String]): ValidationNel[String, Long] = {
-    value map validateLong getOrElse Defaults.Cpu.successNel
+  private def validateCpu(cpu: Option[WdlValue]): ValidationNel[String, Int] = {
```
geoffjentry: I recognize it's preexisting but could you change ValidationNel to ErrorOr (throughout the PR)

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
```diff
@@ -24,156 +27,159 @@ case class CromwellRuntimeAttributes(docker: Option[String],
...
-    value map validateBoolean getOrElse Defaults.Preemptible.successNel
+  private def validatePreemptible(preemptible: Option[WdlValue]): ValidationNel[String, Boolean] = {
+    preemptible.map(validateBoolean).getOrElse(defaults.preemptible.successNel)
   }
 
-  private def validateFailOnStderr(value: Option[String]): ValidationNel[String, Boolean] = {
-    value map validateBoolean getOrElse Defaults.FailOnStderr.successNel
+  private def validateDocker(docker: Option[WdlValue]): ValidationNel[String, Option[String]] = {
+    docker match {
+      case Some(WdlString(s)) => Some(s).successNel
```
geoffjentry: Option(s)

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
scottfrazer: How would you structure the `apply()` method if this returned an `ErrorOr`?  I could either:

* Change the return type of `RuntimeAttributes.evaluate` to return an `ErrorOr` (but then wdl4s returns scalaz objects from its public API).  This also conflicts with your other suggestion of throwing an exception from that method.
* Convert the return value of `RuntimeAttributes.evaluate` to an `ErrorOr` in `apply`

Any other suggestions?

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
```diff
@@ -24,156 +27,159 @@ case class CromwellRuntimeAttributes(docker: Option[String],
...
+      case scalaz.Failure(e) => scala.util.Failure(new IllegalArgumentException(e.list.mkString("\n")))
     }
   }
 
+  def unsupportedKeys(keys: Iterable[String], backendType: BackendType): Seq[String] = {
+    val supportedKeys = backendType.supportedKeys map { _.key }
+    val unsupportedKeys = keys.toSet -- supportedKeys
+
+    if (unsupportedKeys.isEmpty) Vector.empty
+    else Vector(s"Found unsupported keys for backend '$backendType': " + unsupportedKeys.toSeq.sorted.mkString(", "))
```
Horneth: It seems weird to have a method with a `Seq[String]` return type which never returns a Seq with more than one element. Why not just return `unsupportedKeys.toSeq.sorted` in this case ?

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
```diff
@@ -24,156 +27,159 @@ case class CromwellRuntimeAttributes(docker: Option[String],
...
+  val defaults: CromwellRuntimeAttributes = {
+    // .get below because we assume that the default values are valid
+    // There's a unit test in CromwellRuntimeAttributeSpec to test these defaults
+    validateRuntimeAttributes(Map(
+      "cpu" -> WdlInteger(1),
+      "defaultDisks" -> WdlString(LocalizationDisk.toString),
+      "defaultZones" -> WdlString("us-central1-a"),
+      "continueOnReturnCode" -> WdlInteger(0),
+      "failOnStderr" -> WdlBoolean.False,
+      "preemptible" -> WdlBoolean.False,
```
Horneth: preemptible is going to be an Int, but I don't think it impacts you I'll make the change when I'll rebase on top of this

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
scottfrazer: Honestly I'm not sure, I just moved this code around, I didn't actually modify it.  I'll look into it.

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
geoffjentry: I didn't really notice that these were public (but I still say public library API should throw). 

Appy should definitely be throwing, however

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
```diff
@@ -24,156 +28,161 @@ case class CromwellRuntimeAttributes(docker: Option[String],
...
-        }
+      *  Warn if keys are found that are unsupported on this backend.  This is not necessarily an error, at this point
+      *  the keys are known to be supported on at least one backend other than this.
+      */
+    unsupportedKeys(wdlRuntimeAttributes.attrs.keys, backendType) foreach log.warn
+    val requiredKeysNel = validateRequiredKeys(wdlRuntimeAttributes.attrs, backendType)
+    val unknownKeysNel = validateUnknownKeys(wdlRuntimeAttributes.attrs, backendType)
+    (requiredKeysNel |@| unknownKeysNel) { (_, _) } match {
+      case scalaz.Success(x) => scala.util.Success(wdlRuntimeAttributes)
+      case scalaz.Failure(e) => scala.util.Failure(new IllegalArgumentException(e.list.mkString("\n")))
```
Horneth: What about using `AggregatedException` or `ThrowableWithErrors` to keep the list as a String ? It will play better with the API too.
```scala
throw new IllegalArgumentException() with ThrowableWithErrors {
          val message = "RuntimeAttribute is not valid."
          val errors = e
        }
```

src/main/scala/cromwell/engine/backend/runtimeattributes/LocalDisk.scala
```diff
@@ -0,0 +1,51 @@
...
+  private def validateDiskType(diskTypeName: String): ValidationNel[String, DiskType] = {
+    DiskType.values().find(_.diskTypeName == diskTypeName) match {
+      case Some(diskType) => diskType.successNel[String]
+      case None =>
+        val diskTypeNames = DiskType.values.map(_.diskTypeName).mkString(", ")
+        s"Disk TYPE $diskTypeName should be one of $diskTypeNames".failureNel
+    }
+  }
+
+  private def validateLong(value: String): ValidationNel[String, Long] = {
```
Horneth: We don't have this validate method somewhere else already ? I think I've seen it before with a bunch of other ones for Boolean, Int etc... IMO this could even be something to put in lenthall.

cjllanwarne:
Please mention this in changelog.md

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
scottfrazer: Yeah, good idea

src/main/scala/cromwell/engine/backend/runtimeattributes/LocalDisk.scala
geoffjentry: It probably exists in 3 places now :'(

At this point it might make sense to set up a ticket explicitly to go through all of our repos and identify lenthall stuff + make the change. Perhaps have a ticket like that which triggers every N releases

scottfrazer: I didn't add this one, I just moved it

scottfrazer: oh actually, maybe I'm wrong... I removed one validateLong and added one... hmm... I guess that speaks to your point that this should be extracted somewhere at some point :)

Horneth:
:+1:

src/test/scala/cromwell/util/SampleWdl.scala
```diff
@@ -2,11 +2,12 @@ package cromwell.util
 
 import java.io.{File, FileWriter}
 import java.nio.file.{Files, Path}
+import java.util.UUID
 
 import wdl4s._
 import wdl4s.types.{WdlArrayType, WdlStringType}
 import wdl4s.values._
-import cromwell.engine.WorkflowSourceFiles
+import cromwell.engine.{WorkflowId, WorkflowDescriptor, WorkflowSourceFiles}
```
cjllanwarne: That's weird... how was this working before if these weren't imported?

src/test/scala/cromwell/engine/backend/CromwellRuntimeAttributeSpec.scala
```diff
@@ -1,72 +1,68 @@
...
-      |  call wc {
-      |    input: in_file=ps.procs
-      |  }
-      |}
-      |
-    """.stripMargin.replace(DockerValuePlaceholder, WdlValueForDockerAttribute)
 
-  val WorkflowWithoutRuntime =
-    """
+  object WorkflowWithStaticRuntime extends SampleWdl {
```
cjllanwarne: This should be in SampleWdl

src/test/scala/cromwell/engine/backend/CromwellRuntimeAttributeSpec.scala
```diff
@@ -341,241 +351,245 @@ object CromwellRuntimeAttributeSpec {
...
+    val attributes = runtimeAttributes(WorkflowWithRuntimeAttributeExpressions, "test", jesBackend)
+    attributes.memoryGB shouldEqual 7
+    attributes.disks shouldEqual Vector(
+      new Disk().setName("Disk1").setSizeGb(9L).setType("PERSISTENT_SSD").setAutoDelete(true),
+      CromwellRuntimeAttributes.LocalizationDisk.toDisk
+    )
+    attributes.docker shouldEqual Some("ubuntu:latest")
   }
 
-  "Default RuntimeAttributes" should "NOT override WDL values" in {
```
cjllanwarne: I can't see where this test has gone to?

src/test/scala/cromwell/engine/backend/CromwellRuntimeAttributeSpec.scala
```diff
@@ -341,241 +351,245 @@ object CromwellRuntimeAttributeSpec {
...
+  it should "reject a task with an invalid 'disks' parameter (3)" in {
     val ex = intercept[IllegalArgumentException] {
-      NamespaceWithWorkflow.load(WorkflowWithMessedUpDiskType).toCromwellRuntimeAttributes(BackendType.JES)
+      runtimeAttributes(WorkflowWithMessedUpDiskType, "messed_up_disk", jesBackend)
     }
-
     ex.getMessage should include("Disk TYPE SDD should be one of LOCAL, SSD, HDD")
   }
 
-  "Default RuntimeAttributes" should "be accessible for all tasks" in {
```
cjllanwarne: Nor this one

src/test/scala/cromwell/InvalidRuntimeAttributesSpec.scala
```diff
@@ -20,16 +20,40 @@ class InvalidRuntimeAttributesSpec extends CromwellTestkitSpec with BeforeAndAft
...
+    "fail on Local Backend" in {
       runWdl(
         sampleWdl = SampleWdl.HelloWorld,
-        runtime =
-          """ runtime { wrongAttribute: "nop" }""".stripMargin,
-        eventFilter = EventFilter.info(pattern = "transitioning from Running to Succeeded", occurrences = 1),
-        terminalState = WorkflowSucceeded
+        runtime = """ runtime { wrongAttribute: "nop" }""".stripMargin,
+        eventFilter = EventFilter.error(pattern = "Unknown keys found in runtime configuration: wrongAttribute", occurrences = 1),
+        terminalState = WorkflowFailed
```
cjllanwarne: I fixed this in my changes but it's regressed here. An unexpected runtime attribute should NOT fail the workflow - it should cause a warning message but the workflow should proceed successfully

src/test/scala/cromwell/engine/backend/CromwellRuntimeAttributeSpec.scala
scottfrazer: Yeah, very real possibility that some of the tests got removed.  It was a messy rebase.  I'll add these back

scottfrazer: This test should be covered by the `it should "allow workflow options to override defaults"` test.  If I understand correctly, this is testing that specifying workflow options should overwrite default values, specifically `docker`

scottfrazer: Though I suppose your test is different because it asserts it for *all* tasks.  I can alter the test below to test for all tasks in a WDL file instead of just one.

src/test/scala/cromwell/engine/backend/CromwellRuntimeAttributeSpec.scala
scottfrazer: `SampleWdl` is ~1700 lines of code now and moving these over there will be 2000+ lines of code... I thought they might be better here since they'll likely never be used elsewhere.  Any opinions @Horneth or @geoffjentry?

geoffjentry: I don't have a strong opinion. I'll throw out a few bits of my general philosophy which could apply here though (tl;dr - do whatever you want):

- I don't really care that much about code cleanliness of Spec files
- A file like SampleWdl which is intended to be a bundle of WDLs commonly used in tests I'm totally OK with being large. That said, perhaps some sort of subsetting (e.g. a package of sample WDLs, perhaps with some easy object which collates them or something, who knows)
- If this isn't duplicated it doesn't bother me having one off WDLs in their own tests. To me SampleWdl is mostly useful for reused workflows
- Even if it is duplicated I go back to my first bullet point. As long as this isn't a case where one would want to ensure that changing WDL foo affected everything using that specific WDL it's fine.

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
```diff
@@ -24,163 +29,187 @@ case class CromwellRuntimeAttributes(docker: Option[String],
...
 
-    val withDefaults = new AttributeMapTrait {
-      override def keys = attributesMap.keys ++ workflowOptions.getDefaultRuntimeOptionKeys
-      override def unsupportedKeys(bt: BackendType): Seq[String] = attributesMap.unsupportedKeys(bt)
-      override def getSeq(key: RuntimeKey): Option[Seq[String]] = attributesMap.getSeq(key)
-      override def get(key: RuntimeKey): Option[String] = attributesMap.get(key) match {
-        case Some(x) => Some(x)
-        case None => workflowOptions.getDefaultRuntimeOption(key.key).toOption
-      }
+    attributes.get
```
francares: Use getOrElse here.

scottfrazer: Originally I had this return a `Try[CromwellRuntimeAttributes]` but @geoffjentry thought it that apply methods should throw an exception.

what would the 'or else' be in this case?  the defaults?

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
```diff
@@ -24,163 +29,187 @@ case class CromwellRuntimeAttributes(docker: Option[String],
...
-    val validatedRuntimeAttributes = (attributeMapNel |@| runtimeAttributeNel) { (_, r) => r }
-
-    validatedRuntimeAttributes match {
-      case Success(r) => r
-      case Failure(f) =>
-        throw new IllegalArgumentException() with ThrowableWithErrors {
-          val message = "RuntimeAttribute is not valid."
-          val errors = f
-        }
+    unsupportedKeys(wdlRuntimeAttributes.attrs.keys, backendType) foreach log.warn
```
francares: It might be better just to call the function and let the function to log.

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
geoffjentry: yes, an apply should only be returning the actual type (which implies an exception on error)

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
```diff
@@ -24,163 +29,187 @@ case class CromwellRuntimeAttributes(docker: Option[String],
 object CromwellRuntimeAttributes {
   private val log = LoggerFactory.getLogger("RuntimeAttributes")
 
-  def apply(wdlRuntimeAttributes: RuntimeAttributes, backendType: BackendType): CromwellRuntimeAttributes =
-    fromAttributeMap(AttributeMap(wdlRuntimeAttributes.attrs), backendType)
+  def apply(wdlRuntimeAttributes: RuntimeAttributes, backendCall: BackendCall, workflowOptions: Option[WorkflowOptions]): CromwellRuntimeAttributes = {
```
francares: This logic should be part of each backend since they will have their own runtime attributes...

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
francares: It also goes for default runtime attributes.

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
francares: It also goes for default runtime attributes.

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
```diff
@@ -24,163 +29,187 @@ case class CromwellRuntimeAttributes(docker: Option[String],
...
-  }
-
-  private def validateRuntimeAttributes(attributeMap: AttributeMapTrait): ValidationNel[String, CromwellRuntimeAttributes] = {
-    val docker = attributeMap.get(DOCKER)
-    val zones = attributeMap.get(ZONES) map { _.split("\\s+").toVector } getOrElse Defaults.Zones
+  val LocalizationDisk = LocalDisk(LocalDiskName, DiskType.SSD, sizeGb=10)
+
+  private val keys = Set("cpu", "disks", "docker", "zones", "continueOnReturnCode", "failOnStderr", "preemptible", "memory")
+
+  private val defaultValues = Map(
```
francares: Shouldn't it be configurable?

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
```diff
@@ -24,163 +29,187 @@ case class CromwellRuntimeAttributes(docker: Option[String],
...
+        k -> WdlString(workflowOptions.getDefaultRuntimeOption(k).get)
+      case k if defaultValues.contains(k) =>
+        k -> defaultValues.get(k).get
+    }).toMap
+  }
+
+  val defaults: CromwellRuntimeAttributes = {
+    // .get below because we assume that the default values are valid
+    // There's a unit test in CromwellRuntimeAttributeSpec to test these defaults
+    validateRuntimeAttributes(defaultValues).get
```
francares: getOrElse here.

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
```diff
@@ -24,163 +29,187 @@ case class CromwellRuntimeAttributes(docker: Option[String],
...
     }
   }
 
-  private def validateDiskType(diskTypeName: String): ValidationNel[String, DiskType] = {
-    DiskType.values().find(_.diskTypeName == diskTypeName) match {
-      case Some(diskType) => diskType.successNel[String]
-      case None =>
-        val diskTypeNames = DiskType.values.map(_.diskTypeName).mkString(", ")
-        s"Disk TYPE $diskTypeName should be one of $diskTypeNames".failureNel
+  private def validateLocalDisks(value: Option[WdlValue]): ErrorOr[Seq[Disk]] = {
```
francares: Maybe you can create a recursive function here for validation...

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
geoffjentry: I hate raw .gets as much as anyone who'd ever look at this code, but there's a comment 2 lines above as to why it's being called this way

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
```diff
@@ -24,163 +29,187 @@ case class CromwellRuntimeAttributes(docker: Option[String],
...
-    val memory = validateMemory(attributeMap.get(MEMORY))
+    val memory = RuntimeAttributes.validateMemoryValue(attributeMap.getOrElse(MEMORY, WdlString(s"${defaults.memoryGB} GB"))) match {
+      case scala.util.Success(x) => x.to(MemoryUnit.GB).amount.successNel
+      case scala.util.Failure(x) => x.getMessage.failureNel
+    }
 
-    (failOnStderr |@| continueOnReturnCode |@| cpu |@| preemptible |@| disks |@| memory) {
-      CromwellRuntimeAttributes(docker, zones, _, _, _, _, _, _)
+    (docker |@| zones |@| failOnStderr |@| continueOnReturnCode |@| cpu |@| preemptible |@| disks |@| memory) {
+      new CromwellRuntimeAttributes(_, _, _, _, _, _, _, _)
```
francares: Shouldn't it be a Map? It contains 8 attributes and I think with current implementation it will tend to grow...
It would be easier if runtime attributes is a map and then keys from that map are verified and validated by each backend. Each backend also can contains or/and set default values.
The other thing you gain with this is that the Cromwell-engine is not modified so often.

geoffjentry: Considering this is all changing w/ the big PR I don't think we need to be handwringing a whole lot on specifics of how RAs are currently handled

src/test/scala/cromwell/engine/backend/CromwellRuntimeAttributeSpec.scala
francares: I think it would be better to have a package of sample WDLs based on common use and just a few classes. It will be better for reading, refactoring, etc...

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
francares: I see now the seq of Try :).

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
francares: OK but all agree on the new way of do it?

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
scottfrazer: The reason why it was like this is that some tests calls `unsupportedKeys`, but maybe if I can figure out how to make assertions on log messages (outside of Akka), I can change it to your suggestion

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
geoffjentry: It's going to need to be more generically handled by every backend which means that it'll have to be treated as a more general K/V pairing prior to entering backend space

src/test/scala/cromwell/engine/backend/CromwellRuntimeAttributeSpec.scala
cjllanwarne: hmm, that test sounds misleading - the ones in `workflow options` should BE the defaults. They are overridden by values in the WDL

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
scottfrazer: @francares When I was modifying all the code in this file I was thinking the exact same thing.

francares: @geoffjentry @scottfrazer Perfect! I just wanted to be sure we all are in the same page :). Thanks.

francares: I like this feature. It will help a lot on cases about authorization and some scenarios where  backend's runtime configuration values need to be generated on the fly.

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
scottfrazer: I definitely agree, though I'm not sure what the best path forward for that is.  I know @mcovarr and @gauravs90 are currently working on an actor to do this kind of backend-specific validation.  I didn't want to try to reinvent too much considering that's coming soon

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
scottfrazer: Perhaps `defaultValues

francares: I meant get default values from config file.

scottfrazer: Sorry my previous comment got submitted too quickly...

src/test/scala/cromwell/engine/backend/CromwellRuntimeAttributeSpec.scala
scottfrazer: Alright, I'll move these to `SampleWdl` then

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
geoffjentry: @scottfrazer Yeah, considering all of this is going to change radically I don't think that you should be needing to dance around those incoming changes. Coding to the current reality is fine here I think.

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
scottfrazer: > Shouldn't it be configurable?

I'm guessing you mean "configurable via the application.conf file"?  I agree, but given the impending conflicts we'll face, perhaps we should wait?

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
scottfrazer: Oh I see what you mean, I'll try it out

scottfrazer: Honestly I tried to do this, but my editor is completely unresponsive trying to modify this block of code because of all the `scalaz` stuff.  I've already spent 20 minutes on this trying to get it to work so I'm going to drop it unless everybody thinks it's really important...  my half-baked work in progress code is here: https://gist.github.com/scottfrazer/5f2bb0333eee002ce0a1.

src/test/scala/cromwell/engine/backend/CromwellRuntimeAttributeSpec.scala
scottfrazer: New tests now named

```
it should "allow workflow options to override defaults for all tasks in a WDL file"
it should "NOT allow workflow options to override values specified in a task runtime section"
```

geoffjentry:
Considering the context it's being merged into (i.e. what's there now, what will likely be changing and what will definitely be changing and thus not worth completely fixing) I think this is fine as-is

Issue #415, opened 2016-02-02
Replace backend-specific execution tables


mcovarr:
The `JES_JOB`, `LOCAL_JOB`, and `SGE_JOB` tables are specific to the backends currently hardcoded into Cromwell and are not suitable for dynamically pluggable backends.  The data currently captured by these tables should be replaced by a schema that is not backend specific, perhaps something like the following:

```
EXECUTION_INFO

EXECUTION_ID  INT           FK to EXECUTION.EXECUTION_ID, NOT NULL
INFO_KEY      VARCHAR(xyz)  NOT NULL
INFO_VALUE    VARCHAR(xyz)  
```

I'd also propose adding a nullable `BACKEND_TYPE VARCHAR(xyz)` field to `EXECUTION` to record what backend was actually used to run a job.

mcovarr:
Just a heads up to Intel friends @francares  @gauravs90 @jainh that work on this issue is now underway.

gauravs90:
@mcovarr  This is great :smile:

PR #403, opened 2016-01-24
Fixed broken test by semi-syncing liquibase and slick schemas.

kshakir:
Added a database schema diff util using liquibase, and a test to verify the schema stays similar to slick.

PR #371, opened 2016-01-11
SOMEONE PLEASE MERGE (after telling coreone): Automatically run liquibase to migrate the schema.

kshakir:

PR #424, opened 2016-02-05
Custom slick thread pool, plus some async db cleanup

kshakir:


kshakir:
The wheel chooseth @scottfrazer and @mcovarr, but happy for anyone else to take a look. Am leaving it as two commits as the first has the more important change, and the second is verbose cleanup.

src/main/scala/cromwell/engine/backend/jes/Run.scala
```diff
@@ -155,9 +156,17 @@ case class Run(runId: String, pipeline: Pipeline, logger: WorkflowLogger) {
...
+      This doesn't even wait for the future to complete. Pretty sure this should be a message to the workflow actor,
+      that then contacts the database to change the state. For now, updating this end run to the database to pass in the
+      default, global execution context.
+       */
+
       // Update the database state:
       val newBackendInfo = JesCallBackendInfo(Option(JesId(runId)), Option(JesStatus(currentStatus.toString)))
-      globalDataAccess.updateExecutionBackendInfo(workflowId, BackendCallKey(call, pipeline.key.index), newBackendInfo)
+      globalDataAccess.updateExecutionBackendInfo(
+        workflowId, BackendCallKey(call, pipeline.key.index), newBackendInfo)(ExecutionContext.global)
```
mcovarr: I remember discussing this at the time it was written though it seems we kinda forgot to come back and clean it up.  :frowning:

src/main/scala/cromwell/engine/backend/sge/SgeBackend.scala
```diff
@@ -88,7 +88,8 @@ case class SgeBackend(actorSystem: ActorSystem) extends Backend with SharedFileS
       logger.error(s"Failed to update database status: $status, rc: $rc because $e")
   }
 
-  private def updateSgeJobTable(call: BackendCall, status: String, rc: Option[Int], sgeJobId: Option[Int]): Future[Unit] = {
+  private def updateSgeJobTable(call: BackendCall, status: String, rc: Option[Int], sgeJobId: Option[Int])
+                               (implicit ec: ExecutionContext): Future[Unit] = {
```
mcovarr: Hrm I've ripped out these `updateWhatevsBackend` methods with my #415 work, I'll have to remember to thread through the implicit context during my rebase purgatory.  :smile:

src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -169,11 +166,39 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
...
+  override def close(): Unit = {
+    actionThreadPool.shutdown()
+    database.close()
+  }
 
   // Run action with an outer transaction
   private def runTransaction[R](action: DBIOAction[R, _ <: NoStream, _ <: Effect]): Future[R] = {
-    database.run(action.transactionally)
+    //database.run(action.transactionally) <-- https://github.com/slick/slick/issues/1274
+    Future(Await.result(database.run(action.transactionally), Duration.Inf))(actionExecutionContext)
```
mcovarr: "custom" and "slick" in the branch name did not prepare me for this :sob:  Pardon me while I upvote slick/slick#1274

mcovarr:
I'm certainly not a fan of littering a code base with FIXMEs but in this case I do think it's warranted in that one spot (you know where I mean :smile:)

src/main/scala/cromwell/engine/backend/jes/Run.scala
geoffjentry: This will almost certainly run afoul of the incoming intel changes, but that makes it @mcovarr's problem :)

src/main/scala/cromwell/engine/backend/sge/SgeBackend.scala
geoffjentry: This Whatevs backend sounds intriguing

src/main/scala/cromwell/engine/db/DataAccess.scala
```diff
@@ -1,21 +1,20 @@
...
 
-import scala.concurrent.Future
+import scala.concurrent.{ExecutionContext, Future}
 
 object DataAccess {
   val globalDataAccess: DataAccess = new slick.SlickDataAccess()
 }
 
-trait DataAccess {
+trait DataAccess extends AutoCloseable {
```
geoffjentry: I haven't really seen anyone use this in scala, presumably just because it's a new-ish construct. I'm not saying this to poo poo it (it seems like a good idea), but more just to ensure we understand how it's actually working here. Is this something you've seen other slick stuff doing?

src/main/scala/cromwell/engine/db/DataAccess.scala
```diff
@@ -23,93 +22,109 @@ trait DataAccess {
   def createWorkflow(workflowDescriptor: WorkflowDescriptor,
                      workflowInputs: Traversable[SymbolStoreEntry],
                      calls: Traversable[Scope],
-                     backend: Backend): Future[Unit]
+                     backend: Backend)(implicit ec: ExecutionContext): Future[Unit]
```
geoffjentry: I coulda saved myself some time and used all of your implicits instead of doing it myself yesterday ;)

src/main/scala/cromwell/engine/backend/jes/Run.scala
mcovarr: oh definitely it's already conflicting...

src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -169,11 +166,39 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
     case o => wdlType.fromWdlString(dbValue)
   }
 
-  override def shutdown() = database.shutdown
+  /**
+    * Create a special execution context, a fixed thread pool, to run each of our composite database actions. Running
```
geoffjentry: I do still want to talk about some of this stuff w/ an eye toward the long term

geoffjentry:
It's entirely possible that this resolves all of the big bang problems, including the spray responsiveness. From what I saw yesterday the pain and suffering was primarily coming from the DB access.

src/main/scala/cromwell/engine/db/DataAccess.scala
kshakir: I have not seen other slick code using this, but then again-- I haven't seen _most_ typesafe code (slick, spray, akka, etc.) tidying up. Usually one has to go digging deep into the test code / google groups for finding instructions on how to cleanly shutdown.

What happened here was that I was wiring in the `ExecutionContext`s, when I realized I already made `SlickDataAccess` extend `AutoCloseable` back in PRs #371 / #403. The change was made such that when the liquibase/slick comparison finished, the both databases would be [`.autoClosed`](https://github.com/broadinstitute/cromwell/blob/e95f46dcffe73b0795b7d6b2aa999035500cec97/src/test/scala/cromwell/engine/db/slick/SlickDataAccessSpec.scala#L97-L98) using `better.files`' [ARM](https://github.com/pathikrit/better-files#lightweight-arm).

geoffjentry: Yeah ARM & Scala has been a weird thing for a while now :)

As long as you're reasonably sure that it's doing what you think it's doing (or at least not doing anything unfriendly) I'm all for it - as I said, on paper it seems like a good idea. I was just a bit leery about it being fairly unvetted in the scala/slick context.

kshakir:
Added a basic regression test that verifies our deadlock workaround.

mcovarr:
:+1: nice job getting this to work, though I really hope Slick 3.2.0 makes this unnecessary.  :smile:

geoffjentry:
At some point over the weekend I'll see if it seems to resolve the other issues we were seeing as well. If so I'll go ahead and close the ticket I have. I don't think we really need to bulkhead out anything other than the DB IO at the moment, will take a look.

src/main/scala/cromwell/engine/db/DataAccess.scala
scottfrazer: Man I'm so confused :)  I've never seen an `AutoCloseable` before (I'm guessing it's [this](https://docs.oracle.com/javase/8/docs/api/java/lang/AutoCloseable.html)).  I read the API but I guess I'm still not clear on why it was added!

geoffjentry:
@kshakir I want to tech talk on this a bit prior to merging. I think it's fine as-is but want to make sure we think through the nuance and/or logical next steps

geoffjentry:
@kshakir I wonder if there's a way to put in a little check in all of our funcs here which verifies that the EC which comes in is the one that we're hoping it to be? As I mentioned at tech talk I always get sketched out at these implicit ECs. 

Either way, :+1: from me

scottfrazer:
:+1:

PR #425, opened 2016-02-05
Fixed CopyOutputCall for scattered calls

cjllanwarne:


cjllanwarne:
Wheelers: @Horneth @scottfrazer

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -716,9 +716,18 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
       (upstream forall { _.nonEmpty }) && dependenciesResolved
     }
 
+    lazy val nonFinalCallsComplete= executionStore forall {
```
Horneth: space before =

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -716,9 +716,18 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
...
+      case (key: FinalCallKey, _) => true
+      case (_, x: ExecutionStatus) if x.isTerminal => true
+      case _ => false
+    }
+
     def isRunnable(entry: ExecutionStoreEntry) = {
-      val (key, status) = entry
-      status == ExecutionStatus.NotStarted && arePrerequisitesDone(key)
+      entry match {
+        case (key: FinalCallKey, ExecutionStatus.NotStarted) => nonFinalCallsComplete
```
mcovarr: Thanks for fixing this.  What I tried (and failed) to do was make a final call dependent on every other call.  I think I must have confused the scatter placeholder with the collector; the latter is only generated after the scatter explodes, and the former is not sufficient for the final call to depend on.

Anyhoo there is a misleading comment above `prerequisiteScopes` in `FinalCall` that should be corrected.  I'm less certain what do about the code.

cjllanwarne: I updated the comment and removed the (no-longer-used) dependencies

geoffjentry:
:+1:

Horneth:
:+1:

PR #426, opened 2016-02-05
Metadata endpoint works with CopyWorkflowOutputsCalls

cjllanwarne:


src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -326,10 +328,11 @@ class WorkflowManagerActor(backend: Backend) extends LoggingFSM[WorkflowManagerS
 
   private def workflowStdoutStderr(workflowId: WorkflowId): Future[Map[FullyQualifiedName, Seq[Seq[CallLogs]]]] = {
     def logMapFromStatusMap(descriptor: WorkflowDescriptor, statusMap: Map[ExecutionDatabaseKey, ExecutionStatus]): Try[Map[FullyQualifiedName, Seq[Seq[CallLogs]]]] = {
+      import cromwell.engine.finalcall.FinalCall._
       Try {
         val sortedMap = statusMap.toSeq.sortBy(_._1.index)
         val callsToPaths = for {
-          (key, status) <- sortedMap if hasLogs(statusMap.keys)(key)
+          (key, status) <- sortedMap if hasLogs(statusMap.keys)(key) && !key.fqn.isFinalCall
```
mcovarr: Could maybe move the `isFinalCall` invocations up in the filtration order as they're super cheap

mcovarr:
:+1:

cjllanwarne:
"also @geoffjentry", spoke the wheel.

src/main/scala/cromwell/engine/finalcall/CopyWorkflowOutputsCall.scala
```diff
@@ -19,7 +18,7 @@ object CopyWorkflowOutputsCall {
   * Final call implementation that copies workflow outputs to a specified destination.
   */
 case class CopyWorkflowOutputsCall(override val workflow: WorkflowDescriptor) extends FinalCall {
-  override def unqualifiedName = CopyWorkflowOutputsCall.name(workflow.name)
+  override def unqualifiedName = CopyWorkflowOutputsCall.UnqualifiedName
```
geoffjentry: I'd prefer to not need the CopyWorkflowOutputsCall part required here but don't care too much if you'd like to retain it

src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -314,11 +314,13 @@ class WorkflowManagerActor(backend: Backend) extends LoggingFSM[WorkflowManagerS
       BackendCallKey(descriptor.namespace.workflow.findCallByName(callName).get, key.index, key.attempt)
     def backendCallFromKey(descriptor: WorkflowDescriptor, callName: String, key: ExecutionDatabaseKey) =
       backend.bindCall(descriptor, callKey(descriptor, callName, key))
+
+    import cromwell.engine.finalcall.FinalCall._
```
geoffjentry: I say this admittedly as someone who is biased against imports in the middle of code but this is now being imported in a couple of functions. Perhaps move it up?

geoffjentry:
:+1:

src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
cjllanwarne: It was more a scoping thing - since this imports an implicit on String, having it widely available seemed nasty to me

geoffjentry: If it's an enhancing implicit (I'm assuming the .isFinalCall) and not an implicit implicit (magic auto-conversions) I don't see think it's a big deal. You'd only be using it if you intended to. If it is indeed a magic auto-conversion we should probably not have that :)

PR #427, opened 2016-02-06
Remove backend specific tables, create universal EXECUTION_INFO table.  Closes #415

### LINKED ISSUE #415: Replace backend-specific execution tables

mcovarr:
Needs cleanup and some testing around JES restart in particular but AFAICT works.


mcovarr:
@scottfrazer and @cjllanwarne have won the Wheelection!

src/main/scala/cromwell/engine/db/DataAccess.scala
```diff
@@ -33,9 +33,9 @@ trait DataAccess {
 
   def getWorkflowsByState(states: Traversable[WorkflowState]): Future[Traversable[WorkflowDescriptor]]
 
-  def getExecutionBackendInfo(workflowId: WorkflowId, call: Call): Future[CallBackendInfo]
+  def getExecutionBackendInfo(workflowId: WorkflowId, call: Call): Future[Map[String, Option[String]]]
```
mcovarr: `(String, Option[String])` should be a type

src/main/scala/cromwell/engine/db/DataAccess.scala
```diff
@@ -97,19 +97,17 @@ trait DataAccess {
...
-
   def updateWorkflowOptions(workflowId: WorkflowId, workflowOptionsJson: String): Future[Unit]
 
-  def resetNonResumableJesExecutions(workflowId: WorkflowId): Future[Unit]
+  def resetNonResumableExecutions(workflowId: WorkflowId, isResumable: ExecutionAndExecutionInfo => Boolean): Future[Unit]
 
-  def findResumableJesExecutions(workflowId: WorkflowId): Future[Map[ExecutionDatabaseKey, JobKey]]
+  def findResumableExecutions(workflowId: WorkflowId,
+                              isResumable: ExecutionAndExecutionInfo => Boolean,
+                              jobKeyBuilder: ExecutionAndExecutionInfo => JobKey): Future[Map[ExecutionDatabaseKey, JobKey]]
```
mcovarr: `(ExecutionDatabaseKey, JobKey)` should be a type

src/main/scala/cromwell/engine/db/slick/ExecutionInfoComponent.scala
```diff
@@ -0,0 +1,49 @@
+package cromwell.engine.db.slick
+
+case class ExecutionInfo(executionId: Int,
+                         key: String,
+                         value: Option[String] = None,
+                         executionInfoId: Option[Int] = None)
+
+case class ExecutionAndExecutionInfo(execution: Execution, executionInfo: ExecutionInfo)
```
mcovarr: I think this type could/should go away and its usages replaced by the following type

src/test/scala/cromwell/engine/db/slick/SlickDataAccessSpec.scala
```diff
@@ -90,6 +88,8 @@ class SlickDataAccessSpec extends FlatSpec with Matchers with ScalaFutures with
     override def backendType: BackendType = throw new NotImplementedError
     override def workflowContext(workflowOptions: WorkflowOptions, workflowId: WorkflowId, name: String): WorkflowContext = throw new NotImplementedError
     override def pollBackoff: ExponentialBackOff = throw new NotImplementedError
+
+    override def executionInfoKeys: Seq[String] = Seq.empty
```
mcovarr: kill blank line

src/test/scala/cromwell/engine/db/slick/SlickDataAccessSpec.scala
```diff
@@ -682,26 +682,22 @@ class SlickDataAccessSpec extends FlatSpec with Matchers with ScalaFutures with
...
+      val pid2 = Option("987")
 
       (for {
         _ <- dataAccess.createWorkflow(workflowInfo, Nil, Nil, localBackend)
         _ <- dataAccess.insertCalls(workflowId, Seq(BackendCallKey(call1, shardIndex1)), localBackend)
         _ <- dataAccess.insertCalls(workflowId, Seq(BackendCallKey(call2, shardIndex2)), localBackend)
-        _ <- dataAccess.updateExecutionBackendInfo(workflowId, BackendCallKey(call1, shardIndex1), backendInfo1)
-        _ <- dataAccess.updateExecutionBackendInfo(workflowId, BackendCallKey(call2, shardIndex2), backendInfo2)
+        _ <- dataAccess.updateExecutionInfo(workflowId, BackendCallKey(call1, shardIndex1), "PID", pid1)
+        _ <- dataAccess.updateExecutionInfo(workflowId, BackendCallKey(call2, shardIndex2), "PID", pid2)
```
mcovarr: backend info keys should be constants in their Backend or BackendType companion objects

src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -414,68 +405,31 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
...
-        case Some(sgeJobResult: SgeJob) =>
-          new SgeCallBackendInfo(sgeJobResult.sgeJobNumber)
-        case _ =>
-          throw new IllegalArgumentException(
-            s"Unknown backend from db for (uuid, fqn): " +
-              s"($workflowId, ${call.fullyQualifiedName})")
-      }
+      executionInfos <- dataAccess.executionInfosByExecutionId(executionResult.executionId.get).result
+
+      x = executionInfos map { i => i.key -> i.value } toMap
```
mcovarr: cryptic variable name

src/main/scala/cromwell/engine/workflow/CallMetadataBuilder.scala
```diff
@@ -36,11 +34,14 @@ object CallMetadataBuilder {
...
-        case ji: JesJob => BackendValues("JES", jobId = ji.jesRunId, status = ji.jesStatus)
-        case ji: SgeJob => BackendValues("SGE", jobId = ji.sgeJobNumber map { _.toString })
+    // FIXME needs to traverse the list of pluggable backends rather than hardcoding a list.
+    def extract(infos: ExecutionInfosByExecution): BackendValues = {
+      def extractValue(key: String): Option[String] = infos.executionInfos collectFirst { case i if i.key == key => i.value } flatten
+
+      infos.execution.backendType match {
+        case "Local" => BackendValues("Local", jobId = extractValue("PID"))
+        case "JES" => BackendValues("JES", jobId = extractValue("JES_RUN_ID"), status = extractValue("JES_STATUS"))
+        case "SGE" => BackendValues("SGE", jobId = extractValue("SGE_JOB_NUMBER"))
```
mcovarr: even given the current fixedness of backends, this extraction should be done in the backend and not here

src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -814,4 +723,14 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
 
     runTransaction(action)
   }
+
+  override def infosByExecution(id: WorkflowId): Future[Traversable[ExecutionInfosByExecution]] = {
+    val action = for {
+      rawTuples <- dataAccess.executionsAndExecutionInfosByWorkflowId(id.toString).result
+      groupedTuples = rawTuples groupBy { _._1 } mapValues { _ map { _._2 } }
```
mcovarr: could use a comment

cjllanwarne:
Make sure to run Tyburn over this before merging. And check that the /metadata and /timing endpoints still work for JES jobs

cjllanwarne:
Also, if possible, run something in JES beforehand, then run the liquibase update over the existing data and check the /metadata and /timing can still be accessed for the preexisting tasks

src/main/scala/cromwell/engine/backend/BackendType.java
```diff
@@ -1,5 +1,16 @@
...
 public enum BackendType {
-    LOCAL, JES, SGE;
+    LOCAL {
+        @Override
+        public String displayName() {
+            return "Local";
+        }
+    },
+    JES,
+    SGE;
```
cjllanwarne: One of these things is not like the others?

src/main/scala/cromwell/engine/backend/jes/Run.scala
```diff
@@ -156,8 +155,9 @@ case class Run(runId: String, pipeline: Pipeline, logger: WorkflowLogger) {
       logger.info(s"Status change from $prevStateName to $currentStatus")
 
       // Update the database state:
-      val newBackendInfo = JesCallBackendInfo(Option(JesId(runId)), Option(JesStatus(currentStatus.toString)))
-      globalDataAccess.updateExecutionBackendInfo(workflowId, BackendCallKey(call, pipeline.key.index), newBackendInfo)
+      // TODO the database API should probably be returning DBIOs so callers can compose and wrap with a transaction.
```
cjllanwarne: Nah, transactions cause no end of problems!

src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -256,11 +252,22 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
     runTransaction(action)
   }
 
+  private def insertExecutionInfos(insertResult: this.dataAccess.executionsAutoInc.SingleInsertResult,
```
cjllanwarne: I'd prefer `insertEmptyExecutionInfos` - I was staring at this for ages until I realised it's just creating empty items so that all subsequent calls don't need to worry about INSERT vs UPDATE

src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
cjllanwarne: Since this is the last line, why not just do this in the yield?

cjllanwarne: Oh I just read the actual line. Probably never mind!

src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
cjllanwarne: I think your original comment about replacing "tuples" with a real type will go a long way to making this more readable

src/main/scala/cromwell/engine/workflow/CallMetadataBuilder.scala
cjllanwarne: Why not just extract all `info`s for this execution and make them key/value pairs on the JSON object, being totally agnostic of what they actually are?

src/main/scala/cromwell/engine/backend/BackendType.java
```diff
@@ -1,5 +1,16 @@
...
+
+    public String displayName() {
+        return name();
+    }
 }
```
kshakir: Might as well, you know, press return and get rid of this. :wink:

cjllanwarne:
My comments are all optionals, and probably going to be addressed by your comments anyway.

The main structure looks good and I don't want to hold you up waiting for me so :+1: from me.

cjllanwarne:
If (but only if) this makes the metadata look different, please mention it in CHANGELOG.MD

src/main/scala/cromwell/engine/backend/jes/JesBackend.scala
```diff
@@ -68,6 +69,16 @@ object JesBackend {
 
   def authGcsCredentialsPath(gcsPath: String): JesInput = JesInput(ExtraConfigParamName, gcsPath, Paths.get(""), "LITERAL")
 
+  // Only executions in Running state with a recorded operation ID are resumable.
+  private val IsResumable: ExecutionAndExecutionInfo => Boolean = ei => {
```
scottfrazer: `IsResumable` -> `isResumable`?

src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
scottfrazer: +1

scottfrazer:
I give a tentative :+1:.  Like Chris said, I'd like to see it with the changes you got from the code review feedback from yourself :)

src/main/scala/cromwell/engine/backend/BackendType.java
geoffjentry: No, it's in an object

src/main/scala/cromwell/engine/backend/jes/JesBackend.scala
geoffjentry: No, it's in an object

src/main/scala/cromwell/engine/backend/jes/JesBackend.scala
```diff
@@ -709,6 +713,8 @@ case class JesBackend(actorSystem: ActorSystem)
   }
 
   override def findResumableExecutions(id: WorkflowId)(implicit ec: ExecutionContext): Future[Map[ExecutionDatabaseKey, JobKey]] = {
-    globalDataAccess.findResumableJesExecutions(id)
+    globalDataAccess.findResumableExecutions(id, IsResumable, BuildJobKey)
   }
+
+  override def executionInfoKeys: Seq[String] = Seq("JES_RUN_ID", "JES_STATUS")
```
geoffjentry: Insert my new-ish general plea to not use the Seq apply directly

src/main/scala/cromwell/engine/db/DataAccess.scala
geoffjentry: images

src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
geoffjentry: Agreed

src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
geoffjentry: I like this chris guy

geoffjentry:
:+1:

cjllanwarne:
Tyburn's current problems notwithstanding, :+1: from a doesn't break things point of view.

src/main/scala/cromwell/engine/backend/jes/JesBackend.scala
scottfrazer: Why?

src/main/scala/cromwell/engine/backend/BackendType.java
mcovarr: yeah yeah :smile:

src/main/scala/cromwell/engine/backend/BackendType.java
mcovarr: On purpose, the display name for LOCAL needs some beautification but JES and SGE are natural beauties.

src/main/scala/cromwell/engine/backend/jes/JesBackend.scala
geoffjentry: Seq() creates a List, it isn't a generic entity. Calling it with Seq obscures that fact and can lead to bugs (in fact that has already happened in the lifespan of Cromwell). It's admittedly a minor thing but it makes it more difficult to reason about what you have and for the cost of an extra character you're saying explicitly what you're doing. 

As an example I recently asked @Horneth to de-Seq something w/ a List and he pointed out that it really should be a Set. That was a positive example IMO since having to think about the right collection type led to a better concrete instantiation.

scottfrazer: I still don't get it... so Seq creates a List... why can that lead to bugs?  It sounds like Thibault's problem is less to do with `Seq` versus `Vector` constructors and more to do with ordered versus unordered Iterables

geoffjentry: People have a tendency to assume that `Seq()` is really just creating something which only conforms to the Seq interface. However inside of pattern matches and stuff like that scala knows what the object's type is for reals. Since people are generally used to Seq == List what has happened is that people supply List-specific operators in pattern matches which works until a non-List gets passed in. And of course since people love using Seq/Traversable as function signatures you have no way of knowing what *really* got passed in.

Really the larger point is that one should think about what they really want to be using and if it's "a bag of stuff" and you're just using Seq, just make it a List since that's what it really is and it's more clear what you're doing.

src/main/scala/cromwell/engine/workflow/CallMetadataBuilder.scala
mcovarr: I do like this idea, but it's both orthogonal to my changes and alters the metadata response in a way the artists formerly known as Blue might not appreciate.  I'll make a new ticket for this.

src/main/scala/cromwell/engine/db/slick/ExecutionInfoComponent.scala
mcovarr: nope I was wrong about that, needs to stay

src/main/resources/changesets/execution_backend_info.xml
```diff
@@ -0,0 +1,87 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.3.xsd">
+
+    <changeSet author="mcovarr" id="CREATE_EXECUTION_INFO">
```
francares: Is there a way to use a generic model for all backend executions? There are many attributes that are the same for all backends right?

src/main/resources/changesets/execution_backend_info.xml
```diff
@@ -0,0 +1,87 @@
...
+        <createTable tableName="EXECUTION_INFO">
```
francares: The other problem this model creates is that how does the engine will know the specifics for each backend model?

src/main/resources/changesets/execution_backend_info.xml
```diff
@@ -0,0 +1,87 @@
...
+                             columnNames="EXECUTION_ID, INFO_KEY" />
+    </changeSet>
+
+    <changeSet id="execution_backend_type_not_null" author="mcovarr">
+        <addNotNullConstraint tableName="EXECUTION"
+                              columnName="BACKEND_TYPE"
+                              schemaName="CROMWELL"
+                              columnDataType="VARCHAR(255)"/>
+    </changeSet>
+
```
ccarrizo: shouldn't the tables be only  status/job_id/execution? leaving the specifics of each backend isolated in its implementation? by having backend-implementation specific data on the engine I guess we're increasing the coefficient of coupling which eventually will bite us when trying to make changes donwstream (for whatever reason).
if Cromwell needs to be aware of which backend it is running then I guess there will be backend specific behaviour which will cause a chain of updates e.g. if we add an AWS-specific backend. instead if cromwell has no such backends specific behabior, then is cleaner to remove any reference from the code and DB's.
which scenario are we at?

src/main/scala/cromwell/engine/backend/BackendType.java
ccarrizo: do we need an enum? i remember when we had the initial discussion we envisioned all the instances to be dealt with equally, thus we only care of the collection of instances but not an enum of types.
I guess some design changes happened since last time I looked.
besides design styles, my main concern is on the component coupling metrics and the associated effect on cost of maintainability (coding time) of the project. (See Martin's metrics)

src/main/scala/cromwell/engine/workflow/CallMetadataBuilder.scala
```diff
@@ -35,11 +38,14 @@ object CallMetadataBuilder {
...
-        case ji: JesJob => BackendValues("JES", jobId = ji.jesRunId, status = ji.jesStatus)
-        case ji: SgeJob => BackendValues("SGE", jobId = ji.sgeJobNumber map { _.toString })
+    // FIXME needs to traverse the list of pluggable backends rather than hardcoding a list.
+    def extract(infos: ExecutionInfosByExecution): BackendValues = {
+      def extractValue(key: String): Option[String] = infos.executionInfos collectFirst { case i if i.key == key => i.value } flatten
+
+      infos.execution.backendType match {
+        case "Local" => BackendValues("Local", jobId = extractValue(LocalBackend.InfoKeys.Pid))
+        case "JES" => BackendValues("JES", jobId = extractValue(JesBackend.InfoKeys.JesRunId), status = extractValue(JesBackend.InfoKeys.JesStatus))
+        case "SGE" => BackendValues("SGE", jobId = extractValue(SgeBackend.InfoKeys.JobNumber))
```
ccarrizo: what if the backend was a composition? e.g. we trying to run a job on a distributed environment (Broad-clsters and GC and AWS) so we design a backend that is made of e.g. JES+SGE only for the sake of using the one that offers the best $$ for the task, should Cromwell be aware of which backend instance if the composite picking?. if we just had some form of:
BackendValues(someInstanceKey, jobId = extractValue(Backend.InfoKeys.JobNumber))
we could use any weird design of the backends tree and cromwell will still run

src/main/scala/cromwell/engine/package.scala
```diff
@@ -61,6 +64,19 @@ package object engine {
     }
   }
 
+  implicit class EnhancedExecution(val execution: Execution) extends AnyVal {
```
jainh: Why it needs to extends from AnyVal? Looks like it is an adaptor to execution.

geoffjentry: It's a pattern called an implicit value class.

geoffjentry: Specifically the extension of AnyVal makes it a value class which means that the underlying JVM byte code is operating directly on the execution and not the wrapper class

geoffjentry:
@ccarrizo A call wraps a single command line and thus can only be run on a single backend, correct? Composed backends wouldn't apply for a call.

We have a need to be able to report back certain backend specific values and from an operational/regulatory perspective we need to track other things for provenance and such. That information needs to be persisted somewhere - should the backends know about the DB (let's ignore for them oment that one backend does indeed talk directly to the DB)? If not that implies now need to know about the internal persistence. What if we decide to bulkhead all DB access behind a specialized actor/router? You can point to separating out the persistence but that's far beyond the scope of this work and is a separate block entirely.

I viewed this as a transitional PR. There'd been a request from team members on our side for a while now that things be done as bit by bit as possible. To do that implies not having full & sweeping changes as the only way to do *that* is one monolithic PR which causes a ton of problems. The whole process would be a bit smoother if you all could find other portions which could be folded directly into develop *now* which started to move the ball in the right direction instead of requiring that everything be perfect in one fell swoop.

ccarrizo:
@geoffjentry I understand the provenance need, this is what I meant (perhaps too short) when I mentioned "someInstanceKey", so we're able to track which particular backend (instance) executed the job, but we do not care about the TYPE of it, as we deal with them as equals. 
I think I might be missing some datapoints here as I can't see the reason for Cromwell to have type-specifc behavior which I'd like to understand.

geoffjentry:
@ccarrizo It's more that users & operators might want to be able to see that behavior. 

For instance imagine something sitting on top of a cromwell server where they know there's a single backend, they might want to build business logic about specifics of that backend. One could argue over whether or not that's a good thing but the truth is that it's already a reality so one way or the other it needs to remain.

Another example would be for operational/debugging bits. As an example some of the data we track which is JES specific has been invaluable when dealing w/ our friends on the other side of the fence in terms of figuring out what's gone awry as both pieces were being built. 

The primary reason I can think of to record what type of backend was used to dispatch would mainly be for informational purposes, although I could imagine what backend a call had been run on being part of a potential cost function on determining which backend a subsequent call would be applied to. For both situations you'd only need to store the backend name, not fundamentally link metadata to that backend but it's easy enough to do both in a single table schema.

mcovarr:
Different backends produce different metadata key/values, forcing all backends into the same set of keys would require changing all backends if the set of keys needs to be expanded (as is about to happen now for our production backend).

francares:
@ccarrizo @geoffjentry @mcovarr IMHO... the engine does not need to know everything about an specific backend or even keep all data about it (it will not scale in a long term). 
The responsibility of generating and handling specific data should be part of the specific backend. Saying that, the engine does not need to know about informative metadata from specific backend to live (encapsulation).
I think cromwell engine should use minimal set of information from backends needed to run right?... which backend did what and so... and if more information is needed like for doing extensive analytics you can get more information from the specific backend since that analysis will focus in specifics from the backend.

geoffjentry:
@francares I'm not saying the engine itself *needs* to know (although I could envision a scenario where it'd be useful for it to know), I'm saying that the outside world needs to know information particular to the backend. That information needs to be stored somewhere and right now the only somewhere is in the DB which is accessed purely through engine.

I see this PR as a transition point - it's against develop which does *not* have pluggable backends but starts to remove the direct requirements (i.e. the backend specific tables). It's not the final state things will live in but it makes the ultimate changes smaller down the road.

src/main/resources/changesets/execution_backend_info.xml
mcovarr: The engine would not need to know the specific keys for a backend.  It would either receive a set of generic KV pairs from a Backend to be persisted, or load all the KV pairs and serve them back to a caller (e.g. the metadata endpoint).

src/main/resources/changesets/execution_backend_info.xml
mcovarr: Metadata will need to be richer than just those.  Cromwell will want to record the backend type used for a call, but that is treated as just an opaque string.

src/main/scala/cromwell/engine/backend/BackendType.java
mcovarr: We do not need or want an enum for pluggable backends, this is a vestige of compiled-in backends that we're not ready to get rid of yet.

PR #428, opened 2016-02-08
Added liquibase note to changelog.

kshakir:


CHANGELOG.md
```diff
@@ -11,4 +11,12 @@
 }
 ```
 * Changed the JES runtime attributes `defaultDisks` and `defaultZones` to be simply `disks` and `zones` respectively.
-
+* Liquibase scripts now run automatically. Cromwell won't start if it detects evidence of manual migrations in a
+persistent database. Instead, before cromwell will start cleanly, the database should backed up, and then this SQL
+should be manually executed:
```
cjllanwarne: Could you mention who needs to do this? E.g. if I have a local database, do I need to run this? Or only if I initialised it with liquibase?

kshakir:
Text updated

kcibul:
lgtm - thanks Khalid

cjllanwarne:
:+1:

geoffjentry:
:+1:

PR #432, opened 2016-02-08
Turn on stackdriver status in proxy

coreone:


cjllanwarne:
:+1:

Horneth:
:+1:

coreone:
@cjllanwarne @Horneth Forgot that cromwell was different than other repos and needed a port mapping in the compose file instead of in Vault.

PR #434, opened 2016-02-09
File size engine function

Horneth:


Horneth:
Wheel chooses @scottfrazer & @cjllanwarne

cjllanwarne:
Please add to CHANGELOG.MD and README.MD

src/test/scala/cromwell/FileSizeWorkflowSpec.scala
```diff
@@ -0,0 +1,60 @@
...
+  */
+class FileSizeWorkflowSpec extends CromwellTestkitSpec {
+
+  object FileSize extends SampleWdl {
+    override def wdlSource(runtime: String = "") =
+      """task file_size {
+        |  File input_file
+        |
+        |  command {
+        |    echo "this file is 8 bytes" > created_file
```
cjllanwarne: This is probably untrue...

CHANGELOG.md
```diff
@@ -20,6 +20,6 @@ update DATABASECHANGELOG
     FILENAME = substr(FILENAME, instr(FILENAME, "src/main/migrations/") + length("src/main/migrations/"))
   where FILENAME like '%src/main/migrations/%'
 ```
-
 * Added Preemptible VMs support for JES. This has impacts on the API Endpoint responses as a Call/Shard can now be attempted multiple times. Each attempt will have its own entry.
+* Added support for [size](https://github.com/broadinstitute/wdl/blob/develop/SPEC.md#float-sizefile-string) WDL standard library function.
```
cjllanwarne: Probably best to not point at develop.

cjllanwarne:
Please add to Tyburn's engine functions test

:+1:

build.sbt
```diff
@@ -66,6 +67,7 @@ libraryDependencies ++= Seq(
   "org.scalaz" % "scalaz-core_2.11" % "7.1.3",
   "com.github.pathikrit" %% "better-files" % "2.13.0",
   "org.liquibase" % "liquibase-core" % "3.4.2",
+  "com.squants"  %% "squants"  % "0.6.1-SNAPSHOT",
```
scottfrazer: Instead of this, you could use the memory parsing library that's included in wdl4s.  Specifically the [MemoryUnit](https://github.com/broadinstitute/wdl4s/blob/develop/src/main/java/wdl4s/parser/MemoryUnit.java) and [MemorySize](https://github.com/broadinstitute/wdl4s/blob/develop/src/main/scala/wdl4s/MemorySize.scala).  [The tests](https://github.com/broadinstitute/wdl4s/blob/develop/src/test/scala/wdl4s/MemorySizeSpec.scala) have some examples of how to use it.

Squants looks like it does the job too... though it is perhaps a little concerning to have to use a snapshot release.

I guess I'm less concerned with which library is used and more concerned that we how have two of these libraries.  What do you think about choosing one?  I imagine we would still need at least a small piece of `MemorySize`, specifically the `parse` method so it can regex match and split memory strings we get from WDL runtime sections

src/test/scala/cromwell/FileSizeWorkflowSpec.scala
```diff
@@ -0,0 +1,60 @@
+package cromwell
+
+import akka.testkit.EventFilter
+import cromwell.util.SampleWdl
+import wdl4s._
+import wdl4s.values.WdlFloat
+
+/**
+  * Created by thibault on 2/8/16.
+  */
```
scottfrazer: These comments are definitely one of my IntelliJ pet-peeves

src/test/scala/cromwell/FileSizeWorkflowSpec.scala
```diff
@@ -0,0 +1,60 @@
...
+        |    echo "this file is 22 bytes" > created_file
+        |  }
+        |
+        |  output {
+        |    Float input_file_size = size(input_file)
+        |    Float created_file_size = size("created_file")
+        |    Float created_file_size_in_k = size("created_file", "K")
+        |    Float created_file_size_in_m = size("created_file", "M")
+        |    Float created_file_size_in_g = size("created_file", "G")
+        |    Float created_file_size_in_t = size("created_file", "T")
```
scottfrazer: I like that `size` can take a unit

src/test/scala/cromwell/FileSizeWorkflowSpec.scala
Horneth: Right this is obnoxious I need to remove this from the template

build.sbt
Horneth: Just talked about this in person but for the record I agree we should use only one and if there is already one (which I didn't realize) then there's no need to replace it.

src/test/scala/cromwell/FileSizeWorkflowSpec.scala
cjllanwarne: Absolutely, yes!

Since the docs say we support binary units (GiB) can we check for those too?

Horneth: We supported them because the library (squant) I was using supports them, but if fall back to use the one already in wdl4s then it's not supported. So I don't know I could add support for it in wdl4s but I don't know if it's really worth it since no one actually asked for it

CHANGELOG.md
Horneth: Ok but the link will be broken until we merge wdl4s develop into master then

src/test/scala/cromwell/FileSizeWorkflowSpec.scala
scottfrazer: wdl4s supports binary prefixes too.  If you think Squant is better then let's use Squant!

Horneth: Oh, my bad - I don't think it's better in the sense I don't like adding a library dependency where 90% of the code won't be used. Let's stick with MemorySize

mcovarr:
If we can find a library that does what that code in wdl4s is doing I'm all for replacing the code in wdl4s with that.  Whether squant is that library or not I'm not sure.  :smile:

geoffjentry:
Still :+1: for me. 

On the issue of whether or not to use the library my general preference is to use well vetted 3rd party libraries for things instead of rolling our own. That said I've not heard of this library before so can't say much (although I did see it endorsed by some people I trust). I don't really care one way or the other on this particular one.

src/test/scala/cromwell/FileSizeWorkflowSpec.scala
```diff
@@ -0,0 +1,80 @@
...
+        |    Float created_file_size_in_tb = size("created_file", "TB")
+        |    Float created_file_size_in_ki = size("created_file", "Ki")
+        |    Float created_file_size_in_kib = size("created_file", "KiB")
+        |    Float created_file_size_in_mi = size("created_file", "Mi")
+        |    Float created_file_size_in_mib = size("created_file", "MiB")
+        |    Float created_file_size_in_gi = size("created_file", "Gi")
+        |    Float created_file_size_in_gib = size("created_file", "GiB")
+        |    Float created_file_size_in_ti = size("created_file", "Ti")
+        |    Float created_file_size_in_tib = size("created_file", "TiB")
```
scottfrazer: haha, I love this test.

scottfrazer:
:+1:

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ update DATABASECHANGELOG
 * Added custom thread pool to workaround Slick [deadlocks](https://github.com/slick/slick/issues/1274). The thread pool
 size defaults to the Slick configuration value `db.numThreads`, but may be increased up to Slick's
 `db.maxConnections`, via a new property `actionThreadPoolSize`.
+* Added support for [size](https://github.com/broadinstitute/wdl/blob/master/SPEC.md#float-sizefile-string) WDL standard library function.
```
scottfrazer: So as far as I know we don't have a process for when to merge `develop` -> `master` on the wdl repository.  I was under the impression that `develop` would only get merged when we decide that it's a good time for a new version of the specification

geoffjentry: Yeah, IIRC we kinda decided that a lot of this (including what a spec version means) needs to be hashed out and then punted from there. I don't think we can make a snap decision on this and at the moment we have much larger fish to fry so I think what @scottfrazer said is the right thing to do for now.

Horneth: Hmm ok, does that mean we should point this link to develop until `develop` -> `master` ?

scottfrazer: yah

PR #436, opened 2016-02-09
This feels hacky to me

cjllanwarne:
This change makes workflows fail slow instead of fail fast.

I'm uncomfortable committing this as-is because I don't fully understand when startRunnableCalls is used - and is it possible to miss the "are we done (including failures)" check.

This *does* work when a single call fails as part of a longer workflow. No other cases have been tested.

Can anyone give me a quick tour of the control flow in WorkflowActor to make sure I haven't missed anything?

### DIFF
--- a/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
+++ b/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
@@ -5,6 +5,7 @@ import java.sql.SQLException
 import akka.actor.{ActorRef, FSM, LoggingFSM, Props}
 import akka.event.Logging
 import akka.pattern.pipe
+import cromwell.engine.ExecutionStatus.ExecutionStatus
 import cromwell.engine.callactor.CallActor
 import cromwell.engine.finalcall.{CopyWorkflowOutputsCall, FinalCall}
 import cromwell.engine.workflow.WorkflowManagerActor.{WorkflowActorSubmitFailure, WorkflowActorSubmitSuccess}
@@ -14,7 +15,7 @@ import wdl4s.types.WdlArrayType
 import wdl4s.values.{WdlArray, WdlCallOutputsObject, WdlValue}
 import CallActor.CallActorMessage
 import cromwell.engine.ExecutionIndex._
-import cromwell.engine.ExecutionStatus.ExecutionStatus
+import cromwell.engine.ExecutionStatus._
 import cromwell.engine.Hashing._
 import cromwell.engine._
 import cromwell.engine.backend.{Backend, BackendCall, JobKey}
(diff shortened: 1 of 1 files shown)


src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -777,15 +783,26 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
...
+        case (k: FinalCallKey, _) => processRunnableCall(k)
+        case (k, v) =>
+          val message = s"Unknown entry in execution store:\nKEY: $k\nVALUE:$v"
+          logger.error(message)
+          Failure(new UnsupportedOperationException(message))
+      }
+    }
+    else if (
+      !(executionStore exists ( (p: (ExecutionStoreKey, ExecutionStatus)) => p._2 == ExecutionStatus.Running)) &&
+        (executionStore exists ( (p: (ExecutionStoreKey, ExecutionStatus)) => p._2 == ExecutionStatus.Failed))
```
cjllanwarne: This is the "are we done" check

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -517,6 +518,11 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
       val updatedData = startRunnableCalls(data)
       val finalData = updatedData.removePersisting(callKey, ExecutionStatus.Preempted)
       stay using finalData
+    case Event(PersistenceSucceeded(callKey, ExecutionStatus.Failed, callOutputs), data) =>
+      executionStore += callKey -> ExecutionStatus.Failed
+      val updatedData = startRunnableCalls(data)
+      val finalData = updatedData.removePersisting(callKey, ExecutionStatus.Failed)
+      stay using finalData
```
cjllanwarne: This makes the "startRunnableCalls" happen after a failure

geoffjentry:
I came here to make a comment but as I started typing that made me wonder if what I was about to suggest might have been old behavior and explain why (we think) things went from fail slow to fail fast. Or, alternatively it was *always* fail fast w/o us meaning to do so and this is all bollocks ...

At one point startRunnableCalls was called findRunnableCalls, and then it'd kick everything off. I was going to suggest that one thing you could do is find runnable calls, and if the set is empty you know that either the workflow has failed or succeeded based on if the state of all of the calls (all success = success, any failure = failure, some not run ... uh oh!). Now that I go back and look at your code this seems more or less what you're doing, at least in spirit.

I was mainly just excited that this might explain the slow -> fast transition (again, assuming it ever actually happened in the first place!)

PR #441, opened 2016-02-10
Liquibase is really noisy

scottfrazer:
Specifically when you run with the command-line runner using the in-memory DB, it spits out pages and pages of liquibase updates

### DIFF
--- a/src/main/resources/logback.xml
+++ b/src/main/resources/logback.xml
@@ -19,6 +19,7 @@
           <layout class="cromwell.logging.TerminalLayout" />
         </encoder>
       </appender>
+      <logger name="liquibase" level="WARN"/>
     </then>
     <else>
       <appender name="PRETTY_APPENDER" class="ch.qos.logback.core.helpers.NOPAppender" />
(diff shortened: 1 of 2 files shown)


geoffjentry:
@scottfrazer Can you provide an example of what this would remove?

scottfrazer:
I just updated the description

src/main/resources/logback.xml
```diff
@@ -31,4 +31,5 @@
   </root>
   <logger name="com.zaxxer.hikari" level="ERROR"/>
   <logger name="HikariPool" level="ERROR"/>
+  <logger name="liquibase" level="OFF"/>
```
scottfrazer: Maybe should be ERROR...

cjllanwarne: Or even WARNING, as long as the trivial and annoying messages are removed, anything else has value

geoffjentry:
I think this should only apply to the PRETTY logging mode and be set to DEBUG or something like that so it could be turned back on if desired.

cjllanwarne:
@scottfrazer is this assigned to you because you're still working on it, or does this need a wheel?

scottfrazer:
Sorry just haven't gotten around to addressing comments.  I'll wheel it up in a second

scottfrazer:
Comments addressed.

WHEEL: @cjllanwarne @kshakir

cjllanwarne:
:+1:

PR #444, opened 2016-02-10
Workflows now fail slow

cjllanwarne:
Not *quite* as neat as I hoped, hopefully this change makes sense to people. 

In an ideal world, the state data would include executionStore, since it's just state by another name and would make things so much easier to work out.

### DIFF
--- a/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
+++ b/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
@@ -5,6 +5,7 @@ import java.sql.SQLException
 import akka.actor.{ActorRef, FSM, LoggingFSM, Props}
 import akka.event.Logging
 import akka.pattern.pipe
+import cromwell.engine.ExecutionStatus.ExecutionStatus
 import cromwell.engine.callactor.CallActor
 import cromwell.engine.finalcall.{CopyWorkflowOutputsCall, FinalCall}
 import cromwell.engine.workflow.WorkflowManagerActor.{WorkflowActorSubmitFailure, WorkflowActorSubmitSuccess}
@@ -14,7 +15,7 @@ import wdl4s.types.WdlArrayType
 import wdl4s.values.{WdlArray, WdlCallOutputsObject, WdlValue}
 import CallActor.CallActorMessage
 import cromwell.engine.ExecutionIndex._
-import cromwell.engine.ExecutionStatus.ExecutionStatus
+import cromwell.engine.ExecutionStatus._
 import cromwell.engine.Hashing._
 import cromwell.engine._
 import cromwell.engine.backend.{Backend, BackendCall, JobKey}
@@ -27,6 +28,7 @@ import cromwell.engine.{CallOutput, HostInputs, CallOutputs, EnhancedFullyQualif
 import cromwell.logging.WorkflowLogger
 import cromwell.util.TerminalUtil
 
+import scala.annotation.tailrec
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
 import scala.concurrent.duration._
@@ -48,6 +50,7 @@ object WorkflowActor {
   case class ScatterCompleted(callKey: ScatterKey) extends TerminalCallMessage
   case class CallFailedRetryable(callKey: OutputKey, executionEvents: Seq[ExecutionEventEntry], returnCode: Option[Int], failure: Throwable) extends CallFailed
   case class CallFailedNonRetryable(callKey: OutputKey, executionEvents: Seq[ExecutionEventEntry], returnCode: Option[Int], failure: String) extends CallFailed
+  case class CallFailedToInitialize(callKey: ExecutionStoreKey) extends CallFailed
   case object Terminate extends WorkflowActorMessage
   final case class CachesCreated(startMode: StartMode) extends WorkflowActorMessage
   final case class AsyncFailure(t: Throwable) extends WorkflowActorMessage
@@ -70,6 +73,11 @@ object WorkflowActor {
    * state data and passed forward to subsequent message processing.
    */
   case object StartRunnableCalls extends WorkflowActorMessage
+
+  /**
+    * Trigger a Running workflow to check whether or not it is complete (either successfully or not).
+    */
+  case object CheckForWorkflowComplete extends WorkflowActorMessage
 
   /**
    * Message sent to self to actually start a call.  This is necessary to synchronize access to the symbol cache
@@ -323,24 +331,6 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
     }
   }
 
-  /**
-   * Attempt to start all runnable calls and return updated state data.  If successful this will create a new copy
-   * of the state data including new pending persists, otherwise it schedules a transition to `WorkflowFailed` and
-   * returns the original state data.  This should only be called from `WorkflowRunning`.
-   */
-  private def startRunnableCalls(data: WorkflowData): WorkflowData = {
-    tryStartingRunnableCalls() match {
-      case Success(result: ExecutionStartResult) =>
-        // Create an updated version of the state data with all pending persists from the start result added.
-        val updatedData = result.startEntries.foldLeft(data) { case (d, StartEntry(k, s)) => d.addPersisting(k, s) }
-        if (result.newEntries.nonEmpty) startRunnableCalls(updatedData) else updatedData
-      case Failure(e) =>
-        logger.error(e.getMessage, e)
-        scheduleTransition(WorkflowFailed)
-        data
-    }
-  }
-
   private def initializeExecutionStore(startMode: StartMode): Future[Unit] = {
     val initializationCode = startMode.runInitialization(this)
     val futureCaches = for {
@@ -411,7 +401,7 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
     completionWork onFailure {
       case e =>
         logger.error(s"Preemption work failed for call ${callKey.tag}! " + e.getMessage, e)
-        scheduleTransition(WorkflowFailed)
+        self ! CheckForWorkflowComplete
     }
 
     val updatedData = data.addPersisting(callKey, retryStatus)
@@ -438,7 +428,6 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
     completionWork onFailure {
       case e =>
         logger.error(s"Completion work failed for call ${callKey.tag}! " + e.getMessage, e)
-        scheduleTransition(WorkflowFailed)
         self ! PersistenceFailed(callKey, ExecutionStatus.Done)
     }
 
@@ -462,7 +451,7 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
   when(WorkflowRunning) {
     case Event(StartRunnableCalls, data) =>
       val updatedData = startRunnableCalls(data)
-      if (isWorkflowDone) scheduleTransition(WorkflowSucceeded)
+      self ! CheckForWorkflowComplete
       stay() using updatedData
     case Event(CallStarted(callKey), data) if !data.isPending(callKey) =>
       executionStore += callKey -> ExecutionStatus.Running
@@ -486,7 +475,11 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
     case Event(message @ CallFailedNonRetryable(callKey, events, returnCode, failure), data) if data.isPersistedRunning(callKey) =>
       persistStatusThenAck(callKey, ExecutionStatus.Failed, sender(), message, callOutputs = None, returnCode = returnCode)
       val updatedData = data.addPersisting(callKey, ExecutionStatus.Failed)
-      scheduleTransition(WorkflowFailed)
+      stay() using updatedData
+    case Event(message @ CallFailedToInitialize(callKey), data) =>
+      logger.warn(s"Call failed to initialize: ${callKey.tag}")
+      persistStatusThenAck(callKey, ExecutionStatus.Failed, sender(), message, callOutputs = None, returnCode = None)
+      val updatedData = data.addPersisting(callKey, ExecutionStatus.Failed)
       stay() using updatedData
     case Event(message @ CallAborted(callKey), data) if data.isPersistedRunning(callKey) =>
       // Something funky's going on if aborts are coming through while the workflow's still running. But don't second-guess
@@ -509,14 +502,35 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
       callOutputs foreach updateSymbolCache(callKey)
       logger.debug(s"In state WorkflowRunning: Got PersistenceCompleted message for Done call ${callKey.tag}")
       val updatedData = startRunnableCalls(data)
-      if (isWorkflowDone) scheduleTransition(WorkflowSucceeded)
       val finalData = updatedData.removePersisting(callKey, ExecutionStatus.Done)
+      self ! CheckForWorkflowComplete
       stay using finalData
     case Event(PersistenceSucceeded(callKey, ExecutionStatus.Preempted, callOutputs), data) =>
       executionStore += callKey -> ExecutionStatus.Preempted
       val updatedData = startRunnableCalls(data)
       val finalData = updatedData.removePersisting(callKey, ExecutionStatus.Preempted)
+      self ! CheckForWorkflowComplete
       stay using finalData
+    case Event(PersistenceSucceeded(callKey, ExecutionStatus.Failed, callOutputs), data) =>
+      executionStore += callKey -> ExecutionStatus.Failed
+      val finalData = data.removePersisting(callKey, ExecutionStatus.Failed)
+      self ! CheckForWorkflowComplete
+      stay using finalData
+    case Event(CheckForWorkflowComplete, data) =>
+      checkForWorkflowComplete(data)
+      stay()
+  }
+
+  private def checkForWorkflowComplete(data: WorkflowData): Unit = {
+    val executionStatuses = executionStore.values.toSeq
+
+    if (!executionStatuses.contains(ExecutionStatus.Running) && !executionStatuses.contains(ExecutionStatus.Starting) && data.pendingExecutions.isEmpty) {
+      if (executionStatuses.contains(ExecutionStatus.Failed)) {
+        scheduleTransition(WorkflowFailed)
+      } else if (isWorkflowDone) {
+        scheduleTransition(WorkflowSucceeded)
+      }
+    }
   }
 
   private def startBackendCallWithMessage(message: CallStartMessage, backendCallKey: BackendCallKey, callInputs: Map[String, WdlValue]) = {
@@ -540,8 +554,7 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
       case backendCallKey: BackendCallKey => fetchLocallyQualifiedInputs(backendCallKey) match {
         case Success(callInputs) => startBackendCallWithMessage(message, backendCallKey, callInputs)
         case Failure(t) =>
-          logger.error(s"Failed to fetch locally qualified inputs for call ${backendCallKey.tag}", t)
-          scheduleTransition(WorkflowFailed)
+          self ! CallFailedToInitialize(callKey)
       }
       case finalCallKey: FinalCallKey => startActor(finalCallKey, Map.empty, message.startMode)
     }
@@ -602,6 +615,7 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
 
   private def removePendingCallKeyPersistence(data: WorkflowData, message: PersistenceMessage): FSM.State[WorkflowState, WorkflowData] = {
     val updatedData = data.removePersisting(message.callKey, message.executionStatus)
+    self ! CheckForWorkflowComplete
     stay using updatedData
   }
 
@@ -643,7 +657,7 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
       stay()
     case Event(AsyncFailure(t), data) if data.pendingExecutions.isEmpty =>
       logger.error(t.getMessage, t)
-      scheduleTransition(WorkflowFailed)
+      self ! CheckForWorkflowComplete
       stay()
     case Event(message @ AsyncFailure(t), _) =>
       // This is the unusual combination of warn + throwable logging since the expectation is that this will eventually
@@ -679,7 +693,7 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
       case Success(_) => self ! PersistenceSucceeded(storeKey, executionStatus, callOutputs)
       case Failure(t) =>
         logger.error(s"Error persisting status of call ${storeKey.tag} to $executionStatus", t)
-        scheduleTransition(WorkflowFailed)
+        self ! CheckForWorkflowComplete
     }
     persistFuture
   }
@@ -714,7 +728,12 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
     logger.info(s"created call actor for ${callKey.tag}.")
   }
 
-  private def tryStartingRunnableCalls(): Try[ExecutionStartResult] = {
+  /**
+    * Attempt to start all runnable calls and return updated state data.  This will create a new copy
+    * of the state data including new pending persists.
+    */
+  @tailrec
+  private def startRunnableCalls(data: WorkflowData): WorkflowData = {
 
     def upstreamEntries(entry: ExecutionStoreKey, prerequisiteScope: Scope): Seq[ExecutionStoreEntry] = {
       prerequisiteScope.closestCommonAncestor(entry.scope) match {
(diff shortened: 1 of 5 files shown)


mcovarr:
Re execution store as state data DSDEEPB-2293

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -1,10 +1,12 @@
 package cromwell.engine.workflow
 
+import java.awt.dnd.InvalidDnDOperationException
```
mcovarr: Oh cool Cromwell has drag 'n drop now

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -48,6 +50,7 @@ object WorkflowActor {
   case class ScatterCompleted(callKey: ScatterKey) extends TerminalCallMessage
   case class CallFailedRetryable(callKey: OutputKey, executionEvents: Seq[ExecutionEventEntry], returnCode: Option[Int], failure: Throwable) extends CallFailed
   case class CallFailedNonRetryable(callKey: OutputKey, executionEvents: Seq[ExecutionEventEntry], returnCode: Option[Int], failure: String) extends CallFailed
+  case class CallFailedToInitialise(callKey: OutputKey) extends CallFailed
```
mcovarr: :us:

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
Horneth: Easter egg feature

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -509,14 +522,30 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
...
+      stay using finalData
+    case Event(CheckForWorkflowComplete, data) =>
+      logger.info("Checking for workflow complete! executuib")
+      if (!(executionStore exists ( (p: (ExecutionStoreKey, ExecutionStatus)) => p._2 == ExecutionStatus.Running)) && data.pendingExecutions.isEmpty) {
+        if (executionStore exists ( (p: (ExecutionStoreKey, ExecutionStatus)) => p._2 == ExecutionStatus.Failed)) {
+          scheduleTransition(WorkflowFailed)
+        } else if (isWorkflowDone) {
+          scheduleTransition(WorkflowSucceeded)
+        }
+      }
```
mcovarr: can haz function?

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -777,26 +805,45 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
...
-      case _ =>
-        // Coalesce all the ExecutionStartResults into one all-encompassing ExecutionStartResult.
-        val zero = ExecutionStartResult(startEntries = Set.empty)
-        Success(entries.foldLeft(zero) { case (acc, e) => acc plus e.get } )
+    // Handle call key failures:
+    callKeyEntries filter (_._2.isFailure) foreach { case (key: ExecutionStoreKey, failure) =>
+      self ! CallFailedToInitialise(key)
     }
+
+    // Handle non-call key failures:
```
mcovarr: a comment without code?

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -509,14 +522,30 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
...
       stay using finalData
+    case Event(PersistenceSucceeded(callKey, ExecutionStatus.Failed, callOutputs), data) =>
+      executionStore += callKey -> ExecutionStatus.Failed
+      val finalData = data.removePersisting(callKey, ExecutionStatus.Failed)
+      self ! CheckForWorkflowComplete
+      stay using finalData
+    case Event(CheckForWorkflowComplete, data) =>
+      logger.info("Checking for workflow complete! executuib")
+      if (!(executionStore exists ( (p: (ExecutionStoreKey, ExecutionStatus)) => p._2 == ExecutionStatus.Running)) && data.pendingExecutions.isEmpty) {
+        if (executionStore exists ( (p: (ExecutionStoreKey, ExecutionStatus)) => p._2 == ExecutionStatus.Failed)) {
```
mcovarr: it also seems like you're only looking at values, which could clean this up a lot

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -509,14 +522,30 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
...
       val finalData = updatedData.removePersisting(callKey, ExecutionStatus.Preempted)
+      self ! CheckForWorkflowComplete
       stay using finalData
+    case Event(PersistenceSucceeded(callKey, ExecutionStatus.Failed, callOutputs), data) =>
+      executionStore += callKey -> ExecutionStatus.Failed
+      val finalData = data.removePersisting(callKey, ExecutionStatus.Failed)
+      self ! CheckForWorkflowComplete
+      stay using finalData
+    case Event(CheckForWorkflowComplete, data) =>
+      logger.info("Checking for workflow complete! executuib")
```
mcovarr: I can't say I'm familiar with that term

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -777,26 +805,45 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
...
+    // Handle non-call key failures:
+
+    // Handle successes:
+    val successes = (callKeyEntries ++ nonCallKeyEntries).values filter (_.isSuccess)
+    // Coalesce all the ExecutionStartResults into one all-encompassing ExecutionStartResult.
+    val zero = ExecutionStartResult(startEntries = Set.empty)
+    Success(successes.foldLeft(zero) { case (acc, e) => acc plus e.get } )
   }
 
+  case class CallsFailedToInitialiseException(failedCalls: Seq[(CallKey, Throwable)]) extends Throwable
```
mcovarr: :us:

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -777,26 +805,45 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
     if (runnableCalls.nonEmpty)
       logger.info(s"starting calls: " + runnableCalls.map(_.fullyQualifiedName).toSeq.sorted.mkString(", "))
 
-    val entries: Traversable[Try[ExecutionStartResult]] = runnableEntries map {
-      case (k: ScatterKey, _) => processRunnableScatter(k)
-      case (k: CollectorKey, _) => processRunnableCollector(k)
-      case (k: BackendCallKey, _) => processRunnableCall(k)
-      case (k: FinalCallKey, _) => processRunnableCall(k)
+    val (callKeyedRunnableEntries, unexpectedRunnableEntries)= runnableEntries.partition {
```
mcovarr: `unexpected` seems like unexpected terminology

geoffjentry:
@mcovarr re DSDEEPB-2293 are you suggesting that could happen here? That split certainly confounded the two of us when charting WorkflowActor's maze of twisty passages, all alike

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -438,8 +446,8 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
     completionWork onFailure {
       case e =>
         logger.error(s"Completion work failed for call ${callKey.tag}! " + e.getMessage, e)
-        scheduleTransition(WorkflowFailed)
         self ! PersistenceFailed(callKey, ExecutionStatus.Done)
+        self ! CheckForWorkflowComplete
```
geoffjentry: I don't believe it can happen in-JVM but if these messages are processed out of order is there any problem?

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -509,14 +522,30 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
...
+      self ! CheckForWorkflowComplete
       stay using finalData
+    case Event(PersistenceSucceeded(callKey, ExecutionStatus.Failed, callOutputs), data) =>
+      executionStore += callKey -> ExecutionStatus.Failed
+      val finalData = data.removePersisting(callKey, ExecutionStatus.Failed)
+      self ! CheckForWorkflowComplete
+      stay using finalData
+    case Event(CheckForWorkflowComplete, data) =>
+      logger.info("Checking for workflow complete! executuib")
+      if (!(executionStore exists ( (p: (ExecutionStoreKey, ExecutionStatus)) => p._2 == ExecutionStatus.Running)) && data.pendingExecutions.isEmpty) {
```
geoffjentry: If this persists beyond Miguel's below comment - did you swap a pair of {}'s for ()'s? Also the opening ( has a space and the closing ) does not (here and below)

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
geoffjentry: space prior to =

mcovarr:
Sure you could sneak some DSDEEPB-2293 in here and I wouldn't complain

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -777,26 +804,51 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
...
-      case Some(failure) => failure
-      case _ =>
-        // Coalesce all the ExecutionStartResults into one all-encompassing ExecutionStartResult.
-        val zero = ExecutionStartResult(startEntries = Set.empty)
-        Success(entries.foldLeft(zero) { case (acc, e) => acc plus e.get } )
+    // Handle non-call key failures:
+    nonCallKeyEntries foreach {
+      case (_, Failure(e)) =>
+        logger.error(e.getMessage, e)
+        self ! CheckForWorkflowComplete
```
mcovarr: send as many check messages as there are failures?

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
cjllanwarne: Maybe. If there's something in pending, the CheckForCompletions won't trigger. 

So, I'll compose these two, instead of sending both and hoping for the best.

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
cjllanwarne: Uh, just to be sure?

cjllanwarne: (I may change this)

geoffjentry: nukefromorbit

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -411,7 +412,7 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
     completionWork onFailure {
       case e =>
         logger.error(s"Preemption work failed for call ${callKey.tag}! " + e.getMessage, e)
-        scheduleTransition(WorkflowFailed)
+        self ! CheckForWorkflowComplete
```
mcovarr: does `CheckForWorkflowComplete` need to be a message?  looks like you could just call the method.

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -714,7 +740,7 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
     logger.info(s"created call actor for ${callKey.tag}.")
   }
 
-  private def tryStartingRunnableCalls(): Try[ExecutionStartResult] = {
+  private def tryStartingRunnableCalls(): ExecutionStartResult = {
```
mcovarr: `tryStartingRunnableCalls` no longer `try`s

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
geoffjentry: That's my fault. I asked for the logic to be centralised but if all it's doing is calling a method you're right

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -777,26 +803,51 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
...
+    val (callKeyedRunnableEntries, nonCallKeyRunnableEntries) = runnableEntries.partition {
+      case (k: CallKey, _) => true
+      case _ => false
+    }
+
+    val callKeyEntries: Map[CallKey, Try[ExecutionStartResult]] = callKeyedRunnableEntries collect {
+      case (k: BackendCallKey, _) => k -> processRunnableCall(k)
+      case (k: FinalCallKey, _) => k -> processRunnableCall(k)
+      case (k: CallKey, v) =>
+        val message = s"Unknown entry in execution store:\nKEY: $k\nVALUE:$v"
```
mcovarr: IIRC `CallKey`s don't stringify too well as they have `Call`s in them

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
mcovarr: also this could loop over just the values

mcovarr: and why is the handling of completely unrecognized keys so different from unrecognized call keys?

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -509,14 +514,35 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
...
+      stay using finalData
+    case Event(CheckForWorkflowComplete, data) =>
+      checkForWorkflowComplete(data)
+      stay()
+  }
+
+  private def checkForWorkflowComplete(data: WorkflowData): Unit = {
+    val executionStatuses = executionStore.values
+
+    if (!(executionStatuses exists ( _ == ExecutionStatus.Running)) && data.pendingExecutions.isEmpty) {
```
mcovarr: I think `exists` could be simpler as `contains`

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -777,26 +803,51 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
...
     }
+
+    // Handle successes:
+    val successes = (callKeyEntries ++ nonCallKeyEntries).values filter (_.isSuccess)
+    // Coalesce all the ExecutionStartResults into one all-encompassing ExecutionStartResult.
+    val zero = ExecutionStartResult(startEntries = Set.empty)
+    successes.foldLeft(zero) { case (acc, e) => acc plus e.get }
   }
 
+  case class CallsFailedToInitialiseException(failedCalls: Seq[(CallKey, Throwable)]) extends Throwable
```
mcovarr: :us:

mcovarr:
a test would be nice

geoffjentry:
@mcovarr re test ;) https://www.manning.com/books/type-driven-development-with-idris

cjllanwarne:
The Wheel speaks!

@Horneth @mcovarr

cjllanwarne:
@geoffjentry before clicking on that, I was hoping for a Type Driven Development DVD hosted by Idris Elba. I was... disappointed.

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
cjllanwarne: Honestly, I prefer it this way. Discrete chunks of work handled as separable and reusable messages. Nice.

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
cjllanwarne: DoOrDoNotThereIsNoTryStartRunnableCalls?

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
cjllanwarne: But, it was like that when I got here!

geoffjentry:
@cjllanwarne Your ideas intrigue me and I'd like to subscribe to your newsletter

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
cjllanwarne: Wow. Thanks for pointing that out. It's now 100 times nicer

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
mcovarr: yeah that's probably how I noticed it :stuck_out_tongue:  don't feel obligated to fix it, but if you do great!

Horneth:
@cjllanwarne Rumors say he might give a Type Driven Development tutorial in the post-generic scene of the next James Bond

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
mcovarr: How is that encapsulation different from a method, except that now there's a new message type and handler case?  

The motivation for internal messages here is to guarantee that mutable state is managed only in the actor's single thread.  When a composed future wants to manipulate mutable state, that can't be done on an arbitrary `Future` thread that runs asynchronously to the actor, so the actor messages itself to schedule the manipulation to be done synchronously.

cjllanwarne: Hmm, I'm on the edge here. One thing I'd say - at the moment all of the functions make sure they leave database state consistent when the actor is "between" the processing of messages. They mess around in all sorts of ways "during" the processing of messages.

So, if I want to make sure checkForWorkflowComplete looks at a complete and consistent view of the world (which I do), I want to make it happen outside of any other message processing activity. There's a lot less thinking about "is it appropriate to check this now with the current state of the state" required if it's isolated in its own message. And I *hate* thinking.

mcovarr: oh whoops, this particular case is in a completion callback on a future.  soooo here it should be a message after all, for the reasons above.  :smile:  up to you whether you want to do that generally.

geoffjentry: In general I do feel like messages to actors feels more actor-y than futures with callbacks

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
Horneth: Wouldn't 
```
val callKeyEntries: Map[CallKey, Try[ExecutionStartResult]] = callKeyedRunnableEntries collect {
      case (k: CallKey, v) => k -> processRunnableCall(k)
    }
```
be enough ?
The only classes extending `CallKey` are `BackendCallKey` and `FinalCallKey` anyway

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
cjllanwarne: Since it's an Iterable, it doesn't have contains. I could "toSeq" it but that seems like slight overkill

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
Horneth: What about just checking for completion when you receive a `PersistenceSucceeded` / `Failed` message ? I think this is a time when you're supposed to have a real representation of the world.
It just seems to me that there is a lot of those `CheckForWorkflowComplete` flying around. For example in something like this
```scala
  persistStatusThenAck(callKey, ExecutionStatus.Failed, sender(), message, callOutputs = None, returnCode = returnCode)
  val updatedData = data.addPersisting(callKey, ExecutionStatus.Failed)
  stay() using updatedData
```
persistStatusThenAck will send a `PersistenceSucceeded` (or Failure) message which will send already a `CheckForWorkflowComplete` message, so I don't see the point of re-sending one in this code.

cjllanwarne: I replaced everywhere which said `scheduleTransition(WorkflowFailed)` into a `CheckForComplete` message. I see your point and I've removed a few, but I'm not confident enough in avoiding all edge cases to unilaterally remove absolutely everything apart from after `case PersistenceSucceeded` / `Failed`

cjllanwarne:
@Horneth @mcovarr lots of noise has just happened but this is now re-ready for review

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -481,12 +470,16 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
       resendDueToPendingExecutionWrites(message)
       stay()
     case Event(message @ CallFailedRetryable(callKey, events, returnCode, failure), data) if data.isPersistedRunning(callKey) =>
-      // Note: Currently Retryable == Preempted. If another reason than preemption arises that requires retrying at this level, this will need to be updated
+      // Note: Currently Retryable == Preempted. If another reason than preemption arizes that requires retrying at this level, this will need to be updated
```
mcovarr: :gb:

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -731,7 +750,7 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
           executionStore filter { case(k, _) => k.scope == prerequisiteScope && k.index == entry.index } toSeq
 
         /**
-         * Otherwise, simply refer to the entry the collector entry.  This means that 'entry' depends
+         * Otherwize, simply refer to the entry the collector entry.  This means that 'entry' depends
```
mcovarr: search / replace Americanization seems unwize

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -1154,21 +1179,23 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
...
+        Success(ExecutionStartResult(Set(StartEntry(callKey, ExecutionStatus.Starting))))
     }
-    Success(ExecutionStartResult(Set(StartEntry(callKey, ExecutionStatus.Starting))))
   }
 
   private def symbolsAsTable: Future[Traversable[Seq[String]]] = {
     def buildColumnData(entry: SymbolStoreEntry): Seq[String] = {
       // If the value of this symbol store entry is defined, that value truncated to MarkdownMaxColumnChars,
-      // otherwise the empty string.
+      // otherwize the empty string.
```
mcovarr: :no_entry_sign:

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
cjllanwarne: I can't win

src/test/scala/cromwell/WorkflowFailSlowSpec.scala
```diff
@@ -0,0 +1,19 @@
...
+    "complete other tasks but ultimately fail" in {
+      runWdlAndAssertOutputs(
+        sampleWdl = SampleWdl.WorkflowFailSlow,
+        eventFilter = EventFilter.info(pattern = s"persisting status of E to Done.", occurrences = 1),
+        runtime = "",
+        expectedOutputs = Map(),
+        terminalState = WorkflowFailed
+      )
+    }
+  }
```
mcovarr: Doesn't really assert all the conditions here.  Maybe a metadata query would be in order?

src/test/scala/cromwell/BadDockerName.scala
```diff
@@ -0,0 +1,26 @@
...
+                    |runtime {
+                    |  docker: "/fauxbuntu:nosuchversion"
+                    |}
+                  """.stripMargin,
+        expectedOutputs = Map(),
+        terminalState = WorkflowFailed
+      )
+    }
+  }
+}
```
mcovarr: why is this part of these changes?

cjllanwarne: It was failing. On investigation, this test was in completely the wrong place. And as a result, the test it should have been did not exist.

https://github.com/broadinstitute/cromwell/commit/253530a7bfa1f5d7cbcc79fc84057072c682a1aa

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
Horneth: Yes there are surely a few other places where it's needed, but pretty much everytime we call persistStatusAndAck there's no need to send `CheckForWorkflowComplete` just after IMO

mcovarr:
:+1:

Horneth:
:+1:

PR #445, opened 2016-02-11
Fix case on table drops

coreone:


mcovarr:
:+1: Cloud SQL is definitely case sensitive WRT identifiers: https://cloud.google.com/appengine/docs/java/cloud-sql/

PR #446, opened 2016-02-11
Fix various bugs related to unspecified call attempts

Horneth:
This also fixes https://broadinstitute.atlassian.net/browse/DSDEEPB-2873

### DIFF
--- a/src/main/scala/cromwell/engine/db/slick/ExecutionInfoComponent.scala
+++ b/src/main/scala/cromwell/engine/db/slick/ExecutionInfoComponent.scala
@@ -36,6 +36,7 @@ trait ExecutionInfoComponent {
     (executionId: Rep[Int], key: Rep[String]) => for {
       executionInfo <- executionInfos
       if executionInfo.key === key
+      if executionInfo.executionId === executionId
     } yield executionInfo.value)
 
   val executionsAndExecutionInfosByWorkflowId = Compiled(
--- a/src/main/scala/cromwell/engine/workflow/CallMetadataBuilder.scala
+++ b/src/main/scala/cromwell/engine/workflow/CallMetadataBuilder.scala
@@ -77,7 +77,7 @@ object CallMetadataBuilder {
     executionMap => {
       // Remove collector entries for the purposes of this endpoint.
       val inputsNoCollectors = filterCollectorSymbols(callInputs)
-      val inputsByKey = inputsNoCollectors.groupBy(i => ExecutionDatabaseKey(i.scope, i.key.index))
+      val inputsByKey = inputsNoCollectors.groupBy(i => ExecutionDatabaseKey(i.scope, i.key.index, 1))
 
       for {
         (inputKey, inputs) <- inputsByKey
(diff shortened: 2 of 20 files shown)


geoffjentry:
:+1:

cjllanwarne:
Wheel: @kshakir @mcovarr

src/main/scala/cromwell/engine/db/slick/ExecutionInfoComponent.scala
```diff
@@ -36,6 +36,7 @@ trait ExecutionInfoComponent {
     (executionId: Rep[Int], key: Rep[String]) => for {
       executionInfo <- executionInfos
       if executionInfo.key === key
+      if executionInfo.executionId === executionId
```
mcovarr: whoops

src/main/scala/cromwell/engine/workflow/CallMetadataBuilder.scala
```diff
@@ -77,7 +77,7 @@ object CallMetadataBuilder {
     executionMap => {
       // Remove collector entries for the purposes of this endpoint.
       val inputsNoCollectors = filterCollectorSymbols(callInputs)
-      val inputsByKey = inputsNoCollectors.groupBy(i => ExecutionDatabaseKey(i.scope, i.key.index))
+      val inputsByKey = inputsNoCollectors.groupBy(i => ExecutionDatabaseKey(i.scope, i.key.index, 1))
```
mcovarr: so this effectively ignores the attempt in rolling up the inputs?

src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -261,7 +261,7 @@ class WorkflowManagerActor(backend: Backend) extends LoggingFSM[WorkflowManagerS
   }
 
   private def assertCallExistence(id: WorkflowId, callFqn: FullyQualifiedName): Future[Any] = {
-    globalDataAccess.getExecutionStatus(id, ExecutionDatabaseKey(callFqn, None)) map {
+    globalDataAccess.getExecutionStatus(id, ExecutionDatabaseKey(callFqn, None, 1)) map {
```
mcovarr: shouldn't this take the max attempt?

src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -292,7 +292,7 @@ class WorkflowManagerActor(backend: Backend) extends LoggingFSM[WorkflowManagerS
     for {
       _ <- assertWorkflowExistence(workflowId)
       _ <- assertCallExistence(workflowId, callFqn)
-      outputs <- globalDataAccess.getOutputs(workflowId, ExecutionDatabaseKey(callFqn, None))
+      outputs <- globalDataAccess.getOutputs(workflowId, ExecutionDatabaseKey(callFqn, None, 1))
```
mcovarr: same q here

mcovarr: oh nm, this just validates the call name is valid and 1 is fine for that

mcovarr:
:+1:

src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
Horneth: This method is used to respond to API queries which don't support the concept of attempt (and probably shouldn't). It verifies that a call with this FQN and index exists, which should be true if there was at least one attempt so I think 1 is ok here.

src/main/scala/cromwell/engine/workflow/CallMetadataBuilder.scala
Horneth: All attempts have the same inputs so this value doesn't matter I think.

PR #447, opened 2016-02-11
Add JES Return code in EXECUTION_INFO Table

Horneth:


cjllanwarne:
Wheel: @cjllanwarne @kshakir

src/main/scala/cromwell/engine/backend/BackendCall.scala
```diff
@@ -54,21 +54,37 @@ trait ExecutionHandle {
...
+  def asScriptReturnCode = this match {
+    case script: ScriptReturnCode => Option(code)
+    case _ => None
+  }
+
+  def asBackendReturnCode = this match {
+    case script: BackendReturnCode => Option(code)
+    case _ => None
+  }
+}
```
geoffjentry: \n

src/main/scala/cromwell/engine/backend/jes/JesBackend.scala
```diff
@@ -568,7 +569,7 @@ case class JesBackend(actorSystem: ActorSystem)
       status match {
         case Run.Success(events) if backendCall.runtimeAttributes.failOnStderr && stderrLength.intValue > 0 =>
           // returnCode will be None if it couldn't be downloaded/parsed, which will yield a null in the DB
-          FailedExecutionHandle(new Throwable(s"${log.tag} execution failed: stderr has length $stderrLength"), returnCode.toOption).future
+          FailedExecutionHandle(new Throwable(s"${log.tag} execution failed: stderr has length $stderrLength"), returnCode.toOption map ScriptReturnCode).future
```
geoffjentry: Could you somehow functionify `returnCode.toOption map ScriptReturnCode` and use it here and below. Personally I'd love to see it writable as `returnCode.toScriptReturnCode` or something like that but YMMV

src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -617,10 +618,18 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
       e => (e.status, e.endDt, e.rc, e.executionHash, e.dockerImageHash, e.resultsClonedFrom)
 
     val maybeDate = if (callStatus.isStarting || callStatus.isTerminal) Option(new Date().toTimestamp) else None
+    val overallHash = callStatus.hash map { _.overallHash }
+    val dockerHash = callStatus.hash flatMap { _.dockerHash }
+    val scriptRC = callStatus.returnCode flatMap { _.asScriptReturnCode }
+    val backendRCAction = callStatus.returnCode flatMap { _.asBackendReturnCode } map { code =>
```
geoffjentry: backendRcAction

I'm guessing this isn't easily for comp'd?

src/main/scala/cromwell/engine/backend/jes/JesBackend.scala
Horneth: Hmm that would probably be an implicit class around `Try[Int]` since returnCode is a `Try`, which feels a bit hacky to me but why not. Another solution is a
```scala
implicit def toScriptReturnCode(rc: Try[Int]): Option[ScriptReturnCode] = rc.toOption map ScriptReturnCode
```
And then we can just use`returnCode` in the `FailedExecutionHandle`

geoffjentry: Oof, those are the sorts of implicits I hate because it can be very non-obvious that a conversion is happening. A normal function is fine too, I just prefer the .toXYZ kinda syntax.

After I wrote this earlier I noticed that besides the two here there were some variants here and there as well - some were already options, some weren't.

src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -10,7 +10,8 @@ import _root_.slick.driver.JdbcProfile
 import com.typesafe.config.{Config, ConfigFactory, ConfigValueFactory}
 import cromwell.engine.ExecutionIndex._
 import cromwell.engine.ExecutionStatus._
-import cromwell.engine.backend.{Backend, JobKey, WorkflowQueryResult}
+import cromwell.engine.backend._
+import cromwell.engine.backend.jes.JesBackend
```
cjllanwarne: We don't want any backend references in SlickDataAccess

src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
Horneth: Could do something like this 
```scala
val insertJesReturnCode = for {
      rc <- callStatus.returnCode
      scriptRc <- rc.asScriptReturnCode
    } yield { (executionId: Int) => 
      dataAccess.executionInfosAutoInc += new ExecutionInfo(executionId, JesBackend.InfoKeys.JesReturnCode, Option(scriptRc.toString)) 
    }
val backendRCAction = insertJesReturnCode getOrElse { (executionId: Int) => DBIO.successful(0) }
```

src/main/scala/cromwell/engine/backend/jes/JesBackend.scala
Horneth: I thought you might say that :p For a "utility" conversion like that it's typically the kind of conversion I don't really need/care to know that it's happening but I hear your point.
I'll make a function and see if I can make use of it in the other cases.

src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
Horneth: Okay

src/main/scala/cromwell/engine/backend/BackendCall.scala
```diff
@@ -54,21 +54,37 @@ trait ExecutionHandle {
...
+
+  def asScriptReturnCode = this match {
+    case script: ScriptReturnCode => Option(code)
+    case _ => None
+  }
+
+  def asBackendReturnCode = this match {
+    case script: BackendReturnCode => Option(code)
+    case _ => None
+  }
```
cjllanwarne: This feels like an odd abstraction to me. Isn't this what an Either is for?

But on the other hand, can't a call have both a ScriptReturnCode AND a BackendReturnCode?

cjllanwarne: Backend return code feels like it's not a general concept across all backends and might be better in the BACKEND_EXECUTION_INFO table?

src/main/scala/cromwell/engine/backend/BackendCall.scala
```diff
@@ -54,21 +54,37 @@ trait ExecutionHandle {
...
+}
+case class ScriptReturnCode(code: Int) extends ReturnCode
+case class BackendReturnCode(code: Int) extends ReturnCode
```
cjllanwarne: Backend return code feels like it's not a general concept across all backends and might be better in the BACKEND_EXECUTION_INFO table?

cjllanwarne:
Backend return code feels like it's not a general concept across all backends and might be better in the BACKEND_EXECUTION_INFO table?

src/main/scala/cromwell/engine/backend/BackendCall.scala
mcovarr: This looks strange to me too, actually I'm not sure I grok the abstraction.

src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
mcovarr: Ultimately that's true of course, but I wouldn't worry about this too much now if it causes you pain.  There are already explicit references to the three backend implementations here.  :unamused:

src/main/scala/cromwell/engine/callactor/CallActor.scala
```diff
@@ -196,7 +196,7 @@ trait CallActor extends LoggingFSM[CallActorState, CallActorData] with CromwellA
     val message = executionResult match {
       case SuccessfulBackendCallExecution(outputs, executionEvents, returnCode, hash, resultsClonedFrom) =>
         WorkflowActor.CallCompleted(key, outputs, executionEvents, returnCode, if (workflowDescriptor.writeToCache) Option(hash) else None, resultsClonedFrom)
-      case SuccessfulFinalCallExecution => WorkflowActor.CallCompleted(key, Map.empty, Seq.empty, 0, None, None)
+      case SuccessfulFinalCallExecution => WorkflowActor.CallCompleted(key, Map.empty, Seq.empty, ScriptReturnCode(0), None, None)
```
cjllanwarne: Does this work properly for a call with `continueOnReturnCode = 1`? 

It looks like it'll store 0 even if the "successful" return code was 1

mcovarr:
Oh yeah I was assuming this was going to EXECUTION_INFO

Horneth:
It is going to EXECUTION_INFO

Horneth:
Is there another BACKEND_EXECUTION_INFO I don't know about ?

src/main/scala/cromwell/engine/backend/BackendCall.scala
Horneth: I asked myself the same question but BackendReturnCode is really more a BackendErrorCode.
At least the way we use it now is when the Call fails because of an error from the Backend (Preemption for example), we get the code but if it succeeds we only care about the Script code.
I guess we could still store the BackendReturnCode even on success though.

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -1068,7 +1068,7 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
         self ! CallFailedNonRetryable(collector, Seq.empty, None, e.getMessage)
       case Success(outputs) =>
         logger.info(s"Collection complete for Scattered Call ${collector.tag}.")
-        self ! CallCompleted(collector, outputs, Seq.empty, 0, hash = None, resultsClonedFrom = None)
+        self ! CallCompleted(collector, outputs, Seq.empty, ScriptReturnCode(0), hash = None, resultsClonedFrom = None)
```
cjllanwarne: Same comment as above - is this wrongly assuming that a call completion implies return code 0?

cjllanwarne:
@Horneth @mcovarr oh, my mistake. I thought EXECUTION_INFO was EXECUTION. Ignore my last comment :)

src/main/scala/cromwell/engine/callactor/CallActor.scala
Horneth: This is for the `SuccessfulFinalCallExecution` which doesn't really have a return code since it's not even a script (for now at least). Problem is returnCode is mandatory in a bunch of places where it doesn't really make sense. IMO this is unrelated to the ticket and would be a non-trivial refactor I think.

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
Horneth: Same response as above :p - This is the case of a CollectorCall for which the concept of return code doesn't really make sense..

src/main/scala/cromwell/engine/backend/BackendCall.scala
cjllanwarne: Hmm, how about every call returning a `returnCode: Option[ScriptReturnCode]` and a `backendErrorCode: Option[BackendErrorCode]` - with the `EXECUTION_INFO` table only getting an entry if `Option[BackendErrorCode]` is `Some`?

Horneth: I can get rid of the `asXXXReturnCode` which I agree are a bit weird because it requires the sealed trait to have knowledge of its implementations which I don't like either now that I think about it.
However I'm not 100% clear on whether we want to/can have both a Backend code and a Script code

src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
Horneth: It wasn't that bad I just added a "key" attribute in the BackendReturnCode class that we can use as a key for the EXECUTION_INFO table

src/main/scala/cromwell/engine/backend/BackendCall.scala
geoffjentry: commenting purely on @Horneth 's last statement about the trait knowing about implementors and not about this particular bit, FWIW it's viewed as the more functional approach of ADTs to have the trait encode most/all of the logic on behalf of its implementors vs the more OO approach where the trait is the lightweight bit. I don't generally have a strong opinion either way, sometimes I think it looks nicer one way, sometimes the other.

src/main/scala/cromwell/engine/backend/jes/JesBackend.scala
geoffjentry: Hah, sorry I have a lot of scar tissue built up from dealing w/ those functions :)

src/main/scala/cromwell/engine/backend/BackendCall.scala
cjllanwarne: I actually would prefer they didn't implement the same trait at all, and functions make their mind up whether they return one or the other (or both)

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
cjllanwarne: Oh I see. Yeah, this is fine.

src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -324,7 +325,7 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
 
       executionStatuses = executionKeyAndStatusResults map { execution =>
         (ExecutionDatabaseKey(execution.callFqn, execution.index.toIndex, execution.attempt),
-         CallStatus(execution.status.toExecutionStatus, execution.rc, execution.overallHash map { ExecutionHash(_, execution.dockerImageHash )}, None)) }
+         CallStatus(execution.status.toExecutionStatus, execution.rc  map ScriptReturnCode, execution.overallHash map { ExecutionHash(_, execution.dockerImageHash )}, None)) }
```
kshakir: extra space before map in `execution.rc  map ScriptReturnCode`

kshakir:
Nothing new to add. I'm assuming the incoming changes, already discussed, will make this :+1:, but happy to re-review.

Horneth:
@cjllanwarne @kshakir  Thinking more about this, I'm wondering if the "Backend Return Code" is something we want to consider abstract/general enough that every backend should have the option to return one (in which case an `Option[BackendRc]` should be wire from the backend back to the CA, like for the script rc), or if it's something really only specific to JES in which case we should treat it the same way we do the JES status and Run ID (which are updated directly from JES Backend, which is not great but will probably be fixed with intel changes when Backends become actors)..

geoffjentry:
@Horneth I think you're right that most reasonable execution backends should have *some* notion which could be viewed as a numeric return code. Perhaps "return code" isn't the right term there, maybe something like "error code" would be more apropos.

Pinging @mcovarr @gauravs90 and @francares here in case they have thoughts on the matter.

cjllanwarne:
@Horneth check out DSDEEPB-2876.

I think this would be better reported as a stringly-typed error message rather than an explicit "backend return code" field

kshakir:
> every backend should have the option to return one

:+1: 

> Add “reason(s) for failure” to database, metadata

:+1:, but happy if storing strings waits for a future ticket

Horneth:
@geoffjentry In this case I don't think it should be in the ExecutionInfos Table which by my understanding is more of a custom bag of key/values for backend specific info (like JES Run ID and JES status).
If this is general enough to be valid for any backend it should be in the EXECUTION table IMO.
@cjllanwarne Makes sense to have a string yes, which JES returns anyway but for now we only extract the error code from it

cjllanwarne:
@Horneth my suggestion for this PR (happy to be overruled - @geoffjentry @kcibul) would be to get the return code being ONLY the return code. And forget the "backend return code" entirely for now until that upcoming ticket (maybe enhance that other ticket to add "JES return codes appear in failure message" as another AC?)

Horneth:
@cjllanwarne I'm ok with that as long as we're ok with not persisting the JES return code at all until this other ticket is addressed - which IMO is not a huge deal anyway since we print it in the logs so it's not completely lost anyway if we need to debug.

cjllanwarne:
@Horneth sounds good to me

Horneth:
See https://github.com/broadinstitute/cromwell/pull/449

PR #448, opened 2016-02-12
Missing runtime attribute bug fix

cjllanwarne:
Stops the WorkflowActor from waiting forever if a call fails to initialise with invalid Runtime Attributes.


geoffjentry:
"Stops the WorkflowActor from waiting forever if a call fails to initialise with invalid Runtime Attributes."

:us:

geoffjentry:
:+1:

Horneth:
:+1:

PR #449, opened 2016-02-16
Do not mix script/backend RC + refactor setStatus DB method

Horneth:


Horneth:
Re @cjllanwarne @kshakir

src/main/scala/cromwell/engine/backend/local/LocalBackend.scala
```diff
@@ -121,7 +121,7 @@ case class LocalBackend(actorSystem: ActorSystem) extends Backend with SharedFil
     val StatusesNeedingUpdate = ExecutionStatus.values -- Set(ExecutionStatus.Failed, ExecutionStatus.Done, ExecutionStatus.NotStarted)
     def updateNonTerminalCalls(workflowId: WorkflowId, keyToStatusMap: Map[ExecutionDatabaseKey, CallStatus]): Future[Unit] = {
       val callFqnsNeedingUpdate = keyToStatusMap collect { case (callFqn, callStatus) if StatusesNeedingUpdate.contains(callStatus.executionStatus) => callFqn }
-      globalDataAccess.setStatus(workflowId, callFqnsNeedingUpdate, CallStatus(ExecutionStatus.NotStarted, None, None, None))
+      globalDataAccess.setOtherStatus(workflowId, callFqnsNeedingUpdate, ExecutionStatus.NotStarted)
```
cjllanwarne: Not sure whether `setOtherStatus` is the best name here

cjllanwarne:
Could you check what happens when you kill cromwell mid-run and then restart, specifically for a JES call? I have a feeling you may have fixed DSDEEPB-2127 with this status stuff (in which case, hooray!)

src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -604,24 +604,55 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
...
+
+  def setStartingStatusAction(workflowId: WorkflowId, scopeKeys: Traversable[ExecutionDatabaseKey])(implicit ec: ExecutionContext) = {
+    for {
+      executions <- executionsForStatusUpdate(workflowId, scopeKeys)
+      count <- executions.map(e => (e.status, e.startDt)).update((ExecutionStatus.Starting.toString, Option(new Date().toTimestamp)))
+      _ = assertUpdateCount(count, scopeKeys.size)
+    } yield ()
+  }
+
+  def setOtherStatus(workflowId: WorkflowId, scopeKeys: Traversable[ExecutionDatabaseKey], status: ExecutionStatus)(implicit ec: ExecutionContext): Future[Unit] = {
```
cjllanwarne: I think I prefer "updateStatus" - since we can always assume this is an update of an existing value, I think?

cjllanwarne:
A few minor comments but :+1: 
If this also fixes DSDEEPB-2127 then :+1:  :+1:

Horneth:
@cjllanwarne Looking into the WorkflowActor I found this comment 
```
      // The assumption for restart is that the call will have been found in state Running.  Explicitly
      // re-writing the Starting status causes the new start date to be persisted, which is desirable to
      // provide a realistic elapsed execution time.
```
which seems contradictory with DSDEEPB-2127..
This PR doesn't change the current behavior.

I found a bug indirectly related to this though where the resumable and non-resumable calls where not processed correctly leading to weird DB errors.

kshakir:
Opened the files in Intellij, as it was a bit easier to read. A number of files showed warnings due to left over imports or unused values, but at this second I'm not sure that they were introduced by _this_ PR.

LGTM

PR #458, opened 2016-02-18
fix timing diagram for calls that were preempted

kcibul:
To verify this fix, I did the following:

1. rename the attached file (VIR_912.preempt.tieout.json) to 'metadata' and put it in a directory (e.g. /tmp/foo) along with the /src/main/resources/workflowTimings/workflowTimings.html from cromwell.

2. From that directory run:

    python -m SimpleHTTPServer 8080
to start up a simple http server on port 8080 serving out of the local directory

3. Hit http://localhost:8080/workflowTimings.html in your browser.

This works because the html makes a AJAX call to ./metadata to get the metadata for the chart. You can't do this on a local filesystem because of cross-site garbage but this simple server does the trick.
You can see the calls that have retries are labeled as such. For example AggregatedBamMarkDuplicates or the HaplotypeCaller (see attached screenshot)

examplefix

### DIFF
--- a/src/main/resources/workflowTimings/workflowTimings.html
+++ b/src/main/resources/workflowTimings/workflowTimings.html
@@ -51,9 +51,16 @@
                 for (var callIndex in callList)
                 {
                     var index = callList[callIndex]["shardIndex"];
+                    var attempt = callList[callIndex]["attempt"];
+                    var callStatus = callList[callIndex].executionStatus;
+
+                    // add the index of the shard if there is one
+                    var thisCallName = (index == -1 ? callName : callName + "." + index);
+
+                    // add the retry number, unless this was a successfuly first attempt (for brevity)
+                    thisCallName = ( callStatus == "Done" && attempt == 1 ? thisCallName : thisCallName + ".retry-" + attempt)
 
                     // Remove the workflow name
-                    var thisCallName = (index == -1 ? callName : callName + "." + index);
                     thisCallName = thisCallName.replace(new RegExp("^" + workflowName + "\\."), "");
 
                     var executionEvents = callList[callIndex].executionEvents;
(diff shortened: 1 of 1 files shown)


cjllanwarne:
It's very cool that we get the timing breakdown for the preempted ones. As far as I know all of the other failures don't get any events...

cjllanwarne:
Anyway, :+1:

src/main/resources/workflowTimings/workflowTimings.html
```diff
@@ -51,9 +51,16 @@
                 for (var callIndex in callList)
                 {
                     var index = callList[callIndex]["shardIndex"];
+                    var attempt = callList[callIndex]["attempt"];
+                    var callStatus = callList[callIndex].executionStatus;
```
cjllanwarne: I actually have no idea- is ["name"] different from .name?

cjllanwarne:
One interesting thing I just spotted - on the sample image you provided, there's a non-zero overlap between one try finishing and the next one starting. I wonder why.

src/main/resources/workflowTimings/workflowTimings.html
scottfrazer:
```js
 node
> x = {"a": 2}
{ a: 2 }
> x.a
2
> x['a']
2
```

geoffjentry:
@cjllanwarne Perhaps one is spinning down and the next one has already triggered?

@kcibul :+1:

kcibul:
@cjllanwarne yeah I think it's because they happen within the same minute (at least the ones I looked at) and that's our granularity so they show up like that.  For example the ending one is 11:12-11:12 and the starting of the next is also 11:12-11:12

PR #460, opened 2016-02-19
Runtime attrs in metadata

Horneth:


cjllanwarne:
@scottfrazer @geoffjentry

src/main/scala/cromwell/engine/db/slick/RuntimeAttributeComponent.scala
```diff
@@ -0,0 +1,48 @@
...
+    } yield runtimeAttribute.value)
+
+  val runtimeAttributesByWorkflowUUID = Compiled(
+    (workflowUUID: Rep[String]) => for {
+      workflow <- workflowExecutions
+      if workflow.workflowExecutionUuid === workflowUUID
+      execution <- executions
+      if execution.workflowExecutionId === workflow.workflowExecutionId
+      runtimeAttribute <- runtimeAttributes
+    if runtimeAttribute.executionId === execution.executionId
```
geoffjentry: is this supposed to be shifted to the left?

src/main/scala/cromwell/webservice/WorkflowJsonSupport.scala
```diff
@@ -33,7 +33,7 @@ object WorkflowJsonSupport extends DefaultJsonProtocol {
     }
   }
   implicit val executionEventProtocol = jsonFormat3(ExecutionEventEntry)
-  implicit val callMetadataProtocol = jsonFormat15(CallMetadata)
+  implicit val callMetadataProtocol = jsonFormat16(CallMetadata)
```
geoffjentry: General comment, not asking @Horneth to do anything here - but since we don't have a whole lot of more room in CallMetadata that's a sign that stuff needs to be refactored

geoffjentry:
:+1:

src/main/scala/cromwell/webservice/WorkflowJsonSupport.scala
cjllanwarne: I'm going to juggle stuff around in an upcoming ticket

src/test/scala/cromwell/util/SampleWdl.scala
```diff
@@ -2222,6 +2222,7 @@ object SampleWdl {
         |  String c = b + "MB"
         |  command { ps }
         |  runtime {
+        |    docker: "ubuntu:latest"
```
cjllanwarne: Was this committed deliberately?

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
```diff
@@ -66,18 +73,28 @@ object CromwellRuntimeAttributes {
...
 
-  private val keys = Set("cpu", "disks", "docker", "zones", "continueOnReturnCode", "failOnStderr", "preemptible", "memory")
+  private case class ValidKeyType(key: String, validTypes: Set[WdlType] = Set(WdlStringType))
+  private val keys = Set(ValidKeyType("cpu", Set(WdlIntegerType, WdlStringType)),
+                        ValidKeyType("disks"),
+                        ValidKeyType("docker"),
+                        ValidKeyType("zones"),
+                        ValidKeyType("continueOnReturnCode", Set(WdlBooleanType, WdlArrayType(WdlIntegerType))),
+                        ValidKeyType("failOnStderr", Set(WdlBooleanType, WdlStringType)),
+                        ValidKeyType("preemptible", Set(WdlIntegerType, WdlStringType)),
```
cjllanwarne: I think this should just be integer. The strings will get coerced down if appropriate?

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
```diff
@@ -17,7 +17,8 @@ import scala.util.Try
 import scalaz.Scalaz._
 import scalaz._
 
-case class CromwellRuntimeAttributes(docker: Option[String],
+case class CromwellRuntimeAttributes(attributes: Map[String, WdlValue],
```
cjllanwarne: Why have an attributes map AND the list of attributes? Can't you just have one or the other?

Horneth: You're right, I think eventually we should only have the map. But right now all the validation logic returns scala types that are used as attributes, which are then used everywhere in Cromwell. As this part of the code is likely to be reworked anyway by intel changes I thought it was not worth rewiring the entire runtime attribute thing to only use a Map[String, WdlValue], or at least not now.

src/test/scala/cromwell/util/SampleWdl.scala
Horneth: Yes, I changed a few of the tests to use a jesBackend instead of a LocalBackend, because now only supported attributes (for a particular backend) are taken into consideration and validated, and some tests are using the "memory" runtime attribute which is only supported by JES. Because of that the corresponding wdls need to have a docker so the validation doesn't fail.

src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
Horneth: Yep

src/test/scala/cromwell/engine/backend/CromwellRuntimeAttributeSpec.scala
```diff
@@ -137,7 +143,7 @@ class CromwellRuntimeAttributeSpec extends FlatSpec with Matchers with EitherVal
     val descriptor = WorkflowDescriptor(WorkflowId(UUID.randomUUID()), SampleWdl.WorkflowWithFullGooglyConfig.asWorkflowSources())
     val call = descriptor.namespace.workflow.callByName("googly_task").get
     val unsupported = CromwellRuntimeAttributes.unsupportedKeys(call.task.runtimeAttributes.attrs.keys, BackendType.LOCAL)
-    unsupported shouldEqual Some("Found unsupported keys for backend 'LOCAL': cpu, disks, memory, zones")
+    unsupported shouldEqual Set("disks", "cpu", "zones", "memory")
```
scottfrazer: I'm not sure I understand the change to this test... why change from local -> jes?  Also shouldn't these keys be allowed for the JES backend?

Horneth: Hmm so these keys are allowed for JES backend, and only the JES backend. Keys that are specified but not supported by the backend are filtered out and not validated nor stored in the DB anymore.
This is the reason why I changed those from local to JES because they use "memory" which is only supported by JES.

scottfrazer: Sorry my confusion was that I thought that line 123 and 146 were part of the same test

scottfrazer:
:+1:

PR #462, opened 2016-02-19
Fix call caching on JES

Horneth:


Horneth:
wheel: @scottfrazer @kshakir

src/main/scala/cromwell/engine/backend/jes/JesBackend.scala
```diff
@@ -361,27 +361,52 @@ case class JesBackend(actorSystem: ActorSystem)
...
+            FailedExecutionHandle(ex).future
+          case Success(_) => postProcess(backendCall) match {
+            case Success(outputs) => backendCall.hash map { h =>
+              SuccessfulExecutionHandle(outputs, Seq.empty[ExecutionEventEntry], backendCall.downloadRcFile.get.stripLineEnd.toInt, h, Option(cachedCall)) }
+            case Failure(ex: AggregatedException) if ex.exceptions collectFirst { case s: SocketTimeoutException => s } isDefined =>
+              // TODO: What can we return here to retry this operation?
+              // TODO: This match clause is similar to handleSuccess(), though it's subtly different for this specific case
+              val error = "Socket timeout occurred in evaluating one or more of the output expressions"
+              log.error(error, ex)
+              FailedExecutionHandle(new Throwable(error, ex)).future
```
kshakir: If we're not going to create our own named exceptions types, I would highly prefer `new RuntimeException` or `new Exception` over `Throwable`.

src/main/scala/cromwell/engine/backend/jes/JesBackend.scala
```diff
@@ -361,27 +361,52 @@ case class JesBackend(actorSystem: ActorSystem)
...
+        JesBackendCall.jesLogStderrFilename _,
+        JesBackendCall.jesLogFilename _,
+        JesBackendCall.jesReturnCodeFilename _
+      )
+
+      val wrongPaths = namesBuilder map { _(cachedCall.key) } map backendCall.callGcsPath.resolve
+      val correctNames = namesBuilder map { _(backendCall.key) }
+
+      (wrongPaths zip correctNames) map {
+        case (wrongPath, correctName) => wrongPath.renameTo(correctName)
```
kshakir: How about something like `cachedPath.renameTo(backendPath)` instead of `wrongPath.renameTo(backendName)`?

kshakir:
:+1:

src/main/scala/cromwell/engine/backend/jes/JesBackend.scala
```diff
@@ -361,27 +361,52 @@ case class JesBackend(actorSystem: ActorSystem)
...
       }
     }
-  } flatten
+
+    def copyingWork(gcs: GoogleCloudStorage) = for {
```
geoffjentry: add type ascription to function

src/main/scala/cromwell/engine/backend/jes/JesBackend.scala
geoffjentry: agreed.

src/main/scala/cromwell/engine/backend/jes/JesBackend.scala
```diff
@@ -361,27 +361,52 @@ case class JesBackend(actorSystem: ActorSystem)
...
+    Future {
+      val log = workflowLoggerWithCall(backendCall)
+      authenticateAsUser(backendCall.workflowDescriptor) { interface =>
+        copyingWork(interface) match {
+          case Failure(ex) =>
+            log.error(s"Exception occurred while attempting to copy outputs from ${cachedCall.callGcsPath} to ${backendCall.callGcsPath}", ex)
+            FailedExecutionHandle(ex).future
+          case Success(_) => postProcess(backendCall) match {
+            case Success(outputs) => backendCall.hash map { h =>
+              SuccessfulExecutionHandle(outputs, Seq.empty[ExecutionEventEntry], backendCall.downloadRcFile.get.stripLineEnd.toInt, h, Option(cachedCall)) }
```
geoffjentry: I realize this is copypasta but can you make that List.empty or replace w/ a more appropriate collection type if list isn't right?

cjllanwarne:
@Horneth just to get a high-level view, which call caching issue does this fix? Is it the stdout/stderr one?

src/main/scala/cromwell/engine/backend/jes/JesBackend.scala
```diff
@@ -361,27 +361,52 @@ case class JesBackend(actorSystem: ActorSystem)
...
+    def renameCallSpecificFiles = {
+      val namesBuilder = List(
+        JesBackendCall.jesLogStdoutFilename _,
+        JesBackendCall.jesLogStderrFilename _,
+        JesBackendCall.jesLogFilename _,
+        JesBackendCall.jesReturnCodeFilename _
+      )
+
+      val wrongPaths = namesBuilder map { _(cachedCall.key) } map backendCall.callGcsPath.resolve
+      val correctNames = namesBuilder map { _(backendCall.key) }
```
cjllanwarne: Could you add a comment to say why this renaming is needed?

cjllanwarne:
I *think* I follow what's going on here but a tiny bit of extra commentary would be great! Assuming I'm following it correctly: :+1:

src/main/scala/cromwell/engine/backend/jes/JesBackend.scala
Horneth: @cjllanwarne  Yep - Because  the log, stdout, stderr and return code files have names specific to the call/shard, call caching breaks when trying to read those files because they get copied as is, with their original name from the to cached call, and the original does not necessarily match the name expected by the the current call. I'll put this in a comment.
We might want to discuss if it's really necessary to have the name of the call + shard ( + attempt that should be there too probably) in the name of those files.

PR #463, opened 2016-02-20
stdout and stderr as symbols.  Closes #430

### LINKED ISSUE #430: Store stdout/stderr as symbols with call execution

mcovarr:
Cromwell piece of stdout and stderr as symbols.

### DIFF
--- a/src/main/scala/cromwell/engine/CallActor.scala
+++ b/src/main/scala/cromwell/engine/CallActor.scala
@@ -7,11 +7,13 @@ import com.google.api.client.util.ExponentialBackOff
 import cromwell.backend.BackendActor.{FailedComputeHashResult, SuccessfulComputeHashResult, ComputeHash, Prepare}
 import cromwell.backend.config.BackendConfiguration
 import cromwell.backend.{BackendActor, DefaultBackendFactory}
+import cromwell.caching.ExecutionHash
 import cromwell.engine.CallActor.{CallActorData, CallActorState}
 import cromwell.engine.backend._
 import cromwell.engine.db.DataAccess._
 import cromwell.engine.db.ExecutionDatabaseKey
 import cromwell.engine.db.slick.Execution
+import cromwell.engine.workflow.WorkflowActor.PersistStandardStreamSymbols
 import cromwell.engine.workflow.{BackendCallKey, WorkflowActor}
 import cromwell.instrumentation.Instrumentation.Monitor
 import cromwell.logging.WorkflowLogger
@@ -48,6 +50,10 @@ object CallActor {
 
   /** Message to self to retry sending the `callMessage` to the `WorkflowActor` in the absence of an `Ack`. */
   final case class Retry(callMessage: WorkflowActor.CallMessage) extends CallActorMessage
+
+  sealed trait StandardStreamPersistenceMessage extends CallActorMessage
+  case object SuccessfulStandardStreamPersist extends StandardStreamPersistenceMessage
+  final case class FailedStandardStreamPersist(error: Throwable) extends StandardStreamPersistenceMessage
 
   /** FSM data class with an optional abort function and exponential backoff. */
   case class CallActorData(abortFunction: Option[AbortFunction] = None,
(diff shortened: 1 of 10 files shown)


mcovarr:
@francares and @kshakir win this one

src/main/scala/cromwell/engine/db/DataAccess.scala
```diff
@@ -47,19 +51,50 @@ trait DataAccess extends AutoCloseable {
...
+
+  /** Get the standard stream symbols for the scope of this call. */
+  def getStandardStreamSymbols(workflowId: WorkflowId, key: ExecutionDatabaseKey)
+                              (implicit ec: ExecutionContext): Future[CallLogs] = {
+
+    // Folds stdout and stderr symbols into an accumulator CallLogs object.
+    def foldLogs(acc: CallLogs, symbol: SymbolStoreEntry): CallLogs = {
+      symbol match {
+        case s if s.key.name.endsWith("$stdout") => acc.copy(stdout = s.wdlValue.get.asInstanceOf[WdlFile])
+        case s if s.key.name.endsWith("$stderr") => acc.copy(stderr = s.wdlValue.get.asInstanceOf[WdlFile])
```
kshakir: Should these be `endsWith(".$stdout")`? Elsewhere the magic seems to be '.$' and not just '$'.

kshakir:
While this PR does add a number of "FIXME" and "TODO" comments that I don't have better suggestions for at the moment-- LGTM.

src/main/scala/cromwell/engine/db/DataAccess.scala
mcovarr: Yes you're right.  As this is used in > 1 place I should really pull it out to a constant.

src/main/scala/cromwell/engine/db/DataAccess.scala
```diff
@@ -47,19 +51,50 @@ trait DataAccess extends AutoCloseable {
...
-  /** Get all outputs for the scope of this call. */
+  /** Get all output symbols for the scope of this call including standard stream symbols. */
+  protected def getAllOutputs(workflowId: WorkflowId, key: ExecutionDatabaseKey)
+                             (implicit ec: ExecutionContext): Future[Traversable[SymbolStoreEntry]]
+
+  /** Get all "real" outputs for the scope of this call (i.e. exclude standard stream symbols). */
   def getOutputs(workflowId: WorkflowId, key: ExecutionDatabaseKey)
-                (implicit ec: ExecutionContext): Future[Traversable[SymbolStoreEntry]]
+                (implicit ec: ExecutionContext): Future[Traversable[SymbolStoreEntry]] = {
+    getAllOutputs(workflowId, key)(ec) map { symbols => symbols filterNot IsStandardStream }
```
mcovarr: could be `_` instead of `symbols => symbols`

src/main/scala/cromwell/engine/CallActor.scala
```diff
@@ -255,6 +257,7 @@ class CallActor(key: BackendCallKey, locallyQualifiedInputs: CallInputs, workflo
     context.stop(self)
   }
 
+  // FIXME this is unused, that doesn't seem right
```
geoffjentry: no, it doesn't :)

cjllanwarne:
Does this have any relation to and/or interfere with the call caching change w.r.t. stdout/stderr by @Horneth?

mcovarr:
I haven't looked at the other PR but I suspect there is a lot of overlap.  The way stdout and stderr are handled in the old world is significantly different than what's required in the new.  It should be possible to add correct caching behavior to these changes, but unit tests for this functionality on develop would likely need to be rewritten.

Horneth:
The other PR is probably to be considered a temporary patch until your change gets merge.

src/main/scala/cromwell/engine/db/DataAccess.scala
francares: +1

src/main/scala/cromwell/engine/db/DataAccess.scala
```diff
@@ -121,4 +156,11 @@ trait DataAccess extends AutoCloseable {
   def updateCallCaching(cachingParameters: CallCachingParameters)(implicit ec: ExecutionContext): Future[Int]
 
   def infosByExecution(id: WorkflowId)(implicit ec: ExecutionContext): Future[Traversable[ExecutionInfosByExecution]]
+
+  def writeStandardStreamSymbols(id: WorkflowId, key: ExecutionDatabaseKey, stdout: String, stderr: String)(implicit ec: ExecutionContext): Future[Unit] = {
+    // Write the magical $stdout and $stderr symbols.
+    val outputs = Map("stdout" -> stdout, "stderr" -> stderr) map {
```
francares: You might want to extract those strings to descriptive const.

francares:
:+1:

mcovarr:
Closing this as it targets pluggable_backends_develop which we are no longer working on, but the ideas here could still be useful for develop.

PR #466, opened 2016-02-22
Add salt to test wdl for RetryableCallsSpec

Horneth:


cjllanwarne:
Is this really easier than turning off call caching for that test?

Horneth:
Oh yeah we could do that too. Call caching should probably only be turned on for call caching test specs actually and turned off by default.

kshakir:
:+1:

cjllanwarne:
:+1:

geoffjentry:
@Horneth bad news, while your branch worked for me the other day this still doesn't work for me when I pull develop

Horneth:
Yeah so I actually changed the fix because @cjllanwarne suggestion seemed cleaner and I though it would have the same effects, but apparently it doesn't. I can change it back to the first version that worked for you.
On a larger point, I think (hope) this kind of failures will be a lot less happening when we re-factor the test suite infrastructure. IMO it's happening because we keep doing more and more complex tricks in the spec to get it to do what we need but with all the features we keep adding it keeps getting less and less stable.

PR #468, opened 2016-02-23
Adding options to copy call and workflow logs.

kshakir:


src/main/scala/cromwell/engine/WorkflowDescriptor.scala
```diff
@@ -55,10 +56,18 @@ case class WorkflowDescriptor(id: WorkflowId,
   private val props = sys.props
   private val relativeWorkflowRootPath = s"$name/$id"
   private val log = WorkflowLogger("WorkflowDescriptor", this)
+  val workflowLogDir = workflowOptions.get("workflow_log_dir") recover { case e: IllegalArgumentException =>
```
geoffjentry: It'd be good if these three didn't just throw since all three could independently be broken

src/main/scala/cromwell/engine/WorkflowDescriptor.scala
```diff
@@ -82,79 +91,116 @@ case class WorkflowDescriptor(id: WorkflowId,
...
   }
 
   lazy val workflowRootPath = wfContext.root.toPath(gcsFilesystem)
-  def workflowRootPathWithBaseRoot(rootPath: String): Path =
+  def workflowRootPathWithBaseRoot(rootPath: String): Path = {
     WorkflowDescriptor.buildWorkflowRootPath(rootPath, name, id).toPath(gcsFilesystem)
+  }
 
-  def copyWorkflowOutputs(implicit executionContext: ExecutionContext): Future[Unit] = {
+  def copyWorkflowOutputs(workflowMetadataResponse: WorkflowMetadataResponse)
```
geoffjentry: Realizing that the reason the stuff is here is the stuff was already here but I was already pretty concerned about how much stuff WorkflowDescriptor has accumulated over time. What about moving all of the copying logic out elsewhere?

src/main/scala/cromwell/engine/backend/package.scala
```diff
@@ -3,6 +3,11 @@ package cromwell.engine
 import scala.concurrent.{ExecutionContext, Future}
 
 package object backend {
+  // Ordered by shards, and then ordered by attempts
+  type AttemptedCallLogs = Seq[Seq[CallLogs]]
```
geoffjentry: we should probably get in the habit of ensuring our `Seq` references are immutable, you can import scala.collections.imutable.Seq to do that

src/main/scala/cromwell/engine/WorkflowDescriptor.scala
geoffjentry: Along those lines (but not what I originally meant) would it make sense for that logic to be handled by the backend?

src/main/scala/cromwell/engine/finalcall/FinalCall.scala
```diff
@@ -1,32 +1,64 @@
...
+import cromwell.engine.backend.{SuccessfulFinalCallExecution, ExecutionHandle}
+import cromwell.engine.workflow.{FinalCallKey, ExecutionStoreKey}
+import cromwell.webservice.WorkflowMetadataResponse
+import wdl4s.{Workflow, Scope}
 
 import scala.concurrent.{ExecutionContext, Future}
 
 object FinalCall {
 
+  private val finalCallCompanions: Seq[FinalCallCompanion[_ <: FinalCall]] = Seq(
```
geoffjentry: If you want this to be a List, make it a List

src/main/scala/cromwell/engine/finalcall/FinalCall.scala
```diff
@@ -34,4 +66,19 @@ trait FinalCall extends Scope {
   def prerequisiteScopes: Set[Scope] = Set.empty
 
   val parent: Option[Scope] = Option(this.rootWorkflow)
+
+  override def rootWorkflow: Workflow = workflow.namespace.workflow
+
+  override def unqualifiedName = FinalCall.name(workflow.name, companion.finalCallName)
+}
+
+trait FinalCallCompanion[A <: FinalCall] {
```
geoffjentry: Can you explain more what this thing is?

cjllanwarne:
Your wheel needs you! @mcovarr @scottfrazer

README.md
```diff
@@ -979,7 +979,9 @@ Valid keys and their meanings:
 
 * **write_to_cache** - Accepts values `true` or `false`.  If `false`, the completed calls from this workflow will not be added to the cache.  See the [Call Caching](#call-caching) section for more details.
 * **read_from_cache** - Accepts values `true` or `false`.  If `false`, Cromwell will not search the cache when invoking a call (i.e. every call will be executed unconditionally).  See the [Call Caching](#call-caching) section for more details.
+* **workflow_log_dir** - Specifies a path where per-workflow logs will be written.  If this is not specified, per-workflow logs will not be generated.
 * **outputs_path** - Specifies a path where final workflow outputs will be written.  If this is not specified, workflow outputs will not be copied out of the Cromwell workflow execution directory/path.
+* **call_logs_dir** - Specifies a path where final call logs will be written.  If this is not specified, call logs will not be copied out of the Cromwell workflow execution directory/path.
```
scottfrazer: Can these two paths be local or GCS paths?

src/main/scala/cromwell/engine/WorkflowDescriptor.scala
kshakir: Yeah, I spent a bit looking at this... these three are independent and the errors are never really "thrown", only logged.

Each value here is actually of type `Try[String]`.  Their `throw`s are actually [caught](https://github.com/scala/scala/blob/v2.11.7/src/library/scala/util/Try.scala#L214-L220) by `recover` and go right back to being the same `Failure`. Still, during each `WorkflowDescriptor` construction, they each independently log if anything is wrong. After construction, at the time of access, the `Failure` is quietly ignored with a `Failure.getOrElse(Future.successful(()))`.

A NEL validating the `WorkflowDescriptor` may be better, but I had to draw a line somewhere when choosing how much to patch in this PR.

Will leave it up to reviewers to decide if this copypasta behavior should be updated as well in this PR?

src/main/scala/cromwell/engine/backend/package.scala
kshakir: Scala collections-- stay weird. _Why tf are Map and Set immutable by default, but not the other collections??_

geoffjentry: It's mainly Seq which is the problem child since it sits waaaaay up there in the hierarchy.

src/main/scala/cromwell/engine/finalcall/FinalCall.scala
kshakir: Just like `GenericCompanion.newBuilder` or `GenericCompanion.empty`, it provides companion, singleton methods. I'll mention this in scaladoc.

The companion object for each `FinalCall` allows instantiation of the class from the singleton, and also includes a unique locally qualified name for each `FinalCall` type.

README.md
kshakir: Yes, that _should_ work... but we don't have automated tests. Let me get back to you? It'll take me a bit to re-setup JES on my machine.

I just refactored and used the existing "copy workflow outputs via `Files.copy`" to a "copy any resolved `Path` via `Files.copy`", internally invoking the existing `GcsFileSystemProvider`.

src/main/scala/cromwell/engine/WorkflowDescriptor.scala
kshakir: Heh, I backed off a number of times from various rabbit holes. You found another one here. :)

Right now, a lot of copying logic is almost magic:rainbow:, using various implicits to inject custom GCS copy handlers into `java.nio.Path` objects. The singleton-ish `Backend` is instead used to validate per-WorkflowDescriptor options and runtime attributes, but does NOT do any actually copying, especially across filesystems.

I expect our new backend repo will create massive changes, and I only _assume_ someone else is already performing surgery to separate backend usage out of the workflow descriptor. But I can continue down that path of moving "file copying" into a/the backend in this PR, if requested.

kshakir:
The wheel has officially replaced ~~@mcovarr~~ with @geoffjentry.

src/main/scala/cromwell/engine/backend/package.scala
kshakir: The more I look at this stuff, the more wtf... Why do `scala.collection.immutable.(Seq|Traversable|Iterable).newBuilder` each return a `mutable.ListBuffer`?

scottfrazer:
I noticed some new output on this branch

```
[2016-02-24 12:49:51,571] [info] WorkflowActor [6ce7303c]: persisting status of w.$final_call$copy_workflow_log to Starting.
[2016-02-24 12:49:51,571] [info] WorkflowActor [6ce7303c]: persisting status of w.$final_call$copy_workflow_outputs to Starting.
[2016-02-24 12:49:51,572] [info] WorkflowActor [6ce7303c]: persisting status of w.$final_call$copy_call_logs to Starting.
[2016-02-24 12:49:51,697] [info] WorkflowActor [6ce7303c]: no inputs for call 'w.$final_call$copy_workflow_log'
[2016-02-24 12:49:51,711] [info] WorkflowActor [6ce7303c]: no inputs for call 'w.$final_call$copy_workflow_outputs'
[2016-02-24 12:49:51,711] [info] WorkflowActor [6ce7303c]: no inputs for call 'w.$final_call$copy_call_logs'
[2016-02-24 12:49:51,916] [info] WorkflowActor [6ce7303c]: created call actor for w.$final_call$copy_workflow_log.
[2016-02-24 12:49:51,917] [info] WorkflowActor [6ce7303c]: created call actor for w.$final_call$copy_workflow_outputs.
[2016-02-24 12:49:51,917] [info] WorkflowActor [6ce7303c]: created call actor for w.$final_call$copy_call_logs.
[2016-02-24 12:49:51,917] [info] WorkflowActor [6ce7303c]: persisting status of w.$final_call$copy_call_logs to Running.
[2016-02-24 12:49:51,917] [info] WorkflowActor [6ce7303c]: persisting status of w.$final_call$copy_workflow_log to Running.
[2016-02-24 12:49:51,918] [info] WorkflowActor [6ce7303c]: persisting status of w.$final_call$copy_workflow_outputs to Running.
[2016-02-24 12:49:52,42] [info] WorkflowActor [6ce7303c]: persisting status of w.$final_call$copy_call_logs to Done.
[2016-02-24 12:49:52,72] [info] WorkflowActor [6ce7303c]: persisting status of w.$final_call$copy_workflow_outputs to Done.
[2016-02-24 12:49:52,80] [info] WorkflowActor [6ce7303c]: persisting status of w.$final_call$copy_workflow_log to Done.
```

I'm not sure I like the strings like `w.$final_call$copy_workflow_log`... maybe I'll understand soon why they are that way but do you think there's any way to make these better?

scottfrazer:
In `Main.initLogging` we still do this

```scala
if (systemProperties.get("LOG_ROOT").isDefined) {
  Console.err.println("WARNING: The LOG_ROOT parameter is currently ignored.")
}
```

Is that relevant anymore?  WorkflowDescriptor still uses `LOG_ROOT`

src/main/scala/cromwell/engine/WorkflowDescriptor.scala
geoffjentry: eh, ignore me. This class is already a disaster anyways

src/main/scala/cromwell/engine/WorkflowDescriptor.scala
geoffjentry: This one I have more trouble with although I'm also having difficulty staying self-consistent between my idealist and realist hats. I'm getting a little leery of continuing to throw any old thing into workflowdescriptor on the hope that it gets cleaned up as part of the other sweeping changes.

OTOH over the last couple of days it is looking increasingly likely that the plan will be "make it work, now fix it" in regards to the backend changes so it is more likely that cleaning this up will actually happen

scottfrazer:
I am using this code and trying various configurations to write workflow logs, but I find myself confused a lot...

Currently the README has [three environment variables](https://github.com/broadinstitute/cromwell/blob/ks_copy_logs/README.md#logging) listed (`LOG_ROOT`, `LOG_MODE`, `LOG_LEVEL`)

It appears that `logback.xml` expects `LOG_MODE` to be `pretty` or `standard` but `WorkflowDescriptor` is expecting it to be `server`.

I also couldn't figure out how to both output logs to stdout and also write to a workflow log (a feature I'd be particularly interested in!)

Also it appears that we only honor `LOG_ROOT` when `LOG_MODE` is `server`.

Can we maybe have a tech talk whenever you get a chance to understand how all these options play together?

src/main/scala/cromwell/engine/callexecution/FinalCallExecutionActor.scala
```diff
@@ -21,8 +23,9 @@ case class FinalCallExecutionActor(override val call: FinalCall) extends CallExe
 
   override def poll(handle: ExecutionHandle) = call.poll(ec, handle)
   override def execute(mode: ExecutionMode) = mode match {
-    case UseCachedCall(cachedBackendCall) => Future.failed(new UnsupportedOperationException("Cannot use cached results for a FinalCall"))
-    case _ => call.execute
+    case UseCachedCall(cachedBackendCall) =>
+      Future.failed(new UnsupportedOperationException("Cannot use cached results for a FinalCall"))
+    case _ => call.execute(workflowMetadataResponse) map (_ => call.handle)
```
geoffjentry: should be {} and not ()?

src/main/scala/cromwell/engine/finalcall/FinalCall.scala
```diff
@@ -1,32 +1,64 @@
...
+import cromwell.engine.backend.{SuccessfulFinalCallExecution, ExecutionHandle}
+import cromwell.engine.workflow.{FinalCallKey, ExecutionStoreKey}
+import cromwell.webservice.WorkflowMetadataResponse
+import wdl4s.{Workflow, Scope}
 
 import scala.concurrent.{ExecutionContext, Future}
 
 object FinalCall {
 
+  private val finalCallCompanions: Traversable[FinalCallCompanion[_ <: FinalCall]] = List(
```
geoffjentry: bleh, the danger of my war on Seq. You all start using traversable and iterable :)

geoffjentry: BTW note that traversable and iterable ahve the same issue as Seq

src/main/scala/cromwell/engine/finalcall/FinalCall.scala
```diff
@@ -1,32 +1,64 @@
...
+    "%s.$final_call$%s".format(workflowName, finalCallName)
+  }
+
+  def isCallFqn(fqn: String, finalCallName: String) = fqn.endsWith(unqualifiedName(finalCallName))
 }
 
 /** Scope representing a "final call" that is inserted after all other workflow executions. */
 trait FinalCall extends Scope {
+  val companion: FinalCallCompanion[_ <: FinalCall]
+
```
geoffjentry: remove superfluous \ns

README.md
scottfrazer: > If this is not specified, call logs will not be copied out of the Cromwell workflow execution directory/path.

I'm having trouble understanding this

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -703,13 +702,23 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
...
-      case backendCallKey: BackendCallKey => CallActor.props(backendCallKey, locallyQualifiedInputs, backend, workflow)
-      case finalCallKey: FinalCallKey => CallActor.props(finalCallKey)
+    val futureCallActorProps = callKey match {
+      case backendCallKey: BackendCallKey =>
+        Future.successful(CallActor.props(backendCallKey, locallyQualifiedInputs, backend, workflow))
+      case finalCallKey: FinalCallKey =>
+        WorkflowMetadataBuilder.workflowMetadata(workflow.id, backend) map {
+          CallActor.props(finalCallKey, _)
+        }
+    }
```
geoffjentry: \n

scottfrazer:
Here is a random proposal...

`LOG_MODE` can be one of `pretty`, `standard`, or `none`.  This determines how Cromwell will log to stdout.  Defaults are `standard` for server and `pretty` for command line.

Then, independent of `LOG_MODE`, users can turn on workflow logging by setting `LOG_ROOT`, which can be overridden via workflow option (currently called `workflow_log_dir`, but perhaps we can synchronize the names of `LOG_ROOT` and `workflow_log_dir`??).  If neither are set, then don't write workflow log.

thoughts?

README.md
scottfrazer: Also I can test the JES integration for you since I'm all set up to do it

src/main/scala/cromwell/engine/workflow/WorkflowMetadataBuilder.scala
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  private def hasLogs(entries: Iterable[ExecutionDatabaseKey])(key: ExecutionDatabaseKey) = {
+    !key.fqn.isScatter && !key.isCollector(entries)
+  }
+
+  def workflowStdoutStderr(workflowId: WorkflowId, backend: Backend,
+                           workflowDescriptor: WorkflowDescriptor,
+                           executionStatuses: Map[ExecutionDatabaseKey, CallStatus]): WorkflowLogs = {
+    val statusMap = executionStatuses filterKeys (!_.fqn.isFinalCall) mapValues (_.executionStatus)
```
geoffjentry: () => {}?

geoffjentry:
@scottfrazer IIRC the original use case specified that one turned on workflow logging via workflow option, not at a server level

geoffjentry:
Semi :+1:  ... I want Scott's various logging things hashed out and I wasn't particularly thorough, trying to balance not delaying you too long, so it'd be good to get another quick pair of eyes or something like that

kshakir:
Switched reviewer back from ~~@geoffjentry~~ to @mcovarr.

README.md
```diff
@@ -842,7 +853,9 @@ Valid keys and their meanings:
 
 * **write_to_cache** - Accepts values `true` or `false`.  If `false`, the completed calls from this workflow will not be added to the cache.  See the [Call Caching](#call-caching) section for more details.
 * **read_from_cache** - Accepts values `true` or `false`.  If `false`, Cromwell will not search the cache when invoking a call (i.e. every call will be executed unconditionally).  See the [Call Caching](#call-caching) section for more details.
+* **workflow_log_dir** - Specifies a path where per-workflow logs will be written.  If this is not specified, per-workflow logs will not be copied out of the Cromwell workflow log temporary directory/path before they are deleted.
 * **outputs_path** - Specifies a path where final workflow outputs will be written.  If this is not specified, workflow outputs will not be copied out of the Cromwell workflow execution directory/path.
+* **call_logs_dir** - Specifies a path where final call logs will be written.  If this is not specified, call logs will not be copied out of the Cromwell workflow execution directory/path.
```
kshakir: @scottfrazer Re:

> If this is not specified, call logs will not be copied out of the Cromwell workflow execution directory/path.

Let me know if this is still not clear enough. The sentence was mostly copied from the one above, but happy to have it reworded to be clearer. Also I tested basic functionality of copying to/from gcs and local, so let me know if you see unexpected behavior?

src/main/scala/cromwell/Main.scala
```diff
@@ -53,13 +53,29 @@ object Main extends App {
...
+    val config = ConfigFactory.load
+      .withFallback(ConfigFactory.systemEnvironment())
+      .withFallback(ConfigFactory.parseMap(defaultProps.asJava, "Defaults"))
+
+    val props = sys.props
+    defaultProps.keys foreach { key =>
+      props += key -> config.getString(key)
+    }
+
+    ConfigFactory.invalidateCaches()
```
mcovarr: I'm not seeing why this is called

src/main/scala/cromwell/engine/WorkflowDescriptor.scala
mcovarr: In the interest of getting this merged sometime soon I'm going to pretend I didn't see this discussion

src/main/scala/cromwell/engine/finalcall/CopyWorkflowLog.scala
```diff
@@ -0,0 +1,26 @@
...
+  override val finalCallName = "copy_workflow_log"
+
+  override def createCall(workflow: WorkflowDescriptor) = CopyWorkflowLogCall(workflow)
+}
+
+/**
+  * Final call implementation that copies the workflow log to a specified destination.
+  */
+case class CopyWorkflowLogCall(override val workflow: WorkflowDescriptor) extends FinalCall {
+  override val companion = CopyWorkflowLogCall
```
mcovarr: crafty :smile:

src/main/scala/cromwell/engine/finalcall/FinalCall.scala
mcovarr: is this ascription really necessary?

src/main/scala/cromwell/engine/finalcall/FinalCall.scala
```diff
@@ -1,37 +1,89 @@
...
+      finalCallCompanions.find(companion => isCallFqn(fqn, companion.finalCallName)) match {
+        case Some(companion) => FinalCallKey(companion.createCall(workflow))
+        case None => throw new IllegalArgumentException(s"Cannot convert $fqn into a FinalCall store key for $workflow")
+      }
     }
   }
+
+  def unqualifiedName(finalCallName: String): LocallyQualifiedName = "$final_call$" + finalCallName
+
+  def name(workflowName: String, finalCallName: String): LocallyQualifiedName = {
```
mcovarr: return type doesn't seem right

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -736,13 +735,24 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
...
+
+    futureCallActorProps andThen {
```
mcovarr: This method returns `Unit`, why not `onComplete`?

src/main/scala/cromwell/engine/workflow/WorkflowMetadataBuilder.scala
```diff
@@ -0,0 +1,164 @@
...
+      _.executionStatus
+    }
+
+    val sortedMap = statusMap.toSeq.sortBy(_._1.index)
+    val callsToPaths = for {
+      (key, status) <- sortedMap if !key.fqn.isFinalCall && hasLogs(statusMap.keys)(key)
+      callName = assertCallFqnWellFormed(workflowDescriptor, key.fqn)
+      call = workflowDescriptor.namespace.workflow.findCallByName(callName).get
+      callKey = BackendCallKey(call, key.index, key.attempt)
+      // TODO There should be an easier way than going as far as backend.bindCall just to retrieve stdout/err path
```
mcovarr: yeah the pluggable backend version of this was a good bit less horrible, hopefully something like that will be coming here soon

src/main/scala/cromwell/engine/workflow/WorkflowMetadataBuilder.scala
```diff
@@ -0,0 +1,164 @@
...
+    import cromwell.engine.db.DataAccess.globalDataAccess
+
+    for {
+      workflowState <- globalDataAccess.getWorkflowState(id)
+      // TODO: This assertion could also be added to the db layer: "In the future I'll fail if the workflow doesn't exist"
+      _ <- assertWorkflowExistence(id, workflowState)
+      workflowExecution <- globalDataAccess.getWorkflowExecution(id)
+      workflowOutputs <- globalDataAccess.getWorkflowOutputs(id)
+      // TODO: What was this next line doing? Is it because of cached results? For now shortened it. Test and make sure they are equivalent
+      //workflowExecutionAux <- globalDataAccess.getWorkflowExecutionAux(WorkflowId.fromString(workflowExecution.workflowExecutionUuid))
```
mcovarr: I think that code may have been confused

src/main/scala/cromwell/engine/workflow/WorkflowMetadataBuilder.scala
```diff
@@ -0,0 +1,164 @@
...
+    val callMetadata = CallMetadataBuilder.build(nonFinalInfosByExecution, callStandardStreamsMap, callInputs,
+      callOutputs, nonFinalEvents, runtimeAttributes)
+    buildWorkflowMetadata(workflowExecution, workflowExecutionAux, engineWorkflowOutputs, callMetadata)
+  }
+
+  def workflowMetadata(id: WorkflowId, backend: Backend)
+                      (implicit ec: ExecutionContext): Future[WorkflowMetadataResponse] = {
+
+    // TODO: This entire block of chained database actions should be a single request to the db layer.
+    // Also, because it's read only, this would be a candidate for a separate, optimized data service.
```
mcovarr: not sure what you mean by the second comment?

src/test/scala/cromwell/engine/WorkflowDescriptorSpec.scala
```diff
@@ -76,13 +85,50 @@ class WorkflowDescriptorSpec extends FlatSpec with Matchers {
 
   }
 
+  def workflowFile(descriptor: WorkflowDescriptor, file: String): WdlFile = {
+    val path = descriptor.wfContext.root.toPath().resolve(file).toAbsolutePath
+    path.createIfNotExists() // TODO: Delete the files and directories!
```
mcovarr: is this still a TODO?

src/main/scala/cromwell/engine/WorkflowDescriptor.scala
```diff
@@ -52,24 +54,55 @@ case class WorkflowDescriptor(id: WorkflowId,
...
   val actualInputs: WorkflowCoercedInputs = coercedInputs ++ declarations
-  private val props = sys.props
   private val relativeWorkflowRootPath = s"$name/$id"
   private val log = WorkflowLogger("WorkflowDescriptor", this)
-  val workflowOutputsPath = workflowOptions.get("outputs_path") recover { case e: IllegalArgumentException =>
-    log.warn("outputs_path expected to be of type String", e)
+  val workflowLogDir = workflowOptions.get(WorkflowLogDirOptionKey) recover { case e: IllegalArgumentException =>
+    log.warn(s"$WorkflowLogDirOptionKey expected to be of type String", e)
+    throw e
+  }
```
mcovarr: Are callers not catching and logging this again?

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -736,13 +735,24 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
...
 
     val callActorName = s"CallActor-${workflow.id}-${callKey.tag}"
-    val callActorProps = callKey match {
-      case backendCallKey: BackendCallKey => CallActor.props(backendCallKey, locallyQualifiedInputs, backend, workflow)
-      case finalCallKey: FinalCallKey => CallActor.props(finalCallKey)
+    val futureCallActorProps = callKey match {
+      case backendCallKey: BackendCallKey =>
+        Future.successful(CallActor.props(backendCallKey, locallyQualifiedInputs, backend, workflow))
+      case finalCallKey: FinalCallKey =>
+        WorkflowMetadataBuilder.workflowMetadata(workflow.id, backend) map {
```
mcovarr: Urgh, grabbing all the workflow metadata seems like a pretty expensive way of getting the couple of items we care about for these final calls.  Hopefully again the pluggable backends work can make this better.

src/main/scala/cromwell/engine/workflow/WorkflowMetadataBuilder.scala
```diff
@@ -0,0 +1,164 @@
...
+                                    workflowExecution: WorkflowExecution,
+                                    workflowOutputs: Traversable[SymbolStoreEntry],
+                                    workflowExecutionAux: WorkflowExecutionAux,
+                                    workflowDescriptor: WorkflowDescriptor,
+                                    executionStatuses: Map[db.ExecutionDatabaseKey, db.CallStatus],
+                                    callInputs: Traversable[SymbolStoreEntry],
+                                    callOutputs: Traversable[SymbolStoreEntry],
+                                    infosByExecution: Traversable[ExecutionInfosByExecution],
+                                    runtimeAttributes: Map[ExecutionDatabaseKey, Map[String, String]],
+                                    executionEvents: Map[db.ExecutionDatabaseKey, Seq[engine.ExecutionEventEntry]]):
```
mcovarr: This is the same `ExecutionDatabaseKey` type on both lines isn't it?

mcovarr:
:+1: although I could not agree more with @geoffjentry's sentiments on slimming down `WorkflowDescriptor`, I think that's better done in pluggable backend PRs or maybe completely separate PRs.

src/main/scala/cromwell/Main.scala
kshakir: Adding an comment to the code:

>We've possibly copied values from the environment, or our defaults, into the system properties. Make sure that the next time one uses the ConfigFactory that our updated system properties are loaded.

src/main/scala/cromwell/engine/workflow/WorkflowMetadataBuilder.scala
kshakir: Removing the second comment, but it was a just a note that we have room for optimization our database interactions. Specifically, we could have a separate readonly database accessor with its own thread pool.

src/test/scala/cromwell/engine/WorkflowDescriptorSpec.scala
kshakir: Yes. After the tests run, these specs do NOT delete the files. This is actually pretty common in a number of tests.

src/main/scala/cromwell/engine/WorkflowDescriptor.scala
kshakir: Nope. These objects are of type `Try[String]`. The exception is logged once here, and then the object is pretty much treated by callers as an `Option[String]` equal to `None`. They generally call `map` with at `getOrElse`, or `match` on the `Failure[_]` ignoring the error.

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
kshakir: Yes, yes it is. But I couldn't figure out an easier/simpler way to have a common interface for FinalCallActors that accepted different objects depending on the various types of FinalCallActor.

README.md
scottfrazer: :+1: for wording

src/main/resources/application.conf
```diff
@@ -36,6 +36,12 @@ workflow-options {
 
   // AES-256 key to use to encrypt the values in `encrypted-fields`
   base64-encryption-key: "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA="
+
+  // Directory where to write per workflow logs
+  workflow-log-dir: "cromwell-workflow-logs"
+
+  // When true, per workflow logs will be deleted after copying
+  workflow-log-temporary: true
```
scottfrazer: hmm is `true` a good default value for this?

I tried running a workflow through a vanilla Cromwell and it created the `cromwell-workflow-logs` and indeed did remove the logs after the workflow finished... but it seems to me like the default behavior should be to keep workflow logs and set this option to true for the Firecloud case

scottfrazer: I don't know, I vacillate on this.  I'm trying to think about what a user who downloads the JAR file and doesn't dive deep into the configuration would expect

geoffjentry: If I was a naive user I would be streaming my logs to console via pretty mode and not expecting files to be stored

scottfrazer: Fair point... especially if you're using the command-line runner.

I'm just thinking... Apache or MySQL do by default?  Well all server software I can think of logs to a file by default.  Perhaps that should be Cromwell's default behavior when in server mode?

scottfrazer: Now that I think about it I don't actually *know for sure* what the default logging behavior is for other big-name servers.  I just have a feeling (anecdata?) that usually I don't have to turn on file-logging by default.  I'll try to investigate out of curiosity when I have some time today

geoffjentry: Fwiw I know people who refuse to use software which writes files they didn't request due to them supposedly being a Trojan. There's no one right answer.

Personally I don't like stuff leaving detritus I don't know about but I wouldn't call something a Trojan for doing so :)

scottfrazer: yeah, we all have our personal preferences there are arguments for both sides, but perhaps if there's a clear bias one way or another among the big-name server software all having file logging on or off by default, we'd better meet user's expectations if we did what they did.

scottfrazer: Also, Cromwell logs can actually get quite large (4.4GB for `/var/log/messages-20160229` on GOTC).  Good argument for keeping it off by default.

scottfrazer:
I've been playing around with different settings and testing this out a lot.  I like it!

I think I might have found a bug with writing to GCS buckets.  I'm using this as my options file:

```
{
  "workflow_log_dir": "gs://sfrazer-dev/foobar",
  "call_logs_dir": "gs://sfrazer-dev/foobar/calls"
}
```

And I ran a workflow both locally and with JES.  In both cases it seemed to write the call logs just fine.

However, it seems the workflow log to lost somewhere.  It did create a file `gs://sfrazer-dev/foobar` with the contents of the file being `foobar`.

I also got this odd error message.  Indeed the stderr file that was uploaded was zero bytes.  However, it is supposed to be zero bytes!  I'll try again and send something else to stderr.  Might just be a spurious error when your files happen to be zero bytes...

```
[2016-03-04 10:02:32,132] [info] JesBackend [7beff6f6]: Trying to copy output file gs://cromwell-dev/w/7beff6f6-0b44-4b34-8b20-49868740f235/call-arr/arr-stderr.log to gs://sfrazer-dev/foobar/calls/w/7beff6f6-0b44-4b34-8b20-49868740f235/call-arr/arr-stderr.log
[2016-03-04 10:02:34,490] [info] Got 'range not satisfiable' for reading gs://cromwell-dev/w/7beff6f6-0b44-4b34-8b20-49868740f235/call-arr/arr-stderr.log at position 0; assuming empty.
[2016-03-04 10:02:34,692] [info] JesBackend [7beff6f6]: Trying to copy output file gs://cromwell-dev/w/7beff6f6-0b44-4b34-8b20-49868740f235/call-arr/arr.log to gs://sfrazer-dev/foobar/calls/w/7beff6f6-0b44-4b34-8b20-49868740f235/call-arr/arr.log
```

src/main/resources/application.conf
geoffjentry: For now, let's delete by default and if that does manage to cause calamity we can consider changing

scottfrazer:
As I thought, that message was nothing.  Once stderr had more than zero bytes it doesn't show up (and the file on GCS is correct)

scottfrazer:
Once the issue with GCS workflow logs is fixed, then this is good to go!! :+1:

kshakir:
Fixed the workflow log duplication in the latest commit. @mcovarr, I'm going to assume you're ok with the latest diff if @scottfrazer gives his approval.

mcovarr:
yeah absolutely if it's good enough for @scottfrazer it's certainly good enough for me :smile:

mcovarr:
@coreone has to approve this??

coreone:
Approved

@mcovarr my name is only attached here because I enabled the PullApprove plugin.  Chris turned off the "required" piece, so you should still be able to merge.  Once a .pullapprove.yml file is in place, it will have the correct people who should actually be doing the approvals.

mcovarr:
:+1: I endorse this pull request

cjllanwarne:
This already has a plethora of thumbs up so I'll give PullApprove a kick with :+1:

PR #471, opened 2016-02-23
Fix empty string output

Horneth:


src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -396,23 +402,26 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
...
+        self ! CallFailedNonRetryable(callKey, executionEvents, Option(returnCode), e.getMessage)
+      case Success(_) =>
+        self ! PersistStatus(callKey, ExecutionStatus.Done, Option(callOutputs), Option(returnCode), hash, resultsClonedFrom)
     }
 
-    val updatedData = data.addPersisting(callKey, ExecutionStatus.Done)
-    stay() using updatedData
+    /* Ack now to the Call Actor that its message is being processed,
+     * we don't need to actually wait for the status to be persisted as this is internal to the WA processing algorithm.
+     */
```
mcovarr: hrm?  Can't calls now end up without their `Done` statuses written to the db?

mcovarr: actually is this even part of the empty string PR?

cjllanwarne:
@cjllanwarne @geoffjentry

src/test/scala/cromwell/util/SampleWdl.scala
```diff
@@ -124,6 +124,33 @@ object SampleWdl {
...
+        |  }
+        |}
+      """.stripMargin.replaceAll("RUNTIME", runtime)
+
+    val rawInputs = Map.empty[String, Any]
+    val OutputKey = "hello.hello.empty"
+    val OutputValue = ""
+  }
+
+
```
cjllanwarne: As I understand it, this problem only manifests on mysql so this test won't actually check anything... I don't have an answer to this, just commenting...

src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -523,6 +532,10 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
       _ <- dataAccess.symbolsAutoInc ++= callOutputs map {
         case (symbolLocallyQualifiedName, CallOutput(wdlValue, hash)) =>
           val reportableSymbol = key.index.fromIndex == -1 && reportableResultNames.contains(key.scope.fullyQualifiedName + "." + symbolLocallyQualifiedName)
+          val value = wdlValue match {
+            case x if wdlValueToDbValue(x).isEmpty => None
+            case x => Option(wdlValueToDbValue(x).toClob)
+          }
```
cjllanwarne: Extract this duplicated code into a function

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -375,16 +390,17 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
...
     } yield ()
 
-    completionWork onFailure {
-      case e =>
+    completionWork onComplete {
+      case Failure(e) =>
         logger.error(s"Preemption work failed for call ${callKey.tag}! " + e.getMessage, e)
         self ! CheckForWorkflowComplete
+      case Success(_) =>
+        self ! PersistStatus(callKey, retryStatus, None, returnCode, None, None, currentSender, message)
```
cjllanwarne: This is fixing a different problem?

src/test/scala/cromwell/util/SampleWdl.scala
Horneth: Correct, or it will check that this still works on hsql...

src/test/scala/cromwell/util/SampleWdl.scala
```diff
@@ -124,6 +124,33 @@ object SampleWdl {
     val OutputKey = "goodbye.goodbye.out"
   }
 
+  object EmptyString extends SampleWdl {
+    override def wdlSource(runtime: String = "") =
+      """
+        |task hello {
+        |  command {
```
cjllanwarne: The code also handles empty input strings, so might as well have one of those too

src/test/scala/cromwell/util/SampleWdl.scala
cjllanwarne: Fair point!

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
Horneth: Yep different problem but from the same bug ticket.
Because the setOutputs was failing (because of the empty string), the status was never set to "Failed", so the CallActor keep sending its `CallCompleted` message because it was never getting an ack which led to infinite retries of `setOutputs` etc...

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
Horneth: @mcovarr thinking more about this, I think it's actually ok to send the Ack to the CallActor here, instead of waiting for the actual Status persistence to happen. 
If I understand, the reason there is a retry/ack mechanism in the first place is to avoid calling this method before the call has been persisted to "Running", while still ensuring that it will be processed eventually, which makes sense.
But once this message goes through, and the `handleCallCompleted` is called, I don't see why we need to wait for the WA to do its persistence dance (outputs and then status) before Acking to the CA. The WA could potentially do all kind of things / verification etc.. when processing the `CallCompleted` message, but the CA only cares to make sure that this processing happens, it doesn't really need to wait for it to complete.

geoffjentry:
:+1:

Horneth:
@cjllanwarne was your assignment handover to Jeff an implicit :+1: or are you still reviewing ?

geoffjentry:
@Horneth an example of how difficult implicits can make things to reason about

Horneth:
@geoffjentry implicits are great as long as they always work. And since it's implicitly assumed that they do, QED

geoffjentry:
Voltaire, Descartes ... Jeandet?

Horneth:
I've always known deep down that I was born in the wrong century, wait millennium !

cjllanwarne:
:+1:

pgrosu:
@Horneth You're not alone on that one - I have the same feeling many times :)  @geoffjentry It probably is more inference by circumscription by John McCarthy :)

PR #477, opened 2016-02-26
Small retry fix

Horneth:


src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -385,7 +385,6 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
                                   events: Seq[ExecutionEventEntry]) = {
     val currentSender = sender()
     val keyClone = callKey.retryClone
-    executionStore += keyClone -> ExecutionStatus.NotStarted
```
Horneth: I think this might have been causing the transient Retryable Spec failures. The cloned key was inserted in the ExecutionStore before the async DB persistence completed, which might lead to the call being started by the WA before it's actually in the DB.

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -396,8 +395,9 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
       case Failure(e) =>
         logger.error(s"Preemption work failed for call ${callKey.tag}! " + e.getMessage, e)
         currentSender ! CallActor.Ack(message)
-        self ! CheckForWorkflowComplete
+        self ! CallFailedNonRetryable(callKey, events, returnCode, e.getMessage)
```
Horneth: This fails the call and doesn't leave it hanging "Running" if the Preemption work fails.

Horneth:
Wheel: @mcovarr @scottfrazer

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -396,8 +395,9 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
...
       case Success(_) =>
+        executionStore += keyClone -> ExecutionStatus.NotStarted
```
mcovarr: This is not safe, it's modifying the executionStore in an async callback and not in the WA's message processing thread.

Horneth: Hmm ok - We're going to need yet another new message then

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
mcovarr: I think you're right this is causing the error but I'm surprised that dependency exists.  I need to dig a bit more on this one...

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
Horneth: Oh, I think this is actually the wrong order of things, the call should be persisted to Preempted before it creates its clone, because the way it is now if Cromwell is stopped in between we're going to end up with the Preempted call still `Running` and its clone `NotStarted`, which is not good.

mcovarr: yeah that would not be good at all :smile:

scottfrazer:
Not so small anymore ;)

Horneth:
Yeah it was actually worse than I thought, it was not terribly broken but not robust to all edge cases either. I can explain to you if you want.

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -376,35 +375,44 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
...
         self ! CheckForWorkflowComplete
-      case Success(_) =>
-        self ! PersistStatus(callKey, retryStatus, None, returnCode, None, None, currentSender, message)
     }
 
-    val updatedData = data.addProcessing(callKey)
+    data.addPersisting(keyClone, ExecutionStatus.NotStarted)
+  }
+
+  private def handleCallFailedRetryable(callKey: OutputKey,
```
scottfrazer: I'm not sure about this function name... is `handleRetryableCallFailed` better?

scottfrazer: Oh I guess this function name corresponds to the name of the message

Horneth: Hmm not really, `handleCallFailedRetryable` does sound better

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -376,35 +375,44 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
...
-    } yield ()
+    val insertCopy = globalDataAccess.insertCalls(workflow.id, Seq(keyClone), backend)
 
-    completionWork onComplete {
-      case Failure(e) =>
-        logger.error(s"Preemption work failed for call ${callKey.tag}! " + e.getMessage, e)
-        currentSender ! CallActor.Ack(message)
+    insertCopy map { _ => PersistenceSucceeded(keyClone, ExecutionStatus.NotStarted) } pipeTo self
+
+    insertCopy onFailure {
```
mcovarr: deprecated in Scala 2.12!

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -215,6 +215,10 @@ object WorkflowActor {
   val AkkaTimeout = 5 seconds
 
   type ExecutionStore = Map[ExecutionStoreKey, ExecutionStatus]
+  implicit class EnhancedStore(val es: ExecutionStore) extends AnyVal {
+    // An entry is Done if one of its attempts is Done
+    def isDone(e: ExecutionStoreEntry) = es collect { case(k, s) if k.scope == e._1.scope && k.index == e._1.index => s } exists { _ == ExecutionStatus.Done }
```
mcovarr: makes more sense as a `find`

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
Horneth: Not yet then ! :p

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -376,35 +375,44 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
...
 
-    completionWork onComplete {
-      case Failure(e) =>
-        logger.error(s"Preemption work failed for call ${callKey.tag}! " + e.getMessage, e)
-        currentSender ! CallActor.Ack(message)
+    insertCopy map { _ => PersistenceSucceeded(keyClone, ExecutionStatus.NotStarted) } pipeTo self
+
+    insertCopy onFailure {
+      case e =>
+        logger.error(s"Failed to clone ${callKey.tag} for retry attempt " + e.getMessage, e)
```
mcovarr: is this going to result in the cloning being attempted again, potentially successfully?

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
geoffjentry: For reference: https://github.com/viktorklang/blog/blob/master/Futures-in-Scala-2.12-part-5.md

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
Horneth: Yes or even
`es exists { case (k, s) => k.scope == e._1.scope && k.index == e._1.index && s == ExecutionStatus.Done }` actually

mcovarr: yeah that's even better

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -376,35 +375,44 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
...
         self ! CheckForWorkflowComplete
```
scottfrazer: I'm not sure why we `CheckForWorkflowComplete` if an insert of a retryable call into the DB fails...

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
mcovarr: agree with Scott's suggestion

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -376,35 +375,44 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
...
+                                        retryStatus: ExecutionStatus,
+                                        returnCode: Option[Int],
+                                        failure: Throwable,
+                                        message: CallFailedRetryable,
+                                        data: WorkflowData,
+                                        events: Seq[ExecutionEventEntry]) = {
+    val currentSender = sender()
+    val persistEvents = globalDataAccess.setExecutionEvents(workflow.id, callKey.scope.fullyQualifiedName, callKey.index, callKey.attempt, events)
+
+    persistEvents onFailure {
```
mcovarr: another one living on borrowed time

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
Horneth: No :/ 
Retrying would mean implementing retry on database query failure which we do nowhere, but I guess in this particular case it's critical enough to trigger a retry ?

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -510,15 +518,18 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
       stay using finalData
     case Event(PersistenceSucceeded(callKey, ExecutionStatus.Preempted, callOutputs), data) =>
       executionStore += callKey -> ExecutionStatus.Preempted
-      val updatedData = startRunnableCalls(data)
-      val finalData = updatedData.removePersisting(callKey, ExecutionStatus.Preempted)
-      self ! CheckForWorkflowComplete
-      stay using finalData
+      val updatedData = retryCall(callKey, data).removePersisting(callKey, ExecutionStatus.Preempted)
```
scottfrazer: The last thing I'm not really sure I understand is why we seem to do `doSomething.removePersisting(callKey, ExecutionStatus.Preempted)` a few times in this code

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
Horneth: If this is the last call of the workflow, it gets Preempted, WA tries to retry it but fails, then it's going to get stuck `Running` while nothing happens ?

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -376,35 +375,44 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
...
+                        data: WorkflowData) = {
+
     val keyClone = callKey.retryClone
-    executionStore += keyClone -> ExecutionStatus.NotStarted
 
-    val completionWork = for {
-      _ <- globalDataAccess.setExecutionEvents(workflow.id, callKey.scope.fullyQualifiedName, callKey.index, callKey.attempt, events)
-      _ <- globalDataAccess.insertCalls(workflow.id, Seq(keyClone), backend)
-    } yield ()
+    val insertCopy = globalDataAccess.insertCalls(workflow.id, Seq(keyClone), backend)
```
geoffjentry: If you're making a List, might as well call it one :)

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
Horneth: Where else do you see this ?

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
geoffjentry: Guessing here on the reason it was put there (and not saying the logic is actually correct) but I'd bet that the thought was that if that led to the workflow failing you'd want to detect it

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
mcovarr: Well no need to do anything heroic for this one particular case, we have a general problem with lack of robustness to transient db issues.  I was more wondering what the state of the workflow would be if we landed in this chunk of code.

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
scottfrazer: Line 530

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
Horneth: Yeah it might get in a weird state with a Preempted call but no retry. The `CheckForWorkflowComplete` on the line after this one is there to ensure we don't get stuck in this state but calls depending on the preempted one would never become runnable.

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
Horneth: Oh yeah this one is a `removePersisting(callKey, ExecutionStatus.NotStarted)`
It's the mechanism used to keep track of what is currently being persisted and avoid out of order status persistence. I think you're right that in those cases it doesn't really matter as nothing really cares if something is being persisted to `Preempted` or `NotStarted`, I just added it for consistency.

scottfrazer:
:+1: I have nothing to add beyond what has already been said!

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
Horneth: I get the point of not having several methods doing the same thing, but
```
 f onFailure {
  case e => ...
}
```
looks more intuitive to me than
```
f.failed foreach {
 case e => ...
}
```

geoffjentry: On the foreach I think the point is to avoid a callback (which is a good thing IMO), but you could still use onComplete as well. Either way you probably have a few years before it matters

mcovarr: how about `recover`

Horneth: Yep that would be better

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -376,35 +375,42 @@ case class WorkflowActor(workflow: WorkflowDescriptor, backend: Backend)
...
-    completionWork onComplete {
-      case Failure(e) =>
-        logger.error(s"Preemption work failed for call ${callKey.tag}! " + e.getMessage, e)
-        currentSender ! CallActor.Ack(message)
-        self ! CheckForWorkflowComplete
-      case Success(_) =>
-        self ! PersistStatus(callKey, retryStatus, None, returnCode, None, None, currentSender, message)
+    insertCopy map { _ => PersistenceSucceeded(keyClone, ExecutionStatus.NotStarted) } recover {
+      case e =>
+        logger.error(s"Failed to clone ${callKey.tag} for retry attempt " + e.getMessage, e)
```
mcovarr: Is there a reason to do the `e.getMessage`?  The logger should dump that anyway with the exception as a parameter.

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -30,7 +30,7 @@ import scala.annotation.tailrec
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
 import scala.concurrent.duration._
-import scala.language.postfixOps
+import scala.language.{implicitConversions, postfixOps}
```
mcovarr: what's implicitly converting?

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
mcovarr: Actually the workflow won't transition states if there are pending persists, so this is necessary.

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
Horneth: Ah nothing, I added an `implicit def` which would have unleashed hell on me but removed it in the end - the import stayed

src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
Horneth: Right we have enough duplicated logs already

mcovarr:
:+1:

PR #495, opened 2016-03-02
Cromwell-backend as a sub-project of Cromwell

francares:
1. Added Cromwell-backend library as sub-project of Cromwell.
2. Added base interfaces from Cromwell-backend.
3. Tried to use SBT Git plugin for versioning but I couldn't get it work with sbt multi-projects.

@geoffjentry and @scottfrazer could you please review this PR?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -3,11 +3,11 @@ import sbtrelease.ReleasePlugin._
 
 name := "cromwell"
 
-version := "0.19"
-
 organization := "org.broadinstitute"
 
 scalaVersion := "2.11.7"
+
+version := "0.19"
 
 val lenthallV = "0.16"
 
@@ -82,8 +82,6 @@ libraryDependencies ++= Seq(
 )
 
 releaseSettings
-
-shellPrompt := { state => "%s| %s> ".format(GitCommand.prompt.apply(state), version.value)}
 
 assemblyJarName in assembly := "cromwell-" + version.value + ".jar"
 
@@ -165,10 +163,12 @@ lazy val CromwellIntegrationTest = config("integration") extend Test
 
 lazy val CromwellNoIntegrationTest = config("nointegration") extend Test
 
+lazy val CromwellBackend = Project(id = "CromwellBackend", base = file("cromwell-backend"))
+
 // NOTE: The following block may cause problems with IntelliJ IDEA
 // by creating multiple test configurations.
 // May need to comment out when importing the project.
-lazy val root = sbt.project.in(file("."))
+lazy val root = Project(id = "Cromwell", base = file(".")).aggregate(CromwellBackend).dependsOn(CromwellBackend)
   .configs(AllTests).settings(inConfig(AllTests)(Defaults.testTasks): _*)
   .configs(NoTests).settings(inConfig(NoTests)(Defaults.testTasks): _*)
   .configs(DockerTest).settings(inConfig(DockerTest)(Defaults.testTasks): _*)
--- a/cromwell-backend/CHANGELOG.MD
+++ b/cromwell-backend/CHANGELOG.MD
@@ -1,0 +1,5 @@
+# Cromwell-Backend Change Log
+
+1.0:
+=====
+
--- a/cromwell-backend/build.sbt
+++ b/cromwell-backend/build.sbt
@@ -1,0 +1,27 @@
+name := "Cromwell Backend"
+
+scalaVersion := "2.11.7"
+
+organization := "org.broadinstitute"
+
+version := "0.1"
+
+// Shorten the git commit hash
+git.gitHeadCommit := git.gitHeadCommit.value map { _.take(7) }
+
+// Travis will deploy tagged releases, add -SNAPSHOT for all local builds
+git.gitUncommittedChanges := true
+
+versionWithGit
+
+assemblyJarName in assembly := "cromwell-backend-" + version.value + ".jar"
+
+scalacOptions := Seq("-unchecked", "-deprecation", "-encoding", "utf8")
+
+// The reason why -Xmax-classfile-name is set is because this will fail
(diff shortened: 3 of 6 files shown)


build.sbt
```diff
@@ -83,8 +83,6 @@ libraryDependencies ++= Seq(
 
 releaseSettings
 
-shellPrompt := { state => "%s| %s> ".format(GitCommand.prompt.apply(state), version.value)}
-
```
scottfrazer: Why remove this?

cromwell-backend/CHANGELOG.MD
```diff
@@ -0,0 +1,5 @@
+# Cromwell-Backend Change Log
+
+1.0:
+=====
+
```
scottfrazer: Are we planning on releasing cromwell-backend as a JAR by itself?  If not perhaps we should just roll this in to the top level changelog?  I must admit I'm not entirely sure what the deploy plan is for this subproject

geoffjentry:
I'd rather see this PR purely being setting up the sub-project, not carrying over a bunch of code which will eventually live there unless everyone is 100% sure that code is exactly correct

cromwell-backend/CHANGELOG.MD
geoffjentry: Yes, it is a separate entity which the full cromwell jar is but one consumer of it

scottfrazer:
Yeah I'd rather see only code committed that's actually used... can we add these classes in only when they're actually used so we're sure that everything we're adding is hooked up and tested?

geoffjentry:
@scottfrazer thanks, that summed up my point better than my point :)

build.sbt
francares: I removed this due to the new version of the plugin. 
I couldn't resolve the part of base version but I read that the plugin takes the base version as project.version attribute.

francares:
@scottfrazer @geoffjentry They are going to be used in the next 3 tasks based on priority (Excel sheet), so I think It does not hurt anyone to have them as a baseline right know (they just define interfaces and not concrete implementations) since we took the work of defining them in the past.

geoffjentry:
I'd prefer every step be just what is required for that step as we don't know exactly how things will look when subsequent steps are completed

francares:
As you wish. I removed traits.

francares:
@geoffjentry @scottfrazer Please, let me know if it can be merged. Thanks!

scottfrazer:
@francares sorry for the slow responses this week.  We've got people out this week so we're falling a bit behind on code reviews (also this week was a release week, which has taken up a good amount of my time).  I will definitely review again by EOD tomorrow, but I'm not avoiding this!  I need to review @kshakir's PR first because I'm way overdue on that one, but this one is next.

scottfrazer:
@francares So I must admit that I am not the most familiar with SBT build files.  This all seems fairly boilerplate and the parts I do know well look right to me!

My understanding of the build process now is that we still run `sbt run ...` from the top level directory and it works as expected.  On this branch, `sbt assembly` seems to create only a Cromwell JAR file.  Is this supposed to build multiple JARs or are we still making a fat-JAR with the `cromwell-backend` code part of the JAR?

I'm excited to start getting some real code in here!

:+1:

cromwell-backend/build.sbt
```diff
@@ -0,0 +1,43 @@
...
+
+assemblyJarName in assembly := "cromwell-backend-" + version.value + ".jar"
+
+scalacOptions := Seq("-unchecked", "-deprecation", "-encoding", "utf8")
+
+resolvers ++= Seq("Typesafe Repository" at "http://repo.typesafe.com/typesafe/releases/",
+                  "Broad Artifactory Releases" at "https://artifactory.broadinstitute.org/artifactory/libs-release/",
+                  "Broad Artifactory Snapshots" at "https://artifactory.broadinstitute.org/artifactory/libs-snapshot/")
+
+libraryDependencies ++= {
```
geoffjentry: there's no code at the moment so currently no dependencies

geoffjentry:
I only quickly looked at it but it seems to build the correct jars

scottfrazer:
I was looking in the wrong place.  Right now it creates:

* `cromwell-backend/target/scala-2.11/cromwell-backend-0.1.jar`
* `target/scala-2.11/cromwell-0.19.jar`

And everything that's in `cromwell-backend/target/scala-2.11/cromwell-backend-0.1.jar` is included in `target/scala-2.11/cromwell-0.19.jar`

francares:
@scottfrazer Yes, that's right. If you take a look to the build logs you will see what your are saying.

Horneth:
:+1:  Looks good !

francares:
@geoffjentry I removed lib dependencies in Cromwell-backend.

scottfrazer:
Does anybody know what happens when versions of dependencies differ between `cromwell` and `cromwell-backend`?  I can find out empirically, though I wonder if there are any consequences of this we should be aware of...  and if that is a bad thing, is there any way we can define dependencies so they are in lockstep with Cromwell?

geoffjentry:
IIRC you generally want to inherit your deps (and everything else) from the main build.sbt and only vary the things which are actually varying. 

Which raises the point that we should be doing that instead of defining all new plugins, deps, etc in here

geoffjentry:
This is the sort of thing I'm talking about: http://stackoverflow.com/a/23189247

scottfrazer:
Though I guess if they really are pluggable backends, then somebody can create their own SBT project and include whatever they want as long as it exposes a class that adheres to the Actor interface.  So we really can't stop differing versions of Akka or any other shared package, right?

geoffjentry:
There's a difference between code officially supported by cromwell and stuff maintained separately. If someone else wants to wreak havoc with their own stuff that's their choice :)

francares:
After discussion on gitter I merging it.

PR #507, opened 2016-03-04
Add poll loggin to JES

Horneth:


src/main/scala/cromwell/engine/backend/jes/JesBackendCall.scala
```diff
@@ -90,13 +90,19 @@ class JesBackendCall(val backend: JesBackend,
 
   override def execute(implicit ec: ExecutionContext) = backend.execute(this)
 
+  def pollingLog(runId: String, status: RunStatus) = logger.info(s"Polled JES Job $runId. Job status : $status")
+
   override def poll(previous: ExecutionHandle)(implicit ec: ExecutionContext) = Future {
     previous match {
       case handle: JesPendingExecutionHandle =>
         val status = Try(handle.run.checkStatus(this, handle.previousStatus))
```
mcovarr: shouldn't the logging be right above this line where we actually do the status check?

Horneth: Hmm yes but then we can't log the status - although it's probably logged somewhere else before anyway yes

mcovarr: yeah that should already be happening in `Run`

```
      logger.info(s"Status change from $prevStateName to $currentStatus")
```

cjllanwarne:
Is this "always on"?

I would have thought we would want a way to stop this log message happening 10,000 times per JES job?

Horneth:
It doesn't happen that much actually, polling slowly backs off to 1 / 10min for long jobs.

cjllanwarne:
@Horneth it still feels like debug information to me - especially if you have ~100 workflows happening simultaneously? 

Can we make this "debug" and allow the logback.xml to make JesBackendCall log at "debug" level?

But, happy to be overruled by the next reviewer. The wheel chooses @scottfrazer.

cjllanwarne:
:+1: pending a call one way or the other re "always on"

mcovarr:
Agree with @cjllanwarne it would be preferable to log at debug, can this be overridden on the command line to specify debug-level logging rather than requiring changes to logback.xml?

:+1: delta this issue and moving the line where the logging is happening

src/main/scala/cromwell/engine/backend/jes/JesBackendCall.scala
```diff
@@ -93,10 +93,13 @@ class JesBackendCall(val backend: JesBackend,
...
       case handle: JesPendingExecutionHandle =>
+        logger.debug(s"Polling JES Job ${handle.run.runId}.")
         val status = Try(handle.run.checkStatus(this, handle.previousStatus))
         status match {
-          case Success(s: TerminalRunStatus) => backend.executionResult(s, handle)
-          case Success(s) => handle.copy(previousStatus = Option(s)).future // Copy the current handle with updated previous status.
+          case Success(s: TerminalRunStatus) =>
+            backend.executionResult(s, handle)
+          case Success(s) =>
+            handle.copy(previousStatus = Option(s)).future // Copy the current handle with updated previous status.
```
cjllanwarne: These can now be re-inlined

mcovarr:
hmm looks like the approver thing wants more thumbs?

:+1:

kshakir:
Approved :+1:

src/main/scala/cromwell/engine/backend/jes/JesBackendCall.scala
```diff
@@ -90,6 +88,7 @@ class JesBackendCall(val backend: JesBackend,
   override def poll(previous: ExecutionHandle)(implicit ec: ExecutionContext) = Future {
     previous match {
       case handle: JesPendingExecutionHandle =>
+        logger.debug(s"Polling JES Job ${handle.run.runId}.")
```
mcovarr: does this / can this include the workflow ID and call name / index in the output?

scottfrazer:
:+1:

PR #509, opened 2016-03-04
Move methods to JobDescriptor to get it closer to replacing BackendCall.

mcovarr:


mcovarr:
@Horneth and @scottfrazer reprise their roles as reviewers

cjllanwarne:
The *code* looks correct to me. I can't see where it gets used elsewhere but I guess that's already covered in upcoming tickets? :+1:

mcovarr:
@cjllanwarne the goal here is to move all backend-specific logic off the `BackendCall` and into the `Backend`.  This is an evolutionary step that maintains the same `BackendCall` interface but delegates all meaningful work directly or indirectly to the backend.

When this process is complete the `BackendCall` will become useless and methods can become parameterized by `JobDescriptor` instead.  The combination of `JobDescriptor` and `Backend` should be able to accomplish any call-level work in the system.

scottfrazer:
:+1:

PR #510, opened 2016-03-04
Cromwell piece of hashing, closes #490

### LINKED ISSUE #490: Move hash to wdl4s + backend

mcovarr:


src/main/scala/cromwell/engine/WorkflowDescriptor.scala
```diff
@@ -171,6 +171,10 @@ case class WorkflowDescriptor(id: WorkflowId,
   private def logReadDisabled() = workflowLogger.warn(readDisabled)
 
   override def toString = s"WorkflowDescriptor(${id.id.toString})"
+
+  def getHash(wdlValue: WdlValue): Option[SymbolHash] = {
```
geoffjentry: shouldn't this just be `hash`?

mcovarr: Yes.  This is how I found it, but I'll change it.

src/main/scala/cromwell/engine/SymbolStoreEntry.scala
```diff
@@ -1,7 +1,7 @@
 package cromwell.engine
 
 import wdl4s.types.WdlType
-import wdl4s.values.WdlValue
+import wdl4s.values.{SymbolHash, WdlValue}
```
cjllanwarne: I don't see why this would be necessary since nothing else changed?

mcovarr: `SymbolHash` now lives in wdl4s, it used to live in the engine package object.

src/main/scala/cromwell/util/FileUtil.scala
```diff
@@ -4,8 +4,8 @@ import java.io.{File, FileInputStream, Writer}
 import java.nio.file.Path
 
 import better.files._
-import cromwell.engine.Hashable
 import org.apache.commons.codec.digest.DigestUtils
+import wdl4s.values.Hashable
```
cjllanwarne: same here?

cjllanwarne:
`sbt.ResolveException: unresolved dependency: org.broadinstitute#wdl4s_2.11;0.4: not found` ?

cjllanwarne:
:+1:

mcovarr:
re the build breakage, I need to have the wdl4s snapshot published by the other PR's merge to develop and then update the build.sbt to reference that.

build.sbt
```diff
@@ -11,7 +11,7 @@ version := "0.19"
 
 val lenthallV = "0.16"
 
-val wdl4sV = "0.3"
+val wdl4sV = "0.4"
```
mcovarr: actually I think this should be 0.3 at a snapshot hash

scottfrazer: 0.4 + snapshot?  We already released 0.3 I believe.

scottfrazer:
:+1:

build.sbt
mcovarr: uh yeah, that makes a lot more sense :frowning:

PR #511, opened 2016-03-07
New .pullapprove.yml file

cjllanwarne:


cjllanwarne:
@geoffjentry @mcovarr please

.pullapprove.yml
```diff
@@ -0,0 +1,20 @@
+approve_by_comment: true
+approve_regex: ':\+1:'
+reject_regex: '^No'
+reject_value: -1
```
mcovarr: If I'm reading the `reject` settings correctly here and [here](http://docs.pullapprove.com/yaml/reject_value/), this seems kind of weird.  If I write "Nobody has ever PR'd more brilliantly than this" (you never know, that might happen), this would get a -1 score even though I was trying to say it was awesome.  At which point we'd need *three* thumbs up for it to become approvable?

cjllanwarne: Hmm, I don't even think we want negative reviews. I'll just remove this

geoffjentry:
:+1:

mcovarr:
:+1:

mcovarr:
:+1:

PR #514, opened 2016-03-07
Fix spelling of Gaurav's ID

mcovarr:
This pullapprover thing is not at all pleased when user IDs are misspelled


geoffjentry:
:+1:  :+1:

Issue #489, opened 2016-03-01
Validation Endpoint must use ValidateActor with the burden of converting to a StatusCode on itself


mcovarr:
depends on #488

PR #522, opened 2016-03-08
Bringing validation code to ValidateActor.  Closes #488

### LINKED ISSUE #488: Add all required validation to ValidationActor

gauravs90:
Solves #488. 
Initial work for Making a centralized validation. Currently, it brings the methods for inputs/options/runtime attr. validation, and uses it for validation via endpoints (like before). Next step will be to make Workflow descriptor a  little cleaner, by removing (most) of the validation from there and only instantiate the WorkflowDescriptor once the WDL passes validation (in the WFManagerActor).

### DIFF
--- a/engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
@@ -2,10 +2,12 @@ package cromwell.engine.workflow
 
 import akka.actor.{Actor, ActorRef, Props}
 import com.typesafe.scalalogging.LazyLogging
-import cromwell.engine.backend.CromwellBackend
+import cromwell.engine.backend.{Backend, CromwellBackend}
 import cromwell.engine.backend.runtimeattributes.CromwellRuntimeAttributes
+import cromwell.engine.workflow.ValidateActor.{ValidationResult, ValidationFailure, ValidationSuccess}
 import cromwell.util.TryUtil
-import wdl4s._
+import spray.json
+import spray.json.{JsObject, _}
 import cromwell.webservice.APIResponse
 import cromwell.webservice.PerRequest.RequestComplete
 import spray.http.StatusCodes
@@ -21,14 +23,23 @@ import scala.util.{Try, Failure, Success}
 object ValidateActor {
   private val tag = "ValidateActor"
 
-  sealed trait ValidateActorMessage
-  case object ValidateWorkflow extends ValidateActorMessage
-
   def props(wdlSource: WdlSource, wdlJson: Option[WdlJson], workflowOptions: Option[WdlJson]): Props = {
     Props(new ValidateActor(wdlSource, wdlJson, workflowOptions))
   }
+
+  sealed trait ValidateActorMessage
+  sealed trait ValidationResult extends ValidateActorMessage
+  // This ADT contains all the resolved / validated components of the workflow (minus Declarations).
+  // It should be used in the future to indicate if the workflow succeeded validation. The callee should reuse these materialized information.
+  case class ValidationSuccess(namespaceWithWorkflow: NamespaceWithWorkflow,
+                               coercedInputs: Option[WorkflowCoercedInputs],
+                               workflowOptions: Option[WorkflowOptions],
+                               runtimeAttributes: Seq[Set[String]]) extends ValidationResult
+  case class ValidationFailure(reason: Throwable) extends ValidationResult
+  case object ValidateWorkflow extends ValidateActorMessage
 }
 
+// TODO: Declarations cannot be validated here currently because of it's dependency on EngineFunctions, and in turn on IOManager and WfContext
 class ValidateActor(wdlSource: WdlSource, workflowInputs: Option[WdlJson], workflowOptions: Option[String])
   extends Actor with LazyLogging {
 
@@ -38,44 +49,96 @@ class ValidateActor(wdlSource: WdlSource, workflowInputs: Option[WdlJson], workf
   override def receive = {
     case ValidateWorkflow =>
       validateWorkflow(sender())
-      // NOTE: self shuts down when the parent PerRequest shuts down
-  }
-
-  private def validateInputs(namespaceWithWorkflow: NamespaceWithWorkflow, maybeWorkflowInputs: Option[WdlJson]): Future[Unit] = {
-    maybeWorkflowInputs match {
-      case Some(wi) =>
-        for {
-          inputs <- Future(wi.parseJson).map(_.asJsObject.fields)
-          coercedInputs <- Future.fromTry(namespaceWithWorkflow.coerceRawInputs(inputs))
-        } yield ()
-      case None => Future.successful(())
-    }
+    // NOTE: self shuts down when the parent PerRequest shuts down
   }
 
   private def validateWorkflow(sentBy: ActorRef): Unit = {
     logger.info(s"$tag for $sentBy")
-    val futureValidation: Future[Unit] = for {
-      namespaceWithWorkflow <- Future(NamespaceWithWorkflow.load(wdlSource))
-      validatedInputs <- validateInputs(namespaceWithWorkflow, workflowInputs)
-      validatedRuntimeOptions <- Future.fromTry(TryUtil.sequence(namespaceWithWorkflow.workflow.calls map {
-        call => CromwellRuntimeAttributes.validateKeys(call.task.runtimeAttributes.attrs.keySet, CromwellBackend.backend.backendType)
-      }))
-    } yield ()
-
-    // Now validate that this Future completed successfully:
-    futureValidation onComplete {
-      case Success(_) =>
+    // TODO: IMO, the responsibility for returning a status code should be a part of the client. This class should only return a validation result,
+    // and the callee should convert that to a status code or anything that it wants
+    validateAll() map {
+      case _: ValidationSuccess =>
         logger.info(s"$tag success $sentBy")
         sentBy ! RequestComplete(
           StatusCodes.OK,
           APIResponse.success("Validation succeeded."))
 
-      case Failure(ex) =>
-        val messageOrBlank = Option(ex.getMessage).mkString
+      case ValidationFailure(reason) =>
+        val messageOrBlank = Option(reason.getMessage).mkString
         logger.info(s"$tag error $sentBy: $messageOrBlank")
         sentBy ! RequestComplete(
           StatusCodes.BadRequest,
-          APIResponse.fail(ex))
+          APIResponse.fail(reason))
+    }
+  }
+
+  /**
+    * This function collectively validates:
+    * 1.) Loading of the namespace,
+    * 2.) Raw inputs and it's coercion
+    * 3.) Workflow Options
+    * 4.) RuntimeAttributes w.r.t. a backend
+    * @return
+    */
+  private def validateAll(): Future[ValidationResult] = {
+    (for {
+      namespaceWithWorkflow <- Future(NamespaceWithWorkflow.load(wdlSource))
+      validatedInputs <- Future(validateInputs(namespaceWithWorkflow))
+      validatedWorkflowOptions <- Future(validateWorkflowOptions(CromwellBackend.backend()))
+      validatedRuntimeAttrs <- Future(validateRuntimeAttributes(namespaceWithWorkflow))
+    } yield ValidationSuccess(namespaceWithWorkflow, validatedInputs, validatedWorkflowOptions, validatedRuntimeAttrs)) recover {
+      case reason => ValidationFailure(reason)
+    }
+  }
+
+  // TODO: With PBE, this should be defined in the backend.
+  private def validateRuntimeAttributes(namespaceWithWorkflow: NamespaceWithWorkflow): Seq[Set[String]] = {
+    TryUtil.sequence(namespaceWithWorkflow.workflow.calls map {
+      call => CromwellRuntimeAttributes.validateKeys(call.task.runtimeAttributes.attrs.keySet, CromwellBackend.backend().backendType)
+    }) match {
+      case Success(validatedRuntimeAttrs) => validatedRuntimeAttrs
+      case Failure(reason) => throw new IllegalArgumentException("Failed to validate runtime attributes.", reason)
+    }
+  }
+
+  private def validateInputs(namespaceWithWorkflow: NamespaceWithWorkflow): Option[WorkflowCoercedInputs] = {
+    def rawInputs(json: WdlJson): Map[String, JsValue] = {
+      Try(json.parseJson) match {
+        case Success(JsObject(inputs)) => inputs
+        case Failure(reason: Throwable) => throw new Throwable(reason.getMessage)
+        case _ => throw new IllegalArgumentException(s"Failed to parse the workflow inputs JSON: $json")
+      }
+    }
+
+    workflowInputs match {
+      case Some(json) =>
+        namespaceWithWorkflow.coerceRawInputs(rawInputs(json)) match {
+          case Success(coercedInputs) => Option(coercedInputs)
+          case Failure(reason: ThrowableWithErrors) => throw new ValidationException(reason.message, reason.errors)
+          case Failure(reason) => throw new IllegalArgumentException("Failed to coerce inputs", reason)
+        }
+      case _ => None
+    }
+  }
+
+  // TODO: With PBE, this should be defined in the backend.
+  private def validateWorkflowOptions(backend: Backend): Option[WorkflowOptions] = {
+    def validateBackendOptions(options: WorkflowOptions, backend: Backend): WorkflowOptions = {
+      try {
+        backend.assertWorkflowOptions(options)
+        options
+      } catch {
+        case e: Exception => throw new IllegalArgumentException(s"Workflow has invalid options for backend ${backend.backendType}: ${e.getMessage}", e)
+      }
+    }
+
+    workflowOptions match {
+      case Some(optionsJson) =>
+        WorkflowOptions.fromJsonString(optionsJson) match {
+          case Success(o) => Option(validateBackendOptions(o, backend))
+          case Failure(e) => throw new IllegalArgumentException(s"Failed to validate the workflow options JSON: ${e.getMessage}", e)
+        }
+      case _ => None
     }
   }
 }


gauravs90:
@geoffjentry and @cjllanwarne are the lucky ones here

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
```diff
@@ -21,14 +23,22 @@ import scala.util.{Try, Failure, Success}
...
   }
+
+  sealed trait ValidateActorMessage
+  //This is the reply that this actor sends in case of a succsessful validation. It contains all the resolved / validated components of
+  //the workflow (minus Declarations). The callee should reuse these materilized information.
+  case class ValidationSuccess(namespaceWithWorkflow: NamespaceWithWorkflow,
+                               coercedInputs: Option[WorkflowCoercedInputs],
+                               workflowOptions: Option[WorkflowOptions],
+                               runtimeAttributes: Seq[Set[String]]) extends ValidateActorMessage
+  case class ValidationFailure(reason: Throwable) extends ValidateActorMessage
```
mcovarr: `ValidationSuccess` and `ValidationFailure` can both be messages, but they should also both be subtypes of `ValidationResult`.

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
```diff
@@ -38,29 +48,20 @@ class ValidateActor(wdlSource: WdlSource, workflowInputs: Option[WdlJson], workf
...
-        } yield ()
-      case None => Future.successful(())
-    }
+    // NOTE: self shuts down when the parent PerRequest shuts down
   }
 
   private def validateWorkflow(sentBy: ActorRef): Unit = {
     logger.info(s"$tag for $sentBy")
-    val futureValidation: Future[Unit] = for {
+    val futureValidation: Future[ValidationSuccess] = for {
```
mcovarr: This should be `Future[ValidationResult]`, with the result of the forcomp `recover`ed to produce a `ValidationFailure`:

```scala
val futureValidation: Future[ValidationResult] = (for {
 ... 
} yield ValidationSuccess(...)) recover { case e => ValidationFailure(e) } 
```

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
```diff
@@ -38,29 +48,20 @@ class ValidateActor(wdlSource: WdlSource, workflowInputs: Option[WdlJson], workf
...
-        call => CromwellRuntimeAttributes.validateKeys(call.task.runtimeAttributes.attrs.keySet, CromwellBackend.backend.backendType)
-      }))
-    } yield ()
+      validatedInputs: Option[WorkflowCoercedInputs] <- Future(validateInputs(namespaceWithWorkflow))
+      validatedWorkflowOptions: Option[WorkflowOptions] <- Future(validateWorkflowOptions(CromwellBackend.backend()))
+      validatedRuntimeAttrs: Seq[Set[String]] <- Future(validateRuntimeAttributes(namespaceWithWorkflow))
+    } yield ValidationSuccess(namespaceWithWorkflow, validatedInputs, validatedWorkflowOptions, validatedRuntimeAttrs)
+
+    //TODO: IMO, the responsibility for returning a status code should be a part of the client. This class should only return a validation result,
+    // and the callee should convert that to a status code or anything that it wants
```
mcovarr: yeah I would agree with that

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
```diff
@@ -78,4 +79,55 @@ class ValidateActor(wdlSource: WdlSource, workflowInputs: Option[WdlJson], workf
           APIResponse.fail(ex))
     }
   }
+
+  //TODO: With PBE, this should be defined in the backend.
```
mcovarr: any reason not to do that now, not in this PR but a separate one?

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
```diff
@@ -21,14 +23,22 @@ import scala.util.{Try, Failure, Success}
...
+  //the workflow (minus Declarations). The callee should reuse these materilized information.
+  case class ValidationSuccess(namespaceWithWorkflow: NamespaceWithWorkflow,
+                               coercedInputs: Option[WorkflowCoercedInputs],
+                               workflowOptions: Option[WorkflowOptions],
+                               runtimeAttributes: Seq[Set[String]]) extends ValidateActorMessage
+  case class ValidationFailure(reason: Throwable) extends ValidateActorMessage
+  case object ValidateWorkflow extends ValidateActorMessage
 }
 
+//TODO: Declarations cannot be validated here currently because of it's dependency on EngineFunctions, and in turn on IOManager and WfContext
```
mcovarr: validate this on the backend then?

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
mcovarr: Hmm maybe just a `ValidationResult` makes more sense if the `Future` is not going to fail.

I'm actually not sure this needs to be a `Future`, are we doing anything async?  Maybe a `Try` instead?

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
gauravs90: Yep..I'm planning to make that  a part of #489

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
gauravs90: I wanted to move it to the backend trait, but didn't to just have a single file change for this PR. If you think it's okay I don't mind moving it to the backend trait right now.

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
gauravs90: Will do.

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
gauravs90: Will do.

gauravs90:
@mcovarr Any reasons why the build is failing? I ran all the tests locally and it looked like it ran successfully.

mcovarr:
All of our builds are failing, something appears to be wrong with Travis.

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
gauravs90: Actually, I think it'd be much better later on with a WF level validator for backend, which would be a single method (or a separate class/actor responsible for the validation) (PS: refer the PR that I did on the PBE branch a few days ago)

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
gauravs90: About the Future, we are going to have Actors for the backends, right? Validation will then be an 'AND' of the results of the methods used here (e.g. validateInputs, namespace etc.) + methods from backends (e.g. for runtimeattrs, communicated via messages, which need to be async.). Looks to me a Future will be imminent (?)

mcovarr: It's definitely an ANDing, but I wasn't seeing any async so I was thinking it could be a `Try`.  `Future` is more...futureproof though.

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
mcovarr: yeah ok

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
```diff
@@ -38,29 +49,22 @@ class ValidateActor(wdlSource: WdlSource, workflowInputs: Option[WdlJson], workf
...
+      validatedRuntimeAttrs: Seq[Set[String]] <- Future(validateRuntimeAttributes(namespaceWithWorkflow))
+    } yield ValidationSuccess(namespaceWithWorkflow, validatedInputs, validatedWorkflowOptions, validatedRuntimeAttrs)) recover {
+      case reason: Throwable => ValidationFailure(reason)
+    }
+
+    //TODO: IMO, the responsibility for returning a status code should be a part of the client. This class should only return a validation result,
+    // and the callee should convert that to a status code or anything that it wants
 
     // Now validate that this Future completed successfully:
     futureValidation onComplete {
```
mcovarr: Now the `Future` can't fail because of the recover, so I think this should be something like

```scala
    // Now validate that this Future completed successfully:
    futureValidation map {
      case _: ValidationSuccess =>
        logger.info(s"$tag success $sentBy")
        sentBy ! RequestComplete(
          StatusCodes.OK,
          APIResponse.success("Validation succeeded."))

      case ValidationFailure(ex) =>
        val messageOrBlank = Option(ex.getMessage).mkString
        logger.info(s"$tag error $sentBy: $messageOrBlank")
        sentBy ! RequestComplete(
          StatusCodes.BadRequest,
          APIResponse.fail(ex))
    }
```

gauravs90: Oops! Good catch! I didn't re ran the tests after that change. Thanks! :smile:

francares:
:+1:

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
```diff
@@ -21,14 +23,23 @@ import scala.util.{Try, Failure, Success}
 object ValidateActor {
   private val tag = "ValidateActor"
```
jainh: what is the significance of this constant? can you do this.class.getSimpleName() ?

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
```diff
@@ -38,44 +49,88 @@ class ValidateActor(wdlSource: WdlSource, workflowInputs: Option[WdlJson], workf
...
     logger.info(s"$tag for $sentBy")
-    val futureValidation: Future[Unit] = for {
+    val futureValidation: Future[ValidationResult] = (for {
       namespaceWithWorkflow <- Future(NamespaceWithWorkflow.load(wdlSource))
-      validatedInputs <- validateInputs(namespaceWithWorkflow, workflowInputs)
-      validatedRuntimeOptions <- Future.fromTry(TryUtil.sequence(namespaceWithWorkflow.workflow.calls map {
-        call => CromwellRuntimeAttributes.validateKeys(call.task.runtimeAttributes.attrs.keySet, CromwellBackend.backend.backendType)
-      }))
-    } yield ()
+      validatedInputs: Option[WorkflowCoercedInputs] <- Future(validateInputs(namespaceWithWorkflow))
```
jainh: Do we need to define type here or just name. It is obvious you are getting inputs, outputs ..

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
```diff
@@ -38,44 +49,88 @@ class ValidateActor(wdlSource: WdlSource, workflowInputs: Option[WdlJson], workf
...
+      case reason: Throwable => ValidationFailure(reason)
+    }
+
+    //TODO: IMO, the responsibility for returning a status code should be a part of the client. This class should only return a validation result,
+    // and the callee should convert that to a status code or anything that it wants
 
     // Now validate that this Future completed successfully:
-    futureValidation onComplete {
-      case Success(_) =>
+    futureValidation map {
```
jainh: Instead of map, i think it should be onComplete or collect for success and failure.

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
```diff
@@ -38,44 +49,88 @@ class ValidateActor(wdlSource: WdlSource, workflowInputs: Option[WdlJson], workf
...
+  }
+
+  private def validateInputs(namespaceWithWorkflow: NamespaceWithWorkflow): Option[WorkflowCoercedInputs] = {
+    workflowInputs match {
+      case Some(json) =>
+        def rawInputs(): Map[String, JsValue] = {
+          Try(json.parseJson) match {
+            case Success(JsObject(inputs)) => inputs
+            case Failure(reason: Throwable) => throw new Throwable(reason.getMessage)
+            case _ => throw new IllegalArgumentException(s"Failed to parse the workflow inputs JSON: $workflowInputs")
```
jainh: $workflowInputs  or $json ?

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
```diff
@@ -38,44 +49,88 @@ class ValidateActor(wdlSource: WdlSource, workflowInputs: Option[WdlJson], workf
...
+          }
+        }
+
+        namespaceWithWorkflow.coerceRawInputs(rawInputs()) match {
+          case Success(coercedInputs) => Option(coercedInputs)
+          case Failure(reason: ThrowableWithErrors) => throw new ValidationException(reason.message, reason.errors)
+          case Failure(reason) => throw new IllegalArgumentException("Failed to coerce inputs", reason)
```
jainh: Is this catching above  case Failure(reason: Throwable) => throw new Throwable(reason.getMessage)  ?

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
```diff
@@ -38,44 +49,88 @@ class ValidateActor(wdlSource: WdlSource, workflowInputs: Option[WdlJson], workf
...
+        }
+      case _ => None
+    }
+  }
+
+  //TODO: With PBE, this should be defined in the backend.
+  private def validateWorkflowOptions(backend: Backend): Option[WorkflowOptions] = {
+    workflowOptions match {
+      case Some(optionsJson) =>
+        def validateBackendOptions(options: WorkflowOptions, backend: Backend): WorkflowOptions = {
```
jainh: You can do Try[] of and match that in below instead of this try{} catch(){} ...

jainh:
After minor changes it can be merged.

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
gauravs90: I'm not sure what that does(Maybe it is being used for logging).  I've seen *'tag'*s being used around a few files..probably all used for logging.

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
gauravs90: I believe that will make the code harder to read. This Future is guaranteed to succeed (see `recover`), as such the onComplete will always yield a success, which then you'll have to pattern match against `Success(ValidationFailure)` (besides the compiler complaining about potential match errors because of a missing `case Failure(..)` , a line which will never be reached even if added.

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
gauravs90: Yes! Will do.

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
gauravs90: I believe the last case is in fact this same one.

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
mcovarr: Yes it is used for logging to add context to messages, usually workflow ID and perhaps call name / index.  But here we're operating at the workflow level so there is no call context, and this is before a workflow ID has been generated.  So the tag here is just the actor name which would be the same for all workflow submissions and therefore not so useful.

This could be enhanced to try to at least pull out the workflow name, but I don't think that should be part of these changes.

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
mcovarr: Agreed, this doesn't appear to be necessary.  The ascription on the `reason: Throwable` below also appears unnecessary.

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
mcovarr: Yeah the structure of the code with a never-failing `Future` is a bit awkward now, but I think this is a good intermediate step as it does work in the `ValidationResult` ADT.  As Guarav pointed out the responsibility for returning a status code should be on the caller; this code should just message back `ValidationSuccess` or `ValidationFailure`.  Once this code and its callers are refactored to account for that, I think this could become something like:

```scala
val futureValidation: Future[ValidationResult] = for {
...
} yield ValidationSuccess(...)

futureValidation onComplete {
  case Success(s) => sentBy ! s
  case Failure(x) => sentBy ! ValidationFailure(x)
}
```

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
mcovarr: yup

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
```diff
@@ -21,14 +23,23 @@ import scala.util.{Try, Failure, Success}
...
-  case object ValidateWorkflow extends ValidateActorMessage
-
   def props(wdlSource: WdlSource, wdlJson: Option[WdlJson], workflowOptions: Option[WdlJson]): Props = {
     Props(new ValidateActor(wdlSource, wdlJson, workflowOptions))
   }
+
+  sealed trait ValidateActorMessage
+  sealed trait ValidationResult extends ValidateActorMessage
+  //This is the reply that this actor sends in case of a succsessful validation. It contains all the resolved / validated components of
+  //the workflow (minus Declarations). The callee should reuse these materilized information.
```
mcovarr: some minor spelling problems in this comment

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
```diff
@@ -38,44 +49,88 @@ class ValidateActor(wdlSource: WdlSource, workflowInputs: Option[WdlJson], workf
...
         logger.info(s"$tag error $sentBy: $messageOrBlank")
         sentBy ! RequestComplete(
           StatusCodes.BadRequest,
-          APIResponse.fail(ex))
+          APIResponse.fail(reason))
+    }
+  }
+
+  //TODO: With PBE, this should be defined in the backend.
+  private def validateRuntimeAttributes(namespaceWithWorkflow: NamespaceWithWorkflow): Seq[Set[String]] = {
```
mcovarr: At a high level I think these changes should be an evolutionary step in moving validation code from `WorkflowDescriptor` to here rather than adapting the existing code or writing new validations.  There would have to be some minor adjustments to account for the fact that the existing `WorkflowDescriptor` validations assume the workflow ID exists, but the code that's in `WorkflowDescriptor` I think better represents how the validations should work.

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
geoffjentry: Could you add a space after the //'s? (yes, I realize I owe you a review of the stylesheet)

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
geoffjentry: Add me to the list of people raising eyebrows at type ascriptions in a for comprehension.

Also I feel that if one needs to wrap a for comp in parens that it should be pulled into a function call or something like that, as it is a sign that you're doing something more complicated on a single line than just the for comp

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
```diff
@@ -38,44 +49,88 @@ class ValidateActor(wdlSource: WdlSource, workflowInputs: Option[WdlJson], workf
...
-      }))
-    } yield ()
+      validatedInputs: Option[WorkflowCoercedInputs] <- Future(validateInputs(namespaceWithWorkflow))
+      validatedWorkflowOptions: Option[WorkflowOptions] <- Future(validateWorkflowOptions(CromwellBackend.backend()))
+      validatedRuntimeAttrs: Seq[Set[String]] <- Future(validateRuntimeAttributes(namespaceWithWorkflow))
+    } yield ValidationSuccess(namespaceWithWorkflow, validatedInputs, validatedWorkflowOptions, validatedRuntimeAttrs)) recover {
+      case reason: Throwable => ValidationFailure(reason)
+    }
+
+    //TODO: IMO, the responsibility for returning a status code should be a part of the client. This class should only return a validation result,
```
geoffjentry: Agree

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
geoffjentry: Agree w/ @mcovarr except for his championing `onComplete`

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
```diff
@@ -38,44 +49,88 @@ class ValidateActor(wdlSource: WdlSource, workflowInputs: Option[WdlJson], workf
...
         sentBy ! RequestComplete(
           StatusCodes.BadRequest,
-          APIResponse.fail(ex))
+          APIResponse.fail(reason))
+    }
+  }
+
+  //TODO: With PBE, this should be defined in the backend.
+  private def validateRuntimeAttributes(namespaceWithWorkflow: NamespaceWithWorkflow): Seq[Set[String]] = {
+    TryUtil.sequence(namespaceWithWorkflow.workflow.calls map {
```
geoffjentry: :cry:

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
```diff
@@ -38,44 +49,88 @@ class ValidateActor(wdlSource: WdlSource, workflowInputs: Option[WdlJson], workf
...
+    }) match {
+      case Success(validatedRuntimeAttrs) => validatedRuntimeAttrs
+      case Failure(reason) => throw new IllegalArgumentException("Failed to validate runtime attributes.", reason)
+    }
+  }
+
+  private def validateInputs(namespaceWithWorkflow: NamespaceWithWorkflow): Option[WorkflowCoercedInputs] = {
+    workflowInputs match {
+      case Some(json) =>
+        def rawInputs(): Map[String, JsValue] = {
```
geoffjentry: Create a function of `foo => () => Map[String, JsValue]` (where foo is whatever json is, I'm pretty limited at the moment) and have the match call that function with json.

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
geoffjentry: Similar to my comment above. Don't define this function directly in the match.

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
```diff
@@ -21,14 +23,23 @@ import scala.util.{Try, Failure, Success}
...
+
+  sealed trait ValidateActorMessage
+  sealed trait ValidationResult extends ValidateActorMessage
+  //This is the reply that this actor sends in case of a succsessful validation. It contains all the resolved / validated components of
+  //the workflow (minus Declarations). The callee should reuse these materilized information.
+  case class ValidationSuccess(namespaceWithWorkflow: NamespaceWithWorkflow,
+                               coercedInputs: Option[WorkflowCoercedInputs],
+                               workflowOptions: Option[WorkflowOptions],
+                               runtimeAttributes: Seq[Set[String]]) extends ValidationResult
+  case class ValidationFailure(reason: Throwable) extends ValidationResult
```
cjllanwarne: (genuine question :-) - If this is an async message, how does the receiver know which validation has failed?

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
```diff
@@ -38,44 +49,88 @@ class ValidateActor(wdlSource: WdlSource, workflowInputs: Option[WdlJson], workf
...
 
     // Now validate that this Future completed successfully:
-    futureValidation onComplete {
-      case Success(_) =>
+    futureValidation map {
+      case _: ValidationSuccess =>
         logger.info(s"$tag success $sentBy")
         sentBy ! RequestComplete(
           StatusCodes.OK,
           APIResponse.success("Validation succeeded."))
```
cjllanwarne: Yeah I totally agree - having these API responses in the validation actor is really jarring

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
mcovarr: @geoffjentry how would you prefer to do this?

geoffjentry: couldn't you do a map/recover?

mcovarr: I suppose, but yuk:

```scala

futureValidation map { s => 
  sentBy ! s
  s
} recover { case t => sentBy ! ValidationFailure(t) }
```

geoffjentry: That does look like teh sukx

cjllanwarne:
Hey @gauravs90 - for my personal bookkeeping (and so I don't forget to thumbs-up it), could you reassign this to me when you've responded to the comments?

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
gauravs90: I like the map / recover.. Not sure why we want to return a result from the map when we are sending a message.. But this looks good:  :grin: 

```scala
futureValidation map { sentBy ! _ } recover { case t => sentBy ! ValidationFailure(t) }
```

jainh: val futureValidation: Future[ValidationResult] = for {
...
} yield ValidationSuccess(...)

futureValidation.collect{ case Failure(x) => sentBy ! ValidationFailure(t)}

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
gauravs90: Yep, we'll have to bring in a WF ID somehow in this file, or anything that helps uniquely identify what workflow failed.

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
gauravs90: Done. Moved it to a nested method. (Since it's not being used anywhere else)

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
gauravs90: Done, Done.

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
gauravs90: Moved to a nested function being called form the match.

src/main/scala/cromwell/engine/workflow/ValidateActor.scala
geoffjentry: Heh, what I was specifically asking for was a simpler case statement.

geoffjentry:
@gauravs90 It might be good to rebase this prior to final review in case there are any conflicts which need to be worked out. It seems doubtful but you never know - this was originally branched several days ago

gauravs90:
@geoffjentry Done. Squashed and Rebased.

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
```diff
@@ -21,14 +23,23 @@ import scala.util.{Try, Failure, Success}
...
-  case object ValidateWorkflow extends ValidateActorMessage
-
   def props(wdlSource: WdlSource, wdlJson: Option[WdlJson], workflowOptions: Option[WdlJson]): Props = {
     Props(new ValidateActor(wdlSource, wdlJson, workflowOptions))
   }
+
+  sealed trait ValidateActorMessage
+  sealed trait ValidationResult extends ValidateActorMessage
+  // This is the reply that this actor sends in case of a succsesful validation. It contains all the resolved / validated components of
+  // the workflow (minus Declarations). The callee should reuse these materialized information.
```
cjllanwarne: Typo in successful. Also this is factually incorrect atm? The actor still sends an `APIResonse`

cjllanwarne:
One little comment to fix but after that (and another squash - sorry!) :+1:

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
```diff
@@ -38,44 +49,89 @@ class ValidateActor(wdlSource: WdlSource, workflowInputs: Option[WdlJson], workf
...
+        val messageOrBlank = Option(reason.getMessage).mkString
         logger.info(s"$tag error $sentBy: $messageOrBlank")
         sentBy ! RequestComplete(
           StatusCodes.BadRequest,
-          APIResponse.fail(ex))
+          APIResponse.fail(reason))
+    }
+  }
+
+  // Validates: 1.) Loading of the namespace, 2.) Raw inputs and it's coercion, 3.) WF Options, 4.) RuntimeAttributes w.r.t. a backend
```
geoffjentry: Pretty petty, I realize - but I think comment would read better as a multiline bullet pointed list (using `/* */` and not `//`)

geoffjentry:
lgtm

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
cjllanwarne: Why not go the full way and `/** */`

geoffjentry: Yeah, you're right. I lost track that that was actually for a function while I was typing it

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
gauravs90: Eeegh!! I thought I corrected that!

gauravs90: Modified the comment to future tense to indicate that :D

PR #524, opened 2016-03-08
Rework filesystems management.  Closes #497

### LINKED ISSUE #497: Move IoInterface to Backend

Horneth:
I know that this is bigger than what the issue says - I'm sorry about that. It just felt like now or never to rework a little bit our communication with GCS and remove some duplicate code in the process.

This PR removes the `IoInterface`/`IoManager` and uses `java.nio.FileSystem` implementations instead to provide file system interactions.
The (now removed) `IoInterface` was nothing but a wrapper interface around a `FileSystem`.
The (also removed) `IoManager` was a wrapper of `IoInterface`s that selected the most appropriate interface according to the path to be processed. This functionality is now provided by the `cromwell.engine.backend.io.PathString.toPath` method which takes a list of available `FileSystem`s and tries to create a `java.nio.Path`. If it succeeds (ie if there is a suitable `FileSystem` able to parse the raw `String`), the created `Path` can now be used in an abstract way (with `Files.*`) or `better.files` as long as the corresponding `FileSystemProvider` implements the required methods. 

The `WdlStandardLibraryImpl` now has a `List[FileSystem]` at its disposal and uses it to parse raw `String` paths with the `toPath` method. Most of the implementation can then be the same and rely entirely on the `FileSystem`s implementations. Engine functions can always be overridden by their respective backend implementations if a special treatment is needed.

The functionalities previously in `GoogleCloudStorage` have been merged into `GcsFileSystemProvider`.


src/main/scala/cromwell/engine/WdlStandardLibraryImpl.scala
```diff
@@ -126,14 +131,22 @@ trait WdlStandardLibraryImpl extends WdlStandardLibraryFunctions {
...
+    // parameters should yield the same return value.
+    val fullPath = s"$path/$prefix${content.md5Sum}$suffix".toAbsolutePath
+    if (!fullPath.exists) {
+      fullPath.write(content)
+    }
+    fullPath.toString
+  }
 
-  override def glob(path: String, pattern: String): Seq[String] = interface.glob(path, pattern)
+  override def glob(path: String, pattern: String): Seq[String]
```
mcovarr: overridden back to abstraction?  :smile:

src/main/scala/cromwell/engine/WorkflowDescriptor.scala
```diff
@@ -312,22 +276,28 @@ object WorkflowDescriptor {
...
                                       conf: Config): ErrorOr[WorkflowDescriptor] = {
-    val gcsInterface = GoogleCloudStorage.userAuthenticated(options) orElse GoogleCloudStorage.cromwellAuthenticated
-    val ioManager = backend match {
-      case _: JesBackend => gcsInterface match { // JesBackend only supports gcsInterface
-        case Success(x) => x
-        case Failure(t) => throw new Throwable("No GCS interface has been found. When running on JES Backend, Cromwell requires a google configuration to perform GCS operations.", t)
-      }
-      case _ => new IoManager(Seq(gcsInterface.toOption, Option(SharedFileSystemIoInterface.instance)).flatten)
-    }
+//    val gcsInterface = GoogleCloudStorage.userAuthenticated(options) orElse GoogleCloudStorage.cromwellAuthenticated
```
mcovarr: delete

src/main/scala/cromwell/engine/WorkflowDescriptor.scala
```diff
@@ -312,22 +276,28 @@ object WorkflowDescriptor {
...
+    val wfGcsFileSystem = gcsFileSystemProvider map { _.newFileSystem(wfContext.root) }
+    val fileSystems = backend match {
+      case _: JesBackend =>
+        val gcs = wfGcsFileSystem match {
+          case Success(x) => x
+          case Failure(t) => throw new Throwable("No GCS interface has been found. When running on JES Backend, Cromwell requires a google configuration to perform GCS operations.", t)
+        }
+        List(gcs, defaultFileSystem)
+      case _ => List(wfGcsFileSystem.toOption, Option(defaultFileSystem)).flatten
+    }
```
mcovarr: this filesystem determination should really move to the backends soon, if not now

src/main/scala/cromwell/engine/backend/io/filesystem/gcs/GcsFileSystemProvider.scala
```diff
@@ -84,59 +122,106 @@ class GcsFileSystemProvider private (gcs: Try[GoogleCloudStorage], executionCont
...
-      case (s: NioGcsPath, d: NioGcsPath) => googleCloudStorage.move(s, d)
+      case (s: NioGcsPath, d: NioGcsPath) =>
+        val storageObject = client.objects.get(s.bucket, d.objectName).execute
+        client.objects.rewrite(s.bucket, s.objectName, d.bucket, d.objectName, storageObject).execute
       case _ => throw new UnsupportedOperationException(s"Can only copy from GCS to GCS: $source or $target is not a GCS path")
     }
   }
-  override def checkAccess(path: Path, modes: AccessMode*): Unit = {checkExists(path)}
+
+  def crc32Hash(path: Path) = path match {
```
mcovarr: This is [CRC32C](http://stackoverflow.com/questions/26429360/crc32-vs-crc32c)

src/main/scala/cromwell/engine/backend/io/package.scala
```diff
@@ -0,0 +1,61 @@
...
+
+import java.nio.file._
+
+import cromwell.engine.backend.io.filesystem.gcs.{ContentTypeOption, GcsFileSystem, GcsFileSystemProvider, NioGcsPath}
+
+import scala.util.Try
+
+/**
+  * Created by tjeandet on 3/3/16.
+  */
```
mcovarr: delete

src/main/scala/cromwell/engine/backend/io/package.scala
```diff
@@ -0,0 +1,61 @@
...
+    val defaultFileSystem = FileSystems.getDefault
+    private val defaultFileSystemsList = List(defaultGCSFileSystem, defaultFileSystem)
+
+    implicit class UriString(val str: String) extends AnyVal {
+
+      def isGcsUrl: Boolean = str.startsWith("gs://")
+      def isUriWithProtocol: Boolean = "^[a-z]+://".r.findFirstIn(str).nonEmpty
+
+      def toPath(implicit fss: List[FileSystem] = defaultFileSystemsList): Path = {
+        fss.toStream map { fs => Try(fs.getPath(str)) } find { _.isSuccess } map { _.get } getOrElse {
```
mcovarr: `collectFirst`

src/main/scala/cromwell/engine/backend/io/package.scala
```diff
@@ -0,0 +1,61 @@
...
+          throw new IllegalArgumentException(s"Could not find suitable filesystem to parse $str")
+        }
+      }
+
+      def toDirectory(implicit fss: List[FileSystem] = defaultFileSystemsList): Path = {
+        fss.toStream collect {
+          case fs: GcsFileSystem => Try(fs.getPathAsDirectory(str))
+          case fs => Try(fs.getPath(str))
+        } find { _.isSuccess } map { _.get } getOrElse {
```
mcovarr: `collectFirst` here too, also below

src/main/scala/cromwell/engine/WdlStandardLibraryImpl.scala
Horneth: Oops - (Besides, "overridden back to abstraction" does not sound like a cool thing to experience :s)

mcovarr:
Could you please provide a high-level description of what's changed here?  It's a lot of code... :smile:

gauravs90:
I like this PR :smile:

scottfrazer:
Definitely very cool.  Do you think this is big enough of a change that we should run Tyburn tests on it before merging?

Horneth:
Definitely, I already did run it and it passed except the SingleSamplePipeline which failed at the very end for some obscure reason that I'm looking into

cjllanwarne:
wheel: @kshakir @geoffjentry

src/main/scala/cromwell/engine/WorkflowDescriptor.scala
```diff
@@ -44,11 +40,9 @@ case class WorkflowDescriptor(id: WorkflowId,
                               backend: Backend,
                               configCallCaching: Boolean,
                               lookupDockerHash: Boolean,
-                              gcsInterface: Try[GoogleCloudStorage],
-                              ioManager: IoInterface,
+                              implicit val fileSystems: List[FileSystem],
```
geoffjentry: what requires this to be implicit?

src/main/scala/cromwell/engine/backend/io/filesystem/gcs/GcsFileSystem.scala
```diff
@@ -29,16 +34,18 @@ object GcsFileSystem {
...
 class GcsFileSystem private (gcsFileSystemProvider: GcsFileSystemProvider, gcsRoot: String) extends FileSystem {
 
+
+
   import GcsFileSystem._
 
-  val root = gcsRoot match {
-    case GsUriRegex(chunks) => new NioGcsPath(chunks.split(Separator), true)(this)
-    case _ => throw new InvalidPathException(gcsRoot, s"Root of GCS file system must be an absolute GCS path: $gcsRoot")
+  val root: Option[Path] = gcsRoot match {
```
kshakir: Are there Specs you can point me to that explain _why_ we have to keep track of different filesystem objects with different roots? This variable is everywhere, but I can't figure out what functionality it actually provides.

Ideally I'd like to get rid of the `root` as a variable all together. If there's some way of having a singleton `GcsFileSystem` I'd love to see it.

src/main/scala/cromwell/engine/backend/io/filesystem/gcs/GcsFileSystem.scala
```diff
@@ -29,16 +34,18 @@ object GcsFileSystem {
   */
 class GcsFileSystem private (gcsFileSystemProvider: GcsFileSystemProvider, gcsRoot: String) extends FileSystem {
 
+
+
```
kshakir: These extra two lines don't add much to the blank one already there. :wink:

src/main/scala/cromwell/engine/backend/io/filesystem/gcs/GcsFileSystem.scala
```diff
@@ -6,21 +6,26 @@ import java.nio.file.attribute.UserPrincipalLookupService
...
-import scala.util.Try
+case class NotAGcsPathException(path: String) extends IllegalArgumentException(s"$path is not a valid GCS path.")
 
 object GcsFileSystem {
-  import PathString._
-  def apply(interface: Try[GoogleCloudStorage], root: String)(implicit executionContext: ExecutionContext) = {
-    if (root.isGcsUrl) new GcsFileSystem(GcsFileSystemProvider(interface), root)
-    else throw new IllegalArgumentException(s"$root is not an absolute GCS path")
+  def apply(provider: GcsFileSystemProvider, root: String) = {
+    new GcsFileSystem(provider , root)
```
kshakir: Example of an extra space. In general, this file could use a "Reformat Code ⌥⌘L"

src/main/scala/cromwell/engine/backend/io/package.scala
```diff
@@ -0,0 +1,58 @@
...
+        }
+      }
+    }
+
+    implicit class PathEnhanced(val path: Path) extends AnyVal {
+      import better.files._
+
+      def hash = path match {
+        case gcs: NioGcsPath => gcs.getFileSystem.provider().asInstanceOf[GcsFileSystemProvider].crc32cHash(gcs)
+        case _ => path.md5
```
kshakir: Comment only: Hashing the file path instead of the file contents is efficient, but disturbing. I'm not sure what our "spec" is right now for hashing if `my_file.bam.md5` isn't available as a metafile. I almost wonder if it's better to say `None`.

src/main/scala/cromwell/engine/backend/jes/JesBackend.scala
```diff
@@ -237,20 +240,19 @@ case class JesBackend(actorSystem: ActorSystem)
 
     generateAuthJson(dockerConf, getGcsAuthInformation(workflow)) foreach { content =>
 
-      val path = GcsPath(gcsAuthFilePath(workflow))
-      def upload(prev: Option[Unit]) = connection.storage.uploadJson(path, content)
+      val path = gcsAuthFilePath(workflow).toAbsolutePath(List(connection.gcsFileSystem))
+      def upload(prev: Option[Unit]) = {
+        path.writeAsJson(content)
+        ()
+      }
```
kshakir: Instead of explicitly returning a unit, I think I'd prefer to explicitly set the return type:

```scala
def upload(prev: Option[Unit]): Unit = path.writeAsJson(content)
```

src/main/scala/cromwell/engine/backend/io/package.scala
```diff
@@ -0,0 +1,58 @@
...
+    val defaultGCSFileSystem = GcsFileSystem.defaultGcsFileSystem
+    val defaultFileSystem = FileSystems.getDefault
+    private val defaultFileSystemsList = List(defaultGCSFileSystem, defaultFileSystem)
+
+    implicit class UriString(val str: String) extends AnyVal {
+
+      def isGcsUrl: Boolean = str.startsWith("gs://")
+      def isUriWithProtocol: Boolean = "^[a-z]+://".r.findFirstIn(str).nonEmpty
+
+      def toPath(implicit fss: List[FileSystem] = defaultFileSystemsList): Path = {
```
scottfrazer: I'm not fond of the name `toPath` but I also can't think of a better name...

I'm also curious why everything is done with implicits... I know I tend to be more against implicits than other people, so maybe this isn't a cause for concern... I can sometimes get lost figuring out where things are actually being created (most specifically, this `List[FileSystem]` parameter, the `UriString` implicit class is less of an issue for me)

src/main/scala/cromwell/engine/WorkflowDescriptor.scala
kshakir: `PathString.UriString` requires an implict `List[FileSystem]` that must show up automagically.

src/main/scala/cromwell/engine/backend/io/package.scala
```diff
@@ -0,0 +1,58 @@
+package cromwell.engine.backend
+
+import java.nio.file._
+
+import cromwell.engine.backend.io.filesystem.gcs.{ContentTypeOption, GcsFileSystem, GcsFileSystemProvider, NioGcsPath}
+
+import scala.util.{Success, Try}
+
+package object io {
+  object PathString {
```
kshakir: Can this be moved to `PathString.scala` instead of the package object, or is there some import magic happening here?

kshakir:
First pass review complete. Assign back when I should rescan.

src/main/scala/cromwell/engine/backend/io/filesystem/gcs/GcsFileSystem.scala
Horneth: I agree with you, this root thing is bothering me too. What it gives you is the ability to call `toAbsolutePath` on a relative GCS path (all it really does is prepend the root to it).
However I'm unsure how useful that is since I removed the ability to create relative GCS paths from the `GcsFileSystem.getPath` method because it gets too confusing (it could successfully create a GCS path from "my/path" which there is no real way to know if it's a GCS or a Unix relative path..).
I'm gonna try and get rid of the root thing.

src/main/scala/cromwell/engine/backend/io/package.scala
Horneth: Argh this is not going to help my case for implicits :s
This is actually an implicit method in better.files._ that reads the file and hash the content (which by the way will need to be change anyway since it loads everything in memory and blows up for big files...)

src/main/scala/cromwell/engine/backend/io/package.scala
Horneth: Yep

src/main/scala/cromwell/engine/backend/io/package.scala
kshakir: Aha. Yeah... pick your poison.

src/main/scala/cromwell/engine/backend/io/package.scala
Horneth: Yeah, I don't really know what else to call it as it effectively takes a String and builds a Path :/
The filesystems don't have to be `implicit`s, but I think this is a pretty good use case for i, because when using this functionality, you're likely to have a list of available file systems created at a high level, and then to want to use them whenever you want to build a `Path`. If they're implicit you don't have to pass them in every single call to `toPath`, and if you do want to pass them (or others) you still can.

However I know I'm definitely part of a minority here with implicits so I'm not going to argue for ages for it :p

geoffjentry: What's wrong with `toPath`, that's literally what it does.

The implicit parameter seems borderline to me - if it's just being used for aesthetic purposes I'd be inclined to say no to it

Horneth: I feel like `implicit`s are literally just used for aesthetic purposes in 99% of the cases though.
I mean when you `import scala.concurrent.ExecutionContext.Implicits.global`, you could also pass `scala.concurrent.ExecutionContext.global` to every method that needs an execution context. I don't really see why it's ok for execution contexts but not for filesystems.

src/main/scala/cromwell/engine/WdlStandardLibraryImpl.scala
```diff
@@ -126,14 +131,21 @@ trait WdlStandardLibraryImpl extends WdlStandardLibraryFunctions {
   override def write_map(params: Seq[Try[WdlValue]]): Try[WdlFile] = writeToTsv(params, classOf[WdlMap])
   override def write_object(params: Seq[Try[WdlValue]]): Try[WdlFile] = writeToTsv(params, classOf[WdlObject])
   override def write_objects(params: Seq[Try[WdlValue]]): Try[WdlFile] = writeToTsv(params, classOf[WdlArray])
-  override def writeTempFile(path: String,prefix: String,suffix: String,content: String): String = interface.writeTempFile(path, prefix, suffix, content)
+  override def writeTempFile(path: String,prefix: String,suffix: String,content: String): String = {
```
mcovarr: spacing woes

src/main/scala/cromwell/engine/backend/io/package.scala
scottfrazer: Well, I'm able to follow what's going on, and I guess ultimately that's what matters most.  If I were writing this, I'd probably try to see if it can be done cleanly without the implicit parameter... but I'm just one person who has a bias :)

Horneth: We talked about this yesterday with Jeff, I'm going to remove the implicit to see how it looks like.
It definitely can be done without `implicit`s (like I said it just felt like a decent use case to use them). I don't think it's a big deal having to pass the argument every time we use `toPath` or whatever methods needs the filesystems, it just feels unnecessary to me because they're always going to be the same for a Call instance :)

Horneth:
@kshakir @geoffjentry @mcovarr round 2 ready

engine/src/main/scala/cromwell/engine/WdlStandardLibraryImpl.scala
```diff
@@ -127,14 +132,21 @@ trait WdlStandardLibraryImpl extends WdlStandardLibraryFunctions {
...
   override def write_objects(params: Seq[Try[WdlValue]]): Try[WdlFile] = writeToTsv(params, classOf[WdlArray])
   override def write_tsv(params: Seq[Try[WdlValue]]): Try[WdlFile] = writeToTsv(params, classOf[WdlArray])
-  override def writeTempFile(path: String,prefix: String,suffix: String,content: String): String = interface.writeTempFile(path, prefix, suffix, content)
+  override def writeTempFile(path: String, prefix: String, suffix: String, content: String): String = {
+    // This may be called multiple times with the same inputs.  Calling this twice with the same
+    // parameters should yield the same return value.
+    val fullPath = path.toAbsolutePath(fileSystems).resolve(s"$prefix${content.md5Sum}$suffix")
+    if (!fullPath.exists) {
+      fullPath.write(content)
+    }
```
geoffjentry: \n 

@cjllanwarne I'm getting him for you

engine/src/main/scala/cromwell/engine/backend/io/PathFactory.scala
```diff
@@ -0,0 +1,26 @@
+package cromwell.engine.backend.io
+
+import java.nio.file.{Path, FileSystem}
+
+import cromwell.engine.backend.io.filesystem.gcs.GcsFileSystem
+
+import scala.util.{Success, Try}
+
+object PathFactory {
+
```
geoffjentry: - \n

engine/src/main/scala/cromwell/engine/backend/io/PathFactory.scala
```diff
@@ -0,0 +1,26 @@
...
+  def buildPathAsDirectory(rawString: String, fileSystems: List[FileSystem]): Path = {
+    fileSystems.toStream collect {
+      case fs: GcsFileSystem => Try(fs.getPathAsDirectory(rawString))
+      case fs => Try(fs.getPath(rawString))
+    } collectFirst { case Success(p) => p } getOrElse {
+      throw new IllegalArgumentException(s"Could not find suitable filesystem to parse $rawString as a Directory")
+    }
+  }
+
```
geoffjentry: - \n

engine/src/main/scala/cromwell/engine/WorkflowDescriptor.scala
```diff
@@ -146,14 +111,13 @@ case class WorkflowDescriptor(id: WorkflowId,
...
                              (implicit executionContext: ExecutionContext): Future[Unit] = {
-    Future(copyWdlFilesTo(destDirectory, gcsOutputsFilesystem, workflowMetadataResponse.outputs.toSeq.flatMap(_.values)))
+    Future(copyWdlFilesTo(destDirectory, workflowMetadataResponse.outputs.toSeq.flatMap(_.values)))
   }
 
   def copyCallLogs(workflowMetadataResponse: WorkflowMetadataResponse)
                   (implicit executionContext: ExecutionContext): Future[Unit] = {
     callLogsDir map copyCallLogsTo(workflowMetadataResponse) getOrElse Future.successful(())
   }
-
```
kshakir: Please put this line back.

engine/src/main/scala/cromwell/engine/backend/jes/JesBackend.scala
```diff
@@ -555,7 +563,10 @@ case class JesBackend(actorSystem: ActorSystem)
          |echo $$? > $rcPath
        """.stripMargin.trim
 
-    def attemptToUploadObject(priorAttempt: Option[Unit]) = authenticateAsUser(backendCall.workflowDescriptor) { _.uploadObject(backendCall.gcsExecPath, fileContent) }
+    def attemptToUploadObject(priorAttempt: Option[Unit]) = authenticateAsUser(backendCall.workflowDescriptor) { implicit gcsFs =>
+      backendCall.gcsExecPath.write(fileContent)
+      ()
```
kshakir: This is weird. Why not set your method return type explicitly instead of `return Unit`.

Horneth: Yes for some reason I thought that didn't work. I'll change it

engine/src/main/scala/cromwell/engine/backend/jes/JesBackend.scala
```diff
@@ -883,6 +894,18 @@ case class JesBackend(actorSystem: ActorSystem)
     lazy val jesStderrGcsPath = callGcsPath.resolve(jesLogStderrFilename(key)).toString
 
     val callContext = new CallContext(callGcsPath.toString, jesStdoutGcsPath, jesStderrGcsPath)
-    new JesCallEngineFunctions(workflowDescriptor.ioManager, callContext)
+    new JesCallEngineFunctions(workflowDescriptor.fileSystems, callContext)
+  }
+
+  override def fileSystems(options: WorkflowOptions, workflowRootPath: String): List[FileSystem] = {
+    // TODO Should GCS operations have their own execution context ?
+    implicit val executionContext = scala.concurrent.ExecutionContext.global
```
kshakir: I'm going to punt on this for a future ticket... but for discussion:

Yes, they should have either their own `ExecutionContext`, or it should be passed in. If I'm reading correctly, `GoogleCloudStorageWriteChannel` is using an `ExecutionContext` for async upload, only during `GcsFileSystemProvider.newOutputStream(Path)`. But afaik, this still synchronously returns an `OutputStream` that is then used synchronously by the caller. Meanwhile, none of the `FileSystem` operations externally behave asynchronously. Thus if GCS behind the scenes has it's own asynchonicity going on, it should, IMHO, have it's own, internal, thread pool. Also it cleans it up such that these _O(n)_ callers don't need to keep passing in a pool. If the GCS wants to use the "global" pool internally, then it should.

IF on the other hand, the `FileSystem` operations were returning `Future[_]` asynchronous results, then they should also be taking in an execution context as a (probably implicit) parameter.

engine/src/main/scala/cromwell/engine/backend/sge/SgeBackend.scala
```diff
@@ -219,6 +223,16 @@ case class SgeBackend(actorSystem: ActorSystem) extends Backend with SharedFileS
...
+  }
+
+  override def fileSystems(options: WorkflowOptions, workflowRootPath: String): List[FileSystem] = {
+    // TODO Should GCS operations have their own execution context ?
+    implicit val executionContext = scala.concurrent.ExecutionContext.global
+
+    val gcsStorage = StorageFactory.userAuthenticated(options) orElse StorageFactory.cromwellAuthenticated
+    val gcs = gcsStorage map GcsFileSystemProvider.apply map { _.getDefaultFileSystem } toOption
+
+    List(gcs, Option(defaultFileSystem)).flatten
```
kshakir: Also punting, but this "smells". FilesSystems feel separate than Backends.

Yes a backend may _need_ a filesystem, but these _O(n)_ backends shouldn't have to each deal with creating a set of `FileSystem` objects in a similar way.

kshakir:
In the interest of "Move IoInterface to Backend #497", :+1: 

Left a few items for future thought, but also super happy to see the GcsFileSystem roots gone! :tada:

engine/src/main/scala/cromwell/engine/backend/sge/SgeBackend.scala
Horneth: > FilesSystems feel separate than Backends

Yes, but this is a bigger change than the primary purpose of this ticket (which this PR is already over reaching anyway..)

geoffjentry: Agree w/ both of you. @Horneth can you write up an issue to take a look at this?

Horneth: https://github.com/broadinstitute/cromwell/issues/555

engine/src/main/scala/cromwell/engine/backend/jes/JesBackend.scala
geoffjentry: I like the notion of GCS having its own threadpool

@kshakir Can you write this up as a ticket to investigate?

Horneth: For now I removed the ExecutionContext argument from the apply() methods and explicitly gave the global execution context to the constructor, which is functionally equivalent but will make it easier to give GCS its own internal EC as @kshakir suggested.

engine/src/main/scala/cromwell/engine/backend/io/filesystem/gcs/GcsFileSystemProvider.scala
```diff
@@ -0,0 +1,249 @@
...
+import com.google.cloud.hadoop.util.{ApiErrorExtractor, AsyncWriteChannelOptions, ClientRequestHelper}
+
+import scala.collection.JavaConverters._
+import scala.concurrent.duration._
+import scala.concurrent.{ExecutionContext, ExecutionContextExecutorService}
+import scala.language.postfixOps
+import scala.util.{Failure, Success, Try}
+
+object GcsFileSystemProvider {
+  private val executionContext = scala.concurrent.ExecutionContext.global
```
kshakir: Definitely better than before. Thanks!

engine/src/main/scala/cromwell/engine/backend/io/PathFactory.scala
```diff
@@ -0,0 +1,24 @@
...
+object PathFactory {
+  def buildPath(rawString: String, fileSystems: List[FileSystem]): Path = {
+    fileSystems.toStream map { fs => Try(fs.getPath(rawString)) } collectFirst { case Success(p) => p } getOrElse {
+      throw new IllegalArgumentException(s"Could not find suitable filesystem to parse $rawString")
+    }
+  }
+
+  def buildPathAsDirectory(rawString: String, fileSystems: List[FileSystem]): Path = {
+    fileSystems.toStream collect {
```
geoffjentry: This is really similar to `buildPath` - is there a way to unify this a bit?

engine/src/main/scala/cromwell/engine/backend/io/filesystem/gcs/GcsFileSystemProvider.scala
```diff
@@ -0,0 +1,249 @@
...
+      case (s: NioGcsPath, d: NioGcsPath) =>
+        def moveInner = {
+          val storageObject = client.objects.get(s.bucket, s.objectName).execute
+          client.objects.rewrite(s.bucket, s.objectName, d.bucket, d.objectName, storageObject).execute
+        }
+
+        def tryMove(retries: Int): Unit = Try(moveInner) match {
+          case Success(_) =>
+          case Failure(ex: GoogleJsonResponseException) if ex.getStatusCode == 404 && retries > 0 =>
+            /* Could not use TryUtil here because it requires a WorkflowLogger, which we can't get form here
```
geoffjentry: form == from

engine/src/main/scala/cromwell/engine/backend/io/filesystem/gcs/GcsFileSystemProvider.scala
```diff
@@ -0,0 +1,249 @@
...
+    case None =>
+      val fs = GcsFileSystem(this)
+      defaultFileSystem = Option(fs)
+      fs
+  }
+
+  override def isHidden(path: Path): Boolean = throw new NotImplementedError()
+  override def newDirectoryStream(dir: Path, filter: Filter[_ >: Path]): DirectoryStream[Path] = dir match {
+    case gcsDir: NioGcsPath =>
+      val listRequest = client.objects().list(gcsDir.bucket)
```
geoffjentry: can you put this case in a function please

geoffjentry:
:+1:

PR #526, opened 2016-03-08
Added retries on file move to resolve a race condition Closes #517

### LINKED ISSUE #517: Retry Call Caching, JES Error 404
JES gave an intermittent 404 error when trying to copy outputs from the cached call directory to the current call directory (discovered by Vivek when using Call Caching for GoTC) .

AC: Have Cromwell retry Call Caching when upon receipt of the 404 error.

cjllanwarne:


mcovarr:
If I were to review this I'd say :+1: , though you and @Horneth will have some conflicts to sort out... :smile:

geoffjentry:
A little brazilian jiu jitsu match between the two of them after standup?

Horneth:
Bah this is America, I'll bring my :gun:

src/main/scala/cromwell/engine/io/gcs/GoogleCloudStorage.scala
```diff
@@ -148,8 +148,25 @@ case class GoogleCloudStorage private(client: Storage) extends IoInterface {
...
+        /* Could not use TryUtil here because it requires a WorkflowLogger, which we can't get form here
+         * TODO From a more general perspective we may need to add a logging capability to the IOInterface
+         */
+        Thread.sleep(polling.toMillis)
+        tryMove(retries - 1)
+      case Failure(ex: GoogleJsonResponseException) if ex.getStatusCode == 404 =>
+      case Failure(ex) => throw ex
+    }
+
+    tryMove(3)
```
geoffjentry: I must have felt a great disturbance in the force or something as I'm actually going to complain about a magic number. Could you at least say why 3?

Now excuse me while I go bathe the shame off of me for being a magic number naysayer

cjllanwarne: Because 5 is right out?

https://www.youtube.com/watch?v=xOrgLj9lOwk

cjllanwarne:
In that case I brazenly ignore the wheel and choose @Horneth to review! Pistols at dawn!

src/main/scala/cromwell/engine/io/gcs/GoogleCloudStorage.scala
cjllanwarne: More seriously, I chose 3 and 500 because that's what was used elsewhere when retrying JES connections in this file.

geoffjentry: If they're being used in multiple places I'd say it should be pulled out to a constant. i don't care that much though

Horneth:
:+1: This code will basically move to GcsFileSystemProvider once I'll rebase on top of this but it will stay pretty much the same

PR #527, opened 2016-03-09
Moving towards a multi-multi-project cromwell repo, many rough edges

geoffjentry:
Far from perfect, but starts moving towards a repo with multiple sub-projects. 

Key highlights:

- All current code except for Main and Main related stuff has been moved to an `engine` subproject. Yes astute reader, you're correct - not everything actually belongs there. The idea is that we'll peel stuff out as we go along.
- Creates a `core` subproject (currently empty) for code which doesn't have a true home but should be shared. For example, if something would otherwise move from `engine` to `backend` for no other reason than the dependency direction and it really isn't a backend-y thing.
- The root project depends on `core`, `engine` and `backend`, `engine` depends on `backend` and `core` and `backend` on `core`
- Started to decompose our SBT structure into more digestible chunks to make it easier to figure out what is where

Key lowlights:

- All of the code which went into `engine` retained `cromwell` as the package instead of `cromwell.engine`, which causes IntelliJ to whine. I thought it'd be weird as there *is* a `cromwell.engine` package already (`cromwell.engine.engine`?). However I view this as a temporary thing, I figure stuff is going to be moving around and we can figure out a package/naming structure which makes sense once we see what's where.
- Some of the abstractions are a little hokey at the moment as a side effect of 99.9999999% of the code living in one subproject. I expect that to start changing soon, mainly I was trying to give some examples of a path forward
- I'm almost positive there's a lot more stuff going on than necessary (e.g. multiple de-dups, etc). Example - I want to disable assembly on all subprojects but backend, but i still wanted to run tests. Wasn't clear how to have root assembly run tests but not assembly on other subprojects
- I'm 100% certain that some of the sbt directives aren't totally idiomatic (which causes the previous statement)
- Some of the ancillary stuff (e.g. conf files) might need to be rethought. I chalk that up along the same lines as the package naming in that some of this stuff will become more clear as it progresses


geoffjentry:
@kshakir @cjllanwarne

jainh:
Amazing! That is the right way to deal with multi project. I don't know if you remember Geoff, Fransico also asked for this approach and I shared quite few links on gitter. But, anyways that's an amazing start for decoupling projects or in other words defining dependency chain. :)

geoffjentry:
@jainh At some point I mentioned that I was almost positive we were talking past each other and arguing about nothing real ;)

engine/src/test/scala/ReadTsvWorkflowSpec.scala
```diff
@@ -1,12 +1,9 @@
...
-import akka.testkit._
-import wdl4s.types.{WdlFileType, WdlIntegerType, WdlStringType, WdlArrayType}
-import wdl4s.values.{WdlFile, WdlInteger, WdlArray, WdlString}
-import cromwell.CromwellSpec.DockerTest
+import akka.testkit.EventFilter
 import cromwell.util.SampleWdl
-
-import scala.language.postfixOps
+import wdl4s.types.{WdlFileType, WdlStringType, WdlIntegerType, WdlArrayType}
+import wdl4s.values.{WdlFile, WdlString, WdlInteger, WdlArray}
```
kshakir: Did you _need_ to change the imports? If not, file renames go much, much better in a commit when there are zero changes.

cjllanwarne: This seems out of place in this PR?

project/Dependencies.scala
```diff
@@ -0,0 +1,75 @@
...
+    "com.typesafe.slick" %% "slick-hikaricp" % slickV,
+    "org.hsqldb" % "hsqldb" % "2.3.2",
+    "mysql" % "mysql-connector-java" % "5.1.36",
+    "org.liquibase" % "liquibase-core" % "3.4.2",
+    // This is to stop liquibase from being so noisy by default
+    // See: http://stackoverflow.com/questions/20880783/how-to-get-liquibase-to-log-using-slf4j
+    "com.mattbertolini" % "liquibase-slf4j" % "2.0.0"
+  )
+
+  val engineDependencies = List(
```
cjllanwarne: As in, cromwell-engine dependencies?

Shouldn't this be in the cromwell-engine build.sbt?

project/Merging.scala
```diff
@@ -0,0 +1,57 @@
+import sbt._
```
cjllanwarne: Is this file just boilerplate or does it have a goodly purpose?

engine/src/test/scala/ReadTsvWorkflowSpec.scala
geoffjentry: @kshakir @cjllanwarne take a look at where this file was previously (it wasn't in the right spot). The imports were what intellij asked me to do

project/Dependencies.scala
geoffjentry: Generally you want to keep all of your stuff at the root levels. As an example, look at how SBT does it themselves.

project/Merging.scala
geoffjentry: It's used for de-duping during sbt assembly. It's just a bunch of crapola that was cluttering up build.sbt

project/plugins.sbt
```diff
@@ -1,9 +1,7 @@
+addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.14.1")
+addSbtPlugin("io.spray" % "sbt-revolver" % "0.7.2")
+addSbtPlugin("org.scalastyle" %% "scalastyle-sbt-plugin" % "0.7.0")
```
cjllanwarne: stowaway change?

geoffjentry: Brought over from backend's plugins. My (perhaps incorrect) understanding is that sbt doesn't honor subproject plugins.sbt

src/test/scala/cromwell/MainSpec.scala
```diff
@@ -9,17 +9,16 @@ import akka.util.Timeout
 import better.files._
 import cromwell.util.FileUtil._
 import cromwell.util.SampleWdl
-import cromwell.util.SampleWdl._
+import cromwell.util.SampleWdl.{EmptyWorkflow, GoodbyeWorld, ThreeStep}
 import org.apache.commons.io.output.TeeOutputStream
 import org.scalatest.concurrent.TimeLimitedTests
 import org.scalatest.time.Span
-import org.scalatest.{BeforeAndAfterAll, FlatSpec, Matchers}
+import org.scalatest.{Matchers, BeforeAndAfterAll, FlatSpec}
```
cjllanwarne: This is a really bizarre reordering. 

Also beware intellij auto-imports if that's what caused this. They can't handle implicits in certain cases.

geoffjentry: I'll reorder :santa: 

I had to fix a lot of the imports manually (due to lost implicits and such), was probably done by hand

project/Dependencies.scala
cjllanwarne: so even if it's a dependency of a sub-project of a sub-project you list the specific dependencies for everything at the root level?

I mean, sure, ok, I almost agree with that.

cjllanwarne:
:+1:

engine/src/test/scala/ReadTsvWorkflowSpec.scala
geoffjentry: That said, if you want me to put the file back on its own little island I can certainly do that

project/plugins.sbt
cjllanwarne: Ok - I wasn't sure whether we were sneaking in scalastyle (which I *do* want, just not in this PR!)

geoffjentry: It also doubles as a reminder to myself that I need to look at the stylesheet @gauravs90 put together ;)

kshakir:
> All of the code which went into engine retained cromwell as the package instead of cromwell.engine, which causes IntelliJ to whine. I thought it'd be weird as there is a cromwell.engine package already (cromwell.engine.engine?).

I'm still trying to understand that statement in the context of what I'm seeing in diff/intellij. I'm especially not sure about the particular IntelliJ "whine". If it's about the packages being incorrect in all the files, that warning is because there's a missing "cromwell" directory in the file paths.

* engine/src/main/scala/{ => cromwell}/*
* engine/src/test/scala/{ => cromwell}/*

project/Dependencies.scala
geoffjentry: discussed in person

geoffjentry:
@kshakir :facepalm:

build.sbt
```diff
@@ -1,174 +1,21 @@
-import sbtassembly.MergeStrategy
-import sbtrelease.ReleasePlugin._
+import Settings._
+import Testing._
 
-name := "cromwell"
+lazy val core = Project(id="core", base=file("core"))
```
kshakir: Can you switch all of these circa 0.12 `Project.apply()` to the 0.13 macro [`project`](http://www.scala-sbt.org/0.13/docs/ChangeSummary_0.13.0.html#.sbt+format+enhancements)? It should make IntelliJ happier.

```scala
lazy val core = (project in file("core"))
```

project/Dependencies.scala
```diff
@@ -0,0 +1,75 @@
+import sbt._
+import Keys._
```
kshakir: This import isn't used. Mind optimizing imports on all the new files?

kshakir:
Review complete. Once comments addressed :+1:

geoffjentry:
Pushing this despite a travis yellow light - I've seen travis work a couple of times now w/o any actual code change, but people keep slipping in minor merges to develop while i'm waiting and i need to rebase on every merge

gauravs90:
I like this too :+1:

PR #534, opened 2016-03-09
Make sure the test configs can cascade through

geoffjentry:
I'm going to look for syntax to abstract that stanza out into a common location but for now this resolves the problem


build.sbt
```diff
@@ -10,6 +10,12 @@ lazy val backend = (project in file("backend"))
 lazy val engine = (project in file("engine"))
   .dependsOn(core % "test->test;compile->compile", backend)
   .settings(engineSettings:_*)
+  .configs(AllTests).settings(inConfig(AllTests)(Defaults.testTasks): _*)
```
geoffjentry: I made a comment on the PR but ideally we wouldn't need this boilerplate on every subproject which uses it. However I need to fix this problem sooner than I have time to investigate the abstraction

geoffjentry:
@scottfrazer @gauravs90

build.sbt
cjllanwarne: So this is copy/pasted from another file?

geoffjentry: If by another file you mean a few lines below, yeah

gauravs90:
:+1:

francares:
:+1:

PR #548, opened 2016-03-11
Making the Validate API endpoint use the new ValidateActor. Closes #489

### LINKED ISSUE #489: Validation Endpoint must use ValidateActor with the burden of converting to a StatusCode on itself

gauravs90:
CromwellApiHandler now uses a ValidateActor to validate a WF received via the API endpoint

### DIFF
--- a/engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
@@ -21,11 +21,7 @@ import scala.language.postfixOps
 import scala.util.{Try, Failure, Success}
 
 object ValidateActor {
-  private val tag = "ValidateActor"
-
-  def props(wdlSource: WdlSource, wdlJson: Option[WdlJson], workflowOptions: Option[WdlJson]): Props = {
-    Props(new ValidateActor(wdlSource, wdlJson, workflowOptions))
-  }
+  def props(): Props = Props(new ValidateActor())
 
   sealed trait ValidateActorMessage
   sealed trait ValidationResult extends ValidateActorMessage
@@ -36,40 +32,22 @@ object ValidateActor {
                                workflowOptions: Option[WorkflowOptions],
                                runtimeAttributes: Seq[Set[String]]) extends ValidationResult
   case class ValidationFailure(reason: Throwable) extends ValidationResult
-  case object ValidateWorkflow extends ValidateActorMessage
+  case class ValidateWorkflow(wdlSource: WdlSource, workflowInputs: Option[WdlJson], workflowOptions: Option[WdlJson]) extends ValidateActorMessage
 }
 
 // TODO: Declarations cannot be validated here currently because of it's dependency on EngineFunctions, and in turn on IOManager and WfContext
-class ValidateActor(wdlSource: WdlSource, workflowInputs: Option[WdlJson], workflowOptions: Option[String])
+class ValidateActor()
   extends Actor with LazyLogging {
 
-  import ValidateActor.{ValidateWorkflow, tag}
+  import ValidateActor.ValidateWorkflow
   import context.dispatcher
 
   override def receive = {
-    case ValidateWorkflow =>
-      validateWorkflow(sender())
-    // NOTE: self shuts down when the parent PerRequest shuts down
-  }
-
-  private def validateWorkflow(sentBy: ActorRef): Unit = {
-    logger.info(s"$tag for $sentBy")
-    // TODO: IMO, the responsibility for returning a status code should be a part of the client. This class should only return a validation result,
-    // and the callee should convert that to a status code or anything that it wants
-    validateAll() map {
-      case _: ValidationSuccess =>
-        logger.info(s"$tag success $sentBy")
-        sentBy ! RequestComplete(
-          StatusCodes.OK,
-          APIResponse.success("Validation succeeded."))
-
-      case ValidationFailure(reason) =>
-        val messageOrBlank = Option(reason.getMessage).mkString
-        logger.info(s"$tag error $sentBy: $messageOrBlank")
-        sentBy ! RequestComplete(
-          StatusCodes.BadRequest,
-          APIResponse.fail(reason))
-    }
+    case ValidateWorkflow(wdlSource, workflowInputs, workflowOptions) =>
+      val requester = sender()
+        // `validateAll(..)` guarantees we will get a `ValidationResult`, and nothing else
+      validateAll(wdlSource, workflowInputs, workflowOptions) map { requester ! _ }
+    case unknownMsg => logger.error(s"${this.getClass.getName} received an unknown message: $unknownMsg")
   }
 
   /**
@@ -78,13 +56,13 @@ class ValidateActor(wdlSource: WdlSource, workflowInputs: Option[WdlJson], workf
     * 2.) Raw inputs and it's coercion
     * 3.) Workflow Options
     * 4.) RuntimeAttributes w.r.t. a backend
-    * @return
+    * @return The result of the validation as instance of `ValidationResult`
     */
-  private def validateAll(): Future[ValidationResult] = {
+  private def validateAll(wdlSource: WdlSource, workflowInputs: Option[WdlJson], workflowOptions: Option[WdlJson]): Future[ValidationResult] = {
     (for {
       namespaceWithWorkflow <- Future(NamespaceWithWorkflow.load(wdlSource))
-      validatedInputs <- Future(validateInputs(namespaceWithWorkflow))
-      validatedWorkflowOptions <- Future(validateWorkflowOptions(CromwellBackend.backend()))
+      validatedInputs <- Future(validateInputs(workflowInputs, namespaceWithWorkflow))
+      validatedWorkflowOptions <- Future(validateWorkflowOptions(workflowOptions, CromwellBackend.backend()))
       validatedRuntimeAttrs <- Future(validateRuntimeAttributes(namespaceWithWorkflow))
     } yield ValidationSuccess(namespaceWithWorkflow, validatedInputs, validatedWorkflowOptions, validatedRuntimeAttrs)) recover {
       case reason => ValidationFailure(reason)
@@ -101,7 +79,7 @@ class ValidateActor(wdlSource: WdlSource, workflowInputs: Option[WdlJson], workf
     }
   }
 
-  private def validateInputs(namespaceWithWorkflow: NamespaceWithWorkflow): Option[WorkflowCoercedInputs] = {
+  private def validateInputs(workflowInputs: Option[WdlJson], namespaceWithWorkflow: NamespaceWithWorkflow): Option[WorkflowCoercedInputs] = {
     def rawInputs(json: WdlJson): Map[String, JsValue] = {
       Try(json.parseJson) match {
         case Success(JsObject(inputs)) => inputs
@@ -122,7 +100,7 @@ class ValidateActor(wdlSource: WdlSource, workflowInputs: Option[WdlJson], workf
   }
 
   // TODO: With PBE, this should be defined in the backend.
-  private def validateWorkflowOptions(backend: Backend): Option[WorkflowOptions] = {
+  private def validateWorkflowOptions(workflowOptions: Option[WdlJson], backend: Backend): Option[WorkflowOptions] = {
     def validateBackendOptions(options: WorkflowOptions, backend: Backend): WorkflowOptions = {
       try {
         backend.assertWorkflowOptions(options)
--- a/engine/src/main/scala/cromwell/server/CromwellServer.scala
+++ b/engine/src/main/scala/cromwell/server/CromwellServer.scala
@@ -15,7 +15,7 @@ object CromwellServer extends WorkflowManagerSystem {
   import scala.concurrent.ExecutionContext.Implicits.global
 
   val conf = ConfigFactory.load()
-  val service = actorSystem.actorOf(CromwellApiServiceActor.props(workflowManagerActor, conf), "cromwell-service")
+  val service = actorSystem.actorOf(CromwellApiServiceActor.props(workflowManagerActor, validateActor, conf), "cromwell-service")
   val webserviceConf = conf.getConfig("webservice")
 
   def run(): Future[Any] = {
--- a/engine/src/main/scala/cromwell/server/WorkflowManagerSystem.scala
+++ b/engine/src/main/scala/cromwell/server/WorkflowManagerSystem.scala
@@ -1,9 +1,9 @@
 package cromwell.server
 
-import akka.actor.ActorSystem
+import akka.actor.{Props, ActorSystem}
 import com.typesafe.config.ConfigFactory
 import cromwell.engine.backend.{Backend, CromwellBackend}
-import cromwell.engine.workflow.WorkflowManagerActor
+import cromwell.engine.workflow.{ValidateActor, WorkflowManagerActor}
 
 trait WorkflowManagerSystem {
   protected def systemName = "cromwell-system"
@@ -21,4 +21,5 @@ trait WorkflowManagerSystem {
   lazy val backend: Backend = CromwellBackend.initBackend(backendType, actorSystem)
   // For now there's only one WorkflowManagerActor so no need to dynamically name it
   lazy val workflowManagerActor = actorSystem.actorOf(WorkflowManagerActor.props(backend), "WorkflowManagerActor")
+  lazy val validateActor = actorSystem.actorOf(Props[ValidateActor], "ValidationActor")
 }
--- a/engine/src/main/scala/cromwell/webservice/CromwellApiHandler.scala
+++ b/engine/src/main/scala/cromwell/webservice/CromwellApiHandler.scala
@@ -6,18 +6,20 @@ import akka.util.Timeout
 import cromwell.engine
 import cromwell.engine._
 import cromwell.engine.backend.{Backend, CallLogs}
+import cromwell.engine.workflow.ValidateActor.{ValidationFailure, ValidationSuccess, ValidateWorkflow}
 import cromwell.engine.workflow.WorkflowManagerActor
 import cromwell.engine.workflow.WorkflowManagerActor.{CallNotFoundException, WorkflowNotFoundException}
 import cromwell.webservice.PerRequest.RequestComplete
 import spray.http.StatusCodes
 import spray.httpx.SprayJsonSupport._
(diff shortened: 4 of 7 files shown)


gauravs90:
@mcovarr and @Horneth

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
```diff
@@ -48,28 +48,15 @@ class ValidateActor(wdlSource: WdlSource, workflowInputs: Option[WdlJson], workf
...
-        val messageOrBlank = Option(reason.getMessage).mkString
-        logger.info(s"$tag error $sentBy: $messageOrBlank")
-        sentBy ! RequestComplete(
-          StatusCodes.BadRequest,
-          APIResponse.fail(reason))
-    }
+      val requester = sender()
+      logger.info(s"$tag for $requester")
+      validateAll() map {
+        // validateAll guarantees we will get a `ValidationResult`, and nothing else
```
Horneth: nit pick but in that case you can just 
```
validateAll() map { result =>
...
}
```

geoffjentry:
A philosophical question, not a leading one - what about having a persistent validation actor instead of creating them per-validation?

Horneth:
Not really for this PR I think, but looking at the `validateInputs` method, it only validates the inputs if a file was passed (makes sense), but if no input file is specified and the WDL does need an input, validation will return success, which is weird IMO.
@geoffjentry My 2 cents are I personally prefer having small short-lived actors instead of a singleton actor that handle all the requests. I think it's more robust, faster, and less leading towards godlike actors.

Horneth:
:+1:

cjllanwarne:
@Horneth does that match with current behaviour? If it currently checks that workflow inputs are satisfied  at submit-time then we should make sure that remains true.

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
```diff
@@ -48,28 +48,15 @@ class ValidateActor(wdlSource: WdlSource, workflowInputs: Option[WdlJson], workf
...
-
-      case ValidationFailure(reason) =>
-        val messageOrBlank = Option(reason.getMessage).mkString
-        logger.info(s"$tag error $sentBy: $messageOrBlank")
-        sentBy ! RequestComplete(
-          StatusCodes.BadRequest,
-          APIResponse.fail(reason))
-    }
+      val requester = sender()
+      logger.info(s"$tag for $requester")
```
Horneth: Are this log + the one below really useful ?
It gives something like
```
ValidateActor response Actor[akka://cromwell-system/user/cromwell-service/Endpoint-endpointActorName293378273336825/Endpoint-endpointActorName293378273336825#-800082011]
ValidateActor for Actor[akka://cromwell-system/user/cromwell-service/Endpoint-endpointActorName293382176345073/Endpoint-endpointActorName293382176345073#-436231765]
```

geoffjentry:
@Horneth There are certainly tradeoffs and I don't disagree w/ what you said. However consider the flip side - by defining A Validation Actor you're allowing for more granular control over performance and fault tolerance down the road, e.g. you could replace it with a router talking to a bank of VAs and doing load balancing, fiddle with its own threadpool, provide validation specific supervision in case of error, etc.

Horneth:
I totally agree with having a singleton actor for load balancing / supervision / monitoring etc.. but I think the actual validation work itself is better handled by a one-shot do-and-die actor than by a singleton actor. I don't think the actor that is responsible for load balancing, error handling etc.. should also be responsible for doing the work it's supervising.

Horneth:
@cjllanwarne the current behaviour (as in develop branch behaviour) is unchanged by this PR. It does check that workflow inputs are satisfied, but only if you give it an input file.

cjllanwarne:
@Horneth in that case I think this PR is correct to act as it does.
@Horneth @geoffjentry I agree with @Horneth

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
mcovarr: yeah agreed, these don't seem informative

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
```diff
@@ -23,7 +23,7 @@ import scala.util.{Try, Failure, Success}
 object ValidateActor {
   private val tag = "ValidateActor"
 
-  def props(wdlSource: WdlSource, wdlJson: Option[WdlJson], workflowOptions: Option[WdlJson]): Props = {
+  def props(wdlSource: WdlSource, wdlJson: Option[WdlJson] = None, workflowOptions: Option[WdlJson] = None): Props = {
```
cjllanwarne: I think it'd be neater if the parameters were not actor-scoped but instead part of the `ValidateWorkflow` message request. Also can we take this opportunity to rename `wdlJson` to `workflowInputs`?

geoffjentry:
@Horneth The more I'm thinking about this I'm transitioning from "throwing something out there" to "advocating" :)

What bothers me about this is that eventually we're going to want workflow submission also go through the same validation actor instead of doing it in multiple ways. Typically you want actor ownership to be hierarchical in nature but this way you'd have multiple parent types. Something to consider - what happens if a VA throws an exception under this model? You'll now need to have supervision code in multiple places. More abstractly what will be said is that validation is A Thing, but needs to exist independently in multiple places - if that's the case it should be pulled out into its own block.

Furthermore the validation actor should be the sort of actor which is perfect for being its own concept - it's a completely idempotent, stateless operation. Work gets sent to it, it process the work and responds to the querier.

My point about supervision is that by Right Now defining A Validation Actor (presumably owned by the kernel) what you're effectively doing is defining a validation interface. You can change how things are implemented in the future (e.g. it's really a bunch of VAs, it's firing up ephemeral VAs, whatever) and not need to change any code throughout the rest of the system as everything is still talking to the same actorRef that they were before. Alternatively if validation actors are being spun up on demand in multiple places and we decide that we somehow need to handle VAs in a special manner, it'll be a larger refactor.

All that said, at the moment only the validation webservice is talking to the VA. I'm happy to shelve this until when something else is talking to VAs.

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
geoffjentry: To play devils advocate, if you're spinning up a VA per-request, who cares?

Horneth: I actually think it's not only neater but more practical to have them parameters of the actor.
If you look at the Workflow Actor for example right now we have Actor messages case classes  a with never ending list of parameters, because the same actor is doing all the work and needs to handle every possible combination of messages and parameters + needs to maintain a complex mutable state data to know what to do when message A arrives from B but needs to wait for C to answer before processing it and in the meantime keep answering requests from D etc...
If you have smaller dedicated actors, pretty much all they need to know can be given as parameters which reduces their mutable data, which is good.

geoffjentry:
@Horneth To play devil's advocate against myself - a bit, as this doesn't matter if it's a short lived or long lived VA, in a world where we want both the validation API and submission to go through the validation actor (reclal that currently submission validation is done by the WMA) the submission endpoint could send the work to the VA who then sends it to the WMA if it is OK. That would at least make the VA fully under the umbrella of the webservice portion of the actor system

Horneth:
@geoffjentry I think we're saying the same thing. This is what I'm thinking : [link](https://www.lucidchart.com/publicSegments/view/288aace2-3423-4364-a5f6-6eff306e754e/image.png)

(Which is exactly what is in the other "proof of concept" branch but with a DB Actor instead of Validation Actor btw)

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
geoffjentry: @Horneth You're going from one horrible extreme to another (not horrible) extreme here. The (well, A)  problem with WA is that it owns too many responsibilities, not that it doesn't have everything it needs as a parameter. However having a single responsibility doesn't mean it's literally doing *one* thing, just that it does one type of thing which can be parameterized

geoffjentry:
@Horneth Yeah, almost exactly that. The only beef I have w/ that is that I'd call the ephemeral VAs behind it as a premature optimization, but it is certainly a valid optimization. Similarly maybe there are N fixed VAs behind it or something like that (I'm not advocating for either).

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
Horneth: I took the example of the WA because it's the more relevant. I'm not saying that the WA needs to have everything it needs as parameters. I'm saying that steering towards a model where we have more (in numbers), more specialized, smaller (in responsibilities, hence code size which is always nice) actors is better than what we do now, and that putting input data in the attributes of an actor, instead of putting them in huge messages, will drive towards that model.

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
```diff
@@ -78,7 +65,7 @@ class ValidateActor(wdlSource: WdlSource, workflowInputs: Option[WdlJson], workf
     * 2.) Raw inputs and it's coercion
     * 3.) Workflow Options
     * 4.) RuntimeAttributes w.r.t. a backend
-    * @return
+    * @return The result of the validation as instance of `ValidationResult`
     */
   private def validateAll(): Future[ValidationResult] = {
```
mcovarr: This might be the PR for making this method work in a more standard way, i.e. the `Future` doesn't need to always be successful:

```scala
private def validateAll: Future[ValidationResult] = for {
  ...
} yield ValidationSuccess(...)
```

and above:

```scala

def respond(response: ValidationResult): Unit = {
  requester ! response
  context.stop(self)
}

validateAll recover { case t => ValidationFailure(t) } foreach respond
```

Horneth:
For this particular case it might be an optimization, but from a general perspective I think it can be a design choice and would avoid creating new Workflow Actors-like

geoffjentry:
@Horneth Anyone who would disagree with "avoid creating new Workflow Actors-like" should be sent to the guillotine

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
cjllanwarne: If the model for our bottom-level `worker` actors is
create actor > send "start work" > actor works > actor sends "completion" back > actor destroyed
then create and start happen in the same place and it actually doesn't functionally make any difference where the params-sending happens.

However, if they're in the "start work" message, then the message to the router is exactly the same as the message to the worker. Which appeals to me sense of neatness.

cjllanwarne: It also allows easier swap-out from router to worker and vice versa.

Horneth:
@geoffjentry I'll check my history book again but I'm pretty sure that's exactly the reason why our last king lost his head

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
geoffjentry: Good point

geoffjentry:
@Horneth You might be thinking of Marie Antoinette when she said "let them eat a monolithic actor"

cjllanwarne:
So @Horneth @geoffjentry given that everyone is in agreement - the more relevant question is, are you advocating this in *this* PR, or a future ticket?

mcovarr:
Le workflow, c'est moi -- WA

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
Horneth: I agree, my point is as long as it doesn't add unnecessary internal mutable data management / maintenance to the worker actor then I don't really mind if the data is in the message.

One example of something nice to have as a parameter is the actor to send the response to. If you include it in the message, then you have to set a variable somewhere (in the actor data probably), or keep passing it in every single message until you're finally ready to send your response. If you have it as a parameter it's available from the very beginning to the end, for no cost.

geoffjentry:
@cjllanwarne I'd prefer to see it talking to a single point (currently a single VA, as @Horneth & I discussed, if performance, reliability or complexity call for it, to be morphed into something else). I don't care that much if it happens here, but it's not an onerous change to make on this PR. I *will* care once something else is talking to a VA

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
geoffjentry: @Horneth What you're describing in your 2nd paragraph is really more about how to manage transactional behavior, and you're right that a single actor managing other things isn't the right place for it. Take a look (if you haven't already) at the saga pattern, which I think you'll find is not dissimilar to your DB POC.

One way of implementing it in akka would be to have the actor which wants a multi-stage transaction to happen spin up an actor for that specific transaction. When it is done it kills itself and the main actor (e.g. the WorkflowActor) sees that the child has died and knows that transaction is complete.

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
gauravs90: @mcovarr Can you explain how that is more standardized? I think this reads better: 

```scala
validateAll() map { validationResult =>
        requester ! validationResult
        context.stop(self)
      }
```

with `validateAll()` remaining the same. To me, just for sending the response and stopping itself doesn't warrant a new function with only side effects

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
gauravs90: @cjllanwarne @geoffjentry @Horneth 
1. VAs-per-request appeals more to me because of the reason already mentioned by @Horneth .

2.) I think passing the arguments with the actor (and not in the message) makes more sense in this case where we are creating VAs per request, as it ties that VA to just one workflow (request). Passing it in message, you can potentially send messages with different workflows to this actor (which breaks the very tenet that we want single VAs per request). On a tangential note, this same logic applied to instantiation of a Backend in PBE where we used a single TaskDescriptor for instantiating a backend, which not only ties a backend to the philosophy of task-at-a-time model, but avoids passing a TaskDesc each time to prepare(), execute(), cleanup() etc. where it is difficult to guarantee that you are in fact passing the same instance of TaskDesc to each of those methods.

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
gauravs90: I agree with that, but something like that was there already before, and I did't want to remove it. Now that you said I'll remove/replace it :smile:

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
geoffjentry: @gauravs90 Note that while @Horneth was talking about transient VAs, he wasn't talking about it as structured here.

Also since a VA is stateless there is no reason why it should be directly tied to a single WDL, there is no chance of crosstalk causing harm.

gauravs90: @geoffjentry  I'm not sure if I understand what you're saying then. Let's have a short Hangouts about that whenever you're free?

gauravs90:
@geoffjentry @mcovarr If i understood correctly, is there anything else that needs to be a part of this PR?

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
geoffjentry: @gauravs90 - @Horneth was suggesting something like [this](https://www.lucidchart.com/publicSegments/view/288aace2-3423-4364-a5f6-6eff306e754e/image.png), see the link he posted below. What I was arguing was that while I agreed with the drawing he put up there (ignoring the per-use worker or if it was a pool) that I thought that at the moment it'd be a premature optimization to do so and you'd just want to have that first layer VA. In a future where you want more workers behind it that VA address can simply become a router.

Does that make sense? If not, let's catch up on monday.

cjllanwarne:
@gauravs90 is it possible to make ValidateActor a singleton in this PR or would you rather spin that out as a new PR?

gauravs90:
@cjllanwarne Are you talking on the lines of something what @Horneth proposed? ([as mentioned here](https://www.lucidchart.com/publicSegments/view/288aace2-3423-4364-a5f6-6eff306e754e/image.png))

cjllanwarne:
@gauravs90 as per @geoffjentry's final comment - like that diagram but without the bottom row

gauravs90:
@cjllanwarne Cool. I don't completely agree with that design, but I'll do it in this PR. Let's see how to evolve it more like what the diagram proposes sometime later.

geoffjentry:
@gauravs90 including the bottom row is an optimization once/if the original actor is overloaded. The right optimization at that point might not be what we think it is now - but it'll be trivial to change on the future as clients of this actor were talking about can continue to use the same actor ref without being any the wiser

francares:
@gauravs90 @geoffjentry @mcovarr @cjllanwarne @Horneth Hey guys! Can we discuss this during today's meeting before doing any change?

francares:
I think we can use a BalancingDispatcher for this use case.

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
mcovarr: The only standard `Future` that never fails is `Future.successful`.  The `receive` method of an actor returns `Unit`, so this really is all about the side effects.  The use of `map` is misleading if it makes the situation appear otherwise.  IMHO having the `recover` here rather than buried in the method that returns `Future[ValidationResult]` makes this all more clear.

mcovarr:
The major topics were discussed and resolved during our meeting today.  :+1: delta my request to fix the unfailing `Future`.

gauravs90:
@mcovarr I made it singelton but :frowning: 
We wanted it right?

mcovarr:
Yeah that's good for now, feel free to change the implementation behind the interface if you need to.

gauravs90:
@mcovarr Well okay then. Let this remain a singleton as we discussed in the meeting. Later on, if we feel it's a bottleneck, we can make it a kinda router or something.

PR #562, opened 2016-03-15
Added scalaz lib to ValidateActor. Closes #543

### LINKED ISSUE #543: Use scalaz stuff from WorkflowDescriptor in ValidateActor

francares:
Added scalaz lib support to ValidateActor in order to match Workflow descriptor.

### DIFF
--- a/engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
@@ -1,30 +1,29 @@
 package cromwell.engine.workflow
 
-import akka.actor.{Actor, ActorRef, Props}
+import akka.actor.{Actor, Props}
 import com.typesafe.scalalogging.LazyLogging
-import cromwell.engine.backend.{Backend, CromwellBackend}
+import cromwell.engine.ErrorOr
 import cromwell.engine.backend.runtimeattributes.CromwellRuntimeAttributes
-import cromwell.engine.workflow.ValidateActor.{ValidationResult, ValidationFailure, ValidationSuccess}
+import cromwell.engine.backend.{Backend, CromwellBackend}
+import cromwell.engine.workflow.ValidateActor.{ValidationFailure, ValidationResult, ValidationSuccess}
 import cromwell.util.TryUtil
-import spray.json
-import spray.json.{JsObject, _}
-import cromwell.webservice.APIResponse
-import cromwell.webservice.PerRequest.RequestComplete
-import spray.http.StatusCodes
 import spray.json._
 import wdl4s._
-import spray.httpx.SprayJsonSupport._
-import cromwell.webservice.WorkflowJsonSupport._
 
 import scala.concurrent.Future
 import scala.language.postfixOps
-import scala.util.{Try, Failure, Success}
+import scala.util.{Failure, Success, Try}
+import scalaz.Scalaz._
+import scalaz.Validation.FlatMap._
 
 object ValidateActor {
-  def props(): Props = Props(new ValidateActor())
+  private val EmptyJson: String = "{}"
+
+  def props(backend: Backend): Props = Props(new ValidateActor(backend))
 
   sealed trait ValidateActorMessage
   sealed trait ValidationResult extends ValidateActorMessage
+
   // This ADT contains all the resolved / validated components of the workflow (minus Declarations).
   // It should be used in the future to indicate if the workflow succeeded validation. The callee should reuse these materialized information.
   case class ValidationSuccess(namespaceWithWorkflow: NamespaceWithWorkflow,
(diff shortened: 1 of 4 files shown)


francares:
@mcovarr and @cjllanwarne are the chosen ones for this review.

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
```diff
@@ -56,67 +55,80 @@ class ValidateActor()
...
     */
-  private def validateAll(wdlSource: WdlSource, workflowInputs: Option[WdlJson], workflowOptions: Option[WdlJson]): Future[ValidationResult] = {
-    (for {
-      namespaceWithWorkflow <- Future(NamespaceWithWorkflow.load(wdlSource))
-      validatedInputs <- Future(validateInputs(workflowInputs, namespaceWithWorkflow))
-      validatedWorkflowOptions <- Future(validateWorkflowOptions(workflowOptions, CromwellBackend.backend()))
-      validatedRuntimeAttrs <- Future(validateRuntimeAttributes(namespaceWithWorkflow))
-    } yield ValidationSuccess(namespaceWithWorkflow, validatedInputs, validatedWorkflowOptions, validatedRuntimeAttrs)) recover {
-      case reason => ValidationFailure(reason)
+  private def validateAll(wdlSource: WdlSource, workflowInputs: Option[WdlJson], workflowOptions: Option[WdlJson]): Future[ValidationResult] = Future {
```
mcovarr: I'm not seeing anything that would require this to be wrapped in a `Future`

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
```diff
@@ -56,67 +55,80 @@ class ValidateActor()
...
+    val namespace = validateNamespace(wdlSource)
+    val result = namespace match {
```
mcovarr: Why isn't this structured just like `WorkflowDescriptor` where errors are potentially aggregated across both namespace and workflow options validation?

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
```diff
@@ -36,17 +34,18 @@ object ValidateActor {
...
-  import ValidateActor.ValidateWorkflow
+  import ValidateActor.{EmptyJson, ValidateWorkflow}
   import context.dispatcher
 
   override def receive = {
     case ValidateWorkflow(wdlSource, workflowInputs, workflowOptions) =>
       val requester = sender()
-        // `validateAll(..)` guarantees we will get a `ValidationResult`, and nothing else
-      validateAll(wdlSource, workflowInputs, workflowOptions) map { requester ! _ }
+      // `validateAll(..)` guarantees we will get a `ValidationResult`, and nothing else
```
mcovarr: Currently a `Future[ValidationResult]`, but maybe it should be just a `ValidationResult` :smile:

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
```diff
@@ -56,67 +55,80 @@ class ValidateActor()
...
+        val inputs = validateInputs(ns, workflowInputs)
+        val runtimeAttributes = validateRuntimeAttributes(ns)
+        (options |@| inputs |@| runtimeAttributes) ((opt, in, runAttr) => ValidationSuccess(ns, Option(in), Option(opt), runAttr))
+
+      case scalaz.Failure(failure) => failure.failure
+    }
+
+    result match {
+      case scalaz.Success(success) => success
+      case scalaz.Failure(failure) =>
```
mcovarr: This variable should be called something plural like `failures` since it's a `NonEmptyList[String]`.

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
```diff
@@ -56,67 +55,80 @@ class ValidateActor()
     * 2.) Raw inputs and it's coercion
```
mcovarr: it's => its

mcovarr: it's coercion => their coercions

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
```diff
@@ -56,67 +55,80 @@ class ValidateActor()
...
-    def rawInputs(json: WdlJson): Map[String, JsValue] = {
+  private def validateInputs(namespaceWithWorkflow: NamespaceWithWorkflow, workflowInputs: Option[WdlJson]): ErrorOr[WorkflowCoercedInputs] = {
+    def validateRawInputs(json: WdlJson): ErrorOr[Map[String, JsValue]] = {
       Try(json.parseJson) match {
-        case Success(JsObject(inputs)) => inputs
-        case Failure(reason: Throwable) => throw new Throwable(reason.getMessage)
-        case _ => throw new IllegalArgumentException(s"Failed to parse the workflow inputs JSON: $json")
+        case Success(JsObject(inputs)) => inputs.successNel
+        case Failure(reason: Throwable) => s"Workflow contains bad inputs JSON: ${reason.getMessage}".failureNel
+        case _ => s"Workflow contains bad inputs JSON: $json".failureNel
```
mcovarr: Since you've broken this out into a separate case, this message should be a bit more detailed.  Technically a bare string is valid JSON but not a valid `JsObject`.

cjllanwarne:
Can we also have a test to exercise the multiple errors validations?

mcovarr:
Yes if the validation is going to be centralized here there have to be tests that are at least equivalent to what's in `WorkflowDescriptor`.

cjllanwarne:
:+1:

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
gauravs90: @mcovarr My two cents: I think it's better to have a non-blocking calculation here, since this actor is now a singleton and might potentially receive simultaneous requests.

geoffjentry: I really dislike having Futures running around inside of an actor as it breaks down the fundamental simplifying abstraction of the actor model - that inside an actor there is a single threaded event loop - making it easier to reason about the internals. 

OTOH we've monstrously overused Futures running around inside of Cromwell's actors, I was hoping this refactor could be the start of a new age where we moved away from that, not adding to it.

One way of handling it would be to use the pipeTo pattern and pipe the Future directly to the caller, although I'm still uneasy with that unless it is a) necessary or b) bubbling all the way back out to a REST endpoint.

What I had meant the other day was literally handling these validations in the single threaded event loop, realizing that given enough load it'd become insufficient. My suggestion was to wait until it's clear that there's a problem to put further structure behind this actor to handle that load via more workers (e.g. your per-request model or francisco's suggestion of a balancing dispatcher) which would be under the supervision of this actor. My primary point there was that we don't know what number of requests would cause problems - it could be 1/s or it could be 100000000000000000000000000/s, and it didn't seem worth adding the structure until we knew it was likely to be a real world problem.

The above said, if that really makes you uneasy I don't care *that* strongly about putting worker structure behind this actor - I've just been saying all along that IMO it was a premature optimization which was trivially fixed down the road.

mcovarr: Yeah Jeff pretty much summed up my concerns.  If all this actor does is spin off Futures the parallelism is essentially out of control. Eventually this will hit a limit when the threads in this actor's `ExecutionContext` have been exhausted. But if you're being bottlenecked by a single validation actor wouldn't it be better to have the topmost validation actor act as a router to worker validation actors and use knowledge of Cromwell's current state to decide whether to elastically bring up more workers or simply push back on requests?

More parallelism in arbitrary parts of the application is not necessarily better.  The throughputs of the validation and execution stages will likely vary with configuration, so it would be good if the elasticities in these stages responded in reasonable ways with respect to one another rather than always just going as wide as possible.

I have broader issues with the way Cromwell currently validates and executes workflows, written up [here](#564).  While I don't currently think refactoring along these lines is strictly required for pluggable backends, it could certainly help with the issues we've already seen where Cromwell is not well behaved under heavy submission loads.

francares: I will remove the future (agree on Miguel's comment about control) but I still think it's a mistake to have a singleton as the first step in a web-service implementation. In CCC there are people working on a component for creating workflows executions on the fly based on federated data sets and it may be a bottleneck for us and by definition is a bad idea to have a singleton behind a web-service.

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
francares: I did this because I wanted to have a monad with the most number of validations in it and I just need ns to make it work.

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
francares: I will change it.

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
francares: It was there but fixed.

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
geoffjentry: Francisco - if you want to figure out ahead of time what the performance envelope for the ValidationActor is such that the mailbox would get overloaded, the please do so. My only request was that that be understood prior to adding parallelism to the procedure and knowing that in fact it'd fall over under anything approaching realistic load.

OTOH I also said elsewhere in here that if people really want to just add it, go ahead. It's not worth the argument as it would probably need to be put in anyways

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
francares: What about: "Workflow inputs JSON cannot be parsed to JsObject"

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
```diff
@@ -56,67 +55,80 @@ class ValidateActor()
...
-  private def validateInputs(workflowInputs: Option[WdlJson], namespaceWithWorkflow: NamespaceWithWorkflow): Option[WorkflowCoercedInputs] = {
-    def rawInputs(json: WdlJson): Map[String, JsValue] = {
+  private def validateInputs(namespaceWithWorkflow: NamespaceWithWorkflow, workflowInputs: Option[WdlJson]): ErrorOr[WorkflowCoercedInputs] = {
+    def validateRawInputs(json: WdlJson): ErrorOr[Map[String, JsValue]] = {
       Try(json.parseJson) match {
-        case Success(JsObject(inputs)) => inputs
-        case Failure(reason: Throwable) => throw new Throwable(reason.getMessage)
-        case _ => throw new IllegalArgumentException(s"Failed to parse the workflow inputs JSON: $json")
+        case Success(JsObject(inputs)) => inputs.successNel
+        case Failure(reason: Throwable) => s"Workflow contains bad inputs JSON: ${reason.getMessage}".failureNel
```
francares: I will change this message from "Workflow contains bad inputs JSON" to "Workflow contains invalid inputs JSON"

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
```diff
@@ -56,67 +55,80 @@ class ValidateActor()
...
-    workflowOptions match {
-      case Some(optionsJson) =>
-        WorkflowOptions.fromJsonString(optionsJson) match {
-          case Success(o) => Option(validateBackendOptions(o, backend))
-          case Failure(e) => throw new IllegalArgumentException(s"Failed to validate the workflow options JSON: ${e.getMessage}", e)
-        }
-      case _ => None
+    WorkflowOptions.fromJsonString(workflowOptions.getOrElse(EmptyJson)) match {
+      case Success(o) => validateBackendOptions(backend, o)
+      case Failure(e) => s"Workflow contains bad options JSON: ${e.getMessage}".failureNel
```
francares: Same here.

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
mcovarr: The monad is fail-fast and defeats the point of the applicative in accumulating errors.  This should mimic the structure of the validation in `WorkflowDescriptor` as much as possible.

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
```diff
@@ -56,67 +55,80 @@ class ValidateActor()
...
-          case Failure(reason) => throw new IllegalArgumentException("Failed to coerce inputs", reason)
+    validateRawInputs(workflowInputs.getOrElse(EmptyJson)) match {
+      case scalaz.Success(inputs) =>
+        namespaceWithWorkflow.coerceRawInputs(inputs) match {
+          case Success(r) => r.successNel
+          case Failure(e: ThrowableWithErrors) => scalaz.Failure(e.errors)
+          case Failure(e) => e.getMessage.failureNel
         }
-      case _ => None
+      case scalaz.Failure(f) => f.failure
```
mcovarr: I don't understand why the structure here is so different from `WorkflowDescriptor`.  The equivalent to `validateDeclarations` appears to be missing.  Yes that would be backend-dependent, but AFAIK it should be part of this validation.

francares: The goal of this PR is just to integrate Scalaz lib into ValidateActor. @gauravs90 will add missing things in the next PR based on what we discussed today.

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
francares: I will not change that here since it's part of @gauravs90 previous PR's. This PR is just to integrate Scalaz. I will try to work on this after all changes we discussed today are in place.

francares:
There are not tests associated to WorkflowDescriptor related to this topic. 
I will need to create them.

engine/src/main/scala/cromwell/engine/workflow/ValidateActor.scala
```diff
@@ -36,87 +35,99 @@ object ValidateActor {
 }
 
 // TODO: Declarations cannot be validated here currently because of it's dependency on EngineFunctions, and in turn on IOManager and WfContext
-class ValidateActor()
-  extends Actor with LazyLogging {
+class ValidateActor() extends Actor with LazyLogging {
```
francares: JFYI... I will add Backend as a dependency of this actor.

engine/src/test/scala/engine/workflow/ValidateActorSpec.scala
```diff
@@ -0,0 +1,177 @@
...
+    validateActor = system.actorOf(ValidateActor.props(backendMock))
+  }
+
+  after {
+    system.stop(validateActor)
+  }
+
+  "A Validate actor" should {
+    "return ValidationSuccess when Namespace, options, runtime attributes and inputs are valid" in {
+      within(1000 millis) {
```
mcovarr: I'd pull this out to a constant, reduce, and dilate for pokey Travis builds that might need more time:

```scala
val Timeout = 1.second.dilated
```

engine/src/test/scala/engine/workflow/ValidateActorSpec.scala
```diff
@@ -0,0 +1,177 @@
...
+        validateActor ! ValidateWorkflow(validWdlSource, Some(validInputsJson), Some(validOptionsJson))
+        expectMsgPF() {
+          case validationSuccess: ValidationSuccess =>
+            if (validationSuccess.namespaceWithWorkflow.tasks.size != 1) fail("Number of tasks is not equals to the expected one.")
+            if (!validationSuccess.coercedInputs.get.head._1.contains("hello.hello.addressee")) fail("Input is not equals to the expected one.")
+            if (!validationSuccess.workflowOptions.get.get("hello.hello.addressee").get.contains("world")) fail("Workflow option is not equals to the expected one.")
+            if (!validationSuccess.runtimeAttributes.head.head.contains("docker")) fail("Runtime attribute is not equals to the expected one.")
```
mcovarr: I think these can just be `shouldBe`s; if they fail it's clear what the expectation was and that the expectation was not met.  :smile:

engine/src/test/scala/engine/workflow/ValidateActorSpec.scala
```diff
@@ -0,0 +1,177 @@
...
+      when(backendMock.backendType).thenReturn(BackendType.JES)
+      within(1000 millis) {
+        validateActor ! ValidateWorkflow(invalidWdlSource, Some(validInputsJson), Some(validOptionsJson))
+        expectMsgPF() {
+          case validationFailure: ValidationFailure =>
+            validationFailure.reason match {
+              case validationException: ValidationException =>
+                if (!validationException.message.contains("Workflow input processing failed.")) fail("Message coming from validation exception is not the expected one.")
+                if (validationException.errors.size != 1) fail("Number of errors coming from validation exception are not the expected ones.")
+                if (!validationException.errors.head.contains("Failed to validate runtime attributes.")) fail("Message from error nr 1 in validation exception is not the expected one.")
```
mcovarr: is there error generated in this case not any more specific?

engine/src/test/scala/engine/workflow/ValidateActorSpec.scala
```diff
@@ -0,0 +1,177 @@
...
+      when(backendMock.assertWorkflowOptions(WorkflowOptions.fromJsonString(validOptionsJson).get)).thenThrow(new IllegalStateException("Some exception"))
+      within(1000 millis) {
+        validateActor ! ValidateWorkflow(validWdlSource, Some(validInputsJson), Some(validOptionsJson))
+        expectMsgPF() {
+          case validationFailure: ValidationFailure =>
+            validationFailure.reason match {
+              case validationException: ValidationException =>
+                if (!validationException.message.contains("Workflow input processing failed.")) fail("Message coming from validation exception is not the expected one.")
+                if (validationException.errors.size != 1) fail("Number of errors coming from validation exception are not the expected ones.")
+                if (!validationException.errors.head.contains("Workflow has invalid options for backend")) fail("Message from error nr 1 in validation exception is not the expected one.")
```
mcovarr: can this validate that the message text from the exception appears in the final error message?

mcovarr:
A few minor comments on the tests.  Not sure if you meant to address the actor spinning off `Future`s in this PR or a later one.  Otherwise :+1:

gauravs90:
@mcovarr If it doesn't happen here, I'll do it in my PR (#544). (Along with bringing in the declarations)

cjllanwarne:
:+1: after @mcovarr's changes

engine/src/test/scala/engine/workflow/ValidateActorSpec.scala
francares: Done.

engine/src/test/scala/engine/workflow/ValidateActorSpec.scala
francares: I'm just checking part of the error message. The part of the message I'm checking only comes from the function that validates the runtimeReq.

engine/src/test/scala/engine/workflow/ValidateActorSpec.scala
francares: I re-phrased the messages.

engine/src/test/scala/engine/workflow/ValidateActorSpec.scala
francares: It just validates part of that message. I'm just validating the part of the validation functions based on parts of the messages that identifies that the validation went through that function.

PR #561, opened 2016-03-15
Move execute and resume off BackendCall.  Closes #484.  Closes #485.

mcovarr:
@scottfrazer's Tyburning is still pending, but I don't expect this to change much even if there are issues there.

PR #563, opened 2016-03-15
Remove BackendCall Closes #516

### LINKED ISSUE #516: Remove BackendCall and its subtypes

Horneth:
Note: This is branched off of https://github.com/broadinstitute/cromwell/pull/561


Horneth:
@mcovarr @kshakir

engine/src/main/scala/cromwell/engine/backend/ExecutionHandle.scala
```diff
@@ -0,0 +1,69 @@
+package cromwell.engine.backend
+
+import cromwell.engine.{CallOutputs, ExecutionEventEntry, ExecutionHash}
+
+/**
+ * Represents a Call that is intended to be run on a specific Backend.
+ *
+ * A BackendCall is created by calling `Backend.bindCall(c: Call)`.  The
```
mcovarr: Lots of `BackendCall` comments

engine/src/main/scala/cromwell/engine/backend/ExecutionHandle.scala
```diff
@@ -0,0 +1,69 @@
...
+
+final case class RetryableExecutionHandle(throwable: Throwable, returnCode: Option[Int] = None, events: Seq[ExecutionEventEntry] = Seq.empty) extends ExecutionHandle {
+  override val isDone = true
+  override val result = new RetryableExecution(throwable, returnCode, events)
+}
+
+case object AbortedExecutionHandle extends ExecutionHandle {
+  override def isDone: Boolean = true
+  override def result: ExecutionResult = AbortedExecution
+}
```
mcovarr: newline

engine/src/main/scala/cromwell/engine/backend/Backend.scala
```diff
@@ -102,7 +100,7 @@ trait Backend {
    * Essentially turns a Call object + CallInputs into a BackendCall
    */
   def bindCall(jobDescriptor: BackendCallJobDescriptor,
```
mcovarr: The comments are not right, and in addition to its jarring name I don't think we need this anymore.  All this does in the three implementations is copy the supplied abort function into the descriptor.

engine/src/main/scala/cromwell/engine/backend/JobDescriptor.scala
```diff
@@ -29,6 +29,8 @@ case class BackendCallJobDescriptor(workflowDescriptor: WorkflowDescriptor,
                                     locallyQualifiedInputs: CallInputs = Map.empty,
                                     abortRegistrationFunction: Option[AbortRegistrationFunction] = None) extends JobDescriptor[BackendCallKey] {
 
+  def call = key.scope
```
mcovarr: Thanks, I meant to do this but never got around to it.  :smile:

engine/src/main/scala/cromwell/engine/callexecution/BackendCallExecutionActor.scala
```diff
@@ -5,26 +5,27 @@ import cromwell.engine.backend._
...
+    callTag = Option(jobDescriptor.key.tag)
   )
 
-  override val call = backendCall.call
-  override def poll(handle: ExecutionHandle) = backendCall.poll(handle)
-  override def execute(mode: ExecutionMode) = mode match {
-    case Execute => backendCall.execute
-    case Resume(jobKey) => backendCall.resume(jobKey)
-    case UseCachedCall(cachedBackendCall) => backendCall.useCachedCall(cachedBackendCall)
+  override val call = jobDescriptor.key.scope
```
mcovarr: may need to do a replace of `key.scope` with `call` to pick up your spiffy new method

engine/src/main/scala/cromwell/engine/backend/Backend.scala
Horneth: Yeah, I'm not super familiar with the way the abort function registration work so I wasn't sure.
Should we just replace whatever is calling `bindCall` with a `jobDescriptor.copy(abortFunction = abortFunction)` ?

engine/src/main/scala/cromwell/engine/backend/sge/SgeBackend.scala
```diff
@@ -51,7 +47,7 @@ case class SgeBackend(actorSystem: ActorSystem) extends Backend with SharedFileS
   override def adjustInputPaths(jobDescriptor: BackendCallJobDescriptor) = adjustSharedInputPaths(jobDescriptor)
 
   /**
-    * Exponential Backoff Builder to be used when polling for call status.
+    * Exponential Backoff Builder to be used when polling for jobDescriptor status.
```
mcovarr: hmm more like job status

engine/src/main/scala/cromwell/engine/backend/Backend.scala
mcovarr: yeah sounds good

engine/src/main/scala/cromwell/engine/backend/JobDescriptor.scala
Horneth: Yeah I think when I finally decided to do it I had already changed 95% of them to `key.scope`...
On a different note I was wondering why all of those parameterless methods are `def`s and not `val`s here, since this is `case class` and not a `trait` ?

mcovarr: hmm can't think of a great reason for that, maybe `lazy val` them to not force evaluation if callers don't care?

Issue #570, opened 2016-03-16
Remove validation from WorkflowDescriptor


mcovarr:
Already part of the plans for reworked, centralized validation

gauravs90:
This can be a part of what I'm currently working on. (#544)

gauravs90:
This too to be closed?

cjllanwarne:
Closed as duplicate

Issue #569, opened 2016-03-16
ValidationActor returns WorkflowDescriptor


mcovarr:

gauravs90:
Please hold on to this thought! I think the responsibility of that shouldn't lie with the ValidateActor!

gauravs90:
That;s multiple responsibilities within one single entity

geoffjentry:
I'd agree with that. I've always felt that the VA should only return a yes/no, although perhaps my issue is being overly pedantic with the name of the actor.

The argument for this though was that because people (including yourself, IIRC) didn't want to receive the simple yes/no and thus replicate the parsing of the input files the VA was handing back the parsed components. Once the validation is taken out of WorkflowDescriptor's apply() method it becomes a simple case class constructed by those same values so the effect is the same.

gauravs90:
This should be closed?

cjllanwarne:
Closed as duplicate

PR #582, opened 2016-03-17
Creating WorkflowDescriptor after validation in WorkflowManagerActor. Closes #544.

### LINKED ISSUE #544: Lazy creation of WorkflowDescriptor in WorkflowManagerActor only after Validation succeeds

### LINKED ISSUE #568: Submission goes to ValidationActor

gauravs90:
Solves #544. (Might also close #568 #569 #570 )
Intentions: 
1.) Make the erstwhile `ValidateActor` (now `MaterializeWorkflowDescriptorActor`) central to instantiating the `WorkflowDescriptor` which does so only after all validations succeed. In essence, workflow submissions now go via this route.
2.) Clean up validation code from `WorkflowDescriptor.scala` (All of that code comes to the above actor)
3.) WorkflowDescriptor should (ideally) be just a case class. Change relevant files where the constructors were being called (quite a lot ~3-liner changes)


gauravs90:
@mcovarr @geoffjentry  Thoughts / Ideas / Inputs please.

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -168,8 +171,10 @@ class WorkflowManagerActor(backend: Backend, config: Config)
 
   when (Running) {
     case Event(SubmitWorkflow(source), _) =>
-      val updatedData = submitWorkflow(source, replyTo = Option(sender), id = None)
-      stay() using updatedData
+      submitWorkflow(source, replyTo = Option(sender), id = None) onSuccess {
+        case updatedData: WorkflowManagerData => self ! UpdateWorkflowManagerState(updatedData)
+      }
+      stay()
```
mcovarr: This should not be passing data to set later.  There could be messages that come in before this one arrives that change the data in unrelated ways, but the updates from the processing of those messages would be overwritten.

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -207,11 +212,13 @@ class WorkflowManagerActor(backend: Backend, config: Config)
       data.workflows.get(id) foreach {_ ! SubscribeTransitionCallBack(sender())}
       stay()
     case Event(RestartWorkflows(w :: ws), _) =>
-      val updatedData = restartWorkflow(w)
+      val updatedData = restartWorkflow(w) onSuccess {
+        case updatedData: WorkflowManagerData => self ! UpdateWorkflowManagerState(updatedData)
+      }
```
mcovarr: same issue here

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -241,6 +248,9 @@ class WorkflowManagerActor(backend: Backend, config: Config)
   when (Done) { FSM.NullFunction }
 
   whenUnhandled {
+    case Event(UpdateWorkflowManagerState(data), _) =>
+      logger.info(s"Updating WorkflowManager state. New Data: $data")
+      stay() using data
```
mcovarr: this should go

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -887,4 +891,13 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
 
     runTransaction(action)
   }
+
+  private def materializeWorkflowDescriptor(id: WorkflowId, workflowSourceFiles: WorkflowSourceFiles): Future[WorkflowDescriptor] = {
+    import akka.pattern.ask
+    val materializedWfDescriptorFuture = (CromwellServer.validateActor ? MaterializeWorkflowDescriptorActor.MaterializeWorkflow(id, workflowSourceFiles)).mapTo[ValidationResult]
```
mcovarr: Ugh impedance mismatch between the synchronously invoked API and async messaging to the materialization actor :frowning:  At a minimum this shouldn't be using an `ask`, but I think we might have to live with the basic ugliness as long as we have a synchronous API talking to an actor.

mcovarr: The `PromiseActor` in `Main` is an example of how to do this without asks.

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -392,13 +396,15 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
   }
 
   override def getWorkflow(workflowExecutionId: Int)(implicit ec: ExecutionContext): Future[WorkflowDescriptor] = {
-    val action = for {
+    val action: dbio.DBIOAction[WorkflowDescriptor, NoStream, Read with Read] = for {
```
mcovarr: I'd remove the ascription as I don't think it really clarifies anything

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -392,13 +396,15 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
...
-      workflowDescriptor = WorkflowDescriptor(
-        WorkflowId(UUID.fromString(workflowExecutionResult.workflowExecutionUuid)),
-        WorkflowSourceFiles(workflowAux.wdlSource.toRawString, workflowAux.jsonInputs.toRawString, workflowAux.workflowOptions.toRawString)
-      )
+      workflowDescriptor <- materializeWorkflowDescriptor(WorkflowId(UUID.fromString(workflowExecutionResult.workflowExecutionUuid)),
+        WorkflowSourceFiles(workflowAux.wdlSource.toRawString, workflowAux.jsonInputs.toRawString, workflowAux.workflowOptions.toRawString))
+//      workflowDescriptor = WorkflowDescriptor(
+//        WorkflowId(UUID.fromString(workflowExecutionResult.workflowExecutionUuid)),
+//        WorkflowSourceFiles(workflowAux.wdlSource.toRawString, workflowAux.jsonInputs.toRawString, workflowAux.workflowOptions.toRawString)
+//      )
```
mcovarr: remove

engine/src/main/scala/cromwell/engine/workflow/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -0,0 +1,136 @@
...
+import wdl4s._
+
+import scala.language.postfixOps
+import scala.util.{Failure, Success, Try}
+import scalaz.Scalaz._
+
+object MaterializeWorkflowDescriptorActor {
+  def props(conf: Config = ConfigFactory.load): Props = Props(new MaterializeWorkflowDescriptorActor(conf))
+
+  sealed trait ValidateActorMessage
```
mcovarr: vestigial `Validate` terminology throughout

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -347,34 +357,29 @@ class WorkflowManagerActor(backend: Backend, config: Config)
...
-      data = stateData.add(workflowId -> actor)
-    } yield ActorAndStateData(actor, data)
-
-    actorAndData match {
-      case Failure(e) =>
-        val messageOrBlank = Option(e.getMessage).mkString
-        logger.error(e, s"$tag: Workflow failed submission: " + messageOrBlank)
-        replyTo foreach { _ ! WorkflowManagerSubmitFailure(e)}
+    import akka.pattern.ask
+    val materializedWfDescriptorFuture = (CromwellServer.validateActor ? MaterializeWorkflowDescriptorActor.MaterializeWorkflow(workflowId, source)).mapTo[ValidationResult]
```
mcovarr: same `ask` issue here as in the DB api

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
cjllanwarne: ?

cjllanwarne: Oh, because of your previous 2 comments! I see!

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
gauravs90: Yes, I'll revert this file back to it's originality and then modify it, once I know how to act upon it. Was just trying out a couple of things to make for the asynchrony of the WorkflowDescriptor

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
gauravs90: I can use a similar thing here..ask with no timeouts.. But I'm still unsure of how to deal with a `Future[WorkflowDescriptor]` instead of a `WorkflowDescriptor` currently being needed by stuff in this file :confused:

mcovarr: ah ok, let me take a look at that

mcovarr: `flatMap`ping the `Future` from the transaction seems to fix this part:

```scala
  override def getWorkflow(workflowExecutionId: Int)(implicit ec: ExecutionContext): Future[WorkflowDescriptor] = {
    case class ExecutionAndAux(execution: WorkflowExecution, aux: WorkflowExecutionAux)
    val action = for {
      workflowExecutionResult <- dataAccess.workflowExecutionsByPrimaryKey(workflowExecutionId).result.head
      workflowAux <- dataAccess.workflowExecutionAuxesByWorkflowExecutionUuid(workflowExecutionResult.workflowExecutionUuid).result.head
    } yield ExecutionAndAux(workflowExecutionResult, workflowAux)

    def futureDescriptorFromExecutionAndAux(executionAndAux: ExecutionAndAux): Future[WorkflowDescriptor] = {
      val sources = WorkflowSourceFiles(executionAndAux.aux.wdlSource.toRawString, executionAndAux.aux.jsonInputs.toRawString, executionAndAux.aux.workflowOptions.toRawString)
      materializeWorkflowDescriptor(id = WorkflowId(UUID.fromString(executionAndAux.execution.workflowExecutionUuid)), workflowSourceFiles = sources)
    }

    runTransaction(action) flatMap futureDescriptorFromExecutionAndAux
  }
```

gauravs90: Ahh Thanks! That seems to work!

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -887,4 +891,13 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
 
     runTransaction(action)
   }
+
+  private def materializeWorkflowDescriptor(id: WorkflowId, workflowSourceFiles: WorkflowSourceFiles): Future[WorkflowDescriptor] = {
```
geoffjentry: Why is this in the slick data access stuff?

geoffjentry: Oh wait, I get it now.

If it's coming from the DB shouldn't we know that the values are actually valid already?

engine/src/test/scala/cromwell/CromwellTestkitSpec.scala
```diff
@@ -110,11 +101,11 @@ object CromwellTestkitSpec {
   }
 
   /**
-   * Akka TestKit appears to be unable to match errors generated by `log.error(Throwable, String)` with the normal
-   * `EventFilter.error(...).intercept {...}` mechanism since `EventFilter.error` forces the use of a dummy exception
-   * that never matches a real exception.  This method works around that problem by building an `ErrorFilter` more
-   * explicitly to allow the caller to specify a `Throwable` class.
-   */
+    * Akka TestKit appears to be unable to match errors generated by `log.error(Throwable, String)` with the normal
```
gauravs90: This diff might have happened due to the cmd + alt + L. Will revert it :confused:

engine/src/test/scala/cromwell/CromwellTestkitSpec.scala
```diff
@@ -183,59 +200,34 @@ abstract class CromwellTestkitSpec extends TestKit(new CromwellTestkitSpec.TestW
...
-      output shouldEqual expectedOutput
+  def buildWorkflowDescriptor(sampleWdl: SampleWdl, uuid: UUID, materializeWorkflowDescriptorActor: ActorRef): WorkflowDescriptor = {
+    val workflowSources = WorkflowSourceFiles(sampleWdl.wdlSource(), sampleWdl.wdlJson, "{}")
+    val wfDescFut = PromiseActor.askNoWait(MaterializeWorkflowDescriptorActor.MaterializeWorkflow(WorkflowId(uuid), workflowSources), materializeWorkflowDescriptorActor)
+    val wfDesc = wfDescFut map {
+      case MaterializationSuccess(workflowDescriptor) => workflowDescriptor
+      case MaterializationFailure(error) => throw new Throwable(error)
+    }
+    Await.result(wfDesc, timeoutDuration)
+    //    WorkflowDescriptor(WorkflowId(uuid), workflowSources)
```
gauravs90: Oops! Needs to be removed!

engine/src/test/scala/cromwell/CromwellTestkitSpec.scala
```diff
@@ -316,6 +322,23 @@ abstract class CromwellTestkitSpec extends TestKit(new CromwellTestkitSpec.TestW
...
+                                         (implicit ec: ExecutionContext) = {
+    val actor = buildWorkflowManagerActor(sampleWdl, runtime, config)
+    val workflowSources = WorkflowSourceFiles(sampleWdl.wdlSource(runtime), sampleWdl.wdlJson, "{}")
+    runWdlWithWorkflowManagerActor(actor, workflowSources, eventFilter, stdout, stderr, Map.empty, terminalState)
+  }
+
+  private def buildWorkflowManagerActor(sampleWdl: SampleWdl, runtime: String, config: Config) = {
+    TestActorRef(new WorkflowManagerActor(new LocalBackend(system), config))
+  }
+
```
gauravs90: I don't know why these are being shown in the diff :confounded:

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
gauravs90: We are only appending the info to the state data (via messages, which this actor will guarantee handling in a serialized way), via `add` method. I don't think any overwriting is being performed   (or attempted currently). Or have I got it terribly wrong?

gauravs90:
\all:  I thought of rebasing it with the latest develop to bring in quite a few pending changes. I think I may have messed up as it shows a *lot* of files changes with all those commits. Anyway that I can do something to make this more legible? Or is it fine enough to follow it?

@scottfrazer and @kshakir were chosen by the wheel..but I think @cjllanwarne and @mcovarr might also wanna take a look (also since you guys already had given this a look).

engine/src/main/scala/cromwell/engine/WorkflowDescriptor.scala
```diff
@@ -364,8 +249,8 @@ object WorkflowDescriptor {
   private val writeDisabled = disabledMessage("Write to Cache", "WILL NOT be cached")
   private val readDisabled = disabledMessage("Read from Cache", "WILL ALL be executed")
 
-  private def configCallCaching(conf: Config) = lookupBooleanWithDefault(conf, "call-caching", "enabled", DefaultCallCachingValue)
-  private def lookupDockerHash(conf: Config) = lookupBooleanWithDefault(conf, "call-caching", "lookup-docker-hash", DefaultLookupDockerHash)
+  def configCallCaching(conf: Config) = lookupBooleanWithDefault(conf, "call-caching", "enabled", DefaultCallCachingValue)
+  def lookupDockerHash(conf: Config) = lookupBooleanWithDefault(conf, "call-caching", "lookup-docker-hash", DefaultLookupDockerHash)
```
gauravs90: I am thinking to move this methods to the `MaterializeWorkflowDescriptorActor` (where they are being actually called). Please let me know if anyone thinks otherwise.

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -391,58 +398,43 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
...
-        }
-      )
-
-    } yield workflowDescriptors
-
-    runTransaction(action)
+    for {
+      executionAndAuxes <- runTransaction(action)
+      workflowDescSeq <- Future.sequence(executionAndAuxes.map(exAux => futureDescriptorFromExecutionAndAux(exAux)))
+    } yield workflowDescSeq
```
gauravs90: I wish I understood Slick better, but is there a way to do this in a better way?

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -887,4 +879,14 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
...
     runTransaction(action)
   }
+
+  private def materializeWorkflowDescriptor(id: WorkflowId, workflowSourceFiles: WorkflowSourceFiles)(implicit ec: ExecutionContext): Future[WorkflowDescriptor] = {
+    implicit val actorSystem = CromwellServer.actorSystem
+    val materializedWfDescriptorFuture =
+      PromiseActor.askNoWait(MaterializeWorkflowDescriptorActor.MaterializeWorkflow(id, workflowSourceFiles), CromwellServer.workflowDescriptorMaterializeActor).mapTo[MaterializationResult]
+    materializedWfDescriptorFuture map {
+      case MaterializationSuccess(workflowDescriptor) => workflowDescriptor
+      case MaterializationFailure(error) => throw new Throwable(error)
```
gauravs90: Should just be `throw error`

cjllanwarne:
@gauravs90 yeah the rebase does seem to have gone slightly awry - if you look at https://github.com/broadinstitute/cromwell/pull/582/commits you can see all the bonus commits you're getting... :)

engine/src/main/scala/cromwell/engine/WorkflowDescriptor.scala
mcovarr: That certainly sounds reasonable.

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -391,58 +398,43 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
...
   }
 
   override def getWorkflowsByState(states: Traversable[WorkflowState])
                                   (implicit ec: ExecutionContext): Future[Traversable[WorkflowDescriptor]] = {
     val action = for {
       workflowExecutionResults <- dataAccess.workflowExecutionsByStatuses(states.map(_.toString)).result
+      workflowAuxSeq <- DBIO.sequence(workflowExecutionResults.map( workflowExecutionResult =>
+        dataAccess.workflowExecutionAuxesByWorkflowExecutionUuid(workflowExecutionResult.workflowExecutionUuid).result.head
+      ))
+      executionAndAux <- DBIO.sequence((workflowExecutionResults zip workflowAuxSeq) map { case (exec, aux) => DBIO.successful(ExecutionAndAux(exec, aux)) })
```
mcovarr: This line feels strange and unnecessary.  You have both the execution and the aux on line 429, they only need to be put into the `ExecutionAndAux`.

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -391,58 +398,43 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
...
-          }
-        }
-      )
-
-    } yield workflowDescriptors
-
-    runTransaction(action)
+    for {
+      executionAndAuxes <- runTransaction(action)
+      workflowDescSeq <- Future.sequence(executionAndAuxes.map(exAux => futureDescriptorFromExecutionAndAux(exAux)))
```
mcovarr: `Future.sequence(executionsAndAuxes map futureDescriptorFromExecutionAndAux)`

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
mcovarr: looks okay to me

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
mcovarr: Say the state at the time this code executes is `S`.  Call the workflow being submitted here `A`, so the `UpdateWorkflowManagerState` message is sent to self saying to set the state data to `S + A`.  

But at the time this code is executing, there may already a message to submit workflow `B` in the mailbox.  Because this message is already in the mailbox it will arrive before the `S + A` message.  So this code will fire off an `UpdateWorkflowManagerState` message saying to set the state to `S + B`.  

At this point it's a race to see which of those `UpdateWorkflowManagerState`s arrives first, but the last message to arrive will win and either `A` or `B` will be lost.

engine/src/main/scala/cromwell/engine/workflow/WorkflowMetadataBuilder.scala
```diff
@@ -172,5 +173,9 @@ object WorkflowMetadataBuilder {
       runtimeAttributes,
       executionEvents,
       failures)
+
+    val y = Await.result(x, Duration.Inf)
+    x
```
mcovarr: this looks like a mistake

engine/src/main/scala/cromwell/util/PromiseActor.scala
```diff
@@ -0,0 +1,35 @@
...
+
+  def props(promise: Promise[Any]): Props = Props(new PromiseActor(promise))
+}
+
+private class PromiseActor(promise: Promise[Any]) extends Actor {
+  override def receive = {
+    case Status.Failure(f) =>
+      promise.tryFailure(f)
+      context.stop(self)
+    case success: Any =>
```
mcovarr: That `Any` ascription seems unnecessary

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
mcovarr: nm I'm wrong, that `result.head` returns a `DBIO`.

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -391,58 +398,43 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
...
 
   override def getWorkflowsByState(states: Traversable[WorkflowState])
                                   (implicit ec: ExecutionContext): Future[Traversable[WorkflowDescriptor]] = {
     val action = for {
       workflowExecutionResults <- dataAccess.workflowExecutionsByStatuses(states.map(_.toString)).result
+      workflowAuxSeq <- DBIO.sequence(workflowExecutionResults.map( workflowExecutionResult =>
+        dataAccess.workflowExecutionAuxesByWorkflowExecutionUuid(workflowExecutionResult.workflowExecutionUuid).result.head
+      ))
+      executionAndAux <- DBIO.sequence((workflowExecutionResults zip workflowAuxSeq) map { case (exec, aux) => DBIO.successful(ExecutionAndAux(exec, aux)) })
+    } yield executionAndAux
```
mcovarr: But doing the zip/map in the yield would allow you to get rid of the `DBIO.sequence` and `DBIO.successful`.

```scala
    } yield (workflowExecutionResults zip workflowAuxSeq) map ExecutionAndAux.tupled
```

engine/src/main/scala/cromwell/engine/workflow/WorkflowMetadataBuilder.scala
gauravs90: Arghh I was debugging something..it crept it. I''ll remove it

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
geoffjentry: For those who have heard me complaining about managing transactions inside the same actor which is handling requests for those transactions, @mcovarr's example helps illustrate the issue - it's not impossible to make correct, but it's harder to reason about and makes it easy to create race conditions

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
gauravs90: This is so much cleaner! Thanks!

gauravs90: Well, tests seem to fail by replacing that line with this, although everything's the same.

gauravs90:
@scottfrazer @cjllanwarne @mcovarr @kshakir Cleaned this up. It now represents only my commits.

mcovarr:
um looks like it needs to be rebased... :frowning:

gauravs90:
Chris just merged his PR. That's why probably. Let's review this and straighten this out first? I can then rebase again before final merge maybe.

engine/src/main/scala/cromwell/util/PromiseActor.scala
gauravs90: Prepping up for scalastyle checks, which requires these kind of ascriptions to the catch-all kinda cases :confused:

mcovarr:
I'd prefer to do the rebase first, both to confirm this is the only problem and also to make sure your stuff is properly rebased on Chris's stuff.

gauravs90:
@mcovarr Done.

engine/src/main/scala/cromwell/engine/WorkflowDescriptor.scala
```diff
@@ -377,24 +245,4 @@ object WorkflowDescriptor {
...
-  private def workflowLogOptions(conf: Config): Option[WorkflowLogOptions] = {
-    for {
-      workflowConfig <- conf.getConfigOption("workflow-options")
-      dir <- workflowConfig.getStringOption("workflow-log-dir") if !dir.isEmpty
-      temporary <- workflowConfig.getBooleanOption("workflow-log-temporary") orElse Option(true)
-    } yield WorkflowLogOptions(Paths.get(dir), temporary)
-  }
 }
-
-case class WorkflowLogOptions(dir: Path, temporary: Boolean)
```
mcovarr: Really liking all this red on `WorkflowDescriptor`!  :smile:

engine/src/main/scala/cromwell/util/PromiseActor.scala
```diff
@@ -0,0 +1,35 @@
...
+object PromiseActor {
+  /**
+    * Sends a message to an actor and returns the future associated with the fullfilment of the reply
+    * Can be used instead of the akka `ask` semantics, without any timeout
+    * @param sendTo The actor that is wished to be asked something
+    * @param message The message to send to the actor
+    * @param actorSystem ActorSystem from which to create the actor from
+    * @return The future associated with the answer to the `ask`
+    */
+  def askNoWait(message: Any, sendTo: ActorRef)(implicit actorSystem: ActorSystem): Future[Any] = {
```
mcovarr: This name feels off, both because `ask` reflexively raises my hackles and because `NoWait` makes it sound like fire-and-forget which is actually the opposite of what it does.

engine/src/main/scala/cromwell/server/WorkflowManagerSystem.scala
```diff
@@ -23,5 +23,5 @@ trait WorkflowManagerSystem {
   CromwellBackend.initBackends(allowedBackends, defaultBackend, actorSystem)
   // For now there's only one WorkflowManagerActor so no need to dynamically name it
   lazy val workflowManagerActor = actorSystem.actorOf(WorkflowManagerActor.props(), "WorkflowManagerActor")
-  lazy val validateActor = actorSystem.actorOf(Props[ValidateActor], "ValidationActor")
+  lazy val workflowDescriptorMaterializeActor = actorSystem.actorOf(Props[MaterializeWorkflowDescriptorActor], "MaterializeWorkflowDescriptorActor")
```
mcovarr: Let's please not use this reflectiony form of `actorOf` if we can avoid it, as we saw last week this can break things in ways unit tests don't catch.

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
mcovarr: If the message is more event-sourcish in meaning and communicated "add A to whatever the state is at the time you receive this", this should work.

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
mcovarr: I'm seeing the same thing and I don't see why that should be the case, I'll keep looking into it.

engine/src/main/scala/cromwell/server/WorkflowManagerSystem.scala
gauravs90: Gotcha. Will change it to `xyz.props()`

engine/src/main/scala/cromwell/util/PromiseActor.scala
gauravs90: Suggestions?
I used 'ask' because it was inherently trying to capture akka's `?` semantics, while using no timeouts of any sort.

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
gauravs90: Working on it

engine/src/main/scala/cromwell/server/WorkflowManagerSystem.scala
gauravs90: @mcovarr Just curious, what was the exact problem?

mcovarr: #592 the actor's constructor was invoked without the required backend parameter

engine/src/main/scala/cromwell/util/PromiseActor.scala
mcovarr: ok, maybe `askNoTimeout` then

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -6,6 +6,8 @@ import java.util.{Date, UUID}
 import javax.sql.rowset.serial.SerialClob
 
 import _root_.slick.backend.DatabaseConfig
+import _root_.slick.dbio
+import _root_.slick.dbio.Effect.Read
 import _root_.slick.driver.JdbcProfile
```
scottfrazer: Can somebody explain what these `_root_` imports are?  I'm not sure I've ever seen something like this before

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -391,61 +399,43 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
     runTransaction(action)
   }
 
+  private def futureDescriptorFromExecutionAndAux(executionAndAux: ExecutionAndAux)(implicit ec: ExecutionContext): Future[WorkflowDescriptor] = {
```
scottfrazer: not that I *really* care though most methods in this file return futures but don't have 'future' in the name of the method... maybe `workflowDescriptorFromExecutionAndAux`?

geoffjentry: I'm always supportive of removing that 'maybe' and 'future' junk from identifier names

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
geoffjentry: http://stackoverflow.com/questions/687071/what-is-the-root-package-in-scala

That said, when I've had intellij add it for me in the past, it's always been unnecessary and due to some operator error combined w/ intellij

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
gauravs90: @mcovarr A test was annoying me. Please check the fixes related to this issue.

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
gauravs90: Well, it `validates` as a side effect now (hence we changed the name that day), mostly it creates a WorkflowDescriptor from sources.

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
gauravs90: Thanks! I removed that

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
gauravs90: Done

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -890,4 +880,14 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
 
     runTransaction(action)
   }
+
+  private def materializeWorkflowDescriptor(id: WorkflowId, workflowSourceFiles: WorkflowSourceFiles)(implicit ec: ExecutionContext): Future[WorkflowDescriptor] = {
+    implicit val actorSystem = CromwellServer.actorSystem
+    val materializedWfDescriptorFuture =
+      PromiseActor.askNoTimeout(MaterializeWorkflowDescriptorActor.MaterializeWorkflow(id, workflowSourceFiles), CromwellServer.workflowDescriptorMaterializeActor).mapTo[MaterializationResult]
```
kshakir: **TL;DR We should probably pass in the `actorSystem` and this `actor` as parameters.**

Long-reaching out to singleton `CromwellServer` from inside the database touches on longer standing issue.

NOTE: If he sees this, Jeff may be able to correct me, as I believe he's also spent time in this area.

There are two entrance points to run cromwell, either `Main.run()` or `Main. runServer()`. The way it's currently written, `Main.run()` creates its own `actorSystem`. It doesn't touch the `CromwellServer` singleton object, thus (supposedly) not static invoking the server's separate actor system.

Outside of `Main`, our test specs also repeatedly create their own specialty actor systems as well.

IMHO, the caller to this method should be allowed to pass in whatever actor system it likes.

Thus I vote for plumbing up the call tree, likely all the way to the super class `DataAccess`, method signature that also take in, along with the existing execution context, the system + actor. That way tests can pass in their test actor system, and the workflow manager actors can pass in whatever system they're currently running on.

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -348,34 +361,32 @@ class WorkflowManagerActor(config: Config)
...
-        logger.error(e, s"$tag: Workflow failed submission: " + messageOrBlank)
-        replyTo foreach { _ ! WorkflowManagerSubmitFailure(e)}
-        // Return the original state data if the workflow failed submission.
-        stateData
-      case Success(a) =>
-        a.actor ! (if (isRestart) Restart else Start(replyTo))
-        a.data
+    implicit val actorSystem = context.system
+    val materializedWfDescriptorFuture = PromiseActor.askNoTimeout(MaterializeWorkflowDescriptorActor.MaterializeWorkflow(workflowId, source, config),
+      CromwellServer.workflowDescriptorMaterializeActor).mapTo[MaterializationResult]
```
kshakir: Similar to comment in `SlickDataAccess`, don't instantiate the `CromwellServer` singleton. Plumb in an actor reference to `workflowDescriptorMaterializeActor` into the `WorkflowManagerActor` constructor.

kshakir: Alternatively, as actors create other actors, the `WorkflowManagerActor` could conceivably create its own `workflowDescriptorMaterializeActor` running in the same system.

engine/src/test/scala/cromwell/engine/WorkflowDescriptorSpec.scala
```diff
@@ -2,89 +2,58 @@ package cromwell.engine
...
-    }
+  def materializeWorkflowDescriptorFromSources(id: WorkflowId = WorkflowId.randomId(),
+                                               workflowSources: WorkflowSourceFiles,
+                                               conf: Config = ConfigFactory.load): WorkflowDescriptor = {
+    import scala.concurrent.ExecutionContext.Implicits.global
 
-    readCacheOnCombinations foreach {
-      case (config, writeToCache, readFromCache) => makeWorkflowDescriptor(config, writeToCache, readFromCache).readFromCache shouldBe true
+    val wfDescFut = PromiseActor.askNoTimeout(MaterializeWorkflowDescriptorActor.MaterializeWorkflow(id, workflowSources, conf),
+      workflowDescriptorMaterializerActor).mapTo[MaterializationResult]
```
kshakir: In tests, maybe we should timeout.

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -391,61 +399,43 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
...
 
-    runTransaction(action)
+    runTransaction(action) flatMap workflowDescriptorFromExecutionAndAux
   }
 
   override def getWorkflowsByState(states: Traversable[WorkflowState])
                                   (implicit ec: ExecutionContext): Future[Traversable[WorkflowDescriptor]] = {
     val action = for {
       workflowExecutionResults <- dataAccess.workflowExecutionsByStatuses(states.map(_.toString)).result
+      workflowAuxSeq <- DBIO.sequence(workflowExecutionResults.map( workflowExecutionResult =>
```
kshakir: Remove space in `.map( workflowExecutionResult`

engine/src/test/scala/cromwell/engine/workflow/SingleWorkflowRunnerActorSpec.scala
```diff
@@ -83,7 +83,8 @@ class SingleWorkflowRunnerActorWithMetadataSpec extends SingleWorkflowRunnerActo
     }
     TestKit.shutdownActorSystem(system, timeoutDuration)
 
-    val metadata = metadataFile.contentAsString.parseJson.convertTo[WorkflowMetadataResponse]
+    val metadata1 = metadataFile.contentAsString
+    val metadata =metadata1.parseJson.convertTo[WorkflowMetadataResponse]
```
kshakir: Add space in `=metadata1`

kshakir:
Generally looks much cleaner than before! I would prefer actor references passed-in instead of using the actor singletons though.

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -92,6 +98,8 @@ object SlickDataAccess {
   lazy val log = LoggerFactory.getLogger("slick")
 
   val FailureEventMaxMessageLength = 1024
+
+  private final case class ExecutionAndAux(execution: WorkflowExecution, aux: WorkflowExecutionAux)
```
cjllanwarne: Let's call this `WorkflowExecutionAndAux`. Unfortunately the term "Execution" is overloaded and in its unqualified state usually refers to `call`s

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
scottfrazer: Just to make sure I understand correctly, you want a signature like this?

```scala
private def materializeWorkflowDescriptor(
  id: WorkflowId,
  workflowSourceFiles: WorkflowSourceFiles,
  system: ActorSystem,
  actor: ActorRef)(implicit ec: ExecutionContext): Future[WorkflowDescriptor]
```

Then have callers (`getWorkflow`, `getWorkflowByState`) subsequently have the `ActorSystem` and `ActorRef` as parameters too?

geoffjentry: That global actorsystem was a major pita and was only put in as a temporary hack leading up to PBE (months ago in the halcyon days when everyone assumed this would be a quick thing).

While I'm not a huge fan of the akka style implicit EC params, it might be a good use case here for the actor system.

engine/src/main/scala/cromwell/engine/workflow/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -0,0 +1,165 @@
...
+import cromwell.util.TryUtil
+import spray.json._
+import wdl4s._
+
+import scala.language.postfixOps
+import scala.util.{Failure, Success, Try}
+import scalaz.Scalaz._
+import scalaz.Validation.FlatMap._
+
+object MaterializeWorkflowDescriptorActor {
```
cjllanwarne: I'm not sure I like `Materialize`. Or verbs as class names. `WorkflowDescriptorActor` is probably enough?

engine/src/main/scala/cromwell/engine/workflow/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -0,0 +1,165 @@
...
+      } flatMap { case (_, rawInputs) =>
+        validateCoercedInputs(rawInputs, namespace) flatMap { coercedInputs =>
+          val workflowRootPath = backend.buildWorkflowRootPath(backend.rootPath(workflowOptions), namespace.workflow.unqualifiedName, id)
+          val wfContext = new WorkflowContext(workflowRootPath)
+          val fileSystems = backend.fileSystems(workflowOptions)
+          val engineFunctions = backend.engineFunctions(fileSystems, wfContext)
+          validateDeclarations(namespace, workflowOptions, coercedInputs, engineFunctions) flatMap { declarations =>
+            WorkflowDescriptor(id, sourceFiles, workflowOptions, workflowLogOptions(conf), rawInputs, namespace, coercedInputs, declarations, backend,
+              configCallCaching(conf), lookupDockerHash(conf), wfContext, fileSystems).successNel
+          }
```
cjllanwarne: This is an almighty triple-nesting of long anonymous functions. Split them out into `def`s and give them names

engine/src/main/scala/cromwell/engine/workflow/MaterializeWorkflowDescriptorActor.scala
scottfrazer: If actors are used kind of like functions (especially in this case), then why not name them like functions too?

I don't really object to either naming scheme, but I think If I had to choose, I'd probably have made the same decision as Guarav

geoffjentry: @scottfrazer raises a good point. Ideally actors should be thought of more as functions manipulating data as part of a chain of events and less as objects w/ state (but obviously they can do either). This is more the former.

engine/src/main/scala/cromwell/engine/workflow/MaterializeWorkflowDescriptorActor.scala
scottfrazer: Yeah I had a bit of a hard time figuring this out too

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -890,4 +880,14 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
...
+
+  private def materializeWorkflowDescriptor(id: WorkflowId, workflowSourceFiles: WorkflowSourceFiles)(implicit ec: ExecutionContext): Future[WorkflowDescriptor] = {
+    implicit val actorSystem = CromwellServer.actorSystem
+    val materializedWfDescriptorFuture =
+      PromiseActor.askNoTimeout(MaterializeWorkflowDescriptorActor.MaterializeWorkflow(id, workflowSourceFiles), CromwellServer.workflowDescriptorMaterializeActor).mapTo[MaterializationResult]
+    materializedWfDescriptorFuture map {
+      case MaterializationSuccess(workflowDescriptor) => workflowDescriptor
+      case MaterializationFailure(error) => throw error
+    }
+  }
```
mcovarr: Can this be shortened up a bit?

```scala
    import PromiseActor.EnhancedActorRef
    val message = MaterializeWorkflowDescriptorActor.MaterializeWorkflow(id, workflowSourceFiles)
    val materializationActor = CromwellServer.workflowDescriptorMaterializeActor
    
    materializationActor.askNoTimeout(message) map {
      case MaterializationSuccess(workflowDescriptor) => workflowDescriptor
      case MaterializationFailure(error) => throw error
    }
```

In `PromiseActor`:

```scala
  implicit class EnhancedActorRef(val actorRef: ActorRef) extends AnyVal {
    def askNoTimeout(message: Any)(implicit ec: ExecutionContext): Future[Any] = {
      PromiseActor.askNoTimeout(message, actorRef)
    }
  }
```

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -43,6 +48,7 @@ object WorkflowManagerActor {
   final case class WorkflowMetadata(id: WorkflowId) extends WorkflowManagerActorMessage
   final case class RestartWorkflows(workflows: Seq[WorkflowDescriptor]) extends WorkflowManagerActorMessage
   final case class CallCaching(id: WorkflowId, parameters: QueryParameters, call: Option[String]) extends WorkflowManagerActorMessage
+  private final case class UpdateWorkflowManagerState(entry: (WorkflowId, WorkflowActorRef)) extends WorkflowManagerActorMessage
```
mcovarr: This says "Update" but it really means "AddEntry"

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -168,8 +172,11 @@ class WorkflowManagerActor(config: Config)
 
   when (Running) {
     case Event(SubmitWorkflow(source), _) =>
-      val updatedData = submitWorkflow(source, replyTo = Option(sender), id = None)
-      stay() using updatedData
+      submitWorkflow(source, replyTo = Option(sender), id = None) onSuccess {
```
mcovarr: `onSuccess` is going to be deprecated in the next Scala release, use `map` instead

gauravs90: Done.

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
gauravs90: Done.

engine/src/main/scala/cromwell/engine/workflow/MaterializeWorkflowDescriptorActor.scala
cjllanwarne: @geoffjentry But they still **aren't** like function calls in that case. They're more like robotic arms on a production line. You wouldn't name a robotic arm in a car production line the `AddWheel`. You'd name it the `WheelAdder`

engine/src/test/scala/cromwell/engine/workflow/SingleWorkflowRunnerActorSpec.scala
gauravs90: Done.

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
gauravs90: Done.

engine/src/test/scala/cromwell/engine/WorkflowDescriptorSpec.scala
gauravs90: It actually is referring to `CromwellServer.timeout` (30 seconds I believe)

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -890,4 +880,14 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
 
     runTransaction(action)
   }
+
+  private def materializeWorkflowDescriptor(id: WorkflowId, workflowSourceFiles: WorkflowSourceFiles)(implicit ec: ExecutionContext): Future[WorkflowDescriptor] = {
```
Horneth: It looks strange to me haveing a method doing asynchronous calls to an actor inside of the DB layer to return an engine-level object type.
Is it not possible to change `getWorkflow` to `getExecutionAndAux` and have whoever is currently calling `getWorkflow` call `materializeWorkflowDescriptor` instead ?

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -207,11 +214,13 @@ class WorkflowManagerActor(config: Config)
       data.workflows.get(id) foreach {_ ! SubscribeTransitionCallBack(sender())}
       stay()
     case Event(RestartWorkflows(w :: ws), _) =>
-      val updatedData = restartWorkflow(w)
+      val updatedData = restartWorkflow(w) onSuccess {
```
mcovarr: another one here, all should be replaced

engine/src/main/scala/cromwell/engine/workflow/MaterializeWorkflowDescriptorActor.scala
Horneth: `WorkflowDescriptorActorMaterializer` it is

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
mcovarr: The `val` here is not used.

`restartWorkflow` used to return the full state data, now it just returns a new entry to be added.  The name of the variables in the `case` should be updated accordingly, they're currently misleading/wrong.

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -348,34 +361,32 @@ class WorkflowManagerActor(config: Config)
   /** Submit the workflow and return an updated copy of the state data reflecting the addition of a
     * Workflow ID -> WorkflowActorRef entry.
     */
-  private def submitWorkflow(source: WorkflowSourceFiles, replyTo: Option[ActorRef], id: Option[WorkflowId]): WorkflowManagerData = {
+  private def submitWorkflow(source: WorkflowSourceFiles, replyTo: Option[ActorRef], id: Option[WorkflowId]): Future[(WorkflowId, ActorRef)] = {
```
mcovarr: This should use a type rather than this `(WorkflowId, ActorRef)` tuple

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -168,8 +172,11 @@ class WorkflowManagerActor(config: Config)
 
   when (Running) {
     case Event(SubmitWorkflow(source), _) =>
-      val updatedData = submitWorkflow(source, replyTo = Option(sender), id = None)
-      stay() using updatedData
+      submitWorkflow(source, replyTo = Option(sender), id = None) onSuccess {
+        case updatedEntry: (WorkflowId, ActorRef) => self ! UpdateWorkflowManagerState(updatedEntry)
+      }
+      //`onFailure` doesn't do anything since it's been handled already in the `submitWorkflow` function
```
Horneth: I don't think this comment is needed

engine/src/test/scala/cromwell/engine/WorkflowDescriptorSpec.scala
kshakir: `PromiseActor.askNoTimeout` doesn't timeout. You can switch to a regular ask here instead.

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -348,34 +361,32 @@ class WorkflowManagerActor(config: Config)
...
-        a.data
+    implicit val actorSystem = context.system
+    val materializedWfDescriptorFuture = PromiseActor.askNoTimeout(MaterializeWorkflowDescriptorActor.MaterializeWorkflow(workflowId, source, config),
+      CromwellServer.workflowDescriptorMaterializeActor).mapTo[MaterializationResult]
+    materializedWfDescriptorFuture map {
+      case MaterializationSuccess(descriptor) =>
+        val wfActor = context.actorOf(WorkflowActor.props(descriptor), s"WorkflowActor-$workflowId")
+        wfActor ! SubscribeTransitionCallBack(self)
+        wfActor ! (if (isRestart) Restart else Start(replyTo))
+        logger.debug(s"Successfuly stared ${wfActor.path} for Workflow ${workflowId}")
```
Horneth: star**t**ed

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
kshakir: _NOTE: After chatting with a couple other cromwellians, there was a general consensus of "oh, that's what you meant. I didn't get that from your comments". If you want to tech talk what I was referring to, happy to do so._

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
gauravs90: @kshakir In general, I like that idea and that did pass my mind, but for the limits of this PR I refrained to change too much there. Do we want the trait `DataAccess` to define an actorSystem, that the concrete implementations must override (in this In `SlickDataAcess` we override it and assign `CromwellServer.actorSystem` ? )

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
gauravs90: Is this implicit class necessary? I'm not a huge fan of implicit classes unless it absolutely makes things better and cleaner. I don't see many advantages of this here currently :confused:

geoffjentry: I find that enhanced classes almost always make things more readable as saying something like `foo.someTransform.anotherTransform` fits the more modern functional dataflow-y style than `anotherTransform(someTransform(foo))

engine/src/main/scala/cromwell/engine/workflow/MaterializeWorkflowDescriptorActor.scala
gauravs90: I'm sorry but what is the final conclusion here? @Horneth 's final name contradicts the way that we decided to name Actors (by appending the term 'Actor' to the classes)

gauravs90: I prefer the name that we have currently (given the options here)

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
gauravs90: Done

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
gauravs90: Done, as per general consensus.

engine/src/main/scala/cromwell/engine/workflow/MaterializeWorkflowDescriptorActor.scala
Horneth: Sorry I was more trolling than anything else -_-
I personally don't have a strong opinion on this nor against the current name.

cjllanwarne: My real complaint is that `MaterializeWorkflowDescriptorActor` is super clunky. 

So I thought, how about `GetWorkflowDescriptorActor`. But in scala it's frowned upon to have `get` in getter names. Hence, my suggestion of `WorkflowDescriptorActor`...

cjllanwarne: OTOH, if everyone else is happy I don't want to hold up your PR :) :ok_hand:

cjllanwarne:
I think the wheel assignment was lost a little ways back but this looks almost ready for thumbs now. I'll restate the will of the wheel: @scottfrazer and @kshakir

engine/src/test/scala/cromwell/engine/WorkflowDescriptorSpec.scala
kshakir: Discussed and no change needed. @gauravs90 pointed out that the `Await` will timeout just below.

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
gauravs90: @Horneth @kshakir  Was just looking at the scope of this, it looks like `getWorkflow` is being called by WorkflowActor (which currently **does** have the reference to `WorkflowDescriptorMaterializerActor` but **wouldn't** once it is moved from WorkflowManagerSystem to as-a-child-of WorklowManagerActor). Do we want the reference for that actor to be passed in in the constructor to the WorkflowActor ?

Horneth: Hmm I'm not sure I understand, if `WorkflowDescriptorMaterializerActor` is meant to be a singleton throughout Cromwell, then accessing it shouldn't be a problem. If it's not supposed to be a singleton and anyone can pop up its own, then it shouldn't be a problem either to create one from the WorkflowActor.
Or am I missing something ?

geoffjentry: I apologize as I haven't read anything but @Horneth 's last comment so I might be way off base or talking about something different altogether.

The idea is that there's one of these, IMO it should hang directly off of the system's kernel. in the future it might be a simple router w/ stuff behind it but that address would be a singular address that anything needing validation would send stuff to

gauravs90: @geoffjentry That's how it is currently, but then going back to @kshakir 's comments that we should not be using `CromwellServer`'s reference directly to get that actor, (and instead plumbing it's reference down to the actors needing it), how do you see, for instance, the `WorkflowActor` getting hold of the `MaterialieWorkflowDescriptorActor` reference to materialize a workflow from sources?

geoffjentry: The WMA shoudl be passed the actorref as teh WMA is also a child of the kernel

gauravs90: Just to be on the same page:
```scala
class WorkflowManagerActor(config: Config, materializeWfActor: ActorRef) {
...

// WMA created WorkflowActor, and say WorkflowActor also needs the `materializeWfActor` 
val wfActor = WorkflowActor.props(wfDescriptor, materializeWfActor)

//where the constructor of WorkflowActor is changed to also receive a reference to materializeWfActor

}
```

And for other singletons needing to materialize the WorkflowDescriptor (like `WorkflowMetadataBuilder`), does it still use `CromwellServer.materializeWfActor` ? Or is there a way to inject the actor ref to other Scala objects?

Horneth: @geoffjentry I think @gauravs90 is saying that the WA is also going to need a reference to the `MaterialieWorkflowDescriptorActor`. Does that mean we've decided that the WA will be responsible for asking `MaterialieWorkflowDescriptorActor` to create its own WorkflowDescriptor ?

gauravs90: @Horneth Yes, with the change for your comment (the first of this thread i.e.), WA will need the MaterializeActor

gauravs90: @Horneth Fixed w.r.t your comment, based on the existing design (i.e. using CromwellServer.MaterializeActor...`., haven't injected it's references down the chain yet until we agree on how to do that.

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -207,11 +215,13 @@ class WorkflowManagerActor(config: Config)
       data.workflows.get(id) foreach {_ ! SubscribeTransitionCallBack(sender())}
       stay()
     case Event(RestartWorkflows(w :: ws), _) =>
-      val updatedData = restartWorkflow(w)
+      val updatedData = restartWorkflow(w) map {
+        case updatedData: WorkflowIdToActorRef => self ! AddEntryToWorkflowManagerData(updatedData)
+      }
```
mcovarr: unused `val` and misleading variable name

gauravs90: Yukk. Fixed.

engine/src/main/scala/cromwell/engine/package.scala
```diff
@@ -70,4 +75,19 @@ package object engine {
     def toKey: ExecutionDatabaseKey = ExecutionDatabaseKey(execution.callFqn, execution.index.toIndex, execution.attempt)
     def executionStatus: ExecutionStatus = ExecutionStatus.withName(execution.status)
   }
+
+  //TODO: The `CromwellServer.workflowDescriptorMaterializerActor` referenced below should be injected rather than being used like this
```
cjllanwarne: Although I might favour the former, workflowDescriptorMaterializerActor is actually MaterializeWorkflowDescriptorActor

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
Horneth: Awesome, and I agree yes we need to decide on the way we want to access the MaterializeActor

gauravs90: @geoffjentry @mcovarr I hope you are already aware of this, but after talking to @kshakir we came to a conclusion that this actor will no longer be singleton. So it will be created (and stopped) by whoever requests some service from it. Basically it's going back to what it was in the earlier stages of this PR (of not having this singleton). This is a little bigger change now that that there are multiple places where this needs to be modified, so just thought that you should know, and raise a flag *right now* if there's anything.

engine/src/main/scala/cromwell/engine/package.scala
gauravs90: Yep..that todo should go soon :smile:

engine/src/main/scala/cromwell/engine/workflow/WorkflowMetadataBuilder.scala
```diff
@@ -160,19 +160,8 @@ object WorkflowMetadataBuilder {
...
-      workflowDescriptor,
-      callToStatusMap,
-      callInputs,
-      callOutputs,
-      infosByExecution,
-      runtimeAttributes,
-      executionEvents,
-      failures)
+      wfMetadata <- buildWorkflowMetadata(id, workflowExecution, workflowOutputs, workflowExecutionAux,
+        callToStatusMap, callInputs, callOutputs, infosByExecution, runtimeAttributes, executionEvents,failures)
```
kshakir: Need a space in `executionEvents,failures`

kshakir:
One space needed. After review from others, plus rebase+squash, then :+1: from me

engine/src/main/scala/cromwell/engine/workflow/WorkflowMetadataBuilder.scala
gauravs90: Done.

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
geoffjentry: Hi @gauravs90 - sorry for the delay, I was offsite all afternoon at a meeting. 

I chatted with @kshakir and it sounds like this is being done simply as a temporary measure because at the moment the widgets can't all be aligned properly and not viewed as the correct architecture. I'm fine with that, however longer term I still believe that the correct solution is to have a dedicated location in the actor system to provide validation services. 

I don't think we need to continue to block all of the work depending on this PR however and we can make a ticket to rearrange it down the road.

scottfrazer:
:+1:

mcovarr:
Hi @gauravs90 this should be mergeable after a rebase.  We'd like to get this in soonish as it's blocking some other tickets we're working on.  Thanks!

gauravs90:
@mcovarr Yep..It'll be bigger rebase I suppose but I'm hoping to get it rebased tonight. Will let you know once it's ready

gauravs90:
@mcovarr Just rebased with the latest develop. Tests pass locally, and waiting for the results on Travis :sweat_smile: 
With the time gap b/w us, and given that I'll be a little later here at work tomorrow w.r.t you guys, please feel free to merge this to develop so that the blockers because of this PR can be continued elsewhere (given the tests pass here of course and everything looks good). Else, I can come in and do that if there's any corrections required. Thanks!

mcovarr:
assembly fails consistently in Travis with "Connection reset" errors during merge, but it seems to work fine locally.  I'll give it a few more chances to straighten itself out in Travis before merging.

mcovarr:
Reproduced the assembly failure in Travis on a different branch that was green last night.  i.e., the problem isn't here, so I'm merging this.

PR #593, opened 2016-03-18
Added "fail faster" support. Closes #587

### LINKED ISSUE #587: Support 'Fail Fast'
In cases where call caching is not being used (as is the case for Genomes On the Cloud for the time being), the ability to have workflows "fail fast" is desired.  Fail fast in this case mean, once a task has failed no new tasks should be started (currently running tasks can finish) and then the workflow should be set to failed.

This is important because when running a workflow, and you can't reuse the non-failed work, you want to know the workflow will fail as soon as possible.  With a long running workflow, an early failure with few downstream dependencies you have to wait a long time (24 hours for GOTC) to realize it's actually failed.

This should be specifyable as a workflow option, e.g. failure-strategy = fast where the current behavior is 'slow'.  If it's easy to also have this as a server config setting for the default that would be a nice to have but not critical.

cjllanwarne:
Suggested comments for reviewers:
- [x] Needs a test
- [x] Needs docs
- [x] Add to changelog
- [x] How about a proper type for this?
- [x] Best code I've ever seen
- [x] The test is failing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 ## 0.19
 
 * New runtime option for JES: `bootDiskSizeGb`. Allows specification of a boot disk size (as an Integer number of GB) that can be increased to boot a larger docker image.
+* Workflow options now allows you to specify a `workflowFailureMode` to control workflow behavior after a call has failed, for example:`{ "workflowFailureMode": "..." }`. The options are:
+  * `ContinueWhilePossible` - continues to start and process calls in the workflow, as long as they did not depend on the failing call
+  * `NoNewCalls` - no *new* calls are started but existing calls are allowed to finish
+  * The default is `NoNewCalls` but this can be changed using the `workflow-options.workflow-failure-mode` configuration option.
 * Bug fix: Tasks that changed directory would fail on JES because their return code file was written to the new directory instead of an absolute path
 * Bug fix: Using `write_*` functions in a Task's command (e.g. `./my_script --file=${write_file(my_array)}`) will now work with JES
 * Changing format of the 'disks' runtime attribute slightly to allow for mounting disks at specific mountpoints
--- a/README.md
+++ b/README.md
@@ -986,18 +986,24 @@ Example workflow options file:
 
 Valid keys and their meanings:
 
-* **write_to_cache** - Accepts values `true` or `false`.  If `false`, the completed calls from this workflow will not be added to the cache.  See the [Call Caching](#call-caching) section for more details.
-* **read_from_cache** - Accepts values `true` or `false`.  If `false`, Cromwell will not search the cache when invoking a call (i.e. every call will be executed unconditionally).  See the [Call Caching](#call-caching) section for more details.
-* **workflow_log_dir** - Specifies a path where per-workflow logs will be written.  If this is not specified, per-workflow logs will not be copied out of the Cromwell workflow log temporary directory/path before they are deleted.
-* **outputs_path** - Specifies a path where final workflow outputs will be written.  If this is not specified, workflow outputs will not be copied out of the Cromwell workflow execution directory/path.
-* **call_logs_dir** - Specifies a path where final call logs will be written.  If this is not specified, call logs will not be copied out of the Cromwell workflow execution directory/path.
-* **jes_gcs_root** - (JES backend only) Specifies where outputs of the workflow will be written.  Expects this to be a GCS URL (e.g. `gs://my-bucket/workflows`).  If this is not set, this defaults to the value within `backend.jes.baseExecutionBucket` in the [configuration](#configuring-cromwell).
-* **google_project** - (JES backend only) Specifies which google project to execute this workflow.
-* **refresh_token** - (JES backend only) Only used if `localizeWithRefreshToken` is specified in the [configuration file](#configuring-cromwell).  See the [Data Localization](#data-localization) section below for more details.
-* **auth_bucket** - (JES backend only) defaults to the the value in **jes_gcs_root**.  This should represent a GCS URL that only Cromwell can write to.  The Cromwell account is determined by the `google.authScheme` (and the corresponding `google.userAuth` and `google.serviceAuth`)
-* **monitoring_script** - (JES backend only) Specifies a GCS URL to a script that will be invoked prior to the WDL command being run.  For example, if the value for monitoring_script is "gs://bucket/script.sh", it will be invoked as `./script.sh > monitoring.log &`.  The value `monitoring.log` file will be automatically de-localized.
-* **preemptible** - (JES backend only) Specifies the maximum number of times a call should be executed with a preemptible VM. This option can be overridden by [runtime attributes](#preemptible). By default the value is 0, which means no Preemptible VM will be used.
-* **defaultRuntimeOptions** - A JSON object where the keys are [runtime attributes](#runtime-attributes) and the values are defaults that will be used through the workflow invocation.  Individual tasks can choose to override these values.  See the [runtime attributes](#specifying-default-values) section for more information.
+* Global *(use with any backend)*
+    * **write_to_cache** - Accepts values `true` or `false`.  If `false`, the completed calls from this workflow will not be added to the cache.  See the [Call Caching](#call-caching) section for more details.
+    * **read_from_cache** - Accepts values `true` or `false`.  If `false`, Cromwell will not search the cache when invoking a call (i.e. every call will be executed unconditionally).  See the [Call Caching](#call-caching) section for more details.
+    * **workflow_log_dir** - Specifies a path where per-workflow logs will be written.  If this is not specified, per-workflow logs will not be copied out of the Cromwell workflow log temporary directory/path before they are deleted.
+    * **outputs_path** - Specifies a path where final workflow outputs will be written.  If this is not specified, workflow outputs will not be copied out of the Cromwell workflow execution directory/path.
+    * **call_logs_dir** - Specifies a path where final call logs will be written.  If this is not specified, call logs will not be copied out of the Cromwell workflow execution directory/path.
+    * **defaultRuntimeOptions** - A JSON object where the keys are [runtime attributes](#runtime-attributes) and the values are defaults that will be used through the workflow invocation.  Individual tasks can choose to override these values.  See the [runtime attributes](#specifying-default-values) section for more information.
+    * **workflowFailureMode** - What happens after a task fails. Choose from:
+        * **ContinueWhilePossible** - continues to start and process calls in the workflow, as long as they did not depend on the failing call
+        * **NoNewCalls** - no *new* calls are started but existing calls are allowed to finish
+        * The default is `NoNewCalls` but this can be changed using the `workflow-options.workflow-failure-mode` configuration option.
+* JES Backend Only
+    * **jes_gcs_root** - (JES backend only) Specifies where outputs of the workflow will be written.  Expects this to be a GCS URL (e.g. `gs://my-bucket/workflows`).  If this is not set, this defaults to the value within `backend.jes.baseExecutionBucket` in the [configuration](#configuring-cromwell).
+    * **google_project** - (JES backend only) Specifies which google project to execute this workflow.
+    * **refresh_token** - (JES backend only) Only used if `localizeWithRefreshToken` is specified in the [configuration file](#configuring-cromwell).  See the [Data Localization](#data-localization) section below for more details.
+    * **auth_bucket** - (JES backend only) defaults to the the value in **jes_gcs_root**.  This should represent a GCS URL that only Cromwell can write to.  The Cromwell account is determined by the `google.authScheme` (and the corresponding `google.userAuth` and `google.serviceAuth`)
+    * **monitoring_script** - (JES backend only) Specifies a GCS URL to a script that will be invoked prior to the WDL command being run.  For example, if the value for monitoring_script is "gs://bucket/script.sh", it will be invoked as `./script.sh > monitoring.log &`.  The value `monitoring.log` file will be automatically de-localized.
+    * **preemptible** - (JES backend only) Specifies the maximum number of times a call should be executed with a preemptible VM. This option can be overridden by [runtime attributes](#preemptible). By default the value is 0, which means no Preemptible VM will be used.
 
 # Call Caching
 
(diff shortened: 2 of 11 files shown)


engine/src/main/scala/cromwell/engine/WorkflowDescriptor.scala
```diff
@@ -40,6 +40,7 @@ case class WorkflowDescriptor(id: WorkflowId,
                               backend: Backend,
                               configCallCaching: Boolean,
                               lookupDockerHash: Boolean,
+                              configWorkflowFailureMode: Boolean,
```
mcovarr: How about a proper type for this?  Not just because types are teh awesome but also as your title suggests it's not a stretch to imagine a "fail fastest" that aborts whatever's still running.

cjllanwarne: This wasn't on the approved comments list!

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -825,10 +826,16 @@ case class WorkflowActor(workflow: WorkflowDescriptor)
       case _ => false
     }
 
+    lazy val hasAnythingFailed = executionStore exists {
```
mcovarr: `executionStore.values exists { x => x == Failed || x == Aborted }`

mcovarr:
Consider all those boxes checked.

engine/src/main/scala/cromwell/engine/WorkflowDescriptor.scala
geoffjentry: @mcovarr knows how to make me :smile:

cjllanwarne: OTOH, agile is all about incremental changes and not over-engineering solutions for a "maybe, possibly..." future scenario :)

mcovarr:
The one Travis failure in "push" occurs in the "fail slow" test :frowning:

cjllanwarne:
@mcovarr all suggested comments addressed. Ready for round 2 now

cjllanwarne:
The wheel also chooses @Horneth

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## 0.19
 
+* Workflow options now allows you to specify a `workflowFailureMode` to control workflow behaviour after a call has failed, for example:`{ "workflowFailureMode": "..." }`. The options are:
+  * `ContinueWhilePossible` - continues to start and process calls in the workflow, as long as they did not depend on the failing call
+  * `NoNewCalls` - no *new* calls are started but existing calls are allowed to finish
```
mcovarr: nice!

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## 0.19
 
+* Workflow options now allows you to specify a `workflowFailureMode` to control workflow behaviour after a call has failed, for example:`{ "workflowFailureMode": "..." }`. The options are:
```
mcovarr: behaviour -> behavior

CHANGELOG.md
mcovarr: What's the default behavior?

engine/src/test/scala/cromwell/util/SampleWdl.scala
```diff
@@ -2317,7 +2317,7 @@ task shouldCompleteFast {
         |    Int a
         |    command {
         |        echo "The number was: ${a}"
-        |        sleep 0.2
+        |        sleep 0.15
```
mcovarr: why these changes?

engine/src/main/scala/cromwell/engine/WorkflowDescriptor.scala
```diff
@@ -61,6 +62,7 @@ case class WorkflowDescriptor(id: WorkflowId,
     log.warn(s"$CallLogsDirOptionKey expected to be of type String", e)
     throw e
   }
+  val workflowFailureMode: WorkflowFailureMode = workflowOptions.get(WorkflowFailureModeKey) flatMap {WorkflowFailureMode.tryParse} getOrElse configWorkflowFailureMode
```
mcovarr: It looks as if `WorkflowFailureMode` is specified incorrectly this would silently ignore that and just go with what's in the config?

engine/src/test/scala/cromwell/WorkflowFailSlowSpec.scala
```diff
@@ -5,13 +5,39 @@ import cromwell.engine.WorkflowFailed
...
+      |}
+    """.stripMargin
+
+  val FailFastOptions =
+    """
+      |{
+      |  "workflowFailureMode": "NoNewCalls"
+      |}
+    """.stripMargin
+
```
mcovarr: add a test for a bad value

mcovarr: and a test for no value

README.md
```diff
@@ -975,18 +975,23 @@ Example workflow options file:
...
+    * **outputs_path** - Specifies a path where final workflow outputs will be written.  If this is not specified, workflow outputs will not be copied out of the Cromwell workflow execution directory/path.
+    * **call_logs_dir** - Specifies a path where final call logs will be written.  If this is not specified, call logs will not be copied out of the Cromwell workflow execution directory/path.
+    * **defaultRuntimeOptions** - A JSON object where the keys are [runtime attributes](#runtime-attributes) and the values are defaults that will be used through the workflow invocation.  Individual tasks can choose to override these values.  See the [runtime attributes](#specifying-default-values) section for more information.
+    * **workflowFailureMode** - What happens after a task fails. Choose from:
+        * **ContinueWhilePossible** - continues to start and process calls in the workflow, as long as they did not depend on the failing call
+        * **NoNewCalls** - no *new* calls are started but existing calls are allowed to finish
```
mcovarr: doc default behavior here too

mcovarr: shouldn't there be a bit about the config setting in here somewhere?

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -825,10 +826,14 @@ case class WorkflowActor(workflow: WorkflowDescriptor)
       case _ => false
     }
 
+    lazy val hasAnythingFailed = executionStore.values exists { x => x == Failed || x == Aborted }
+
+    lazy val workflowFailFastTrigger = workflowFailureMode == NoNewCalls && hasAnythingFailed
```
mcovarr: it feels like most of this logic should live in the `WorkflowFailureMode` subtypes

engine/src/test/scala/cromwell/util/SampleWdl.scala
cjllanwarne: Turns out Fail Fast-ish lets there be a race condition here, if the failing and successful tasks take the same time, sometimes the successful task's successor has already started by the time the failure is noticed. So this increases (from 0) the gap between the preceding failure and success.

mcovarr: Can you please leave a comment to this effect?

README.md
```diff
@@ -975,18 +975,23 @@ Example workflow options file:
...
+* JES Backend Only
+    * **jes_gcs_root** - (JES backend only) Specifies where outputs of the workflow will be written.  Expects this to be a GCS URL (e.g. `gs://my-bucket/workflows`).  If this is not set, this defaults to the value within `backend.jes.baseExecutionBucket` in the [configuration](#configuring-cromwell).
```
Horneth: Totally off PR but I'm thinking about it as I see it now. The "root" concept is actually also totally valid for local backend and this option should probably be generalized.

engine/src/main/scala/cromwell/engine/WorkflowDescriptor.scala
```diff
@@ -299,8 +302,8 @@ object WorkflowDescriptor {
     val validatedDescriptor = for {
       c <- validateCoercedInputs(id, rawInputs, namespace).disjunction
       d <- validateDeclarations(id, namespace, options, c, engineFunctions).disjunction
-    } yield WorkflowDescriptor(id, sourceFiles, options, workflowLogOptions(conf), rawInputs, namespace, c, d, backend, configCallCaching(conf), lookupDockerHash(conf),
-      wfContext, fileSystems)
+    } yield new WorkflowDescriptor(id, sourceFiles, options, workflowLogOptions(conf), rawInputs, namespace, c, d, backend, configCallCaching(conf), lookupDockerHash(conf),
```
Horneth: just wondering, why the `new` ?

engine/src/test/scala/cromwell/WorkflowFailSlowSpec.scala
cjllanwarne: No test added. Bad value test added to centaur

README.md
cjllanwarne: Hmm, apparently the config is "self-documenting" so I've mentioned it here and the config has some more explanation.

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
cjllanwarne: Hmm, I guess I can move the behaviour definition inside the object. The implementation of that behaviour itself though, I think belongs here

engine/src/main/scala/cromwell/engine/WorkflowDescriptor.scala
```diff
@@ -40,6 +40,7 @@ case class WorkflowDescriptor(id: WorkflowId,
                               backend: Backend,
                               configCallCaching: Boolean,
                               lookupDockerHash: Boolean,
+                              configWorkflowFailureMode: WorkflowFailureMode,
```
Horneth: I think that if this is going to be an attribute of the class it should be the final definitive value.
Now it is used as a default when computing the actual value on line 65.
The full logic of determining the mode should be done int eh apply IMO. It would avoid having `configWorkflowFailureMode` and `workflowFailureMode` where really you only care about the latter.

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
Horneth: Hmm I don't see why ? I can't see how we would have different classes that implement this behaviour differently in Cromwell ?
there could be a 
`def shouldFail(hasAnythingFailed: Boolean)` method in `WorkflowFailureMode` that would return `hasAnythingFailed` for `NoNewCalls` and `false` for `ContinueWhilePossible`

mcovarr: @Horneth yes that's what I meant, thanks.  "most of this logic" was not a good way for me to describe that.  :smile:

engine/src/main/scala/cromwell/engine/WorkflowDescriptor.scala
cjllanwarne: This is actually in a slightly weird spot w.r.t validation, defaults, etc. I'm going to wait until @gauravs90 commits the ValidationActor stuff because it's all going to be so much neater in the brave new world.

Horneth:
:+1:

mcovarr:
looks like a test is failing for realz

engine/src/main/scala/cromwell/engine/WorkflowDescriptor.scala
```diff
@@ -61,6 +62,8 @@ case class WorkflowDescriptor(id: WorkflowId,
     log.warn(s"$CallLogsDirOptionKey expected to be of type String", e)
     throw e
   }
+  // This exception should be validated in ValidationActor
```
mcovarr: it's called something else now

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -825,10 +826,14 @@ case class WorkflowActor(workflow: WorkflowDescriptor)
       case _ => false
     }
 
+    lazy val hasAnythingFailed = executionStore.values exists { x => x == Failed || x == Aborted }
+
+    lazy val disallowNewCalls = !workflowFailureMode.allowNewCallsAfterFailure && hasAnythingFailed
```
mcovarr: maybe flip the operands

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -825,10 +826,14 @@ case class WorkflowActor(workflow: WorkflowDescriptor)
...
+
     def isRunnable(entry: ExecutionStoreEntry) = {
       entry match {
-        case (key: FinalCallKey, ExecutionStatus.NotStarted) => nonFinalCallsComplete
-        case (key, ExecutionStatus.NotStarted) => arePrerequisitesDone(key)
+        case (key: FinalCallKey, ExecutionStatus.NotStarted) => nonFinalCallsComplete && !disallowNewCalls
+        case (key, ExecutionStatus.NotStarted) => arePrerequisitesDone(key) && !disallowNewCalls
```
mcovarr: can this be rephrased positively rather than doubly negative?

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
mcovarr: eh scratch that, it's probably more efficient the way you've got it

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
cjllanwarne: `I don't see why not`

Horneth:
Do we want this for the release ?

engine/src/main/scala/cromwell/engine/backend/WorkflowDescriptor.scala
```diff
@@ -61,6 +62,8 @@ case class WorkflowDescriptor(id: WorkflowId,
     log.warn(s"$CallLogsDirOptionKey expected to be of type String", e)
     throw e
   }
+  // This exception should be validated in ValidationActor
```
mcovarr: The actor formerly known as ValidationActor

geoffjentry: nay, née

cjllanwarne: I haven't rebased this yet! Give me a chance!

engine/src/test/scala/cromwell/util/SampleWdl.scala
```diff
@@ -2306,7 +2306,8 @@ task shouldCompleteFast {
         |    Int a
         |    command {
         |        echo "The number was: ${a}"
-        |        sleep 0.2
+        |        # More than 0.15 so this should finish second
```
mcovarr: This 0.1 second stuff still makes me kind of uneasy, wouldn't be surprised if Travis gets wonky with this

mcovarr:
:+1: assuming Travis approves

mcovarr:
unhappy Travis :cry:

PR #594, opened 2016-03-21
Adding some missing information to the README

scottfrazer:


README.md
```diff
@@ -750,6 +751,43 @@ task runtime_test {
...
+  "defaultRuntimeOptions": {
+    "docker": "ubuntu:latest",
+    "zones": "us-central1-a us-central1-b"
+  }
+}
+```
+
+Then these values for `docker` and `zones` will be used for any task that does not explicitly override them in the WDL file.
+
+In the example above, `task first` will inherit from the `defaultRuntimeOptions` section of the workflow options.  `task second` will only use the "zones" information from the workflow options.
```
cjllanwarne: `task second` will use the zones from the WDL

cjllanwarne:
Otherwise good. :+1:

README.md
scottfrazer: but `task second` doesn't specify `zones` in its runtime attributes, so it should use the value in `defaultRuntimeOptions`, right?

cjllanwarne: Oh, I got confused. Maybe `task second will use zones from the default in workflow options but because docker is specified in the WDL, that value will be used`

scottfrazer: okay, will change

cjllanwarne:
Oh, also in the Command Line Options section under Run - it says "workflow options only apply to JES" - this is certainly no longer true

scottfrazer:
@cjllanwarne I'll update that too

README.md
Horneth: I would even maybe provide the final composition of attributes that will actually be run for task first and second:
```
task first at runtime will be:

task first {
   command { ... }
  runtime {
   docker: "ubuntu:latest",
   zones: "us-central1-a us-central1-b"
 }
 }
```
...

cjllanwarne:
I'll absorb this change into an upcoming PR

cjllanwarne:
I've taken this over from @scottfrazer. @Horneth's self-selected himself for review. Bravo!

Horneth:
:+1:

PR #604, opened 2016-03-22
Initial commit to start discussing interfaces and implementation. Closes #550.

### LINKED ISSUE #550: [LCM] Backend interface definition
1. Match functions between old Backend and new Backend traits
2. Add missing functions in new Backend

francares:
This is the first iteration on the Cromwell-backend LCM design and implementation.

The idea is to start commenting on this in order to get a final initial version to start the implementation/refactoring of different backends.

Please, DO NOT MERGE.


backend/src/main/scala/cromwell/backend/BackendActorFactory.scala
```diff
@@ -0,0 +1,17 @@
...
+  */
+trait BackendActorFactory {
+  /**
+    * Returns a backend actor instance based on the name, Cromwell actor system and specific task.
+    *
+    * @param initClass Initialization class for the specific backend.
+    * @param actorSystem Cromwell Engine actor system.
+    * @return An actor of type BackendActor.
+    */
+  def getBackend(initClass: String, actorSystem: ActorSystem): ActorRef
```
francares: WorkflowDescriptor is missing here.

core/src/main/scala/cromwell/core/eventbus/PubSubMediator.scala
```diff
@@ -0,0 +1,48 @@
+package cromwell.core.eventbus
+
+import akka.actor.{ActorRef, ActorSystem}
+
+final case class EventMsg[A](topic: A, payload: Any)
+
+/**
+  * This is a wrapper to EventStream as one of the event bus with SubChannel trait.
+  */
+trait PubSubMediator {
```
mcovarr: Per Carlos's email I thought we weren't going to use this?

backend/src/main/scala/cromwell/backend/Backend.scala
```diff
@@ -0,0 +1,36 @@
+package cromwell.backend
+
+import cromwell.backend.model.{WorkflowDescriptor, JobDescriptor}
+
+/**
+  * Defines basic structure and functionality to work with a backend.
+  */
+trait Backend {
```
mcovarr: It seems there's parallel but distinct nomenclature on the engine and backend sides for workflow-level versus call-level.   e.g.

WorkflowActor -> BackendActor
CallActor -> BackendJobExecutorActor

Could we try to make this more consistent to reduce confusion?

backend/src/main/scala/cromwell/backend/BackendActor.scala
```diff
@@ -0,0 +1,41 @@
...
+/**
+  * Defines basic structure and functionality to initialize and make use of a backend through an Akka actor system.
+  * Backend functions should be implemented by each custom backend.
+  */
+trait BackendActor extends Backend with Actor with ActorLogging {
+  def receive: Receive = LoggingReceive {
+    case BeforeAll => beforeAll()
+    case CreateJobExecutor(jobDescriptor: JobDescriptor) => createJobExecutor(jobDescriptor)
+    case AfterAll => afterAll()
+    case Validate => validate() //TODO: to be discussed.
```
mcovarr: I thought these were supposed to be PBE :smile:

backend/src/main/scala/cromwell/backend/BackendActor.scala
```diff
@@ -0,0 +1,41 @@
...
+  case class CreateJobExecutor(jobDescriptor: JobDescriptor) extends BackendActorMessage
+  case object AfterAll extends BackendActorMessage
+  case object Validate extends BackendActorMessage
+
+  // Events
+  sealed trait BackendActorEvent
+  case object BeforeAllCompleted extends BackendActorEvent
+  case class BeforeAllFailed(throwable: Throwable) extends BackendActorEvent
+  case class JobExecutorCreated(backendJobExecutor: ActorRef) extends BackendActorEvent
+  case class JobExecutorCreationFailed(throwable: Throwable) extends BackendActorEvent
```
mcovarr: I would probably be good to pass the `JobDescriptor` with both of these so the sender will know which one failed.

backend/src/main/scala/cromwell/backend/Backend.scala
```diff
@@ -0,0 +1,36 @@
...
+
+  /**
+    * Factory for creating Jobs in the specific backend.
+    *
+    * @param jobDescriptor All information needed to execute a task in the backend.
+    * @return A BackendJobExecutor with functionality to handle the life cycle management of the task.
+    */
+  def createJobExecutor(jobDescriptor: JobDescriptor): BackendJobExecutor
+
+  /**
```
francares: Caching functionality is missing here.

backend/src/main/scala/cromwell/backend/BackendJobExecutorActor.scala
```diff
@@ -0,0 +1,42 @@
...
+  case object Stop extends BackendJobExecutorActorMessage
+  case object CleanUp extends BackendJobExecutorActorMessage
+
+  //TODO: identify other generic events within job execution functionality.
+  // Events
+  sealed trait BackendJobExecutorActorEvent
+  case object PrepareCompleted extends BackendJobExecutorActorEvent
+  case class PrepareFailed(throwable: Throwable) extends BackendJobExecutorActorEvent
+  case object ExecutionStarted extends BackendJobExecutorActorEvent
+  case class ExecutionCompleted(result: String) extends BackendJobExecutorActorEvent
```
mcovarr: What are you envisioning for this "result"?

backend/src/main/scala/cromwell/backend/BackendJobExecutorActor.scala
```diff
@@ -0,0 +1,42 @@
...
+  case object Prepare extends BackendJobExecutorActorMessage
+  case object Execute extends BackendJobExecutorActorMessage
+  case object Stop extends BackendJobExecutorActorMessage
+  case object CleanUp extends BackendJobExecutorActorMessage
+
+  //TODO: identify other generic events within job execution functionality.
+  // Events
+  sealed trait BackendJobExecutorActorEvent
+  case object PrepareCompleted extends BackendJobExecutorActorEvent
+  case class PrepareFailed(throwable: Throwable) extends BackendJobExecutorActorEvent
```
cjllanwarne: For clarity, I would swap `Complete` for `Successful` throughout

backend/src/main/scala/cromwell/backend/Backend.scala
cjllanwarne: I still don't quite get what the point of "Backend" is compared to BackendActor. Especially since this isn't on the diagram.

backend/src/main/scala/cromwell/backend/BackendActor.scala
```diff
@@ -0,0 +1,41 @@
...
+  case object ValidationCompleted extends BackendActorEvent
+  case class ValidationFailed(throwable: Throwable) extends BackendActorEvent
+
+}
+
+/**
+  * Defines basic structure and functionality to initialize and make use of a backend through an Akka actor system.
+  * Backend functions should be implemented by each custom backend.
+  */
+trait BackendActor extends Backend with Actor with ActorLogging {
```
cjllanwarne: Since this is the difficult bit, but would be identical in each of the various backends, why not make this concrete, with a `BackendActor(backend: Backend)` constructor.

That way, the difficult actor bit can be done once, here. And for every Backend, the implementer can just write a bunch of blocking functions. All of the special corralling around that blocking can be done once in the general purpose `BackendActor`?

backend/src/main/scala/cromwell/backend/Backend.scala
cjllanwarne: NB: My later comment on BackendActor is me trying to make sense of this split :)

backend/src/main/scala/cromwell/backend/BackendActor.scala
```diff
@@ -0,0 +1,41 @@
+package cromwell.backend
+
+import akka.actor.{ActorRef, Actor, ActorLogging}
+import akka.event.LoggingReceive
+import cromwell.backend.BackendActor.{Validate, AfterAll, BeforeAll, CreateJobExecutor}
+import cromwell.backend.model.JobDescriptor
+
+object BackendActor {
```
cjllanwarne: @mcovarr I'll bite! How about calling this `WorkflowExecutionActor`?

project/Dependencies.scala
```diff
@@ -54,13 +54,19 @@ object Dependencies {
     "com.mattbertolini" % "liquibase-slf4j" % "2.0.0"
   )
 
+  val coreDependencies = List(
+    "com.typesafe.akka" %% "akka-actor" % akkaV
+  ) ++ testDependencies
+
+  val backendDependencies = List(
+    "org.broadinstitute" %% "wdl4s" % wdl4sV
```
cjllanwarne: Is this move right? You're saying that the engine will never read the workflow definition but the backend will?

If anything (expressions notwithstanding and I know that backend **does** need this for now) I would have thought it would be the other way round?

backend/src/main/scala/cromwell/backend/BackendActor.scala
mcovarr: how would we deal with an arbitrary number of calls to blocking functions?

cjllanwarne: How is it handled in *this* implementation?

mcovarr: Asynchronously, not locking up threads in the fixed size context.  If there does need to be a bulkheaded pool of sacrificial threads to cope with an underlying blocking API, that would be up to the backend implementation to provide.

cjllanwarne: Another reading of that is:

> Every backend *must* be trusted to handle blocking correctly. Furthermore any blocking in a 3rd party backend will cause the Cromwell engine to break. 

Are we going to put that in our backend developer guide?

mcovarr: We'll have to spell out the contract explicitly WRT usage of threads.  That's my current mental model of how this would work, but I'm happy to discuss it further.

cjllanwarne: Is it at all possible to pass an execution context to the backend and say "this one's yours. Don't make threads anywhere else"?

backend/src/main/scala/cromwell/backend/Backend.scala
francares: The idea is that a developer who is creating a new backend, implements a new CustomBackenActor extending from BackendActor.
After that he/she just need to override backend interface with specific functionality.
With this you are saying that all new custom backends need to implement that interface.

backend/src/main/scala/cromwell/backend/BackendActor.scala
francares: WorkflowExecutionActor => it may not execute the entire workflow.
WorkflowPreparationActor => this is not a workflow.
WorkflowLevelBackendSpecificPreparationCleanupAndJobDescriptorMakerActor => too long :)

I think Backend is representative.

backend/src/main/scala/cromwell/backend/BackendActor.scala
francares: Second one yes but first one you can get that from backendJobExecutor... I think we can define a message in BackendJobExecutorActor for this since in the BackendJobExecutor trait is already defined.

backend/src/main/scala/cromwell/backend/BackendActor.scala
francares: Yes, but since we said to remove validation init class from config I tough you wanted here for backend specific validations.
Just remember this will not be instantiated in WorkflowManagerSystem.

backend/src/main/scala/cromwell/backend/BackendJobExecutorActor.scala
francares: What about *Succeeded since it will represent an event?

backend/src/main/scala/cromwell/backend/BackendJobExecutorActor.scala
francares: I just defined that because I could not get to this point yet. I need to take a look to what we have right now and propose a new structure if needed.

core/src/main/scala/cromwell/core/eventbus/PubSubMediator.scala
francares: As Carlos mentioned in his e-mail, this is there because we will need to use for workflow events logging.

project/Dependencies.scala
francares: Cromwell-backend needs from wdl4s to live... and Cromwell-engine will need from Cromwell-backend to live... then in the way the build mechanism is defined you have two options in order to not repeat the dependency entry:
1. As I defined.
2. Define a new list of dependencies with just wdl4s and add that to backend and engine.

In my opinion this should not be a big deal... we can define this using any of the two options.

backend/src/main/scala/cromwell/backend/BackendActor.scala
cjllanwarne: Hmm, I agree with your comments but I think there's a middle ground here. For example, to me the name `BackendActor` implies that it's shared between Workflows. But if I understand correctly you get a new BackendActor per backend per workflow?

backend/src/main/scala/cromwell/backend/Backend.scala
cjllanwarne: Are you expecting that receive: Receive is final method?

cjllanwarne: Why not just put these defs in the `BackendActor` trait?

backend/src/main/scala/cromwell/backend/BackendActor.scala
cjllanwarne: If you're running a workflow and try to run `BeforeAll` on two backends at the same time, if you then get a `BeforeAllFailed` back, how do you know which it was that failed to start?

backend/src/main/scala/cromwell/backend/BackendActor.scala
francares: Yes, maybe WorkflowBackendActor?

backend/src/main/scala/cromwell/backend/Backend.scala
francares: 1st comment: I'm defining the messages this actor will understand and the engine will send. Later if someone wants to create a composition of receive shouldn't be a problem.
2nd: I defined Backend because later if we implement BackendPersistentActor, I will need to implement this.
But... if we don't want to use PA's or EA's in the future I can move these as you describe.

francares: 2nd: I will remove backend trait to make it simple. Later if we implement PA's modify this. I don't see we will support PA's / EA's in a short term.

backend/src/main/scala/cromwell/backend/BackendActor.scala
```diff
@@ -0,0 +1,85 @@
...
+      sndr ! afterAll()
+    case Validate => //TODO: to be discussed.
+      val sndr = sender()
+      sndr ! validate()
+  }
+
+  /**
+    * Registers code to be executed before the backend is ready for executing jobs for the specific workflow.
+    */
+  def beforeAll(): BeforeAllEvent
```
francares: Making sure that at least one event is received by the sender.

backend/src/main/scala/cromwell/backend/BackendJobExecutorActor.scala
```diff
@@ -0,0 +1,83 @@
...
+
+object BackendJobExecutorActor {
+
+  // Commands
+  sealed trait BackendJobExecutorActorMessage
+  case object Prepare extends BackendJobExecutorActorMessage
+  case object Execute extends BackendJobExecutorActorMessage
+  case object Stop extends BackendJobExecutorActorMessage
+  case object CleanUp extends BackendJobExecutorActorMessage
+
```
francares: Should BackendJobExecutorActor notify for intermediate status of a task? Such as running, onHold, aborting?
If so, we need to define intermediate statuses events and work with subscriptions here.

francares:
Please, check hidden comments on removed traits.

backend/src/main/scala/cromwell/backend/BackendActor.scala
francares: During this week I will work on events content.

backend/src/main/scala/cromwell/backend/BackendJobExecutorActor.scala
francares: Done.

backend/src/main/scala/cromwell/backend/BackendActorFactory.scala
```diff
@@ -0,0 +1,19 @@
...
+
+import akka.actor.{ActorRef, ActorSystem}
+import cromwell.backend.model.WorkflowDescriptor
+
+/**
+  * Creates Backend instances based on backend name.
+  */
+trait BackendActorFactory {
+  /**
+    * Returns a backend actor instance based on the name, Cromwell actor system and specific task.
```
cjllanwarne: task => workflow :)

backend/src/main/scala/cromwell/backend/BackendActor.scala
```diff
@@ -0,0 +1,85 @@
...
+  case object ValidationSucceeded extends ValidationEvent
+  case class ValidationFailed(throwable: Throwable) extends ValidationEvent
+
+}
+
+/**
+  * Defines basic structure and functionality to initialize and make use of a backend through an Akka actor system.
+  * Backend functions should be implemented by each custom backend.
+  */
+trait BackendActor extends Actor with ActorLogging {
```
cjllanwarne: I like your suggestion of `WorkflowBackendActor `

backend/src/main/scala/cromwell/backend/BackendActor.scala
```diff
@@ -0,0 +1,85 @@
...
+import cromwell.backend.model.{JobDescriptor, WorkflowDescriptor}
+
+object BackendActor {
+
+  // Commands
+  sealed trait BackendActorMessage
+  case object BeforeAll extends BackendActorMessage
+  case class CreateJobExecutorActor(jobDescriptor: JobDescriptor) extends BackendActorMessage
+  case object AfterAll extends BackendActorMessage
+  case object Validate extends BackendActorMessage
```
cjllanwarne: It's hidden now but you mentioned caching before as something that needs to be added. Did you mean for the caching writing or reading

Did you foresee other commands being necessary for that/those?

core/src/main/scala/cromwell/core/eventbus/PubSubMediator.scala
cjllanwarne: My reading of the email was that the pub/sub mediator is no longer the primary way of passing control messages between engine and backend - but rather a logging and tracking mechanism. 

If that's a correct and true reading, I think I'd lobby for focusing this PR just on the engine/backend interface and adding the pub/sub mediator in another PR focussed on that? If I've missed the point of the email, maybe a clarification? Thanks!

backend/src/main/scala/cromwell/backend/BackendActorFactory.scala
francares: Fixed.

cjllanwarne:
My comment to resurrect a hidden comment was also just hidden - so I'll drop it here instead :)

You mentioned caching before as something that needs to be added to the BackendActor. Did you mean for the caching writing or reading? And did you foresee other command messages for the BackendActor being necessary for that/those?

francares:
@cjllanwarne yes, comments were hidden due to file change name.
Ans 1. Adding hidden comment I did: Caching functionality is missing here. Shouldn't each backend implement caching and when engine ask for a jobExecutor, backend may return BackendCachedJobExecutor?
Doing that we can get rid of the engine responsibility to deal with cached data...
IMO, Cache should be encapsulated in each backend. The only thing I'm not sure if we should expose a standard message to force not to use cached data. So with that you tell to each backend to not use cached data but instead to process data again.
Ans 2. I'm not seeing any new msg for WorkflowBackendActor right now. That will depend on the UCs... for CCC backend those msgs are OK.

cjllanwarne:
@francares Ah! I see. Yeah, I asked because I wasn't sure why we'd need another command type, we could either make a different type of ExecutorActor or have a branch in the existing ExecutorActor for caches. So we agree there.

As for another message type, I believe this is all configured in (a) global config file and (b) workflow options so the actors should already have everything they need to decide whether to allow caching or not. So IMO there's no special case at this layer to handle caching.

project/Dependencies.scala
```diff
@@ -70,5 +76,5 @@ object Dependencies {
     "org.codehaus.janino" % "janino" % "2.7.8",
     "org.scalaz" % "scalaz-core_2.11" % "7.1.3",
     "com.github.pathikrit" %% "better-files" % "2.13.0"
-  ) ++ testDependencies ++ kamonDependencies ++ sprayDependencies ++ googleDependencies ++ dbDependencies
```
Horneth: why did the `testDependencies` disappear ?

Horneth: Nevermind I just saw they moved to core

backend/src/main/scala/cromwell/backend/BackendActor.scala
```diff
@@ -0,0 +1,146 @@
...
+  def receive: Receive = LoggingReceive {
+    case Execute(jobDescriptor: JobDescriptor) =>
+      val sndr = sender()
+      execute(jobDescriptor) onComplete {
+        case Success(executionEvent: ExecutionEvent) =>
+          sndr ! executionEvent
+        case Failure(exception: Throwable) =>
+          sndr ! ExecutionFailed(exception)
+      }
+    case Abort(jobDescriptor: JobDescriptor) =>
```
francares: Do you see any scenario where Abort should be a future?

backend/src/main/scala/cromwell/backend/BackendActor.scala
```diff
@@ -0,0 +1,146 @@
...
+import scala.concurrent.Future
+import scala.util.{Success, Failure}
+
+object BackendActor {
+
+  // Commands
+  sealed trait BackendActorMessage
+  case class Execute(jobDescriptor: JobDescriptor) extends BackendActorMessage
+  case class Abort(jobDescriptor: JobDescriptor) extends BackendActorMessage
+  case object AbortAll extends BackendActorMessage
```
cjllanwarne: I don't foresee a case where we abort a single call, so I would merge these two into a single `abort` message which aborts everything.

backend/src/main/scala/cromwell/backend/BackendActor.scala
```diff
@@ -0,0 +1,146 @@
...
+  case class AbortAllSucceeded(jobDescriptor: List[JobDescriptor]) extends AbortAllEvent //Attributes TBD.
+  case class AbortAllFailed(throwable: Throwable) extends AbortAllEvent
+
+  sealed trait RecoverEvent extends BackendActorEvent
+  case class RecoverSucceeded(jobDescriptor: JobDescriptor) extends RecoverEvent //Same happens here.
+  case class RecoverFailed(throwable: Throwable) extends RecoverEvent
+
+  sealed trait ValidationEvent extends BackendActorEvent
+  case object ValidationSucceeded extends ValidationEvent
+  case class ValidationFailed(throwable: Throwable) extends ValidationEvent
```
cjllanwarne: I thought the decision was that WorkflowBackendActor-level validation was going to be in the constructor and we'd just throw an exception if it failed?

backend/src/main/scala/cromwell/backend/BackendActor.scala
```diff
@@ -0,0 +1,146 @@
...
+
+  sealed trait BeforeAllEvent extends BackendActorEvent
+  case object BeforeAllSucceeded extends BeforeAllEvent
+  case class BeforeAllFailed(throwable: Throwable) extends BeforeAllEvent
+
+  sealed trait AfterAllEvent extends BackendActorEvent
+  case object AfterAllSucceeded extends AfterAllEvent
+  case class AfterAllFailed(throwable: Throwable) extends AfterAllEvent
```
cjllanwarne: I thought we'd use the akka beforeStart and postStop (names probably different) for these?

backend/src/main/scala/cromwell/backend/BackendActor.scala
geoffjentry: plus-one to chris

geoffjentry: although to play devils advocate how is everything aborted, perhpas with per-call abort signals internally

backend/src/main/scala/cromwell/backend/BackendActor.scala
cjllanwarne: Actually I'm not sure I'm right. @mcovarr @geoffjentry ?

backend/src/main/scala/cromwell/backend/BackendActor.scala
geoffjentry: We *might* be able to do away with our up front fast validation business requirement which simplifies this

backend/src/test/scala/cromwell/backend/BackendActorSpec.scala
```diff
@@ -0,0 +1,156 @@
...
+    *
+    * @return An AbortAllEvent with the result.
+    */
+  override def abortAll(): AbortAllEvent = ???
+
+  /**
+    * Executes validation on workflow descriptor in order to see if the workflow can be executed by the backend.
+    */
+  override def validate(): Unit = {
+    context.parent ! ValidationSucceeded
```
cjllanwarne: This test spec only seems to validate that this test actor is implement correctly?

backend/src/main/scala/cromwell/backend/BackendActor.scala
```diff
@@ -0,0 +1,146 @@
...
+
+  /**
+    * Defines needed data to be able to execute a workflow.
+    */
+  val workflowDescriptor: WorkflowDescriptor
+
+  def receive: Receive = LoggingReceive {
+    case Execute(jobDescriptor: JobDescriptor) =>
+      val sndr = sender()
+      execute(jobDescriptor) onComplete {
```
geoffjentry: Could you use a pipeTo and pipe the execution event directly to the sender?

backend/src/main/scala/cromwell/backend/BackendActor.scala
cjllanwarne: Yeah, abort may well be an API call to a remote service

backend/src/main/scala/cromwell/backend/BackendActor.scala
```diff
@@ -0,0 +1,146 @@
...
+    */
+  def abortAll(): AbortAllEvent
+
+  /**
+    * Tries to recover last status and information on an on going job in the backend.
+    *
+    * @return An RecoverEvent with the result.
+    */
+  //TODO: Need more details on this in order to define it.
+  def recover(): RecoverEvent
```
cjllanwarne: This is probably also better as a Future

backend/src/main/scala/cromwell/backend/BackendActor.scala
Horneth: @cjllanwarne Noooo

backend/src/main/scala/cromwell/backend/BackendActor.scala
geoffjentry: In general I'm wondering if the general pattern on these should be `Future[FooEvent]` which is then pipeTo'd back to sender

backend/src/main/scala/cromwell/backend/BackendActorFactory.scala
```diff
@@ -0,0 +1,19 @@
...
+import akka.actor.{ActorRef, ActorSystem}
+import cromwell.backend.model.WorkflowDescriptor
+
+trait BackendActorFactory {
+
+  /**
+    * Returns a backend actor instance based on the init class, Cromwell actor system and specific workflow.
+    *
+    * @param initClass          Initialization class for the specific backend.
+    * @param actorSystem        Cromwell Engine actor system.
```
cjllanwarne: Technically this would be the actor system designated for backend actors, not engine actors :)

backend/src/main/scala/cromwell/backend/BackendActor.scala
```diff
@@ -0,0 +1,146 @@
...
+
+  // Commands
+  sealed trait BackendActorMessage
+  case class Execute(jobDescriptor: JobDescriptor) extends BackendActorMessage
+  case class Abort(jobDescriptor: JobDescriptor) extends BackendActorMessage
+  case object AbortAll extends BackendActorMessage
+  case class Recover(jobDescriptor: JobDescriptor) extends BackendActorMessage
+
+  // Events
+  sealed trait BackendActorEvent
```
kcibul: Forgive me if I didn't understand this properly, but on the call yesterday it sounded like form a contract perspective the event messages are irrelevant.  They are not part of the contract.  If so, lets move them out of of here since the are important only to specific backends.  If they're here they're part of the contract.

backend/src/test/scala/cromwell/backend/BackendActorSpec.scala
francares: I just created this test as an example of how it would looks like preStart and postStop in BackendActor.

backend/src/main/scala/cromwell/backend/BackendActor.scala
kcibul: Resolved in discussion, let's leave it as is

backend/src/main/scala/cromwell/backend/BackendActor.scala
```diff
@@ -0,0 +1,146 @@
...
+      val sndr = sender()
+      execute(jobDescriptor) onComplete {
+        case Success(executionEvent: ExecutionEvent) =>
+          sndr ! executionEvent
+        case Failure(exception: Throwable) =>
+          sndr ! ExecutionFailed(exception)
+      }
+    case Abort(jobDescriptor: JobDescriptor) =>
+      sender() ! abort(jobDescriptor)
+    case AbortAll =>
```
francares: Remove this.

backend/src/main/scala/cromwell/backend/BackendActor.scala
francares: Remove jobDescriptor.

backend/src/main/scala/cromwell/backend/BackendActor.scala
```diff
@@ -0,0 +1,146 @@
...
+      sender() ! recover()
+  }
+
+  /**
+    * Makes mandatory the execution of validation and beforeAll.
+    * Validation: intention here is to check if runtime attributes complies with backend specifications.
+    * BeforeAll: Execute any required pre-condition before making backend available for executions.
+    */
+  override def preStart(): Unit = {
+    validate()
```
francares: Remove this line from here.

backend/src/main/scala/cromwell/backend/BackendActor.scala
francares: TBD on Monday.

backend/src/main/scala/cromwell/backend/BackendActorFactory.scala
francares: Yes, we need to think if there will be any other communication between BackendActors and any other actor that belongs to Engine's actor system.

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
```diff
@@ -0,0 +1,137 @@
...
+  case class ExecutionSucceeded(result: String) extends ExecutionEvent //Result TBD.
+  case class ExecutionFailed(throwable: Throwable) extends ExecutionEvent
+
+  sealed trait AbortEvent extends WorkflowBackendActorEvent
+  case object AbortSucceeded extends AbortEvent
+  case class AbortFailed(throwable: Throwable) extends AbortEvent
+
+  sealed trait RecoverEvent extends WorkflowBackendActorEvent
+  case class RecoverSucceeded(jobDescriptor: JobDescriptor) extends RecoverEvent //Same happens here.
+  case class RecoverFailed(throwable: Throwable) extends RecoverEvent
```
mcovarr: Per 4/4 meeting, we may be able to get rid of this assuming the backend can persist the association b/w JobDescriptor and the implementation-specific run ID.

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
```diff
@@ -0,0 +1,137 @@
...
+  case class Execute(jobDescriptor: JobDescriptor) extends WorkflowBackendActorMessage
+  case object Abort extends WorkflowBackendActorMessage
+  case class Recover(jobDescriptor: JobDescriptor) extends WorkflowBackendActorMessage
+
+  // Events
+  sealed trait WorkflowBackendActorEvent
+
+  sealed trait ExecutionEvent extends WorkflowBackendActorEvent
+  case class ExecutionSucceeded(result: String) extends ExecutionEvent //Result TBD.
+  case class ExecutionFailed(throwable: Throwable) extends ExecutionEvent
```
mcovarr: add Job ID to all responses

mcovarr: add call outputs for ExecutionSucceeded

mcovarr: Split ExecutionFailed into transient (or retryable) / non-transient (non-retryable)

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
```diff
@@ -0,0 +1,137 @@
...
+
+  sealed trait AbortEvent extends WorkflowBackendActorEvent
+  case object AbortSucceeded extends AbortEvent
+  case class AbortFailed(throwable: Throwable) extends AbortEvent
```
mcovarr: parameterize by jobs that failed to abort

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
```diff
@@ -0,0 +1,137 @@
...
+
+import scala.concurrent.Future
+import scala.util.{Success, Failure}
+
+object WorkflowBackendActor {
+
+  // Commands
+  sealed trait WorkflowBackendActorMessage
+  case class Execute(jobDescriptor: JobDescriptor) extends WorkflowBackendActorMessage
+  case object Abort extends WorkflowBackendActorMessage
```
mcovarr: parmeterize this by JobDescriptor like the other messages

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
mcovarr: add JobID to these responses too

francares:
- WA will receive a WdlSource in preStart will try to validate it.
- BA will receive a list of jobDescriptors and will try to validate runtime requirements.
- Needs to defined if the will be a MIM in validation state in WA.

francares:
BA will receive (Calls, workflowInputs, workflowOptions), config, serviceReg.

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
ccarrizo: as per the meeting on 4/6 recover is actually needed, although in the case of a local BE it might mean only execute, while in case of remote execution (e.g. JES) it has a more comprehensive behaviour

cjllanwarne:
Superceded by #677

PR #605, opened 2016-03-23
Sub engine function Closes #597

### LINKED ISSUE #597: sub() expression language function
As a pipeline author, I don't enjoy having to spin up a VM, with docker in order to do string substitutions on my parameters.  In the GOTC pipeline, we do this in order to strip off the extension of the input file in order to get a base-name, which happens ~40 times in a 20-plex workflow.  

This causes a real problem because by requiring so many VMs to spun up, we spend more money (although that cost is quite small) but also eat into our quotas and QPS limits, which actually does hurt our scalability.

The proposal is to add a new expression language function which allows for regex substitutions:

sub(string, pattern, replacement)

For example,

to strip off an extension from a file you could use `sub(filename, ".bam$","")`
to swap an extension, you could use `sub(filename, ".bam$", ".metrics")`

By being constrained to a regex (unlike an arbitrary code block) we don't have concerns about security or evaluating these in the cromwell engine.

This does not eliminate the need for a generally more expressive expression language or user defined functions, but does solve a large class of common usages that impact ease of use and performance

Horneth:
- [x] Needs https://github.com/broadinstitute/wdl4s/pull/19 merged to compile
- [x] Changelog
- [x] Update WDL SPEC

~~Add documentation~~ Documentation is in WDL SPEC

Also https://github.com/broadinstitute/wdl/pull/32

### DIFF
--- a/engine/src/main/scala/cromwell/engine/WdlStandardLibraryImpl.scala
+++ b/engine/src/main/scala/cromwell/engine/WdlStandardLibraryImpl.scala
@@ -95,6 +95,26 @@ trait WdlStandardLibraryImpl extends WdlStandardLibraryFunctions {
     */
   override def read_float(params: Seq[Try[WdlValue]]): Try[WdlFloat] = read_string(params) map { s => WdlFloat(s.value.trim.toDouble) }
 
+  override def sub(params: Seq[Try[WdlValue]]): Try[WdlString] = {
+    def extractArguments = params.size match {
+      case 3 => Success((params.head, params(1), params(2)))
+      case n => Failure(new IllegalArgumentException(s"Invalid number of parameters for engine function sub: $n. sub takes exactly 3 parameters."))
+    }
+
+    def validateArguments(values: (Try[WdlValue], Try[WdlValue], Try[WdlValue])) = values match {
+      case (Success(strValue), Success(WdlString(pattern)), Success(replaceValue))
+        if WdlStringType.isCoerceableFrom(strValue.wdlType) &&
+           WdlStringType.isCoerceableFrom(replaceValue.wdlType) =>
+        Success((strValue.valueString, pattern, replaceValue.valueString))
+      case _ => Failure(new IllegalArgumentException(s"Invalid parameters for engine function sub: $values."))
+    }
+
+    for {
+      args <- extractArguments
+      (str, pattern, replace) <- validateArguments(args)
+    } yield WdlString(pattern.r.replaceAllIn(str, replace))
+  }
+
   /**
     * Try to read a boolean from the file referenced by the specified `WdlValue`.
     */
--- a/engine/src/test/scala/cromwell/SubFunctionWorkflowSpec.scala
+++ b/engine/src/test/scala/cromwell/SubFunctionWorkflowSpec.scala
@@ -1,0 +1,59 @@
+package cromwell
+
+import java.nio.file.Files
+
+import akka.testkit.EventFilter
+import cromwell.util.SampleWdl
+import wdl4s._
+import wdl4s.types.{WdlArrayType, WdlFileType}
+import wdl4s.values.{WdlArray, WdlFile, WdlString}
+
+class SubFunctionWorkflowSpec extends CromwellTestkitSpec {
+
+  object SubEngineFunction extends SampleWdl {
+    override def wdlSource(runtime: String = "") =
+      """task sub {
+        |  String myBamString = "myfilename.bam"
+        |  File myBamFile
+        |  String swappedStr = sub(myBamString, ".bam$", ".txt")
+        |  String swappedFile = sub(myBamFile, ".bam$", ".txt")
+        |
+        |  command {
+        |    echo ${sub(myBamFile, ".bam$", ".txt")}
+        |  }
+        |
+        |  output {
+        |    Array[File] o = [read_string(stdout()), swappedFile]
+        |    String o2 = swappedStr
+        |  }
+        |}
+        |
+        |workflow wf {
+        |  String test = sub("ab", "a", "b")
+        |  call sub
+        |}
+      """.stripMargin
+
+    val tempDir = Files.createTempDirectory("SubEngineFunction")
+    override val rawInputs: WorkflowRawInputs = Map("wf.sub.myBamFile" -> createFile("myfilename.bam", tempDir, "arbitrary content").getAbsolutePath)
+  }
+
+  "sub engine function" should {
+    "apply a regex to a string-like WdlValue" in {
+      val outputs = Map(
+        "wf.sub.o" -> WdlArray(WdlArrayType(WdlFileType), Seq(
+          WdlFile(SubEngineFunction.tempDir.resolve("myfilename.txt").toString),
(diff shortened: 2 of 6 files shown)


Horneth:
wheel @scottfrazer @mcovarr

cjllanwarne:
Could you add the Closes #xxx tag?

project/Dependencies.scala
```diff
@@ -2,7 +2,7 @@ import sbt._
 
 object Dependencies {
   lazy val lenthallV = "0.16"
-  lazy val wdl4sV = "0.4-b1562df-SNAPSHOT"
+  lazy val wdl4sV = "0.4-171214a-SNAPSHOT"
```
mcovarr: Is this in artifactory?  If no, that explains the test failures.

mcovarr: nm, I see it's in your description...

mcovarr:
What's here looks good, but our tech talk left me wondering if this is actually going to address the problems the ticket is looking to solve.  i.e., if we tell people there's this new `sub()` function but when they try to write real-world WDLs that use it they can't because of existing language restrictions that aren't really related to the changes here.  I'm thinking we might want to ask Green for samples of their workflows that have been `sub`bed and see if Cromwell can deal with them on this branch.

scottfrazer:
Update WDL spec too

engine/src/test/scala/cromwell/engine/EngineFunctionsSpec.scala
```diff
@@ -15,10 +19,57 @@ class EngineFunctionsSpec extends FlatSpec with Matchers {
...
+    }
+
+    val engineFn = new TestEngineFn
+
+    val table = Table(
+      ("str", "pattern", "replace", "result"),
+      ("somefilename.bam", ".bam$", ".txt", "somefilename.txt"),
+      ("somefilename.bam.bam", ".bam$", ".txt", "somefilename.bam.txt"),
+      ("somefilenamewith.baminside.ext", ".bam$", ".txt", "somefilenamewith.baminside.ext"),
+      ("somefilename", "^some", "other", "otherfilename")
```
kcibul: can you also add a test case for replacing ".bam$" with "" (basically to strip off an extension)?

engine/src/test/scala/cromwell/SubFunctionWorkflowSpec.scala
```diff
@@ -0,0 +1,59 @@
+package cromwell
```
cjllanwarne: @geoffjentry is centaur ready-enough for the prime-time for me to say `add this in centaur instead`?

Horneth:
So in order to replace `StripBamExtension` with `sub` without making other changes, this is what I had to do in the single sample wdl:

```
String file_name_regex = "gs://.*/"
String ext_regex = unmapped_bam_suffix + "$"

# Align flowcell-level unmapped input bams in parallel
  scatter (unmapped_bam in flowcell_unmapped_bams) {
    call CollectQualityYieldMetrics {
      input:
        input_bam = unmapped_bam,
        metrics_filename = sub(sub(unmapped_bam, file_name_regex, ""), ext_regex, ".quality_yield_metrics")
    }
    call SamToFastqAndBwaMem {
      input:
        input_bam = unmapped_bam,
        output_bam_basename = sub(sub(unmapped_bam, file_name_regex, ""), ext_regex, ".unmerged"),
        ref_fasta = ref_fasta,
        ref_fasta_index = ref_fasta_index
...
```

It's not very pretty but it does get rid of `StripBamExtension`. We can make this better by allowing scatter-scoped declarations which would allow for  
```
scatter (unmapped_bam in flowcell_unmapped_bams) {
  String base_name = sub(sub(unmapped_bam, file_name_regex, "")

  call CollectQualityYieldMetrics {
      input:
        input_bam = unmapped_bam,
        metrics_filename = sub(base_name, ext_regex, ".quality_yield_metrics")
    }
}
```

Or even maybe do the `sub` directly in the task command.
But even as is this allow for removal of `StripBamExtension`.

engine/src/test/scala/cromwell/engine/EngineFunctionsSpec.scala
Horneth: Will do

engine/src/main/scala/cromwell/engine/WdlStandardLibraryImpl.scala
```diff
@@ -95,6 +95,26 @@ trait WdlStandardLibraryImpl extends WdlStandardLibraryFunctions {
...
   override def read_float(params: Seq[Try[WdlValue]]): Try[WdlFloat] = read_string(params) map { s => WdlFloat(s.value.trim.toDouble) }
 
+  override def sub(params: Seq[Try[WdlValue]]): Try[WdlString] = {
+    def extractArguments = params.size match {
+      case 3 => Success((params.head, params(1), params(2)))
+      case n => Failure(new IllegalArgumentException(s"Invalid number of parameters for engine function sub: $n. sub takes exactly 3 parameters."))
+    }
+
+    def validateArguments(values: (Try[WdlValue], Try[WdlValue], Try[WdlValue])) = values match {
+      case (Success(strValue), Success(WdlString(pattern)), Success(replaceValue))
```
cjllanwarne: Any reason param 2 is forced to be a literal string but the others can be coerced on the fly?

engine/src/test/scala/cromwell/SubFunctionWorkflowSpec.scala
geoffjentry: I'd still say both simply because it's not yet automated like tyburn is. However that's in flight (waiting on devops) and centaur is already as feature-ful as tyburn

engine/src/main/scala/cromwell/engine/WdlStandardLibraryImpl.scala
Horneth: First I did that, but the second one being expected to be a regex it felt weird to allow it being derived from a `File` for example or something like that..

cjllanwarne: Oh! That's a regex? So if the file name contains a "." that'll be treated as "any character"?

cjllanwarne: i.e. will `sub("metxt.txt", ".txt", ".hello")` create `"m.hello.hello"`

Horneth: Yep. I guess it's a matter of definition, but if we do want to have regex capabilities then I don't see any way around that except having a different function that consider `pattern` as a simple string.

cjllanwarne: Well that explains the $ at the end of the pattern I guess... I was wondering about that! OK.

Horneth: yes, but `sub("metxt.txt", "\\.txt", ".hello")` or even `sub("metxt.txt", "txt$", "hello")` will create `"metxt.hello"`

scottfrazer:
:+1:

mcovarr:
:+1:

Issue #596, opened 2016-03-21
Call Wars Episode I: The Status Menace


mcovarr:
The complicated retry/ack dance between WA and CA is meant to ensure the proper sequencing of status writes.  A more straightforward way to accomplish this would be to use CRDTs as per Kuhn and Allen Reactive Design Patterns.  This way we don't have to care about the order of the writes, executions will always end up with the right status no matter which write attempt happens last.

A superior sequel to this could use the EXECUTION_INFO table in an event sourcing-like append-only manner to record all statuses and the times at which they were generated within Cromwell.  A database view could present an image of all this that looked the same as the EXECUTION table does now, except we'd have conflict-free writes, simplified Cromwell internals, and event data in the DB that ops currently has to scour logs to find.  But the first paragraph here is a prerequisite for all this and a good place to start.

geoffjentry:
:+1: 

![image](https://cloud.githubusercontent.com/assets/961771/13930865/1bdf4c9c-ef77-11e5-978c-8f73a11ff882.png)

PR #617, opened 2016-03-25
Storing stdout/stderr as execution info. Closes #487

### LINKED ISSUE #487: Move stdoutStderr
Look at code in existing but closed PRs in cromwell-backend/cromwell.  This will need to be greatly adapted for develop, but should be doable.

kshakir:


engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -491,6 +490,35 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
...
+
+  private def upsertExecutionInfo(executionId: Int)(keyValue: (String, Option[String]))
+                                 (implicit ec: ExecutionContext): DBIO[Unit] = {
+    val (key, value) = keyValue
+    for {
+      backendUpdate <- dataAccess.executionInfoValueByExecutionAndKey(executionId, key).update(value)
+      _ <- backendUpdate match {
+        case 0 => dataAccess.executionInfosAutoInc += ExecutionInfo(executionId, key, value)
+        case 1 => DBIO.successful(Unit)
+        case _ => DBIO.failed(new RuntimeException(s"Unexpected backend update count $backendUpdate"))
```
mcovarr: Very cool.  I hope to make `EXECUTION_INFO` append-only [one day soon](#596), but this is still cool.  :smile:

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -466,9 +467,14 @@ case class WorkflowActor(workflow: WorkflowDescriptor)
       val updatedData = startRunnableCalls(data)
       self ! CheckForWorkflowComplete
       stay() using updatedData
-    case Event(CallStarted(callKey), data) if !data.isPending(callKey) =>
+    case Event(CallStarted(callKey, maybeCallLogs), data) if !data.isPending(callKey) =>
       executionStore += callKey -> ExecutionStatus.Running
-      persistStatus(callKey, ExecutionStatus.Running)
+      // TODO: We're not using/waiting for the future results
+      // TODO: This should be in a single transaction!
```
mcovarr: This [inspired me](#618). :smile:

engine/src/test/scala/cromwell/engine/db/ExecutionInfosByExecutionSpec.scala
```diff
@@ -0,0 +1,222 @@
...
+          CallLogs(WdlFile("/my/stdout_1_2.txt"), WdlFile("/my/stderr_1_2.txt"))
+        )
+      ),
+      "my.other_scope.name" -> List(
+        List(
+          CallLogs(WdlFile("/my/stdout_other.txt"), WdlFile("/my/stderr_other.txt"))
+        )
+      )
+    ))
+  }
```
mcovarr: Very nice tests!  :star:

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -466,9 +467,14 @@ case class WorkflowActor(workflow: WorkflowDescriptor)
       val updatedData = startRunnableCalls(data)
       self ! CheckForWorkflowComplete
       stay() using updatedData
-    case Event(CallStarted(callKey), data) if !data.isPending(callKey) =>
+    case Event(CallStarted(callKey, maybeCallLogs), data) if !data.isPending(callKey) =>
       executionStore += callKey -> ExecutionStatus.Running
-      persistStatus(callKey, ExecutionStatus.Running)
+      // TODO: We're not using/waiting for the future results
```
mcovarr: The persist of the execution status is being waited for by the `data.addPersisting` stuff, but you have a point on the call logs.

mcovarr: Actually you should be able to flip the order of these statements so at least Cromwell wouldn't exit with the call logs write pending.

engine/src/main/scala/cromwell/engine/db/ExecutionInfosByExecution.scala
```diff
@@ -0,0 +1,101 @@
...
+case class ExecutionInfosByExecution(execution: Execution, executionInfos: Seq[ExecutionInfo]) {
+
+  lazy val callLogs: Option[CallLogs] = {
+    import ExecutionInfosByExecution._
+
+    case class Accumulator(stdout: Option[WdlFile], stderr: Option[WdlFile], logs: Map[String, WdlFile])
+
+    val acc = executionInfos.foldLeft(Accumulator(None, None, Map.empty)) { (acc, executionInfo) =>
+      executionInfo.key.split("_", 2) match {
+        case Array(CallLogPrefix, StdoutSuffix) => acc.copy(stdout = executionInfo.value map {
```
mcovarr: `map WdlFile.apply` might read more smoothly, here and below

engine/src/main/scala/cromwell/engine/db/ExecutionInfosByExecution.scala
```diff
@@ -0,0 +1,101 @@
...
+          WdlFile(_)
+        })
+        case Array(CallLogPrefix, StderrSuffix) => acc.copy(stderr = executionInfo.value map {
+          WdlFile(_)
+        })
+        case Array(CallLogPrefix, callLogName) =>
```
mcovarr: maybe roll this into a guard and let the `None` case fall through?

```scala
case Array(CallLogPrefix, callLogName) if executionInfo.value.isDefined =>
  acc.copy(logs = acc.logs + (callLogName -> WdlFile(executionInfo.value.get)))
```

engine/src/main/scala/cromwell/engine/db/ExecutionInfosByExecution.scala
mcovarr: nm, this doesn't seem to play as nicely with the defaulted arg on `WdlFile` as I had hoped :frowning:

engine/src/main/scala/cromwell/engine/db/ExecutionInfosByExecution.scala
```diff
@@ -0,0 +1,101 @@
...
+            case None => acc
+          }
+        case _ => acc
+      }
+    }
+
+    for {
+      stdout <- acc.stdout
+      stderr <- acc.stderr
+    } yield CallLogs(stdout, stderr, Option(acc.logs).filterNot(_.isEmpty))
```
mcovarr: we don't want to warn for stdout xor stderr being defined?

mcovarr:
:+1:

engine/src/main/scala/cromwell/engine/db/ExecutionInfosByExecution.scala
kshakir: By "warn", we could throw an exception instead. It does qualify possibly as an illegal state.

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
kshakir: Heh, nice. Let's discuss post standup where it may be easier to sketch out.

I do see additional `cromwell.engine.db.slick` methods returning `DBIO[_]` objects, such that they could be more easily composed into different transactions. But I feel that anywhere we have a composed database `Future` should be encapsulated and pushed down into `cromwell.engine.db` methods, instead of pulled up into the actors, as part of SoC.

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
kshakir: Flipped

engine/src/main/scala/cromwell/engine/db/ExecutionInfosByExecution.scala
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+}
+
+object ExecutionInfosByExecution {
+  private val CallLogPrefix = "$log"
+  private val StdoutSuffix = "stdout"
+  private val StderrSuffix = "stderr"
+
+  def toCallLogMap(callLogs: CallLogs): Map[String, Option[String]] = {
```
scottfrazer: Is there a reason why this can't return a Map[String, String]?  Unless I'm reading this wrong, the value will always be a `Some`?

scottfrazer:
:+1: 

[Approved with PullApprove](https://pullapprove.com/broadinstitute/cromwell/pull-request/617/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=broadinstitute/cromwell)

engine/src/main/scala/cromwell/engine/db/ExecutionInfosByExecution.scala
```diff
@@ -0,0 +1,107 @@
...
+import wdl4s.values.WdlFile
+
+case class ExecutionInfosByExecution(execution: Execution, executionInfos: Seq[ExecutionInfo]) {
+
+  lazy val callLogs: Option[CallLogs] = {
+    import ExecutionInfosByExecution._
+
+    case class Accumulator(stdout: Option[WdlFile], stderr: Option[WdlFile], logs: Map[String, WdlFile])
+
+    val acc = executionInfos.foldLeft(Accumulator(None, None, Map.empty)) { (acc, executionInfo) =>
```
geoffjentry: I'd love you even *more* than I already do if you were folding over a function, but you can limit yourself to your current level of jeff love if you'd prefer

geoffjentry: err, not over a function. you know what i mean :)

engine/src/main/scala/cromwell/engine/db/ExecutionInfosByExecution.scala
kshakir: Here is the original `DataAccess.updateExecutionInfo` and its new sibling `DataAccess.upsertExecutionInfo`:

```scala
  def updateExecutionInfo(workflowId: WorkflowId, callKey: BackendCallKey,
                          key: String, value: Option[String])
                         (implicit ec: ExecutionContext): Future[Unit]

  def upsertExecutionInfo(workflowId: WorkflowId, callKey: BackendCallKey,
                          keyValues: Map[String, Option[String]])
                         (implicit ec: ExecutionContext): Future[Unit]
```

The `Map` returned by this function is passed directly into the upsert.

kshakir:
> map WdlFile.apply might read more smoothly, here and below

@mcovarr I'm happy to go over to `wdl4s` and properly overload the `WdlFile.apply` such that this will work, if we want to do that fix now?

mcovarr:
@kshakir nah don't worry about the wdl4s bit, we have bigger fish to fry. :smile: 

:+1: as is

PR #623, opened 2016-03-25
NioGcsPath's toFile throws an exception :-/

scottfrazer:
Noticed when using `write_tsv` on JES recently

### DIFF
--- a/engine/src/main/scala/cromwell/engine/WdlStandardLibraryImpl.scala
+++ b/engine/src/main/scala/cromwell/engine/WdlStandardLibraryImpl.scala
@@ -156,7 +156,7 @@ trait WdlStandardLibraryImpl extends WdlStandardLibraryFunctions {
     // This may be called multiple times with the same inputs.  Calling this twice with the same
     // parameters should yield the same return value.
     val rootDir = path.toAbsolutePath(fileSystems)
-    if (!rootDir.exists) rootDir.toFile.mkdirs()
+    rootDir.createDirectories
 
     val fullPath = rootDir.resolve(s"$prefix${content.md5Sum}$suffix")
     if (!fullPath.exists) fullPath.write(content)


engine/src/main/scala/cromwell/engine/WdlStandardLibraryImpl.scala
```diff
@@ -156,7 +157,10 @@ trait WdlStandardLibraryImpl extends WdlStandardLibraryFunctions {
     // This may be called multiple times with the same inputs.  Calling this twice with the same
     // parameters should yield the same return value.
     val rootDir = path.toAbsolutePath(fileSystems)
-    if (!rootDir.exists) rootDir.toFile.mkdirs()
+    Try(rootDir.toFile) match {
+      case Success(javaFile) if !javaFile.exists => javaFile.mkdirs
+      case _ =>
+    }
```
scottfrazer: This is just a placeholder solution in case Jose needs it soon, but I'm not sure this is the best solution

kshakir: Use nio's [`Files.createDirectories(rootDir)`](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Files.html#createDirectories-java.nio.file.Path-java.nio.file.attribute.FileAttribute...-), or if a class is already using [`better.files`](https://github.com/pathikrit/better-files#better-files----), [`rootDir.createDirectories()`](https://github.com/pathikrit/better-files/blob/v2.13.0/core/src/main/scala/better/files/File.scala#L85).

In general, all usage of `java.io.File` should (probably) be removed from the cromwell ecosystem to ensure that our custom `java.nio.file.spi.FileSystemProvider` works.

scottfrazer: Okay I will change

cjllanwarne:
Any way we can test this? Centaur?

cjllanwarne:
Otherwise, :+1:

scottfrazer:
@cjllanwarne 

https://github.com/broadinstitute/tyburn/pull/30
https://github.com/broadinstitute/centaur/pull/29

Horneth:
Yes as @kshakir mentioned, `io.File` has the default OS filesystem hardcoded inside the class, which makes it impossible to use with GCS unfortunately..

Horneth:
:+1:

PR #624, opened 2016-03-25
Move where the kamon stuff is started to fix a race condition

geoffjentry:
I'm curious if travis succeeds, but if it does this is intended to resolve our tyburn woes


geoffjentry:
Ok - Travis succeeded. This *should* resolve the massive tyburn problems and it should now work about as well as it did previously

mcovarr:
if that's the case then :+1: :+1: :+1:...

cjllanwarne:
:+1:

geoffjentry:
@mcovarr there's only one way to know for sure, but ...

PR #625, opened 2016-03-28
Add retries on creating Input/Output Sreams for GCS

Horneth:
Adds retries on creation of Input / Output streams for GCS files.
It's not an "end-to-end" retry though as something might fail while reading/writing the stream that won't be caught by this retry.

### DIFF
--- a/engine/src/main/scala/cromwell/engine/backend/io/filesystem/gcs/GcsFileSystemProvider.scala
+++ b/engine/src/main/scala/cromwell/engine/backend/io/filesystem/gcs/GcsFileSystemProvider.scala
@@ -95,8 +95,7 @@ class GcsFileSystemProvider private (storageClient: Try[Storage], executionConte
     case gcsPath: NioGcsPath =>
       Try(withRetry(client.objects.get(gcsPath.bucket, gcsPath.objectName).execute)) recover {
         case ex: GoogleJsonResponseException
-          if ex.getStatusCode == 404 =>
-          if (!gcsPath.isDirectory) throw new FileNotFoundException(path.toString)
+          if ex.getStatusCode == 404 => if (!gcsPath.isDirectory) throw new FileNotFoundException(path.toString)
       } get
     case _ => throw new FileNotFoundException(path.toString)
   }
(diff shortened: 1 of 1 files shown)


engine/src/main/scala/cromwell/engine/backend/io/filesystem/gcs/GcsFileSystemProvider.scala
```diff
@@ -95,8 +95,7 @@ class GcsFileSystemProvider private (storageClient: Try[Storage], executionConte
     case gcsPath: NioGcsPath =>
       Try(withRetry(client.objects.get(gcsPath.bucket, gcsPath.objectName).execute)) recover {
         case ex: GoogleJsonResponseException
-          if ex.getStatusCode == 404 =>
-          if (!gcsPath.isDirectory) throw new FileNotFoundException(path.toString)
+          if ex.getStatusCode == 404 => if (!gcsPath.isDirectory) throw new FileNotFoundException(path.toString)
```
mcovarr: why not `&&` these two tests in the guard

mcovarr:
:+1:

mcovarr:
Per our post-standup, I'm not sure if this PR is supposed to have a "closes" on it or not.

Horneth:
I don't think it closes anything no. It should be enough for the release though ? We'll see if/where we need more retries in the logs if this error pops up again.

engine/src/main/scala/cromwell/engine/backend/io/filesystem/gcs/GcsFileSystemProvider.scala
Horneth: Because if `&&` is in the guard, the `case` will only match if gcsPath is *not* a directory.
So if it *is* a directory it will not recover and the `get` below will throw an exception, which we don't want (if it's a directory we need to always consider that it exists - even if doesn't make sense for GCS, as a lot of Nio operations check for file/directory existence before doing something)

mcovarr: OK I misunderstood, I thought it was a predicate but it's really an assertion, if the file does not exist the method throws.

Horneth: Yes, that's what the Nio stuff expects - I got confused too at first, I did as you suggested and had a bunch of random errors when manipulating directories

engine/src/main/scala/cromwell/engine/backend/io/filesystem/gcs/GcsFileSystemProvider.scala
```diff
@@ -183,13 +183,12 @@ class GcsFileSystemProvider private (storageClient: Try[Storage], executionConte
         }
 
         withRetry(moveInner)
-      case _ => throw new UnsupportedOperationException(s"Can only copy from GCS to GCS: $source or $target is not a GCS path")
+      case _ => throw new UnsupportedOperationException(s"Can only move from GCS to GCS: $source or $target is not a GCS path")
```
cjllanwarne: I didn't think we ever moved? I thought we always copied?

Horneth: We rename stdout/stderr/call logs files when copying them in call caching so they have the expected name in the new call context. Rename is considered a `move`

cjllanwarne:
:+1:

mcovarr:
Yes, as far as I'm concerned it's :+1: for the release.

PR #627, opened 2016-03-29
Bringing in BackendConfiguration, beginning changes to config file.  Closes #580

### LINKED ISSUE #580: Introduce BackendConfiguration like pluggable_backends.

mcovarr:
Warning: kind of a trainwreck.  This PR is aimed in the general direction of #580 but "closes" might be too strong a term for this.  

This has a DO NOT MERGE on it since it definitely shouldn't go into 0.19, and maybe also for the trainwreck comment above.

Naughty list:

- Might break GCS with Local and SGE backends
- Doesn't cleanly deal with filesystems, so configuration is currently duplicated among backends.  Blech.

Nice list:

- Does get the config closer to what it should look like for PBEs.
- Might be able to plug in a backend type that isn't compiled in.  Maybe.
- Might be able to do sneaky things like have the LocalBackend specified twice with two different names to test call-granular backends.  Again, maybe.


mcovarr:
@Horneth and @cjllanwarne are the reviewers, though you might want to let your lunches settle a bit first.

engine/src/main/resources/application.conf
```diff
@@ -68,85 +61,121 @@ call-caching {
...
+    },
+    {
+      name = "JES"
+      class = "cromwell.engine.backend.jes.JesBackend"
+      config {
+        // Google project
+        project = "my-cromwell-workflows"
+
+        // Base bucket for workflow executions
+        baseExecutionBucket = "gs://my-cromwell-workflows-bucket"
```
Horneth: I think this is the function equivalent of "root" in the two other backend configs, if we're going to break config retro-compatibility anyway maybe it's a good time to harmonize the name ?

geoffjentry:
@mcovarr regarding some of the badness, if this isn't going into 0.19 (and thus causing ppl to need to repeatedly modify config) I feel like the nice likely outweighs the naughty ... he says w/o having his lunch settled yet

engine/src/main/resources/application.conf
mcovarr: good idea

mcovarr:
@geoffjentry yes I'm fine with merging this after release (or the release branch is cut), barring any substantive problems turned up by eagle-eyed reviewers.

engine/src/main/resources/application.conf
```diff
@@ -68,85 +61,121 @@ call-caching {
...
+        endpointUrl = "https://genomics.googleapis.com/"
+
+        // Polling for completion backs-off gradually for slower-running jobs.
+        // This is the maximum polling interval (in seconds):
+        maximumPollingInterval = 600
+
+        applicationName = "cromwell"
+
+        // Optional Dockerhub Credentials. Can be used to access private docker images.
+        docker {
```
Horneth: Is there any advantage in putting this at the root of the config ? Only JES supports it now but other backends might want to support it too. Although leaving it here each backend could have different docker credentials I guess..
Also we might want to rename it `dockerHub`

mcovarr: I like `dockerHub`, then maybe the values within can be just `account` and `token`?

I moved this largely because it needed to be visible for the JesBackend config parsing, but given that this is only supported by JesBackend it didn't seem like it should be global anyway?  I don't feel too strongly about this and could be convinced either way.

Horneth: Yeah I'm not too sure either, let's let it here for now I guess as only JES supports it for now anyway as you said.
:+1:  to `account` and `token`

geoffjentry: We'll probably see a future where we want this elsewhere but I wouldn't worry about it for now

mcovarr:
Also I have not updated the README whatsoever

engine/src/main/resources/application.conf
mcovarr: Fun fact: the code was already changed over to use `account` and `token`, this config was never updated for the switch :frowning:

mcovarr: in some spots...

Horneth:
I was trying to find a way to re-arrange the conf wrt file systems and came up with this
https://gist.github.com/Horneth/42d5656afced5795f0d2cafe3e9ef0f2

engine/src/main/resources/application.conf
```diff
@@ -72,85 +65,121 @@ call-caching {
...
-  //   client_id = ""
-  //   client_secret = ""
-  // }
+backend {
+  default = "Local"
+  providers = [
+    {
+      name = "Local"
+      class = "cromwell.engine.backend.local.LocalBackend"
+      config {
```
francares: I think I commented this before in other pull request but this should not be config. Instead of that it should be filesystem or IO or something descriptive about FS.

engine/src/main/resources/application.conf
```diff
@@ -72,85 +65,121 @@ call-caching {
...
+      class = "cromwell.engine.backend.jes.JesBackend"
+      config {
+        // Google project
+        project = "my-cromwell-workflows"
+
+        // Base bucket for workflow executions
+        root = "gs://my-cromwell-workflows-bucket"
+
+        // Endpoint for APIs, no reason to change this unless directed by Google.
+        endpointUrl = "https://genomics.googleapis.com/"
```
francares: This is specific of JES backend implementation and there is no need to generalized. I would recommend to put this in other section specific for JES backend implementation.

engine/src/main/scala/cromwell/engine/backend/BackendConfiguration.scala
```diff
@@ -0,0 +1,25 @@
...
+
+import scala.collection.JavaConverters._
+
+/**
+  * Defines a backend configuration.
+  *
+  * @param name      Backend name.
+  * @param className Implementation class for the specific backend.
+  */
+case class BackendConfigurationEntry(name: String, className: String, config: Config)
```
francares: I think BackendConfigurationEntry is already a representation of configuration... why are you defining again config?

engine/src/main/scala/cromwell/engine/backend/local/LocalBackend.scala
```diff
@@ -58,6 +57,28 @@ object LocalBackend {
     def script = jobDescriptor.callRootPath.resolve("script")
     def returnCode = jobDescriptor.callRootPath.resolve("rc")
   }
+}
+
+
+/**
+ * Handles both local Docker runs as well as local direct command line executions.
+ */
+case class LocalBackend(config: Config, actorSystem: ActorSystem) extends Backend with SharedFileSystem {
```
francares: I think Local backend should take care for it's own way of defining configuration and we should not standarize the configuration for all backends. I.e.: Getting configuration from companion object.

engine/src/main/resources/application.conf
Horneth: Hmm isn't this precisely the place to put backend specific configuration ?

mcovarr:
Some more hypothetical confs after discussion with @Horneth:

- [Analyst](https://gist.github.com/mcovarr/a2d23c79fcd694093597948222ff2440) using Local backend with local filesystem and GCS.
- [GotC](https://gist.github.com/mcovarr/e920d24840ae1788cd0aeb69dd11b3d9) using the JES backend for most things and a Local backend for final calls (@kshakir's forthcoming work)
- [FireCloud](https://gist.github.com/mcovarr/53d48d0008727a961dfeca164ed57fd8) with JES + Local backends, two auth modes.

engine/src/main/resources/application.conf
mcovarr: I've changed this around in the latest proposed version, check out the three Analyst / GotC / FireCloud confs that have been linked.

engine/src/main/resources/application.conf
mcovarr: There wouldn't be another section for JES configuration; this is it.

engine/src/main/scala/cromwell/engine/backend/BackendConfiguration.scala
mcovarr: This `config` is only the chunk of the configuration specific to this backend instance.

francares: OK. I didn't realize you are using Typesafe config.

engine/src/main/scala/cromwell/engine/backend/BackendConfiguration.scala
```diff
@@ -0,0 +1,25 @@
...
+
+import com.typesafe.config.{Config, ConfigFactory}
+
+import scala.collection.JavaConverters._
+
+/**
+  * Defines a backend configuration.
+  *
+  * @param name      Backend name.
+  * @param className Implementation class for the specific backend.
```
francares: @param config entry is missing.

engine/src/main/resources/application.conf
francares: OK.

engine/src/main/resources/application.conf
francares: I didn't see you were using typesafe config. Looks OK.

README.md
```diff
@@ -291,7 +291,7 @@ backend {
 }
 ```
 
-Or, via `-Dbackend.abortJobsOnTerminate=true` command line option.
+Or, via `-Dserver.abortJobsOnTerminate=true` command line option.
```
cjllanwarne: Ironically, I suspect that `run` mode is far more likely to use this than `server`!

engine/src/main/resources/application.conf
```diff
@@ -72,85 +65,121 @@ call-caching {
...
+  default = "Local"
+  providers = [
+    {
+      name = "Local"
+      class = "cromwell.engine.backend.local.LocalBackend"
+      config {
+        // Root directory where Cromwell writes job results.  This directory must be
+        // visible and writeable by the Cromwell process as well as the jobs that Cromwell
+        // launches.
+        root: "cromwell-executions"
```
cjllanwarne: So reading this description a question jumped up at me - this directory relative to... where?

cjllanwarne:
@mcovarr I like your example configs, excepting the (hopefully) upcoming backend/filesystem split. Are they represented in the code as it is now?

Horneth:
@mcovarr It's more concise but it makes several implicit assumptions to fit exactly our 2 current JES use cases (GotC and FireCloud).
IMO it should be flexible enough so that GotC and Firecloud are just combinations of configuration entries among all the possible.

> // The JES backend is assumed to use the GCS filesystem with user authentication, dropping back to Cromwell
        // authentication if user authentication is not defined.  In GotC there is no user authentication, so
        // Cromwell authentication it is.

This is tailor made to accommodate GotC and FireCloud with as few configuration changes as possible, but I think we should move towards something more generic, even if the conf looks a bit more different between the two.
For example what if you want to use refresh token auth mode for creating the pipeline as well ?

README.md
mcovarr: Hmm okay.  Can you think of a better choice than `server` that encompasses both?

engine/src/main/resources/application.conf
mcovarr: Currently, the current directory for wherever Cromwell is started.  This is not a change from the current behavior.

mcovarr:
@Horneth @cjllanwarne sure we can make the choice of auths explicit in `genomics` and `filesystems`.  I did want to keep it clear in the config which auth was for Cromwell and which was for the user so we didn't make it impossible to implement call log copying in FireCloud (in case someday we want to use that feature there).  How about something like the following for FireCloud:

```hocon

// Same as the preceding FireCloud sample conf
google {
  application-name = "cromwell"

  // There may be instances like the final call that copies call logs which will need to be able to generate both
  // Cromwell and user authentication, so making these explicit.
  cromwellAuthentication {
    scheme = "application_default"
  }

  // Used for engine functions involving the filesystem.
  userAuthentication {
    scheme = "refresh"
    client-id = "secret_id"
    client-secret = "secret_secret"
  }
}

// genomics with explicitly selected conf
genomics {
  ...
  auth = "cromwell"
  ...
}

...
  filesystems = [
    // gcs filesystem with explicitly selected conf
    gcs {
      auth = "user"
    }
  ]
...


```

Horneth:
Actually what is bugging me I think is precisely the distinction between "cromwell" and "user" auth I think :smile: 
I think we called them like this because it makes sense in the current way we use them. We use "cromwell" auth pretty much always with service-account and "user" auth with a "refreh token mode"  (falling back to the "cromwell" auth if there's no "user" auth).

But really what we need is an auth for pipelines and another one for gcs (they might be the same or not, use service account or refresh tokens, but we shouldn't care from a cromwell perspective).

mcovarr:
@Horneth gotcha, I'll work up another sample.  :smile:

Horneth:
I'm thinking having more a "google auth template" in the conf, with all the auth mode supported by cromwell.
And each section of the conf that needs google authentication (JES genomics, gcs filesystem,...) would provide the auth scheme that it wants to use for this particular feature.

cjllanwarne:
@mcovarr @Horneth I haven't really followed the discussion but I'd be worried about having "GOOGLE" auth as a top-level item.
I'd much rather see this be general purpose, although I admit that relies on fliesystems and backends actually being separated...

cjllanwarne:
i.e. something like:
```
auths {
  googleauth1: {
    class: "path.to.googleauth"
    otherstuff: "blah"
  }
}

filesystems {
  fs1 {
    class: "path.to.gcs.filesystem"
    auth: googleauth1
  }
}

backends {
  googlepipelineservice {
    class: "path.to.pipelineservicejar"
    gcsfilesystem: fs1
    auth: googleauth1
  }
}
```

mcovarr:
OK, after discussion with @Horneth and @cjllanwarne:

```hocon
google {
  application-name = "cromwell"

  auths = [
    {
      name = "cromwell-service-account"
      scheme = "application_default"
    },
    {
      name = "user-via-refresh"
      scheme = "refresh"
      client-id = "secret_id"
      client-secret = "secret_secret"
    }
  ]
}

backend {
  default = "JES"
  providers = [
    {
      name = "JES"
      class = "cromwell.engine.backend.jes.JesBackend"
      config {
        ...
        genomics {
          // Endpoint for APIs, no reason to change this unless directed by Google.
          endpoint-url = "https://genomics.googleapis.com/"
          // Use the Cromwell service account for creating the Pipeline and manipulating auth JSONs.
          auth = "cromwell-service-account"
        }
        filesystems = [
          {
            gcs {
              auth = "user-via-refresh"
            }
          }
        ]
      }
    },
    ...
```

mcovarr:
Since the code doesn't actually work like the proposed conf, I'm going to close this PR until all is in harmony again.

README.md
mcovarr: How about `system.abort-jobs-on-terminate`?  And yes I'm dashifying identifiers in here as long as I'm solving all the other problems.  :stuck_out_tongue:

PR #633, opened 2016-03-31
Split file systems Closes #555

### LINKED ISSUE #555: Decouple Filesystems from Backends
Filesystems and Backends are still too tightly coupled at the moment. Filesystems might even been extracted to their own subdirectory.

Horneth:
This needs https://github.com/broadinstitute/lenthall/pull/13 to compile


gcsfilesystem/src/main/scala/cromwell/gcsfilesystem/GoogleConfigurationAdapter.scala
```diff
@@ -1,34 +1,26 @@
...
-// Google Authentication modes supported for Cromwell
-sealed trait GoogleCromwellAuthMode
-final case class ServiceAccountMode(accountId: String, pemPath: String) extends GoogleCromwellAuthMode
-final case class UserMode(user: String, secretsFile: String, datastoreDir: String) extends GoogleCromwellAuthMode
-case object ApplicationDefaultMode extends GoogleCromwellAuthMode
-
-// Google Authentication modes supported for User
-sealed trait GoogleUserAuthMode
-final case class Refresh(clientSecrets: ClientSecrets) extends GoogleUserAuthMode
+final case class GoogleConfigurationAdapter(appName: String, cromwellAuthMode: GoogleAuthMode, userAuthMode: Option[GoogleAuthMode]) {
```
Horneth: This class is the old `GoogleConfiguration` and is there temporarily as a compatibility workaround, knowing that the config template is going to change soon anyway.

Horneth:
@cjllanwarne @mcovarr - also https://github.com/broadinstitute/lenthall/pull/13 pls

build.sbt
```diff
@@ -7,8 +7,11 @@ lazy val backend = (project in file("backend"))
   .dependsOn(core % "test->test;compile->compile")
   .settings(backendSettings:_*)
 
+lazy val gcsfilesystem = (project in file("gcsfilesystem"))
+  .settings(gcsFileSystemSettings:_*)
+
 lazy val engine = (project in file("engine"))
-  .dependsOn(core % "test->test;compile->compile", backend)
+  .dependsOn(core % "test->test;compile->compile", gcsfilesystem % "test->test;compile->compile", backend)
```
mcovarr: is the long-term plan to have only the backend project depend on gcsfilesystem?

Horneth: yes, currently backend project is empty and everything is in engine, but eventually this should probably move to be a backend-only dependency.

project/Dependencies.scala
```diff
@@ -45,6 +45,17 @@ object Dependencies {
...
 
+  val gcsFileSystemDependencies = List(
+    "com.google.cloud.bigdataoss" % "gcsio" % "1.4.4",
+    "com.google.api-client" % "google-api-client-java6" % googleClientApiV,
+    "com.google.api-client" % "google-api-client-jackson2" % googleClientApiV,
+    "com.google.oauth-client" % "google-oauth-client" % googleClientApiV,
+    "com.typesafe" % "config" % "1.3.0",
+    "org.broadinstitute" %% "lenthall" % lenthallV,
+    "org.scalaz" % "scalaz-core_2.11" % "7.1.3"
+//    "org.scalatest" %% "scalatest" % "2.2.5" % Test
```
mcovarr: ?

project/Testing.scala
```diff
@@ -17,6 +17,10 @@ object Testing {
   lazy val UseCromwellIntegrationTaggedTests = Tests.Argument("-n", CromwellIntegrationTestTag)
   lazy val DontUseCromwellIntegrationTaggedTests = Tests.Argument("-l", CromwellIntegrationTestTag)
 
+  lazy val GcsIntegrationTestTag = "GcsIntegrationTest"
+  lazy val UseGcsIntegrationTaggedTests = Tests.Argument("-n", CromwellIntegrationTestTag)
```
mcovarr: looks like a copy/paste fail on the tag

mcovarr:
:+1: looks good, though it's going to be quite the rebase on my branch :weary:

build.sbt
cjllanwarne: I don't understand that - I thought the whole point of this was that engine could do FS stuff without worrying about a backend?

engine/src/main/scala/cromwell/engine/backend/jes/JesConfiguration.scala
```diff
@@ -1,23 +1,27 @@
...
-  def googleConf: GoogleConfiguration
+  def genomicsConf: GoogleConfiguration
+  def gcsConf: GoogleConfiguration
   def dockerConf: Option[JesDockerCredentials]
 }
 
 object ProductionJesConfiguration {
   lazy val jesConf =  JesAttributes()
-  lazy val googleConf =  GoogleConfiguration.gcloudConf.get
+  private lazy val oldGoogleConf =  GoogleConfigurationAdapter.gcloudConf.get
```
cjllanwarne: uh, ??
Is anything still using this?

cjllanwarne: Oh, I see they still are. Is there any better way of naming it?

engine/src/main/scala/cromwell/engine/backend/sge/SgeBackend.scala
```diff
@@ -36,6 +36,9 @@ object SgeBackend {
     def script = jobDescriptor.callRootPath.resolve("script.sh")
     def returnCode = jobDescriptor.callRootPath.resolve("rc")
   }
+
+  private lazy val oldGoogleConf = GoogleConfigurationAdapter.gcloudConf.get
+  lazy val gcsConf = Try(GoogleConfiguration(oldGoogleConf.appName, oldGoogleConf.userAuthMode.getOrElse(oldGoogleConf.cromwellAuthMode)))
```
cjllanwarne: I'm not too keen on having copy/pasted google blocks in every backend. Is it possible to pass in the generic FS ref instead of having each backend forced to support each filesystem?

engine/src/test/scala/cromwell/util/docker/DockerHashableSpec.scala
```diff
@@ -45,12 +45,12 @@ class DockerHashableSpec extends FlatSpec with Matchers {
   it should "not create hashes for invalid hashables" in {
     val hashables = Table(
       ("hashable", "exceptionType", "exceptionMessage"),
-      (DockerRegistryImageId("bad_hash"), an[AggregatedException],
-        "hashString 'bad_hash' is not valid: contains illegal character for hexBinary: bad_hash"),
+      (DockerRegistryImageId("bad_hash"), an[CromwellAggregatedException],
```
cjllanwarne: I presume there's a possible `a[CromwellAggregatedException]` here? Otherwise readability has taken an unfortunate hit!

engine/src/test/scala/cromwell/util/docker/DockerHashableSpec.scala
```diff
@@ -45,12 +45,12 @@ class DockerHashableSpec extends FlatSpec with Matchers {
...
+        "\nhashString 'bad_hash' is not valid: contains illegal character for hexBinary: bad_hash"),
```
cjllanwarne: What's added this newline?

gcsfilesystem/src/test/scala/cromwell/gcsfilesystem/GoogleConfigurationSpec.scala
```diff
@@ -0,0 +1,161 @@
+package cromwell.gcsfilesystem
```
cjllanwarne: I was going to complain that the same `google` block is now used by both `engine` and `gcsfilesystem` but... if it's ok with @mcovarr then it's ok with me!

build.sbt
Horneth: Well, if the engine does need filesystem capabilities then yes, it can depend on this, but then it would probably have to depend on every possible filesystem available.
I think if we can avoid having the engine needing filesystems it'll be better

cjllanwarne: I agree it shouldn't require individual filesystems - but it will always need some sort of filesystem operations?

gcsfilesystem/src/test/scala/cromwell/gcsfilesystem/GoogleConfigurationSpec.scala
mcovarr: Not sure what you mean?  I didn't think there were config changes in this PR.  Lots of them in the one I'm working on though...

engine/src/main/scala/cromwell/engine/backend/sge/SgeBackend.scala
Horneth: Hmm I'm not sure I understand, every backend is supposed to be totally isolated in theory, so if several backends use the same features there is going to be some copy/paste at some point, unless we create ©MetaBackends ?

mcovarr: That is sort of how it will work per the proposals [here](https://github.com/broadinstitute/cromwell/pull/627#issuecomment-203502798).

gcsfilesystem/src/test/scala/cromwell/gcsfilesystem/GoogleConfigurationSpec.scala
Horneth: `engine` should not need any `google` related thing when PBE is finished

engine/src/main/scala/cromwell/engine/backend/jes/JesConfiguration.scala
Horneth: Yeah I didn't want to touch too much on the configuration parsing since @mcovarr is going to shuffle everything anyway. This is all temporary and when we have the final conf, `GoogleConfigurationAdapter` will disappear.
I can try to think of a better name though.

build.sbt
Horneth: Hmm what for ?

engine/src/main/scala/cromwell/engine/backend/sge/SgeBackend.scala
mcovarr: Oh I think I misunderstood what @cjllanwarne was saying. :frowning: 

Over on my branch, for the specific case of GcsFilesystemSupport I've pulled out a trait that both SharedFilesystemBackend and JesBackend mix in to keep things DRY.  I was thinking this could be just a utility trait that hangs out in the backend subproject and backend impls could mix it in if they want to support GCS.  Would that address your concern without introducing any structural badness?

```scala
trait CanUseGcsFilesystem { self: Backend =>

  protected def gcsFilesystem(options: WorkflowOptions): Option[FileSystem] = {

    def gcsFilesystemMustHaveAuth: Option[String] = {
      backendConfig.getConfigOption("filesystems.gcs") map {
        _.getStringOr("auth", throw new RuntimeException(s"No auth in filesystems.gcs for backend '$name.")) }
    }

    def gcsFilesystemAuthMustBeLegit(authName: String): GoogleAuthMode = {
      def prependName(errs: NonEmptyList[String]): String = s"For backend '$name': " + errs.list.mkString("\n")

      GoogleConfiguration.Instance.auth(authName) leftMap prependName valueOr { msg => throw new RuntimeException(msg) }
    }

    val gcsAuth: Option[GoogleAuthMode] = for {
      // Throw if there's a filesystems.gcs but no auth within that.
      authName <- gcsFilesystemMustHaveAuth
      // Throw if there's an auth named in this backend's `filesystems.gcs` which isn't in the global `google` list of auths.
      auth = gcsFilesystemAuthMustBeLegit(authName)
    } yield auth

    val maybeStorage = gcsAuth map { _.buildStorage(options) }
    maybeStorage map GcsFileSystemProvider.apply map { _.getFileSystem }
  }
}
```

cjllanwarne:
Hmm, @mcovarr @Horneth, given both of your responses I think I misunderstood what this PR is for...

Am I now right in thinking that it's doing this:
- Remove filesystem operations from individual backends so that any backend can use them
- Adding the GCS filesystem into the local and SGE backends ?

Horneth:
Yep, that's what it's doing, at the exception that currently all backends are still in the `engine` subproject, so currently `engine` depends on gcsfilesystem. Eventually individual backends that want to use gcs will have to list it as a dependency (or like @mcovarr suggested there could be a top-level backend gcs trait that backends could choose to use or not, although I'm not sure how this is going to work if we eventually have isolated independent backend jars).

cjllanwarne:
OK- :+1:

PR #641, opened 2016-04-05
A possible version of the Metadata Service in the ServiceRegistry

cjllanwarne:


engine/src/main/scala/cromwell/services/MetadataServiceActor.scala
```diff
@@ -0,0 +1,65 @@
...
+import cromwell.engine.{ExecutionEventEntry, WorkflowId}
+import cromwell.engine.backend.JobKey
+import cromwell.services.MetadataServiceActor._
+import cromwell.webservice.WorkflowMetadataResponse
+import org.joda.time.DateTime
+
+object MetadataServiceActor {
+
+  sealed trait MetadataServiceActorMessage
+  case class SetStartTime(jobKey: JobKey, startTime: DateTime)
```
mcovarr: workflow ID doesn't appear to be specified anywhere

engine/src/main/scala/cromwell/services/MetadataServiceActor.scala
```diff
@@ -0,0 +1,65 @@
...
+import cromwell.engine.workflow.WorkflowManagerActor.WorkflowMetadata
+import cromwell.engine.{ExecutionEventEntry, WorkflowId}
+import cromwell.engine.backend.JobKey
+import cromwell.services.MetadataServiceActor._
+import cromwell.webservice.WorkflowMetadataResponse
+import org.joda.time.DateTime
+
+object MetadataServiceActor {
+
+  sealed trait MetadataServiceActorMessage
```
mcovarr: probably stuff should extend this trait?  😛

engine/src/main/scala/cromwell/services/MetadataServiceActor.scala
```diff
@@ -0,0 +1,65 @@
...
+  case class SetStartTime(jobKey: JobKey, startTime: DateTime)
+  case class SetEndTime(jobKey: JobKey, time: DateTime)
+  case class AddExecutionEvent(jobKey: JobKey, executionEventEntry: ExecutionEventEntry)
+  /*
+  We'd have a "set" message for each of these:
```
mcovarr: a(nother) plea to try to do all this event-sourcing-ish as append-only rather than set

geoffjentry: Very strong +1

kcibul: Can you give an example/suggestion what you'd like to see (specifically, like instead of the set)

mcovarr: Something that makes clear that a value has changed at a specified point in time, without clobbering whatever current value might be set.  Something like:

```scala
sealed trait MetadataEvent[V] {
  def workflowId: WorkflowId
  def jobKey: JobKey
  def eventName: String
  def eventTime: DateTime
  def value: V
}

/** Every instance of this received by the MetadataServiceActor is simply appended to the DB table; previously existing rows
    for this workflowId + jobKey are not modified.  This retains the full record of what changed and when.  `submission`, `start`, and `end` might
    be derived from the appropriate rows with eventName = "WorkflowStatus".  */
final case class WorkflowStatusEvent(workflowId: WorkflowId, jobKey: JobKey, eventTime: DateTime, value: String) extends MetadataEvent[String] {
  override def eventName = "Workflow Status"
}

```

engine/src/main/scala/cromwell/services/MetadataServiceActor.scala
```diff
@@ -0,0 +1,65 @@
...
+                        backendStatus: Option[String],
+                        outputs: Option[Map[String, WdlValue]],
+                        start: Option[DateTime],
+                        end: Option[DateTime],
+                        jobId: Option[String],
+                        returnCode: Option[Int],
+                        shardIndex: Int,
+                        stdout: Option[WdlFile],
+                        stderr: Option[WdlFile],
+                        backendLogs: Option[Map[String, WdlFile]],
```
mcovarr: Did we decide how to store map types like this (JSON map)?

engine/src/main/scala/cromwell/services/MetadataServiceActor.scala
cjllanwarne: In JobKey? Otherwise oops!

engine/src/main/scala/cromwell/services/MetadataServiceActor.scala
cjllanwarne: I think this is a different model - I'll see if I can mock it up in a separate PR so we can see them side by side

engine/src/main/scala/cromwell/services/MetadataServiceActor.scala
cjllanwarne: In this scheme, you'd have an `AddBackendLog(wfId, jobKey, key, value)` message specifically for appending another backend log (c.f. `case class AddExecutionEvent(jobKey: JobKey, executionEventEntry: ExecutionEventEntry)` above)

cjllanwarne:
Closed so it doesn't distract from M'sMM

PR #642, opened 2016-04-06
MySQL transaction rollback detection and retries.  Better upsert support

scottfrazer:


engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -43,6 +42,7 @@ object SlickDataAccess {
   private lazy val rootDatabaseConfig = rootConfig.getConfig("database")
   private lazy val databaseConfigName = rootDatabaseConfig.getStringOption("config")
   lazy val defaultDatabaseConfig = databaseConfigName.map(getDatabaseConfig).getOrElse(rootDatabaseConfig)
+  val actorSystem = ActorSystem("slick-retry-scheduler")
```
scottfrazer: Before you flog me... I wasn't sure what to do about this.  In order to use `akka.pattern.after` in [line 209](https://github.com/broadinstitute/cromwell/pull/642/files#diff-e3d71de750517c4e29193d05735f0cdcR209), I need a scheduler and I wasn't able to create a scheduler manually, it seemed like the way to get one was from an `ActorSystem`, so I just created this one here just so I could get this working.  It was non-obvious to me how to get the main `ActorSystem` piped in here because currently.  I could use `CromwellServer.actorSystem` instead, but again, I'm not sure.

engine/src/test/scala/cromwell/engine/db/slick/SlickDataAccessSpec.scala
```diff
@@ -113,6 +113,79 @@ class SlickDataAccessSpec extends FlatSpec with Matchers with ScalaFutures with
...
+                  """.stripMargin
+
+  val table = Table(
+    ("db", "dbConfig"),
+    ("mysql", mysql),
+    ("hsqldb", hsqldb)
+  )
+
+  forAll(table) { (db, dbConfig) =>
+    it should s"not deadlock with upserts in $db" in {
```
scottfrazer: I need to move this to `databaseWithConfig`

scottfrazer:
@kshakir and @Horneth

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
mcovarr: I would think we want to share the ActorSystem; might it be possible to pass it as an implicit where required?

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -190,9 +194,25 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
   }
 
   // Run action with an outer transaction
-  private def runTransaction[R](action: DBIOAction[R, _ <: NoStream, _ <: Effect]): Future[R] = {
+  private def runTransaction[R](action: DBIOAction[R, _ <: NoStream, _ <: Effect])(implicit ec: ExecutionContext): Future[R] = {
```
kshakir: There are too many `ExecutionContext` flying around. Do you mind making this not-`implicit` and pass it in explicitly where needed?

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
kshakir: Ask for the actor system in your method parameters. I could even sign off on a patch that implements two different `runTransaction` methods for this patch. 1) a new one that retries by scheduling a message to come back over the actor system, 2) the original that just tries once and fails. The former would only be used by the upsert methods for now, the latter by everything else.

For the love of all that's good, please don't use `CromwellServer.actorSystem`. That singleton needs go away asap or undergo refactoring as it's not even supposed to be around during the current 'cromwell run my.wdl'.

geoffjentry: Right, that singleton was a temporary hack for something which no longer matters and we shouldn't continue to feed it. Also agree that there shouldn't be a new system created here

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
scottfrazer: not only do I not mind, but I'd much prefer it!

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -181,16 +180,39 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
...
   override def close(): Unit = {
     actionThreadPool.shutdown()
     database.close()
   }
 
-  // Run action with an outer transaction
-  private def runTransaction[R](action: DBIOAction[R, _ <: NoStream, _ <: Effect]): Future[R] = {
+  private def runTransactionWithRollbackRetry[R](action: DBIOAction[R, _ <: NoStream, _ <: Effect], actorSystem: ActorSystem): Future[R] = {
+    implicit val ec = actorSystem.dispatcher
+    def retry[T](f: => Future[T], delay: FiniteDuration, retries: Int): Future[T] = {
```
kshakir: Regarding our chat on an explicit execution context: I would also be ok if you swapped the above two lines.

The swap would then scope the `implict ec` **_only_** within the `def retry {}` block. I don't want any implicit execution contexts floating around within the scope of the wider `def runTransactionWithRollbackRetry {}` block.

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -181,15 +180,38 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
...
+
+  private lazy val useSlickUpserts = dataAccess.driver.capabilities.contains(JdbcProfile.capabilities.insertOrUpdate)
 
   override def close(): Unit = {
     actionThreadPool.shutdown()
     database.close()
   }
 
-  // Run action with an outer transaction
+  private def runTransactionWithRollbackRetry[R](action: DBIOAction[R, _ <: NoStream, _ <: Effect], actorSystem: ActorSystem): Future[R] = {
```
scottfrazer: @kshakir is this okay for `runTransactionWithRollbackRetry` and `runTransaction`?

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -181,15 +180,38 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
...
+    def retry[T](f: => Future[T], delay: FiniteDuration, retries: Int): Future[T] = {
+      implicit val ec = actorSystem.dispatcher
+      f recoverWith {
+        case e: MySQLTransactionRollbackException if retries > 0 =>
```
mcovarr: Can this be less explicitly MySQL flavored and operate on a `SQLTransactionRollbackException` instead?

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -181,15 +180,38 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
...
+            *  prepared to re-issue a transaction if it gets rolled back because of a deadlock."
+            */
+          println(s"MySQL Transaction rollback detected.  Retrying in $delay ($retries retries remaining)")
+          after(delay, actorSystem.scheduler)(retry(f, delay, retries - 1))
+      }
+    }
+
+    def queryAttempt = database.run(action.transactionally)
+
+    Future(Await.result(retry(queryAttempt, delay=200 millis, retries=10), Duration.Inf))(actionExecutionContext)
```
mcovarr: why is there a Future wrapping an Await.result?  Why not just return `retry(...)`?

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
mcovarr: looks like this can be `action: DBIO[R]`, at least when `runTransaction` below is changed similarly

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
scottfrazer: I think I had it that way at some point, then I had some reason for changing it back to `MySQLTransactionRollbackException`... though I can't think of the reason for it

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
scottfrazer: If I understand correctly, the reason is that the outer `Future` is using `actionExecutionContext`, which fixed the issue we had a while ago about the thread pool starving.  Each `retry(...)` attempt is using the `ActorSystem`'s execution context / scheduler.

I believe this code is the same structure as the code in `runTransaction`

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -181,16 +179,39 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
...
+  private lazy val useSlickUpserts = dataAccess.driver.capabilities.contains(JdbcProfile.capabilities.insertOrUpdate)
 
   override def close(): Unit = {
     actionThreadPool.shutdown()
     database.close()
   }
 
-  // Run action with an outer transaction
-  private def runTransaction[R](action: DBIOAction[R, _ <: NoStream, _ <: Effect]): Future[R] = {
+  private def runTransactionWithRollbackRetry[R](action: DBIO[R], actorSystem: ActorSystem): Future[R] = {
```
Horneth: Just to make sure I understand, the only reason we pass an actor system through is to have a `actorSystem.scheduler` on line 211 to schedule retries right ?

scottfrazer: Currently, yes.  I suppose the idea is that we can also use `actorSystem.dispatcher` instead of passing in `ExecutionContext` into the other methods on `DataAccess`.

Horneth: Ok, it just seemed a bit odd to inject akka stuff into slickDataAccess only to use a scheduler, but since we're probably going to actorify a lot of DB IO it's going to the right direction

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
Horneth: The `Await.result` is going to block until the retry succeeds or failed though, which might take up to 10 retries ?
Is there a way to / would it still work by retrying asynchronously ? Maybe if you could do something like `retry(runTransaction(...))` instead

Horneth: Hmm we probably do want it to block actually, so disregard that, I was just thinking that 10 retries might take a long time, if for some reason they all fail, until we finally fail the action, but maybe not..

scottfrazer: Yeah, I see what you mean.  The code does do `Await.result(retry(queryAttempt, delay=200 millis, retries=10))` so it does seem that in the worst case it will be waiting 2 seconds if all 10 retries happen.  And that whole time it's waiting, other queries are not running in that slot.  And actually, if my understanding is correct, then the retries as I have them are really no better than `Thread.sleep()`.

I guess an alternative would be to yield that slot to other queries if a deadlock occurs?  Though I'm not sure if that fixes anything because wouldn't the caller of `upsertExecutionInfo()` just have to wait longer for the results?

I don't know...  I feel like I'm not understanding something.  Maybe this is good to talk over with @kshakir

kshakir: Good catch Miguel! Sorry Scott. More changes that I missed, again-

In short, I concur that this composite future is inverted.

The whole `f()` to be scheduled for retry should be `DataAccess.upsertWhatever()`. 

If it's easier to see in code, I'll mock up something for discussion as soon as I'm off mobile. I'm roughly picturing a generic method on DataAccess allowing restart on any of the DA methods.

kshakir: #690

scottfrazer:
I filed [a ticket](https://github.com/broadinstitute/cromwell/issues/693) to try to address the retry functionality more completely, but I will not implement it as part of this PR.

@Horneth and @kshakir: I could use some thumbs

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -181,16 +177,26 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
...
-  // Run action with an outer transaction
-  private def runTransaction[R](action: DBIOAction[R, _ <: NoStream, _ <: Effect]): Future[R] = {
+  override protected def isTransient(throwable: Throwable): Boolean = {
+    throwable match {
+      case e: SQLTransientException => true
+      case _ => false
+    }
+  }
+
+  private def runTransaction[R](action: DBIO[R]): Future[R] = {
```
kshakir: Thanks for fixing this parameter type. :)

kshakir:
👍

mcovarr:
👍

PR #643, opened 2016-04-06
Miguel's Metadata Masterplan

cjllanwarne:
Warning: May not be 100% representative of Miguel's Metadata Masterplan

### DIFF
--- a/engine/src/main/scala/cromwell/services/MetadataServiceActor.scala
+++ b/engine/src/main/scala/cromwell/services/MetadataServiceActor.scala
@@ -7,13 +7,32 @@ import cromwell.engine.backend.JobKey
 import cromwell.services.MetadataServiceActor._
 import cromwell.webservice.WorkflowMetadataResponse
 import org.joda.time.DateTime
+import wdl4s.values.WdlValue
 
 object MetadataServiceActor {
 
-  sealed trait MetadataServiceActorMessage
-  case class SetStartTime(jobKey: JobKey, startTime: DateTime)
-  case class SetEndTime(jobKey: JobKey, time: DateTime)
-  case class AddExecutionEvent(jobKey: JobKey, executionEventEntry: ExecutionEventEntry)
+  sealed trait MetadataEvent[V] {
+    def workflowId: WorkflowId
+    def jobKey: JobKey
+    def eventName: String
+    def eventTime: DateTime
+    def value: V
+  }
+
+  /** Every instance of this received by the MetadataServiceActor is simply appended to the DB table; previously existing rows
+    * for this workflowId + jobKey are not modified.  This retains the full record of what changed and when.  `submission`, `start`, and `end` might
+    * be derived from the appropriate rows with eventName = "Workflow Status".  */
+  final case class WorkflowStatusEvent(workflowId: WorkflowId, jobKey: JobKey, eventTime: DateTime, value: String) extends MetadataEvent[String] {
+    override def eventName = "Workflow Status"
+  }
+
+  /**
+    * Note the additional fqn argument in this one:
+    */
+  final case class WorkflowOutputGenerated(workflowId: WorkflowId, jobKey: JobKey, eventTime: DateTime, fqn: String, value: WdlValue) extends MetadataEvent[WdlValue] {
+    override def eventName = "Output"
+  }
+
   /*
   We'd have a "set" message for each of these:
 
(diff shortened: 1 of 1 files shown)


engine/src/main/scala/cromwell/services/MetadataServiceActor.scala
```diff
@@ -0,0 +1,89 @@
...
+import cromwell.engine.{ExecutionEventEntry, WorkflowId}
+import cromwell.engine.backend.JobKey
+import cromwell.services.MetadataServiceActor._
+import cromwell.webservice.WorkflowMetadataResponse
+import org.joda.time.DateTime
+import wdl4s.values.WdlValue
+
+object MetadataServiceActor {
+
+  sealed trait MetadataEvent[V] {
```
cjllanwarne: Move this somewhere that other actors can receive them
Make it unsealed

geoffjentry: Can you explain why other actors would be receiving these?

engine/src/main/scala/cromwell/services/MetadataServiceActor.scala
```diff
@@ -0,0 +1,89 @@
...
+
+  /**
+    * Note the additional fqn argument in this one:
+    */
+  final case class WorkflowOutputGenerated(workflowId: WorkflowId, jobKey: JobKey, eventTime: DateTime, fqn: String, value: WdlValue) extends MetadataEvent[WdlValue] {
+    override def eventName = "Output"
+  }
+
+  /*
+  We'd have a "set" message for each of these:
```
geoffjentry: Is this still true within Miguel's Master Magnificence?

engine/src/main/scala/cromwell/services/MetadataServiceActor.scala
mcovarr: The metadata service

geoffjentry: Isn't this in the metadata service though? I'm still trying to understand how this whole thing is working.

mcovarr: Oops I committed a halfway written comment.  I think this is mostly here for metadata incoming to the metadata service.  I don't think we have a lot of detail how the other usages are to work.

engine/src/main/scala/cromwell/services/MetadataServiceActor.scala
mcovarr: I hope not, appends only if possible

engine/src/main/scala/cromwell/services/MetadataServiceActor.scala
```diff
@@ -0,0 +1,89 @@
...
+                        runtimeAttributes: Map[String, String],
+                        preemptible: Option[Boolean],
+                        failures: Option[Seq[FailureEventEntry]])
+   */
+
+}
+
+class MetadataServiceActor extends Actor {
+
+  def receive = {
```
geoffjentry: Another indicator that I don't really understand how this is intended to be working. Why wouldn't the metadata service just be handling generic requests? Why does it have special knowledge of workflows and such?

engine/src/main/scala/cromwell/services/MetadataServiceActor.scala
geoffjentry: It certainly seemed the opposite of your masterification

engine/src/main/scala/cromwell/services/MetadataServiceActor.scala
geoffjentry: I'd assumed this service was just taking arbitrary metadata values from any requester and the ability to pull them back out, leaving responsibility for understanding the semantics of anything in there to someone else. Is that not how it's been drawn up?

engine/src/main/scala/cromwell/services/MetadataServiceActor.scala
```diff
@@ -0,0 +1,89 @@
...
+import cromwell.services.MetadataServiceActor._
+import cromwell.webservice.WorkflowMetadataResponse
+import org.joda.time.DateTime
+import wdl4s.values.WdlValue
+
+object MetadataServiceActor {
+
+  sealed trait MetadataEvent[V] {
+    def workflowId: WorkflowId
+    def jobKey: JobKey
```
geoffjentry: What if engine wanted to store workflow level metadata in here, not for a specific job?

engine/src/main/scala/cromwell/services/MetadataServiceActor.scala
cjllanwarne: As I envisioned it, the simplest thing we can do would be a fully-workflow-aware metadata service which takes exactly the right data and stores them in exactly fitting tables (i.e. the already-existing ones). With all the problems which that entails, but it's an easy first-step and it's a service so we can replace it later if necessary

engine/src/main/scala/cromwell/services/MetadataServiceActor.scala
geoffjentry: disregard above. My brief dalliance with worrying about scoping these services to backend calls is on hiatus

geoffjentry:
@cjllanwarne Does this still need to be a PR?

PR #677, opened 2016-04-06
WorkflowBackendActor trait and some package classes. Closes #550

### LINKED ISSUE #550: [LCM] Backend interface definition

cjllanwarne:


backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
```diff
@@ -0,0 +1,83 @@
...
+object WorkflowBackendActor {
+
+  // Commands
+  sealed trait WorkflowBackendActorMessage
+
+  case class Execute(jobDescriptor: BackendJobDescriptor) extends WorkflowBackendActorMessage
+  case class Recover(jobDescriptor: BackendJobDescriptor) extends WorkflowBackendActorMessage
+  case class Abort(jobDescriptor: BackendJobDescriptor) extends WorkflowBackendActorMessage
+
+  // Events
```
mcovarr: // Responses

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
```diff
@@ -0,0 +1,83 @@
...
+  case class AbortFailed(jobDescriptor: BackendJobDescriptor, throwable: Throwable) extends AbortResponse
+}
+
+/**
+  * Workflow-level actor for executing, recovering and aborting jobs.
+  */
+trait WorkflowBackendActor extends Actor with ActorLogging {
+
+  implicit def ec: ExecutionContext
+  def workflowDescriptor: BackendWorkflowDescriptor
```
mcovarr: these two could/should be `protected`

cjllanwarne: Do we also want the service registry here?

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
```diff
@@ -0,0 +1,83 @@
...
+  private def abortAndReply(abortFunction: BackendJobDescriptor => Future[AbortResponse],
+                            jobDescriptor: BackendJobDescriptor,
+                            sndr: ActorRef = sender) = {
+    abortFunction(jobDescriptor) onComplete {
+      case Success(abortResponse: AbortResponse) =>
+        sndr ! abortResponse
+      case Failure(exception: Throwable) =>
+        sndr ! AbortFailed(jobDescriptor, exception)
+    }
+  }
```
mcovarr: slightly DRYer version of the above

```scala
  def receive: Receive = LoggingReceive {
    case Execute(jobDescriptor) => respond(execute, onFailure = ExecutionFailed, jobDescriptor)
    case Recover(jobDescriptor) => respond(recover, onFailure = ExecutionFailed, jobDescriptor)
    case Abort(jobDescriptor) => respond(abort, onFailure = AbortFailed, jobDescriptor)
  }

  private def respond(operation: BackendJobDescriptor => Future[WorkflowBackendActorResponse],
                      onFailure: (BackendJobDescriptor, Throwable) => WorkflowBackendActorResponse,
                      jobDescriptor: BackendJobDescriptor, sndr: ActorRef = sender()) = {
    operation(jobDescriptor) onComplete {
      case Success(r) => sndr ! r
      case Failure(t) => sndr ! onFailure(jobDescriptor, t)
    }
  }
```

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
mcovarr: Eventually yes but that should probably be left to whoever is the lucky recipient of that ticket.

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
```diff
@@ -0,0 +1,83 @@
...
+  /**
+    * Restart or resume a previously-started job.
+    */
+  def recover(jobDescriptor: BackendJobDescriptor): Future[ExecutionResponse]
+
+  /**
+    * Abort a running job.
+    */
+  def abort(jobDescriptor: BackendJobDescriptor): Future[AbortResponse]
+}
```
mcovarr: \n

backend/src/main/scala/cromwell/backend/package.scala
```diff
@@ -0,0 +1,36 @@
...
+
+package object backend {
+
+  trait BackendJobDescriptorKey {
+    def callFqn: String
+    def index: Option[Int]
+    def attempt: Int
+  }
+
+  // Can become a case class once Khalid's work simplifies the world so all calls run on the backend.
```
mcovarr: actually this lives in `cromwell.backend` and is not the `JobDescriptor` used in the rest of the Cromwell codebase, so this case-classification could happen now

mcovarr:
👍

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
cjllanwarne: I knew someone would feel compelled to improve that eventually... I didn't expect it so quickly!

cjllanwarne: Much DRYer!

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
```diff
@@ -0,0 +1,75 @@
...
+  case class Recover(jobDescriptor: BackendJobDescriptor) extends WorkflowBackendActorMessage
+  case class Abort(jobKey: BackendJobDescriptorKey) extends WorkflowBackendActorMessage
+
+  // Responses
+  sealed trait WorkflowBackendActorResponse
+
+  sealed trait ExecutionResponse extends WorkflowBackendActorResponse
+  case class ExecutionSucceeded(jobKey: BackendJobDescriptorKey, callOutputs: CallOutputs) extends ExecutionResponse
+  object ExecutionFailed {
+    def make(backendJobDescriptor: BackendJobDescriptor, throwable: Throwable) = ExecutionFailed(backendJobDescriptor.key, throwable)
```
mcovarr: not sure I see what this bit is giving us?

mcovarr: Oh this might be because the optimization I pasted before didn't quite work?  😄   You should be able to pass `ExecutionFailed.apply` instead of `ExecutionFailed.make` below, and then this object can go away.

Horneth: Looks like ExecutionFailed needs a jobKey but `performActionThenRespond` get a jobDescriptor so apply might not work

mcovarr: Ah okay, I (finally) get it.  Maybe something that makes this more clear in case others are slow like I was?  😄 

```scala
 private def descriptorToFailure: (BackendJobDescriptor, Throwable) => ExecutionFailed =
   (jd, t) => ExecutionFailed(jd.key, t)
```

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
```diff
@@ -0,0 +1,75 @@
...
+  def receive: Receive = LoggingReceive {
+    case Execute(jobDescriptor) => performActionThenRespond(execute, jobDescriptor, onFailure = ExecutionFailed.make)
+    case Recover(jobDescriptor) => performActionThenRespond(recover, jobDescriptor, onFailure = ExecutionFailed.make)
+    case Abort(jobKey)          => performActionThenRespond(abort, jobKey, onFailure = AbortFailed)
+  }
+
+  private def performActionThenRespond[A](operation: A => Future[WorkflowBackendActorResponse],
+                                          operationArgument: A,
+                                          onFailure: (A, Throwable) => WorkflowBackendActorResponse,
+                                          sndr: ActorRef = sender()) = {
```
scottfrazer: Why have `sndr` as a parameter if it's never anything besides `sender()`?

geoffjentry: So you don't close over sender. Wouldn't be a huge deal  if we were good about not having multiple computations at once in an actor but that's how we roll.

That said I really dislike how it is done here, seems unintuitive and potentially problematic - I'm leery that it'll run afoul of evaluation rules

backend/src/main/scala/cromwell/backend/package.scala
```diff
@@ -0,0 +1,31 @@
...
+
+  /**
+    * For uniquely identifying a job which has been or will be sent to the backend.
+    */
+  case class BackendJobDescriptorKey (callFqn: String, index: Option[Int], attempt: Int)
+
+  /**
+    * For passing to a BackendWorkflowActor at job execution, recovery or aborting time
+    */
+  case class BackendJobDescriptor(key: BackendJobDescriptorKey, task: Task, locallyQualifiedInputs: CallInputs)
```
scottfrazer: Any way that the second parameter can be a `Call` instead?  This will be especially useful if we ever use some of my proposed changes for wdl4s.  For example, the appropriate lookup function for instantiating commands and evaluating outputs would be a method on the `Call`.

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
scottfrazer: oh, right... I always forget about that.  Though you could just `val sndr = sender()`... but I guess if you're doing that, a default parameter is pretty much the same thing.  :shrug:

geoffjentry: Right, and that's how I'd rather see it :)

cjllanwarne: Would you prefer if it was an implicit :-p

geoffjentry: Don't tempt me, Llanwarne

cjllanwarne: I'd like to see a design where we don't need to store an evaluated sender as part of the closure - 
Presumably by forwarding it on directly to you backend execution actors?... but then you'd be implying implementation for the backends?

That said... I want to merge this ASAP so maybe we can revisit that in slower time :)

backend/src/main/scala/cromwell/backend/package.scala
cjllanwarne: I think that's also more correct. Especially given that the `BackendJobDescriptorKey` contains a `callFqn`...

backend/src/main/scala/cromwell/backend/package.scala
```diff
@@ -0,0 +1,31 @@
...
+
+  /**
+    * For uniquely identifying a job which has been or will be sent to the backend.
+    */
+  case class BackendJobDescriptorKey (callFqn: String, index: Option[Int], attempt: Int)
+
+  /**
+    * For passing to a BackendWorkflowActor at job execution, recovery or aborting time
+    */
+  case class BackendJobDescriptor(key: BackendJobDescriptorKey, call: Call, locallyQualifiedInputs: CallInputs)
```
cjllanwarne: I think this is wrong - the Engine cannot evaluate `locallyQualifiedInputs` - only the backend can

cjllanwarne: Since we're passing the input mappings in via the Call, maybe we just need to provide the backend with a lookup function?

backend/src/main/scala/cromwell/backend/package.scala
```diff
@@ -0,0 +1,38 @@
...
+  /**
+    * For uniquely identifying a symbol in the symbol map:Not
+    */
+  case class SymbolKey(fullyQualifiedName: FullyQualifiedName, index: Option[Int])
+
+  /**
+    * For passing to a BackendWorkflowActor at job execution, recovery or aborting time
+    */
+  //case class BackendJobDescriptor(key: BackendJobDescriptorKey, call: Call, locallyQualifiedInputs: CallInputs)
+  case class BackendJobDescriptor(key: BackendJobDescriptorKey, call: Call, symbolMap: Map[SymbolKey, WdlValue])
```
cjllanwarne: I think this works...?

backend/src/main/scala/cromwell/backend/package.scala
```diff
@@ -0,0 +1,40 @@
...
+    */
+  case class SymbolKey(fullyQualifiedName: FullyQualifiedName, index: Option[Int])
+
+  type BackendSymbolMap = Map[SymbolKey, WdlValue]
+
+  /**
+    * For passing to a BackendWorkflowActor for job execution or recovery
+    */
+  //case class BackendJobDescriptor(key: BackendJobDescriptorKey, call: Call, locallyQualifiedInputs: CallInputs)
+  case class BackendJobDescriptor(key: BackendJobDescriptorKey, call: Call, symbolMap: BackendSymbolMap)
```
cjllanwarne: Recommenting since my previous comment was lost. I *think* this is ok...?

Horneth: So we went back to passing evaluated wdl values to the backend ?

cjllanwarne: `call: Call` contains `inputs: Map[String, WdlExpression]` - but the `symbolMap` is now necessary to allow that to be evaluated

Horneth: Hmm ok, so symbolMap will contain all the current symbols for this workflow ?

cjllanwarne: Yes, that would be the idea

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
```diff
@@ -0,0 +1,77 @@
...
+  case class Execute(jobDescriptor: BackendJobDescriptor) extends WorkflowBackendActorMessage
+  case class Recover(jobDescriptor: BackendJobDescriptor) extends WorkflowBackendActorMessage
+  case class Abort(jobKey: BackendJobDescriptorKey) extends WorkflowBackendActorMessage
+
+  // Responses
+  sealed trait WorkflowBackendActorResponse
+
+  sealed trait ExecutionResponse extends WorkflowBackendActorResponse
+  case class ExecutionSucceeded(jobKey: BackendJobDescriptorKey, callOutputs: CallOutputs) extends ExecutionResponse
+  case class ExecutionFailed(jobKey: BackendJobDescriptorKey, throwable: Throwable) extends ExecutionResponse
```
francares: Since both exceptions contains the same, can we have something like this?
` ExecutionFailed(jobKey: BackendJobDescriptorKey, throwable: Throwable, retryable: Boolean = false) `

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
```diff
@@ -0,0 +1,77 @@
...
+  /**
+    * Execute a new job.
+    */
+  def execute(jobDescriptor: BackendJobDescriptor): Future[ExecutionResponse]
+
+  /**
+    * Restart or resume a previously-started job.
+    */
+  def recover(jobDescriptor: BackendJobDescriptor): Future[ExecutionResponse]
+
```
francares: I think preStart, postStop, beforeAll, afterAll, validate functions are missing here.

backend/src/main/scala/cromwell/backend/package.scala
```diff
@@ -0,0 +1,35 @@
+package cromwell
+
+import com.typesafe.config.Config
+import cromwell.core.{WorkflowId, WorkflowOptions}
+import wdl4s._
+import wdl4s.values.WdlValue
+
+package object backend {
```
francares: What happened with the idea of having a model package?

backend/src/main/scala/cromwell/backend/package.scala
```diff
@@ -0,0 +1,35 @@
...
+
+  /**
+    * For uniquely identifying a job which has been or will be sent to the backend.
+    */
+  case class BackendJobDescriptorKey (callFqn: String, index: Option[Int], attempt: Int)
```
francares: Remove extra space.

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
geoffjentry: This is inherently a sum type due to the boolean switch. IMO it is better to be explicit about this in the type system

backend/src/main/scala/cromwell/backend/package.scala
francares: I think this is something that may be complicated to a backend developer to understand. I also understand this is an intermediate step until we discuss engine functions so we can continue like this until we discuss engine functions.

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
```diff
@@ -0,0 +1,77 @@
...
+}
+
+/**
+  * Workflow-level actor for executing, recovering and aborting jobs.
+  */
+trait WorkflowBackendActor extends Actor with ActorLogging {
+
+  protected implicit def ec: ExecutionContext
+  protected def workflowDescriptor: BackendWorkflowDescriptor
+
```
francares: BackendConfigurationDescriptor is missing here.

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
cjllanwarne: The thinking is - if you want to implement these, just make a preStart yourself in the BE...

backend/src/main/scala/cromwell/backend/package.scala
cjllanwarne: Yeah I agree a BE implementer probably doesn't care about implementing backend engine functions themselves... but for now it's in here... :)

backend/src/main/scala/cromwell/backend/package.scala
```diff
@@ -0,0 +1,35 @@
...
+  //case class BackendJobDescriptor(key: BackendJobDescriptorKey, call: Call, locallyQualifiedInputs: CallInputs)
+  case class BackendJobDescriptor(key: BackendJobDescriptorKey, call: Call, symbolMap: BackendSymbolMap)
+
+  /**
+    * For passing to a BackendActor construction time
+    */
+  case class BackendWorkflowDescriptor(id: WorkflowId,
+                                       workflowOptions: WorkflowOptions,
+                                       workflowNamespace: NamespaceWithWorkflow,
+                                       declarations: Map[String, WdlExpression])
```
francares: Is there any difference between this `declarations` and the `workflowNamespace.workflow.declarations`?

cjllanwarne: Great point! I think this was supposed to be inputs

backend/src/main/scala/cromwell/backend/package.scala
cjllanwarne: I'm open to the idea... but if it's only to define the interface that Backend is presenting to the world, then I don't mind leaving it in backend. 

OTOH as I say, if there's a bunch of other stuff to throw in there, I'm open to the idea

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
scottfrazer: I :-1: without looking at the code... forget that... sorry.  I have no opinion.  I saw "sum type" and it was a reflex :)

cjllanwarne: Something to discuss or something to "it's not that important let's just pick one"?

geoffjentry: Keep it as is.

mcovarr: I think we should use the richer algebra like it is now

scottfrazer: not important to me.  I'm just objecting to the idea that all case classes with `Option` and `Boolean` would be better off as sum types

geoffjentry: But that's what they inherently are.

scottfrazer: lol...... how can I argue with assertions like that?!

cjllanwarne: It's fine if it multiples 1 into 2 cases. If there's 7 booleans then that's better than 2^7 case classes... IMO

geoffjentry: Yeah, I've never found a reasonable solution when there are more than 1 variable elements although it makes me shed a little tear every time

mcovarr: Real ADTs are great when not all of the cases are valid, just don't have types for those.

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
geoffjentry: I thought the idea was that the trait would have a validate method and a preStart which would call it? or did that change at some point?

PR #678, opened 2016-04-07
Workflow backend actor factory.

francares:
This is not the complete code. I'm waiting for dependencies: ShadowWorkflowActor and BackendConfiguration impl. to finish with it.

### DIFF
--- a/backend/src/main/scala/cromwell/backend/WorkflowBackendActorFactory.scala
+++ b/backend/src/main/scala/cromwell/backend/WorkflowBackendActorFactory.scala
@@ -1,0 +1,27 @@
+package cromwell.backend
+
+import akka.actor.{ActorRef, ActorSystem}
+import wdl4s.Call
+
+trait WorkflowBackendActorFactory {
+
+  /**
+    * Creates a workflow backend actor ref based on the init class, Cromwell actor system,
+    * backend workflow descriptor, backend configuration and Cromwell service registry.
+    *
+    * @param initClass                      Initialization class for the specific backend.
+    * @param actorSystem                    Cromwell Engine actor system.
+    * @param backendWorkflowDescriptor      Workflow level data needed to be able to execute a workflow.
+    * @param backendConfigurationDescriptor Backend configuration data.
+    * @param serviceRegistryActor           Cromwell service registry.
+    * @param calls                          A sequence of WDL Calls used to validate runtime requirements by the backend.
+    * @return An actor ref pointing to a WorkflowBackendActor implementation.
+    */
+  def getBackend(initClass: String,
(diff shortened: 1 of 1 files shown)


backend/src/main/scala/cromwell/backend/WorkflowBackendActorFactory.scala
```diff
@@ -0,0 +1,24 @@
...
+    * @param backendWorkflowDescriptor      Workflow level data needed to be able to execute a workflow.
+    * @param backendConfigurationDescriptor Backend configuration data.
+    * @param serviceRegistryActor           Cromwell service registry.
+    * @return An actor ref pointing to a WorkflowBackendActor implementation.
+    */
+  def getBackend(initClass: String,
+                 actorSystem: ActorSystem,
+                 backendWorkflowDescriptor: BackendWorkflowDescriptor,
+                 backendConfigurationDescriptor: BackendConfigurationDescriptor,
+                 serviceRegistryActor: ActorRef): ActorRef
```
cjllanwarne: Also the list of calls which this Backend will be asked to execute

francares: Right! I will add it.

francares:
I would like to merge this interface so I can continue with other task until all dependencies are satisfied.

@cjllanwarne and @scottfrazer Can you review this small piece of code?

cjllanwarne:
:+1:

backend/src/main/scala/cromwell/backend/WorkflowBackendActorFactory.scala
```diff
@@ -0,0 +1,27 @@
...
+    * @param serviceRegistryActor           Cromwell service registry.
+    * @param calls                          A sequence of WDL Calls used to validate runtime requirements by the backend.
+    * @return An actor ref pointing to a WorkflowBackendActor implementation.
+    */
+  def getBackend(initClass: String,
+                 actorSystem: ActorSystem,
+                 backendWorkflowDescriptor: BackendWorkflowDescriptor,
+                 backendConfigurationDescriptor: BackendConfigurationDescriptor,
+                 serviceRegistryActor: ActorRef,
+                 calls: Seq[Call]): ActorRef
```
scottfrazer: I don't think we need to pass in `calls`... We can just do `backendWorkflowDescriptor.workflowNamespace.workflow.calls`

scottfrazer: Oh wait, I think I'm wrong, forget that... these are a subset of the calls that will run on that backend.

scottfrazer:
:+1:

PR #680, opened 2016-04-07
Create a shadow WorklfowActor. Closes #679

### LINKED ISSUE #679: Create a shadow workflow actor
To exercise the new route of execution using pluggable backends, create a secondary workflow actor (currently with empty behavior).

gauravs90:


gauravs90:
@scottfrazer @cjllanwarne

engine/src/main/resources/application.conf
```diff
@@ -73,6 +73,10 @@ call-caching {
 }
 
 backend {
+  // Use the new route of execution starting ShadowWorkflowActor
+  // Once all features related to pluggable backends are in place, this will be deprecated
+  shadowExecutionEnabled = false
```
mcovarr: I wouldn't put this in `backend` stanza because a) I'm in the process of ripping it apart for `BackendConfiguration` and b) this is actually a directive for `WorkflowManagerActor`.  Can I suggest a `system` stanza instead?  I'm already relocating at least one displaced configuration option to that in my branch.

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -342,11 +342,16 @@ class WorkflowManagerActor(config: Config)
...
+
     val message  = MaterializeWorkflowDescriptorActor.MaterializeWorkflow(workflowId, source, config)
     val materializeWorkflowDescriptorActor = context.actorOf(MaterializeWorkflowDescriptorActor.props())
     materializeWorkflowDescriptorActor.askNoTimeout(message) map {
       case MaterializationSuccess(descriptor) =>
-        val wfActor = context.actorOf(WorkflowActor.props(descriptor), s"WorkflowActor-$workflowId")
+        val wfActor = shadowExecutionEnabled match {
+          case true => context.actorOf(ShadowWorkflowActor.props(), s"WorkflowActor-$workflowId")
+          case _ => context.actorOf(WorkflowActor.props(descriptor), s"WorkflowActor-$workflowId")
+        }
```
mcovarr: I'm assuming the plan is to refactor both the "classic" WorkflowActor as well as the new shadow WorkflowActor to manage their own validation since the classic/shadow distinction is more about interfacing with the backend workflow actors than validation?

If so this could pass the `WorkflowDescriptor` to the shadow `props` method to keep these two symmetric and the above can be simplified:

```scala
val buildProps = if (shadowExecutionEnabled) ShadowWorkflowActor.props else WorkflowActor.props
context.actorOf(buildProps(descriptor), s"WorkflowActor-$workflowId")
```

gauravs90: I was hoping its more more like adding the missing pieces to the ShadowWorkflowActor and when once ready, rename it to WorkflowActor. That ways we also make sure that it doesn't become as complex as it is today :-/

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
```diff
@@ -0,0 +1,19 @@
...
+import akka.actor.{Props, ActorLogging, Actor}
+
+object ShadowWorkflowActor {
+  def props(): Props = Props(new ShadowWorkflowActor)
+}
+
+/**
+  * Class that will [eventually] orchestrate a single workflow.
+  */
+class ShadowWorkflowActor extends Actor with ActorLogging{
```
Horneth: I know this is just the shadow's skeleton (this is starting to get kinda creepy :smile: ), but it'll be an FSM eventually right ?

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -342,11 +342,16 @@ class WorkflowManagerActor(config: Config)
...
 
+    val shadowExecutionEnabled = config.getBoolean("system.shadowExecutionEnabled")
+
     val message  = MaterializeWorkflowDescriptorActor.MaterializeWorkflow(workflowId, source, config)
     val materializeWorkflowDescriptorActor = context.actorOf(MaterializeWorkflowDescriptorActor.props())
     materializeWorkflowDescriptorActor.askNoTimeout(message) map {
       case MaterializationSuccess(descriptor) =>
-        val wfActor = context.actorOf(WorkflowActor.props(descriptor), s"WorkflowActor-$workflowId")
+        val wfActor = shadowExecutionEnabled match {
+          case true => context.actorOf(ShadowWorkflowActor.props(), s"WorkflowActor-$workflowId")
```
Horneth: Should we give it the wdl source ?

Horneth: Maybe the workflowId too ?

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -342,11 +342,16 @@ class WorkflowManagerActor(config: Config)
...
+          case _ => context.actorOf(WorkflowActor.props(descriptor), s"WorkflowActor-$workflowId")
+        }
         wfActor ! SubscribeTransitionCallBack(self)
         wfActor ! (if (isRestart) Restart else Start(replyTo))
```
Horneth: Also this might be problematic now, because before we assumed the WorfklowActor was immediately ready to process the start/restart message, now it's going to go through WorkflowDescriptor materialization + Backend validation before being ready

Horneth: Or maybe not if we do all validation in `preStart()`

Horneth: We might have to have a `isRestart` in the constructor too so it knows from the begging if it's a restart or not

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
cjllanwarne: I would have thought that having them operate totally different FSMs is expected and OK? Ideally, I would only want to change OldeStyleeWorkflowActor if tests fail or it won't compile.

IIRC, this descriptor is actually going to be `WdlSources` for the new actor?

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
cjllanwarne: I like the idea of telling a workflow what it's ID will be. Sources will certainly be passed in here, maybe part of the set-up FSM or Validation ticket?

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
cjllanwarne: Yup. I believe there's a "set it up as an FSM" ticket?

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
cjllanwarne: I personally think the WMA shouldn't care what state a WA is in. It should just send it messages. The WA should be the one who's responsible for making sure it acts on them... eventually.

Horneth: So the WA should be responsible for figuring out if it's being restarted or if it's fresh ?

cjllanwarne: It could still get a `Start` or `Restart`, but the WMA shouldn't need to ask "have you initialised yet" before sending the `Start` or `Restart` *message*

cjllanwarne:
This look correct to me :+1:

cjllanwarne:
@mcovarr or @Horneth for the second thumb?

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
Horneth: Right but then the WA needs to "remember" that it got a start/restart message using state data or something else, which feels strange. Why not give the information directly at construction time since we know it already.

Horneth:
If we don't want to figure out the constructor parameters yet in this PR, then :+1: from me

cjllanwarne:
NB I'm adding the FSM states for https://github.com/broadinstitute/cromwell/issues/683

PR #681, opened 2016-04-08
Backend stubs Closes #650

### LINKED ISSUE #650: StubOut Concrete backends

Horneth:


build.sbt
```diff
@@ -7,6 +7,18 @@ lazy val backend = (project in file("backend"))
   .dependsOn(core % "test->test;compile->compile")
   .settings(backendSettings:_*)
 
+lazy val localBackend = (project in file("local-backend"))
```
geoffjentry: Do we really want each of these backends to have a top level directory in the cromwell repo?

Horneth: I couldn't see any reason not to. They'll be aggregated in the global cromwell jar anyway, but they'll also have their own jar which will give us the ability to fix/update only one backend if necessary. Plus it helps validating that this one jar per backend model will work with future external backends

Horneth: Oh or do you mean still separate them but put the code in different backend subdirectories ?

geoffjentry: Right but as we add supported backends, if we want to keep them under the cromwell repo umbrella it'll really blossom out. I don't know how off the top of my head but I'm pretty sure you can have sub-structure in the sbt tree to still have the individual backend jars.

geoffjentry: @Horneth I think "yeah" to the second one but I'm not sure :) I meant to have some dir which holds the officially supported backends and then each supported backend has its own dir inside there

Horneth: > I meant to have some dir which holds the officially supported backends 

I was thinking that dir could be the current `backend` dir but that won't work, I'm gonna try with another `default-backends` or something

geoffjentry: Sweet. 

I think the SBT magic I was referring to could be handled by setting up whichever dir you set up as its own SBT project, but there might need to be some extra glue at the root to make it work right. Let me know if it's problematic and I can take a closer look.

Horneth:
@geoffjentry seems to work that way

Horneth:
Also @francares for review

project/Settings.scala
```diff
@@ -59,6 +59,24 @@ object Settings {
     assemblyJarName in assembly := name.value + "-" + version.value + ".jar"
   ) ++ commonSettings
 
+  val localBackendSettings = List(
```
cjllanwarne: OOI, how are these going to be added for "plugged" backends?

cjllanwarne:
OOI, how are these in-scala changes going to be added for each "plugged" backend?

cjllanwarne:
Otherwise, :+1:

Horneth:
@cjllanwarne  They're not - those are here because they are "cromwell default backends" and part of the cromwell jar / repo.
Other backends will need to append their jar in the classpath.

francares:
@Horneth Can you add HTCondor as well?

francares:
@Horneth Can we use commons instead of defaults?

francares:
:+1:

geoffjentry:
@francares @Horneth What about "supportedBackends"?

build.sbt
```diff
@@ -7,6 +7,18 @@ lazy val backend = (project in file("backend"))
   .dependsOn(core % "test->test;compile->compile")
   .settings(backendSettings:_*)
 
+lazy val localBackend = (project in file("default-backends/local"))
```
geoffjentry: I think this is fine for now but at some point we might want to think about also breaking this stuff out a bit

geoffjentry:
:+1: potentially modulo the name of that subproject, I'm partial to the supported verbiage

Horneth:
Are we good with `supportedBackends` @francares @geoffjentry ? I honestly don't have an opinion on this

cjllanwarne:
@Horneth oops, I meant in-SBT for other backends. But I think it's still a stupid question. LGTM!

cjllanwarne:
Also FWIW I would have *every* backend plugged in from the "pluggables" directory from the start. For the sake of (a) dog-fooding and (b) not creating two tiers of backend - supported and ... (presumably) unsupported?

Horneth:
Agreed, as it is now they're all (local, sge, jes, htcondor) aggregated into the cromwell jar so they are effectively plugged in no ?

cjllanwarne:
I would *prefer* - but not necessarily in this PR - if **all** backends were separate jars from the start. So that we dog-food on the whole "pluggable" thing. And so that we can test (e.g.) a cromwell which doesn't have the concept of LocalBackend anywhere in its classpath to make sure we didn't accidentally rely on it somewhere

cjllanwarne:
Also a heads up to @Horneth or anyone planning on using these - they are going to change a little bit when #688 arrives

Horneth:
> if all backends were separate jars from the start

they are :)
`sbt assembly` on this will create:
cromwell-core.jar
cromwell-engine.jar 
cromwell-backend.jar // Backend trait
cromwell-local-backend.jar
cromwell-jes-backend.jar
cromwell-sge-backend.jar
cromwell-htcondor-backend.jar
cromwell.jar // Aggregation of all of the above

supportedBackends/htcondor/src/main/scala/cromwell/htcondorbackend/HtCondorBackend.scala
```diff
@@ -0,0 +1,29 @@
+package cromwell.htcondorbackend
```
mcovarr: These package names are all two words smashed together, can they be broken out into two components?

geoffjentry:
@cjllanwarne I disagree. IMO officially supported backends should be provided by the downloadable fat jar, we should not require a user to need to download multiple jars if all they want to do is fart around w/ the officially supported stuff.

geoffjentry:
@cjllanwarne Although perhaps the solution there is to have a release assembly which bundles everything together and a lighter weight one as well, which we can use for the situation you describe

supportedBackends/htcondor/src/main/scala/cromwell/htcondorbackend/HtCondorBackend.scala
Horneth: I was worried that naming them `cromwell.backend.<backendname>` would be confusing as we already have `cromwell.backend` for the global backend project containing backend definition

mcovarr: Yeah I think you're right that would be confusing.  Maybe something like `cromwell.backend.impl.<backendname>`?

PR #682, opened 2016-04-08
[Info Only] Per Call Backend Changes

cjllanwarne:
This is just half-completed as far as I got trying to make Per-Call backends.

For information only. Do not Merge


geoffjentry:
@cjllanwarne Why is this still open?

PR #686, opened 2016-04-08
Added validation back into the preStart of a BackendWorkflowActor

cjllanwarne:


cjllanwarne:
Probably @francares. Also @scottfrazer please!

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
```diff
@@ -29,6 +30,10 @@ object WorkflowBackendActor {
   sealed trait AbortResponse extends WorkflowBackendActorResponse
   case class AbortSucceeded(jobKey: BackendJobDescriptorKey) extends AbortResponse
   case class AbortFailed(jobKey: BackendJobDescriptorKey, throwable: Throwable) extends AbortResponse
+
+  sealed trait ValidationResponse extends WorkflowBackendActorResponse
```
francares: I think these are not needed.

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
```diff
@@ -37,29 +42,35 @@ object WorkflowBackendActor {
 trait WorkflowBackendActor extends Actor with ActorLogging {
 
   protected implicit def ec: ExecutionContext
+  protected def calls: Seq[Call]
```
francares: I think it will be worthy to have some description of why calls are being added so the backend developer knows about that.

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
```diff
@@ -75,4 +86,9 @@ trait WorkflowBackendActor extends Actor with ActorLogging {
     * Abort a running job.
     */
   def abort(jobKey: BackendJobDescriptorKey): Future[AbortResponse]
+
+  /**
+    * Validate that this WorkflowBackendActor can run all of the calls that it's been assigned
+    */
+  def validate(): Future[ValidationResponse]
```
francares: This should be blocking. In other words Unit. If there is a problem should throw an exception otherwise nothing.

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
```diff
@@ -37,29 +42,35 @@ object WorkflowBackendActor {
...
   protected implicit def ec: ExecutionContext
+  protected def calls: Seq[Call]
   protected def workflowDescriptor: BackendWorkflowDescriptor
   protected def configurationDescriptor: BackendConfigurationDescriptor
 
   def receive: Receive = LoggingReceive {
-    case Execute(jobDescriptor) => performActionThenRespond(execute, jobDescriptor, onFailure = descriptorToExecutionFailure)
-    case Recover(jobDescriptor) => performActionThenRespond(recover, jobDescriptor, onFailure = descriptorToExecutionFailure)
-    case Abort(jobKey)          => performActionThenRespond(abort, jobKey, onFailure = AbortFailed)
+    case Execute(jobDescriptor) => performActionThenRespond( () => execute(jobDescriptor), onFailure = executionFailed(jobDescriptor.key))
```
francares: I like this change.

francares:
:+1: after changes.

francares:
I think beforeAll and afterAll can be added in the future. It would be great to make that behavior explicit to the developer.

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
cjllanwarne: Re this and a comment further down, I'm about to commit my FSM for ShadowWorkflowActor which builds on this. Hopefully you'll see how -

- WorkflowBackendActors start up immediately, and in the preStart will begin their validation (which might be slow if they need to make remote FS calls, but usually not)
- On `ValidationSuccess` or `ValidationFailed`, report back to the WorkflowActor which created them. The `ShadowWorkflowActor` aggregates the responses asynchronously and proceeds once it gets something back from every backend

cjllanwarne:
For simplicity, I'm making my two changes into a single PR - they're more entwined than I realised. Please see #688 instead!

PR #688, opened 2016-04-08
ShadowWorkflowActor FSM and Lifecycle Actors. Closes #683

### LINKED ISSUE #683: Add FSM and states to Shadow Workflow Actor

cjllanwarne:


cjllanwarne:
@francares @scottfrazer Sorry for the confusion. Could you review this one instead? Thanks!

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
```diff
@@ -1,19 +1,128 @@
...
+      ???
+  }
+
+  when(ExecutingWorkflow) {
+    case Event(StartRunnableCallsCommand, data) => ??? // TODO: Hold onto your butts, 'cos here we go!
+    case Event(_, _) => ??? // TODO: I'm sure there'll be other events to catch!
+  }
+
+  // TODO: This may not even be a real state
+  when(CliveDunn) {
```
scottfrazer: I don't get why this is here...

engine/src/main/scala/cromwell/engine/workflow/package.scala
```diff
@@ -0,0 +1,5 @@
+package cromwell.engine
+
+package object workflow {
```
scottfrazer: Does this need to exist?

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
mcovarr: yeah should be Nora Dunn

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
```diff
@@ -1,19 +1,128 @@
...
+  when(Unstarted) {
+    case Event(StartCommand(sm: StartMode), data) =>
+      // TODO: Begin MaterializingWorkflowDescriptor (by outsourcing to MaterialiseWorkflowDescriptorActor)
+      goto(MaterializingWorkflowDescriptor) using data.copy(startMode = sm)
+  }
+
+  when(MaterializingWorkflowDescriptor) {
+    case Event(MaterializationSuccess(workflowDescriptor: WorkflowDescriptor), emptyMap) =>
+      // TODO: Begin the WorkflowBackendActor materializification!
+      goto(MaterializingWorkflowBackendActors) using emptyMap
```
scottfrazer: is this `using emptyMap` correct?

scottfrazer:
Seems fine to me... assuming that the TODOs are not supposed to be filled in on this PR

cjllanwarne:
I've had a request from @geoffjentry and @mcovarr to convert the "validation response aggregation" into a separate actor - see https://docs.google.com/a/broadinstitute.com/document/d/1qTXiPtiJcmfmWghLC_uBeWlL2SuCE2Ek1xBiT0j8-iQ/edit?usp=sharing - I'll rework this PR then re-open

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
cjllanwarne: Ah no, probably not!

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
```diff
@@ -29,37 +31,68 @@ object WorkflowBackendActor {
   sealed trait AbortResponse extends WorkflowBackendActorResponse
   case class AbortSucceeded(jobKey: BackendJobDescriptorKey) extends AbortResponse
   case class AbortFailed(jobKey: BackendJobDescriptorKey, throwable: Throwable) extends AbortResponse
+
+  sealed trait InitializationResponse extends WorkflowBackendActorResponse
+  case object InitializationSuccess extends InitializationResponse
+  case class InitializationFailed(reason: Throwable) extends InitializationResponse
```
geoffjentry: This could be made final (i think)

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
```diff
@@ -29,37 +31,68 @@ object WorkflowBackendActor {
...
 
 /**
   * Workflow-level actor for executing, recovering and aborting jobs.
   */
 trait WorkflowBackendActor extends Actor with ActorLogging {
 
+  /**
+    * The execution context for the actor
+    */
   protected implicit def ec: ExecutionContext
```
geoffjentry: Is this too allow a backend implementer to not need to use the actor system's ec?

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
```diff
@@ -29,37 +31,68 @@ object WorkflowBackendActor {
...
-    // To avoid sending a response to the wrong `sender` if more than one `execute` arrives before the first one finishes,
-    // for example:
-    val sndr = sender()
-    operation(operationArgument) onComplete {
-      case Success(r) => sndr ! r
-      case Failure(t) => sndr ! onFailure(operationArgument, t)
+  private def performActionThenRespond(operation: () => Future[WorkflowBackendActorResponse],
+                                       onFailure: (Throwable) => WorkflowBackendActorResponse) = {
+    val respondTo: ActorRef = sender
+    operation.apply onComplete {
```
geoffjentry: The Akka devs frequently advise against using callbacks on Futures inside of an Actor like this. At the moment all of the alternate schemes I've thought of would incur more of a hassle on the part of a backend implementer (although one could argue that requiring the implementer to be sure to not F themselves by screwing up their internal state is just as bad). 

I'm still looking at alternatives but if someone else sees this and has a good idea please chime in.

geoffjentry: Disregard comment which was previously here. Since we're controlling all the dice here we just need to be careful that the onFailure function doesn't close over anything.

engine/src/main/scala/cromwell/engine/workflow/InitializeWorkflowBackendsActor.scala
```diff
@@ -0,0 +1,86 @@
...
+  case object BackendInitializationsSucceeded extends InitializeWorkflowBackendsActorResponse
+  case class BackendInitializationsFailed(reasons: Map[ActorRef, Throwable]) extends InitializeWorkflowBackendsActorResponse
+
+  def props(workflowTag: String, backendsToInitialize: Seq[ActorRef]): Props = Props(new InitializeWorkflowBackendsActor(workflowTag, backendsToInitialize))
+
+}
+
+case class InitializeWorkflowBackendsActor(workflowTag: String, backendsToInitialize: Seq[ActorRef]) extends LoggingFSM[InitializeWorkflowBackendsActorState, InitializeWorkflowBackendsActorData] {
+
+  startWith(Unstarted, InitializeWorkflowBackendsActorData(Seq.empty, Map.empty))
```
geoffjentry: If you want an empty `List`, make an empty `List`

engine/src/main/scala/cromwell/engine/workflow/InitializeWorkflowBackendsActor.scala
```diff
@@ -0,0 +1,86 @@
...
+    case Event(InitializationSuccess, stateData) =>
+      val newData = InitializeWorkflowBackendsActorData(stateData.successes ++ Seq(sender), stateData.failures)
+      checkForDoneAndTransition(newData)
+    case Event(InitializationFailed(reason), stateData) =>
+      val newData = InitializeWorkflowBackendsActorData(stateData.successes, stateData.failures ++ Map(sender -> reason))
+      checkForDoneAndTransition(newData)
+  }
+
+  private def checkForDoneAndTransition(newData: InitializeWorkflowBackendsActorData): State = {
+    if(checkForDone(newData)) {
```
geoffjentry: need some spaces in these ifs

engine/src/main/scala/cromwell/engine/workflow/InitializeWorkflowBackendsActor.scala
```diff
@@ -0,0 +1,86 @@
...
+      if(newData.failures.isEmpty) {
```
geoffjentry: You could kinda collapse this a bit by assigning `BackendInitializationsSucceeded` and `BackendInit..Failed` to a val based on the `newData.failures` and then send that val to the parent

engine/src/main/scala/cromwell/engine/workflow/InitializeWorkflowBackendsActor.scala
```diff
@@ -0,0 +1,86 @@
...
+    allActorResponses.size == backendsToInitialize.size
+  }
+
+  whenUnhandled {
+    case unhandledMessage =>
+      log.info(s"Backends initializer for $workflowTag received an unhandled message: $unhandledMessage")
+      stay
+  }
+
+  // TODO: Kill on transition to Done?
```
geoffjentry: I'd say either that or have parent stop it

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
```diff
@@ -1,19 +1,124 @@
...
 
   private val tag = this.getClass.getSimpleName
 
-  override def receive: Receive = {
-    case unhandledMessage: Any => log.info(s"$tag received an unhandled message: $unhandledMessage")
+  startWith(Unstarted, ShadowWorkflowActorData(NoStartMode, Map.empty))
+
+  when(Unstarted) {
+    case Event(StartCommand(sm: StartMode), data) =>
+      // TODO: Begin MaterializingWorkflowDescriptor (by outsourcing to MaterialiseWorkflowDescriptorActor)
```
geoffjentry: :us:

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
```diff
@@ -1,19 +1,124 @@
...
+
+  when(MaterializingWorkflowDescriptor) {
+    case Event(MaterializationSuccess(workflowDescriptor: WorkflowDescriptor), stateData) =>
+      // TODO: Begin the WorkflowBackendActor materializification!
+      val newBackendAssignments: Map[Call, ActorRef] = ???
+      val initializerActor = context.actorOf(InitializeWorkflowBackendsActor.props(tag, newBackendAssignments.values.toSeq))
+      initializerActor ! StartInitializations
+      goto(MaterializingWorkflowBackendActors) using stateData.copy(backendMap = newBackendAssignments)
+    case Event(MaterializationFailure(reason: Throwable), _) =>
+      context.parent ! ShadowWorkflowFailed(reason)
```
geoffjentry: If you're going to message parent w/ an error and die, why not let it crash?

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
```diff
@@ -1,19 +1,124 @@
...
+      ??? // TODO: Kill this workflow actor?
+  }
+
+  when(MaterializingWorkflowBackendActors) {
+    case Event(BackendInitializationsSucceeded, _) =>
+      self ! StartRunnableCallsCommand // TODO: Hold onto your butts, here we go!
```
geoffjentry: butt is held

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
```diff
@@ -1,19 +1,124 @@
...
+  }
+
+  when(ExecutingWorkflow) {
+    case Event(StartRunnableCallsCommand, data) => ??? // TODO: Hold onto your butts, 'cos here we go!
+    case Event(_, _) => ??? // TODO: I'm sure there'll be other events to catch!
+  }
+
+  // Exists so that we can record the state transition.
+  when(Done) {
+    case Event(_, _) => ??? // TODO: Why are you still talking to me!? Go away!
```
geoffjentry: I will do just that.

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
cjllanwarne: It could, as could all these messages... but I don't think it needs to be? It wouldn't be protecting anything and what if a backend implementor want to do something clever with `InitializationFailed` subtypes between its internal actors?

geoffjentry: Do you see extending init failed for internal use as a real use case of just throwing out the one possible use?

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
cjllanwarne: The `onComplete`s need an implicit execution context. If we can drag that out of the actor system by default then so much the better!

engine/src/main/scala/cromwell/engine/workflow/InitializeWorkflowBackendsActor.scala
cjllanwarne: What if I don't care what type of empty it is?

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
cjllanwarne: I'm saying that I'd want a reason for making it final vs the default of not making it final. I can imagine a BE implementor *might* want to use subtypes in the future. I *cannot* see a strong reason for making it final.

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
cjllanwarne: Fail fast vs fail slow

engine/src/main/scala/cromwell/engine/workflow/InitializeWorkflowBackendsActor.scala
```diff
@@ -0,0 +1,92 @@
...
+
+object InitializeWorkflowBackendsActor {
+
+  /**
+    * States
+    */
+  sealed trait InitializeWorkflowBackendsActorState
+  case object Unstarted extends InitializeWorkflowBackendsActorState
+  case object InitializingBackends extends InitializeWorkflowBackendsActorState
+  case object NoraDunn extends InitializeWorkflowBackendsActorState
```
scottfrazer: I hate to be the party pooper but can we choose a better name for `NoraDunn`?

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
```diff
@@ -1,19 +1,125 @@
 package cromwell.engine.workflow
 
-import akka.actor.{Props, ActorLogging, Actor}
+import akka.actor.{ActorRef, LoggingFSM, Props, ActorLogging}
+import cromwell.backend.WorkflowBackendActor.{InitializationFailed, InitializationSuccess}
```
scottfrazer: Unused imports

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
```diff
@@ -1,19 +1,125 @@
...
+import cromwell.core.WorkflowId
+import cromwell.engine.WorkflowSourceFiles
+import cromwell.engine.backend.WorkflowDescriptor
+import cromwell.engine.workflow.InitializeWorkflowBackendsActor.{StartInitializations, BackendInitializationsFailed, BackendInitializationsSucceeded}
+import cromwell.engine.workflow.MaterializeWorkflowDescriptorActor.{MaterializationFailure, MaterializationSuccess}
+import cromwell.engine.workflow.ShadowWorkflowActor._
+import cromwell.engine.workflow.WorkflowManagerActor.WorkflowOutputs
+import wdl4s.{Call, WdlSource}
```
scottfrazer: `wdl4s.WdlSource` is unused

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
cjllanwarne: Oh wait, this is in materialisation, not execution. I guess we could kill children and crash?

mcovarr: wow that sounds awful

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
```diff
@@ -29,37 +31,68 @@ object WorkflowBackendActor {
...
     }
   }
 
+  /**
+    * Our predefined sequence to run during preStart
+    */
+  final def initSequence() = for {
+    _ <- beforeAll()
+    result <- validate()
+  } yield result
```
mcovarr: Perhaps I'm missing some stuff, but this is making me wonder:

1. Why does validation happen after beforeAll, which does things like upload auth files that really shouldn't be uploaded if the workflow doesn't validate?
2. Why is this implemented with `Future`s?  This `WorkflowActor` is open for business once `preStart` returns, but if we've spun off a `Future` we're not waiting to see if these prerequisite operations have completed.
3. It feels weird that we dictate to an implementer that they must write a `beforeAll` and then that they must have it call this particular method.

engine/src/main/scala/cromwell/engine/workflow/InitializeWorkflowBackendsActor.scala
mcovarr: seems like an ideal application for Cockney rhyming slang

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
cjllanwarne: 1) Good point, I can switch the order. And maybe rename `beforeAll` to something more accurate - `postValidationBeforeAnythingElse`?
2) As per our conversation Friday. This is no longer called from `preStart` so you'd have an actor even *before* this is called. But the only thing that would use it is `ShadowWorkfowActor` which we'll have to trust not to start executions when it's in the `MaterializingBackendWorkflowActors` state
3) This one I'm not sure I understand. We're making them write a `beforeAll`, but it's the `ShadowWorkflowActor` (technically the `InitializeWorkflowBackendsActor`) who's causing it to be run

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
```diff
@@ -75,4 +108,19 @@ trait WorkflowBackendActor extends Actor with ActorLogging {
...
+
+  /**
+    * Validate that this WorkflowBackendActor can run all of the calls that it's been assigned
+    */
+  def validate(): Future[InitializationResponse]
+
+  /**
+    * A call which happens before anything else runs
+    */
+  def beforeAll(): Future[Unit]
```
francares: With the new changes you will need to add messages here I think.

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
mcovarr: I like keeping this non-final, implementers can throw classes of failures more specifically by subtype.  For JES in particular we can draw distinctions between "could not upload auth file" and "you're missing a docker runtime attribute in call 'foo'.".

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
francares: One use case that may happen is when the backend developer does not deal well with exceptions, so if an unhandled exception happen, SWA should be able to recover from that situation and fail gracefully.

engine/src/main/scala/cromwell/engine/workflow/InitializeWorkflowBackendsActor.scala
```diff
@@ -0,0 +1,91 @@
...
+    */
+  sealed trait InitializeWorkflowBackendsActorResponse
+  case object BackendInitializationsSucceeded extends InitializeWorkflowBackendsActorResponse
+  case class BackendInitializationsFailed(reasons: Map[ActorRef, Throwable]) extends InitializeWorkflowBackendsActorResponse
+
+  def props(workflowTag: String, backendsToInitialize: Seq[ActorRef]): Props = Props(new InitializeWorkflowBackendsActor(workflowTag, backendsToInitialize))
+
+}
+
+case class InitializeWorkflowBackendsActor(workflowTag: String, backendsToInitialize: Seq[ActorRef]) extends LoggingFSM[InitializeWorkflowBackendsActorState, InitializeWorkflowBackendsActorData] {
```
francares: Is this going to be refactored based on the new idea?.
Calling to BeforeAll and then Validate?

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
```diff
@@ -1,19 +1,127 @@
...
+  startWith(Unstarted, ShadowWorkflowActorData(NoStartMode, Map.empty))
+
+  when(Unstarted) {
+    case Event(StartCommand(sm: StartMode), data) =>
+      // TODO: Begin MaterializingWorkflowDescriptor (by outsourcing to MaterializeWorkflowDescriptorActor)
+      goto(MaterializingWorkflowDescriptor) using data.copy(startMode = sm)
+    case Event(x, data) => throw new InvalidMessageException(s"Didn't expect $x message in state $Unstarted")
+  }
+
+  when(MaterializingWorkflowDescriptor) {
```
francares: Can we change the name to InitializingWorkflow or something more descriptive?

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
cjllanwarne: It's part of the same InitializationSequence as validate. So we do:
```
final def initSequence() = for {
    _ <- validate()
    _ <- beforeAll()
  } yield InitializationSuccess
```
And if either throws an exception, we get an InitializationFailure instead

engine/src/main/scala/cromwell/engine/workflow/InitializeWorkflowBackendsActor.scala
cjllanwarne: Yes, except it's wrapped up in the same command - `Initialize`

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
```diff
@@ -1,19 +1,124 @@
...
+    */
+  case object MaterializingWorkflowBackendActors extends ShadowWorkflowActorState
+
+  /**
+    * The WorkflowActor has a descriptor and a set of backends. It now needs to execute the hell out of its jobs.
+    */
+  case object ExecutingWorkflow extends ShadowWorkflowActorState
+
+  /**
+    * We're done. But we're more than done. We're Clive Dunn.
```
mcovarr: Done and Doner.

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
```diff
@@ -1,19 +1,124 @@
...
+    */
+  sealed trait ShadowWorkflowActorResponse
+  case class ShadowWorkflowSucceeded(outputs: WorkflowOutputs)
+  case class ShadowWorkflowFailed(reason: Throwable)
+
+  /**
+    * Mode in which the workflow should be started:
+    */
+  sealed trait StartMode
+  case object NoStartMode extends StartMode
```
mcovarr: what is this?

backend/src/main/scala/cromwell/backend/WorkflowBackendActor.scala
geoffjentry: It's coo', @cjllanwarne convinced me. I just wanted to make sure there was an actual use case in mind.

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
```diff
@@ -1,19 +1,127 @@
...
+
+  when(MaterializingWorkflowDescriptor) {
+    case Event(MaterializationWorkflowDescriptorSuccess(workflowDescriptor: WorkflowDescriptor), stateData) =>
+      // TODO: Begin the WorkflowBackendActor materializification!
+      // TODO: Subscribe to MaterializeWorkflowDescriptorActor crashes.
+      val newBackendAssignments: Map[Call, ActorRef] = ???
+      val initializerActor = context.actorOf(InitializeWorkflowBackendsActor.props(tag, newBackendAssignments.values.toSeq))
+      initializerActor ! StartInitializations
+      goto(MaterializingWorkflowBackendActors) using stateData.copy(backendMap = newBackendAssignments)
+    // TODO: Convert this "Event" into a "on crash" handler for MaterializeWorkflowDescriptorActor:
```
scottfrazer: This TODO seems like it should implemented in this PR

engine/src/main/scala/cromwell/engine/package.scala
```diff
@@ -93,8 +93,8 @@ package object engine {
     val materializeWorkflowDescriptorActor = actorSystem.actorOf(MaterializeWorkflowDescriptorActor.props())
 
     materializeWorkflowDescriptorActor.askNoTimeout(MaterializeWorkflow(id, sources))  map {
-      case MaterializationSuccess(workflowDescriptor) => workflowDescriptor
-      case MaterializationFailure(error) => throw error
+      case MaterializationWorkflowDescriptorSuccess(workflowDescriptor) => workflowDescriptor
+      case MaterializeWorkflowDescriptorFailure(error) => throw error
```
scottfrazer: These two classes don't seem to be named consistently... maybe `MaterializeWorkflowDescriptorFailure` and `MaterializeWorkflowDescriptorSuccess`?

scottfrazer:
Is there any way to get even just some simple tests for `ShadowWorkflowActor`?  I feel like there are some things in `ShadowWorkflowActor` that could be restructured for better testing.

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
cjllanwarne: OK

cjllanwarne:
Closing while I implement some "Let It Crash"

cjllanwarne:
Reopening with the new LifecycleActors

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
```diff
@@ -1,19 +1,166 @@
...
   */
-class ShadowWorkflowActor extends Actor with ActorLogging{
+class ShadowWorkflowActor(workflowId: WorkflowId, wdlSource: WorkflowSourceFiles) extends LoggingFSM[ShadowWorkflowActorState, ShadowWorkflowActorData] with ActorLogging{
 
   private val tag = this.getClass.getSimpleName
 
-  override def receive: Receive = {
-    case unhandledMessage: Any => log.info(s"$tag received an unhandled message: $unhandledMessage")
+  startWith(Unstarted, ShadowWorkflowActorData(NoStartMode, Map.empty))
+
```
cjllanwarne: You buffoon! You still need to add the supervision strategy!

gauravs90:
I like this design change 😄

backend/src/main/scala/cromwell/backend/BackendWorkflowLifecycleActor.scala
```diff
@@ -0,0 +1,66 @@
...
+
+  protected def performActionThenRespond(operation: () => Future[BackendWorkflowLifecycleActorResponse],
+                                         onFailure: (Throwable) => BackendWorkflowLifecycleActorResponse) = {
+    val respondTo: ActorRef = sender
+    operation.apply onComplete {
+      case Success(r) => respondTo ! r
+      case Failure(t) => respondTo ! onFailure(t)
+    }
+  }
+}
```
mcovarr: \n

backend/src/main/scala/cromwell/backend/BackendWorkflowLifecycleActor.scala
```diff
@@ -0,0 +1,66 @@
...
+    * The workflow descriptor for the workflow in which this Backend is being used
+    */
+  protected def workflowDescriptor: BackendWorkflowDescriptor
+
+  /**
+    * The configuration for the backend, in the context of the entire Cromwell configuration file.
+    */
+  protected def configurationDescriptor: BackendConfigurationDescriptor
+
+  protected def performActionThenRespond(operation: () => Future[BackendWorkflowLifecycleActorResponse],
```
mcovarr: I think if you make this `operation: => Future[...]` you can get rid of the `() =>` at the call sites

backend/src/main/scala/cromwell/backend/BackendWorkflowFinalizationActor.scala
```diff
@@ -0,0 +1,46 @@
...
+    case AbortWorkflow => performActionThenRespond(abortFinalization, onFailure = WorkflowAbortFailed)
+  }
+
+  /**
+    * Abort all initializations.
+    */
+  def abortFinalization(): Future[WorkflowAbortResponse]
+
+  /**
+    * A call which happens before anything else runs
```
mcovarr: actually, the opposite of that

backend/src/main/scala/cromwell/backend/BackendWorkflowFinalizationActor.scala
```diff
@@ -0,0 +1,46 @@
...
+  */
+trait BackendWorkflowFinalizationActor extends BackendWorkflowLifecycleActor with ActorLogging {
+
+  def receive: Receive = LoggingReceive {
+    case Finalize => performActionThenRespond(afterAll, onFailure = FinalizationFailed)
+    case AbortWorkflow => performActionThenRespond(abortFinalization, onFailure = WorkflowAbortFailed)
+  }
+
+  /**
+    * Abort all initializations.
```
mcovarr: finalizations

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
```diff
@@ -1,19 +1,166 @@
...
+import cromwell.engine.workflow.lifecycle.FinalizeWorkflowBackendsActor.{BackendFinalizationsFailed, StartFinalizations}
+import cromwell.engine.workflow.lifecycle.{FinalizeWorkflowBackendsActor, ExecuteWorkflowActor, InitializeWorkflowBackendsActor}
+import InitializeWorkflowBackendsActor.{StartInitializations, BackendInitializationsFailed, BackendInitializationsSucceeded}
+import cromwell.engine.workflow.MaterializeWorkflowDescriptorActor.{MaterializeWorkflowDescriptorFailure, MaterializationWorkflowDescriptorSuccess}
+import cromwell.engine.workflow.ShadowWorkflowActor._
+import cromwell.engine.workflow.WorkflowManagerActor.WorkflowOutputs
+import cromwell.util.AggregatedException
+import wdl4s.Call
 
 object ShadowWorkflowActor {
```
mcovarr: Per earlier discussion, I think it's maybe time for this to come out of the shadows.

backend/src/main/scala/cromwell/backend/BackendWorkflowExecutionActor.scala
```diff
@@ -0,0 +1,57 @@
...
+  case class ExecutionFailedRetryable(jobKey: BackendJobDescriptorKey, throwable: Throwable) extends ExecutionResponse
+}
+
+/**
+  * Workflow-level actor for executing, recovering and aborting jobs.
+  */
+trait BackendWorkflowExecutionActor extends BackendWorkflowLifecycleActor with ActorLogging {
+
+  def receive: Receive = LoggingReceive {
+    case Execute(jobDescriptor) => performActionThenRespond( () => execute(jobDescriptor), onFailure = executionFailed(jobDescriptor.key))
```
francares: Remove extra space.

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
cjllanwarne: Is that now an Officially Endorsed Decision (TM)?

cjllanwarne: I'd probably rather do that in a subsequent PR though in any case

geoffjentry: It is indeed, as badass as 'shadow' sounds

cjllanwarne:
@scottfrazer @mcovarr I think your two remaining comments are related - having some tests for this, and replacing the existing WMA with the new version. I'd lobby for doing them together in the very next ticket I'm picking up - i.e. #692

mcovarr:
@cjllanwarne sounds good, unofficial thumb from me

francares:
:+1: for this iteration. There are some topics that are not so clear at this moment (such as sharing information between beforeAll->execute->afterAll) but I think we can move forward with this. It might suffer changes in the future but it's a starting point.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/InitializeWorkflowBackendsActor.scala
```diff
@@ -0,0 +1,114 @@
+package cromwell.engine.workflow.lifecycle
+
+import akka.actor.{ActorRef, LoggingFSM, Props}
+import cromwell.backend.BackendWorkflowInitializationActor._
+import cromwell.engine.workflow.lifecycle.InitializeWorkflowBackendsActor._
+import wdl4s.Call
+
+object InitializeWorkflowBackendsActor {
```
scottfrazer: Okay this was confusing to me... we have a `InitializeWorkflowBackendsActor` and a `BackendWorkflowInitializationActor`... same 4 words... I'd certainly not be opposed to some better terminology

cjllanwarne:
I've changed the names from the diagram a little:
`InitializeWorkflowBackendsActor` => `EngineWorkflowInitializationActor`
`ExecuteWorkflowActor` => `EngineWorkflowExecutionActor`
`FinalizeWorkflowBackendsActor` => `EngineWorkflowFinalizationActor`
`BackendWorkflowExecutionActor` => `BackendJobExecutionActor` (now also per-Job not per-workflow)

geoffjentry:
@cjllanwarne Please name actors when creating them

cjllanwarne:
@geoffjentry done!

geoffjentry:
@cjllanwarne awesome, thanks.

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
```diff
@@ -1,19 +1,166 @@
...
+  private val tag = s"$workflowId-${this.getClass.getSimpleName}"
 
-  private val tag = this.getClass.getSimpleName
+  startWith(Unstarted, ShadowWorkflowActorData(Map.empty))
+
+  val decider: PartialFunction[Throwable, Directive] = {
+    case _: Throwable => Escalate
+  }
+  override def supervisorStrategy: SupervisorStrategy =
+    OneForOneStrategy()(decider.orElse(SupervisorStrategy.defaultStrategy.decider))
```
scottfrazer: So this means... "If an actor that I'm supervising dies, escalate the exception to my supervisor"?

If I'm understanding this correctly, the decider that you've defined doesn't need an `orElse`, so why can't we just write `OneForOneStrategy()(decider)`?

scottfrazer: maybe even this?

```scala
override def supervisorStrategy: SupervisorStrategy = OneForOneStrategy() {
  case _: Throwable => Escalate
}
```

engine/src/main/scala/cromwell/engine/workflow/lifecycle/EngineWorkflowInitializationActor.scala
```diff
@@ -0,0 +1,118 @@
...
+
+  private def checkForDone(stateData: EngineWorkflowInitializationActorData) = {
+    // TODO: Check these are actually the same?
+    val allActorResponses = stateData.successes ++ stateData.failures.keys
+    allActorResponses.size == backendsToInitialize.size
+  }
+
+  whenUnhandled {
+    case unhandledMessage =>
+      log.info(s"Backends initializer for $workflowTag received an unhandled message: $unhandledMessage")
```
scottfrazer: We shouldn't be seeing any unhandled messages if I'm not mistaken... so maybe `log.warn`?

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
```diff
@@ -1,19 +1,166 @@
...
+      ???
+    case Event(AbortWorkflowCommand, stateData) => ??? // TODO: Handle abort
+  }
+
+  when(ExecutingWorkflow) {
+    case Event(BackendJobExecutionSucceededResponse, stateData) =>
+      val finalizationActor = context.actorOf(EngineWorkflowFinalizationActor.props(tag, stateData.backendAssignments), name = s"EngineWorkflowFinalizationActor-$workflowId")
+      finalizationActor ! StartEngineFinalizationCommand
+      goto(FinalizingWorkflow)
+    case Event(WorkflowInitializationFailed(reasons), _) =>
```
scottfrazer: Is this supposed to be:

```scala
case Event(BackendJobExecutionFailedResponse(key, throwable), _) =>
```

scottfrazer:
Okay so my understanding of this so far (I'm just reiterating what you have in your diagram for my own benefit):

`ShadowWorkflowActor` has 4 "lifecycle" states as actors:

* `MaterializeWorkflowDescriptorActor` - basic validation and `WorkflowDescriptor` creation
* `EngineWorkflowInitializationActor` - spawns 1 or more `BackendWorkflowInitializationActor`, then aggregates results from all of these and message back `ShadowWorkflowActor`
* `EngineWorkflowExecutionActor` - sent a Start or Restart message, performs the execution, sends back result of execution.  Spawns `BackendWorkflowExecutionActor`s
* `EngineWorkflowFinalizationActor` - do post-workflow termination actions.  Spawns `BackendWorkflowFinalizationActor`s

*I'm just thinking out loud here... if you think this is stupid, I won't feel bad if you ignore me*

I guess the only comments on this scheme are about naming...  

I feel like `MaterializeWorkflowDescriptorActor` should follow the same naming scheme and maybe be called something like `EngineWorkflowDescriptorActor` or `EngineWorkflowParserActor`.  I'm also not a huge fan of prepending `Engine` onto these actors... maybe we can just drop `Engine`?

* `WorkflowDescriptorActor`
* `WorkflowInitializationActor`
* `WorkflowExecutionActor`
* `WorkflowFinalizationActor`

engine/src/main/scala/cromwell/engine/workflow/lifecycle/EngineWorkflowExecutionActor.scala
```diff
@@ -0,0 +1,86 @@
...
+    */
+  sealed trait EngineWorkflowExecutionActorResponse
+  case object WorkflowExecutionSucceededResponse extends EngineWorkflowExecutionActorResponse
+  case object WorkflowExecutionAbortedResponse extends EngineWorkflowExecutionActorResponse
+  final case class WorkflowExecutionFailedResponse(reasons: Map[ActorRef, Throwable]) extends EngineWorkflowExecutionActorResponse
+
+  def props(workflowTag: String, backendAssignments: Map[Call, String]): Props = Props(EngineWorkflowExecutionActor(workflowTag, backendAssignments))
+}
+
+final case class EngineWorkflowExecutionActor(workflowTag: String, backendAssignments: Map[Call, String]) extends LoggingFSM[EngineWorkflowExecutionActorState, EngineWorkflowExecutionActorData] {
```
scottfrazer: I noticed that `EngineWorkflow*Actor` does not define a `SupervisorStrategy`.  Is this part of the contract between the `BackendWorkflow*Actor`?  I feel it should be... and if it, should it be explicitly spelled out?

scottfrazer:
What was the rationale behind the `supportedBackends` directory?  Right now it seems [incredibly sparse](https://gist.github.com/scottfrazer/2fc865fab8f200cf6bb9cf09eb0f1905) and I don't see what benefit we gain from putting them there.  I also have a hard time seeing how we share code between backends in this scheme, like in the case of Local and SGE using a lot of the same functionality.

I realize you're all going to pile on and flog me for even suggesting this... but I kinda thought it was going to be something kind of similar to this (but obviously structured better):

```
$ tree backend/src
backend/src
├── main
│   ├── java
│   ├── resources
│   ├── scala
│   │   └── cromwell
│   │       └── backend
│   │           ├── BackendJobExecutionActor.scala
│   │           ├── BackendLifecycleActor.scala
│   │           ├── BackendWorkflowFinalizationActor.scala
│   │           ├── BackendWorkflowInitializationActor.scala
│   │           ├── WorkflowBackendActorFactory.scala
│   │           ├── package.scala
│   │           └── supported
│   │               ├── HtCondorBackend.scala
│   │               ├── JesBackend.scala
│   │               ├── LocalBackend.scala
│   │               ├── SgeBackend.scala
│   │               └── SharedFileSystem.scala
│   └── scala-2.11
└── test
    ├── java
    ├── resources
    ├── scala
    └── scala-2.11
```

scottfrazer:
So we're producing these JARs during assembly:

* backend/target/scala-2.11/cromwell-backend-0.1.jar
* core/target/scala-2.11/cromwell-core-0.20.jar
* engine/target/scala-2.11/cromwell-engine-0.20.jar
* filesystems/gcs/target/scala-2.11/cromwell-gcsfilesystem-0.1.jar
* supportedBackends/htcondor/target/scala-2.11/cromwell-htcondor-backend-0.1.jar
* supportedBackends/jes/target/scala-2.11/cromwell-jes-backend-0.1.jar
* supportedBackends/local/target/scala-2.11/cromwell-local-backend-0.1.jar
* supportedBackends/sge/target/scala-2.11/cromwell-sge-backend-0.1.jar
* target/scala-2.11/cromwell-0.20.jar

Are we going to publish all of them to Artifactory?  Who are the consumers are these `cromwell-*-backend` JARs?  I have a hard time seeing why somebody would need a `cromwell-[sge|local|jes|htcondor]-backend-0.1.jar` if it's already built into `cromwell-0.20.jar` by default

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
scottfrazer: Or maybe the svelt

```scala
override def supervisorStrategy: SupervisorStrategy = OneForOneStrategy() { case _ => Escalate }
```

Horneth:
I may not know sbt enough to know if there is a way around, but by putting the backend implementations into the backend project, I think the backend jar will contain all the backend implementations, which is not its role.
It's providing an interface for any potential backend implementor and they might not want to also get all our "supported" backends with it.
My understanding of pluggable backend is that we have self-contained jars that only need the backend interface (and messages etc..) and cromwell-core. Those jars can be plugged-in and out seemingly in the conf file, and will be used by the engine. I think it's a good idea to also do that for our own backend implementations, even if they're then aggregated into the master cromwell jar.

scottfrazer:
So I guess we share code between the backends through the `core` project?

Horneth:
Hmm core doesn't feel like the right place for this to me. We currently don't have a really good place to share backend code between backends. Maybe we need a meta-backend concept :/ ?
Although if you're thinking about the SharedFileSystem trait, it could very well fit in the filesystems project IMO, and be re-used by both local and SGE.

scottfrazer:
is `filesystem` included in `core`?  I just did a quick check of the `core` JAR and it doesn't appear to be

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
```diff
@@ -1,19 +1,166 @@
...
+      // TODO: Handle workflow failure
+      ???
+    case Event(AbortWorkflowCommand, stateData) => ??? // TODO: Handle abort
+  }
+
+  when(FinalizingWorkflow) {
+    case Event(BackendJobExecutionSucceededResponse, stateData) =>
+      goto(WorkflowSuccessful)
+    case Event(EngineFinalizationFailedResponse(reasons), _) =>
```
scottfrazer: The two events here to be coming from different places, is this right?

Horneth:
No `filesystem` is separated from everything

scottfrazer:
Okay, I'm having a little bit of trouble keeping track of what JARs are allowed to included where... but I'll avoid hijacking this PR to talk about that.

geoffjentry:
@scottfrazer FWIW my vision of the world was a lot closer to the diagram you drew up but I don't have strong feelings on that.

In terms of distributed jars what I'd like to see distributed to the world would be:

- cromwell.jar: full fat jar like we have today which also includes all of the supported backends built in
- cromwell-backend.jar: a jar providing the interface stuff which someone can use to build their own backend jar

I'd be totally okay with (and could see value in):

- cromwell-lite.jar (or something like that): a stripped down fat jar w/o any supported backends or maybe local
- foo-backend.jar for each supported backend.

My main concern though is that we always make the most obvious download for a naive user of cromwell to be the one with all of the supported (or perhaps a 'very common supported' subset) backends built in so as to minimize the work someone needs to do to get rolling.

The other jars are really an artifact of the multi-project model and can be ignored.

The side discussion about `core` is exactly why I was picturing the hierarchy stemming from `backend` all along. Despite the name of `supportedBackends` being my request I'll admit I was just looking at the name when I said that, not thinking the whole thing through critically :)

IMO `core` should be code which is shared between all components, I'd call the filesystem concept a component, not something in core itself.

geoffjentry:
@cjllanwarne I apologize for the long comment which has nothing to do with your PR :)

engine/src/main/scala/cromwell/engine/workflow/lifecycle/EngineWorkflowExecutionActor.scala
cjllanwarne: So I was convinced that we weren't really Letting anything Crash, since we were accumulating errors and then passing a Message (just as an exception...) so for this PR I'll say "we use Messages" but it should be not too difficult to transition that in the future

cjllanwarne:
Apologies, had to squash this for the rebase

cjllanwarne:
@scottfrazer I prepended `Engine` to make them more obviously distinct from the `Backend` varieties. But I'll happily remove the incredibly awkward `Engine` :)

cjllanwarne:
@geoffjentry @Horneth @scottfrazer has that whole backends' packages discussion been recorded anywhere? Maybe in another ticket?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/EngineWorkflowExecutionActor.scala
scottfrazer: Agreed, no need to implement something that we don't actually need right now!

scottfrazer:
@cjllanwarne I'm happy with whatever naming scheme you want.  I'm just thinking out loud.

scottfrazer:
:+1: 

[Approved with PullApprove](https://pullapprove.com/broadinstitute/cromwell/pull-request/688/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=broadinstitute/cromwell)

PR #689, opened 2016-04-08
Service registry + KeyValue service.  Closes #653, Closes #665

### LINKED ISSUE #653: Implement Service Registry

### LINKED ISSUE #665: Implement K-V Service

scottfrazer:
NEW!  Designed by committee!  Implemented reluctantly!

### DIFF
--- a/backend/src/main/scala/cromwell/backend/package.scala
+++ b/backend/src/main/scala/cromwell/backend/package.scala
@@ -7,16 +7,27 @@ import wdl4s.values.WdlValue
 
 package object backend {
 
+  trait JobKey {
+    def scope: Scope
+    def index: Option[Int]
+    def attempt: Int
+    def tag: String
+  }
+
   /**
     * For uniquely identifying a job which has been or will be sent to the backend.
     */
-  case class BackendJobDescriptorKey(callFqn: String, index: Option[Int], attempt: Int)
+  case class BackendJobDescriptorKey(call: Call, index: Option[Int], attempt: Int) extends JobKey {
+    val scope = call
+    val tag = s"${call.fullyQualifiedName}:$index:$attempt"
+  }
 
   /**
     * For passing to a BackendWorkflowActor for job execution or recovery
     */
-  //case class BackendJobDescriptor(key: BackendJobDescriptorKey, call: Call, locallyQualifiedInputs: CallInputs)
-  case class BackendJobDescriptor(key: BackendJobDescriptorKey, call: Call, symbolMap: Map[FullyQualifiedName, WdlValue])
+  case class BackendJobDescriptor(descriptor: BackendWorkflowDescriptor,
+                                  key: BackendJobDescriptorKey,
+                                  symbolMap: Map[FullyQualifiedName, WdlValue])
 
   /**
     * For passing to a BackendActor construction time
--- a/engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
+++ b/engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
@@ -6,6 +6,7 @@ import java.util.{Date, UUID}
 import javax.sql.rowset.serial.SerialClob
 
 import com.typesafe.config.{Config, ConfigFactory, ConfigValueFactory}
+import cromwell.backend.JobKey
 import cromwell.core.{CallOutput, CallOutputs, WorkflowId}
 import cromwell.engine.ExecutionIndex._
 import cromwell.engine.ExecutionStatus._
@@ -459,6 +460,19 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
     runTransaction(action)
   }
 
+  override def getExecutionInfoByKey(workflowId: WorkflowId, call: Call, attempt: Int, key: String)
+                                    (implicit ec: ExecutionContext): Future[Option[Option[String]]] = {
+    val action = for {
+      executionResult <- dataAccess.executionsByWorkflowExecutionUuidAndCallFqnAndAttempt(
+        workflowId.toString, call.fullyQualifiedName, attempt
+      ).result.head
+
+      executionInfo <- dataAccess.executionInfoValueByExecutionAndKey(executionResult.executionId.get, key).result.headOption
+    } yield executionInfo
+
+    runTransaction(action)
+  }
+
   override def updateExecutionInfo(workflowId: WorkflowId,
                                    callKey: BackendCallKey,
                                    key: String,
@@ -476,22 +490,21 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
   }
 
   override def upsertExecutionInfo(workflowId: WorkflowId,
-                                   callKey: BackendCallKey,
+                                   key: JobKey,
                                    keyValues: Map[String, Option[String]])
                                   (implicit ec: ExecutionContext): Future[Unit] = {
     import ExecutionIndex._
 
     val action = for {
       executionResult <- dataAccess.executionsByWorkflowExecutionUuidAndCallFqnAndShardIndexAndAttempt(
-        workflowId.toString, callKey.scope.fullyQualifiedName, callKey.index.fromIndex, callKey.attempt).result.head
-
+        workflowId.toString, key.scope.fullyQualifiedName, key.index.fromIndex, key.attempt).result.head
       _ <- DBIO.sequence(keyValues map upsertExecutionInfo(executionResult.executionId.get))
     } yield ()
 
     runTransaction(action)
   }
 
-  private def upsertExecutionInfo(executionId: Int)(keyValue: (String, Option[String]))
+  protected def upsertExecutionInfo(executionId: Int)(keyValue: (String, Option[String]))
                                  (implicit ec: ExecutionContext): DBIO[Unit] = {
     val (key, value) = keyValue
 
--- a/engine/src/main/scala/cromwell/services/KeyValueServiceActor.scala
+++ b/engine/src/main/scala/cromwell/services/KeyValueServiceActor.scala
@@ -1,0 +1,72 @@
+package cromwell.services
+
+import akka.actor.{Actor, ActorRef, Props}
+import com.typesafe.config.Config
+import cromwell.backend.BackendJobDescriptor
+import cromwell.engine.db.DataAccess
+import cromwell.services.KeyValueServiceActor._
+import cromwell.services.ServiceRegistryActor.ServiceRegistryMessage
+
+import scala.concurrent.{Await, Future}
+import scala.util.{Failure, Success}
+
+object KeyValueServiceActor {
+  trait KvMessage
+  trait KvAction extends KvMessage with ServiceRegistryMessage {
+    def serviceName = "KeyValue"
+  }
+  trait KvResponse extends KvMessage
+
+  case class ScopedKey(jobDescriptor: BackendJobDescriptor, key: String)
+  case class KvPut(pair: KvPair) extends KvAction
+  case class KvGet(key: ScopedKey) extends KvAction
+
+  case class KvPair(key: ScopedKey, value: Option[String]) extends KvResponse
+  case class KvFailure(action: KvAction, failure: Throwable) extends KvResponse
+  case class KvKeyLookupFailed(action: KvGet) extends KvResponse
+  case class KvPutSuccess(action: KvPut) extends KvResponse
+
+  def props(serviceConfig: Config, globalConfig: Config) = {
+    Props(KeyValueServiceActor(serviceConfig, globalConfig))
+  }
+}
+
+case class KeyValueServiceActor(serviceConfig: Config, globalConfig: Config) extends Actor {
+  private implicit val ec = context.dispatcher
+  private val dataAccess = DataAccess.globalDataAccess
+
+  def receive = {
+    case action: KvGet => respond(sender(), action, doGet(action))
+    case action: KvPut => respond(sender(), action, doPut(action))
+  }
+
+  private def respond(replyTo: ActorRef, action: KvAction, response: Future[KvResponse]): Unit = {
+    response.onComplete {
+      case Success(x) => replyTo ! x
+      case Failure(ex) => replyTo ! KvFailure(action, ex)
+    }
+  }
+
+  private def doPut(put: KvPut): Future[KvResponse] = {
+    dataAccess.upsertExecutionInfo(
+      put.pair.key.jobDescriptor.descriptor.id,
+      put.pair.key.jobDescriptor.key,
+      Map(put.pair.key.key -> put.pair.value),
+      context.system
+    ).map(_ => KvPutSuccess(put))
+  }
+
+  private def doGet(get: KvGet): Future[KvResponse] = {
+    val executionInfo = dataAccess.getExecutionInfoByKey(
+      get.key.jobDescriptor.descriptor.id,
+      get.key.jobDescriptor.key.call,
+      get.key.jobDescriptor.key.attempt,
+      get.key.key
+    )
+
+    executionInfo map {
+      case Some(maybeValue) => KvPair(ScopedKey(get.key.jobDescriptor, get.key.key), maybeValue)
+      case None => KvKeyLookupFailed(get)
+    }
+  }
+}
--- a/engine/src/main/scala/cromwell/services/ServiceRegistryActor.scala
+++ b/engine/src/main/scala/cromwell/services/ServiceRegistryActor.scala
@@ -1,0 +1,53 @@
+package cromwell.services
+
+import akka.actor.{Actor, ActorRef, Props}
+import com.typesafe.config.{ConfigFactory, Config, ConfigObject}
+import cromwell.engine.db.DataAccess
+import cromwell.services.ServiceRegistryActor._
+import lenthall.config.ScalaConfig._
+
+import scala.collection.JavaConverters._
+
+object ServiceRegistryActor {
+  case class ServiceRegistryFailure(serviceName: String)
+  trait ServiceRegistryMessage {
+    def serviceName: String
+  }
+
+  def props(config: Config) = {
+    Props(ServiceRegistryActor(config))
+  }
+
+  def serviceNameToPropsMap(globalConfig: Config): Map[String, Props] = {
+    val serviceNamesToConfigStanzas = globalConfig.getObject("services").entrySet.asScala.map(x => x.getKey -> x.getValue).toMap
+    serviceNamesToConfigStanzas map {
+      case (serviceName, config: ConfigObject) => serviceName -> serviceProps(serviceName, globalConfig, config.toConfig)
+      case (serviceName, _) => throw new Exception(s"Invalid configuration for service $serviceName")
+    }
+  }
+
+  private def serviceProps(serviceName: String, globalConfig: Config, serviceStanza: Config): Props = {
+    val serviceConfigStanza = serviceStanza.getConfigOption("config").getOrElse(ConfigFactory.parseString(""))
+    val className = serviceStanza.getStringOr(
+      "class",
+      throw new Exception(s"Invalid configuration for service $serviceName: missing 'class' definition")
+    )
+    Props.create(Class.forName(className), serviceConfigStanza, globalConfig)
+  }
+}
+
+case class ServiceRegistryActor(globalConfig: Config) extends Actor {
+  import ServiceRegistryActor._
+
+  val services: Map[String, ActorRef] = serviceNameToPropsMap(globalConfig) map {
+    case (name, props) => name -> context.actorOf(props, "ServiceRegistryActor")
+  }
+
+  def receive = {
+    case msg: ServiceRegistryMessage =>
+      services.get(msg.serviceName) match {
+        case Some(ref) => ref.tell(msg, sender)
+        case None => sender ! ServiceRegistryFailure(msg.serviceName)
+      }
+  }
+}
--- a/engine/src/test/scala/cromwell/services/KeyValueServiceActorSpec.scala
+++ b/engine/src/test/scala/cromwell/services/KeyValueServiceActorSpec.scala
@@ -1,0 +1,152 @@
+package cromwell.services
+
+import akka.pattern.ask
+import com.typesafe.config.ConfigFactory
+import cromwell.engine.WorkflowSourceFiles
+import cromwell.services.KeyValueServiceActor._
+
+class KeyValueServiceActorSpec extends CromwellServicesSpec {
+
+  val cromwellConfig = ConfigFactory.parseString(
+    s"""services: {
+        |  KeyValue: {
+        |    class: "cromwell.services.KeyValueServiceActor"
+        |    config {
+        |      option1: "value1"
+        |    }
+        |  }
+        |}
+     """.stripMargin
+  )
+
+  val sources = WorkflowSourceFiles(
+    wdlSource="""task a {command{}}
+                |workflow w {
+                |  call a
+                |  call a as b
+                |}
+              """.stripMargin,
+    inputsJson="{}",
+    workflowOptionsJson="{}"
+  )
+
+  it should "insert a key/value" in {
+    val descriptor = makeWorkflowDescriptor(sources)
+    val kvActor = makeKeyValueActor(cromwellConfig, descriptor)
+    val callKey = getBackendJobDescriptorKey(descriptor, "a")
+    val put = KvPut(KvPair(ScopedKey(callKey, "k"), Option("v")))
+    val get = KvGet(ScopedKey(callKey, "k"))
+
+    (for {
+      _ <- dataAccess.createWorkflow(descriptor, Nil, descriptor.namespace.workflow.calls, localBackend)
+      put0 <- ask(kvActor, put).mapTo[KvResponse]
+      _ = put0 shouldEqual KvPutSuccess(put)
+      get0 <- ask(kvActor, get).mapTo[KvResponse]
+      _ = get0 shouldEqual KvPair(ScopedKey(callKey, "k"), Option("v"))
+    } yield ()).futureValue
+  }
+
+  it should "return error if key doesn't exist" in {
+    val descriptor = makeWorkflowDescriptor(sources)
+    val kvActor = makeKeyValueActor(cromwellConfig, descriptor)
+    val callKey = getBackendJobDescriptorKey(descriptor, "a")
+    val get = KvGet(ScopedKey(callKey, "k"))
+
+    (for {
+      _ <- dataAccess.createWorkflow(descriptor, Nil, descriptor.namespace.workflow.calls, localBackend)
+      get0 <- ask(kvActor, get).mapTo[KvKeyLookupFailed]
+      _ = get0.action shouldEqual get
+    } yield ()).futureValue
+  }
+
+  it should "be able to overwrite values" in {
+    val descriptor = makeWorkflowDescriptor(sources)
+    val kvActor = makeKeyValueActor(cromwellConfig, descriptor)
+    val callAKey = getBackendJobDescriptorKey(descriptor, "a")
+    val putA = KvPut(KvPair(ScopedKey(callAKey, "k"), None))
+    val putB = KvPut(KvPair(ScopedKey(callAKey, "k"), Option("v")))
+    val get = KvGet(ScopedKey(callAKey, "k"))
+
+    (for {
+      _ <- dataAccess.createWorkflow(descriptor, Nil, descriptor.namespace.workflow.calls, localBackend)
+      put0 <- ask(kvActor, putA).mapTo[KvPutSuccess]
+      _ = put0 shouldEqual KvPutSuccess(putA)
+      get0 <- ask(kvActor, get).mapTo[KvPair]
+      _ = get0 shouldEqual KvPair(ScopedKey(callAKey, "k"), None)
+      put1 <- ask(kvActor, putB).mapTo[KvPutSuccess]
+      _ = put1 shouldEqual KvPutSuccess(putB)
+      get1 <- ask(kvActor, get).mapTo[KvPair]
+      _ = get1 shouldEqual KvPair(ScopedKey(callAKey, "k"), Option("v"))
+    } yield ()).futureValue
+  }
+
+  it should "be able to store NULL values" in {
+    val descriptor = makeWorkflowDescriptor(sources)
+    val kvActor = makeKeyValueActor(cromwellConfig, descriptor)
+    val callAKey = getBackendJobDescriptorKey(descriptor, "a")
+    val put = KvPut(KvPair(ScopedKey(callAKey, "k"), None))
+    val get = KvGet(ScopedKey(callAKey, "k"))
+
+    (for {
+      _ <- dataAccess.createWorkflow(descriptor, Nil, descriptor.namespace.workflow.calls, localBackend)
+      put0 <- ask(kvActor, put).mapTo[KvPutSuccess]
+      _ = put0 shouldEqual KvPutSuccess(put)
+      get0 <- ask(kvActor, get).mapTo[KvPair]
+      _ = get0 shouldEqual KvPair(ScopedKey(callAKey, "k"), None)
+    } yield ()).futureValue
+  }
+
+  it should "partition keys by call" in {
+    val descriptor = makeWorkflowDescriptor(sources)
+    val kvActor = makeKeyValueActor(cromwellConfig, descriptor)
+    val callAKey = getBackendJobDescriptorKey(descriptor, "a")
+    val callBKey = getBackendJobDescriptorKey(descriptor, "b")
+    val putA = KvPut(KvPair(ScopedKey(callAKey, "k"), Option("value_a")))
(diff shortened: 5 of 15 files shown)


backend/src/main/scala/cromwell/backend/package.scala
```diff
@@ -7,16 +7,27 @@ import wdl4s.values.WdlValue
...
+    def tag: String
+  }
+
   /**
     * For uniquely identifying a job which has been or will be sent to the backend.
     */
-  case class BackendJobDescriptorKey(callFqn: String, index: Option[Int], attempt: Int)
+  case class BackendJobDescriptorKey(call: Call, index: Option[Int], attempt: Int) extends JobKey {
+    def scope = call
+    def tag = s"${call.fullyQualifiedName}:$index:$attempt"
```
scottfrazer: yeah yeah yeah... these can be vals...

engine/src/main/scala/cromwell/services/KeyValueServiceActor.scala
```diff
@@ -0,0 +1,69 @@
...
+
+object KeyValueServiceActor {
+  trait KvMessage
+  trait KvAction extends KvMessage
+  trait KvResponse extends KvMessage
+
+  case class KvPut(jobDescriptor: BackendJobDescriptor, key: String, value: String) extends KvAction
+  case class KvGet(jobDescriptor: BackendJobDescriptor, key: String) extends KvAction
+
+  case class KvPair(jobDescriptor: BackendJobDescriptor, key: String, value: Option[String]) extends KvResponse
```
geoffjentry: Under what circumstance(s) would `value` be `None`?

engine/src/main/scala/cromwell/services/KeyValueServiceActor.scala
```diff
@@ -0,0 +1,69 @@
...
+                                serviceConfig: Config,
+                                entireConfig: Config) extends Actor {
+  private implicit val ec = context.dispatcher
+
+  def receive = {
+    case action: KvGet => respond(sender(), action, doGet(action))
+    case action: KvPut => respond(sender(), action, doPut(action))
+  }
+
+  private def respond(replyTo: ActorRef, action: KvAction, future: Future[KvResponse]): Unit = {
```
geoffjentry: I think `future` would be better named something like `response`

engine/src/main/scala/cromwell/services/ServiceRegistryActor.scala
```diff
@@ -0,0 +1,46 @@
...
+  }
+}
+
+case class ServiceRegistryActor(dataAccess: DataAccess, globalConfig: Config) extends Actor {
+  val services: Map[String, ActorRef] = {
+    val serviceNamesToConfigStanzas = globalConfig.getObject("services").entrySet.asScala.map(x => x.getKey -> x.getValue).toMap
+    serviceNamesToConfigStanzas map {
+      case (serviceName, config: ConfigObject) =>
+        serviceName -> makeActor(config.toConfig)
+      case (serviceName, config) =>
```
geoffjentry: `config` can be `_` I believe?

engine/src/main/scala/cromwell/services/ServiceRegistryActor.scala
```diff
@@ -0,0 +1,46 @@
...
+        throw new Exception("invalid configuration")
```
geoffjentry: add `serviceName` to the exception message

geoffjentry: What's supervising this actor? what's handling the exception thrown here?

engine/src/main/scala/cromwell/services/ServiceRegistryActor.scala
```diff
@@ -0,0 +1,46 @@
...
+    }
+  }
+
+  private def makeActor(serviceStanza: Config): ActorRef = {
+    val serviceConfigStanza = serviceStanza.getConfigOption("config").getOrElse(ConfigFactory.parseString(""))
+    val className = serviceStanza.getStringOption("class").getOrElse(throw new Exception("invalid config"))
```
geoffjentry: Can the exception message be more informative?

engine/src/main/scala/cromwell/services/ServiceRegistryActor.scala
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  def receive = {
+    case Send(serviceName, message) =>
+      services.get(serviceName) match {
+        case Some(ref) => ref.tell(message, sender)
+        case None => sender ! SendFailure(serviceName)
+      }
+  }
+}
```
geoffjentry: At some point @kshakir would ask that a `\n` be added at the end

engine/src/main/scala/cromwell/services/KeyValueServiceActor.scala
```diff
@@ -0,0 +1,69 @@
...
+
+  private def doGet(get: KvGet): Future[KvResponse] = {
+    val executionInfos = dataAccess.getExecutionInfos(
+      get.jobDescriptor.descriptor.id,
+      get.jobDescriptor.key.call,
+      get.jobDescriptor.key.attempt
+    )
+
+    executionInfos map {
+      case infos => infos.find(_.key == get.key) match {
```
Horneth: Maybe we could have a slick query that filters by key as well ?

engine/src/main/scala/cromwell/services/ServiceRegistryActor.scala
```diff
@@ -0,0 +1,46 @@
...
+
+import scala.collection.JavaConverters._
+
+object ServiceRegistryActor {
+  case class Send(serviceName: String, message: Any)
+  case class SendFailure(serviceName: String)
+
+  def props(dataAccess: DataAccess, config: Config) = {
+    Props(ServiceRegistryActor(dataAccess, config))
+  }
```
Horneth: We might want to make all this private and create a singleton actor instance accessible from here to avoid accidentally creating several registries ?

engine/src/main/scala/cromwell/services/KeyValueServiceActor.scala
scottfrazer: We currently allow `NULL` in the database for the corresponding column.  I forgot to make `KvPut` take an `value: Option[String]`

backend/src/main/scala/cromwell/backend/package.scala
```diff
@@ -7,16 +7,27 @@ import wdl4s.values.WdlValue
 
 package object backend {
 
+  trait JobKey {
+    def scope: Scope
```
cjllanwarne: I don't know which to change but it seems strange to have a `Job` key contain a `Scope`. For the sake of consistency, could we either call it `ScopeKey` or make it contain a `Call`?

engine/src/main/scala/cromwell/services/KeyValueServiceActor.scala
geoffjentry: My real point was that ignoring what the DB currently has (as we could change that) - what use case(s) are there for something inserting a blank value into a KV store. However thinking about it over the weekend I could see value where a client was modeling an Option or something like that, or simply using a key as a flag.

backend/src/main/scala/cromwell/backend/package.scala
scottfrazer: It seems the reason why it's a `Scope` is because there's a [`ScatterKey`](https://github.com/broadinstitute/cromwell/blob/494aa86713ee3a19c3d944355672862efd292e3f/engine/src/main/scala/cromwell/engine/workflow/ExecutionStoreKey.scala#L35)

The only reason why I (slightly) prefer `JobKey` over `ScopeKey` is that implementers of this trait typically use the word "job" to describe themselves.  But I can see good arguments for using `ScopeKey`.  Anybody else have a preference?

engine/src/main/scala/cromwell/services/ServiceRegistryActor.scala
cjllanwarne: Is he really that much of a tyrant though?

engine/src/main/scala/cromwell/services/KeyValueServiceActor.scala
scottfrazer: Yeah I guess so... let me rebase on top of the `deadlocks_and_upserts` first so I don't create even more merge conflicts

engine/src/main/scala/cromwell/services/ServiceRegistryActor.scala
```diff
@@ -0,0 +1,46 @@
...
+    val className = serviceStanza.getStringOption("class").getOrElse(
+      throw new Exception(s"invalid configuration for service $serviceName: missing 'class' definition")
+    )
+    val method = Class.forName(className).getMethod("props", classOf[DataAccess], classOf[Config], classOf[Config])
+    val props = method.invoke(None, dataAccess, serviceConfigStanza, globalConfig).asInstanceOf[Props]
+    context.actorOf(props)
+  }
+
+  def receive = {
+    case Send(serviceName, message) =>
```
cjllanwarne: Just a thought which might be horrible... It might be possible to remove this awkward "Send" message by including the serviceName in the message. You could have:
```
trait ServiceRegistryMessage { def serviceName: String }
```

And then every service's messages could just have (eg):
```
sealed trait KvMessage extends ServiceRegistryMessage { val serviceName = "KeyValue" }
```

engine/src/main/scala/cromwell/services/ServiceRegistryActor.scala
```diff
@@ -0,0 +1,46 @@
...
+import akka.actor.{Actor, ActorRef, Props}
+import com.typesafe.config.{ConfigFactory, Config, ConfigObject}
+import cromwell.engine.db.DataAccess
+import cromwell.services.ServiceRegistryActor._
+import lenthall.config.ScalaConfig._
+
+import scala.collection.JavaConverters._
+
+object ServiceRegistryActor {
+  case class Send(serviceName: String, message: Any)
```
geoffjentry: What about having a `ServiceMessage` (or somesuch) type and having `message` accept that. Clients of the service registry would have access to it and could extend as appropriate.

engine/src/main/scala/cromwell/services/ServiceRegistryActor.scala
geoffjentry: This reminds me that I've been waffling over the weekend on if having the registry double as an api gateway is really the best idea. Of course every time i decide 'no', i bring myself back. Discussions like this are part of why i think 'no', as it's just adding a lot of complication in a single spot.

engine/src/test/scala/cromwell/services/KeyValueServiceActorSpec.scala
```diff
@@ -0,0 +1,186 @@
...
+      _ = r2 shouldEqual KvPutSuccess(putA1)
+      r3 <- ask(kv0, getA0).mapTo[KvPair]
+      _ = r3 shouldEqual KvPair(callA0, "k", Option("v"))
+      r4 <- ask(kv1, getA1).mapTo[KvPair]
+      _ = r4 shouldEqual KvPair(callA1, "k", Option("v"))
+    } yield ()
+
+    Await.result(future, 2.seconds)
+  }
+}
```
cjllanwarne: @geoffjentry will at some point comment that @kshakir will at some point comment about newlines here :)

engine/src/test/scala/cromwell/services/KeyValueServiceActorSpec.scala
```diff
@@ -0,0 +1,186 @@
...
+  val actorSystem = workflowManagerSystem.actorSystem
+  val dataAccess = new SlickDataAccess(ConfigFactory.parseString(
+    s"""
+       |db.url = "jdbc:hsqldb:mem:$${slick.uniqueSchema};shutdown=false;hsqldb.tx=mvcc"
+       |db.driver = "org.hsqldb.jdbcDriver"
+       |driver = "slick.driver.HsqldbDriver$$"
+       |schema.manager = slick
+     """.stripMargin
+  ))
+  val localBackend = LocalBackend(actorSystem)
```
cjllanwarne: I think this is necessary because of foreign key constraints in the database? 

Just pointing out that it's weird that for us to test our key/value store, we need to create a backend instance :-/

engine/src/main/scala/cromwell/services/ServiceRegistryActor.scala
scottfrazer: okay... whatever'll get this merged faster

cjllanwarne: Yeah, I'm not 100% sold on "make it an actor" tbh... I could go either way

geoffjentry: @cjllanwarne at no point was I questioning if the registry should be an actor. Just if it should also double as an API gateway

scottfrazer: @cjllanwarne but over-engineering is just SO MUCH FUN!

cjllanwarne:
:+1:

scottfrazer:
Wheel chooses @kshakir

engine/src/main/scala/cromwell/services/ServiceRegistryActor.scala
```diff
@@ -0,0 +1,50 @@
...
+    serviceNamesToConfigStanzas map {
+      case (serviceName, config: ConfigObject) => serviceName -> makeActor(serviceName, config.toConfig)
+      case (serviceName, _) => throw new Exception(s"invalid configuration for service $serviceName")
+    }
+  }
+
+  private def makeActor(serviceName: String, serviceStanza: Config): ActorRef = {
+    val serviceConfigStanza = serviceStanza.getConfigOption("config").getOrElse(ConfigFactory.parseString(""))
+    val className = serviceStanza.getStringOption("class").getOrElse(
+      throw new Exception(s"invalid configuration for service $serviceName: missing 'class' definition")
```
kshakir: `.getStringOption(a).getOrElse(b)` should be `.getStringOr(a, b)`. Not sure why lenthall doesn't have a `.getConfigOr`.

Also, "invalid" -> "Invalid"

engine/src/test/scala/cromwell/services/KeyValueServiceActorSpec.scala
```diff
@@ -0,0 +1,186 @@
...
+import cromwell.engine.WorkflowSourceFiles
+import cromwell.engine.backend.local.LocalBackend
+import cromwell.engine.db.slick.SlickDataAccess
+import cromwell.services.KeyValueServiceActor._
+import org.scalatest.{FlatSpec, Matchers}
+
+import scala.concurrent.Await
+import scala.concurrent.duration._
+
+class KeyValueServiceActorSpec extends FlatSpec with Matchers with CromwellServicesTest {
```
kshakir: Mix in `BeforeAndAfterAll`, then clean up after yourself. Each test like this that leaks an actor system (below) is leaving behind a pool of threads among other resources.

```scala
  override protected def afterAll() = {
    testWorkflowManagerSystem.shutdownTestActorSystem()
    super.afterAll()
  }
```

engine/src/test/scala/cromwell/services/KeyValueServiceActorSpec.scala
```diff
@@ -0,0 +1,186 @@
...
+
+    val future = for {
+      _ <- dataAccess.createWorkflow(descriptor, Nil, descriptor.namespace.workflow.calls, localBackend)
+      r0 <- ask(kvActor, put).mapTo[KvResponse]
+      _ = r0 shouldEqual KvPutSuccess(put)
+      r1 <- ask(kvActor, get).mapTo[KvResponse]
+      _ = r1 shouldEqual KvPair(callKey, "k", Option("v"))
+    } yield ()
+
+    Await.result(future, 2.seconds)
```
kshakir: These are ok if you prefer `Await.result`, but I've leaned towards using the trait `ScalaFutures`, as its implicit `.futureValue` theoretically allows test timeouts to be "[flexibly configured](http://doc.scalatest.org/2.2.4/index.html#org.scalatest.concurrent.Futures)" on slower systems. For example, Travis isn't as zippy as our local machines.

engine/src/test/scala/cromwell/services/KeyValueServiceActorSpec.scala
```diff
@@ -0,0 +1,186 @@
...
+import cromwell.services.KeyValueServiceActor._
+import org.scalatest.{FlatSpec, Matchers}
+
+import scala.concurrent.Await
+import scala.concurrent.duration._
+
+class KeyValueServiceActorSpec extends FlatSpec with Matchers with CromwellServicesTest {
+  val workflowManagerSystem = new TestWorkflowManagerSystem
+  val actorSystem = workflowManagerSystem.actorSystem
+  val dataAccess = new SlickDataAccess(ConfigFactory.parseString(
```
kshakir: Are you testing some specific variation of data access? Otherwise, convention should be to use the "main" data access code, that for now is the singleton.

engine/src/main/scala/cromwell/services/KeyValueServiceActor.scala
```diff
@@ -0,0 +1,72 @@
...
+  trait KvMessage
+  trait KvAction extends KvMessage with ServiceRegistryMessage {
+    def serviceName = "KeyValue"
+  }
+  trait KvResponse extends KvMessage
+
+  case class KvPut(jobDescriptor: BackendJobDescriptor, key: String, value: Option[String]) extends KvAction
+  case class KvGet(jobDescriptor: BackendJobDescriptor, key: String) extends KvAction
+
+  case class KvPair(jobDescriptor: BackendJobDescriptor, key: String, value: Option[String]) extends KvResponse
```
geoffjentry: A `KvPair` has 3 elements?

Also if this is identical to the contents of a `KvPut`, would it make sense for the `KvPut` to have a single element of a `KvPair`?

engine/src/test/scala/cromwell/services/ServiceRegistryActorSpec.scala
```diff
@@ -0,0 +1,65 @@
...
+import cromwell.engine.WorkflowSourceFiles
+import cromwell.engine.backend.local.LocalBackend
+import cromwell.engine.db.slick.SlickDataAccess
+import cromwell.services.KeyValueServiceActor._
+import org.scalatest.{FlatSpec, Matchers}
+
+import scala.concurrent.Await
+import scala.concurrent.duration._
+
+class ServiceRegistryActorSpec extends FlatSpec with Matchers with CromwellServicesTest {
```
kshakir: Same comments about data access copies, Await vs futureValue, cleaning up your thread pool.

engine/src/test/scala/cromwell/services/CromwellServicesTest.scala
```diff
@@ -0,0 +1,45 @@
...
+      descriptor.namespace,
+      Map.empty,
+      WorkflowOptions(JsObject())
+    )
+    val key = BackendJobDescriptorKey(call, None, 1)
+    BackendJobDescriptor(backendWorkflowDescriptor, key, Map.empty)
+  }
+
+  protected def makeWorkflowDescriptor(sources: WorkflowSourceFiles): WorkflowDescriptor = {
+    val workflowId = WorkflowId(UUID.randomUUID())
```
kshakir: `WorkflowId.randomId()`

engine/src/main/scala/cromwell/services/KeyValueServiceActor.scala
mcovarr: should there be a `ScopedKey` type that combines the first two parameters?

engine/src/test/scala/cromwell/services/CromwellServicesTest.scala
```diff
@@ -0,0 +1,45 @@
...
+import cromwell.engine.WorkflowSourceFiles
+import cromwell.engine.backend.{WorkflowDescriptor, WorkflowDescriptorBuilder}
+import cromwell.engine.db.DataAccess
+import spray.json.JsObject
+
+trait CromwellServicesTest extends WorkflowDescriptorBuilder {
+  val actorSystem: ActorSystem
+
+  protected def getBackendJobDescriptorKey(descriptor: WorkflowDescriptor, callName: String): BackendJobDescriptor = {
+    val call = descriptor.namespace.workflow.calls.find(_.unqualifiedName == callName).get
```
kshakir: _This is ~~PTSD~~ (Current)TSD dealing with merging FinalCallKey to Backend/CallKey._

How about a more descriptive error, `.getOrElse(throw new Exception("you messed up…"))`.

kshakir:
First pass complete. Main code looks good. Just a few suggestions / preference for test code.

engine/src/main/scala/cromwell/services/ServiceRegistryActor.scala
```diff
@@ -0,0 +1,51 @@
...
+  trait ServiceRegistryMessage {
+    def serviceName: String
+  }
+
+  def props(dataAccess: DataAccess, config: Config) = {
+    Props(ServiceRegistryActor(dataAccess, config))
+  }
+}
+
+case class ServiceRegistryActor(dataAccess: DataAccess, globalConfig: Config) extends Actor {
```
gauravs90: IMHO, this class should not be getting a DataAccess as a parameter. I see that it's required by the KV Service actor in the constructor, but when we start adding other service actor implementations, we will begin to notice a discrete seemingly unrelated params being passed onto here as a constructor. 

I feel this actor should receive a `Seq[Props]` (or a Seq of some ADT with props) as a param, and all it should do is create the 'services` map of String -> ActorRef using those props. The responsibility of reading the config and creating the props object can can be delegated to something like a Factory, which will create these props and know the args required by the services.

engine/src/main/scala/cromwell/services/KeyValueServiceActor.scala
gauravs90: Are we going to create a separate service for the Metadata stuff? Or is this class supposed to be reused there?

engine/src/test/scala/cromwell/services/CromwellServicesTest.scala
```diff
@@ -0,0 +1,63 @@
...
+import cromwell.engine.backend.local.LocalBackend
+import cromwell.engine.backend.{WorkflowDescriptor, WorkflowDescriptorBuilder}
+import cromwell.engine.db.DataAccess
+import org.scalatest.{Matchers, FlatSpec, BeforeAndAfterAll}
+import scala.concurrent.duration._
+import org.scalatest.time.{Seconds, Span, Millis}
+import spray.json.JsObject
+import org.scalatest.concurrent.ScalaFutures
+
+trait CromwellServicesTest extends FlatSpec with Matchers with BeforeAndAfterAll with WorkflowDescriptorBuilder with ScalaFutures {
```
gauravs90: `CromwellServicesTest` -> `CromwellServicesSpec`

engine/src/main/scala/cromwell/services/ServiceRegistryActor.scala
scottfrazer: Not that I disagree... I just think we should not over-engineer this a whole bunch and let it evolve a little more...  is this sort of what you want?

```scala
object ServiceRegistryActor {
  case class ServiceRegistryFailure(serviceName: String)
  trait ServiceRegistryMessage {
    def serviceName: String
  }

  def props(config: Config) = {
    Props(ServiceRegistryActor(config))
  }

  def serviceProps(globalConfig: Config): Map[String, Props] = {
    val serviceNamesToConfigStanzas = globalConfig.getObject("services").entrySet.asScala.map(x => x.getKey -> x.getValue).toMap
    serviceNamesToConfigStanzas map {
      case (serviceName, config: ConfigObject) => serviceName -> makeServiceProps(serviceName, globalConfig, config.toConfig)
      case (serviceName, _) => throw new Exception(s"Invalid configuration for service $serviceName")
    }
  }

  private def makeServiceProps(serviceName: String, globalConfig: Config, serviceStanza: Config): Props = {
    val serviceConfigStanza = serviceStanza.getConfigOption("config").getOrElse(ConfigFactory.parseString(""))
    val className = serviceStanza.getStringOr(
      "class",
      throw new Exception(s"Invalid configuration for service $serviceName: missing 'class' definition")
    )
    val method = Class.forName(className).getMethod("props", classOf[DataAccess], classOf[Config], classOf[Config])
    method.invoke(None, DataAccess.globalDataAccess, serviceConfigStanza, globalConfig).asInstanceOf[Props]
  }
}

case class ServiceRegistryActor(globalConfig: Config) extends Actor {
  import ServiceRegistryActor._

  val services: Map[String, ActorRef] = serviceProps(globalConfig) map {
    case (name, props) => name -> context.actorOf(props)
  }

  def receive = {
    case msg: ServiceRegistryMessage =>
      services.get(msg.serviceName) match {
        case Some(ref) => ref.tell(msg, sender)
        case None => sender ! ServiceRegistryFailure(msg.serviceName)
      }
  }
}
```

gauravs90: I don't know about other people here, but personally I prefer this one more than the previous one. Some further improvements can be: 
1.) To use Props's reflection thingy here instead of explicit method invocation (something like [this](https://github.com/broadinstitute/cromwell-backend/blob/c7440eee7ddf34599ff5abff52e6d3e44a8efa0d/src/main/scala/cromwell/backend/DefaultBackendFactory.scala#L15) that I previously found simpler to use) 

2.) Have that `DataAccess.globalDataAccess` reference directly in the KVServiceActor instead of passing it in the props (since ideally in a distant utopia, that actor should be responsible for creating it's own DB access stuff)

engine/src/main/scala/cromwell/services/KeyValueServiceActor.scala
```diff
@@ -0,0 +1,72 @@
...
+import scala.util.{Failure, Success}
+
+object KeyValueServiceActor {
+  trait KvMessage
+  trait KvAction extends KvMessage with ServiceRegistryMessage {
+    def serviceName = "KeyValue"
+  }
+  trait KvResponse extends KvMessage
+
+  case class KvPut(jobDescriptor: BackendJobDescriptor, key: String, value: Option[String]) extends KvAction
```
geoffjentry: To throw a monkeywrench into all of this - why are we namespacing the KV store on a backend job descriptor? What if I want to use it for something else?

engine/src/main/scala/cromwell/services/KeyValueServiceActor.scala
geoffjentry: @gauravs90 I was going to write up some thoughts on the metadata thing

engine/src/main/scala/cromwell/services/KeyValueServiceActor.scala
gauravs90: @geoffjentry That would have been my next question in case the answer to my previous question down below would have been to 'Re-use' 😄

geoffjentry: My own thinking which brought me to asking that has gone down an even further rabbit hole, but the long and short of it is that at the very least I think this is too limiting. I still need to get my own thoughts together on anything beyond that statement.

scottfrazer: Honestly I'm kind of sick of refactoring this into oblivion.  I already feel zero ownership of this code, does somebody who feels at least *some* ownership of this code want to just take it over so you don't have to tell me to change everything line-by-line?  I was just implementing what you guys laid out anyway. AND I asked Chris and Miguel if they liked the design before I pushed it and *this is the design we all liked*.  I can't please everybody, sorry.

geoffjentry: I think it is fine for now, I just want to have a larger conversation about the specificity of the KV store. That doesn't block merging this PR but the overall scheme might need to change afterwards

scottfrazer: also the reason why I passed dataAccess into the KV service is because Khalid doesn't want `globalDataAccess` to be globally accessed anymore, so I was just trying to facilitate that.

kshakir: Appreciate the thought. :)

Still, current Cromwell data access convention is to use the singleton, so until a full replacement arrives, I'm fine with continuing to use the singleton.

engine/src/main/scala/cromwell/services/ServiceRegistryActor.scala
scottfrazer: in my mind, it's the exact same thing... but if it'll make you happy

engine/src/main/scala/cromwell/services/KeyValueServiceActor.scala
scottfrazer: I don't know, I'm dragging my feet because I *really* don't agree but I feel if I try to make my case I'll risk feeling even more alienated (considering it's now 3 against 1), so as long as I don't have to take ownership of this code I'll feel *much* better about these changes...

Just so you know though... we're now going to have this interface for constructing `KvPut`s and `KvGet`s:

```scala
KvPut(KvPair(ScopedKey(callBKey, "k"), Option("value_b")))
KvGet(ScopedKey(callAKey, "k"))
```

geoffjentry: @scottfrazer I was asking questions, not giving directives. I just want to know why things are the way they are. That *might* lead to requests for changes or it might be that I just don't understand the underlying scheme and am missing something.

scottfrazer: You're not missing anything.  The reason why I wrote it the way I did initially is that I perceived that as the simplest approach given the constraints I had to work with.  That's all.

I've already pushed the new design... I couldn't help but be reminded of the quote "a camel is a horse designed by a committee" :)

scottfrazer:
sorry @cjllanwarne and @kshakir (and whoever else cares) the code changed enough that I need it re-reviewed...

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -459,6 +460,19 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
     runTransaction(action)
   }
 
+  override def getExecutionInfoByKey(workflowId: WorkflowId, call: Call, attempt: Int, key: String)
+                                    (implicit ec: ExecutionContext): Future[Option[Option[String]]] = {
```
scottfrazer: Oops, I forgot to add a test for this (this is what @Horneth requested)

engine/src/test/scala/cromwell/services/CromwellServicesTest.scala
```diff
@@ -0,0 +1,64 @@
...
+import cromwell.engine.db.DataAccess
+import cromwell.engine.db.slick.SlickDataAccess
+import org.scalatest.concurrent.ScalaFutures
+import org.scalatest.time.{Millis, Seconds, Span}
+import org.scalatest.{BeforeAndAfterAll, FlatSpec, Matchers}
+import spray.json.JsObject
+
+import scala.concurrent.duration._
+
+trait CromwellServicesTest extends FlatSpec with Matchers with BeforeAndAfterAll with WorkflowDescriptorBuilder with ScalaFutures {
```
kshakir: Previous comment mentioned that our current convention for test code is `*Spec` not `*Test`.

engine/src/test/scala/cromwell/services/ServiceRegistryActorSpec.scala
```diff
@@ -0,0 +1,46 @@
...
+    val kvGet = KvGet(ScopedKey(callKey, "k"))
+
+    val future = for {
+      _ <- dataAccess.createWorkflow(descriptor, Nil, descriptor.namespace.workflow.calls, localBackend)
+      r2 <- ask(serviceRegistryActor, kvGet).mapTo[KvKeyLookupFailed]
+      _ = r2.action shouldEqual kvGet
+      r0 <- ask(serviceRegistryActor, kvPut).mapTo[KvResponse]
+      _ = r0 shouldEqual KvPutSuccess(kvPut)
+      r1 <- ask(serviceRegistryActor, kvGet).mapTo[KvResponse]
+      _ = r1 shouldEqual KvPair(ScopedKey(callKey, "k"), Option("v"))
```
kshakir: Not a huge deal.

The `r2`, `r0`, `r1` names aren't descriptive. But I'm more curious (and others may in the [future](http://stackoverflow.com/questions/876089/who-wrote-this-programing-saying-always-code-as-if-the-guy-who-ends-up-maintai))… why the 2,0,1 order?

scottfrazer: Because I refactored this PR like 10 times.  I'm surprised it isn't worse!

geoffjentry:
LGTM

geoffjentry:
@scottfrazer Make sure to throw names on all of the actor creations (I think I saw 3)

cjllanwarne:
That's weird - @kshakir 's :+1: didn't count towards the pull approval? I can see two thumbs on this now

PR #690, opened 2016-04-08
Refactored DataAccess retries.

kshakir:
Extracted the retries out of SlickDataAccess.
Having trouble producing a true positive deadlock test that this code fixes, so left that out and moving on for now.

### DIFF
--- a/engine/src/main/scala/cromwell/engine/db/DataAccess.scala
+++ b/engine/src/main/scala/cromwell/engine/db/DataAccess.scala
@@ -1,6 +1,7 @@
 package cromwell.engine.db
 
 import akka.actor.ActorSystem
+import akka.pattern.after
 import cromwell.core.{CallOutputs, WorkflowId}
 import cromwell.engine.ExecutionStatus.ExecutionStatus
 import cromwell.engine._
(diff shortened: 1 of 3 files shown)


engine/src/main/scala/cromwell/engine/db/DataAccess.scala
```diff
@@ -46,8 +72,25 @@ trait DataAccess extends AutoCloseable {
...
+   */
+  def upsertExecutionInfo(workflowId: WorkflowId,
+                          callKey: BackendCallKey,
                           keyValues: Map[String, Option[String]],
-                          actorSystem: ActorSystem): Future[Unit]
+                          actorSystem: ActorSystem): Future[Unit] = {
+    implicit val system = actorSystem
+    implicit val ec = actorSystem.dispatcher
+    withRetry(upsertExecutionInfo(workflowId, callKey, keyValues))
+  }
```
scottfrazer: Do we need two `upsertExecutionInfo` methods here?

scottfrazer: Oh, this one calls the other one.

scottfrazer:
I'm (obviously) not terribly good at spotting problems in the logic, but it seems good to me... I share your desire to make the interface a little better somehow... But perhaps that's a ticket for later?

kshakir:
I'm definitely ok with any "ticket(s) for later", as long as Miguel's comments in the parent PR are addressed. Retries should be _outside_ of the `Await.result()`, and by association should not retry within `runTransaction`.

This PR is only one example that does an outer retry, choosing implementation details such as:

1. Communicating "whoops, please retry" via a functional parameter `shouldRetry`, instead of wrapping the JDBC exception to our own expected-exception-for-retries, such as ~~`cromwell.actor.TransientException`~~ `lethall.actor.TransientException`.
2. Invoking `withRetry` via an overloaded `DataAccess` method, instead having the caller invoke `withRetry`.

In general, if a similar but different version of `withRetry` outside `runTransaction` appears, it'll get a thumbsup from me.

geoffjentry:
Can someone describe what exactly is being described as a TBD and what parts are for this project? I thought I knew but don't think o do

engine/src/main/scala/cromwell/engine/db/DataAccess.scala
```diff
@@ -9,19 +10,44 @@ import cromwell.engine.db.DataAccess.{ExecutionKeyToJobKey, WorkflowExecutionAnd
...
+  private def withRetry[A](f: => Future[A], delay: FiniteDuration, retries: Int)
+                          (shouldRetry: (Throwable) => Boolean)
+                          (implicit actorSystem: ActorSystem): Future[A] = {
+    // TODO: May want to pass a separate ec, or is the actor system's dispatcher ok?
+    implicit val ec = actorSystem.dispatcher
+    f recoverWith {
+      case throwable if shouldRetry(throwable) && retries > 0 =>
+        log.warn(
+          s"""Transient exception detected: ${throwable.getMessage}
+              |Retrying in $delay ($retries retries remaining)""".stripMargin)
```
mcovarr: should this pass the throwable to `log.warn` to get the stack trace, or would that be too noisy?

engine/src/main/scala/cromwell/engine/db/DataAccess.scala
mcovarr: Perhaps the non-retrying abstract version of that method should be made `protected` so callers don't mistakenly invoke that.

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -191,24 +189,11 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
...
-            *
-            * "Normally, you must write your applications so that they are always
-            *  prepared to re-issue a transaction if it gets rolled back because of a deadlock."
-            */
-          log.warn(s"Transaction rollback detected.  Retrying in $delay ($retries retries remaining)")
-          after(delay, actorSystem.scheduler)(retry(f, delay, retries - 1))
-      }
+  override protected def isTransient(throwable: Throwable): Boolean = {
+    throwable match {
+      case e: SQLTransientException => true
```
Horneth: `case _:`

engine/src/main/scala/cromwell/engine/db/DataAccess.scala
scottfrazer: Maybe also make the message one-line so it's more grep-friendly?

scottfrazer:
@geoffjentry my understanding is that the logic here is what we want but the interface could be a little better, and a little more fully implemented.  Correct me if I'm wrong, but I *think* we want all `DataAccess` methods to automatically (and transparently) have deadlock retries.  If we were to do that for all methods in `DataAccess` using the scheme laid out in this PR, then we'd be duplicating all of the methods in `DataAccess`.

Perhaps we could construct `DataAccess` with an `ActorSystem`... though I can't remember if @kshakir objected to this for a reason that is currently escaping me.  We could also just make the whole thing an actor, or put an actor in front of `DataAccess`.

kshakir:
No current opinions on further refactoring DataAccess, as those discussions are outside of the parent PR, though happy participate in a tech talk if desired.

I will admit I "started it" by embellishing beyond simply inverting the futures within `runTransactionWithRollbackRetry`.

scottfrazer:
:+1: 

@kshakir do you think we can merge this into my branch and then just continue with the normal review process on that PR?  Or would you rather get the :+1:'s here first?

geoffjentry:
@scottfrazer Thanks.

PR #694, opened 2016-04-12
JobDescriptor merging checkpoint

kshakir:
`FinalCallJobDescriptor`, `JobDescriptor`, and `BackendCallJobDescriptors` merged into one.
BIG TODO: Final calls "run" on the local backend via a `FinalCallCopyHack` that runs them in process.
Moved final call tests from `WorkflowDescriptorSpec` to `FinalCallSpec`.

Creating a pull request to review current changes, while discussing best ways to execute copies out of process.


kshakir:
Choosing @mcovarr and @cjllanwarne for this initial review of Episode VII

engine/src/main/scala/cromwell/engine/callactor/CallActor.scala
```diff
@@ -87,13 +86,12 @@ object CallActor {
   val CallCounter = Monitor.minMaxCounter("calls-running")
 
   def props(backendCallDescriptor: BackendCallJobDescriptor): Props = Props(new BackendCallActor(backendCallDescriptor))
-  def props(finalCallDescriptor: FinalCallJobDescriptor) = Props(new FinalCallActor(finalCallDescriptor))
 }
 
 /** Actor to manage the execution of a single call. */
 // FIXME It feels like I shouldn't need to restate the type bounds of JobDescriptor's CallKey type variable.
```
mcovarr: comment should go

engine/src/main/scala/cromwell/engine/backend/Backend.scala
```diff
@@ -54,7 +55,12 @@ object Backend {
 
   def callRootPathWithBaseRoot(jobDescriptor: BackendCallJobDescriptor, baseRoot: String): Path = {
     import io._
-    val call = s"$CallPrefix-${jobDescriptor.key.scope.unqualifiedName}"
+    import FinalCall._
```
mcovarr: Maybe put the "enhancer" on the `descriptor` or `scope` rather than `descriptor.key.scope.name`?  It feels a bit strange to ask a name if it's a final call.

engine/src/main/scala/cromwell/engine/db/ExecutionInfosByExecution.scala
```diff
@@ -51,7 +52,12 @@ object ExecutionInfosByExecution {
   }
 
   def toWorkflowLogs(executionInfosByExecutions: Traversable[ExecutionInfosByExecution]): WorkflowLogs = {
-    executionInfosByExecutions groupBy { _.execution.callFqn } mapValues toAttemptedCallLogs filterNot {
+    import FinalCall._
+    executionInfosByExecutions groupBy {
+      _.execution.callFqn
+    } filterKeys {
+      !_.isFinalCall
+    } mapValues toAttemptedCallLogs filterNot {
```
mcovarr: feels a bit FP-soupy 😛

mcovarr:
@kshakir I'm curious as to your thoughts on how to move forward from here.  Not being able to do final call work on the local backend plus all the architectural changes subsequent to this ticket reluctantly make me think we should take a step back.

Maybe an in-JVM backend to which all the final call descriptors are assigned, whose Scopes encapsulate the work to be done and all relevant parameters including required auths?  That might be closer to what's currently on develop than what's here, but now we know from your experience we don't have an answer for how to do final call work on the local backend.

kshakir:
Let's chat IRL later today. To be clear, I'm totally fine mothballing this PR. In the context of PBE, conceivably this ticket was introduced so one could implement other final calls that run on a backend?

kshakir:
Leaving branch around, but closing this PR while we revisit the original issue.

PR #704, opened 2016-04-14
Refactoring restarts... move globalDataAccess away from backends.  closes #581

### LINKED ISSUE #581: Move + generalize DB access for restart / resume
Currently this happens in the JES backend, this needs to be made generic to backend type and put into engine.

scottfrazer:
This was the original proposed design, so the requirements might have changed since I wrote this weeks ago.

### DIFF
--- a/engine/src/main/scala/cromwell/engine/backend/Backend.scala
+++ b/engine/src/main/scala/cromwell/engine/backend/Backend.scala
@@ -11,7 +11,6 @@ import cromwell.engine.backend.jes.JesBackend
 import cromwell.engine.backend.local.LocalBackend
 import cromwell.engine.backend.runtimeattributes.{ContinueOnReturnCodeFlag, ContinueOnReturnCodeSet, CromwellRuntimeAttributes}
 import cromwell.engine.backend.sge.SgeBackend
-import cromwell.engine.db.DataAccess.ExecutionKeyToJobKey
 import cromwell.logging.WorkflowLogger
 import cromwell.util.docker.SprayDockerRegistryApiClient
 import org.slf4j.LoggerFactory
@@ -64,8 +63,6 @@ object Backend {
   }
 }
 
-trait BackendJobKey
-
 final case class AttemptedLookupResult(name: String, value: Try[WdlValue]) {
   def toPair = name -> value
 }
@@ -108,9 +105,10 @@ trait Backend {
   def engineFunctions(fileSystems: List[FileSystem], workflowContext: WorkflowContext): WorkflowEngineFunctions
 
   /**
-   * Do any work that needs to be done <b>before</b> attempting to restart a workflow.
-   */
-  def prepareForRestart(restartableWorkflow: WorkflowDescriptor)(implicit ec: ExecutionContext): Future[Unit]
+    * Assume JobKey was in a 'Running' state when the server was shut down.
+    */
+  def isResumable(key: JobKey, executionInfo: Map[String, Option[String]]): Boolean
+  def isRestartable(key: JobKey, executionInfo: Map[String, Option[String]]): Boolean
 
   /**
    * Return CallLogs which contains the stdout/stderr of the particular call
@@ -126,9 +124,6 @@ trait Backend {
   def assertWorkflowOptions(options: WorkflowOptions): Unit = {}
 
   private[backend] def backendClassString = backendType.toString.toLowerCase.capitalize + "Backend"
-
-  /** Default implementation assumes backends do not support resume, returns an empty Map. */
-  def findResumableExecutions(id: WorkflowId)(implicit ec: ExecutionContext): Future[Traversable[ExecutionKeyToJobKey]] = Future.successful(List.empty)
 
   def workflowLogger(descriptor: WorkflowDescriptor) = WorkflowLogger(
     backendClassString,
@@ -225,5 +220,5 @@ trait Backend {
 
   def execute(jobDescriptor: BackendCallJobDescriptor)(implicit ec: ExecutionContext): Future[ExecutionHandle]
 
-  def resume(descriptor: BackendCallJobDescriptor, jobKey: BackendJobKey)(implicit ec: ExecutionContext): Future[ExecutionHandle]
+  def resume(descriptor: BackendCallJobDescriptor, executionInfos: Map[String, Option[String]])(implicit ec: ExecutionContext): Future[ExecutionHandle]
 }
--- a/engine/src/main/scala/cromwell/engine/db/slick/ExecutionInfoComponent.scala
+++ b/engine/src/main/scala/cromwell/engine/db/slick/ExecutionInfoComponent.scala
@@ -1,5 +1,7 @@
 package cromwell.engine.db.slick
 
+import cromwell.engine.ExecutionStatus
+
 case class ExecutionInfo(executionId: Int,
                          key: String,
                          value: Option[String] = None,
@@ -42,6 +44,19 @@ trait ExecutionInfoComponent {
       execution <- executionInfo.execution
       workflowExecution <- execution.workflowExecution
       if workflowExecution.workflowExecutionUuid === id
+    } yield (execution, executionInfo)
+  )
+
+  val runningExecutionsAndExecutionInfosByWorkflowId = Compiled(
+    // Used by restart workflow code to get all executions that were
+    // in-flight when the engine is restarted.
+    (id: Rep[String]) => for {
+      executionInfo <- executionInfos
+      execution <- executionInfo.execution
+      workflowExecution <- execution.workflowExecution
+      if workflowExecution.workflowExecutionUuid === id
+      if execution.status === ExecutionStatus.Running.toString ||
+         execution.status === ExecutionStatus.Starting.toString
     } yield (execution, executionInfo)
   )
 
--- a/engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
+++ b/engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
@@ -12,7 +12,7 @@ import cromwell.engine.ExecutionIndex._
 import cromwell.engine.ExecutionStatus._
 import cromwell.engine._
 import cromwell.engine.backend._
-import cromwell.engine.db.DataAccess.{ExecutionKeyToJobKey, WorkflowExecutionAndAux}
+import cromwell.engine.db.DataAccess.WorkflowExecutionAndAux
 import slick.backend.DatabaseConfig
 import slick.driver.JdbcProfile
 import cromwell.engine.db._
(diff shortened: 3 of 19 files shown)


engine/src/main/resources/application.conf
```diff
@@ -192,7 +192,7 @@ database {
 
     mysql {
       db.url = "jdbc:mysql://localhost/cromwell_test"
-      db.user = "travis"
+      db.user = "root"
```
scottfrazer: I'll revert this

scottfrazer:
Wheel chooses @mcovarr and @Horneth... I know @mcovarr will be out next week so if you want me to re-roll let me know

mcovarr:
No problem, I can take this.

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -125,31 +126,83 @@ object WorkflowActor {
...
+      }
+
       for {
-        _ <- actor.backend.prepareForRestart(actor.workflow)
-        _ <- actor.dumpTables()
+        allExecutions <- globalDataAccess.getExecutions(actor.workflow.id)
+        _ <- isWorkflowRestartable(allExecutions)
+        runningExecutionsAndInfos <- globalDataAccess.runningExecutionsAndExecutionInfos(actor.workflow.id)
+        executionsAndInfosMap = runningExecutionsAndInfos.map(x => x.execution -> x.executionInfos).toMap
+        _ <- restartOrResume(actor, executionsAndInfosMap)
```
scottfrazer: Everything added above here was moved here [from the JesBackend](https://github.com/broadinstitute/cromwell/pull/704/files#diff-02fb95673814eb1a1aca084e078b5113L795).  It's definitely messy and even I have a hard time following it... I'll try to clean it up but any suggestions would be welcome

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -125,31 +126,83 @@ object WorkflowActor {
 
   case object Restart extends WorkflowActorMessage with StartMode {
```
geoffjentry: You didn't start this behavior although this PR made it large enough for me to notice. 

I don't feel that actor message types should themselves be balls of functionality - IMO the message type should be either a simple flag to do something or maybe a bit of state as well, and then the actor itself should be working out what to do with it.

scottfrazer: It's only here because that was the obvious place for it given our current design.  I'm happy to hear suggestions for alternatives or I can try to think of some myself... though considering there's this whole shadow world emerging, I wonder if it makes sense to figure this out now or figure it out when we move this to the shadow world

engine/src/main/scala/cromwell/engine/backend/Backend.scala
```diff
@@ -108,9 +105,10 @@ trait Backend {
   def engineFunctions(fileSystems: List[FileSystem], workflowContext: WorkflowContext): WorkflowEngineFunctions
 
   /**
-   * Do any work that needs to be done <b>before</b> attempting to restart a workflow.
-   */
-  def prepareForRestart(restartableWorkflow: WorkflowDescriptor)(implicit ec: ExecutionContext): Future[Unit]
+    * Assume JobKey was in a 'Running' state when the server was shut down.
+    */
+  def isResumable(key: JobKey, executionInfo: Map[String, Option[String]]): Boolean = false
+  def isRestartable(key: JobKey, executionInfo: Map[String, Option[String]]): Boolean = false
```
mcovarr: I'm a little uncomfortable about this being split into two methods.  The execution should return true for one and only one of these, but with two methods there isn't the sanity checking that there could be with one.

mcovarr:
wth is going on with Coveralls, we get green ticks for -0.2% now?

engine/src/main/scala/cromwell/engine/db/slick/ExecutionInfoComponent.scala
```diff
@@ -45,6 +47,16 @@ trait ExecutionInfoComponent {
     } yield (execution, executionInfo)
   )
 
+  val runningExecutionsAndExecutionInfosByWorkflowId = Compiled(
+    (id: Rep[String]) => for {
+      executionInfo <- executionInfos
+      execution <- executionInfo.execution
+      workflowExecution <- execution.workflowExecution
+      if workflowExecution.workflowExecutionUuid === id
```
mcovarr: Should check for workflow execution in status running.  Sometimes ops will set a workflow to a terminal state if they want it ignored.

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -5,14 +5,15 @@ import java.util.concurrent.{ExecutorService, Executors}
 import java.util.{Date, UUID}
 import javax.sql.rowset.serial.SerialClob
 
+import com.mysql.jdbc.exceptions.MySQLTransactionRollbackException
```
mcovarr: I thought this became a more generic exception type in one of Khalid's PRs that was merged?

engine/src/main/scala/cromwell/engine/db/slick/ExecutionInfoComponent.scala
mcovarr: oh this is probably done elsewhere isn't it?  might want to drop a comment here so nobody else gets confused.  😄

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
mcovarr: wait for the shadow world

engine/src/main/scala/cromwell/engine/backend/Backend.scala
mcovarr: Why are there defaults for these?  I think the implementation should be compelled to choose, even if it's as simple as what LocalBackend does. 

So SGE is saying no to both which causes WorkflowActor to give a somewhat vague error and makes an SGE restart impossible, restart or resume.

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
mcovarr: I'd wait for the shadows here too

mcovarr:
👍 

Yeah this is going to change a lot in the shadow world, but these changes do eliminate a lot of db stuff from the backend, integrate retries on the upserts, and test the heck out of restarts. 😄

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
geoffjentry: Agree w/ @mcovarr provided that we actually intend to go back after the structure is settled a bit more and make messages actually be messages

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
scottfrazer: I was debugging something and this might just be left-over.

scottfrazer:
@mcovarr regarding test coverage, I'm sure I can add some tests, or at least verify that all paths through that restart code are hit

engine/src/main/scala/cromwell/engine/backend/Backend.scala
scottfrazer: I'm not married to the way I implemented it, I ended up trying both ways and had an easier time working with this approach.  The way I thought about it is that you *can* return true from both of these.  However, the engine will always prefer to resume over restart.  Again, I'll change it if up really want.

The alternatives, as I see them:

```scala
def partitionRunningJobs(runningJobs: Set[ExecutionInfosByExecution]): (Set[Execution], Set[Execution])
```
or
```scala
def partitionRunningJobs(runningJobs: Map[Execution, Set[ExecutionInfo]]): (Set[Execution], Set[Execution])
```
or 
```scala
def partitionRunningJobs(runningJobs: Map[Execution, Map[String, Option[String]]): (Set[Execution], Set[Execution])
```
or
```scala
case class RestartableOrResumableExecutions(restartable: Set[Execution], resumable: Set[Execution])
def partitionRunningJobs(runningJobs: Set[ExecutionInfosByExecution]): RestartableOrResumableExecutions
```

... or any other combination of parameter(s) and return type.  Perhaps it's just a difference in personal taste, but I prefered the other scheme on mostly just aesthetics.  I'm not opposed to changing it, especially if @Horneth prefers one of the above approaches.

Another question that I didn't have a great answer for is whether or not to assume "if it's not resumable, then it's restartable".  If we take that assumption, then the whole function signature can be simplified.  So I just went with the more flexible model that you laid out in the ticket.

But given that we went with the backend deciding if something is 'restartable' or 'resumable', there was an issue of what to do with executions that fell into *neither* category of restartable or resumable.  Neither scheme handles it perfectly, but with this current scheme, I just [catch it in an else statement](https://github.com/broadinstitute/cromwell/pull/704/files#diff-b4e5dbe581881528f7565792c2766390R152) and fail the job.  With something more akin to partitioning, I guess I'd just have to check if the running execution is in neither `Set[Execution]`.

scottfrazer: Err those function signatures are actually wrong because we can't pass `Execution` and `ExecutionInfo`s to the backend

scottfrazer: So maybe this?

```scala
def partitionRunningJobs(runningJobs: Map[JobKey, Map[String, Option[String]])
: (Set[JobKey], Set[JobKey])
```

cjllanwarne:
@mcovarr did you use a non-standard thumb in your comment? The pullapprove looks specifically for `: +1:`

mcovarr:
Github has gotten cute with autocompleting emoji and unfortunately this seems to break pullapprove.  :frowning:

mcovarr:
:+1: take that pullapprove

engine/src/main/scala/cromwell/engine/db/slick/ExecutionInfoComponent.scala
scottfrazer: I'll check to make sure this is working as you described and I'll add a comment and/or refactor it a bit

scottfrazer: Yeah seems the check for workflow status happens [all the way up in the WorkflowManagerActor](https://github.com/broadinstitute/cromwell/blob/cd4f74dd0c9fee9c0b91ecb9786082d8a7741833/engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala#L406).

engine/src/main/scala/cromwell/engine/backend/Backend.scala
Horneth: The new interface only has one method for this feature no ? (called "resume") - I feel like this distinction will be lost anyway when we switch to PBE :/

scottfrazer: How are we probing the backends to ask if an execution is resumable in PBE?

scottfrazer:
Seems that it's not (easily) possible to correctly test restarts the way the code is structured now.  

I need to be able to pass in a modified `LocalBackend` which overrides `isResumable`, and `isRestartable`, and `resume`.  The problem is that `WorkflowDescriptor` has a reference to the `Backend` but also the `MaterializedWorkflowDescriptorActor` reaches into the global map of backends too... I had a very hard time constructing a custom `LocalBackend` that could be used for a workflow submission.

So I ended up adding a happy-path test but I'll wait until the shadow world emerges before adding other tests I guess...

engine/src/main/scala/cromwell/engine/backend/Backend.scala
Horneth: I don't think so, I think we just send it to the backend and it figures out by itself if it can/can't resume and act appropriately.

engine/src/main/scala/cromwell/engine/db/slick/ExecutionInfoComponent.scala
```diff
@@ -45,6 +47,18 @@ trait ExecutionInfoComponent {
...
 
+  val runningExecutionsAndExecutionInfosByWorkflowId = Compiled(
+    // Used by restart workflow code to get all executions that were
+    // in-flight when the engine is restarted.
+    (id: Rep[String]) => for {
+      executionInfo <- executionInfos
+      execution <- executionInfo.execution
+      workflowExecution <- execution.workflowExecution
+      if workflowExecution.workflowExecutionUuid === id
+      if execution.status === ExecutionStatus.Running.toString
```
mcovarr: what about `Starting`?

scottfrazer: I will add that

Horneth:
👍

PR #706, opened 2016-04-14
Have upstream test clean up after itself.

mcovarr:
May fix weird test failures.

I began getting test failures on my BackendConfiguration branch after I rebased on develop unrelated to anything I had changed.  It looks like the workflow from the KV store test is visible to the test that runs after it, which led to more success messages than the assertions expected.  Making this a separate PR in case the problem is waiting to bite anyone else.

### DIFF
--- a/engine/src/test/scala/cromwell/services/KeyValueServiceActorSpec.scala
+++ b/engine/src/test/scala/cromwell/services/KeyValueServiceActorSpec.scala
@@ -2,7 +2,7 @@ package cromwell.services
 
 import akka.pattern.ask
 import com.typesafe.config.ConfigFactory
-import cromwell.engine.WorkflowSourceFiles
+import cromwell.engine.{WorkflowSucceeded, WorkflowSourceFiles}
 import cromwell.services.KeyValueServiceActor._
 
 class KeyValueServiceActorSpec extends CromwellServicesSpec {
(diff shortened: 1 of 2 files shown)


mcovarr:
Also @geoffjentry

scottfrazer:
woops... I think this also needs to happen in `KeyValueServiceActorSpec`... I can handle that if you want

mcovarr:
@scottfrazer all yours  😄

scottfrazer:
Okay I'll take over this branch

geoffjentry:
:+1:

engine/src/test/scala/cromwell/services/KeyValueServiceActorSpec.scala
```diff
@@ -43,6 +43,7 @@ class KeyValueServiceActorSpec extends CromwellServicesSpec {
       _ = put0 shouldEqual KvPutSuccess(put)
       get0 <- ask(kvActor, get).mapTo[KvResponse]
       _ = get0 shouldEqual KvPair(ScopedKey(callKey, "k"), Option("v"))
+      _ <- dataAccess.updateWorkflowState(descriptor.id, WorkflowSucceeded)
```
cjllanwarne: I thought we got a new dataAccess per test!?

cjllanwarne:
:+1: from me

PR #707, opened 2016-04-14
Processing messages in the Shadow Hierarchy. Closes #692. Closes #648.

### LINKED ISSUE #648: Parse WDL Source, get Workflow Descriptor
First part of the validation, verify that the submission is a proper workflow.

### LINKED ISSUE #692: Update WorkflowManagerActor to *only* support the ShadowWorkflowActor. Erase olde stylee WorkflowActor

cjllanwarne:
To set expectations, this PR sets up shadow mode to actually process messages through the new set of lifecycle actors. It can be seen that submissions return an ID and make it through the `MaterializeWorkflowDescriptor` phase. However not the `Initialize` phase. This is all done in a shadow hierarchy which has not affected the normal operation yet...

We still need:
* The part which assigns backends to calls (currently blocked out to happen in `ShadowMaterializeWorkflowDescriptorActor`)
* The part which generates actors from backend name strings (blocked out at the bottom of both the `InitializeWorkflowActor` and `FinalizeWorkflowActor`)
* The entire `WorkflowExecutionActor`

If you try to run anything in shadow mode, it'll fail in the `InitializeWorkflow` phase with an exception saying that no backend assignments were found.

### DIFF
--- a/backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
+++ b/backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
@@ -30,7 +30,7 @@ object BackendWorkflowInitializationActor {
   */
 trait BackendWorkflowInitializationActor extends BackendLifecycleActor with ActorLogging {
 
-  def receive: Receive = LoggingReceive {
+  final def receive: Receive = LoggingReceive {
     case Initialize    => performActionThenRespond(initSequence, onFailure = InitializationFailed)
     case AbortWorkflow => performActionThenRespond(abortInitialization, onFailure = BackendWorkflowAbortFailedResponse)
   }
--- a/engine/src/main/scala/cromwell/engine/backend/dummy/DummyBackendJobExecutionActor.scala
+++ b/engine/src/main/scala/cromwell/engine/backend/dummy/DummyBackendJobExecutionActor.scala
@@ -1,0 +1,46 @@
+package cromwell.engine.backend.dummy
+
+import cromwell.backend.BackendJobExecutionActor.{BackendJobExecutionSucceededResponse, BackendJobExecutionResponse}
+import cromwell.backend.BackendLifecycleActor.{BackendJobExecutionAbortSucceededResponse, JobAbortResponse}
+import cromwell.backend._
+import cromwell.core.CallOutput
+import wdl4s.types._
+import wdl4s.values._
+import wdl4s.{TaskOutput, Call}
+
+import scala.concurrent.Future
+
+case class DummyBackendJobExecutionActor(workflowDescriptor: BackendWorkflowDescriptor, configurationDescriptor: BackendConfigurationDescriptor, calls: Seq[Call]) extends BackendJobExecutionActor {
+  /**
+    * Execute a new job.
+    */
+  override def execute(jobDescriptor: BackendJobDescriptor): Future[BackendJobExecutionResponse] = {
+    val outputs = (jobDescriptor.call.task.outputs map taskOutputToJobOutput _).toMap
+    Future.successful(BackendJobExecutionSucceededResponse(jobDescriptor.key, outputs))
+  }
+
+  private def taskOutputToJobOutput(taskOutput: TaskOutput) =
+    taskOutput.name -> CallOutput(sampleValue(taskOutput.wdlType), None)
+
+  private def sampleValue(wdlType: WdlType): WdlValue = wdlType match {
+    case WdlIntegerType => WdlInteger(3)
+    case WdlFloatType => WdlFloat(55.55)
+    case WdlStringType => WdlString("The rain in Spain falls mainly in the plain")
+    case WdlBooleanType => WdlBoolean(true)
+    case WdlFileType => WdlFile("/root/of/all/evil")
+    case WdlArrayType(memberType) => WdlArray(WdlArrayType(memberType), List(sampleValue(memberType)))
+    case WdlObjectType => WdlObject(Map("a" -> WdlString("1"), "b" -> WdlString("2")))
+    case WdlMapType(keyType, valueType) => WdlMap(WdlMapType(keyType, valueType), Map(sampleValue(keyType) -> sampleValue(valueType)))
+  }
+
+  /**
+    * Restart or resume a previously-started job.
+    */
+  override def recover(jobDescriptor: BackendJobDescriptor) = execute(jobDescriptor)
+
+  /**
+    * Abort a running job.
+    */
+  override def abortJob(jobKey: BackendJobDescriptorKey): Future[JobAbortResponse] = Future.successful(BackendJobExecutionAbortSucceededResponse(jobKey))
+
+}
--- a/engine/src/main/scala/cromwell/engine/package.scala
+++ b/engine/src/main/scala/cromwell/engine/package.scala
@@ -1,6 +1,7 @@
 package cromwell
 
 import akka.actor.ActorSystem
+import cromwell.backend.BackendWorkflowDescriptor
 import cromwell.core.{CallOutput, WorkflowId}
 import cromwell.engine.ExecutionStatus._
 import cromwell.engine.backend.WorkflowDescriptor
@@ -98,5 +99,13 @@ package object engine {
     } andThen {
       case _ => actorSystem.stop(materializeWorkflowDescriptorActor)
     }
+  }
+
+  final case class EngineWorkflowDescriptor(backendDescriptor: BackendWorkflowDescriptor,
+                                            declarations: WorkflowCoercedInputs,
+                                            backendAssignments: Map[Call, String],
+                                            failureMode: WorkflowFailureMode) {
+    def id = backendDescriptor.id
+    def namespace = backendDescriptor.workflowNamespace
   }
 }
--- a/engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
@@ -2,17 +2,17 @@ package cromwell.engine.workflow
 
 import akka.actor.SupervisorStrategy.Escalate
 import akka.actor._
+import com.typesafe.config.Config
 import cromwell.core.WorkflowId
-import cromwell.engine.WorkflowSourceFiles
-import cromwell.engine.backend.WorkflowDescriptor
-import cromwell.engine.workflow.lifecycle.WorkflowExecutionActor.{WorkflowExecutionFailedResponse, RestartExecutingWorkflowCommand, StartExecutingWorkflowCommand}
+import cromwell.engine.{EngineWorkflowDescriptor, WorkflowSourceFiles}
+
+import cromwell.engine.workflow.lifecycle.ShadowMaterializeWorkflowDescriptorActor.{ShadowMaterializeWorkflowDescriptorFailureResponse, ShadowMaterializeWorkflowDescriptorSuccessResponse, ShadowMaterializeWorkflowDescriptorCommand}
+import cromwell.engine.workflow.lifecycle.WorkflowExecutionActor.{WorkflowExecutionSucceededResponse, WorkflowExecutionFailedResponse, RestartExecutingWorkflowCommand, StartExecutingWorkflowCommand}
 import cromwell.engine.workflow.lifecycle.WorkflowFinalizationActor.{WorkflowFinalizationSucceededResponse, WorkflowFinalizationFailedResponse, StartEngineFinalizationCommand}
 import cromwell.engine.workflow.lifecycle.WorkflowInitializationActor.{WorkflowInitializationFailedResponse, WorkflowInitializationSucceededResponse, StartInitializationCommand}
-import cromwell.engine.workflow.lifecycle.{WorkflowFinalizationActor, WorkflowExecutionActor, WorkflowInitializationActor}
-import cromwell.engine.workflow.MaterializeWorkflowDescriptorActor.{MaterializeWorkflowDescriptorFailure, MaterializeWorkflowDescriptorSuccess}
+import cromwell.engine.workflow.lifecycle.{ShadowMaterializeWorkflowDescriptorActor, WorkflowFinalizationActor, WorkflowExecutionActor, WorkflowInitializationActor}
 import cromwell.engine.workflow.ShadowWorkflowActor._
 import cromwell.engine.workflow.WorkflowManagerActor.WorkflowOutputs
-import wdl4s.Call
 
 object ShadowWorkflowActor {
 
@@ -27,8 +27,8 @@ object ShadowWorkflowActor {
     * Responses from the ShadowWorkflowActor
     */
   sealed trait ShadowWorkflowActorResponse
-  case class ShadowWorkflowSucceededResponse(outputs: WorkflowOutputs) extends ShadowWorkflowActorResponse
-  case class ShadowWorkflowFailedResponse(reasons: Seq[Throwable]) extends Exception with ShadowWorkflowActorResponse
+  case class ShadowWorkflowSucceededResponse(workflowId: WorkflowId, outputs: WorkflowOutputs) extends ShadowWorkflowActorResponse
+  case class ShadowWorkflowFailedResponse(workflowId: WorkflowId, inState: ShadowWorkflowActorState, reasons: Seq[Throwable]) extends Exception with ShadowWorkflowActorResponse
 
   /**
     * States for the ShadowWorkflowActor FSM
@@ -76,7 +76,9 @@ object ShadowWorkflowActor {
     */
   case object WorkflowFailedState extends ShadowWorkflowActorTerminalState
 
-  case class ShadowWorkflowActorData(backendAssignments: Map[Call, String])
+  sealed trait ShadowWorkflowActorData
+  case object EmptyShadowWorkflowActorData extends ShadowWorkflowActorData
+  case class ShadowWorkflowActorDataWithDescriptor(workflowDescriptor: EngineWorkflowDescriptor) extends ShadowWorkflowActorData
 
   /**
     * Mode in which the workflow should be started:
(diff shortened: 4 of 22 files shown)


cjllanwarne:
@gauravs90 and @mcovarr please!

engine/src/main/scala/cromwell/engine/backend/dummy/DummyBackendJobExecutionActor.scala
```diff
@@ -0,0 +1,46 @@
...
+import cromwell.backend.BackendLifecycleActor.{BackendJobExecutionAbortSucceededResponse, JobAbortResponse}
+import cromwell.backend._
+import cromwell.core.CallOutput
+import wdl4s.types._
+import wdl4s.values._
+import wdl4s.{TaskOutput, Call}
+
+import scala.concurrent.Future
+
+case class DummyBackendJobExecutionActor(workflowDescriptor: BackendWorkflowDescriptor, configurationDescriptor: BackendConfigurationDescriptor, calls: Seq[Call]) extends BackendJobExecutionActor {
```
cjllanwarne: I didn't actually get as far as making one of these in this PR :'(

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
```diff
@@ -76,7 +77,7 @@ object ShadowWorkflowActor {
     */
   case object WorkflowFailedState extends ShadowWorkflowActorTerminalState
 
-  case class ShadowWorkflowActorData(backendAssignments: Map[Call, String])
+  case class ShadowWorkflowActorData(workflowDescriptor: Option[EngineWorkflowDescriptor])
```
geoffjentry: Could we have a case object for the empty case and the class for the descriptor-ful case?

geoffjentry: I wondered about that. :star: for at least thinking of me the first time :) If you could give it another whirl just to be sure that'd be fantastic.

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowManagerActor.scala
```diff
@@ -0,0 +1,255 @@
...
+  sealed trait ShadowWorkflowManagerState
+  case object Running extends ShadowWorkflowManagerState
+  case object Aborting extends ShadowWorkflowManagerState
+  case object Done extends ShadowWorkflowManagerState
+
+  /**
+    * Data
+    */
+  case class ShadowWorkflowManagerData(workflows: Map[WorkflowId, ActorRef]) {
+    def idFromActor(actor: ActorRef): Option[WorkflowId] =
```
geoffjentry: Is this just a find?

cjllanwarne: It's a `find` followed by a `map` to remove the unwanted key from the result.

geoffjentry: ah good point

gauravs90:
Hey @cjllanwarne ! I do like some of the stuff in your PR, especially with that EngineWorkflowDescriptor. The old one has a lot of stuff in it which may not be needed. I was more going towards the path where we incrementally remove stuff from the original workflow descriptor, while your PR goes the other way around by adding stuff in it in stages. I think I prefer the direction that you are going.

gauravs90:
@cjllanwarne I would like to rebase my PR with your branch, and merge it into yours (stuff where I'm populating the Call -> Assignment map + some tests). Do you think that's fine?

cjllanwarne:
@gauravs90 sounds perfect. I'll leave it up to you whether you want to rebase or just re-apply the changes directly into this branch

mcovarr:
:+1: looks like a great start!  Really looking to moving out of the shadows and getting rid of unnecessary switches and crazy long class names.  😄

gauravs90:
Hey @cjllanwarne ! I have re-applied my changes to this branch..please checkout the latest commit before further making your changes. I will close my PR. Also, it'd be great if you can review it 😄

engine/src/main/scala/cromwell/engine/package.scala
```diff
@@ -99,4 +100,12 @@ package object engine {
       case _ => actorSystem.stop(materializeWorkflowDescriptorActor)
     }
   }
+
+  final case class EngineWorkflowDescriptor(// TODO: Re-add:: declarations: WorkflowCoercedInputs
```
geoffjentry: Would it make sense to add this todo as a ticket w/ details so we don't forget what it was?

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
```diff
@@ -149,19 +155,33 @@ class ShadowWorkflowActor(workflowId: WorkflowId, startMode: StartMode, wdlSourc
...
     case Event(WorkflowFinalizationFailedResponse(reasons), _) =>
-      context.parent ! ShadowWorkflowFailedResponse(reasons)
+      context.parent ! ShadowWorkflowFailedResponse(workflowId, FinalizingWorkflowState, reasons)
       goto(WorkflowFailedState)
     case Event(AbortWorkflowCommand, stateData) => ??? // TODO: Handle abort
   }
 
+
+  when(WorkflowAbortedState) {
+    case Event(_, _) => throw new Exception("No messages please!")
```
geoffjentry: What exactly happens when these exceptions are thrown?

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowManagerActor.scala
```diff
@@ -0,0 +1,255 @@
...
+  case class WorkflowIdToActorRef(workflowId: WorkflowId, workflowActor: ActorRef)
+
+  sealed trait ShadowWorkflowManagerActorMessage
+  /**
+    * Commands
+    */
+  sealed trait ShadowWorkflowManagerActorCommand extends ShadowWorkflowManagerActorMessage
+
+  case class SubmitWorkflowCommand(source: WorkflowSourceFiles) extends ShadowWorkflowManagerActorCommand
+  // TODO: This SubscribeToWorkflowCommand is a candidate for deletion (see the handler if you don't believe me!)
```
geoffjentry: Can we determine a plan for when this is clearly a go or no-go on the todo and capture that in a ticket?

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowManagerActor.scala
```diff
@@ -0,0 +1,255 @@
...
+        Await.ready(donePromise.future, Duration.Inf)
+      }
+    }
+  }
+
+  private def replyToAbort(id: WorkflowId): Unit = {
+    // Access the mutable state here synchronously, not in the callback.
+    val workflowActor = stateData.workflows.get(id)
+    val sndr = sender()
+    globalDataAccess.getWorkflowState(id) onComplete {
```
geoffjentry: Is this a preexisting onComplete or a freshie?

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowManagerActor.scala
```diff
@@ -0,0 +1,255 @@
...
+  private def restartWorkflow(restartableWorkflow: WorkflowDescriptor): WorkflowIdToActorRef = {
+    logger.info("Invoking restartableWorkflow on " + restartableWorkflow.id.shortString)
+    submitWorkflow(restartableWorkflow.sourceFiles, replyTo = None, Option(restartableWorkflow.id))
+  }
+
+  private def restartIncompleteWorkflows(): Unit = {
+    def logRestarts(restartableWorkflows: Traversable[WorkflowDescriptor]): Unit = {
+      val num = restartableWorkflows.size
+      val displayNum = if (num == 0) "no" else num.toString
+      val plural = if (num == 1) "" else "s"
```
geoffjentry: I thought we had an enhanced something or other floating around to do the pluralization somewhere. There's also something in scalaz but IIRC it is kinda hokey

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -343,43 +345,24 @@ class WorkflowManagerActor(config: Config)
...
+      case MaterializeWorkflowDescriptorSuccess(descriptor) =>
+        val wfActor = context.actorOf(WorkflowActor.props(descriptor), name = s"WorkflowActor-$workflowId")
+        wfActor ! SubscribeTransitionCallBack(self)
+        wfActor ! (if (isRestart) Restart else Start(replyTo))
+        logger.debug(s"Successfuly started ${wfActor.path} for Workflow $workflowId")
+        context.stop(materializeWorkflowDescriptorActor)
+        workflowId -> wfActor
+      case MaterializeWorkflowDescriptorFailure(error) =>
+        val messageOrBlank = Option(error.getMessage).mkString
+        logger.error(error, s"$tag: Workflow failed submission: " + messageOrBlank)
```
geoffjentry: I think this would look better if the Some case on the message had the `: ` instead of potentially having a dangling colon here

engine/src/main/scala/cromwell/engine/workflow/lifecycle/ShadowMaterializeWorkflowDescriptorActor.scala
```diff
@@ -0,0 +1,300 @@
...
+  sealed trait ShadowWorkflowDescriptorMaterializationResult extends ShadowMaterializeWorkflowDescriptorActorMessage
+  case class ShadowMaterializeWorkflowDescriptorSuccessResponse(workflowDescriptor: EngineWorkflowDescriptor) extends ShadowWorkflowDescriptorMaterializationResult
+  case class ShadowMaterializeWorkflowDescriptorFailureResponse(reason: Throwable) extends Exception with ShadowWorkflowDescriptorMaterializationResult
+
+  /*
+  States
+   */
+  sealed trait ShadowMaterializeWorkflowDescriptorActorState { def terminal = false }
+  sealed trait ShadowMaterializeWorkflowDescriptorActorTerminalState extends ShadowMaterializeWorkflowDescriptorActorState {
+    override def terminal = true
```
geoffjentry: val terminal

engine/src/main/scala/cromwell/engine/workflow/lifecycle/ShadowMaterializeWorkflowDescriptorActor.scala
```diff
@@ -0,0 +1,300 @@
...
+  /**
+    * Map a call to a backend name depending on the runtime attribute key
+    */
+  private def mapCallToBackend(call: Call): String = {
+
+    val runtimeAttributesMap = call.task.runtimeAttributes.attrs
+    val backend = runtimeAttributesMap.get(RuntimeBackendKey) match {
+      case Some(backendNameAsExp) =>
+        // Don't allow *any* lookup or functions. The runtime must be static!
+        val backendNameAsString: String = backendNameAsExp.evaluate(NoLookup, NoFunctions) match {
```
geoffjentry: pull this inner match into its own function

engine/src/main/scala/cromwell/engine/workflow/lifecycle/ShadowMaterializeWorkflowDescriptorActor.scala
```diff
@@ -0,0 +1,300 @@
...
+  }
+
+  /**
+    * Checks if a given backend name is plugged-in in to the Cromwell sub-system and returns if true.
+    * Returns the default backend if `backendName` is not found in the list
+     */
+  private def validateBackend(backendName: String): String = {
+    if (PluggedInBackends.exists(_.equalsIgnoreCase(backendName)))
+      backendName
+    else {
```
geoffjentry: don't mix the non-brace on the if with braces on the else. I prefer with braces but the scala style guide disagrees with me.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowFinalizationActor.scala
```diff
@@ -90,29 +115,46 @@ case class WorkflowFinalizationActor(workflowTag: String, backendsToFinalize: Ma
...
+    backendsToFinalize map { backend => (finalizationActor(backend, callAssignmentMap(backend)), backend) } toMap
+  }
+
+  /**
+    * Turns a mapping from Call => backendName into a mapping from backendName => List[Call]
+    */
+  private def callAssignments(backendAssignments: Map[Call, String]): Map[String, Seq[Call]] = {
+    // Group the assignments by backendName:
+    val asTuples = backendAssignments.toSeq
+    val groupedByBackendName: Map[String, Seq[(Call, String)]] = asTuples groupBy { case (call, backendName) => backendName }
```
geoffjentry: do you need the `asTuples` here? W/o it you'd end up wiht a Map[String, Map[Call, String]] I believe and it shouldn't make a big difference below

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowInitializationActor.scala
```diff
@@ -89,30 +119,51 @@ case class WorkflowInitializationActor(workflowTag: String, backendsToInitialize
...
+    backendsToInitialize
+      .map { backend => (initializationActor(backend, callAssignmentMap(backend)), backend) } // Create the initializationActors
+      .collect { case (Some(actorRef), backend) => actorRef -> backend } // Only track the backends which need initialization
+      .toMap
+  }
+
+  /**
+    * Turns a mapping from Call => backendName into a mapping from backendName => List[Call]
+    */
+  private def callAssignments(backendAssignments: Map[Call, String]): Map[String, Seq[Call]] = {
```
geoffjentry: Is this the same as the code I commended on earlier? If so can it be extracted out?

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
```diff
@@ -42,6 +42,7 @@ trait CromwellApiService extends HttpService with PerRequestCreator {
 
   import CromwellApiServiceActor._
 
+  def shadowMode: Boolean
   val workflowManager: ActorRef
```
geoffjentry: make this a def while you're here

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowManagerActor.scala
cjllanwarne: I believe this is a remnant from the pre-shadow days. 

OOI, is there a good sell-it-to-the-masses rationale for "never use onComplete"?

geoffjentry: One of the key abstractions of the actor model is that inside the actor only one thing is being processed at once. When you have futures (including callbacks) kicking off inside of an actor that no longer is the case, you're back to needing to reason about things like shared mutable state and other concurrency issues - there are better schemes for that than actors.

It's not the end of the world and sometimes it is unavoidable (e.g. if you look at things like pipeTo, they're also using callbacks albeit in a more controlled fashion) but the fewer places it is used the fewer places one needs to reason about that class of concurrency issue.

An example like this one would appear to be in the 'no big deal' category but the problem is that as we've seen it's easy for stuff to slowly creep in

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowManagerActor.scala
cjllanwarne: I'll just remove it. If it's needed, we can re-add with a working implementation

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowManagerActor.scala
cjllanwarne: I'm sold. I'll see if this can be reorganised around an anonymous actor... Or something

cjllanwarne: Actually, this is al based around non-shadow things. I'm going to remove for now. And we can re-add it properly during the `abort()` tickets

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowManagerActor.scala
geoffjentry: I wouldn't worry too much about it, it's just one if those patterns I'd like people to think carefully about prior to adding to the mayhem. Otoh if you see an easy way to avoid it go for it

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowManagerActor.scala
cjllanwarne: Nothing obvious on my CMD-SHFT-F. Since this is as-was in WorkflowManagerActor, at least it's not a regression. No action in this PR.

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
cjllanwarne: This is legacy in that it only affects the non-shadow world. And it's not a change, except in removing it from a `{...}` block.

Unless you feel strongly otherwise, I intend to not action in this PR.

cjllanwarne:
@gauravs90 I'm happy to handle the tidy-up of this PR. Could you though review it (including my integration of your changes) and thumbs-up if you're happy? Cheers!

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowFinalizationActor.scala
cjllanwarne: I'd be very interested in a tidier version here. It's like this because there's no groupBy on Map (since Map doesn't extend Seq).
... @mcovarr is usually very good at suggesting tidier versions of my messy code...?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowInitializationActor.scala
cjllanwarne: Probably. We can probably make a WorkflowLifecycleActor trait to include a lot of common code. Watch this space...

engine/src/main/scala/cromwell/engine/package.scala
cjllanwarne: Actually I can do this right now...

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowLifecycleActor.scala
```diff
@@ -0,0 +1,111 @@
...
+
+  /**
+    * Turns a mapping from Call => backendName into a mapping from backendName => List[Call]
+    */
+  protected def callAssignments(backendAssignments: Map[Call, String]): Map[String, Seq[Call]] = {
+    // Group the assignments by backendName:
+    val asTuples = backendAssignments.toSeq
+    val groupedByBackendName: Map[String, Seq[(Call, String)]] = asTuples groupBy { case (call, backendName) => backendName }
+    groupedByBackendName mapValues { _ map { case (call, backendName) => call } }
+  }
```
mcovarr: not sure this much of an improvement

```scala
  /**
    * Turns a mapping from Call => backendName into a mapping from backendName => List[Call]
    */
  protected def callAssignments(backendAssignments: Map[Call, String]): Map[String, List[Call]] = {
    backendAssignments.foldLeft(Map.empty[String, List[Call]]) {
      case (acc, (call, backendName)) =>
        acc + (backendName -> (acc.getOrElse(backendName, List.empty[Call]) :+ call))
    }
  }
```

cjllanwarne: Thanks!

backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
```diff
@@ -30,7 +30,7 @@ object BackendWorkflowInitializationActor {
   */
 trait BackendWorkflowInitializationActor extends BackendLifecycleActor with ActorLogging {
```
gauravs90: Is this name okay? Since it does more than just initialization (like abort et. al) ?

engine/src/main/scala/cromwell/engine/backend/dummy/DummyBackendJobExecutionActor.scala
gauravs90: This is just for a quick proof of concept check? Or can we move this to the tests?

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
```diff
@@ -85,62 +87,65 @@ object ShadowWorkflowActor {
...
       }
       executionActor ! commandToSend
       goto(ExecutingWorkflowState)
-    case Event(WorkflowInitializationFailedResponse(reasons), _) =>
-      // TODO: Handle workflow failure
-      ???
+    case Event(WorkflowInitializationFailedResponse(reason), _) =>
+      context.parent ! ShadowWorkflowFailedResponse(workflowId, InitializingWorkflowState, reason)
+      goto(WorkflowFailedState)
     case Event(AbortWorkflowCommand, stateData) => ??? // TODO: Handle abort
```
gauravs90: I think all these `AbortWorkflowCommand` can move to the `whenUnhandled` state? Unless it is expected that each state handles it differently in which case these are just fine as it is.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/ShadowMaterializeWorkflowDescriptorActor.scala
```diff
@@ -0,0 +1,282 @@
...
+    * Throws an exception if `backendName` is not found in the list
+    */
+  private def validateBackend(backendName: String): String = {
+    if (PluggedInBackends.exists(_.equalsIgnoreCase(backendName))) backendName
+    else throw new Exception(s"$backendName not found in the list of plugged in backends!")
+
+  }
+}
+
+class ShadowMaterializeWorkflowDescriptorActor() extends LoggingFSM[ShadowMaterializeWorkflowDescriptorActorState, ShadowMaterializeWorkflowDescriptorActorData] with LazyLogging {
```
gauravs90: Is  this needed to be an FSM? It only has one actionable state and I cannot think of any other states that can be here 😕

engine/src/main/scala/cromwell/engine/workflow/lifecycle/ShadowMaterializeWorkflowDescriptorActor.scala
```diff
@@ -0,0 +1,282 @@
...
+                                      failureMode: WorkflowFailureMode): ErrorOr[EngineWorkflowDescriptor] = {
+    for {
+      coercedInputs <- validateCoercedInputs(rawInputs, namespace)
+      declarations <- validateDeclarations(namespace, workflowOptions, coercedInputs, NoWorkflowEngineFunctions)
+      declarationsAndInputs = declarations ++ coercedInputs
+      backendDescriptor = BackendWorkflowDescriptor(id, namespace, declarationsAndInputs, workflowOptions)
+    } yield EngineWorkflowDescriptor(backendDescriptor, declarations, backendAssignments, failureMode)
+  }
+
+  private def validateBackendAssignments(calls: Seq[Call], workflowOptions: WorkflowOptions): ErrorOr[Map[Call, String]] = {
```
gauravs90: Woah! This turned into a Scalaz thingy 😁

geoffjentry: good point, sadly it's not a cats thingy but this is good enough ;)

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowLifecycleActor.scala
```diff
@@ -0,0 +1,111 @@
...
+    case fromState -> toState =>
+      // Only log this at debug - these states are never seen or used outside of the CallActor itself.
+      log.debug(s"State is transitioning from $fromState to $toState.")
+  }
+
+  /**
+    * Turns a mapping from Call => backendName into a mapping from backendName => List[Call]
+    */
+  protected def callAssignments(backendAssignments: Map[Call, String]): Map[String, List[Call]] = {
+    backendAssignments.foldLeft(Map.empty[String, List[Call]]) {
```
gauravs90: I may be wrong on how the result of this method should look like but this little one liner should do the same too:
```scala
protected def callAssignments(backendAssignments: Map[Call, String]): Map[String, List[Call]] =  backendAssignments.groupBy(_._2).mapValues(_.map(_._1))
```

gauravs90:
I like it overall:  👍

gauravs90:
@cjllanwarne Let me know if you want me to clean up any of the review comments received, either from me or Jeff or anything else. I'll be happy to do it :-)

gauravs90:
How does pullapprove work?!

👍

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowLifecycleActor.scala
gauravs90: Ohh..never mind. I see this got evolved from something like this itself. Ignore it then.

gauravs90:
👍

backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
cjllanwarne: Hmm, I think it's OK - it *can* be aborted but its *purpose* is to initialise the backend for the workflow

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
cjllanwarne: I assumed that they would be different depending on the stage. TBH I wouldn't spend time worrying about these aborts since it's all going to have to change regardless

engine/src/main/scala/cromwell/engine/workflow/lifecycle/ShadowMaterializeWorkflowDescriptorActor.scala
cjllanwarne: Haha, i do enjoy a good FSM... Right now it's just a single state and you're probably right. But,
- I don't think it *hurts* it to be a FSM? (is there a hidden cost?)
- I can predict a future in which backend assignment is not so synchronous as it is now, in which case this actor might be calling out to another (Analysis?) actor and awaiting a response. In that case an FSM will be better IMO.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowLifecycleActor.scala
cjllanwarne: Did you make that work? I got problems because backendAssignments wasn't a Seq and so didn't have `groupBy` - but if you got it to work I think that way's more intuitive

cjllanwarne:
PullApprove has stopped noticing the 👍 s recently. I think we'll need a new PullApprove incantation...

engine/src/main/scala/cromwell/engine/workflow/lifecycle/ShadowMaterializeWorkflowDescriptorActor.scala
cjllanwarne: happy to do either, but IMO if we want to go cats we should make that choice soon

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowLifecycleActor.scala
gauravs90: Just typed this into a scala interpreter and it gives me out:

```scala
scala> val dumbMap = Map(1 -> "local", 3 -> "jes", 5 -> "local", 7 -> "ccc", 9 -> "jes", 11 -> "local")
dumbMap: scala.collection.immutable.Map[Int,String] = Map(5 -> local, 1 -> local, 9 -> jes, 7 -> ccc, 3 -> jes, 11 -> local)

scala> val res = dumbMap.groupBy(_._2).mapValues(_.map(_._1))
res: scala.collection.immutable.Map[String,scala.collection.immutable.Iterable[Int]] = Map(ccc -> List(7), jes -> List(9, 3), local -> List(5, 1, 11))
```

This is kinda what we want right?

cjllanwarne: Huh, yeah that's right and I much prefer that way, if we can use `case` matches or something to avoid the `_`s

gauravs90: I think it's intuitive even with the _s, but that maybe because I have been looking at it for a long time.

cjllanwarne: Fair enough. I think with the comment above to explain it, it's not going to be hard to work out. So I vote we go with yours and just move on :) Commit upcoming...

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowLifecycleActor.scala
```diff
@@ -0,0 +1,107 @@
...
+    case fromState -> toState =>
+      // Only log this at debug - these states are never seen or used outside of the CallActor itself.
+      log.debug(s"State is transitioning from $fromState to $toState.")
+  }
+
+  /**
+    * Turns a mapping from Call => backendName into a mapping from backendName => List[Call]
+    */
+  protected def callAssignments(backendAssignments: Map[Call, String]): Map[String, List[Call]] =
+    backendAssignments.groupBy(_._2).mapValues(_.keys.toList)
```
cjllanwarne: @gauravs90 I'm very happy with this! Needed a `toList` because it was a `Iterable` otherwise

gauravs90: Let's merge this PR then ! 🎉

cjllanwarne: Will rebase and merge

PR #708, opened 2016-04-14
Backend side Validation. Closes #651.

### LINKED ISSUE #651: Move Backend Specific Validation into constructor
Depends upon the Stubbed out backends.

francares:
Hi All,
This PR is intended to review the implementation to perform "runtime attributes" validation in the backend side.

Please, feel free to add any comment/idea that can add value to this.

### DIFF
--- a/backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
+++ b/backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
@@ -2,10 +2,17 @@ package cromwell.backend
 
 import akka.actor.ActorLogging
 import akka.event.LoggingReceive
-import cromwell.backend.BackendWorkflowInitializationActor._
 import cromwell.backend.BackendLifecycleActor._
+import cromwell.backend.BackendWorkflowInitializationActor._
+import cromwell.backend.validation.TryUtils
+import cromwell.backend.validation.exception.ValidationAggregatedException
+import cromwell.core._
+import lenthall.exception.AggregatedException
+import wdl4s.expression.{NoFunctions, ValueEvaluator}
+import wdl4s.{Call, WdlExpression}
 
 import scala.concurrent.Future
+import scalaz.{Validation, Failure, NonEmptyList}
 
 object BackendWorkflowInitializationActor {
 
@@ -13,14 +20,18 @@ object BackendWorkflowInitializationActor {
   sealed trait BackendWorkflowInitializationActorCommand extends BackendWorkflowLifecycleActorCommand
 
   case object Initialize extends BackendWorkflowInitializationActorCommand
+
   final case class Abort(jobKey: BackendJobDescriptorKey) extends BackendWorkflowInitializationActorCommand
 
   // Responses
   sealed trait BackendWorkflowInitializationActorResponse extends BackendWorkflowLifecycleActorResponse
 
   sealed trait InitializationResponse extends BackendWorkflowInitializationActorResponse
+
   case object InitializationSuccess extends InitializationResponse
+
   case class InitializationFailed(reason: Throwable) extends Exception with InitializationResponse
+
 }
 
 /**
@@ -28,32 +39,96 @@ object BackendWorkflowInitializationActor {
   */
 trait BackendWorkflowInitializationActor extends BackendWorkflowLifecycleActor with ActorLogging {
 
-  final def receive: Receive = LoggingReceive {
-    case Initialize    => performActionThenRespond(initSequence, onFailure = InitializationFailed)
-    case AbortWorkflow => performActionThenRespond(abortInitialization, onFailure = BackendWorkflowAbortFailedResponse)
+  def receive: Receive = LoggingReceive {
+    case Initialize => performActionThenRespond(initSequence(), onFailure = InitializationFailed)
+    case AbortWorkflow => performActionThenRespond(abortInitialization(), onFailure = BackendWorkflowAbortFailedResponse)
   }
 
   /**
     * Our predefined sequence to run during preStart
     */
-  final def initSequence = for {
-    _ <- validate
-    _ <- beforeAll
+  final def initSequence() = for {
+    _ <- validate()
+    _ <- beforeAll()
   } yield InitializationSuccess
 
   /**
     * Abort all initializations.
     */
-  def abortInitialization: Future[WorkflowAbortResponse]
-
-  /**
-    * Validate that this WorkflowBackendActor can run all of the calls that it's been assigned
-    */
-  def validate: Future[Unit]
+  def abortInitialization(): Future[WorkflowAbortResponse]
 
   /**
     * A call which happens before anything else runs
     */
-  def beforeAll: Future[Unit]
+  def beforeAll(): Future[Unit]
 
+  /**
+    * Validates runtime attributes for one specific call.
+    *
+    * @param runtimeAttributes Runtime Attributes with already evaluated values.
+    * @return If all entries from runtime attributes section are valid Success otherwise
+    *         Failure with the aggregation of errors.
+    */
+  def validateRuntimeAttributes(runtimeAttributes: EvaluatedRuntimeAttributes): Future[scalaz.Validation[NonEmptyList[String], Unit]]
+
+  /**
+    * Validate that this WorkflowBackendActor can run all of the calls that it's been assigned
+    */
+  protected def validate(): Future[Unit] = {
+    def eval(call: Call) = validateRuntimeAttributes(evaluateRuntimeAttributesWdlExpressions(call))
+
+    val validationResult: Future[Seq[(Call, Validation[NonEmptyList[String], Unit])]] =
+      Future.sequence(calls map { call => eval(call) map { valRes => (call, valRes) } })
+
+    val failedValidations = validationResult.map(evaluateValidationResultPerCall)
+
+    //TODO: Add recover or not but refactor this as part of https://github.com/broadinstitute/cromwell/issues/725.
+    failedValidations map { failures =>
+      if (failures.nonEmpty) {
+        throw new AggregatedException("Runtime attribute validation failed", failures)
+      }
+    }
+  }
+
+  /**
+    * Creates WDL expression evaluator for the specific call.
+    *
+    * @param call Call which contains WDL expressions in it.
+    * @return A WDL ValueEvaluator.
+    */
+  protected def createEvaluator(call: Call): ValueEvaluator = {
+    val declarations = workflowDescriptor.workflowNamespace.workflow.declarations ++ call.task.declarations
+    val knownInputs = workflowDescriptor.inputs
+    val lookup = WdlExpression.standardLookupFunction(knownInputs, declarations, NoFunctions)
+    new ValueEvaluator(lookup, NoFunctions)
+  }
+
+  /**
+    * Evaluates WDL expressions within call runtime attributes.
+    *
+    * @param call Call which contains WDL expressions in it.
+    * @return Evaluated runtime attributes.
+    */
+  private def evaluateRuntimeAttributesWdlExpressions(call: Call): EvaluatedRuntimeAttributes = {
+    val evaluator = createEvaluator(call)
+    def evaluate(wdlExpression: WdlExpression) = evaluator.evaluate(wdlExpression.ast)
+    val evaluateAttrs = call.task.runtimeAttributes.attrs mapValues evaluate
+    TryUtils.sequenceMap(evaluateAttrs, "Runtime attributes evaluation").get
+  }
+
+  /**
+    * Checks if there was any validation failure.
+    *
+    * @param evaluation Tuple with a call and related validation result.
+    * @return A list with all validation exceptions.
+    */
+  private def evaluateValidationResultPerCall(evaluation: Seq[(Call, ErrorOr[Unit])]): Seq[ValidationAggregatedException] = {
+    evaluation collect {
+      case (call, validation: Failure[NonEmptyList[String]]) =>
+        validation match {
+          case failureWithStringErr: Failure[NonEmptyList[String]] =>
+            ValidationAggregatedException(s"Runtime attribute validation failed for task '${call.taskFqn}'.", failureWithStringErr.e.list)
+        }
+    }
+  }
 }
--- a/backend/src/main/scala/cromwell/backend/validation/ContinueOnReturnCode.scala
+++ b/backend/src/main/scala/cromwell/backend/validation/ContinueOnReturnCode.scala
@@ -1,7 +1,7 @@
-package cromwell.engine.backend.runtimeattributes
+package cromwell.backend.validation
 
 /**
-  * Decides is a call/job continues upon a specific return code.
+  * Decides if a call/job continues upon a specific return code.
   */
 sealed trait ContinueOnReturnCode {
   /**
--- a/backend/src/main/scala/cromwell/backend/validation/package.scala
+++ b/backend/src/main/scala/cromwell/backend/validation/package.scala
@@ -1,0 +1,30 @@
+package cromwell.backend
+
+import lenthall.exception.AggregatedException
+
+import scala.util.{Success, Failure, Try}
+
+package object validation {
+  object TryUtils {
+    private def sequenceIterable[T](tries: Iterable[Try[_]], unbox: () => T, prefixErrorMessage: String) = {
+      tries collect { case f: Failure[_] => f } match {
+        case failures if failures.nonEmpty => Failure(new AggregatedException(prefixErrorMessage, failures.map(_.failed.get)))
+        case _ => Success(unbox())
+      }
+    }
+
+    def sequence[T](tries: Seq[Try[T]], prefixErrorMessage: String = ""): Try[Seq[T]] = {
+      def unbox = tries map {
+        _.get
+      }
+      sequenceIterable(tries, unbox _, prefixErrorMessage)
+    }
+
+    def sequenceMap[T, U](tries: Map[T, Try[U]], prefixErrorMessage: String = ""): Try[Map[T, U]] = {
+      def unbox = tries mapValues {
+        _.get
+      }
+      sequenceIterable(tries.values, unbox _, prefixErrorMessage)
+    }
+  }
+}
--- a/build.sbt
+++ b/build.sbt
@@ -25,11 +25,12 @@ lazy val sgeBackend = (project in backendRoot / "sge")
   .settings(sgeBackendSettings:_*)
 
 lazy val jesBackend = (project in backendRoot / "jes")
-  .dependsOn(backend % "test->test;compile->compile", gcsfilesystem)
+  .dependsOn(backend % "test->test;compile->compile", gcsfilesystem % "test->test;compile->compile")
   .settings(jesBackendSettings:_*)
 
+//TODO: remove jesBackend once refactoring has finished.
 lazy val engine = (project in file("engine"))
-  .dependsOn(core % "test->test;compile->compile", gcsfilesystem % "test->test;compile->compile", backend)
+  .dependsOn(core % "test->test;compile->compile", jesBackend % "test->test;compile->compile")
   .settings(engineSettings:_*)
   .configs(AllTests).settings(inConfig(AllTests)(Defaults.testTasks): _*)
   .configs(DockerTest).settings(inConfig(DockerTest)(Defaults.testTasks): _*)
--- a/supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorInitializationActor.scala
+++ b/supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorInitializationActor.scala
@@ -1,0 +1,55 @@
+package cromwell.backend.impl.htcondor
+
+import akka.actor.Props
+import cromwell.backend.BackendLifecycleActor.WorkflowAbortResponse
+import cromwell.backend.impl.htcondor.HtCondorInitializationActor._
+import cromwell.backend.validation.ContinueOnReturnCodeSet
+import cromwell.backend.validation.RuntimeAttributesKeys._
+import cromwell.backend.validation.RuntimeAttributesValidation._
+import cromwell.backend.{BackendConfigurationDescriptor, BackendWorkflowDescriptor, BackendWorkflowInitializationActor}
+import cromwell.core._
+import wdl4s.Call
+
+import scala.concurrent.Future
+import scalaz.Scalaz._
+
+object HtCondorInitializationActor {
+  val FailOnStderrDefaultValue = true
+  val ContinueOnRcDefaultValue = 0
+
+  def props(workflowDescriptor: BackendWorkflowDescriptor, calls: Seq[Call], configurationDescriptor: BackendConfigurationDescriptor): Props =
+    Props(new HtCondorInitializationActor(workflowDescriptor, calls, configurationDescriptor))
+}
+
+class HtCondorInitializationActor(override val workflowDescriptor: BackendWorkflowDescriptor,
+                                  override val calls: Seq[Call],
+                                  override val configurationDescriptor: BackendConfigurationDescriptor) extends BackendWorkflowInitializationActor {
+  /**
+    * Abort all initializations.
+    */
+  override def abortInitialization(): Future[WorkflowAbortResponse] = ???
+
+  /**
+    * A call which happens before anything else runs
+    */
+  override def beforeAll(): Future[Unit] = Future.successful(())
+
+  /**
+    * Validates runtime attributes for one specific call.
+    *
+    * @param runtimeAttributes Runtime Attributes with already evaluated values.
+    * @return If all entries from runtime attributes section are valid Success otherwise
+    *         Failure with the aggregation of errors.
+    */
+  override def validateRuntimeAttributes(runtimeAttributes: EvaluatedRuntimeAttributes): Future[ErrorOr[Unit]] = {
+    Future {
+      val docker = validateDocker(runtimeAttributes.get(Docker), () => "Failed to get Docker mandatory key from runtime attributes".failureNel)
+      val failOnStderr = validateFailOnStderr(runtimeAttributes.get(FailOnStderr), () => FailOnStderrDefaultValue.successNel)
+      val continueOnReturnCode = validateContinueOnReturnCode(runtimeAttributes.get(ContinueOnReturnCode),
+        () => ContinueOnReturnCodeSet(Set(ContinueOnRcDefaultValue)).successNel)
+      (docker |@| failOnStderr |@| continueOnReturnCode) {
+        (_, _, _)
+      }
+    }
+  }
+}
--- a/supportedBackends/htcondor/src/test/scala/HtCondorInitializationActorSpec.scala
+++ b/supportedBackends/htcondor/src/test/scala/HtCondorInitializationActorSpec.scala
@@ -1,0 +1,82 @@
+import akka.actor.ActorSystem
+import akka.testkit.{ImplicitSender, TestDuration, TestKit}
+import com.typesafe.config.ConfigFactory
+import cromwell.backend.BackendWorkflowInitializationActor.{InitializationFailed, Initialize}
+import cromwell.backend.impl.htcondor.HtCondorInitializationActor
+import cromwell.backend.{BackendConfigurationDescriptor, BackendWorkflowDescriptor}
+import cromwell.core.{WorkflowId, WorkflowOptions}
+import lenthall.exception.AggregatedException
+import org.scalatest.{BeforeAndAfterAll, Matchers, WordSpecLike}
+import spray.json.{JsObject, JsValue}
+import wdl4s.values.WdlValue
+import wdl4s.{Call, NamespaceWithWorkflow, WdlSource}
+
+import scala.concurrent.duration._
+
+class HtCondorInitializationActorSpec extends TestKit(ActorSystem("HtCondorInitializationActorSpec"))
+  with WordSpecLike with Matchers with BeforeAndAfterAll with ImplicitSender {
+  val Timeout = 5.second.dilated
+
+  val HelloWorld =
+    """
+      |task hello {
+      |  String addressee = "you"
+      |  command {
+      |    echo "Hello ${addressee}!"
+      |  }
+      |  output {
+      |    String salutation = read_string(stdout())
+      |  }
+      |
+      |  RUNTIME
+      |}
+      |
+      |workflow hello {
+      |  call hello
+      |}
+    """.stripMargin
+
+  val defaultBackendConfig = new BackendConfigurationDescriptor("htcondor-config", ConfigFactory.load())
+
+  private def buildWorkflowDescriptor(wdl: WdlSource,
+                                      inputs: Map[String, WdlValue] = Map.empty,
+                                      options: WorkflowOptions = WorkflowOptions(JsObject(Map.empty[String, JsValue])),
+                                      runtime: String = "") = {
+    new BackendWorkflowDescriptor(
+      WorkflowId.randomId(),
+      NamespaceWithWorkflow.load(wdl.replaceAll("RUNTIME", runtime)),
+      inputs,
+      options
+    )
+  }
+
+  override def afterAll {
+    system.shutdown()
+  }
+
+  private def getHtCondorBackend(workflowDescriptor: BackendWorkflowDescriptor, calls: Seq[Call], conf: BackendConfigurationDescriptor) = {
+    system.actorOf(HtCondorInitializationActor.props(workflowDescriptor, calls, conf))
+  }
+
+  "HtCondorInitializationActor" should {
+    "return InitializationFailed when there are no runtime attributes defined." in {
+      within(Timeout) {
+        val workflowDescriptor = buildWorkflowDescriptor(HelloWorld, runtime = """runtime { }""")
+        val backend = getHtCondorBackend(workflowDescriptor, workflowDescriptor.workflowNamespace.workflow.calls, defaultBackendConfig)
+        backend ! Initialize
+        expectMsgPF() {
+          case InitializationFailed(failure) =>
+            failure match {
+              case exception: AggregatedException =>
+                if (!exception.exceptionContext.equals("Runtime attribute validation failed"))
+                  fail("Exception message does not contains 'Runtime attribute validation failed'.")
+                if (exception.throwables.size != 1)
+                  fail("Number of errors coming from AggregatedException is not equals to one.")
+                if (!exception.throwables.head.getMessage.contains("Failed to get Docker mandatory key from runtime attributes"))
+                  fail("Message from error nr 1 in validation exception does not contains 'Failed to get Docker mandatory key from runtime attributes'.")
+            }
+        }
+      }
+    }
+  }
+}
--- a/supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesInitializationActor.scala
+++ b/supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesInitializationActor.scala
@@ -1,0 +1,164 @@
+package cromwell.backend.impl.jes
+
+import akka.actor.Props
+import cromwell.backend.BackendLifecycleActor.WorkflowAbortResponse
+import cromwell.backend.impl.jes.JesInitializationActor._
+import cromwell.backend.impl.jes.io.{JesAttachedDisk, JesWorkingDisk}
+import cromwell.backend.impl.jes.model.MemorySize
+import cromwell.backend.validation.ContinueOnReturnCodeSet
+import cromwell.backend.validation.RuntimeAttributesKeys._
+import cromwell.backend.validation.RuntimeAttributesValidation._
+import cromwell.backend.{BackendConfigurationDescriptor, BackendWorkflowDescriptor, BackendWorkflowInitializationActor}
+import cromwell.core._
+import wdl4s.Call
+import wdl4s.parser.MemoryUnit
+import wdl4s.types.{WdlIntegerType, WdlStringType}
+import wdl4s.values.{WdlArray, WdlInteger, WdlString, WdlValue}
+
+import scala.concurrent.Future
+import scalaz.Scalaz._
+
+object JesInitializationActor {
+  //TODO: check if these need to be configurable.
+  val CpuKey = "cpu"
+  val CpuDefaultValue = 1
+  val ZonesKey = "zones"
+  val ZoneDefaultValue = Seq("us-central1-a")
+  val PreemptibleKey = "preemptible"
+  val PreemptibleDefaultValue = 0
+  val BootDiskSizeKey = "bootDiskSizeGb"
+  val BootDiskSizeDefaultValue = 10
+  val MemoryKey = "memory"
+  val MemoryDefaultValue = "2 GB"
+  val DisksKey = "disks"
+  val DisksDefaultValue = s"${JesWorkingDisk.Name} 10 SSD"
+  val DefaultJesWorkingDisk = JesAttachedDisk.parse(DisksDefaultValue).get
+  val FailOnStderrDefaultValue = false
+  val ContinueOnRcDefaultValue = 0
+
+  def props(workflowDescriptor: BackendWorkflowDescriptor, calls: Seq[Call], configurationDescriptor: BackendConfigurationDescriptor): Props =
+    Props(new JesInitializationActor(workflowDescriptor, calls, configurationDescriptor))
+}
+
+class JesInitializationActor(override val workflowDescriptor: BackendWorkflowDescriptor,
+                             override val calls: Seq[Call],
+                             override val configurationDescriptor: BackendConfigurationDescriptor) extends BackendWorkflowInitializationActor {
+  /**
+    * Abort all initializations.
+    */
+  override def abortInitialization(): Future[WorkflowAbortResponse] = ???
+
+  //TODO: Workflow options may need to be validated for JES.
+
+  /**
+    * A call which happens before anything else runs
+    */
+  override def beforeAll(): Future[Unit] = Future.successful(())
+
+  /**
+    * Validates runtime attributes for one specific call.
+    *
+    * @param runtimeAttributes Runtime Attributes with already evaluated values.
+    * @return If all entries from runtime attributes section are valid Success otherwise
+    *         Failure with the aggregation of errors.
+    */
+  override def validateRuntimeAttributes(runtimeAttributes: EvaluatedRuntimeAttributes): Future[ErrorOr[Unit]] = {
+    Future {
+      val cpu = validateCpu(runtimeAttributes.get(CpuKey))
+      val zones = validateZone(runtimeAttributes.get(ZonesKey))
+      val preemptible = validatePreemptible(runtimeAttributes.get(PreemptibleKey))
+      val bootDiskSize = validateBootDisk(runtimeAttributes.get(BootDiskSizeKey))
+      val memory = validateMemory(runtimeAttributes.get(MemoryKey))
+      val disks = validateLocalDisks(runtimeAttributes.get(DisksKey))
+      val docker = validateDocker(runtimeAttributes.get(Docker), () => "Failed to get Docker mandatory key from runtime attributes".failureNel)
+      val failOnStderr = validateFailOnStderr(runtimeAttributes.get(FailOnStderr), () => FailOnStderrDefaultValue.successNel)
+      val continueOnReturnCode = validateContinueOnReturnCode(runtimeAttributes.get(ContinueOnReturnCode),
+        () => ContinueOnReturnCodeSet(Set(ContinueOnRcDefaultValue)).successNel)
+      (cpu |@| zones |@| preemptible |@| bootDiskSize |@| memory |@| disks |@| docker |@| failOnStderr |@| continueOnReturnCode) {
+        (_, _, _, _, _, _, _, _, _)
+      }
+    }
+  }
+
+  private def validateCpu(cpu: Option[WdlValue]): ErrorOr[Int] = {
+    val cpuValidation = cpu.map(validateInt).getOrElse(CpuDefaultValue.successNel)
+    if (cpuValidation.isFailure) {
+      s"Expecting $CpuKey runtime attribute to be an Integer".failureNel
+    }
+    else {
+      cpuValidation
+    }
+  }
+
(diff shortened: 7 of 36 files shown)


supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorInitializationActor.scala
```diff
@@ -0,0 +1,65 @@
...
+  /**
+    * Abort all initializations.
+    */
+  override def abortInitialization: Future[WorkflowAbortResponse] = ???
+
+  /**
+    * Validate that this WorkflowBackendActor can run all of the calls that it's been assigned
+    */
+  override def validate: Future[Unit] = {
+    //TODO: Expressions need to be resolved. EvalFunctions is needed. TBD in the PR.
```
francares: I need some opinions on how to resolve the evaluation of run-time attributes expressions.

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorInitializationActor.scala
```diff
@@ -0,0 +1,65 @@
...
+import cromwell.backend.validation.Validation._
+import cromwell.backend.{BackendConfigurationDescriptor, BackendWorkflowDescriptor, BackendWorkflowInitializationActor}
+import wdl4s.util.TryUtil
+import wdl4s.{Call, ThrowableWithErrors, WdlExpression}
+
+import scala.concurrent.Future
+import scala.util.{Failure, Success}
+import scalaz.Scalaz._
+
+class HtCondorInitializationActor(override val workflowDescriptor: BackendWorkflowDescriptor,
```
francares: This is just an example... once I have a validated model, I will continue with other backends.

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesInitializationActor.scala
```diff
@@ -0,0 +1,87 @@
...
+  override def beforeAll: Future[Unit] = ???
+
+  /**
+    * Validates runtime attributes for each related call.
+    *
+    * @param taskAttributes Runtime attributes.
+    * @return If all entries from runtime requirements section are valid Success otherwise
+    *         Failure with the aggregation of errors.
+    */
+  private def validateRuntimeAttributes(taskAttributes: Map[String, WdlExpression]) = {
```
francares: Work in progress here with other validations.

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorInitializationActor.scala
```diff
@@ -0,0 +1,74 @@
...
+  /**
+    * Validate that this WorkflowBackendActor can run all of the calls that it's been assigned
+    */
+  override def validate: Future[Unit] = {
+    //TODO: Expressions need to be resolved. EvalFunctions is needed. TBD in the PR.
+    //TODO: refactor old code to use validation functions from Validation object in Backend.
+    //TODO: Workflow options may need to be validated for JES.
+    //TODO: Move specifics from Jes to JesBackendInitializationActor.
+    Future {
+      TryUtil.sequence(calls map {
```
cjllanwarne: Unfortunately, I believe this TryUtil will only give you the error in the first call which fails RuntimeAttributeValidation.

In a scheme where we're already using scalaz, is there some way to fold the calls' vallidations together? I'm going to have to summon @geoffjentry or @mcovarr to give a real answer here...

cjllanwarne:
I like the scheme of providing a standard functions library to do some common validations for a user. 

I'm a little wary of listing backend-specific keys in that library though. In particular I don't want to give any backend developers the impression that either they (a) cannot use their own custom keys or (b) must use every "standard" key.

But mostly, this looks good!

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorInitializationActor.scala
francares: I can do it but that means to return ErrorOr[same type] in all validation functions.
If we follow the idea of just returning success and failure and not the resolved value it may work since I does not need to return the value.

build.sbt
```diff
@@ -28,8 +28,9 @@ lazy val jesBackend = (project in backendRoot / "jes")
   .dependsOn(backend % "test->test;compile->compile", gcsfilesystem)
   .settings(jesBackendSettings:_*)
 
+//TODO: remove jesBackend once refactoring has finished.
 lazy val engine = (project in file("engine"))
-  .dependsOn(core % "test->test;compile->compile", gcsfilesystem % "test->test;compile->compile", backend)
+  .dependsOn(core % "test->test;compile->compile", jesBackend % "test->test;compile->compile", backend % "test->test;compile->compile")
```
francares: I can remove backend and keep jesBackend only.

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorInitializationActor.scala
```diff
@@ -0,0 +1,74 @@
+package cromwell.backend.impl.htcondor
```
francares: Can we remove impl package name?

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorInitializationActor.scala
mcovarr: The current structure also doesn't say which call failed validation.  I think the following addresses these problems, though I'm sure it can be improved:

```scala
  override def validate: Future[Unit] = {
    case class ErrorMessageAggregatedException(override val exceptionContext: String,
                                               override val errorMessages: Traversable[String]) extends MessageAggregation
    Future {
      val callsWithValidations = calls map { call => call -> validateRuntimeAttributes(call) }
      val failedValidations = callsWithValidations collect { case (call, validation) if validation.isFailure =>
        ErrorMessageAggregatedException(s"Runtime attribute validation failed for task '${call.taskFqn}': ", validation.left.toList) }
      
      if (failedValidations.nonEmpty) {
        throw new exception.AggregatedException("Runtime attribute validation failed", failedValidations)
      }
    }
  }

  private def validateRuntimeAttributes(call: Call): ErrorOr[(Option[String], Boolean)] = {
    val taskAttributes: Map[String, WdlExpression] = call.task.runtimeAttributes.attrs
    val docker = validateDocker(taskAttributes.get(Docker), () => "Failed to get Docker mandatory key from runtime attributes".failureNel)
    val failOnStderr = validateFailOnStderr(taskAttributes.get(FailOnStderr), () => FailOnStderrDefaultValue.successNel)
    (docker |@| failOnStderr) { (_, _) }
  }
```

francares: Thanks, I will use your recommendation.

backend/src/main/scala/cromwell/EngineFunctions.scala
```diff
@@ -0,0 +1,24 @@
+package cromwell
```
francares: Reusing EngineFunctions impl from branch: tj_PluggableLocal.

geoffjentry: Now that we have the subprojects should we be better about full package names (yuck) - e.g. `org.broadinstitute.blah.blah.backend.cromwell`? Specifically my concern is how all of this works within the fat jar

backend/src/main/scala/cromwell/backend/WdlStandardLibraryImpl.scala
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  override def write_lines(params: Seq[Try[WdlValue]]): Try[WdlFile] = writeToTsv(params, classOf[WdlArray])
+  override def write_map(params: Seq[Try[WdlValue]]): Try[WdlFile] = writeToTsv(params, classOf[WdlMap])
+  override def write_object(params: Seq[Try[WdlValue]]): Try[WdlFile] = writeToTsv(params, classOf[WdlObject])
+  override def write_objects(params: Seq[Try[WdlValue]]): Try[WdlFile] = writeToTsv(params, classOf[WdlArray])
+  override def write_tsv(params: Seq[Try[WdlValue]]): Try[WdlFile] = writeToTsv(params, classOf[WdlArray])
+  override def writeTempFile(path: String, prefix: String, suffix: String, content: String): String = {
+    // This may be called multiple times with the same inputs.  Calling this twice with the same
+    // parameters should yield the same return value.
```
geoffjentry: should or will?

backend/src/main/scala/cromwell/backend/validation/ContinueOnReturnCode.scala
```diff
@@ -0,0 +1,31 @@
+package cromwell.backend.validation
+
+/**
+  * Decides is a call/job continues upon a specific return code.
```
geoffjentry: Decides if?

core/src/main/scala/cromwell/core/PathFactory.scala
```diff
@@ -0,0 +1,110 @@
+package cromwell.core
+
+import java.io.Writer
+import java.nio.file.{FileSystem, Path}
+
+import scala.collection.immutable.Queue
+import scala.util.{Success, Try}
+
+object PathFactory {
+  private def findFileSystem(rawString: String, fss: List[FileSystem], mapping: PartialFunction[FileSystem, Try[Path]]) = {
```
geoffjentry: If you're going to take in a `List` and immediately transform it to a `Stream`, why take in a `List` in the first place?

core/src/main/scala/cromwell/core/PathFactory.scala
```diff
@@ -0,0 +1,110 @@
...
+  def swapExt(filePath: String, oldExt: String, newExt: String): String = {
+    filePath.stripSuffix(oldExt) + newExt
+  }
+
+  implicit class EnhancedPath(val path: Path) extends AnyVal {
+    def swapExt(oldExt: String, newExt: String): Path = {
+      path.getFileSystem.getPath(s"${path.toString.stripSuffix(oldExt)}$newExt")
+    }
+
+    def untailed = new UntailedWriter(path)
```
geoffjentry: I'd rather see these as `untailedWriter` and `tailedWriter`.

core/src/main/scala/cromwell/core/PathFactory.scala
```diff
@@ -0,0 +1,110 @@
...
+
+    def tailed(tailedSize: Int) = TailedWriter(path, tailedSize)
+  }
+
+  implicit class FlushingAndClosingWriter(writer: Writer) {
```
geoffjentry: extends `AnyVal`

core/src/main/scala/cromwell/core/PathFactory.scala
```diff
@@ -0,0 +1,110 @@
...
+case class UntailedWriter(path: Path) extends PathWriter
+
+/**
+  * Used with a `ProcessLogger`, queues up the `tailedSize` number of lines.
+  *
+  * @param path Path to the log file.
+  * @param tailedSize Maximum number of lines to save in the internal FIFO queue.
+  */
+case class TailedWriter(path: Path, tailedSize: Int) extends PathWriter {
+  var isTailed = false
```
geoffjentry: 😦

geoffjentry: Let's find a way to do these without having public mutable state (ideally without mutable state at all)

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesInitializationActor.scala
```diff
@@ -0,0 +1,157 @@
...
+import wdl4s.parser.MemoryUnit
+import wdl4s.types.{WdlIntegerType, WdlStringType}
+import wdl4s.values.{WdlArray, WdlInteger, WdlString, WdlValue}
+import wdl4s.{Call, RuntimeAttributes, WdlExpression}
+
+import scala.concurrent.Future
+import scalaz.Scalaz._
+
+object JesInitializationActor {
+  //TODO: check if these need to be configurable.
```
geoffjentry: Not necessarily right here but prior to us wrapping up it'd be nice if we had a cleaner pattern to handle this than spamming a bunch of junk at the top

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesInitializationActor.scala
```diff
@@ -0,0 +1,157 @@
...
+
+  private def validatePreemptible(preemptible: Option[WdlValue]): ErrorOr[Int] = {
+    preemptible.map(validateInt).getOrElse(PreemptibleDefaultValue.successNel)
+  }
+
+  private def validateBootDisk(diskSize: Option[WdlValue]): ErrorOr[Int] = diskSize match {
+    case Some(x) if WdlIntegerType.isCoerceableFrom(x.wdlType) =>
+      WdlIntegerType.coerceRawValue(x) match {
+        case scala.util.Success(x: WdlInteger) => x.value.intValue.successNel
+        case scala.util.Success(coercionhasgoneawry) => s"Coercion was expected to create an Integer but instead got $coercionhasgoneawry".failureNel
```
geoffjentry: not sure if coercionhasgoneawry was yours or already existed but it made me 😺

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesInitializationActor.scala
```diff
@@ -0,0 +1,157 @@
+package cromwell.backend.impl.jes
```
geoffjentry: If you do want to remove the impl remember to do it here too

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorInitializationActor.scala
geoffjentry: This was discussed on the call the other day, right? If not I don't understand this :)

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalInitializationActor.scala
```diff
@@ -0,0 +1,73 @@
...
+    * Abort all initializations.
+    */
+  override def abortInitialization: Future[WorkflowAbortResponse] = ???
+
+  /**
+    * Validate that this WorkflowBackendActor can run all of the calls that it's been assigned
+    */
+  override def validate: Future[Unit] = {
+    //TODO: Expressions need to be resolved. EvalFunctions is needed. TBD in the PR.
+    Future {
```
geoffjentry: I'm pretty sure this code block exists in a few places - extract it out and put it somewhere general

geoffjentry:
👍 overall

geoffjentry:
Stupid PullApprove - :+1:

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorInitializationActor.scala
francares: Yes, this comment was before of the meeting :).

backend/src/main/scala/cromwell/backend/WdlStandardLibraryImpl.scala
francares: This code comes from Thib branch. Should I modify this here?

core/src/main/scala/cromwell/core/PathFactory.scala
francares: This code comes from Thib branch. Should I modify this here?

core/src/main/scala/cromwell/core/PathFactory.scala
francares: This code comes from Thib branch. Should I modify this here?

core/src/main/scala/cromwell/core/PathFactory.scala
francares: Same as above.

core/src/main/scala/cromwell/core/PathFactory.scala
francares: Same as above.

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesInitializationActor.scala
francares: I didn't change anything in validation code. The code was already there.

backend/src/main/scala/cromwell/EngineFunctions.scala
francares: Added to Waffle => https://github.com/broadinstitute/cromwell/issues/714.

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalInitializationActor.scala
francares: Yes, that is one of the topics I would like to discuss since this code may need to be executed in JobExecutionActor.
I'll wait until I have a clear idea on how to share objects or dependencies within InitializationActors and ExecutionActors.

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesInitializationActor.scala
francares: I will rename it.

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesInitializationActor.scala
francares: OK, from your comment I'm deducing you are OK with the change. So I will make the change.

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesInitializationActor.scala
geoffjentry: Please don't j thought it was cool :)

backend/src/main/scala/cromwell/backend/WdlStandardLibraryImpl.scala
francares: It would a good idea to wait for @Horneth PR merge and after that I will rebase and merge my PR (when we solve the issue of sharing data and code duplication.

core/src/main/scala/cromwell/core/PathFactory.scala
Horneth: The reasoning was that It's not necessarily intuitive to a caller of this method why they should provide a `Stream` of FileSystems, it seemed more an implementation detail of the method itself. But I agree it seems better to pass in a `Stream` if that's what we need so I don't really mind.

geoffjentry:
@cjllanwarne I'm assigning you as it looks like you might have been an official reviewer at some point (you were assigned), so it might be awaiting some thumbage

cjllanwarne:
@geoffjentry @francares is this a PR for the idea (:+1: from me) or is this now a PR for the code itself to merge into develop?

backend/src/main/scala/cromwell/backend/WdlStandardLibraryImpl.scala
geoffjentry: @francares I think your rebase statement is the right call

Horneth:
@francares here is the No-op implementation of Engine functinos in wdl4s https://github.com/broadinstitute/wdl4s/blob/develop/src/main/scala/wdl4s/expression/WdlStandardLibraryFunctions.scala#L129

francares:
@cjllanwarne @Horneth I did the changes in order to evaluate expressions based on @Horneth recommendation (Could you review it?). I also created a couple of tests cases for LocalBackend so you can see the whole idea behind it.
I will continue adding all needed tests for all supported backends and once they are in place I will let you know to review the whole PR so then I can merge it.

backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
```diff
@@ -31,31 +37,86 @@ object BackendWorkflowInitializationActor {
...
+    *
+    * @param call Call which contains WDL expressions in it.
+    * @return Evaluated runtime attributes.
     */
-  def beforeAll: Future[Unit]
+  private def evaluateRuntimeAttributesWdlExpressions(call: Call): EvaluatedRuntimeAttributes= {
+    val evaluator = createEvaluator(call)
+    def evaluate(wdlExpression: WdlExpression) = evaluator.evaluate(wdlExpression.ast)
+    val evaluateAttrs = call.task.runtimeAttributes.attrs mapValues evaluate
+    TryUtils.sequenceMap(evaluateAttrs, "Runtime attributes evaluation").get
```
francares: @cjllanwarne @Horneth @geoffjentry Should I catch exception and log or fail here?
If the case is to catch the exception and log, should this return 'InitializationSuccess' after that? or just fail and return a InitializationFailed?

Use case: task1 output is the input for task2 and that value is used within RuntimeAttributes.

francares:
@geoffjentry I'm working on JES unit test cases now... which are the cases validation should fail if entry is not present?
Let's say Docker entry is not provided in RuntimeAttributes, should validation fails in this case for JES?

francares:
Formal reviewers: @cjllanwarne @Horneth 

Thanks!

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesInitializationActor.scala
francares: Created https://github.com/broadinstitute/cromwell/issues/722 to perform this change in another PR due to several changed/new files in this PR.

backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
```diff
@@ -31,31 +40,86 @@ object BackendWorkflowInitializationActor {
...
+    * @return If all entries from runtime attributes section are valid Success otherwise
+    *         Failure with the aggregation of errors.
+    */
+  def validateRuntimeAttributes(runtimeAttributes: EvaluatedRuntimeAttributes): ErrorOr[Product]
 
   /**
     * Validate that this WorkflowBackendActor can run all of the calls that it's been assigned
     */
-  def validate: Future[Unit]
+  private def validate(): Future[Unit] = {
```
cjllanwarne: We should allow backend implementors to do whatever they like in `validate()` and just provide `validateRuntimeAttributes` as a utility method in case they want to use it.

backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
```diff
@@ -31,31 +40,86 @@ object BackendWorkflowInitializationActor {
...
+  def beforeAll(): Future[Unit]
+
+  /**
+    * Validates runtime attributes for one specific call.
+    *
+    * @param runtimeAttributes Runtime Attributes with already evaluated values.
+    * @return If all entries from runtime attributes section are valid Success otherwise
+    *         Failure with the aggregation of errors.
+    */
+  def validateRuntimeAttributes(runtimeAttributes: EvaluatedRuntimeAttributes): ErrorOr[Product]
```
cjllanwarne: What should the validated `Product` be?

geoffjentry:
@francares I believe the only attribute strictly required is docker as all of the other ones should have defaults set by the google genomics pipeline API (e.g. disk defaults to 500GB). Anything provided should be valid however.

backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
```diff
@@ -31,31 +40,86 @@ object BackendWorkflowInitializationActor {
...
+    *
+    * @param call Call which contains WDL expressions in it.
+    * @return Evaluated runtime attributes.
     */
-  def beforeAll: Future[Unit]
+  private def evaluateRuntimeAttributesWdlExpressions(call: Call): EvaluatedRuntimeAttributes = {
+    val evaluator = createEvaluator(call)
+    def evaluate(wdlExpression: WdlExpression) = evaluator.evaluate(wdlExpression.ast)
+    val evaluateAttrs = call.task.runtimeAttributes.attrs mapValues evaluate
+    TryUtils.sequenceMap(evaluateAttrs, "Runtime attributes evaluation").get
```
Horneth: This is going to throw if one of the runtime attributes can't be evaluated. So the first call for which evaluation fails will fail the future in the `validate` method.
Maybe that's fine for now, but it could fail slower by not calling `get` now but rather returning the `Try` and aggregate all the potential failures in the validate method.

backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
Horneth: the `Product` looks strange, if we don't care about the result it should be a `Unit`

backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
cjllanwarne: After looking at your `LocalInitializationActor`, I've reconsidered. I like this structure a lot but think we should have a some sort of `customValidation` in the `initSequence` just in case someone want some totally orthogonal custom validation of their own devising.

backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
```diff
@@ -31,31 +40,86 @@ object BackendWorkflowInitializationActor {
...
-  final def initSequence = for {
-    _ <- validate
-    _ <- beforeAll
-  } yield InitializationSuccess
+  final def initSequence() = {
+    for {
+      _ <- validate()
+      _ <- beforeAll()
+    } yield InitializationSuccess
+  }
```
cjllanwarne: Bracketing overload! :)

backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
```diff
@@ -31,31 +40,86 @@ object BackendWorkflowInitializationActor {
...
+    TryUtils.sequenceMap(evaluateAttrs, "Runtime attributes evaluation").get
+  }
 
+  /**
+    * Creates WDL expression evaluator for the specific call.
+    *
+    * @param call Call which contains WDL expressions in it.
+    * @return A WDL ValueEvaluator.
+    */
+  private def createEvaluator(call: Call): ValueEvaluator = {
```
cjllanwarne: We should allow an implementor override this

backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
```diff
@@ -31,31 +40,86 @@ object BackendWorkflowInitializationActor {
...
   /**
     * Validate that this WorkflowBackendActor can run all of the calls that it's been assigned
     */
-  def validate: Future[Unit]
+  private def validate(): Future[Unit] = {
+    Future {
+      val callsWithValidations = calls map { call =>
+        call -> validateRuntimeAttributes(evaluateRuntimeAttributesWdlExpressions(call))
+      }
+      val failedValidations = callsWithValidations collect {
```
Horneth: this could probably be simplified to something like 

```
val failedValidations = callsWithValidations collect {
        case (call, v: Failure[NonEmptyList[String]]) => 
          ValidationAggregatedException(s"Runtime attribute validation failed for task '${call.taskFqn}'.", v.e.list)
       }
```

backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
cjllanwarne: What happens if this `.get` throws an exception?

It might be that we just couldn't evaluate something yet because we haven't started running the workflow. That doesn't mean validation should fail.

backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
Horneth: Why not just make this protected and let them override it if they want

backend/src/main/scala/cromwell/backend/validation/package.scala
```diff
@@ -0,0 +1,30 @@
+package cromwell.backend
+
+import lenthall.exception.AggregatedException
+
+import scala.util.{Success, Failure, Try}
+
+package object validation {
+  object TryUtils {
```
cjllanwarne: Is this copy/pasted from somewhere? Ideally we'd use the original.

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorInitializationActor.scala
```diff
@@ -0,0 +1,57 @@
...
+                                  override val configurationDescriptor: BackendConfigurationDescriptor) extends BackendWorkflowInitializationActor {
+  /**
+    * Abort all initializations.
+    */
+  override def abortInitialization(): Future[WorkflowAbortResponse] = ???
+
+  /**
+    * A call which happens before anything else runs
+    */
+  override def beforeAll(): Future[Unit] = {
```
cjllanwarne: There's a lot of unneeded brackets here. Just `override def beforeAll(): Future[Unit] = Future.Successful(())`

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorInitializationActor.scala
```diff
@@ -0,0 +1,57 @@
...
+    Future.successful {
+      //Add missing code here.
```
cjllanwarne: Not necessarily. There might not be a beforeAll action.

backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
Horneth: But then if we don't throw when this fails, what's the point of doing it ?

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorInitializationActor.scala
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  /**
+    * Validates runtime attributes for one specific call.
+    *
+    * @param runtimeAttributes Runtime Attributes with already evaluated values.
+    * @return If all entries from runtime attributes section are valid Success otherwise
+    *         Failure with the aggregation of errors.
+    */
+  override def validateRuntimeAttributes(runtimeAttributes: EvaluatedRuntimeAttributes): ErrorOr[(Option[String], Boolean, ContinueOnReturnCode)] = {
```
cjllanwarne: This is a really non-obvious return type from the overridden `def` which just wants a `Product`

supportedBackends/htcondor/src/test/scala/HtCondorInitializationActorSpec.scala
```diff
@@ -0,0 +1,78 @@
...
+      options
+    )
+  }
+
+  private def getHtCondorBackend(workflowDescriptor: BackendWorkflowDescriptor, calls: Seq[Call], conf: BackendConfigurationDescriptor) = {
+    system.actorOf(HtCondorInitializationActor.props(workflowDescriptor, calls, conf))
+  }
+
+  "HtCondorInitializationActor" should {
+    "return InitializationSuccess when there are no runtime attributes defined." in {
```
cjllanwarne: This doesn't seem to match the rest of what the test actually does?

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesInitializationActor.scala
```diff
@@ -0,0 +1,141 @@
...
+      case scala.util.Success(x) => x.to(MemoryUnit.GB).amount.successNel
+      case scala.util.Failure(x) => x.getMessage.failureNel
+    }
+    val disks = validateLocalDisks(runtimeAttributes.get(DisksKey))
+    val docker = validateDocker(runtimeAttributes.get(Docker), () => None.successNel)
+    val failOnStderr = validateFailOnStderr(runtimeAttributes.get(FailOnStderr), () => FailOnStderrDefaultValue.successNel)
+    val continueOnReturnCode = validateContinueOnReturnCode(runtimeAttributes.get(ContinueOnReturnCode),
+      () => ContinueOnReturnCodeSet(Set(ContinueOnRCDefaultValue)).successNel)
+    (cpu |@| zones |@| preemptible |@| bootDiskSize |@| memory |@| disks |@| docker |@| failOnStderr |@| continueOnReturnCode) {
+      (_, _, _, _, _, _, _, _, _)
```
cjllanwarne: It'd be nicer for backend implementors if they just return a List[ErrorOr[Unit]] and this ugly scalaz `fold` is done in the `BackendInitializationActor` trait

supportedBackends/jes/src/test/scala/cromwell/backend/impl/jes/JesInitializationActorSpec.scala
```diff
@@ -0,0 +1,70 @@
...
+      options
+    )
+  }
+
+  private def getJesBackend(workflowDescriptor: BackendWorkflowDescriptor, calls: Seq[Call], conf: BackendConfigurationDescriptor) = {
+    system.actorOf(JesInitializationActor.props(workflowDescriptor, calls, conf))
+  }
+
+  "JesInitializationActor" should {
+    "return InitializationSuccess when there are no runtime attributes defined." in {
```
cjllanwarne: Not in this PR perhaps, but could you add a ticket for adding significantly more tests here! :)

backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
cjllanwarne: Haha, I agree with Thibault!

backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
francares: I can recall I had a discussion about this with @geoffjentry and he point me to => http://docs.scala-lang.org/style/method-invocation.html
Validate has side efects so should carry parenthesis if I'm not wrong.

gauravs90: I think @cjllanwarne meant about the extra `{` before the actual for comprehension.

backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
francares: Product is a tuple and the first element of the tuple is used to create an aggregated exception in case there is an exception.

backend/src/main/scala/cromwell/backend/validation/package.scala
geoffjentry: Also let's not jam stuff like this into package objects

backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
francares: I did same question but since I moved this code from other file it was hidden: 
"@cjllanwarne @Horneth @geoffjentry Should I catch exception and log or fail here?
If the case is to catch the exception and log, should this return 'InitializationSuccess' after that? or just fail and return a InitializationFailed?

Use case: task1 output is the input for task2 and that value is used within RuntimeAttributes."

backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
francares: are you saying that because you see a need that the functions can be modified (such as backend specific functions)?

backend/src/main/scala/cromwell/backend/validation/package.scala
francares: Yes, from engine. We should put this in Core or Lenthal. I think @Horneth is moving this in his PR.

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorInitializationActor.scala
francares: I forgot to change that.

supportedBackends/jes/src/test/scala/cromwell/backend/impl/jes/JesInitializationActorSpec.scala
francares: Today I will add missing ones based on the Jes validation functions.

backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
francares: ohh I see. :)

francares:
@geoffjentry @scottfrazer It seems the parser is validating memory runtime attribute entry when it tries to load namespaces.
I think it should be removed to follow the current idea.
This is the link to the code that is causing related test to fail => https://github.com/broadinstitute/wdl4s/blob/d7e19c9f4dfbc5ad912cf641af9c640eb8a9a9c7/src/main/scala/wdl4s/RuntimeAttributes.scala
Let me know how to proceed with this...

geoffjentry:
@francares At one point it was the only runtime attribute WDL was opinionated about: https://github.com/broadinstitute/wdl/blob/develop/SPEC.md#runtime-section although it looks like docker is in that class now too. That's why it's there. Whether or not that's correct or if the WDL spec should be more/less opinionated is a separate matter.

scottfrazer:
Honestly I'd really like it if that validation were removed from wdl4s.  Even though it "makes sense" for it to be there because "the spec says memory should be formatted like this so we should validate it down at the wdl4s level".  I still would rather change the spec to be a suggestion rather than something that's mandatory so we don't have this one outlier which actually makes the code a lot harder to write.

If we do change that now, I'll shed a tear because I have to rebase those changes on my super long lived branch that we were going to get to once PBE was finished.

geoffjentry:
@francares Let's remove that check. I just checked with @scottfrazer and he was up for ripping the bandaid off now on the tear shedding front.

francares:
Ok, just for clarification I will do the following:
1. Remove validateRuntimeValue from wdl4s and create a PR in that repo.
2. Add validateMemoryValue to JesInitializationActor.
3. Refactor validateMemoryValue to return similar message than the other ones.

geoffjentry:
@francares Sounds good to me

backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
francares: Changed.

backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
francares: I will make it protected.

backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
francares: It would be better to work on this in another PR. I think this PR has increased it size a lot.

francares:
I realized I need to move all ContinueOnReturnCode tests and do the refactor in CromwellRuntimeRequirements. I will add that tomorrow but that should be the last thing to do.

backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
cjllanwarne: Could you add a ticket to waffle for that?

backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
```diff
@@ -2,58 +2,116 @@ package cromwell.backend
...
+    *         Failure with the aggregation of errors.
+    */
+  def validateRuntimeAttributes(runtimeAttributes: EvaluatedRuntimeAttributes): ErrorOr[Unit]
 
   /**
     * Validate that this WorkflowBackendActor can run all of the calls that it's been assigned
     */
-  def validate: Future[Unit]
+  protected def validate(): Future[Unit] = {
+    Future {
```
cjllanwarne: We should let the `validateRuntimeAttributes()` method return a Future and compose the response from there instead of wrapping everything in this method into a block Future.

cjllanwarne:
After that method is changed to compose Futures, 👍

Horneth:
👍  with a ticket to not forget about slow failing validation

backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
francares: Done: https://github.com/broadinstitute/cromwell/issues/725.

backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
```diff
@@ -2,58 +2,127 @@ package cromwell.backend
...
 
   /**
     * Validate that this WorkflowBackendActor can run all of the calls that it's been assigned
     */
-  def validate: Future[Unit]
+  protected def validate(): Future[Unit] = {
+    def eval(call: Call) = validateRuntimeAttributes(evaluateRuntimeAttributesWdlExpressions(call))
+
+    val validationResult: Future[Seq[(Call, Validation[NonEmptyList[String], Unit])]] =
+      Future.sequence(calls map { call => (call, eval(call)) } map { entry => entry._2.map(valRes => (entry._1, valRes)) })
```
cjllanwarne: The middle map seems superfluous:
Would just `Future.sequence(calls map { call => eval(call).map(valRes => (call, valRes)) })` work?

cjllanwarne:
Looks good to me now.

backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
```diff
@@ -2,58 +2,127 @@ package cromwell.backend
...
+  /**
+    * Evaluates WDL expressions within call runtime attributes.
+    *
+    * @param call Call which contains WDL expressions in it.
+    * @return Evaluated runtime attributes.
     */
-  def beforeAll: Future[Unit]
+  private def evaluateRuntimeAttributesWdlExpressions(call: Call): EvaluatedRuntimeAttributes = {
+    val evaluator = createEvaluator(call)
+    def evaluate(wdlExpression: WdlExpression) = evaluator.evaluate(wdlExpression.ast)
```
scottfrazer: I personally think we should prefer to not use `ValueEvaluator`s directly because it exposes ASTs unnecessarily.  Instead maybe `createEvaluator(call: Call): ValueEvaluator` could become `lookupFunction(call: Call)(variable: String): WdlValue`

Then this line could become `wdlExpression.evaluate(lookupFunction(call), NoFunctions)`

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/model/MemorySize.scala
```diff
@@ -0,0 +1,46 @@
...
+        (amount |@| unit) { (a, u) => new MemorySize(a, u) } match {
+          case scalaz.Success(memorySize) => Success(memorySize)
+          case scalaz.Failure(nel) => Failure(new UnsupportedOperationException(nel.list.mkString("\n")))
+        }
+      case _ => Failure(new UnsupportedOperationException(s"$unparsed should be of the form 'X Unit' where X is a number, e.g. 8 GB"))
+    }
+  }
+}
+
+case class MemorySize(amount: Double, unit: MemoryUnit) {
```
scottfrazer: I'm very confused about why this is in `cromwell.backend.impl.jes.model`  Seems to me like we could (and probably should) support memory sizes for local jobs too!

backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
```diff
@@ -2,58 +2,127 @@ package cromwell.backend
...
-    * A call which happens before anything else runs
+    * Creates WDL expression evaluator for the specific call.
+    *
+    * @param call Call which contains WDL expressions in it.
+    * @return A WDL ValueEvaluator.
+    */
+  protected def createEvaluator(call: Call): ValueEvaluator = {
+    val declarations = workflowDescriptor.workflowNamespace.workflow.declarations ++ call.task.declarations
+    val knownInputs = workflowDescriptor.inputs
+    val lookup = WdlExpression.standardLookupFunction(knownInputs, declarations, NoFunctions)
```
scottfrazer: I think `knownInputs` is not quite right... `standardLookupFunction` is looking for non-fully-qualified names but you're passing it fully-qualified names

Horneth:
@francares Are you going to merge this soon ? I need to rebase on top of it but I'd rather do it directly from develop if you're going to merge it now.

francares:
@Horneth Yes, I will do it within one hour.

backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
francares: @Horneth will work on this after I merge this PR since this function will be re-used by different classes.

backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
francares: @Horneth will work on this after I merge this PR since this function will be re-used by different classes.

Horneth:
Awesome thanks

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/model/MemorySize.scala
francares: I created a ticket to work on this => https://github.com/broadinstitute/cromwell/issues/729.

francares:
@Horneth Waiting for builds to pass then I'll merge.

PR #710, opened 2016-04-15
Allow listing more than 1000 files in GCS.  Closes #699.

### LINKED ISSUE #699: wdl glob / Array[File] only seems to grab 1000 elements
I tried running a these two glob output commands
`    Array[File] gvcf_list = glob("split_gvcfs/*.gz")`
`    Array[File] gvcf_index_list = glob("split_gvcfs/*.tbi")`
Which should have had 3457 elements in each array.  When I go to the output directory bucket of the above task, I see each glob directory does indeed have 3457 elements but when I go to the symbols table in the Cromwell db the arrays only have 1000 elements.  This causes tasks further down in the workflow to fail because I am expecting an array of 3457 elements but I am only getting 1000.

mcovarr:
@jsotobroad the default limit on results for the GCS list API is 1000.  I made this configurable and bumped up the default to 50K.  That number is basically pulled out of thin air, if you have a more informed suggestion of a good value to use that would be great.  😄 

I still need to figure out the least awkward place to shoehorn the documentation for this value in the README.md.

### DIFF
--- a/filesystems/gcs/src/main/scala/cromwell/filesystems/gcs/GcsFileSystemProvider.scala
+++ b/filesystems/gcs/src/main/scala/cromwell/filesystems/gcs/GcsFileSystemProvider.scala
@@ -16,7 +16,10 @@ import com.google.api.services.storage.Storage
 import com.google.api.services.storage.model.StorageObject
 import com.google.cloud.hadoop.gcsio.{GoogleCloudStorageReadChannel, GoogleCloudStorageWriteChannel, ObjectWriteConditions}
 import com.google.cloud.hadoop.util.{ApiErrorExtractor, AsyncWriteChannelOptions, ClientRequestHelper}
+import com.typesafe.config.ConfigFactory
+import lenthall.config.ScalaConfig.EnhancedScalaConfig
 
+import scala.annotation.tailrec
 import scala.collection.JavaConverters._
 import scala.concurrent.duration._
 import scala.concurrent.{ExecutionContext, ExecutionContextExecutorService}
@@ -201,15 +204,36 @@ class GcsFileSystemProvider private (storageClient: Try[Storage], executionConte
   override def isHidden(path: Path): Boolean = throw new NotImplementedError()
 
   private def list(gcsDir: NioGcsPath) = {
-    val listRequest = client.objects().list(gcsDir.bucket)
+    val maxResults = ConfigFactory.load().getIntOr("google.list-max-results", 1000).toLong
+    val listRequest = client.objects().list(gcsDir.bucket).setMaxResults(maxResults)
     listRequest.setPrefix(gcsDir.objectName)
 
-    val paths = for {
-      listedFile <- withRetry(listRequest.execute()).getItems.asScala
-    } yield NioGcsPath(s"$getScheme${listedFile.getBucket}${GcsFileSystem.Separator}${listedFile.getName}")(gcsDir.getFileSystem.asInstanceOf[GcsFileSystem])
+    def objectToPath(storageObject: StorageObject): Path = {
+      NioGcsPath(s"$getScheme${storageObject.getBucket}${GcsFileSystem.Separator}${storageObject.getName}")(gcsDir.getFileSystem.asInstanceOf[GcsFileSystem])
+    }
+
+    // Contains a Seq corresponding to the current page of objects, plus a token for the next page of objects, if any.
+    case class ListPageResult(objects: Seq[StorageObject], nextPageToken: Option[String])
+
+    def requestListPage(pageToken: Option[String] = None): ListPageResult = {
+      val objects = withRetry(listRequest.setPageToken(pageToken.orNull).execute())
+      ListPageResult(objects.getItems.asScala, Option(objects.getNextPageToken))
+    }
+
+    @tailrec
+    def remainingObjects(pageToken: Option[String], acc: Seq[StorageObject]): Seq[StorageObject] = {
+      if (pageToken.isEmpty) acc
+      else {
+        val page = requestListPage(pageToken)
+        remainingObjects(page.nextPageToken, acc ++ page.objects)
+      }
+    }
+
+    val firstPage = requestListPage(pageToken = None)
+    val allObjects = remainingObjects(firstPage.nextPageToken, firstPage.objects)
 
     new DirectoryStream[Path] {
-      override def iterator(): util.Iterator[Path] = paths.toIterator.asJava
+      override def iterator(): util.Iterator[Path] = (allObjects map objectToPath).toIterator.asJava
       override def close(): Unit = {}
     }
   }


mcovarr:
@scottfrazer and @kshakir

scottfrazer:
Okay how in the world did you figure that out? :+1:

filesystems/gcs/src/main/scala/cromwell/filesystems/gcs/GcsFileSystemProvider.scala
```diff
@@ -201,7 +203,8 @@ class GcsFileSystemProvider private (storageClient: Try[Storage], executionConte
   override def isHidden(path: Path): Boolean = throw new NotImplementedError()
 
   private def list(gcsDir: NioGcsPath) = {
-    val listRequest = client.objects().list(gcsDir.bucket)
+    val maxResults = ConfigFactory.load().getIntOr("google.list-max-results", 50000).toLong
+    val listRequest = client.objects().list(gcsDir.bucket).setMaxResults(maxResults)
```
cjllanwarne: Have you verified that google will actually honour such a high number? 

Do you need to do an equivalent fix on `glob()`?

cjllanwarne:
Assuming that you've tested it, my previous comment is probably irrelevant. So :+1:

filesystems/gcs/src/main/scala/cromwell/filesystems/gcs/GcsFileSystemProvider.scala
mcovarr: glob() is implemented by calling list(), so this where it ends up

mcovarr:
Upon further reflection I think I might want to work on this a bit more before merging.  This addresses the problems of the list limit being too low and not configurable, but it doesn't yet address the problem that when the list limit is hit the code silently truncates.  I'm not sure if GCS provides an API to say how many object have a given prefix, but I'd like to see if something like that is available.  If the number of objects that would be returned by list() exceeds our limit, I think the code should fail noisily and not silently.

cjllanwarne:
Or page the results back somehow? But then glob which relies on it will have to change too :(

kcibul:
As an example of the paging... could you just accumulate them and then pass them back that way?
https://cloud.google.com/storage/docs/json_api/v1/objects/list

mcovarr:
@kcibul yup, that's what I'm working on now.  😄

mcovarr:
This has been re-written for paging so it'll need re-review.  Sorry... 😢

filesystems/gcs/src/main/scala/cromwell/filesystems/gcs/GcsFileSystemProvider.scala
```diff
@@ -201,15 +204,37 @@ class GcsFileSystemProvider private (storageClient: Try[Storage], executionConte
   override def isHidden(path: Path): Boolean = throw new NotImplementedError()
 
   private def list(gcsDir: NioGcsPath) = {
-    val listRequest = client.objects().list(gcsDir.bucket)
+    val maxResults = ConfigFactory.load().getIntOr("google.list-max-results", 1000).toLong
```
mcovarr: Not sure if we still want to expose this configuration option.

kcibul: meh -- might as well ;)

geoffjentry: If it is paging anyways why not just hardcode it to the max size?

cjllanwarne: I agree with "meh, might as well", as a safety valve in case we suddenly need to change it in production... but I wouldn't bother documenting it.

filesystems/gcs/src/main/scala/cromwell/filesystems/gcs/GcsFileSystemProvider.scala
```diff
@@ -201,15 +204,37 @@ class GcsFileSystemProvider private (storageClient: Try[Storage], executionConte
...
     listRequest.setPrefix(gcsDir.objectName)
 
-    val paths = for {
-      listedFile <- withRetry(listRequest.execute()).getItems.asScala
-    } yield NioGcsPath(s"$getScheme${listedFile.getBucket}${GcsFileSystem.Separator}${listedFile.getName}")(gcsDir.getFileSystem.asInstanceOf[GcsFileSystem])
+    def objectToPath(storageObject: StorageObject): Path = {
+      NioGcsPath(s"$getScheme${storageObject.getBucket}${GcsFileSystem.Separator}${storageObject.getName}")(gcsDir.getFileSystem.asInstanceOf[GcsFileSystem])
+    }
+
+    type PageToken = String
```
cjllanwarne: Does this type really add anything?

cjllanwarne:
I reiterate my :+1:

filesystems/gcs/src/main/scala/cromwell/filesystems/gcs/GcsFileSystemProvider.scala
geoffjentry: And if it does, perhaps a more robust type-ification than a type alias

filesystems/gcs/src/main/scala/cromwell/filesystems/gcs/GcsFileSystemProvider.scala
mcovarr: oh I like this answer.  😄

filesystems/gcs/src/main/scala/cromwell/filesystems/gcs/GcsFileSystemProvider.scala
mcovarr: meh, as they say.  I generally stay away from type aliases but this seemed like a reasonable application.  I can get rid of it if it feels unwarranted.

geoffjentry: I've really soured on type aliases except for a few key categories. If you want the name is prefer the full belt and suspenders treatment, otherwise just use the string

filesystems/gcs/src/main/scala/cromwell/filesystems/gcs/GcsFileSystemProvider.scala
scottfrazer: Should this new option be in our default application.conf?

scottfrazer: Not sure if that's what you meant by "document it"

scottfrazer:
:+1:

PR #712, opened 2016-04-15
Pluggable Local Backend

Horneth:
First implementation of a pluggable LocalBackend. This is more a light basic implementation and a starting point to iterate over.
What is implemented : 
- Support for non-docker jobs
- Support for docker jobs
- Support for "ContinueOnReturnCode" "FailOnStderr" and "docker" runtime attributes
- Engine functions
- Abort

Things to think about:
- How to share code between backends ?  runtime attributes validation, engine functions, shared filesystem code.. 
- Testing

Note: some code is duplicated from the engine as it's still used by the current non-PBE implementation. Eventually this will replace all local backend code in the engine.

Currently adding more tests for 

- [x] abort

~~- [ ] engine functions~~

- [x] input localization

- [x] expression evaluation

- [x] coercion 

- [x] scatter


backend/src/main/scala/cromwell/attributes/ContinueOnReturnCode.scala
```diff
@@ -1,4 +1,4 @@
-package cromwell.engine.backend.runtimeattributes
+package cromwell.attributes
```
francares: I moved this to a different package in my branch if you want to take a look...

backend/src/main/scala/cromwell/attributes/RuntimeAttributesValidation.scala
```diff
@@ -0,0 +1,53 @@
+package cromwell.attributes
+
+import cromwell.core._
+import wdl4s.types.WdlIntegerType
+import wdl4s.values._
+
+import scalaz.Scalaz._
+trait RuntimeAttributesValidation {
```
francares: Same here.

engine/src/main/scala/cromwell/engine/backend/local/LocalBackend.scala
```diff
@@ -7,6 +7,7 @@ import akka.actor.ActorSystem
 import better.files._
```
francares: I renamed this actor in my PR, could you take a look if that works for you?... I had to rename it due to the initialization actor.

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalExecutor.scala
```diff
@@ -0,0 +1,165 @@
...
+  val logger = LoggerFactory.getLogger("LocalExecutor")
+
+  val call = jobDescriptor.key.call
+  val jobPaths = new JobPaths(workflowDescriptor, config, filesystems, jobDescriptor.key)
+  val callContext = new CallContext(jobPaths.callRoot.toString,
+    jobPaths.stdout.toAbsolutePath.toString,
+    jobPaths.stderr.toAbsolutePath.toString
+  )
+  val callEngineFunction = new LocalCallEngineFunctions(filesystems, callContext)
+  val evaluator = {
```
francares: This is the issue I'm referring to on BackendInitializationActor... to create an evaluator a jobDescriptor is needed :(.

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalRuntimeAttributes.scala
```diff
@@ -0,0 +1,27 @@
...
+
+import cromwell.attributes.{ContinueOnReturnCode, RuntimeAttributesValidation}
+import wdl4s.values.WdlValue
+import scalaz._
+import Scalaz._
+import lenthall.exception.MessageAggregation
+
+object LocalRuntimeAttributes extends RuntimeAttributesValidation {
+  def apply(attrs: Map[String, WdlValue]): LocalRuntimeAttributes = {
+    val continueOnReturnCode = validateContinueOnReturnCode(attrs.get("continueOnReturnCode"))
```
francares: This will duplicates what is in InitializationActor... This is the shared data between actors I was talking about.

cjllanwarne: Rather than discuss here, I've started a doc for this - https://docs.google.com/a/broadinstitute.com/document/d/1feRDusWXQQ2pJ03sNHTNmrrnnwL3y-vtyF1fv_RdogU/edit?usp=sharing

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalExecutor.scala
Horneth: You can create more or less "powerful" evaluators. You can have an evaluator with no lookup function and no engine functions, so it will only evaluate simple expression like `Int x = 3 + 6` or `File f = "path/to/some/file"`
Depending on what information you have available (workflow level declarations, call inputs, call level declarations, ...) you can construct more or less aware lookup functions.
Same with engine functions, at workflow level you can create WorkflowLevel engine functions with a workflow context and filesystems (practically the only difference is that WorkflowEngine function can't evaluate stdout and stderr but pretty much everything else will work).

core/src/main/scala/cromwell/core/package.scala
```diff
@@ -5,9 +5,12 @@ import wdl4s.values.{SymbolHash, WdlValue}
 import scalaz._
 
 package object core {
+  class WorkflowContext(val root: String)
```
geoffjentry: I'd prefer to not add classes and such to the package object unless there's a good reason to put it in here

core/src/main/scala/cromwell/core/package.scala
```diff
@@ -5,9 +5,12 @@ import wdl4s.values.{SymbolHash, WdlValue}
...
+  class CallContext(override val root: String, val stdout: String, val stderr: String) extends WorkflowContext(root)
   type ErrorOr[+A] = ValidationNel[String, A]
 
   type LocallyQualifiedName = String
   case class CallOutput(wdlValue: WdlValue, hash: Option[SymbolHash])
   type CallOutputs = Map[LocallyQualifiedName, CallOutput]
+  type EvaluatedRuntimeAttributes = Map[String, WdlValue]
```
geoffjentry: I'd also prefer that cases where one thinks a type is warranted that a real type (i.e. case class) is used instead of a type alias

engine/src/main/scala/cromwell/engine/backend/package.scala
```diff
@@ -6,8 +6,6 @@ import scala.concurrent.{ExecutionContext, Future}
 import scala.collection.immutable.Seq
 
 package object backend {
-  class WorkflowContext(val root: String)
```
geoffjentry: eh, I see - all the stuff I commented on above was just moved from a separate package object. blech. disregard above unless you want to be amazing :)

Horneth: I don't mind putting it somewhere else, creating a file to have a one line case class seemed overkill that's why I put it there.

geoffjentry: It's more avoiding the slow creep we previously saw. Feel free to disregard entirely

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
```diff
@@ -1,34 +1,59 @@
...
     */
-  override protected def calls: Seq[Call] = ???
+  override def execute(jobDescriptor: BackendJobDescriptor): Future[BackendJobExecutionResponse] = {
+    Try(
+      new LocalExecutor(workflowDescriptor, jobDescriptor, fileSystems, config)
+    ) map { _.execute } recover {
+      case e => Future.successful(BackendJobExecutionFailedResponse(jobDescriptor.key,
+        new RuntimeException(s"Failed to initialize job ${jobDescriptor.key}", e))
+      )
+    } get
```
scottfrazer: I can't quite tell yet but it seems like callers of `execute` are not equipped to receive an exception if the `Try`.  Very possible I'm missing something though

core/src/main/scala/cromwell/core/PathFactory.scala
```diff
@@ -0,0 +1,111 @@
...
+case class UntailedWriter(path: Path) extends PathWriter
+
+/**
+  * Used with a `ProcessLogger`, queues up the `tailedSize` number of lines.
+  *
+  * @param path Path to the log file.
+  * @param tailedSize Maximum number of lines to save in the internal FIFO queue.
+  */
+case class TailedWriter(path: Path, tailedSize: Int) extends PathWriter {
+  var isTailed = false
```
geoffjentry: I accidentally blamed @francares for these in his PR :)

There shouldn't be public mutable state in here (and ideally, no mutable state at all)

Horneth: This comes from here https://github.com/broadinstitute/cromwell/blob/494aa86713ee3a19c3d944355672862efd292e3f/engine/src/main/scala/cromwell/util/FileUtil.scala#L63

geoffjentry: 🙍 

Could you throw in some privates for now?

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
Horneth: I think that the recover takes care of that, if the Try fails it's mapped to a `Future[BackendJobExecutionFailedResponse]` with the exception that crashed the try.

engine/src/main/scala/cromwell/engine/backend/local/LocalBackend.scala
Horneth: Yep no worries. I'm not worrying too much about that right now, I prefer to focus on making sure that the implementation works. I'll rebase when your PR is merged and use your version.

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
scottfrazer: Yes yes, I see now

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
```diff
@@ -1,34 +1,56 @@
...
 
-  /**
-    * Execute a new job.
-    */
-  override def execute(jobDescriptor: BackendJobDescriptor): Future[BackendJobExecutionResponse] = ???
+class LocalBackend(override val workflowDescriptor: BackendWorkflowDescriptor,
+                   override val configurationDescriptor: BackendConfigurationDescriptor,
+                   override val calls: Seq[Call]) extends BackendJobExecutionActor {
+
+  override protected implicit def ec: ExecutionContext = scala.concurrent.ExecutionContext.global
```
cjllanwarne: This feels icky. Doesn't it already acquire the actor system context by default?

scottfrazer: `BackendLifecycleActor` already has this...

```scala
protected implicit def ec: ExecutionContext = context.dispatcher
```

Horneth: Hmm yeah, I don't think the default context was here when I wrote it, I'll remove this.

Horneth:
@cjllanwarne feel free to have a first pass. There are still TODOs and things that needs to be moved around but it's getting close to the final result

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
```diff
@@ -1,51 +1,229 @@
...
-    workflowPaths.workflowRoot.createDirectories()
+    // workflowPaths.workflowRoot.createDirectories() TODO move to initialize actor
+    jobPaths.callRoot.createDirectories()
   }
-   /**
+
+  lazy val executionPromise = Promise[BackendJobExecutionResponse]()
+
+  // We don't want the job to run on the same execution context as the actor, because it would block the actor message processing abilities while the job is running
+  override implicit val ec = scala.concurrent.ExecutionContext.global
```
geoffjentry: I don't think it's worth doing at this moment but could you add a ticket to get this off of the global context?

cjllanwarne: Actually you should be able to just delete this line and it'll just work in the actor's context

Horneth: So I removed it, but abort wasn't working. Although I put the blocking part of waiting for the process exit value in a future, it was still blocking.

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
```diff
@@ -1,29 +1,227 @@
...
+          }
+        }
+
+        lazy val badReturnCodeMessage =
+          s"""Call ${call.fullyQualifiedName}, Workflow ${workflowDescriptor.id}: return code was ${returnCode.getOrElse("(none)")}
+              |
+            |Full command was: $processArgs
+              |
+            |${stderrTailed.tailString}
+          """.stripMargin
```
scottfrazer: indentation weirdness

backend/src/main/scala/cromwell/attributes/RuntimeAttributesValidation.scala
```diff
@@ -0,0 +1,53 @@
...
+
+  val defaultContinueOnReturnCode = ContinueOnReturnCodeSet(Set(0)).successNel
+  val defaultFailOnStderr = false.successNel
+
+  def validateContinueOnReturnCode(value: Option[WdlValue]): ErrorOr[ContinueOnReturnCode] = {
```
cjllanwarne: If possible, can we rebase on @francares' (#708) and remove this from this PR?

engine/src/main/scala/cromwell/engine/backend/jes/JesEngineFunctions.scala
```diff
@@ -1,20 +1,26 @@
...
+class JesWorkflowEngineFunctions(fileSystems: List[FileSystem], context: WorkflowContext) extends WorkflowEngineFunctions(context) {
+  import cromwell.core.PathFactory._
+
   override def globPath(glob: String): String = context.root.toAbsolutePath(fileSystems).resolve(JesBackend.globDirectory(glob)).toString
   override def glob(path: String, pattern: String): Seq[String] = {
     path.toAbsolutePath(fileSystems).asDirectory.glob("**/*") map { _.path.fullPath } filterNot { _.toString == path } toSeq
   }
-  override def adjustFilePath(path: String) = if (!path.isGcsUrl) context.root.toAbsolutePath(fileSystems).resolve(path).toString else path
+
+  override def adjustFilePath(path: Path) = if (!path.isAbsolute) context.root.toAbsolutePath(fileSystems).resolve(path) else path
```
cjllanwarne: In keeping with the filesystem split, should this be part of the GCS filesystem rather than the JES backend?

cjllanwarne: If you agree, just add it as a new ticket, no action req'd here :)

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
```diff
@@ -1,29 +1,227 @@
...
+
+  import LocalBackend._
+  import better.files._
+  import cromwell.core.PathFactory._
+
+  import scala.sys.process._
+
+  // Mutable process variable assigned when the execute method is called.
+  // Needs to be accessible to be killed when the job is aborted.
+  private var process: Option[Process] = None
```
cjllanwarne: Put it in state data!

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
```diff
@@ -1,29 +1,227 @@
...
+      case (fqn, v) => splitFqn(fqn)._2 -> v
+    }
+    val knownInputs = unqualifiedWorkflowInputs ++ jobDescriptor.symbolMap
+    val lookup = WdlExpression.standardLookupFunction(knownInputs, declarations, callEngineFunction)
+    new ValueEvaluator(lookup, callEngineFunction)
+  }
+
+  private def evaluate(wdlExpression: WdlExpression) = evaluator.evaluate(wdlExpression.ast)
+
+  // TODO determine how this plays with the init/validation work
```
cjllanwarne: Is this a TODO in this PR, dependent on Francisco?

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
```diff
@@ -1,29 +1,227 @@
...
+    val dockerRun = runtimeAttributes.dockerImage map buildDockerRunCommand getOrElse ""
+    Command(Seq("/bin/bash", "-c", s"cat ${jobPaths.script} | $dockerRun /bin/bash <&0"))
+  }
+
+  // Stream Writers
+  lazy val stdoutWriter = jobPaths.stdout.untailed
+  lazy val stderrTailed = jobPaths.stderr.tailed(100)
+
+  override def preStart() = {
+    // workflowPaths.workflowRoot.createDirectories() TODO move to initialize actor
```
cjllanwarne: Is this TODO still valid? Since you don't know all the Jobs at Initialization time, I guess probably not?

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
```diff
@@ -1,29 +1,227 @@
...
+
+  // TODO determine how this plays with the init/validation work
+  val runtimeAttributes = {
+    val evaluateAttrs = call.task.runtimeAttributes.attrs mapValues evaluate
+    // Fail the call if runtime attributes can't be evaluated
+    val evaluatedAttributes = TryUtils.sequenceMap(evaluateAttrs, "Runtime attributes evaluation").get
+    LocalRuntimeAttributes(evaluatedAttributes)
+  }
+
+  lazy val runsOnDocker = runtimeAttributes.dockerImage.isDefined
```
cjllanwarne: TOL: It would be super-neat if we could have `LocalDocker` as a separate backend:
- To avoid all of this messy decision making in a single BE file. 
- To make extending to RemoteDocker backends much easier
- It isn't a big deal now that we can mix and match Docker and Non-docker backends in the same workflow now
- It'd be an awesome early test case for PBE (since the FS is genuinely shared, we won't have to worry about localisation/delocalisation)
- It would make test cases more focused
- This code file is smaller and simpler

(Genuine TOL - if you disagree or want to merge this PR first, no problem)

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
```diff
@@ -1,29 +1,227 @@
...
+  }
 
   /**
     * Execute a new job.
     */
-  override def execute: Future[BackendJobExecutionResponse] = ???
+  override def execute: Future[BackendJobExecutionResponse] = {
 
-  override protected def jobDescriptor: BackendJobDescriptor = ???
+    val instantiatedCommand = {
```
cjllanwarne: Move this block out to a def

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
```diff
@@ -1,29 +1,227 @@
...
+      val evaluatedInputMappings = call.inputMappings mapValues evaluate
+
+      TryUtils.sequenceMap(evaluatedInputMappings, "Job Input evaluation") flatMap { inputs =>
+        val localizedInputs = localizeInputs(jobPaths, runsOnDocker, fileSystems, inputs ++ workflowInputEntries)
+        call.task.instantiateCommand(localizedInputs, callEngineFunction, pathTransformFunction)
+      }
+    }
+
+    instantiatedCommand match {
+      case Success(command) =>
```
cjllanwarne: Move this success block out into a def

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
```diff
@@ -1,29 +1,227 @@
...
+      case Failure(ex) => Future.successful(BackendJobExecutionFailedResponse(jobDescriptor.key, ex))
+    }
+  }
 
   /**
     * Abort a running job.
     */
-  override def abortJob: Future[JobAbortResponse] = ???
+  override def abortJob: Future[JobAbortResponse] = {
+    Future {
```
cjllanwarne: Is any of this slow or async or likely to fail? If not, can we just do `Future.successful` and avoid spinning up a new thread for the 3 lines?

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
```diff
@@ -1,29 +1,227 @@
...
+    val dockerDir = jobPaths.callDockerRoot
+    s"docker run --rm -v ${jobPaths.callRoot.toAbsolutePath}:$dockerDir -i $image"
+  }
+
+  private def waitAndPostProcess(ec: ExecutionContext): Future[BackendJobExecutionResponse] = Future {
+    val processReturnCode = process map { _.exitValue() } getOrElse (throw new RuntimeException("No process was running"))
+    stdoutWriter.writer.flushAndClose()
+    stderrTailed.writer.flushAndClose()
+
+    def postProcessJob(): BackendJobExecutionResponse = {
```
cjllanwarne: Can `postProcessJob` be its own function? i.e. a top level `private def`?

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
```diff
@@ -1,29 +1,227 @@
...
+    * --rm automatically deletes the container upon exit
+    * -v maps the host workflow executions directory to /root/<workflow id> on the container.
+    * -i makes the run interactive, required for the cat and <&0 shenanigans that follow.
+    */
+  private def buildDockerRunCommand(image: String): String = {
+    val dockerDir = jobPaths.callDockerRoot
+    s"docker run --rm -v ${jobPaths.callRoot.toAbsolutePath}:$dockerDir -i $image"
+  }
+
+  private def waitAndPostProcess(ec: ExecutionContext): Future[BackendJobExecutionResponse] = Future {
```
cjllanwarne: Is the implicit `ExecutionContext` not good enough here?

supportedBackends/local/src/test/scala/cromwell/backend/impl/local/LocalBackendSpec.scala
```diff
@@ -0,0 +1,164 @@
...
+
+    testWorkflow(wf)
+  }
+
+  it should "execute calls with input files and localize them appropriately" in {
+
+    def templateConf(localizers: String) = BackendConfigurationDescriptor("local-config",
+      ConfigFactory.parseString(
+        s"""
+           |local-config {
```
scottfrazer: should this be just `config` instead of `local-config`?

supportedBackends/local/src/test/resources/application.conf
```diff
@@ -0,0 +1,10 @@
+local-config {
+  root = "local-cromwell-executions"
+  filesystems = {
+    local {
+      localization: [
+        "hard-link", "soft-link", "copy"
+      ]
+    }
+  }
+}
```
scottfrazer: I'm not sure how this is used...

supportedBackends/local/src/test/scala/cromwell/backend/impl/local/LocalBackendSpec.scala
```diff
@@ -0,0 +1,164 @@
...
+           |  }
+           |}
+        """.stripMargin)
+    )
+
+    val hardConf = templateConf("hard-link")
+    val symConf = templateConf("soft-link")
+    val copyConf = templateConf("copy")
+
+    val localFs = List(FileSystems.getDefault)
```
scottfrazer: I think this is unused

supportedBackends/local/src/test/scala/cromwell/backend/impl/local/BackendTestkitSpec.scala
```diff
@@ -0,0 +1,88 @@
...
+  implicit val defaultPatience = PatienceConfig(timeout = Span(5, Seconds), interval = Span(500, Millis))
+
+  def testWorkflow(workflow: TestWorkflow) = {
+    val backend = localBackend(jobDescriptorFromSingleCallWorkflow(workflow.workflowDescriptor), workflow.config)
+    executeJobAndAssertOutputs(backend, workflow.expectedResponse)
+  }
+
+  def buildWorkflowDescriptor(wdl: WdlSource,
+                              inputs: Map[String, WdlValue] = Map.empty,
+                              options: WorkflowOptions = WorkflowOptions(JsObject(Map.empty[String, JsValue])),
```
scottfrazer: just thinking out loud...  but might these first three parameters better be a `WorkflowSourceFiles`?

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
```diff
@@ -1,29 +1,227 @@
...
+
+  val evaluator = {
+    val declarations = workflowDescriptor.workflowNamespace.workflow.declarations ++ call.task.declarations
+    val unqualifiedWorkflowInputs = workflowDescriptor.inputs map {
+      case (fqn, v) => splitFqn(fqn)._2 -> v
+    }
+    val knownInputs = unqualifiedWorkflowInputs ++ jobDescriptor.symbolMap
+    val lookup = WdlExpression.standardLookupFunction(knownInputs, declarations, callEngineFunction)
+    new ValueEvaluator(lookup, callEngineFunction)
+  }
```
scottfrazer: I'm just offering some alternatives here...

I was surprised to see that you're creating a `ValueEvaluator`.  What about instead of `val evaluator`, it was `val lookup` then the `evaluate()` method a few lines below can be:

```scala
private def evaluate(wdlExpression: WdlExpression) = wdlExpression.evaluate(lookup, callEngineFunctions)
```

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
```diff
@@ -1,29 +1,227 @@
...
+  lazy val stderrTailed = jobPaths.stderr.tailed(100)
+
+  override def preStart() = {
+    // workflowPaths.workflowRoot.createDirectories() TODO move to initialize actor
+    jobPaths.callRoot.createDirectories()
+  }
 
   /**
     * Execute a new job.
     */
```
scottfrazer: This comment feels unnecessary

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
```diff
@@ -1,29 +1,227 @@
...
+    )
+
+    new LocalCallEngineFunctions(fileSystems, callContext)
+  }
+
+  val evaluator = {
+    val declarations = workflowDescriptor.workflowNamespace.workflow.declarations ++ call.task.declarations
+    val unqualifiedWorkflowInputs = workflowDescriptor.inputs map {
+      case (fqn, v) => splitFqn(fqn)._2 -> v
+    }
```
scottfrazer: Can we change the third parameter on `BackendWorkflowDescriptor` to be `Map[FullyQualifiedName, WdlValue]`?

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
```diff
@@ -1,29 +1,227 @@
...
+    val knownInputs = unqualifiedWorkflowInputs ++ jobDescriptor.symbolMap
```
scottfrazer: based on the [discussion about expression evaluation](https://docs.google.com/document/d/1Hgq2f90iVDZC-3LrFvvLP638PFNcZYWB4rlAzaisjd8/edit?ts=57192cd2&pli=1#)... one proposal was to pass in the lookup function as part of `BackendJobDescriptor` (instead of `symbolMap`).  I just checked, and this is the *only* place we use `symbolMap`.

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
Horneth: Which state data ? BackendJobExecutionActor is an Actor not an FSM :/

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
Horneth: Yes, and also your document about sharing data between stages

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
Horneth: This is only the workflow root : `cromwell-executions/helloWorld/11gfxg513-vcx21vx-fdfdsd/`
We don't need to know the jobs for this

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
Horneth: okay - just curious why ?

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
Horneth: Yeah I'm not 100% sure of what process.destroy() is doing, if it's blocking or just sending SIGTERM and returning. I'll look into it but if it's unnecessary I'll remove the `Future`

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
Horneth: Ok

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
Horneth: If you mean the default implicit `ExecutionContext` (context.dispatcher), no, because it blocks message processing while this future is waiting for the process to finish.
Instead of overriding the ec with another one, I just passed it explicitly to the method to avoid confusion.

supportedBackends/local/src/test/resources/application.conf
Horneth: It's the default conf used in the tests. When you do ConfigFactory.load() this is what you get.

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
Horneth: That would work too yes, I was looking for a way to create something that would allow me to evaluate things everywhere in the class and stumbled upon `ValueEvaluator`. I'll change it

engine/src/main/scala/cromwell/engine/backend/jes/JesEngineFunctions.scala
Horneth: I don't think this should/can be in GCS filesystem, we're prepending the workflow root path to this path if we detect that it's not absolute. For that we need the workflow root which we don't have in GCS Filesystem.
Moreover this might actually be broken I need to double-check.

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
Horneth: I like the idea, a lot would be shared between Local and Docker though so we'd need to abstract that. I'll try to do sthg and if I see it's too big for this I'll ticket it.

supportedBackends/local/src/test/resources/application.conf
scottfrazer: I see `LocalBackendSpec` that you're parsing a similar config stanza manually, so I was just wondering if this was actually used

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
Horneth: I like the idea of passing a lookup function, it'll have to be shard aware though to be able to return the correct index if we're trying to resolve a scatter variable for example. The backend could tell it what its key is.
It could look like 
```scala
lookupFunction: JobDescriptorKey => String => WdlValue
```

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
geoffjentry: I think it is a fantastic idea but out of scope for the moment. At this point I'd say to do the fastest of (in order of my assumed fastness):

- Keeping the local be the dual backend
- Making this local only one or the other
- Separating the backends

And if the answer isn't the last one, open a ticket describing what needs to be done

supportedBackends/local/src/test/scala/cromwell/backend/impl/local/BackendTestkitSpec.scala
Horneth: Hmm yeah, I like to keep it separate because this way you can easily reuse the same Wdl with different inputs without having to create a new `WorkflowSourceFiles`, but it's no big deal either way

supportedBackends/local/src/test/resources/application.conf
Horneth: Yeah I thought I would use this one as a default when I don't really care about the config, and then override it with a custom one when I do.

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
cjllanwarne: I guess it's a personal preference thing, but I much prefer:
```
def instantiateCommand = { ... }

...

def foo {
  instantiatedCommand = instantiateCommand()
}
```

geoffjentry: I almost made the same comment the other day on a much smaller val closure. I'm not a big fan of multi-line vals although a couple of lines don't bother me *that* much. However, when one feels the need to split out a full function inside a val closure it seems excessive

For me it has more to do with the closure than it being a val, if something is building up to anything beyond trivial complexity IMO it should have a name.

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
cjllanwarne: Can it become an FSM?

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
scottfrazer: We talked about this in person quite extensively and came up with something we both like, it will be added to the linked design doc as a proposal.

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
Horneth: I'm ok with putting it outside, it's just never used anywhere else.
I'd rather make `val`s things that can be though. Even though in this case this will very likely be called only once, since the command can be completely built at instatiation time, making it a val or lazy val makes more sense to me.

engine/src/main/scala/cromwell/engine/backend/jes/JesEngineFunctions.scala
cjllanwarne: Fair enough. I'll bow to your knowledge on this stuff!

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
cjllanwarne: Oh right! In that case yes, certainly Initialization actor if possible

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
cjllanwarne: Same reason as above. Not a blocker.

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
cjllanwarne: Do you mean the engine creates a sub-loookup which only contains a specific scatter's variables...? And that's the function which is passed to the backend at execute time?

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
cjllanwarne: Ouch. That's awkward. Can we achieve the same thing by farming the work out to a sub- or anonymous actor?

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
Horneth: I guess so - we would have to let the backend implementor provide it's own state data ADT etc though.. We can definitely look into it but for this PR it'll be fine I think, especially since abort is not really part of the MVP for this. Also, state data is really a disguised `var` (with more thread-safeness, granted).

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
geoffjentry: @Horneth For this PR do whatever is the quickest thing to implement which will demonstrate a working local backend for at least some workflows, even if it can't handle everything (which hopefully is whatever is here now). We'll loop back around w/ the design doc discussion and clean up afterwards as necessary

engine/src/main/scala/cromwell/engine/backend/jes/JesEngineFunctions.scala
Horneth: I'll remove this method to merge it with `toPath` though

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalBackend.scala
Horneth: > working local backend for at least some workflows

Yes it's what this is - I'd just like to rebase on top of francisco's PR to remove the duplicated code but as far as I'm concerned this is good to be reviewed / thumbs uped / merged whenever everyone is happy

Horneth:
@cjllanwarne @scottfrazer @geoffjentry Rebased and most comments addressed - I left the docker split untouched for now

cjllanwarne:
Hmm... it's a shame the `var` is still there. I may see if I can FSM this if I have a spare moment but otherwise this looks great! 👍

scottfrazer:
👍

PR #713, opened 2016-04-17
Allow listing more than 1000 files in GCS.  Closes #699 against master.

### LINKED ISSUE #699: wdl glob / Array[File] only seems to grab 1000 elements

mcovarr:
Same code as #710 but for master.  I really should have done this fix [gitflow](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow) style and merged this branch to both master and develop.  Oh well.


geoffjentry:
For once you could have actually used that favorite gitflow link of yours ....

:+1:

cjllanwarne:
👍

PR #720, opened 2016-04-19
Demonstrates all lifecycle states for a single-call workflow

cjllanwarne:
To set expectations:

- Demonstrates the ShadowWorkflowActor progressing through every lifecycle state
- Hard-codes the DummyBackend for every call
- Doesn't store anything persistently
- Will certainly need to evolve from here...

### DIFF
--- a/backend/src/main/scala/cromwell/backend/BackendJobExecutionActor.scala
+++ b/backend/src/main/scala/cromwell/backend/BackendJobExecutionActor.scala
@@ -28,7 +28,7 @@ object BackendJobExecutionActor {
 /**
   * Workflow-level actor for executing, recovering and aborting jobs.
   */
-trait BackendJobExecutionActor extends BackendLifecycleActor with ActorLogging {
+trait BackendJobExecutionActor extends BackendJobLifecycleActor with ActorLogging {
 
   def receive: Receive = LoggingReceive {
     case ExecuteJobCommand(jobDescriptor) => performActionThenRespond(execute(jobDescriptor), onFailure = executionFailed(jobDescriptor.key))
--- a/backend/src/main/scala/cromwell/backend/BackendJobExecutionActor.scala
+++ b/backend/src/main/scala/cromwell/backend/BackendJobExecutionActor.scala
@@ -40,18 +40,20 @@ trait BackendJobExecutionActor extends BackendJobLifecycleActor with ActorLoggin
   private def executionFailed(key: BackendJobDescriptorKey) = (t: Throwable) => BackendJobExecutionFailedResponse(key, t)
   private def abortFailed(key: BackendJobDescriptorKey) = (t: Throwable) => BackendJobExecutionAbortFailedResponse(key, t)
 
+  override def jobDescriptor: BackendJobDescriptor
+
   /**
     * Execute a new job.
     */
-  def execute(jobDescriptor: BackendJobDescriptor): Future[BackendJobExecutionResponse]
+  def execute(): Future[BackendJobExecutionResponse]
 
   /**
     * Restart or resume a previously-started job.
     */
-  def recover(jobDescriptor: BackendJobDescriptor): Future[BackendJobExecutionResponse]
+  def recover(): Future[BackendJobExecutionResponse]
 
   /**
     * Abort a running job.
     */
-  def abortJob(jobKey: BackendJobDescriptorKey): Future[JobAbortResponse]
+  def abortJob(): Future[JobAbortResponse]
 }
--- a/backend/src/main/scala/cromwell/backend/BackendJobExecutionActor.scala
+++ b/backend/src/main/scala/cromwell/backend/BackendJobExecutionActor.scala
@@ -13,8 +13,8 @@ object BackendJobExecutionActor {
 
   // Commands
   sealed trait BackendJobExecutionActorCommand extends BackendWorkflowLifecycleActorCommand
-  final case class ExecuteJobCommand(jobDescriptor: BackendJobDescriptor) extends BackendJobExecutionActorCommand
-  final case class RecoverJobCommand(jobDescriptor: BackendJobDescriptor) extends BackendJobExecutionActorCommand
+  case object ExecuteJobCommand extends BackendJobExecutionActorCommand
+  case object RecoverJobCommand extends BackendJobExecutionActorCommand
 
   // Responses
   sealed trait BackendJobExecutionActorResponse extends BackendWorkflowLifecycleActorResponse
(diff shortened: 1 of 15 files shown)


cjllanwarne:
@gauravs90 again and @Horneth please!

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -64,6 +80,13 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId, workflowDescript
   }
 
   when(WorkflowExecutionInProgressState) {
+    case Event(BackendJobExecutionSucceededResponse(jobKey, callOutputs), stateData) =>
+      log.info(s"Job ${jobKey.call.fullyQualifiedName} succeeded! Outputs: ${callOutputs.mkString(" |:::| ")}")
```
geoffjentry: what the heck is this?

cjllanwarne: The "Pipe Triple-colon Pipe"? I call it the accordion operator

cjllanwarne: Would you prefer a `\n`?

backend/src/main/scala/cromwell/backend/BackendLifecycleActor.scala
```diff
@@ -63,3 +53,19 @@ trait BackendLifecycleActor extends Actor {
...
+trait BackendWorkflowLifecycleActor extends BackendLifecycleActor {
+  /**
+    * The workflow descriptor for the workflow in which this Backend is being used
+    */
+  protected def workflowDescriptor: BackendWorkflowDescriptor
+
+  /**
+    * The subset of calls which this backend will be expected to run
+    */
+  protected def calls: Seq[Call]
```
gauravs90: Why do we need it separately? Can't it be accessed from the namespace? (`namespace.workflow.calls`)

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -64,6 +80,13 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId, workflowDescript
...
+      context.parent ! WorkflowExecutionSucceededResponse
+      goto(WorkflowExecutionSuccessfulState)
+    case Event(BackendJobExecutionFailedResponse(jobKey, reason), stateData) =>
+      log.warning(s"Job ${jobKey.call.fullyQualifiedName} succeeded! Reason: $reason")
```
gauravs90: Job success shouldn't be a warning.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
geoffjentry: That seems like it'd be more readable. Although I'm questioning if the outputs really need to be in the log in the first place, I suppose we could always remove it later if it ends up being too chatty

backend/src/main/scala/cromwell/backend/BackendLifecycleActor.scala
Horneth: `namespace.workflow.calls` contains all the calls in the workflow, this will be the subset of calls that this particular backend will execute

cjllanwarne: Exactly

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowFinalizationActor.scala
```diff
@@ -50,26 +50,30 @@ case class WorkflowFinalizationActor(workflowId: WorkflowId, workflowDescriptor:
...
-      // Create each FinalizationActor with the list of calls which ere assigned to that backend.
-      val backendFinalizationActors = backendWorkflowActors(backendAssignments)
-      backendFinalizationActors.keys foreach { _ ! Finalize }
-      goto(WorkflowFinalizing) using WorkflowFinalizationActorData(backendFinalizationActors, List.empty, Map.empty)
-    case Event(AbortEngineFinalizationCommand, _) =>
+  startWith(FinalizationPendingState, WorkflowFinalizationActorData(Map.empty, List.empty, Map.empty))
+
+  when(FinalizationPendingState) {
+    case Event(StartFinalizationCommand, _) =>
+      val backendFinalizationActors = backendWorkflowActors(workflowDescriptor.backendAssignments)
```
gauravs90: `backendWorkflowActors(...)` actually creates `BackendWorkflowInitializationActors`, no?

gauravs90:
Some minor stuff to take care of, but PullApprove: 👍

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
Horneth: Hmm this looks more like a wrong copy/paste, should be `failed` instead of `succeeded` here

backend/src/main/scala/cromwell/backend/BackendJobExecutionActor.scala
```diff
@@ -28,7 +28,7 @@ object BackendJobExecutionActor {
 /**
   * Workflow-level actor for executing, recovering and aborting jobs.
   */
-trait BackendJobExecutionActor extends BackendLifecycleActor with ActorLogging {
+trait BackendJobExecutionActor extends BackendJobLifecycleActor with ActorLogging {
```
Horneth: So if this actor is now job specific why not make the jobDescriptor a `def` in the trait and remove it from the signature of `execute` `recover` and `abort` ?

Horneth: It makes it more friendly on the implementation side because you don't have to pass around a jobDescriptor to every method you write although you know for a fact that your actor is scoped with this job already

cjllanwarne: Great point!

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowFinalizationActor.scala
cjllanwarne: Oh, oops - the variables were poorly named in WorkflowLifecycleActor due to copy/paste. But this method indirectly calls the `backendActor()` method which in `WorkflowFinalizationActor` creates `BackendWorkflowFinalizationActor`s.

backend/src/main/scala/cromwell/backend/BackendJobExecutionActor.scala
cjllanwarne: Ha, in fact it was already in the trait. I think the method-level `jobDescriptor`s were just left over from when it was workflow scoped

cjllanwarne:
@Horneth should be ready now for another round

Horneth:
👍

PR #721, opened 2016-04-19
No longer materializing workflows when testing `SlickDataAccess`.

kshakir:
MySQL tests are no longer optional when running DbmsTests.
Removed duplicate hsqldb test.


kshakir:
Another weekend hack, no rush on the PR review.

On my test mysql database with hundreds of dummy workflows, the spec was materializing _all_ of the dummy workflows before then filtering, and timing out.

geoffjentry:
:+1:

engine/src/main/resources/application.conf
```diff
@@ -218,27 +218,13 @@ database {
...
-      db.url = "jdbc:hsqldb:mem:testdb;shutdown=false;hsqldb.tx=mvcc"
-      db.driver = "org.hsqldb.jdbcDriver"
-      db.connectionTimeout = 1000 // NOTE: 1000ms (the default) is ok for a small hsqldb, but often too short for mysql
-      driver = "slick.driver.HsqldbDriver$"
-      liquibase = {
-        changelog = "src/main/migrations/changelog.xml"
-        connection = "liquibase.database.jvm.HsqlConnection"
-      }
-    }
-
```
Horneth: Why is this disappearing ?

kshakir: It was used to test liquibase-on-hsqldb when mysql wasn't available.

Now that we use liquibase-on-hsqldb for all schema creation, test.hsqldb is an exact duplicate of the main.hsqldb tests.

Horneth:
👍

PR #727, opened 2016-04-22
Ensuring directories are created in the SFS localizers. Closes #646

### LINKED ISSUE #646: When cromwell runs a job, it must copy the input files (local backend, cromwell 0.19)
Development of WDL scripts would be *much* easier if we did not have to copy input files during execution of a job.  This causes hardships, since the bam files might be quite large.  

I tried modifying ``application.conf`` to eliminate the "copy" localization strategy.  However, this just caused a crash ("unable to localize...").  In other words, hard links and soft links seem to be supported, but do not function.

"Use smaller bam files for testing WDL" is not a good answer, since:

1.  This would require a lot of setup work solely for testing WDL scripts.
2.  This would rule out WDL/cromwell for load testing and automated evaluations, which are run frequently.  We regularly have to process hundreds of bam files for a single evaluation.
3.  We have several auxiliary files that are quite large and these are copied, as well.  Add this to point #1 above, since we would have to strip these down as well.
4.  This would not suss out problems that appear with larger bam files -- a common problem.  
5.  Our tools already run quickly on large files.

Many thanks!

kshakir:
Updated the travis configuration to again use copy-only localization.
As the other localizers were previously broken, travis was _only_ able to run copy.

### DIFF
--- a/engine/src/main/scala/cromwell/engine/backend/local/SharedFileSystemBackend.scala
+++ b/engine/src/main/scala/cromwell/engine/backend/local/SharedFileSystemBackend.scala
@@ -1,15 +1,13 @@
 package cromwell.engine.backend.local
 
-import java.io.File
-import java.nio.file.{FileSystem, Files, Path, Paths}
+import java.nio.file._
 
-import better.files.{File => ScalaFile, _}
+import better.files._
 import cromwell.core.{CallOutput, CallOutputs, WorkflowOptions}
 import cromwell.engine.backend._
 import cromwell.engine.io.gcs.GcsPath
 import cromwell.engine.{ExecutionEventEntry, backend}
 import cromwell.util.TryUtil
-import org.apache.commons.io.FileUtils
 import org.apache.commons.lang3.exception.ExceptionUtils
 import wdl4s.types.{WdlArrayType, WdlFileType, WdlMapType}
 import wdl4s.values.{WdlValue, _}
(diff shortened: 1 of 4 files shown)


engine/src/test/scala/cromwell/engine/backend/local/SharedFileSystemBackendSpec.scala
```diff
@@ -0,0 +1,312 @@
...
+    val result = SharedFileSystemBackend.localizePathViaSymbolicLink(orig.fullPath, dest.path, workflowDescriptor)
+    result should be(Success(()))
+
+    dest.append(".")
+
+    orig.contentAsString should be("hello world.")
+    countLinks(orig) should be(1)
+    countLinks(dest) should be(1)
+    isSymLink(dest) should be(true)
+    dest.contentAsString should be("hello world.")
```
Horneth: You can also `Files.isSymbolicLink(dest.toPath) shouldBe true`

engine/src/test/scala/cromwell/engine/backend/local/SharedFileSystemBackendSpec.scala
```diff
@@ -0,0 +1,312 @@
...
+
+    origDir.delete(ignoreIOExceptions = true)
+    destDir.delete(ignoreIOExceptions = true)
+  }
+
+  it should "localize from gcs to a directory" taggedAs IntegrationTest in {
```
Horneth: nice

engine/src/main/scala/cromwell/engine/backend/local/SharedFileSystemBackend.scala
```diff
@@ -23,31 +21,41 @@ import scala.util.{Failure, Success, Try}
 object SharedFileSystemBackend {
   type LocalizationStrategy = (String, Path, WorkflowDescriptor) => Try[Unit]
 
-  private def localizeFromGcs(originalPath: String, executionPath: Path, descriptor: WorkflowDescriptor): Try[Unit] = Try {
+  private[local] def localizeFromGcs(originalPath: String, executionPath: Path,
```
Horneth: Just FYI, In the new LocalBackend I made `originalPath` be a `Path` too, so we don't have to have this special case for GCS, it'll be built in the path already.

Horneth:
👍

engine/src/main/scala/cromwell/engine/backend/local/SharedFileSystemBackend.scala
geoffjentry: @Horneth Can you add a ticket about this?

geoffjentry:
👍

engine/src/main/scala/cromwell/engine/backend/local/SharedFileSystemBackend.scala
Horneth: Hmm yes but I'm not sure what it should say ?

geoffjentry: @Horneth I was reading your bit as meaning that if this were being developed in the not too distant future it should be done differently due to changes you're implementing. If that's the case it's really just needs to be a reminder w/ enough info that people will know what it means, or if I misunderstood then forget about it :)

Issue #695, opened 2016-04-12
Cromwell out of memory error large array


jsotobroad:
We are trying to scatter over a large array and it seems we ran out of heap space.  The object was an array of arrays.  The outer array had a length of 3457 and the inner arrays had a length of 3 each.  The values of these were gs paths.  This happened in the process of starting scattered calls over this array.

Here is the part of the wdl where this is happening.  `partition_indexing_list` is a list of integers from 0-3456 used to index into the array of arrays `RotateGVCF.output_matrix`, `RotateGVCFIndex.output_matrix` and the array of strings `tiledb_partitions_list`

  scatter (idx in partition_indexing_list) {
    call TileDBCombineGVCF {
      input:
        output_gvcf_filename = "output.vcf.gz",
        gvcf_list = RotateGVCF.output_matrix[idx],
        gvcf_index_list = RotateGVCFIndex.output_matrix[idx],
        ref_fasta = ref_fasta,
        ref_fasta_index = ref_fasta_index,
        partition = tiledb_partitions_list[idx],
        size_per_column_partition = size_per_column_partition,
        vid_mapping_file = vid_mapping_file,
        vcf_header = vcf_header,
        sample_name_map = sample_name_map,
        disk_size = medium_disk
    }

**snippet of the array in question:**

[...]

PR #728, opened 2016-04-22
Lowering per upload memory for GCS to the minimum 256K.

kshakir:
Added an undocumented config option to tweak this value if we need to adjust it for further testing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
 # Cromwell Change Log
 
 ## 0.20
+
+* The default per-upload bytes size for GCS is now the minumum 256K
+instead of 64M. There is also an undocumented config key
+`google.upload-buffer-bytes` that allows adjusting this internal value.
--- a/filesystems/gcs/src/main/scala/cromwell/filesystems/gcs/GcsFileSystemProvider.scala
+++ b/filesystems/gcs/src/main/scala/cromwell/filesystems/gcs/GcsFileSystemProvider.scala
@@ -12,6 +12,7 @@ import java.util.Collections
 import java.util.concurrent.{AbstractExecutorService, TimeUnit}
 
 import com.google.api.client.googleapis.json.GoogleJsonResponseException
+import com.google.api.client.googleapis.media.MediaHttpUploader
 import com.google.api.services.storage.Storage
 import com.google.api.services.storage.model.StorageObject
 import com.google.cloud.hadoop.gcsio.{GoogleCloudStorageReadChannel, GoogleCloudStorageWriteChannel, ObjectWriteConditions}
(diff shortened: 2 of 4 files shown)


kshakir:
Blocked by https://github.com/broadinstitute/lenthall/pull/14

kshakir:
A similar patch was tested by @jsotobroad. Once this PR against develop is reviewed & merged, we can submit another PR against master, and _that one_ should make #695 good to go.

kshakir:
@geoffjentry & @Horneth for review please, including the lenthall patch needed to get this cromwell build repaired

filesystems/gcs/src/main/scala/cromwell/filesystems/gcs/GcsFileSystemProvider.scala
```diff
@@ -82,12 +83,21 @@ object ExecutionContextExecutorServiceBridge {
   * This implementation is not complete and mostly a proof of concept that it's possible to *copy* around files from/to local/gcs.
   * Copying is the only functionality that has been successfully tested (same and cross filesystems).
   *
+  * If/when switching to Google's GCS NIO implementation, callers may need to implement various utilities built into
+  * this implementation, including:
+  *
+  * - Minimizing the upload buffer size, assuming the default is also on the order of megabytes of memory per upload
+  * - Automatically retrying transient errors
+  * - etc.
+  *
```
Horneth: Yes, I think when we switch to google's NIO we should still write a wrapper around it to take care of these things

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
 # Cromwell Change Log
 
 ## 0.20
+
+* The default per-upload bytes size for GCS is now the minumum 256K
+instead of 64M. There is also an undocumented config key
```
Horneth: Is it undocumented on purpose ? If we want to make it configurable (which I'm not sure we need to) then we should drop a line in the README IMO.

Horneth:
👍

CHANGELOG.md
geoffjentry: At the moment we weren't sure how to handle this or what the ideal default would be. Putting it in the config allows gotc to fiddle with it easily. I'm guessing undocumented so that we can remove it later if it makes sense do handle this some other way

Horneth: Ok then

geoffjentry:
In case @kshakir isn't the one managing the master patch today, reminder that lenthall needs to be deal with in that situation as well

geoffjentry:
:+1:

kshakir:
This should go green soon and be good to merge. I'll submit the master patch tomorrow, unless someone else gets to it first.

PR #734, opened 2016-04-25
FSM local job executor to remove process var

Horneth:
This gets rid of the `var` in Local backend by delegating the work to an FSM that can hold on to the process reference using state data instead.

### DIFF
--- a/supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalJobExecutionActor.scala
+++ b/supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalJobExecutionActor.scala
@@ -1,18 +1,16 @@
 package cromwell.backend.impl.local
 
-import java.nio.file.{FileSystems, Path, Paths}
+import java.nio.file.FileSystems
 
-import cromwell.backend.BackendJobExecutionActor.{BackendJobExecutionAbortedResponse, BackendJobExecutionFailedResponse, BackendJobExecutionResponse, BackendJobExecutionSucceededResponse}
-import cromwell.backend.BackendLifecycleActor.{BackendJobExecutionAbortFailedResponse, BackendJobExecutionAbortSucceededResponse, JobAbortResponse}
+import cromwell.backend.BackendJobExecutionActor.BackendJobExecutionResponse
+import cromwell.backend.BackendLifecycleActor.JobAbortResponse
 import cromwell.backend._
-import cromwell.core.CallContext
+import cromwell.backend.impl.local.LocalJobExecutionFSM.{Abort, Run}
 import org.slf4j.LoggerFactory
 import wdl4s._
-import wdl4s.values.{WdlFile, WdlValue}
 
-import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.{Future, Promise}
 import scala.language.postfixOps
-import scala.util.{Failure, Success, Try}
 
 object LocalJobExecutionActor {
   val ProcessKilledCode = 143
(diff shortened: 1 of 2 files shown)


supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalJobExecutionFSM.scala
```diff
@@ -0,0 +1,248 @@
...
+  case object Running extends JobState
+  case object Aborting extends JobState
+  protected case class StateData(process: Option[Process])
+
+  val ProcessKilledCode = 143
+  val logger = LoggerFactory.getLogger("LocalBackend")
+  // TODO Support GCS ?
+  val fileSystems = List(FileSystems.getDefault)
+
+  private def splitFqn(fullyQualifiedName: FullyQualifiedName): (String, String) = {
```
mcovarr: case class plz

mcovarr:
Not sure I see why this needs to be broken up into two actors?  The outer one is a super-thin shim over the inner one that does all the heavy lifting.

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalJobExecutionFSM.scala
```diff
@@ -0,0 +1,248 @@
...
+import org.slf4j.LoggerFactory
+import wdl4s._
+import wdl4s.values.{WdlFile, WdlValue}
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.{Future, Promise}
+import scala.language.postfixOps
+import scala.sys.process._
+import scala.util.{Failure, Success, Try}
+
+object LocalJobExecutionFSM {
```
mcovarr: This is an FSM, but to my broader point I'm not sure what the responsibilities of this class actually are in contrast with its parent.

Horneth:
Maybe I'm missing something, but the outer actor is completely unusable as an actor for the backend implementor. First it's an actor, not an FSM, but even if we'd make it an FSM we wouldn't control the state flow of this actor. The only think the backend does is implement methods that returns future. So we can't change its state data.

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalJobExecutionFSM.scala
```diff
@@ -0,0 +1,248 @@
...
+  sealed trait JobMessages
+  case class Run(executePromise: Promise[BackendJobExecutionResponse])
+  case class Abort(abortPromise: Promise[JobAbortResponse])
```
mcovarr: probably these two meant to extend `JobMessages`?

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalJobExecutionFSM.scala
```diff
@@ -0,0 +1,248 @@
...
+  }
+
+  when(Running) {
+    case Event(Abort(abortPromise), data) => abort(abortPromise, data)
+    case Event(Run(_), data) =>
+      logger.warn("Received Run message before but job is already running. Ignoring.")
+      stay()
+  }
+
+  whenUnhandled { NullFunction }
```
mcovarr: silently ignoring unrecognized messages doesn't seem like the best policy... 😄

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalJobExecutionFSM.scala
mcovarr: which for consistency should maybe be `JobMessage`

mcovarr:
I think technically these could be collapsed into one class, but from your description maybe it's preferable to keep this structure with the roles of the classes more clearly articulated?  The outer actor has the responsibility of implementing the `BackendJobExecutionActor` trait, but can be implemented as a simple adapter to any backend-specific means of executing jobs, which here just happens to be an FSM.  Not sure what the best nomenclature would be for this distinction.

geoffjentry:
@mcovarr If I understand the issue correctly, which I probably don't at the moment, this is falling under the umbrella of "and then the backend implementor chooses how to do this" (aside: apparently both implementor and implementer are correct) and in this case there's a deeper actor structure. Does that sound right here?

mcovarr:
@geoffjentry yes that sounds right.

On the spelling front, I was suggesting we pick one of the two accepted spellings and stick with that.  😄

geoffjentry:
@mcovarr I hadn't even seen you talking about it. I wrote it one way and thought it didn't look right and then just googled it to realize why my spellcheck wasn't getting angry with me.

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalJobExecutionActor.scala
```diff
@@ -31,190 +29,21 @@ object LocalJobExecutionActor {
 }
 
 class LocalJobExecutionActor(override val jobDescriptor: BackendJobDescriptor,
-                   override val configurationDescriptor: BackendConfigurationDescriptor) extends BackendJobExecutionActor with SharedFileSystem {
+                   override val configurationDescriptor: BackendConfigurationDescriptor) extends BackendJobExecutionActor {
 
-  import LocalJobExecutionActor._
-  import better.files._
-  import cromwell.core.PathFactory._
+  val executionFSM = context.actorOf(LocalJobExecutionFSM.props(jobDescriptor, backendConfiguration))
```
cjllanwarne: I'd make this a lazy val, in case it's never used

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalJobExecutionFSM.scala
cjllanwarne: I asked for this FSM... but it looks like you can't extend the JobExecutionActor base trait with FSM. So the parent's job it interfacing with the engine. This one's job is actually doing the work as an FSM.

If it turns out in the future that every JobExecutionActor wants to be an FSM, maybe we can modify the trait and remove the need for the thin layer

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalJobExecutionFSM.scala
```diff
@@ -0,0 +1,248 @@
...
+      case Success(command) =>
+        val process = spawnProcess(command)
+        val futureResult = waitAndPostProcess(process)
+        futureResult onComplete {
+          case _ => self ! PoisonPill
+        }
+        executePromise completeWith futureResult
+        goto(Running) using data.copy(process = Option(process))
+      case Failure(ex) =>
+        executePromise success BackendJobExecutionFailedResponse(jobDescriptor.key, ex)
```
cjllanwarne: `success`?

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalJobExecutionFSM.scala
```diff
@@ -0,0 +1,248 @@
...
+        context.stop(self)
```
cjllanwarne: Is there a reason for using `context.stop(self)` here, but `self ! PoisonPill` above?

Horneth:
@mcovarr Yeah the reason for this PR is that there is currently a mutable `var` that holds the process when it's created, so it can be accessed by the abort method. I know `var`s are evil but in this case creating an underlying FSM just to get rid of this `var` does seem an overkill to me. I couldn't find a way to use the `BackendJobExecutionActor` itself to encapsulate this mutable state though, which is why I ended up with this. I can give it another shot.

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalJobExecutionFSM.scala
Horneth: `success` completes the promise (and hence the future) successfully, but with a `BackendJobExecutionFailedResponse`

cjllanwarne:
@Horneth Since I asked for this originally - I would say it's not a high-priority and if you've looked at it and it feels like overkill just to make it work, no problem to me in just quietly dropping it

Horneth:
@cjllanwarne ok, I'm thinking the same, I'll close the PR but leave the branch in case we realize later we can't possibly live with this :p

geoffjentry:
@Horneth having a private var inside an actor is IMO completely fine *iff* there aren't futures kicking off inside of it.

PR #739, opened 2016-04-26
Databases are hard.  closes #736

### LINKED ISSUE #736: Fix the slow left joins
`ExecutionComponent.executionsWithCacheHitWorkflowAndCall` is slow and it's making people sad

scottfrazer:
- [x] Test with MySQL
- [x] Test with in-memory DB
- [x] Test against `master` too
- [x] Create PR against `master`

### DIFF
--- a/engine/src/main/scala/cromwell/engine/db/slick/ExecutionComponent.scala
+++ b/engine/src/main/scala/cromwell/engine/db/slick/ExecutionComponent.scala
@@ -7,11 +7,13 @@ import cromwell.engine.ExecutionStatus
 import cromwell.engine.ExecutionStatus.ExecutionStatus
 import cromwell.engine.db.ExecutionDatabaseKey
 import cromwell.engine.workflow.BackendCallKey
+import slick.dbio.Effect.Read
+import slick.jdbc.GetResult
 import slick.profile.RelationalProfile.ColumnOption.Default
 import wdl4s.{NamespaceWithWorkflow, Scatter}
 
 import scala.language.postfixOps
-import scala.util.{Success, Failure, Try}
+import scala.util.{Failure, Success, Try}
 
 case class CallCacheHit(workflowId: String, callName: String)
 case class ExecutionWithCacheData(execution: Execution, cacheHit: Option[CallCacheHit])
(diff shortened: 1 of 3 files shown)


scottfrazer:
@dvoet and @kshakir, can you take a look?  I'm not sure if this can be done in a better way!

engine/src/main/scala/cromwell/engine/db/slick/ExecutionComponent.scala
```diff
@@ -167,39 +170,34 @@ trait ExecutionComponent {
...
-        case (((e, _), _), we) => e.workflowExecutionId === we.workflowExecutionId
-      }
-      if workflowExecution.map(_.workflowExecutionUuid) === workflowExecutionUuid
-    } yield (execution, cacheHitWorkflow.map(_.workflowExecutionUuid), cacheHitExecution.map(_.callFqn))
-  )
+  import slick.dbio.DBIOAction
+
+  def executionsWithCacheHitWorkflowAndCall(workflowExecutionUuid: String): DBIOAction[Vector[ExecutionWithCacheData], NoStream, Read] = {
+    implicit val converter = GetResult { r =>
+      val execution = Execution(r.<<, r.<<, r.<<, r.<<, r.<<, r.<<, r.<<, r.<<, r.<<, r.<<, r.<<, r.<<, r.<<, r.<<)
```
scottfrazer: @abaumann so many alligators.... at least they're not fighting this time!

engine/src/main/scala/cromwell/engine/db/slick/ExecutionComponent.scala
```diff
@@ -167,39 +170,34 @@ trait ExecutionComponent {
...
+    sql"""
+       SELECT e.EXECUTION_ID, e.WORKFLOW_EXECUTION_ID, e.CALL_FQN, e.IDX,
+              e.STATUS, e.RC, e.START_DT, e.END_DT, e.BACKEND_TYPE, e.ALLOWS_RESULT_REUSE,
+              e.DOCKER_IMAGE_HASH, e.RESULTS_CLONED_FROM, e.EXECUTION_HASH, e.ATTEMPT,
+              we1.WORKFLOW_EXECUTION_UUID, e1.CALL_FQN
+       FROM execution e
+       LEFT JOIN execution e1 ON e.results_cloned_from=e1.execution_id
+       LEFT JOIN workflow_execution we1 ON e1.workflow_execution_id=we1.workflow_execution_id
+       LEFT JOIN workflow_execution we ON e.workflow_execution_id=we.workflow_execution_id
+       WHERE we.workflow_execution_uuid=$workflowExecutionUuid
```
scottfrazer: Do I need `quoteIdentifier` here?

engine/src/main/scala/cromwell/engine/db/slick/ExecutionComponent.scala
abaumann: good god, what the hell is this alligator conga line supposed to mean

geoffjentry: I thought they were spooning

scottfrazer: haha!  I stole this idea from [the Rawls codebase](https://github.com/broadinstitute/rawls/blob/develop/src/main/scala/org/broadinstitute/dsde/rawls/dataaccess/slick/EntityComponent.scala#L51)

I'm definitely not opposed to a better way to do it!

dvoet: I think this is how it is done. I think they are done fighting and one is chasing and about to eat the other.

engine/src/main/scala/cromwell/engine/db/slick/ExecutionComponent.scala
dvoet: only if it does not work in the databases you support.

engine/src/main/scala/cromwell/engine/db/slick/ExecutionComponent.scala
```diff
@@ -7,16 +7,20 @@ import cromwell.engine.ExecutionStatus
...
 import scala.language.postfixOps
-import scala.util.{Success, Failure, Try}
+import scala.util.{Failure, Success, Try}
 
 case class CallCacheHit(workflowId: String, callName: String)
 case class ExecutionWithCacheData(execution: Execution, cacheHit: Option[CallCacheHit])
 
-case class Execution(workflowExecutionId: Int,
+/** NOTE: the attributes should be in column-order so that executionsWithCacheHitWorkflowAndCall can work */
+case class Execution(executionId: Option[Int] = None,
```
dvoet: you don't need to change the order here, you just need the column order in your select statement to match your constructor order.

engine/src/main/scala/cromwell/engine/db/slick/ExecutionComponent.scala
jmthibault79: wackwack or bust!

dvoet:
I think the changes to the case class and column ordering are not required but not harmful either.

👍

engine/src/main/scala/cromwell/engine/db/slick/ExecutionComponent.scala
scottfrazer: I'll at least update the comment

engine/src/main/scala/cromwell/engine/db/slick/ExecutionComponent.scala
```diff
@@ -7,16 +7,19 @@ import cromwell.engine.ExecutionStatus
...
 
 import scala.language.postfixOps
-import scala.util.{Success, Failure, Try}
+import scala.util.{Failure, Success, Try}
 
 case class CallCacheHit(workflowId: String, callName: String)
 case class ExecutionWithCacheData(execution: Execution, cacheHit: Option[CallCacheHit])
 
-case class Execution(workflowExecutionId: Int,
+case class Execution(executionId: Option[Int] = None,
```
kshakir: I'll allow it I as this class is already borked, but in general I'm not a fan of default arguments before non-default. I'm not quite sure how and when this class ended up with this hybrid approach. Having defaults before non-defaults forces the caller to use named arguments _only_, i.e. no short form `Execution("fqn", 1, "Starting")`

scottfrazer: Honestly I just blindly did an IntelliJ refactor because I thought it was necessary but it turns out it's not necessary.  I can change it back to the original signature and it'll also make the diff nicer

kshakir:
👍

PR #743, opened 2016-04-27
Extend single call workflow execution to N-step execution. Closes #652.

### LINKED ISSUE #652: Implement Workflow Execution using Backend Interface

gauravs90:
Changes are related only to the Shadow world.

The expectations of this PR is to extend the current state of things in Workflow Execution (i.e currently we only run a single call workflow) to allow arbitrarily sized workflows (i.e. an N-call workflow). The intention is *not* to support scatters in this PR, but allow it to be extensible for scatters (or Inception-esque nested scatters, which I hope to take up as my next ticket).

~~The original WA used Data Access and symbol store to pass around information between tasks. I am not quite sure how that would work with the shadow world, also considering we don't (yet) have engine functions at that level. So I have used a little different algorithm to orchestrate the calls in a workflow (preparing a small call graph and sorting that graph to obtain the logical ordering among tasks, and then orchestrate that via information in the FSM state data).~~

~~I might wait for @Horneth for getting the engine functions in and have thoughts from you guys on plugging in outputs of a task to it's dependent  task.~~

~~I talked with Thibault about this and I honestly don't mind if this PR does't get merged at all if we see a problem with this, just need a fresh pair of eyes to look through it.~~

~~Currently, I'm adding tests for all this new code.~~

### DIFF
--- a/engine/src/main/scala/cromwell/engine/workflow/SingleWorkflowRunnerActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/SingleWorkflowRunnerActor.scala
@@ -8,6 +8,7 @@ import better.files._
 import cromwell.core.{CallOutput, WorkflowId}
 import cromwell.engine
 import cromwell.engine._
+import cromwell.engine.workflow.ShadowWorkflowManagerActor.SubmitWorkflowCommand
 import cromwell.engine.workflow.SingleWorkflowRunnerActor._
 import cromwell.engine.workflow.WorkflowManagerActor._
 import cromwell.webservice.CromwellApiHandler._
@@ -17,8 +18,8 @@ import spray.json._
 import scala.util._
 
 object SingleWorkflowRunnerActor {
-  def props(source: WorkflowSourceFiles, metadataOutputFile: Option[Path], workflowManager: ActorRef): Props = {
-    Props(classOf[SingleWorkflowRunnerActor], source, metadataOutputFile, workflowManager)
+  def props(source: WorkflowSourceFiles, metadataOutputFile: Option[Path], workflowManager: ActorRef, shadowMode: Boolean = false): Props = {
+    Props(classOf[SingleWorkflowRunnerActor], source, metadataOutputFile, workflowManager, shadowMode)
   }
 
   sealed trait RunnerMessage
@@ -52,7 +53,8 @@ object SingleWorkflowRunnerActor {
  */
 case class SingleWorkflowRunnerActor(source: WorkflowSourceFiles,
                                      metadataOutputPath: Option[Path],
-                                     workflowManager: ActorRef) extends LoggingFSM[RunnerState, RunnerData] with CromwellActor {
+                                     workflowManager: ActorRef,
+                                     shadowMode: Boolean = false) extends LoggingFSM[RunnerState, RunnerData] with CromwellActor {
 
   import SingleWorkflowRunnerActor._
 
@@ -73,7 +75,8 @@ case class SingleWorkflowRunnerActor(source: WorkflowSourceFiles,
   when (NotStarted) {
     case Event(RunWorkflow, data) =>
       log.info(s"$tag: launching workflow")
-      workflowManager ! SubmitWorkflow(source)
+      val submitMessage = if (shadowMode) SubmitWorkflowCommand(source) else SubmitWorkflow(source)
+      workflowManager ! submitMessage
       goto (RunningWorkflow) using data.copy(replyTo = Option(sender()))
   }
 
--- a/engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
@@ -2,15 +2,20 @@ package cromwell.engine.workflow.lifecycle
 
 import akka.actor.{FSM, LoggingFSM, Props}
 import cromwell.backend.BackendJobExecutionActor.{BackendJobExecutionFailedResponse, BackendJobExecutionSucceededResponse, ExecuteJobCommand}
-import cromwell.backend.{BackendConfigurationDescriptor, BackendJobDescriptor, BackendJobDescriptorKey, BackendLifecycleActorFactory}
-import cromwell.core.WorkflowId
-import cromwell.engine.EngineWorkflowDescriptor
+import cromwell.backend.{BackendConfigurationDescriptor, BackendJobDescriptor, BackendJobDescriptorKey, BackendLifecycleActorFactory, JobKey}
+import cromwell.core.{WorkflowId, _}
+import cromwell.engine.ExecutionStatus._
+import cromwell.engine.{EngineWorkflowDescriptor, ExecutionStatus, _}
 import cromwell.engine.backend.{BackendConfiguration, CromwellBackend}
 import cromwell.engine.workflow.lifecycle.WorkflowExecutionActor._
+import cromwell.webservice.WdlValueJsonFormatter
+import lenthall.exception.ThrowableAggregation
 import wdl4s._
 import wdl4s.values.WdlValue
 
-import scala.util.Success
+import scala.annotation.tailrec
+import scala.language.postfixOps
+import scala.util.{Failure, Success, Try}
 
 object WorkflowExecutionActor {
 
@@ -26,10 +31,49 @@ object WorkflowExecutionActor {
   case object WorkflowExecutionFailedState extends WorkflowExecutionActorTerminalState
   case object WorkflowExecutionAbortedState extends WorkflowExecutionActorTerminalState
 
+  case class SymbolCacheKey(scopeName: String, input: Boolean)
+
+  type SymbolCache = Map[SymbolCacheKey, Traversable[SymbolStoreEntry]]
+  type ExecutionStore = Map[JobKey, ExecutionStatus]
+  type ExecutionStoreEntry = (JobKey, ExecutionStatus)
+
   /**
     * State data
     */
-  final case class WorkflowExecutionActorData()
+  final case class WorkflowExecutionActorData(executionStore: ExecutionStore,
+                                              symbolCache: SymbolCache) {
+
+    /** This method updates: The ExecutionStore with the updated status, the symbol cache with the new outputs, and appends to the overall CallOutputs */
+    def updateJob(jobKey: JobKey,
+                  outputs: CallOutputs,
+                  status: ExecutionStatus): WorkflowExecutionActorData = {
+      this.copy(executionStore = executionStore + (jobKey -> status),
+        symbolCache = symbolCache ++ updateSymbolStoreEntry(jobKey, outputs))
+    }
+
+    /** Add the outputs for the specified `JobKey` to the symbol cache. */
+    private def updateSymbolStoreEntry(jobKey: JobKey, outputs: CallOutputs) = {
+      val newEntriesMap = outputs map { case (lqn, value) =>
+        val storeKey = SymbolStoreKey(jobKey.scope.fullyQualifiedName, lqn, jobKey.index, input = false)
+        // TODO: SymbolStoreEntry should no longer contain the symbol hashes
+        new SymbolStoreEntry(storeKey, value.wdlValue.wdlType, Option(value.wdlValue), None)
+      } groupBy { entry => SymbolCacheKey(entry.scope, entry.isInput) }
+
+      newEntriesMap map { case (key, entries) =>
+        // SymbolCache is essentially a MultiMap, but that's a trait only for mutable Maps.
+        key -> (symbolCache.getOrElse(key, Seq.empty) ++ entries)
+      }
+    }
+
+    /** Checks if the workflow is completed by scanning through the executionStore */
+    def isWorkflowComplete: Boolean = {
+      def isDone(executionStatus: ExecutionStatus): Boolean = executionStatus == ExecutionStatus.Done
+      executionStore.values.forall(isDone)
+    }
+
+    /** Updates the status of a a job by the new entry */
+    def updateExecutionStoreStatus(entry: (JobKey, ExecutionStatus)): WorkflowExecutionActorData = this.copy(executionStore = executionStore + entry)
+  }
 
   /**
     * Commands
@@ -47,13 +91,27 @@ object WorkflowExecutionActor {
   case object WorkflowExecutionAbortedResponse extends WorkflowExecutionActorResponse
   final case class WorkflowExecutionFailedResponse(reasons: Seq[Throwable]) extends WorkflowExecutionActorResponse
 
+  case class WorkflowExecutionException(override val throwables: List[Throwable]) extends ThrowableAggregation {
+    override val exceptionContext = s"WorkflowExecutionActor"
+  }
+
   def props(workflowId: WorkflowId, workflowDescriptor: EngineWorkflowDescriptor): Props = Props(WorkflowExecutionActor(workflowId, workflowDescriptor))
 }
 
 final case class WorkflowExecutionActor(workflowId: WorkflowId, workflowDescriptor: EngineWorkflowDescriptor) extends LoggingFSM[WorkflowExecutionActorState, WorkflowExecutionActorData] {
 
+  import WorkflowExecutionActor._
+
   val tag = self.path.name
-  startWith(WorkflowExecutionPendingState, WorkflowExecutionActorData())
+
+  private val calls = workflowDescriptor.backendDescriptor.workflowNamespace.workflow.calls
+
+  // Initialize the StateData with ExecutionStore (all calls as NotStarted) and SymbolStore
+  startWith(
+    WorkflowExecutionPendingState,
+    WorkflowExecutionActorData(
+      executionStore = (calls map (BackendJobDescriptorKey(_, None, 1) -> NotStarted)) toMap,
+      symbolCache = buildSymbolStoreEntries.groupBy(entry => SymbolCacheKey(entry.scope, entry.isInput))))
 
   /** PBE: the return value of WorkflowExecutionActorState is just temporary.
     *      This should probably return a Try[BackendJobDescriptor], Unit, Boolean,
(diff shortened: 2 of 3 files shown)


gauravs90:
@cjllanwarne  and @Horneth.

engine/src/main/scala/cromwell/engine/workflow/SingleWorkflowRunnerActor.scala
```diff
@@ -8,6 +8,7 @@ import better.files._
 import cromwell.core.{CallOutput, WorkflowId}
```
gauravs90: I'll remove the changes in this file. It just adds the submission capability of a workflow to the Shadow world via command line, forgot to remove it, was just for my debugging purposes.

engine/src/main/scala/cromwell/engine/backend/dummy/DummyInitializationActor.scala
```diff
@@ -0,0 +1,50 @@
...
+
+// The sequence of calls that this receives are the ones this backend is expected to run, and it should validate only those
+class DummyInitializationActor(override val workflowDescriptor: BackendWorkflowDescriptor,
+                               override val calls: Seq[Call],
+                               override val configurationDescriptor: BackendConfigurationDescriptor) extends  BackendWorkflowInitializationActor{
+
+  private val tag = s"[${self.path.name}]"
+
+  override def validateRuntimeAttributes(runtimeAttributes: EvaluatedRuntimeAttributes): Future[scalaz.Validation[NonEmptyList[String], Unit]] =
+  Future.successful(scalaz.Success(Unit))
```
cjllanwarne: If this isn't validating anything, we don't need a `DummyInitializationActor`. `InitializationActor`s should be optional

engine/src/main/scala/cromwell/engine/workflow/SingleWorkflowRunnerActor.scala
cjllanwarne: 😐 - Eh, it's not hurting anyone...

Horneth:
@gauravs90 I haven't forgotten this, I'm just deep in the engine function / call inputs stuff but I'll look into it today :)

gauravs90:
@Horneth No problemo.

engine/src/main/scala/cromwell/engine/backend/dummy/DummyInitializationActor.scala
gauravs90: I'll remove this file. (Think I added it to validate the flow, and have it move to the tests, but did't do it 😓 )

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -51,26 +81,45 @@ object WorkflowExecutionActor {
...
+      call.prerequisiteCalls foreach (callGraph.addEdge(call, _))
+    }
+    callGraph
+  }
+
+  private def startJob(call: Call) = {
     // TODO: Support indexes and retries:
+    log.info(s"Starting Call: [${call.fullyQualifiedName}]")
     val jobKey = BackendJobDescriptorKey(call, None, 1)
     val jobDescriptor: BackendJobDescriptor = BackendJobDescriptor(workflowDescriptor.backendDescriptor, jobKey, Map.empty)
```
gauravs90: This `Map.empty` was for a symbol map..which currently I'm not sure how to populate or how it should look like 😕

Horneth: So I've changed how this works in this PR https://github.com/broadinstitute/cromwell/pull/748
If I understand you keep track of calls outputs in a map in the state data. All you'd have to do is to convert this map to a `Map[LocallyQualifiedName, WdlValue]` (which should be trivial), and adding it to the `symbolsFor` method in my PR.

cjllanwarne:
@gauravs90 a couple of global comments:

- The symbol store and execution store in the old WorkflowActor were not necessarily database-backed. They just stored which calls were completed and which were in flight.
- The graph is a nice idea but currently isn't as eager as it could be. Consider:
```
A -> B -> C
X -> Y -> Z
```
-- I believe this will run this in pairs, `(A,X)`, `(B,Y)`, `(C,Z)`. But what if A and B are really quick but X and Y are really slow - we're slowed down from implementing C because the unrelated tasks X and Y haven't started yet.
- I think I would prefer the existing method of determining (after every job completes) the set of jobs which have now become runnable. There's already an implicit DAG there. 
-- A major reason is, it has already been shown to work with the scatter/gather but I can't see how that would work with this rather static graph approach?

cjllanwarne:
Sorry all for the repeated almost-identical edits to the above comment. Markdown is hard... :(

cjllanwarne:
@gauravs90 @geoffjentry 

I think there's a couple of possible design choices for this executer and I'm afraid of starting off a big design discussion in a PR... Would it be better to have a skype/hangouts discussion of the various options? Maybe after the tasks discussion tomorrow?

gauravs90:
@cjllanwarne I certainly do not want to reinvent the wheel either. I was trying to bring over those stuff from old WA, but had some trouble along the way, and then decided to proceeded with a proof of concept with this approach since it felt like the least LoC and least ETA. 😓 (Also with this, @francares envisioned that the nodes of this graph can eventually be Actor'ified and allowed to handle everything autonomously)
I can give the old approach another shot and ask you at the points that I get stuck  at.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
Horneth: So actually no, I tried to do it and it's much more annoying than that

geoffjentry:
@gauravs90 I'm :-1: on the actual graph implementation. If we were to stick with this scheme I'd want something much different.

In terms of the scheme, I'm ok with the general concept (it's not unlike what I wanted to do originally as it turns out) but I question the utility of making this change. The old findRunnableCalls scheme was relatively simple and did the job - changing that seems out of scope for this ticket.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -25,9 +25,39 @@ object WorkflowExecutionActor {
...
   /**
-    * State data
+    * State Data
+    * @param callQueue The ordered (and grouped) List of Lists with safe ordering among them
+    * @param callOutputs A Map of Call -> Outputs
+    * @param inFlight Set of calls that are currently running
     */
-  final case class WorkflowExecutionActorData()
+  final case class WorkflowExecutionActorData(private var callQueue: List[List[Scope]],
+                                              private var callOutputs: Map[Call, cromwell.core.CallOutputs],
```
Horneth: Regardless of the outcome of this PR, I like this. It doesn't need to be a var I think, but having call outputs in the state data is a good idea. We can pass it around as a read-only thing whenever needed to create new jobs for example, but it should only need to be updated when a call finishes.

cjllanwarne: I totally agree. And what used to be the executionStore (or the execution graph, or whatever) should be in the stateData too.

Hooray for reducing the amount of hidden mutable state :D

gauravs90: @geoffjentry I'm fine with that too, had just put it out there for opinions :)
Been working on pulling the old stuff out.

gauravs90: @cjllanwarne @Horneth Yes..I've been trying to push that into the state data itself 😄

gauravs90:
@Horneth @cjllanwarne Have brought in the relevant (only) code from the WorkflowActor, with some improvements related to mutable state. Good to review again from my side.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -20,14 +28,58 @@ object WorkflowExecutionActor {
 
   case object WorkflowExecutionPendingState extends WorkflowExecutionActorState
   case object WorkflowExecutionInProgressState extends WorkflowExecutionActorState
-  case object WorkflowExecutionSuccessfulState extends WorkflowExecutionActorTerminalState
-  case object WorkflowExecutionFailedState extends WorkflowExecutionActorTerminalState
+  case object WorkflowExecutionSuccessfulState extends WorkflowExecutionActorTerminalState{
+    override val terminal = true
```
cjllanwarne: You don't need these, they already extend `WorkflowExecutionActorTerminalState`

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -20,14 +28,58 @@ object WorkflowExecutionActor {
...
   /**
     * State data
     */
-  final case class WorkflowExecutionActorData()
+  final case class WorkflowExecutionActorData(executionStore: ExecutionStore,
+                                              symbolCache: SymbolCache,
+                                              callOutputs: Map[JobKey, CallOutputs] = Map.empty) {
+
+    /** This method updates: The ExecutionStore with the updated status, the symbol cache with the new outputs, and appends to the overall CallOutputs */
+    def handleCompletedCall(jobKey: JobKey,
```
cjllanwarne: To be pedantic, should be `Job` rather than `Call`.

I also wasn't too thrilled with `handle` since this isn't really a handler...

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -20,14 +28,58 @@ object WorkflowExecutionActor {
...
+                            outputs: CallOutputs,
+                            status: ExecutionStatus): WorkflowExecutionActorData = {
+      this.copy(executionStore = executionStore + (jobKey -> status),
+        symbolCache = symbolCache ++ updatedSymbolStoreEntry(jobKey, outputs),
+        callOutputs = callOutputs + (jobKey -> outputs))
+    }
+
+    /** Add the outputs for the specified `BackendJobDescriptorKey` to the symbol cache. */
```
cjllanwarne: for the specified `JobKey` now

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -20,14 +28,58 @@ object WorkflowExecutionActor {
...
+    private def updatedSymbolStoreEntry(jobKey: JobKey, outputs: CallOutputs) = {
+      val newEntriesMap = outputs map { case (lqn, value) =>
+        val storeKey = SymbolStoreKey(jobKey.scope.fullyQualifiedName, lqn, jobKey.index, input = false)
+        new SymbolStoreEntry(storeKey, value.wdlValue.wdlType, Option(value.wdlValue), value.hash)
```
cjllanwarne: Maybe out of scope in this ticket, but we no longer want hashes coming back with symbols

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -20,14 +28,58 @@ object WorkflowExecutionActor {
...
+                                              symbolCache: SymbolCache,
+                                              callOutputs: Map[JobKey, CallOutputs] = Map.empty) {
+
+    /** This method updates: The ExecutionStore with the updated status, the symbol cache with the new outputs, and appends to the overall CallOutputs */
+    def handleCompletedCall(jobKey: JobKey,
+                            outputs: CallOutputs,
+                            status: ExecutionStatus): WorkflowExecutionActorData = {
+      this.copy(executionStore = executionStore + (jobKey -> status),
+        symbolCache = symbolCache ++ updatedSymbolStoreEntry(jobKey, outputs),
+        callOutputs = callOutputs + (jobKey -> outputs))
```
Horneth: I don't think it's necessary to store the outputs in the symbolCache and also the outputs map. If we want to split them up then we should have an inputMap and an outputMap, or if we don't just put everything in the symbol cache

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -45,32 +97,56 @@ object WorkflowExecutionActor {
   case object WorkflowExecutionAbortedResponse extends WorkflowExecutionActorResponse
   final case class WorkflowExecutionFailedResponse(reasons: Seq[Throwable]) extends WorkflowExecutionActorResponse
 
+  /**
+    * Captures the `ExecutionStoreKey`/`ExecutionStatus` combination for a call that has been created or started
```
cjllanwarne: I think this comment needs an update. 
1. `JobKey / ExecutionStatus`
2. We can't only be using these for persistence if there isn't persistence in the actor yet?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -45,32 +97,56 @@ object WorkflowExecutionActor {
...
   val tag = self.path.name
-  startWith(WorkflowExecutionPendingState, WorkflowExecutionActorData())
 
-  def startJob(call: Call) = {
+  private val calls = workflowDescriptor.backendDescriptor.workflowNamespace.workflow.calls
+
+  // Initialize the StateData with ExecutionStore (all calls as NotStarted) and SymbolStore
+  startWith(WorkflowExecutionPendingState,
+    WorkflowExecutionActorData(executionStore = (calls map (BackendJobDescriptorKey(_, None, 1) -> NotStarted)) toMap,
+      symbolCache = buildSymbolStoreEntries.groupBy(entry => SymbolCacheKey(entry.scope, entry.isInput))))
```
cjllanwarne: Slightly tough to follow the tabulation. I prefer:
```
startWith(
  WorkflowExecutionPendingState,
  WorkflowExecutionActorData(
    executionStore = (calls map (BackendJobDescriptorKey(_, None, 1) -> NotStarted)) toMap,
    symbolCache = buildSymbolStoreEntries.groupBy(entry => SymbolCacheKey(entry.scope, entry.isInput))))
```

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -45,32 +97,56 @@ object WorkflowExecutionActor {
...
+
+  def startJob(backendJobDescriptorKey: BackendJobDescriptorKey, stateData: WorkflowExecutionActorData): Try[ExecutionStartResult] = {
```
cjllanwarne: `private def`

Also, stateData could just be `symbolCache` since that's the only part of it that's actually used

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -45,32 +97,56 @@ object WorkflowExecutionActor {
...
     // TODO: Support indexes and retries:
-    val jobKey = BackendJobDescriptorKey(call, None, 1)
-    val jobDescriptor: BackendJobDescriptor = BackendJobDescriptor(workflowDescriptor.backendDescriptor, jobKey, Map.empty)
-    val executionActor = backendForExecution(jobDescriptor, workflowDescriptor.backendAssignments(call))
-    executionActor ! ExecuteJobCommand
+    val call = backendJobDescriptorKey.call
+    fetchLocallyQualifiedInputs(backendJobDescriptorKey, stateData.symbolCache) match {
+      case Success(symbolStoreForCall) =>
+        val jobDescriptor: BackendJobDescriptor = BackendJobDescriptor(workflowDescriptor.backendDescriptor, backendJobDescriptorKey, symbolStoreForCall)
+        val executionActor = backendForExecution(jobDescriptor, workflowDescriptor.backendAssignments(call))
```
cjllanwarne: You might want to coordinate with @scottfrazer on how to wire up the actor factory properly (I believe his PR is about to merge)

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -82,18 +158,35 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId, workflowDescript
   when(WorkflowExecutionInProgressState) {
     case Event(BackendJobExecutionSucceededResponse(jobKey, callOutputs), stateData) =>
       log.info(s"Job ${jobKey.call.fullyQualifiedName} succeeded! Outputs: ${callOutputs.mkString("\n")}")
-      context.parent ! WorkflowExecutionSucceededResponse
-      goto(WorkflowExecutionSuccessfulState)
+      val newData = stateData.handleCompletedCall(jobKey, callOutputs, ExecutionStatus.Done) // addPersisting(jobKey, ExecutionStatus.Done)
```
cjllanwarne: What's going on with the comment here?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -82,18 +158,35 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId, workflowDescript
...
+      if (newData.isWorkflowComplete) {
+        // Printing the final outputs, temporarily here until SingleWorkflowManagerActor is made in-sync with the shadow mode
```
cjllanwarne: Could you at least move it out into a separate `generateAndPrintWorkflowOutputs` method?

It's just I find all this stuff is getting in the way of parsing the state machine.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -118,8 +211,131 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId, workflowDescript
...
+        /**
+          * Otherwise, simply refer to the entry the collector entry.  This means that 'entry' depends
+          * on every shard of the pre-requisite scope to finish.
+          */
+        case _ =>
+          executionStore filter { case (k, _) => k.scope == prerequisiteScope && k.index.isEmpty } toSeq
+      }
+    }
+
+    def isDone(e: JobKey, status: ExecutionStatus): Boolean =
```
cjllanwarne: Duplicates a method already in `stateData`

cjllanwarne: `status: ExecutionStatus` is never used

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -118,8 +211,131 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId, workflowDescript
...
+
+      /**
+        * We need to make sure that all prerequisiteScopes have been resolved to some entry before going forward.
+        * If a scope cannot be resolved it may be because it is in a scatter that has not been populated yet,
+        * therefore there is no entry in the executionStore for this scope.
+        * If that's the case this prerequisiteScope has not been run yet, hence the (upstream forall {_.nonEmpty})
+        */
+      (upstream forall {
+        _.nonEmpty
+      }) && dependenciesResolved
```
cjllanwarne: Just thinking out loud: I don't know the origin of all this but it seems kinda backwards. It all seems to come about because val dependenciesResolved = executionStore.filter(dependencies) forall (x => isDone(x._1, x._2))` is true even if the executionStore doesn't contain a dependency.

But anyway, maybe for this non-scatter ticket can we just remove this and whoever follows with the scatter work can re-introduce it if necessary?

cjllanwarne: In other words, I'm suggestion:
```
def arePrerequisitesDone(key: JobKey): Boolean = {
    val upstream = key.scope.prerequisiteScopes.map(s => upstreamEntries(key, s))
    val dependencies = upstream.flatten
    executionStore.filter(dependencies) forall (x => isDone(x._1, x._2))
}
```

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -118,8 +211,131 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId, workflowDescript
...
+
+    inputSymbols.toSet ++ callSymbols.toSet
+  }
+
+  /**
+    * Attempt to start all runnable calls and return updated state data.  This will create a new copy
+    * of the state data including new pending persists.
+    */
+  @tailrec
+  private def startRunnableCalls(data: WorkflowExecutionActorData): WorkflowExecutionActorData = {
```
cjllanwarne: Personally I really dislike having a function that's almost 1/4 of the entire file, and which contains a whole bunch of sub-functions.

Would it be possible to extract these sub-functions into separate methods so that the flow of `startRunnableCalls` is more obvious?

cjllanwarne:
@gauravs90 this is really exciting. I suggest you rebase on top of @scottfrazer's #766 (or develop when he merges it) which means we'll have all the pieces to start testing end-to-end workflows!!

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
gauravs90: I'll change it to `None` and add a note here.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
gauravs90: It is being used in the `startRunnableCalls` but I guess and I can do without it just as well. I'll remove this.

gauravs90:
@cjllanwarne @Horneth Addressed the comments. What do you guys want me to do now? Wait and rebase with Scott's PR or just merge into develop?

Horneth:
@scottfrazer is currently rebasing on top of @cjllanwarne's abort merge, if you merge now he'll need to rebase again on top of this. Maybe we can wait to get the factory mergable/merged to rebase this one ?

gauravs90:
@Horneth Great.

gauravs90:
@cjllanwarne and @Horneth I have rebased with the latest develop. Awaiting you guys approvals now. 
Also for some reason the tests have started failing locally for mw (files which are irrelevant for this PR). I checked out the latest develop separately and it also seems to fail there. Still looking into it but not sure how to proceed with that.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -62,57 +120,58 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId, workflowDescript
...
-    )
-    jobExecutionActor ! ExecuteJobCommand
-    WorkflowExecutionInProgressState
-  }
-
-  private def startJob(jobKey: BackendJobDescriptorKey, inputs: Map[FullyQualifiedName, WdlValue]): WorkflowExecutionActorState = {
+                       factory: BackendLifecycleActorFactory): Try[ExecutionStatus] = {
+
+      val call = jobKey.call
+      fetchLocallyQualifiedInputs(jobKey, symbolCache) match {
```
gauravs90: @Horneth Question: Do we need to use this function `fetchLocallyQualifiedInputs` or will it work if we just convert the entries in the SymbolCache and pass it as a Map[FQZ -> WdlValue] to the backend?

Horneth: That's an open discussion. There are issues with both that we need to figure out. Just passing the whole symbol cache to the backend puts a lot of work on the backend side, all of which I'm not even sure is doable in the context of a Job only. On the other side fetching the outputs (even unveluated) for a call and giving only that to the backend makes it harder in terms of lookup if expressions reference variables in different scopes. 
I'd leave it as is for now and we can have a discussion about it later.

cjllanwarne:
👍

Horneth:
👍

PR #746, opened 2016-04-27
Fmc 725 validation exp eval. Closes #725.

### LINKED ISSUE #725: Backend validation expression evaluation exceptions
Class: BackendWorkflowInitializationActor
Line: 115
Function: evaluateRuntimeAttributesWdlExpressions
Code: TryUtils.sequenceMap(evaluateAttrs, "Runtime attributes evaluation").get

This line is producing an exception when it cannot evaluate an expression in runtime attributes.
There are open questions to work on:
Should this catch exception and log or just fail here?
If the case it catch the exception and log, should this return 'InitializationSuccess' after that? or just fail and return a InitializationFailed?

Use case: task1 output is the input for task2 and that value is used within RuntimeAttributes."

One this is solve, enable ignored test case.

francares:
Main purpose of this PR is to validate the idea of "Runtime Attributes validation" at Backend **initialization** an **execution** level.

[x] LocalBackend
- Initialization => No runtime attribute keys validation. Just a warning for those attributes that are not supported.
- Execution => Validate values of supported runtime attributes.

[x] JesBackend
- Initialization => Validate Docker key is present and warn for those attributes that are not supported.
- Execution => Validate values of supported runtime attributes.

[x] HtCondorBackend
- Initialization => No runtime attribute keys validation. Just a warning for those attributes that are not supported.
- Execution => Validate values of supported runtime attributes.

[x] SgeBackend
- Initialization => No runtime attribute keys validation. Just a warning for those attributes that are not supported.
- Execution => Validate values of supported runtime attributes.

The idea is to have this public while I implement this functionality so we all are in the same page.

### DIFF
--- a/supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorInitializationActor.scala
+++ b/supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorInitializationActor.scala
@@ -14,8 +14,7 @@ import scala.concurrent.Future
 import scalaz.Scalaz._
 
 object HtCondorInitializationActor {
-  val FailOnStderrDefaultValue = true
-  val ContinueOnRcDefaultValue = 0
+  val SupportedKeys = Set(Docker, FailOnStderr, ContinueOnReturnCode)
 
   def props(workflowDescriptor: BackendWorkflowDescriptor, calls: Seq[Call], configurationDescriptor: BackendConfigurationDescriptor): Props =
     Props(new HtCondorInitializationActor(workflowDescriptor, calls, configurationDescriptor))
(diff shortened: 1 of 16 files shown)


francares:
Reviewers: @geoffjentry and @Horneth.

geoffjentry:
👍 

Once the dust settles w/ PBE I think we should go back and abstract common validation patterns which might exist across the backends into some place of optional utility functions for other backend implementers but that's not for now IMO

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorInitializationActor.scala
```diff
@@ -35,20 +32,21 @@ class HtCondorInitializationActor(override val workflowDescriptor: BackendWorkfl
...
     Future {
-      val docker = validateDocker(runtimeAttributes.get(Docker), "Failed to get Docker mandatory key from runtime attributes".failureNel)
-      val failOnStderr = validateFailOnStderr(runtimeAttributes.get(FailOnStderr), FailOnStderrDefaultValue.successNel)
-      val continueOnReturnCode = validateContinueOnReturnCode(runtimeAttributes.get(ContinueOnReturnCode),
-        ContinueOnReturnCodeSet(Set(ContinueOnRcDefaultValue)).successNel)
-      (docker |@| failOnStderr |@| continueOnReturnCode) {
-        (_, _, _)
+      calls foreach { call =>
+        val runtimeAttributes = call.task.runtimeAttributes.attrs
+        val supportedKeys = Set(Docker, FailOnStderr, ContinueOnReturnCode)
```
Horneth: The list of supported keys could be moved to the `object` instead of being re-instantiated every time

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorInitializationActor.scala
```diff
@@ -35,20 +32,21 @@ class HtCondorInitializationActor(override val workflowDescriptor: BackendWorkfl
...
+        val notSupportedAttributes = runtimeAttributes filterKeys {
+          !supportedKeys.contains(_)
+        }
+
+        if (notSupportedAttributes.size > 0) {
```
Horneth: `!notSupportedAttributes.isEmpty` seems a bit more idiomatic but not a big deal

Horneth:
A couple nitpicks but 👍

PR #748, opened 2016-04-27
Call inputs and evaluation

Horneth:


core/src/main/scala/cromwell/core/package.scala
```diff
@@ -1,17 +1,42 @@
...
 import scalaz._
 
 package object core {
   // root can be a Path instead of a String in PBE. stdout / err too but it doesn't really bring values since they're just stringified to WdlFiles
   class WorkflowContext(val root: String)
   class CallContext(override val root: String, val stdout: String, val stderr: String) extends WorkflowContext(root)
 
+  // Can evaluate a wdl value
+  class Evaluator(evaluator: WdlValue => Try[WdlValue]) {
+    def evaluate(wdlValue: WdlValue): Try[WdlValue] = evaluator(wdlValue)
```
francares: Maybe we can add evaluate as an extension method for WdlValue in WDL4S... What do you think about that?...

Horneth: Yeah that could be interesting, I'm trying to make everything work in Cromwell first but when we have that we could definitely move some of this code to WDL4s

Horneth:
@cjllanwarne @kshakir

Horneth:
I'm still working on this actually, as it is now it wouldn't work with JES so I'm cooking something else. It's close but still need a bit more refinement, I'll push the new version over the week-end.

PR #759, opened 2016-04-28
Abort wiring in shadow world. Closes #671

### LINKED ISSUE #671: Have Workflow Actor call Abort to Lifecycle Actors

cjllanwarne:
The idea being, sending an abort message does not get an `AbortSuccess` or `AbortFailed` response directly.

Instead, the actor being aborted will cause its operation to fail and the parent should expect an "Aborted" response instead of a Success or Failure (although a Success or Failure may still be returned anyway).

In this way, it's acceptable for a backend implementor to simply do nothing when `abort` is received. The parent will just continue waiting for a result of some sort.

### DIFF
--- a/backend/src/main/scala/cromwell/backend/BackendJobExecutionActor.scala
+++ b/backend/src/main/scala/cromwell/backend/BackendJobExecutionActor.scala
@@ -34,12 +34,11 @@ trait BackendJobExecutionActor extends BackendJobLifecycleActor with ActorLoggin
   def receive: Receive = LoggingReceive {
     case ExecuteJobCommand => performActionThenRespond(execute, onFailure = executionFailed)
     case RecoverJobCommand => performActionThenRespond(recover, onFailure = executionFailed)
-    case AbortJob          => performActionThenRespond(abortJob, onFailure = abortFailed)
+    case AbortJobCommand => abortJob
   }
 
   // We need this for receive because we can't do `onFailure = ExecutionFailure` directly - because BackendJobDescriptor =/= BackendJobDescriptorKey
   private def executionFailed = (t: Throwable) => BackendJobExecutionFailedResponse(jobDescriptor.key, t)
-  private def abortFailed = (t: Throwable) => BackendJobExecutionAbortFailedResponse(jobDescriptor.key, t)
 
   /**
     * Execute a new job.
@@ -54,5 +53,5 @@ trait BackendJobExecutionActor extends BackendJobLifecycleActor with ActorLoggin
   /**
     * Abort a running job.
     */
-  def abortJob: Future[JobAbortResponse]
+  def abortJob: Unit
 }
--- a/backend/src/main/scala/cromwell/backend/BackendLifecycleActor.scala
+++ b/backend/src/main/scala/cromwell/backend/BackendLifecycleActor.scala
@@ -15,21 +15,14 @@ object BackendLifecycleActor {
    * Commands
    */
   trait BackendWorkflowLifecycleActorCommand extends BackendWorkflowLifecycleActorMessage
-  case object AbortWorkflow extends BackendWorkflowLifecycleActorCommand
-  final case class AbortJob(jobKey: BackendJobDescriptorKey) extends BackendWorkflowLifecycleActorCommand
+  case object AbortWorkflowCommand extends BackendWorkflowLifecycleActorCommand
+  case object AbortJobCommand extends BackendWorkflowLifecycleActorCommand
 
   /*
    * Responses
    */
   trait BackendWorkflowLifecycleActorResponse extends BackendWorkflowLifecycleActorMessage
-
-  sealed trait JobAbortResponse extends BackendWorkflowLifecycleActorResponse
-  sealed trait WorkflowAbortResponse extends BackendWorkflowLifecycleActorResponse
-
-  case object BackendWorkflowAbortSucceededResponse extends WorkflowAbortResponse
-  final case class BackendWorkflowAbortFailedResponse(throwable: Throwable) extends WorkflowAbortResponse
-  final case class BackendJobExecutionAbortSucceededResponse(jobKey: BackendJobDescriptorKey) extends JobAbortResponse
-  final case class BackendJobExecutionAbortFailedResponse(jobKey: BackendJobDescriptorKey, throwable: Throwable) extends JobAbortResponse
+  case object BackendActorAbortedResponse extends BackendWorkflowLifecycleActorResponse
 }
 
 trait BackendLifecycleActor extends Actor {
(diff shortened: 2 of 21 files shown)


cjllanwarne:
The wheel has chosen @Horneth and @mcovarr. I hope the fundamental choice of how abort now works isn't too controversial (as I understand it, it's how things work currently) but I'm happy to do a meet 'n' discuss if you want.

backend/src/main/scala/cromwell/backend/BackendLifecycleActor.scala
```diff
@@ -15,21 +15,14 @@ object BackendLifecycleActor {
    * Commands
    */
   trait BackendWorkflowLifecycleActorCommand extends BackendWorkflowLifecycleActorMessage
-  case object AbortWorkflow extends BackendWorkflowLifecycleActorCommand
-  final case class AbortJob(jobKey: BackendJobDescriptorKey) extends BackendWorkflowLifecycleActorCommand
+  case object AbortWorkflowCommand extends BackendWorkflowLifecycleActorCommand
+  case object AbortJobCommand extends BackendWorkflowLifecycleActorCommand
```
mcovarr: nothing appears to send one of these

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
```diff
@@ -76,9 +83,16 @@ object ShadowWorkflowActor {
...
-  case object EmptyShadowWorkflowActorData extends ShadowWorkflowActorData
-  case class ShadowWorkflowActorDataWithDescriptor(workflowDescriptor: EngineWorkflowDescriptor) extends ShadowWorkflowActorData
+  /**
+    * @param currentLifecycleStateActor The current lifecycle stage, represented by an ActorRef.
+    */
+  case class ShadowWorkflowActorData(currentLifecycleStateActor: Option[ActorRef],
+                                     workflowDescriptor: Option[EngineWorkflowDescriptor])
+  object ShadowWorkflowActorData {
+    def empty = ShadowWorkflowActorData(None, None)
+    def apply(currentLifecycleStateActor: ActorRef): ShadowWorkflowActorData = ShadowWorkflowActorData(Some(currentLifecycleStateActor), None)
```
mcovarr: Option

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
```diff
@@ -76,9 +83,16 @@ object ShadowWorkflowActor {
...
+    def apply(currentLifecycleStateActor: ActorRef, workflowDescriptor: EngineWorkflowDescriptor): ShadowWorkflowActorData = ShadowWorkflowActorData(Some(currentLifecycleStateActor), Some(workflowDescriptor))
+  }
```
mcovarr: you could get rid of all this by using `.copy`

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowLifecycleActor.scala
```diff
@@ -9,38 +9,47 @@ object WorkflowLifecycleActor {
...
+                                         successes: Seq[ActorRef],
+                                         failures: Map[ActorRef, Throwable],
+                                         aborted: Seq[ActorRef]) {
+
+    def withBackendActors(backendActors: Map[ActorRef, String]) = this.copy(
+      backendActors = this.backendActors ++ backendActors
+    )
+    def withSuccess(successfulActor: ActorRef) = this.copy(
+      backendActors = this.backendActors - successfulActor,
+      successes = successes ++ List(successfulActor))
```
mcovarr: I think you can `:+` to avoid putting stuff in `List`s just to `++`

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowLifecycleActor.scala
```diff
@@ -82,18 +91,25 @@ trait WorkflowLifecycleActor[S <: WorkflowLifecycleActorState, D <: WorkflowLife
...
+        }
+        else {
+          context.parent ! failureResponse(newData.failures.values.toSeq)
+          goto(failureState) using newData
+        }
       }
     }
-    else {
+    else
+    {
```
mcovarr: join lines

supportedBackends/local/src/test/scala/cromwell/backend/impl/local/LocalJobExecutionActorSpec.scala
```diff
@@ -134,24 +133,21 @@ class LocalJobExecutionActorSpec extends FlatSpec with BackendTestkitSpec with M
...
     }
   }
 
-  it should "fail to abort if execution hasn't been requested" in {
+  it should "preemptively abort if execution hasn't been requested yet" in {
     val wf = buildWorkflowDescriptor(Sleep10)
     val jobDescriptor: BackendJobDescriptor = jobDescriptorFromSingleCallWorkflow(wf)
     val backend = localBackend(jobDescriptor, defaultBackendConfig)
 
     val abort = backend.abortJob
```
mcovarr: looks like this variable is unused

backend/src/main/scala/cromwell/backend/BackendLifecycleActor.scala
cjllanwarne: Yeah, I think that "`abort()` within the `WorkflowExecutionActor`" is going to need to be a new ticket following @gauravs90's scoping-out of execution.

cjllanwarne: Now #762

backend/src/main/scala/cromwell/backend/BackendJobExecutionActor.scala
```diff
@@ -54,5 +53,5 @@ trait BackendJobExecutionActor extends BackendJobLifecycleActor with ActorLoggin
   /**
     * Abort a running job.
     */
-  def abortJob: Future[JobAbortResponse]
+  def abortJob: Unit
```
Horneth: Couldn't it still be useful to have some kind of response from that actually ? Say you try to abort a JES job, but for some reason there's an error while requesting the abort to JES (the jib can't be found or whatever..) Then you'll never know about this error :/

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
cjllanwarne: This is just a shorthand to not have to type `Option(...)` every time.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowLifecycleActor.scala
cjllanwarne: Woohoo!

backend/src/main/scala/cromwell/backend/BackendJobExecutionActor.scala
cjllanwarne: @mcovarr @Horneth Haha, ok, somebody mentioned it! 

If you guys would like, let's tech talk this. The questions would be:
a) what would be returned
b) what would you actually do with the result

In the Ye Olde world, we never check for this or do anything special as a result so it would have to be a new feature.

Horneth: I was thinking it would be useful only if it fails. The engine could decide to try again, or just fail the call, or do nothing, but at least it would have the information.

mcovarr:
:+1:

Horneth:
👍

PR #766, opened 2016-04-29
BackendLifecycleActorFactory... go ahead and tear it apart

scottfrazer:
- [x] update README
- [x] let everybody know they need to update their configs
- [x] create ticket describing surprising `src` + `test` config overlay, not to be fixed as part of this work! (https://github.com/broadinstitute/cromwell/issues/768)


scottfrazer:
Wheel chose @kshakir and @cjllanwarne ... but maybe @mcovarr might want to take a look too?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/ShadowMaterializeWorkflowDescriptorActor.scala
```diff
@@ -185,24 +168,27 @@ class ShadowMaterializeWorkflowDescriptorActor() extends LoggingFSM[ShadowMateri
...
+        val backendPriorities = Seq(
+          workflowOptions.get(RuntimeBackendKey).toOption,
+          assignBackendUsingRuntimeAttrs(call),
+          Option(CromwellBackend.shadowDefaultBackend)
+        )
+
+        backendPriorities.flatten.headOption match {
+          case Some(backendName) if CromwellBackend.isValidBackendName(backendName) => call -> backendName
+          case Some(backendName) => throw new Exception(s"INVALID BACKEND: $backendName")
+          case None => throw new Exception(s"NO BACKEND FOR CALL $call")
```
mcovarr: lots of shouting 😄

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -53,19 +57,64 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId, workflowDescript
...
+            startJob(jobKey, inputs, configDescriptor, factory)
+          case (Failure(ex), Success(_)) =>
+            context.parent ! WorkflowExecutionFailedResponse(Seq(configurationDescriptorException(ex)))
+            WorkflowExecutionFailedState
+          case (Success(_), Failure(ex)) =>
+            context.parent ! WorkflowExecutionFailedResponse(Seq(backendLifecycleActorException(ex)))
+            WorkflowExecutionFailedState
+          case (Failure(ex1), Failure(ex2)) =>
+            context.parent ! WorkflowExecutionFailedResponse(Seq(configurationDescriptorException(ex1), backendLifecycleActorException(ex2)))
+            WorkflowExecutionFailedState
```
mcovarr: this feels like it could be DRYed but I don't have a concrete suggestion... 😞

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowFinalizationActor.scala
```diff
@@ -56,17 +56,23 @@ case class WorkflowFinalizationActor(workflowId: WorkflowId, workflowDescriptor:
...
   when(FinalizationPendingState) {
     case Event(StartFinalizationCommand, _) =>
-      val backendFinalizationActors = backendWorkflowActors(workflowDescriptor.backendAssignments)
-      if (backendFinalizationActors.isEmpty) {
-        sender ! WorkflowFinalizationSucceededResponse // Nothing more to do
-        goto(FinalizationSucceededState)
-      } else {
-        backendFinalizationActors.keys foreach { _ ! Finalize }
-        goto(FinalizationInProgressState) using WorkflowFinalizationActorData(backendFinalizationActors, List.empty, Map.empty)
+      val finalizationActors = for {
```
mcovarr: I think the original variable name was more clear

engine/src/main/scala/cromwell/engine/workflow/lifecycle/ShadowMaterializeWorkflowDescriptorActor.scala
scottfrazer: oops

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowInitializationActor.scala
```diff
@@ -65,13 +68,25 @@ case class WorkflowInitializationActor(workflowId: WorkflowId, workflowDescripto
...
+      initializationActors match {
+        case Failure(ex) =>
+          sender ! WorkflowInitializationFailedResponse(Seq(ex))
+          goto(InitializationFailedState)
+        case Success(Nil) =>
+          sender ! WorkflowInitializationSucceededResponse
+          goto(InitializationSucceededState)
+        case Success(actors) =>
+          actors.foreach(_ ! Initialize)
+          goto(InitializingWorkflowState) using WorkflowInitializationActorData(actors.toSet, List.empty, Map.empty)
```
mcovarr: `using stateData.copy(backendActors = actors.toSet)`

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowLifecycleActor.scala
```diff
@@ -57,23 +61,21 @@ trait WorkflowLifecycleActor[S <: WorkflowLifecycleActorState, D <: WorkflowLife
...
 
-  /**
-    * Turns a mapping from Call => backendName into a mapping from backendName => List[Call]
-    */
-  protected def callAssignments(backendAssignments: Map[Call, String]): Map[String, List[Call]] =
-    backendAssignments.groupBy(_._2).mapValues(_.keys.toList)
+  protected def finalizationActorProps(backendName: String): Option[Props] = {
+    CromwellBackend.shadowBackendLifecycleFactory(backendName) match {
+      case Success(factory) => factory.workflowFinalizationActorProps()
+      case Failure(ex) => throw ex // TODO: what do we do?
```
scottfrazer: I forgot to finish this...

engine/src/test/resources/application.conf
```diff
@@ -1,44 +1,141 @@
...
+  // Docker image that was used to run the call, but at a cost of having to make a
+  // request to an external service (DockerHub, GCR).  If a call fails to lookup a
+  // Docker hash, it will fail.
+  lookup-docker-hash = false
+}
+
+system {
+  // If 'true', a SIGINT will trigger Cromwell to attempt to abort all currently running jobs before exiting
+  abort-jobs-on-terminate = false
 }
```
mcovarr: there shouldn't be two system blocks

engine/src/main/scala/cromwell/engine/backend/BackendConfiguration.scala
```diff
@@ -1,25 +1,48 @@
...
-case class BackendConfigurationEntry(name: String, className: String, config: Config)
+case class BackendConfigurationEntry(name: String, className: String, shadowLifecycleActorFactoryClass: String, config: Config) {
+  def asBackendLifecycleActorFactory: BackendLifecycleActorFactory = {
+    Class.forName(shadowLifecycleActorFactoryClass)
+         .getConstructor(classOf[Config])
+         .newInstance(config)
+         .asInstanceOf[BackendLifecycleActorFactory]
+  }
+
+  val asBackendConfigurationDescriptor = BackendConfigurationDescriptor(config, CromwellConfiguration.config)
```
mcovarr: is there a reason one is a val and one is a def?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
scottfrazer: I hear ya... I figured that since this has to be refactored anyway (see comment above about the return value), somebody can figure out how to be more clever.

engine/src/main/scala/cromwell/engine/backend/BackendConfiguration.scala
```diff
@@ -1,25 +1,48 @@
...
-  private val BackendConfig = ConfigFactory.load().getConfig("backend")
+  private val BackendConfig = CromwellConfiguration.config.getConfig("backend")
   private val DefaultBackendName = BackendConfig.getString("default")
-  private val BackendProvidersConfigs = BackendConfig.getConfigList("providers").asScala.toList
+  private val BackendProviders = BackendConfig.getConfig("providers")
+  private val BackendNames: Set[String] = BackendProviders.entrySet().asScala.map(_.getKey.split("\\.").toSeq.head).toSet
+
+  val AllBackendEntries: List[BackendConfigurationEntry] = BackendNames.toList map { backendName =>
+    val entry = BackendProviders.getConfig(backendName)
+    // TODO PBE: "class" can be removed when switching to the shadow world; "actor-factor" renamed?
```
mcovarr: Digging the "actor-factor" 😄

engine/src/main/scala/cromwell/engine/backend/BackendConfiguration.scala
```diff
@@ -1,25 +1,48 @@
...
-  val AllBackendEntries: List[BackendConfigurationEntry] = BackendProvidersConfigs map { entry =>
-    BackendConfigurationEntry(entry.getString("name"), entry.getString("class"), entry.getConfig("config")) }
+  val DefaultBackendEntry: BackendConfigurationEntry = AllBackendEntries.find(_.name == DefaultBackendName) getOrElse {
+    throw new IllegalArgumentException(s"Could not find specified default backend name '$DefaultBackendName'.")
+  }
 
-  val DefaultBackendEntry: BackendConfigurationEntry =
-    AllBackendEntries find { _.name == DefaultBackendName } getOrElse { throw new IllegalArgumentException(s"Could not find specified default backend name '$DefaultBackendName'.") }
+  def backendConfigurationDescriptor(backendName: String): Try[BackendConfigurationDescriptor] = {
+    AllBackendEntries.find(_.name == backendName).map(_.asBackendConfigurationDescriptor) match {
```
mcovarr: find + map could be collect, maybe getOrElse this wrapped in a Try?

engine/src/main/scala/cromwell/engine/backend/BackendConfiguration.scala
scottfrazer: it's left-over from The Great Debugging Fiasco of 4/28

engine/src/main/scala/cromwell/engine/backend/BackendConfiguration.scala
scottfrazer: for real or sarcastic?  I don't mind it I guess, I didn't have it in me to think of a better name at the time.

engine/src/main/scala/cromwell/engine/backend/CromwellBackend.scala
```diff
@@ -16,24 +18,52 @@ import scala.util.Try
...
+  }
+
+  def shadowDefaultBackend = _shadowDefaultBackendFactory
+
+  def isValidBackendName(name: String): Boolean = _shadowBackendFactories.exists(_.contains(name))
+
+  private def backend(backendName: String): Backend = _backends map { _(backendName) } getOrElse { throw new IllegalStateException("backend() called prior to initBackends") }
 
-  def tryBackend(backendName: String) = Try(backend(backendName))
+  private def tryBackend(backendName: String) = Try(backend(backendName))
```
mcovarr: maybe can just inline this

engine/src/main/scala/cromwell/engine/backend/BackendConfiguration.scala
scottfrazer: oh lol, misspelled.. heh

I'm way too tired

mcovarr:
:+1: apart from a few nits and the parts you pointed out as unfinished.  I added a couple of quick items to your TODO list for you too.  😄

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -53,19 +57,64 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId, workflowDescript
...
+      ),
+      s"$workflowId-BackendExecutionActor-${jobDescriptor.key.tag}"
+    )
+    jobExecutionActor ! ExecuteJobCommand
+    WorkflowExecutionInProgressState
+  }
+
+  private def startJob(jobKey: BackendJobDescriptorKey, inputs: Map[FullyQualifiedName, WdlValue]): WorkflowExecutionActorState = {
+    workflowDescriptor.backendAssignments.get(jobKey.call) match {
+      case None =>
```
cjllanwarne: I assume this is a case which we expect to never reach?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
cjllanwarne: It's a little ugly but:
```
        (attemptedConfigurationDescriptor, attemptedActorFactory) match {
          case (Success(configDescriptor), Success(factory)) =>
            startJob(jobKey, inputs, configDescriptor, factory)
          case (_, _) =>
            val errors = List(
              attemptedActorFactory.failed map { new Exception(s"Could not get BackendConfigurationDescriptor for backend $backendName", _) } toOption,
              attemptedActorFactory.failed map { new Exception(s"Could not get BackendLifecycleActor for backend $backendName", _) } toOption
            ) collect { case Some(ex) => ex }
            context.parent ! WorkflowExecutionFailedResponse(errors)
            WorkflowExecutionFailedState
        }
```

cjllanwarne:
Looks good to me

cjllanwarne:
👍

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
```diff
@@ -96,7 +96,7 @@ object ShadowWorkflowActor {
 class ShadowWorkflowActor(workflowId: WorkflowId,
                           startMode: StartMode,
                           workflowSources: WorkflowSourceFiles,
-                          conf: Config) extends LoggingFSM[ShadowWorkflowActorState, ShadowWorkflowActorData] with ActorLogging{
+                          conf: Config) extends LoggingFSM[ShadowWorkflowActorState, ShadowWorkflowActorData] with ActorLogging {
```
kshakir: Revert

geoffjentry: Why?

engine/src/test/resources/application.conf
```diff
@@ -1,44 +1,139 @@
...
+    Local {
       class = "cromwell.engine.backend.local.LocalBackend"
+      actor-factory = "cromwell.backend.impl.local.LocalBackendLifecycleActorFactory"
+      config {
+        // Root directory where Cromwell writes job results.  This directory must be
+        // visible and writeable by the Cromwell process as well as the jobs that Cromwell
+        // launches.
+        root: "cromwell-executions"
+
+        filesystems = {
```
kshakir: Don't use the equals assignment for a sub-config here, and for the other `filesystems` below. See `confg`, `Local`, `providers`, etc. above.

engine/src/main/scala/cromwell/engine/package.scala
```diff
@@ -43,6 +42,10 @@ package object engine {
   class CromwellFatalException(exception: Throwable) extends Exception(exception)
   class PreemptedException(msg: String) extends Exception(msg)
 
+  object CromwellConfiguration {
+    val config: Config = ConfigFactory.load
```
kshakir: I don't feel that this extra layer of singleton adds any benefit, since `ConfigFactory.load` is already [cached](http://typesafehub.github.io/config/v1.3.0/com/typesafe/config/ConfigFactory.html#load--).

Why don't callers who like using singletons (instead of passing in configurations) just use the shorter `ConfigFactory.load` instead of `CromwellConfiguration.Config`?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -53,19 +57,64 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId, workflowDescript
...
+    */
+  private def startJob(jobKey: BackendJobDescriptorKey,
+                       inputs: Map[FullyQualifiedName, WdlValue],
+                       configDescriptor: BackendConfigurationDescriptor,
+                       factory: BackendLifecycleActorFactory): WorkflowExecutionActorState = {
+    val jobDescriptor = BackendJobDescriptor(workflowDescriptor.backendDescriptor, jobKey, inputs)
+    val jobExecutionActor = context.actorOf(
+      factory.jobExecutionActorProps(
+        jobDescriptor,
+        BackendConfigurationDescriptor(configDescriptor.backendConfig, CromwellConfiguration.config)
```
kshakir: Why swap out the passed in `configDescriptor.globalConfig`, instead of passing through the `configDescriptor` as is to the `factory.jobExecutionActorProps`?

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
kshakir: I'm the reviewer and it's my humble opinion. :)

When git barfs with a merge conflict, I'm generally searching through the conflicting lines for "why did someone touch this line?" I personally wouldn't expect someone just to add a single space in the middle of 30+ other files. This line _will_ change eventually, and should be fixed then.

Similarly, when I do git blame, and the line is marked as "[cjllanwarne](https://github.com/broadinstitute/cromwell/blame/e11ca390495247eea2f3c46409faeb47a3ab5f25/engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala#L99)" I don't end up chasing down "[scottfrazer](https://github.com/broadinstitute/cromwell/blame/8b8f7ed10d41a66f3afd8a6dbb6a8500b241d4c3/engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala#L99)" instead.

Lastly, another way of putting it-- if one is fixing this one whitespace character in this one file, why doesn't this PR also fix other spaces in other files? Usually when one is on a cleanup hunt, the entire commit is labeled as such.

kshakir:
A few personal-prefs/questions from me, but generally 👍

engine/src/main/scala/cromwell/engine/workflow/ShadowWorkflowActor.scala
geoffjentry: I disagree that "doesn't clean up everything" is a good reason to avoid "clean up one thing", as the former then relies on a clean up PR that's likely to never happen. If you'd prefer I can submit a cleanup specific PR to add that space w/o having it in the midst of other changes.

kshakir: In the event of an differing opinions, the decision falls back to the submitter Scott. More importantly, it's a single space character, and only my opinion/preference. The PR has my go for merge as is.

engine/src/main/scala/cromwell/engine/package.scala
scottfrazer: Fair enough

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
scottfrazer: Probably just an oversight

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
scottfrazer: that is correct

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
scottfrazer: well it's better than what I had

engine/src/test/resources/application.conf
scottfrazer: Actually I'm a bit confused about the different syntaxes... Seems that below we also use a colon

```
services: {
   KeyValue: {
     class: "cromwell.services.KeyValueServiceActor"
   }
 }
```

I'm not sure what the difference is between `=`, `:`, and space... I guess I have some more reading to do!

kshakir: There's no difference between `=`, `:`, and space. HOCON is to JSON as HTML is to XML. Just like when one writes html you can use `<br>` instead of `<br/>`, HOCON gives config files a lot of flexibility on style. IMHO, just keep it consistent. The reason `=` and `:` are both allowed is because HOCON attempts to be a superset of both JSON and ([mostly](https://github.com/typesafehub/config/blob/master/HOCON.md#note-on-java-properties-similarity)) Java's Property Files.



> Key-value separator
> The = character can be used anywhere JSON allows :, i.e. to separate keys from values.
> If a key is followed by {, the : or = may be omitted. So "foo" {} means "foo" : {}

https://github.com/typesafehub/config/blob/master/HOCON.md#key-value-separator

The reason you're seeing different optional delimiters is because different coders/reviewers let different things pass. In the end, like my other prefs in this PR, it really doesn't matter to the correctness of your patch.

scottfrazer: Thanks!  I'll update this now

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -53,19 +55,59 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId, workflowDescript
...
+        val attemptedActorFactory = CromwellBackend.shadowBackendLifecycleFactory(backendName)
+
+        (attemptedConfigurationDescriptor, attemptedActorFactory) match {
+          case (Success(configDescriptor), Success(factory)) =>
+            startJob(jobKey, inputs, configDescriptor, factory)
+          case (_, _) =>
+            val errors = List(
+              attemptedActorFactory.failed.map(new Exception(s"Could not get BackendLifecycleActor for backend $backendName", _)).toOption,
+              attemptedConfigurationDescriptor.failed.map(new Exception(s"Could not get BackendConfigurationDescriptor for backend $backendName", _)).toOption
+            ) collect { case Some(ex) => ex }
```
mcovarr: I think you can get this with just `flatten`

engine/src/test/scala/cromwell/CromwellTestkitSpec.scala
```diff
@@ -169,7 +184,11 @@ object CromwellTestkitSpec {
     """.stripMargin)
 
   lazy val DefaultLocalBackendConfigEntry = BackendConfigurationEntry(
-    name = "local", className = "cromwell.engine.backend.local.LocalBackend", DefaultLocalBackendConfig)
+    name = "local",
```
mcovarr: are these `name`s in backend entries still required?

mcovarr: there seem to be a few of these throughout

scottfrazer: Yeah it's not really used a lot, but mostly in [initBackends](https://github.com/broadinstitute/cromwell/pull/766/files#diff-aea7de37b083e8ed03b84dd9960ea9b3R44)... maybe it can be restructured so we don't need it but I struggle to see how currently... let me give it some more thought... do you have any ideas?

mcovarr: oh sorry, I thought you had removed the need for this already.  It seemed the key was redundant with `name` and the src/test/application.conf config above appears to be `name`less.  I guess that was not intentional?

README.md
```diff
@@ -320,18 +320,18 @@ Backends are specified in the `backend` configuration block.  Each backend has a
...
 configured backends:
 
 ```hocon
 backend {
   default = "Local"
-  providers = [
-    {
+  providers {
+    Local {
       name = "Local"
```
mcovarr: ☝️

PR #761, opened 2016-04-29
Adding query parameters for metadata endpoint.

kshakir:

PR #767, opened 2016-04-29
Adding query parameters for metadata endpoint. Closes #719 as hotfix

### LINKED ISSUE #719: Large workflow causes metadata endpoint to time out
The same workflow that caused #695 and #699 also has a problem when trying to grab the metadata for the workflow.  We always get a `The server was not able to produce a timely response to your request.` response.  This makes it so we cannot see a timing diagram either.

Example workflow in gotc-dev: 1b3418a6-724b-4e1a-81c5-acdd1bdfe0ee

kshakir:
Workflow execution info retrieval now optionally retrieves events.
Only (lazily) retrieving the workflow metadata future once instead of per-call.

### DIFF
--- a/engine/src/main/resources/swagger/cromwell.yaml
+++ b/engine/src/main/resources/swagger/cromwell.yaml
@@ -373,6 +373,18 @@ paths:
           required: true
           type: string
           in: path
+        - name: outputs
+          description: Include output metadata
+          required: false
+          type: boolean
+          in: query
+          default: true
+        - name: timings
+          description: Include timing metadata
+          required: false
+          type: boolean
+          in: query
+          default: true
       tags:
         - Workflows
       responses:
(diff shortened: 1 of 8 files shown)


geoffjentry:
👍

engine/src/main/resources/swagger/cromwell.yaml
```diff
@@ -373,6 +373,18 @@ paths:
           required: true
           type: string
           in: path
+        - name: outputs
+          description: Include output metadata
+          required: false
+          type: boolean
+          in: query
+          default: true
+        - name: timings
```
kcibul: TOL: nickpick if it's not too late... but these might be better referred to as 'events'.  events do have times associated with them, but this makes me think of performance timings etc not executions state transitions

Horneth:
👍

engine/src/main/resources/swagger/cromwell.yaml
kshakir: Too late for this PR, as this PR is a hotfix/copypasta of #761.

Overall, this was implemented based on comments from #719, but a new ticket that outlines expected behavior could easily be done.

PR #769, opened 2016-05-02
Batch submit as a hotfix. Closes #634

### LINKED ISSUE #634: Batch submission endpoint
As a user (FireCloud) I have the use case of wanting to submit a number of workflows (~1000) that are all the same WDL with different inputs.  For example, running the same pipeline on all samples in my cohort.  Currently I have to do this with 1000 separate calls which is inefficient.

I would like a new batch submission endpoint which allows me to submit a single WDL, a single workflowOptions and an array of inputJsons.

The server will take this and create the individual submissions and return an array of workflow ids are in the same order as the submitted jsons (e.g. the first returned id will be for the workflow that uses the first supplied json and so on).

kshakir:
~~TODO~~ DONE:

- [x] Confirm existing spec TODOs are ok
- [x] Confirm reviewers are ok with naming
- [x] Add same patch to develop

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,3 +5,10 @@
 * The default per-upload bytes size for GCS is now the minumum 256K
 instead of 64M. There is also an undocumented config key
 `google.upload-buffer-bytes` that allows adjusting this internal value.
+
+* Updated Docker Hub hash retriever to parse json with [custom media
+types](https://github.com/docker/distribution/blob/05b0ab0/docs/spec/manifest-v2-1.md).
+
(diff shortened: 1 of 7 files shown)


kshakir:
This should look very familiar to the thumbs of @cjllanwarne / @kcibul / @Horneth

Horneth:
👍

kcibul:
:+1:

kcibul:
except for that part about the failing build :(

kshakir:
My cherry-pick picked out too much. One more time!

kshakir:
Going to squash and merge, even though @kcibul's thumb wasn't recognized as "pull approved"

PR #770, opened 2016-05-02
Batch submit for develop branch

kshakir:


kshakir:
Wheel chooses @Horneth and @cjllanwarne as reviewers

engine/src/main/scala/cromwell/webservice/WorkflowSubmitBatchActor.scala
```diff
@@ -0,0 +1,47 @@
...
+
+  private val responses = Array.fill[Option[WorkflowManagerResponse]](sources.size)(None)
+
+  override def preStart() = {
+    sources.zipWithIndex foreach {
+      case (source, index) =>
+        context.actorOf(
+          Props(new WorkflowSubmitIndexedActor(self, requestHandlerActor, source, index)),
+          "WorkflowSubmitIndexedActor_" + index)
+    }
```
Horneth: Nitpick: If this is done in the `preStart` then I think it's equivalent to messaging directly the WorkflowManagerActor from here instead of creating an actor to do it, and then handling the response in this actor too, because it won't start processing new messages before `preStart` is done.

Horneth:
👍

engine/src/main/scala/cromwell/webservice/WorkflowSubmitBatchActor.scala
kshakir: Not sure I completely follow, but happy to whiteboard out IRL. I thought about this for a while before re-implementing the previous commit that used asks, but had to get something to Alex asap.

> I think it's equivalent to messaging directly the WorkflowManagerActor from here…

The responses **MUST** be kept in the same order as the requested sources. The only thing the `WorkflowSubmitIndexedActor` adds is that it tracks the index order for the responses, that are _not_ guaranteed by akka to arrive in order.

Horneth: Oh ok I understand, I thought the main reason for the `WorkflowSubmitIndexedActor` was that the `WorkflowSubmitBatchActor` could start filling its array as soon as the first workflow is sent, but it's actually for the indexing. 👍

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
```diff
@@ -91,6 +91,23 @@ trait CromwellApiService extends HttpService with PerRequestCreator {
       }
     }
 
+  def submitBatchRoute =
+    path("workflows" / Segment / "batch") { version =>
```
cjllanwarne: Thinking out loud: Is there a reason for making this a separate endpoint? I would think the existing submit could be modified to allow either single or multiple inputs?

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
```diff
@@ -91,6 +91,23 @@ trait CromwellApiService extends HttpService with PerRequestCreator {
...
+      post {
+        formFields("wdlSource", "workflowInputs".?, "workflowOptions".?) {
+          (wdlSource, workflowInputs, workflowOptions) =>
+            requestContext =>
+              import spray.json._
+              workflowInputs.getOrElse("[]").parseJson match {
+                case JsArray(inputses) =>
+                  val sources = inputses.map(inputs => WorkflowSourceFiles(wdlSource, inputs.compactPrint, workflowOptions.getOrElse("{}")))
```
cjllanwarne: Thinking out loud: I take it we never need to support variation in workflowOptions for each input set?

engine/src/main/scala/cromwell/webservice/WorkflowSubmitBatchActor.scala
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  override def receive = {
+    case WorkflowManagerBatchIndexedResponse(response, index) =>
+      responses(index) = Option(response)
```
cjllanwarne: No change required (it would be massive overkill)... I just want to mention a slight sense of sadness that this isn't an FSM :)

cjllanwarne:
Some out-loud thinking but no action required to change anything unless you want to.
👍

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
kshakir: Wasn't requested by the original PR submitter.

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
kshakir: Discussed IRL. Leaving it separate.

kshakir:
As I was adding docs, I realized one shouldn't be using the batch submit w/o an inputs list json. Need a quick re-review by the reviewers.

README.md
```diff
@@ -1273,6 +1274,175 @@ Content-Disposition: form-data; name="workflowOptions"; filename="options.json"
 --f3fd038395644de596c460257626edd7--
 ```
 
+## POST /api/workflows/:version/batch
+
+This endpoint accepts a POST request with a `multipart/form-data`
+encoded body.  The form fields that may be included are:
```
kcibul: can you add somewhere the contract that the order of the input jsons will be the order of the returned workflow ids (with better english than I used!)?

README.md
```diff
@@ -1273,6 +1274,175 @@ Content-Disposition: form-data; name="workflowOptions"; filename="options.json"
...
+    input: in_file=ps.procs
+  }
+}
+
+--64128d499e9e4616adea7d281f695dcb
+Content-Disposition: form-data; name="workflowInputs"; filename="inputs_array.json"
+
+[
+    {
+        "three_step.cgrep.pattern": "..."
```
cjllanwarne: Maybe have n > 1 since it's the example of batch submit?

README.md
```diff
@@ -1273,6 +1274,175 @@ Content-Disposition: form-data; name="workflowOptions"; filename="options.json"
...
+    input: example="gs://my-bucket/cromwell-executions/sfrazer/example.txt", int=3000
+  }
+}
+
+--f3fd038395644de596c460257626edd8
+Content-Disposition: form-data; name="workflowInputs"; filename="jes0_array.json"
+
+[
+  {
+    "sfrazer.x.x": "100"
```
cjllanwarne: Real name in docs

README.md
```diff
@@ -1295,7 +1465,7 @@ $ curl -v "localhost:8000/api/workflows/v1/validate" -F wdlSource=@src/main/reso
 HTTPie:
 
 ```
-$ http --print=hbHB --form POST localhost:8000/api/workflows/v1 wdlSource=@src/main/resources/3step.wdl workflowInputs@inputs.json
+$ http --print=hbHB --form POST localhost:8000/api/workflows/v1/validate wdlSource=@src/main/resources/3step.wdl workflowInputs@inputs.json
```
cjllanwarne: Validate hasn't changed has it? There's no batch validation?

cjllanwarne:
Tiny comments then 👍

engine/src/main/scala/cromwell/webservice/CromwellApiHandler.scala
```diff
@@ -121,6 +123,24 @@ class CromwellApiHandler(requestHandlerActor: ActorRef) extends Actor {
...
+      val requestResponse: Seq[Either[WorkflowSubmitResponse, FailureResponse]] = responses.map {
+        case WorkflowManagerSubmitSuccess(id) => Left(WorkflowSubmitResponse(id.toString, engine.WorkflowSubmitted.toString))
+        case WorkflowManagerSubmitFailure(e) =>
+          Right(e match {
+            case _: IllegalArgumentException => APIResponse.fail(e)
+            case _ => APIResponse.error(e)
+          })
+        case unexpected => Right(FailureResponse("error", s"unexpected message: $unexpected", None))
+      }
+      context.parent ! RequestComplete(StatusCodes.OK, requestResponse)
```
kcibul: This should be a 201/CREATED I think (that's what your docs say also)

engine/src/main/resources/swagger/cromwell.yaml
```diff
@@ -86,6 +86,49 @@ paths:
...
+          in: formData
+        - name: workflowOptions
+          description: Workflow Options JSON
+          required: false
+          type: file
+          in: formData
+      tags:
+        - Workflows
+      responses:
+        '201':
```
kcibul: 200, not 201

kcibul:
👍

README.md
kshakir: This is in the "/validate" documentation section. Scroll up to the curl doc just above. I found this missing-"/validate"-in-the-url typo when doing a search and replace for the regular wdl submit url.

kshakir:
All fixed and squashed. Will merge once tests pass unless someone else has comments before then

PR #777, opened 2016-05-03
Input Evaluation.  Closes #654

### LINKED ISSUE #654: PBE-ize Engine Functions
How to evaluate engine functions in a Pluggable Backends World?

**Acceptance Criteria**:

Use the following WDLs files to test isolated components of engine function implementation:

Read/Write file at a task level when running against the Local Backend:
```
task t {
  Array[String] a = ["a", "b", "c"]
  File x = write_lines(a)
  String y = read_string("/foo/bar/some-public-object.txt")
  command {
    cat ${x}
    echo ${y}
  }
}

workflow w { call t }
```

read functions from the task output section when running against the Local Backend:

```
task t {
  command {
    echo foo
    echo bar
  }
  output {
    Array[String] array = read_lines(stdout())
  }
}

workflow w { call t }
```

**Unanswered Questions**

What do we do when a user wants to call `write_lines` from the workflow level?  If multiple filesystems are supported, then it's unclear where this gets written to.  Perhaps all of them?  If `x` is used by a JES job, then the file *must* be written to at least GCS.

```
workflow w {
  String x = write_lines(["a", "b", "c"])
}
```

** Decision **
We should make this operation unsupported until we have a concrete use case for it, and likely this will be easier when we have filesystem separated out entirely from backend.

Horneth:


engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -105,46 +110,43 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId, workflowDescript
...
-            ),
-            actorName
-          )
-          jobExecutionActor ! ExecuteJobCommand
-          Success(ExecutionStatus.Starting)
-        case Failure(reason) =>
-          log.error(s"Failed to fetch locally qualified inputs for call ${call.fullyQualifiedName}", reason)
-          throw new WorkflowExecutionException(List(reason))
-      }
+    val call = jobKey.call
```
scottfrazer: `call` is unused

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -105,46 +110,43 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId, workflowDescript
...
+    inputResolver.resolveAndEvaluate(jobKey, factory.engineFunctions(workflowDescriptor.backendDescriptor, jobKey, configDescriptor), outputStore) map { inputs =>
+        val jobDescriptor = BackendJobDescriptor(workflowDescriptor.backendDescriptor, jobKey, inputs)
+        val actorName = s"${jobDescriptor.descriptor.id}-BackendExecutionActor-${jobKey.call.fullyQualifiedName}-${jobKey.index.getOrElse("")}-${jobKey.attempt}"
```
scottfrazer: can `-${jobKey.call.fullyQualifiedName}-${jobKey.index.getOrElse("")}-${jobKey.attempt}` be `jobKey.tag`?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -105,46 +110,43 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId, workflowDescript
...
       executionStore = (calls map (BackendJobDescriptorKey(_, None, 1) -> NotStarted)) toMap,
-      symbolCache = buildSymbolStoreEntries.groupBy(entry => SymbolCacheKey(entry.scope, entry.isInput))))
+      outputStore = Map.empty))
 
   /** PBE: the return value of WorkflowExecutionActorState is just temporary.
     *      This should probably return a Try[BackendJobDescriptor], Unit, Boolean,
     *      Try[ActorRef], or something to indicate if the job was started
     *      successfully.  Or, if it can fail to start, some indication of why it
     *      failed to start
     */
```
scottfrazer: I don't think this comment makes sense anymore

Horneth:
also @cjllanwarne

engine/src/main/scala/cromwell/engine/workflow/JobInputEvaluator.scala
```diff
@@ -0,0 +1,176 @@
...
+    outputStore.get(OutputCallKey(call, index)) match {
+      case Some(outputs) =>
+        TryUtil.sequenceMap(outputEntriesToMap(outputs), s"Output fetching for call ${call.unqualifiedName}") map { outputsMap =>
+          WdlCallOutputsObject(call, outputsMap)
+        }
+      case None => Failure(new RuntimeException(s"Could not find call ${call.unqualifiedName}"))
+    }
+  }
+
+  private def buildInputMappingsLookup(jobKey: BackendJobDescriptorKey, outputStore: OutputStore)(identifier: String): WdlValue = {
```
scottfrazer: regarding `(identifier: String)`... some functions here use `identifier: String`, some use `name: String`... maybe unify this?

engine/src/main/scala/cromwell/engine/workflow/JobInputEvaluator.scala
```diff
@@ -0,0 +1,176 @@
...
+    // Will need to be enhanced with scatter variable resolvers
```
scottfrazer: Feels to me like this is an important thing to implement!

engine/src/main/scala/cromwell/engine/workflow/JobInputEvaluator.scala
```diff
@@ -0,0 +1,176 @@
...
+    val resolvers = Stream(lookupWorkflowDeclaration _, lookupCall(jobKey, outputStore) _)
```
scottfrazer: hmm I'm not sure if we lost functionality or not here.  It'd be nice if `lookupWorkflowDeclaration` was instead something like `traverseScopeHierarchy(root=jobKey.call)`, and this could walk up the hierarchy (`root.ancestry`) looking for the variable.  I don't remember if we supported this before, but [scatter blocks can have declarations too](https://github.com/broadinstitute/wdl4s/blob/develop/src/main/resources/grammar.hgr#L339).

`traverseScopeHierarchy` could also be smart enough to address your comment above.

Maybe not something for this PR...

engine/src/main/scala/cromwell/engine/workflow/JobInputEvaluator.scala
```diff
@@ -0,0 +1,176 @@
...
+  /*
+   *  /|\ this could be very time consuming, as it potentially involves IO when using engine functions
+   * /_._\
+   */
+  def resolveAndEvaluate(jobKey: BackendJobDescriptorKey,
+                         wdlFunctions: WdlStandardLibraryFunctions,
+                         outputStore: OutputStore): Try[Map[LocallyQualifiedName, WdlValue]] = {
+    val call = jobKey.call
+    lazy val callInputsFromFile = unqualifiedInputsFromInputFile(call)
+    lazy val inputMappingsLookup = buildInputMappingsLookup(jobKey, outputStore) _
```
scottfrazer: (thinking out loud) I feel like `inputMappingsLookup` and `buildInputMappingsLookup` could be named better... perhaps something like `buildWorkflowScopedLookup(root: BackendJobDescriptorKey, outputStore: OutputStore)`?  I don't know if this suggestion is really any better though.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -185,13 +187,16 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId, workflowDescript
...
       if (newData.isWorkflowComplete) {
-        printOutputs(newData)
+        printOutputs(newData.outputStore)
         goto(WorkflowExecutionSuccessfulState) using newData
       }
       else
         stay() using startRunnableJobs(newData)
     case Event(BackendJobExecutionFailedResponse(jobKey, reason), stateData) =>
-      log.warning(s"Job ${jobKey.call.fullyQualifiedName} failed! Reason: $reason")
+      log.warning(s"Job ${jobKey.call.fullyQualifiedName} failed: $reason")
```
scottfrazer: `jobKey.tag` maybe?

backend/src/main/scala/cromwell/backend/BackendJobExecutionActor.scala
```diff
@@ -54,4 +57,22 @@ trait BackendJobExecutionActor extends BackendJobLifecycleActor with ActorLoggin
     * Abort a running job.
     */
   def abortJob: Unit
+
+  def evaluateOutputs(wdlFunctions: WdlStandardLibraryFunctions,
+                      postMapper: WdlValue => WdlValue = identity) = {
```
scottfrazer: Is it worth keeping this in this PR if it's unused?

engine/src/main/resources/3step.wdl
```diff
@@ -8,7 +8,7 @@ task ps {
 }
 
 task cgrep {
-  String pattern
+  String pattern = "ttys"
```
scottfrazer: I'm curious about why this change was made

engine/src/main/scala/cromwell/engine/workflow/JobInputEvaluator.scala
Horneth: There's no scatter logic in the WorkflowExecutionActor yet

backend/src/main/scala/cromwell/backend/BackendJobExecutionActor.scala
Horneth: I have an update to use it in local backend

engine/src/main/resources/3step.wdl
Horneth: I was just trying stuff and it got caught in the `git add`, I'll remove it

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
Horneth: It's still returning an ExecutionStatus which is weird I think

engine/src/main/scala/cromwell/engine/workflow/JobInputEvaluator.scala
Horneth: Good point yes, I don't think we can run anything that would fail because of this yet but it should be addressed.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
scottfrazer: Perhaps the comment can be updated then?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -105,58 +110,56 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId, workflowDescript
...
+        val actorName = s"${jobDescriptor.descriptor.id}-BackendExecutionActor-${jobKey.tag}"
+        val jobExecutionActor = context.actorOf(
+          factory.jobExecutionActorProps(
+            jobDescriptor,
+            BackendConfigurationDescriptor(configDescriptor.backendConfig, configDescriptor.globalConfig)
+          ),
+          actorName
+        )
+        jobExecutionActor ! ExecuteJobCommand
+        ExecutionStatus.Starting
```
scottfrazer: is this block indented too far?

backend/src/main/scala/cromwell/backend/BackendJobExecutionActor.scala
```diff
@@ -54,4 +57,22 @@ trait BackendJobExecutionActor extends BackendJobLifecycleActor with ActorLoggin
...
+        case (name, value) if value.isSuccess => name -> value.get.wdlValue
+      }
+      def lookup = (currentOutputs ++ inputs).apply _
+
+      val coerced = output.requiredExpression.evaluate(lookup, wdlFunctions) flatMap output.wdlType.coerceRawValue
+
+      val callOutput = output.name -> (coerced flatMap postMapper map { CallOutput(_, None) })
+
+      outputMap + callOutput
+
```
cjllanwarne: There's a lot of empty lines here which could probably be tidied up

backend/src/main/scala/cromwell/backend/BackendJobExecutionActor.scala
```diff
@@ -54,4 +57,22 @@ trait BackendJobExecutionActor extends BackendJobLifecycleActor with ActorLoggin
...
   def abortJob: Unit
+
+  def evaluateOutputs(wdlFunctions: WdlStandardLibraryFunctions,
+                      postMapper: WdlValue => Try[WdlValue] = v => Success(v)) = {
+    val inputs = jobDescriptor.inputs
+    jobDescriptor.call.task.outputs.foldLeft(Map.empty[LocallyQualifiedName, Try[CallOutput]])((outputMap, output) => {
+      val currentOutputs = outputMap collect {
+        case (name, value) if value.isSuccess => name -> value.get.wdlValue
+      }
+      def lookup = (currentOutputs ++ inputs).apply _
```
cjllanwarne: Does "inputs" contain declarations too?

backend/src/main/scala/cromwell/backend/BackendLifecycleActorFactory.scala
```diff
@@ -2,11 +2,15 @@ package cromwell.backend
...
 import wdl4s.Call
+import wdl4s.expression.WdlStandardLibraryFunctions
 
 trait BackendLifecycleActorFactory {
   def workflowInitializationActorProps(workflowDescriptor: BackendWorkflowDescriptor,
                                        calls: Seq[Call],
                                        configurationDescriptor: BackendConfigurationDescriptor): Option[Props]
   def jobExecutionActorProps(jobDescriptor: BackendJobDescriptor, configurationDescriptor: BackendConfigurationDescriptor): Props
   def workflowFinalizationActorProps(): Option[Props]
+  def engineFunctions(workflowDescriptor: BackendWorkflowDescriptor,
```
cjllanwarne: Could we take this opportunity to rename `engine` functions?

engine/src/main/scala/cromwell/engine/workflow/JobInputEvaluator.scala
```diff
@@ -0,0 +1,176 @@
...
+   *              s = a + read_string(f), #(3)
+   *              q = f
+   *          }
+   * }
+   *
+   * will be evaluated such that
+   *
+   * read_string (1) will be evaluated using BFN
+   * read_string (2) will be evaluated using EFN
+   * read_string (3) will be evaluated using BFN
```
cjllanwarne: What is BFN? And are they all really evaluated the same way?

backend/src/main/scala/cromwell/backend/BackendJobExecutionActor.scala
Horneth: It only contains declarations (task input declarations)

engine/src/main/scala/cromwell/engine/workflow/JobInputEvaluator.scala
```diff
@@ -0,0 +1,176 @@
...
+   * the value of q has already been resolved in the appropriate scope and evaluated by the appropriate wdl functions.
+   */
+class JobInputEvaluator(workflowDescriptor: EngineWorkflowDescriptor) {
+
+  private val calls = workflowDescriptor.backendDescriptor.workflowNamespace.workflow.calls
+
+  /*
+   *  /|\ this could be very time consuming, as it potentially involves IO when using engine functions
+   * /_._\
+   */
```
cjllanwarne: Is this is ASCII art going on here?

engine/src/main/scala/cromwell/engine/workflow/JobInputEvaluator.scala
```diff
@@ -0,0 +1,176 @@
...
+      // If it can't be resolved it's ignored and won't appear in the final input map
+      val evaluated: Option[Try[WdlValue]] = declaration.expression match {
+        // Static expression in the declaration
+        case Some(expr) => Option(expr.evaluate(buildMapBasedLookup(inputs), wdlFunctions))
+        // Expression found in the input mappings
+        case None if call.inputMappings.contains(name) => Option(call.inputMappings(name).evaluate(workflowScopedLookup, wdlFunctions))
+        // Expression found in the input file
+        case None if callInputsFromFile.contains(name) => Option(Success(callInputsFromFile(name)))
+        // Expression can't be found
+        case _ => None
```
cjllanwarne: Isn't a missing value a cause for an error?

Horneth: No because of optional inputs

backend/src/main/scala/cromwell/backend/BackendLifecycleActorFactory.scala
Horneth: Yep ! Suggestions ? I was thinking `WdlFunctions`

engine/src/main/scala/cromwell/engine/workflow/JobInputEvaluator.scala
```diff
@@ -0,0 +1,176 @@
...
+
+    TryUtil.sequenceMap(inputEvaluationAttempt, s"Input evaluation for Call ${call.fullyQualifiedName} failed")
+  }
+
+  private def splitFqn(fullyQualifiedName: FullyQualifiedName): (String, String) = {
+    val lastIndex = fullyQualifiedName.lastIndexOf(".")
+    (fullyQualifiedName.substring(0, lastIndex), fullyQualifiedName.substring(lastIndex + 1))
+  }
+
+  // Split inputs map (= evaluated workflow declarations + coerced json inputs) into [init\.*].last
```
cjllanwarne: This is a little confusing. And shouldn't it be `init[\.tail]*` since you don't start with a `.` for a single element name?

engine/src/main/scala/cromwell/engine/workflow/JobInputEvaluator.scala
Horneth: I think any sentence to qualify this with the word "art" in it is most likely far from reality

engine/src/main/scala/cromwell/engine/workflow/lifecycle/ShadowMaterializeWorkflowDescriptorActor.scala
```diff
@@ -216,7 +216,7 @@ class ShadowMaterializeWorkflowDescriptorActor() extends LoggingFSM[ShadowMateri
                                    coercedInputs: WorkflowCoercedInputs,
                                    engineFunctions: WorkflowEngineFunctions): ErrorOr[WorkflowCoercedInputs] = {
     // TODO: Need to create engine-only engine functions!
```
cjllanwarne: Is this still true?

engine/src/main/scala/cromwell/engine/workflow/JobInputEvaluator.scala
Horneth: You're right about the regexp, and it is confusing yes I should put an example instead

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -71,6 +70,10 @@ object WorkflowExecutionActor {
       executionStore.values.forall(isDone)
     }
 
+    def hasFailedJob: Boolean = {
```
cjllanwarne: Perhaps `containsFailedJob` is easier to parse?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/ShadowMaterializeWorkflowDescriptorActor.scala
Horneth: Yes, I mean currently they exist but they're kinda hidden. I'm creating "real" ones in the scatter support branch since we need them also to evaluate scatter collections.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -105,58 +110,56 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId, workflowDescript
...
-            actorName
-          )
-          jobExecutionActor ! ExecuteJobCommand
-          Success(ExecutionStatus.Starting)
-        case Failure(reason) =>
-          log.error(s"Failed to fetch locally qualified inputs for call ${call.fullyQualifiedName}", reason)
-          throw new WorkflowExecutionException(List(reason))
-      }
+    // FIXME This is a potential bottleneck as input evaluation can be arbitrarily long and executes code coming from the backend.
+    // It should probably be actorified / isolated
```
cjllanwarne: Since the process of running a job is looking a lot like a self-contained state machine (i.e. you have to init, evaluate inputs, begin executing, poll for completion, evaluate outputs) then this idea is awesome and I fully support a sub-actor!

Plus it allows this actor to worry about co-ordinating calls, while the sub-actor can handle the tedium of actually running the jobs.

engine/src/test/scala/cromwell/CromwellTestkitSpec.scala
```diff
@@ -40,6 +41,8 @@ case class TestBackendLifecycleActorFactory(config: Config) extends BackendLifec
   }
 
   override def workflowFinalizationActorProps(): Option[Props] = None
+
+  override def engineFunctions(workflowDescriptor: BackendWorkflowDescriptor, jobKey: BackendJobDescriptorKey, configurationDescriptor: BackendConfigurationDescriptor): WdlStandardLibraryFunctions = ???
```
cjllanwarne: Is this a good idea? Is there a simple no-op engine function we can use here instead?

backend/src/main/scala/cromwell/backend/BackendLifecycleActorFactory.scala
cjllanwarne: How about Expression Language Functions (`ELF`s)

(and that's the last time you make the mistake of asking me to name anything!)

cjllanwarne:
This is awesome! A few comments but none should be considered blockers. 👍

engine/src/main/scala/cromwell/engine/workflow/JobInputEvaluator.scala
cjllanwarne: https://en.wikibooks.org/wiki/Unicode/List_of_useful_symbols#Health_and_safety

backend/src/main/scala/cromwell/backend/BackendJobExecutionActor.scala
```diff
@@ -54,4 +57,22 @@ trait BackendJobExecutionActor extends BackendJobLifecycleActor with ActorLoggin
     * Abort a running job.
     */
   def abortJob: Unit
+
+  def evaluateOutputs(wdlFunctions: WdlStandardLibraryFunctions,
```
geoffjentry: when would postMapper be something else?

Horneth: JES gcs paths conversion

scottfrazer:
👍

Issue #668, opened 2016-04-06
Update Engine to call Metadata Service API


ruchim:

geoffjentry:
With an n of 2 I've now heard 2 opposing viewpoints on how this should happen but yet still don't have a good handle on what the actual issue is, technical pros/cons, etc. I'd like to talk about this further before anyone acts on it.

cjllanwarne:
Depends on #782, #783

gauravs90:
Hey @kcibul / @geoffjentry ! Can we add a little more details to this ticket? For instance, what all stuff from the engine do we think should go into the MetadataService? I'd like to take this up but not sure what stuff should we be adding.

PR #795, opened 2016-05-06
Add retry support in WorkflowExecutionActor. Closes #655.

### LINKED ISSUE #655: Implement Retryable in Workflow Actor
The (edit by Chris: ~~Workflow Actor~~ WorkflowExecutionActor) should take action based on the kind of failure returned by the BE and retry if so indicated

gauravs90:
Adds an event handler for a backend job that failed with a retry able failure. The engine attempts to restart the failed job again on the same backend for a `max` number of times (where that number is configurable and controlled by the engine)


engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -221,6 +233,19 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId, workflowDescript
       log.info(s"$tag transitioning from $fromState to $toState.")
   }
 
+  private def handleRetryableFailure(jobKey: BackendJobDescriptorKey) = {
```
gauravs90: This will probably involve reporting the failed event to a metadata reporting service or something later on. Currently, it simply increases the number of attempts for the JobKey and resets it's status as `NotStarted` (so that it's picked up by the startRunnableCalls for re-execution). In essence, the ExecutionStore, at any given time, will only have jobs with either `NotStarted`, `Starting` / `Running`, `Failed` or `Done` statuses. If the use is more complex than this, please do let me know.

engine/src/test/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActorSpec.scala
```diff
@@ -0,0 +1,52 @@
...
+    """
+      |runtime {
+      | backend: "Mock"
+      |}
+    """.stripMargin
+
+  val workflowId = WorkflowId.randomId()
+  val engineWorkflowDescriptor = createMaterializedEngineWorkflowDescriptor(workflowId, SampleWdl.HelloWorld.asWorkflowSources(runtime = runtimeSection))
+
+  "WorkflowExecutionActor" should {
```
gauravs90: I wanted to write more unit tests but it's gotten very difficult to write more without repeating lots of code (like creating new Mock backends). The `CromwellBackend` singleton is used in the WorkflowExecutionActor (and a few other places) which prevent's its (stubbed-out) injection into these classes.

gauravs90:
@mcovarr and @Horneth please!

engine/src/test/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActorSpec.scala
cjllanwarne: Would it make things easier if we did inject the `CromwellBackend` into the class instead of relying on a global reference?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -95,6 +99,8 @@ object WorkflowExecutionActor {
     override val exceptionContext = s"WorkflowExecutionActor"
   }
 
+  val MaxRetries = ConfigFactory.load().getInt("system.maxRetries")
```
cjllanwarne: It'd be good to either make sure this value exists at start-up time, or make this optional.

It'd be pretty awkward to have to wait until a workflow is submitted to find out the config file is invalid

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -182,6 +188,7 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId, workflowDescript
 
   when(WorkflowExecutionInProgressState) {
     case Event(BackendJobExecutionSucceededResponse(jobKey, callOutputs), stateData) =>
+      sender() ! PoisonPill
```
cjllanwarne: In theory the sub-actors are already killing themselves so this should be redundant. But that doesn't necessarily mean its a bad idea!

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
cjllanwarne: Could you add this as a note to #668

engine/src/main/resources/application.conf
```diff
@@ -36,6 +36,9 @@ system {
 
   // If 'true', a SIGINT will trigger Cromwell to attempt to abort all currently running jobs before exiting
   abort-jobs-on-terminate = false
+
+  // Max number of retries per job that the engine will attempt in case of a retryable failure received from the backend
+  maxRetries = 3
```
Horneth: I guess it doesn't hurt to have that threshold, although in the case of JES for example where this is used for preemptible VMs, the backend decides how many times a job needs to be retried if it gets preempted, based on the runtime attributes. So a WDL task with `preemptible: 5` runtime attribute running on JES, would actually only be retried 3 times if it runs on a Cromwell configured as above.

gauravs90: I can remove that cap from the engine side, and retry as long as I am receiving retry able failures.. It was like that before since there backed was a part of engine. But since it is outside now, I thought it'll also be better to have a limit on the engine side.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
gauravs90: Will do that. 
Perhaps, do you think it's fair to overwrite the old entry in the execution store in this way?

cjllanwarne: I would coordinate with @Horneth since his PR is also in this area.

kcibul:
Since @cjllanwarne has been making a lot of comments can we switch the reviewer to him and @Horneth (rather than @mcovarr)?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
kcibul: I believe currently (and the in the future) we want to treat these as separate calls and not overwrite them, at the very least from a metadata perspective, so that after the fact you know things about the failed tasks (how long they ran, etc)

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
kcibul: Currently -- the use case for this functionality is via the WDL "preemptible" count... is that what we are handling here?

engine/src/main/resources/application.conf
mcovarr: While this feature is obviously supported on the engine side, I agree with @Horneth that this should be specified within the backend and not at the engine level.

Also a nitpick that wherever this is implemented, this file is generally moving over to `lower-case-separated-by-dashes` for key names.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
mcovarr: @cjllanwarne in conjunction with your earlier observation that this is more appropriately a backend property, I might suggest having the actor factory sanity check this on startup for fail fast.  I had similar concerns in this forthcoming JES PR.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
mcovarr: Uh hmm, I almost I agree with that; if we start sending `PoisonPill`s we might start to see code that relies on that pattern.  I think it would be preferable to require the offending actors to poison themselves.

engine/src/main/resources/application.conf
gauravs90: @mcovarr Are you talking just about the property, or this feature as a whole? As in, do we want the engine to not have a limit and *always* retry a job if we receive a retry able failure? (relying on the backend to have a limit on it's side)

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
mcovarr: I would defer to @Horneth as the author of preemptibility, but I thought the original key would get status "preempted" and a new key would be created with status "not started".

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
gauravs90: Gotcha. I can remove these. (As a note though, currently the children are not poisoning themselves [PS> Not to be taken literally] )

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
gauravs90: I'll add a new entry with 'Preempted' then. (unless @Horneth  has a different opinion)

engine/src/test/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActorSpec.scala
gauravs90: I'll take a look at this but I'm afraid that it would still have a few problems. (I'll see for other potential solutions in case)

engine/src/main/resources/application.conf
mcovarr: I was thinking the property should be on the backend, but then the "feature as a whole" question gets interesting.  If the engine is made responsible for enforcing this limit, how does it know what the backend's retry limit is?

gauravs90: My opinion here was to have a decent enough default value in the engine side to accommodate our current use cases (current value of `3` is just a start.. we can and should probably increase it). But relying solely on the backend doesn't sound like a very good idea in the world of 3rd party developers. This can later (quite later) on also have the added benefit that after certain number of retries, if the job still fails, we can schedule it on a different backend maybe.

gauravs90:
@kcibul I think @mcovarr and @cjllanwarne are equal now 😄 
Anyways, i'll take the votes of the first two in this case I guess.

engine/src/main/resources/application.conf
kcibul: That's a good point @mcovarr maybe the way to think about this is that as long as the backend is saying "this is retryable" the engine should retry it.  If the backend always returns retryable.. then we will retry forever... which would be a bad-backend.  And perhaps this "max" is really just a safety valve on that (although I would set it to like 10-20 or something).  The backend should know (a) the current attempt # and (b) the runtime attributes for a call.  So for the JES PBE it would decide that on failure # < runtime attribute for premptible it would return "retryable".  And then also when running if attempt # < premptible # (from runtime attributes) to launch a preemptible VM otherwise launch an on-demand one.

Maybe y'all already knew this and I'm just stating the obvious..

gauravs90: @kcibul A backend already knows that (the attempt # is embedded in the jobkey). And I would set this number to a number higher or equal to the largest that you've seen so far in your workloads. Just pick something and let me know, I'll update it's value here.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
gauravs90: \all : Keeping the failed Job in the execution store with `Preempted` status now, and creating a new entry for subsequent retries.

engine/src/main/resources/application.conf
geoffjentry: I like the idea of leaving a safety valve in the engine although primarily relying on the backend. Someone (I think it was @mcovarr) made the point on a different topic a week or so ago that we're going to have to keep bulletproofing from craptastic backends in mind from here on out.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -106,6 +110,15 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId, workflowDescript
 
   private val calls = workflowDescriptor.backendDescriptor.workflowNamespace.workflow.calls
 
+  // TODO: We should probably create a trait which loads all the configuration (once per application), and let classes mix it in
+  // to avoid doing ConfigFactory.load() at multiple places
+  val MaxRetries = Try(ConfigFactory.load().getInt("system.max-retries")) match {
```
gauravs90: @cjllanwarne Something like this? I didn't want to jam this value in CromwellBackend or BackendConfiguration since it doesn't quite belong there and I'm trying to avoid more singleton accesses in this.

engine/src/main/resources/application.conf
mcovarr: OK so assuming a JES backend with a preemptible limit of 3, does the following sound right?

If the JES backend runs attempt 3 on a preemptible VM that gets preempted, the backend should message the engine that the call has failed retriably.  The engine would launch the 4th attempt for the call, but now the JES backend would launch this attempt on a non-preemptible VM.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -106,6 +110,15 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId, workflowDescript
 
   private val calls = workflowDescriptor.backendDescriptor.workflowNamespace.workflow.calls
 
+  // TODO: We should probably create a trait which loads all the configuration (once per application), and let classes mix it in
```
geoffjentry: Can you create a ticket for this - specifically the part about not calling `.load` everywhere although include your trait idea as an example (I like the idea, but don't want to specify that as the definitive solution at the moment)

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
geoffjentry: There should be a getIntOption from the lenthall implicits, it's the same idea but a little cleaner looking IMO

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -106,6 +110,15 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId, workflowDescript
...
+  // to avoid doing ConfigFactory.load() at multiple places
+  val MaxRetries = Try(ConfigFactory.load().getInt("system.max-retries")) match {
+    case Success(value) => value
+    case Failure(reason) =>
+      log.warning(s"Failed to load the max-retries value from the configuration. Defaulting back to a value of `10` : ${reason.getMessage}")
+      10
```
geoffjentry: Weren't you the one who wanted the 'no magic numbers' check in the scalastyle xml? :)

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
```diff
@@ -67,12 +68,15 @@ object WorkflowExecutionActor {
...
-      def isDone(executionStatus: ExecutionStatus): Boolean = executionStatus == ExecutionStatus.Done
+      def isDone(executionStatus: ExecutionStatus): Boolean = (executionStatus == ExecutionStatus.Done) || (executionStatus == ExecutionStatus.Preempted)
       executionStore.values.forall(isDone)
     }
 
     /** Updates the status of a a job by the new entry */
     def updateExecutionStoreStatus(entry: (JobKey, ExecutionStatus)): WorkflowExecutionActorData = this.copy(executionStore = executionStore + entry)
+
+    def addNewJob(jobKey: JobKey, executionStatus: ExecutionStatus = ExecutionStatus.NotStarted): WorkflowExecutionActorData =
+      this.copy(executionStore = executionStore + (jobKey -> executionStatus))
```
mcovarr: Do you think there's value in having this method?  it seems kind of redundant with the above, they both just add a KV entry.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
Horneth: @gauravs90 This looks good to me, it's how it's working in `OldWorld` at least. I believe the code that's here now is not the one your first comment was referring to. I think you're right that strictly speaking we could overwrite the existing key in the store with NotStarted. The engine doesn't need to remember the failed attempts. The reason for the duplication was that we had a 1:1 mapping between the execution store and the Database, and the metadata came from the executions table.
Maybe keep it like this now, and we can remove it once Restart / Metadata is implemented and we're sure it's really useless.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
gauravs90: Eeek! I'll secretly sneek `10` to the list of the ignored parameters for MagicNumber checks.

gauravs90: Or does the style check get circumvented if we use something on the following lines ?

```scala
val MaxRetries = Seq.fill(10)(scala.util.Random.nextInt(1)).size
```

mcovarr: can't you do a constant `val DefaultMaxRetries = 10` and return that?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
gauravs90: Thanks! There was a getIntOption from lenthal.config.ScalaConfig. Have used it now

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
gauravs90: Ahh yes you're right. I spent a lot of time to not create the very function (the one that was there before this commit) but gave up and had to create one. Then when I changed the stuff to keep the Preempted jobs,  I just changed it little. I can definitely remove this now. Thanks!

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
gauravs90: Done.  #796.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowExecutionActor.scala
gauravs90: @mcovarr I was kidding mostly there 😕 
I have used something like that currently now.

mcovarr: phew

gauravs90:
Hey @Horneth and ( @mcovarr / @cjllanwarne ), I have squashed rebased with the latest develop. It'd be great if you guys can take a look. Thanks!

Horneth:
👍

engine/src/main/resources/application.conf
```diff
@@ -32,6 +32,9 @@ spray.can {
 system {
   // If 'true', a SIGINT will trigger Cromwell to attempt to abort all currently running jobs before exiting
   abort-jobs-on-terminate = false
+
+  // Max number of retries per job that the engine will attempt in case of a retryable failure received from the backend
+  max-retries = 3
```
mcovarr: The engine-level max should probably be higher than this.  For JES I believe we usually allow 3 preemption attempts, which means the engine would be asked to kick off a fourth attempt (which the JES backend would run non-preemptibly).

mcovarr:
:+1: with a nitpick

engine/src/main/resources/application.conf
gauravs90: Think @kcibul was mentioning somewhere around 10-20. Do you think 10 should be fine?

mcovarr: Yes that sounds like plenty.  😄

PR #804, opened 2016-05-09
Add the docker command to the Cromwell container

coreone:


mcovarr:
@coreone could you describe what's going on in these changes?

ruchim:
@coreone once we have a little bit of a description, I'd love to review this PR (short & sweet), and you can choose any one other person from the team.

coreone:
@ruchim @mcovarr This is just to add the docker executable to the Cromwell container such that Cromwell can run local workflows using the docker command when Cromwell itself is running inside a container. The magic to make it all work is to mount the /var/run/docker.sock file into the Cromwell container so that the internal docker command can communicate with docker running on the host.

This is the PR to get around having to change code to use the docker API instead of shelling out to *docker* that we discussed in standup last week.

mcovarr:
:+1:

coreone:
Added some cleanup after the apt-get install to clean up that layer due to suggestions from @hjfbynara

hjfbynara:
👍

coreone:
@mcovarr do you want to re-approve after my changes, or are we going to say a cumulative 2 thumbs is good?  If good, I'll just merge.

mcovarr:
:+1: even better

PR #818, opened 2016-05-10
Shadow scatter Closes #790

### LINKED ISSUE #790: Add Scatter Support to WorkflowExecutionActor

Horneth:
Implement scatter execution and re-organize Execution related code

- [x] Write engine scatter test
- [ ] Write more ?


engine/src/main/scala/cromwell/engine/EngineELF.scala
```diff
@@ -0,0 +1,29 @@
...
+import java.nio.file.{FileSystem, FileSystems}
+
+import cromwell.backend.wdl.{PureFunctions, ReadLikeFunctions}
+import cromwell.core.WorkflowOptions
+import wdl4s.expression.WdlStandardLibraryFunctions
+import wdl4s.values.{WdlFile, WdlValue}
+
+import scala.util.{Failure, Try}
+
+class EngineELF(workflowOptions: WorkflowOptions) extends WdlStandardLibraryFunctions with ReadLikeFunctions with PureFunctions {
```
scottfrazer: I really liked CromwellWdlFunctions better :-/

Horneth: Yeah me too :/ I was trying to be consistent.
It's easy to rename anyway, maybe we could take a poll on what they should be renamed to ? :p

scottfrazer: 👍 to having a poll

geoffjentry: even if it remains as ELF it should be Elf

backend/src/main/scala/cromwell/backend/wdl/WriteFunctions.scala
```diff
@@ -0,0 +1,54 @@
...
+
+  private lazy val absoluteDirectory = {
+    val abs = writeDirectory.toAbsolutePath
+    abs.createDirectories
+    abs
+  }
+
+  override def tempFilePath = absoluteDirectory.toString
+
+  def writeTempFile(path: String,prefix: String,suffix: String,content: String): String = throw new NotImplementedError("This method is not used anywhere and should be removed")
```
scottfrazer: Why not remove it?

Horneth: Yes I wanted to but I forgot, and it's as Wdl4s method so it needs to be removed from there first.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
```diff
@@ -0,0 +1,78 @@
...
+
+      /**
+        * Otherwise, simply refer to the entry the collector entry.  This means that 'entry' depends
+        * on every shard of the pre-requisite scope to finish.
+        */
+      case _ =>
+        store filter { case (k, _) => k.scope == prerequisiteScope && k.index.isEmpty } toSeq
+    }
+  }
+}
```
mcovarr: \n

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/InputEvaluation.scala
```diff
@@ -0,0 +1,73 @@
...
+  def workflowDescriptor: EngineWorkflowDescriptor
+
+  // Split inputs map (= evaluated workflow declarations + coerced json inputs) into [init\.*].last
+  private lazy val splitInputs = workflowDescriptor.backendDescriptor.inputs map {
+    case (fqn, v) => splitFqn(fqn) -> v
+  }
+
+  /*
+ *  /|\ this could be very time consuming, as it potentially involves IO when using engine functions
+ * /_._\
```
mcovarr: return `Future` instead?

core/src/main/scala/cromwell/core/package.scala
```diff
@@ -1,14 +1,17 @@
...
 import wdl4s.values.{SymbolHash, WdlValue}
 
 import scalaz._
 
 package object core {
   // root can be a Path instead of a String in PBE. stdout / err too but it doesn't really bring values since they're just stringified to WdlFiles
-  class WorkflowContext(val root: String)
-  class CallContext(override val root: String, val stdout: String, val stderr: String) extends WorkflowContext(root)
+  class OldWorkflowContext(val root: String)
+  class OldCallContextOld(override val root: String, val stdout: String, val stderr: String) extends OldWorkflowContext(root)
```
mcovarr: Why a trailing `Old`?

engine/src/main/resources/application.conf
```diff
@@ -193,6 +193,15 @@ backend {
...
+filesystems {
+  "gcs" {
+    auth = ""
+  }
+  "local" {
+
+  }
+}
+
 services {
```
mcovarr: stowaway?

engine/src/main/scala/cromwell/engine/EngineELF.scala
mcovarr: why not just `EngineFunctions` like everything it extends?

engine/src/main/scala/cromwell/engine/backend/jes/JesEngineFunctions.scala
```diff
@@ -3,30 +3,25 @@ package cromwell.engine.backend.jes
...
-import cromwell.{CallEngineFunctions, WorkflowEngineFunctions}
 import wdl4s.values._
 
 import scala.language.postfixOps
 import scala.util.Try
 @deprecated(message = "This class will not be part of the PBE universe", since = "May 2nd 2016")
-class JesWorkflowEngineFunctions(fileSystems: List[FileSystem], context: WorkflowContext) extends WorkflowEngineFunctions(context) {
-  import cromwell.core.PathFactory._
-
+class JesOldWorkflowEngineFunctions(val fileSystems: List[FileSystem], context: OldWorkflowContext) extends OldWorkflowEngineFunctions(context) with PathFactory {
```
mcovarr: `Old` in a weird place here

engine/src/main/scala/cromwell/engine/backend/jes/JesEngineFunctions.scala
```diff
@@ -3,30 +3,25 @@ package cromwell.engine.backend.jes
...
-  override def toPath(str: String): Path = {
-    val path = buildPath(str, fileSystems)
-    if (!path.isAbsolute) context.root.toAbsolutePath(fileSystems).resolve(path) else path
-  }
+  override def postMapping(path: Path) = if (!path.isAbsolute) context.root.toAbsolutePath(fileSystems).resolve(path) else path
 }
 
 @deprecated(message = "This class will not be part of the PBE universe", since = "May 2nd 2016")
-class JesCallEngineFunctions(fileSystems: List[FileSystem], context: CallContext) extends JesWorkflowEngineFunctions(fileSystems, context) with CallEngineFunctions {
+class JesCallEngineFunctionsOld(override val fileSystems: List[FileSystem], context: OldCallContextOld) extends JesOldWorkflowEngineFunctions(fileSystems, context) with OldCallEngineFunctions {
```
mcovarr: `Old` should be prefix

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
```diff
@@ -0,0 +1,78 @@
...
+import cromwell.backend.{BackendJobDescriptorKey, JobKey}
+import cromwell.engine.ExecutionStatus
+import cromwell.engine.ExecutionStatus._
+import cromwell.engine.workflow.lifecycle.execution.ExecutionStore.ExecutionStoreEntry
+import cromwell.engine.workflow.lifecycle.execution.WorkflowExecutionActor.CollectorKey
+import wdl4s.{Scatter, Scope}
+
+import scala.language.postfixOps
+
+object ExecutionStore {
```
mcovarr: yay breaking this out into its own class

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/JobStarterActor.scala
```diff
@@ -0,0 +1,43 @@
...
+                           factory: BackendLifecycleActorFactory,
+                           configDescriptor: BackendConfigurationDescriptor) extends Actor with InputEvaluation {
+
+  override val workflowDescriptor: EngineWorkflowDescriptor = executionData.workflowDescriptor
+  override val executionStore: ExecutionStore = executionData.executionStore
+  override val outputStore: OutputStore = executionData.outputStore
+  override val expressionLanguageFunctions = factory.expressionLanguageFunctions(workflowDescriptor.backendDescriptor, jobKey, configDescriptor)
+
+  override def receive = {
+    case Start => resolveAndEvaluate(jobKey, expressionLanguageFunctions) map { inputs =>
```
mcovarr: if you `Future`ize the inputs evaluation, this cool little actor might become an FSM which could transition through its states as it evaluates.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/OutputStore.scala
```diff
@@ -0,0 +1,58 @@
...
+    val shardsOutputs = shards.toSeq sortBy { _.index.fromIndex } map { e =>
+      fetchCallOutputEntries(e.scope, e.index) map { _.outputs } getOrElse(throw new RuntimeException(s"Could not retrieve output for shard ${e.scope} #${e.index}"))
+    }
+    collector.scope.task.outputs map { taskOutput =>
+      val wdlValues = shardsOutputs.map(s => s.getOrElse(taskOutput.name, throw new RuntimeException(s"Could not retrieve output ${taskOutput.name}")))
+      val arrayOfValues = new WdlArray(WdlArrayType(taskOutput.wdlType), wdlValues)
+      taskOutput.name -> CallOutput(arrayOfValues, None)
+    } toMap
+  }
+}
```
mcovarr: \n

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/OutputStore.scala
```diff
@@ -0,0 +1,58 @@
...
+import cromwell.engine.workflow.lifecycle.execution.WorkflowExecutionActor.CollectorKey
+import cromwell.util.TryUtil
+import wdl4s.{Call, Scope}
+import wdl4s.types.{WdlArrayType, WdlType}
+import wdl4s.values.{WdlArray, WdlCallOutputsObject, WdlValue}
+
+import scala.language.postfixOps
+import scala.util.{Failure, Success, Try}
+
+object OutputStore {
```
mcovarr: more yay for broken-out class

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -0,0 +1,312 @@
...
+    */
+  private case class JobInitializationFailed(throwable: Throwable)
+  private case class ScatterCollectionFailedResponse(collectorKey: CollectorKey, throwable: Throwable)
+  private case class ScatterCollectionSucceededResponse(collectorKey: CollectorKey, outputs: CallOutputs)
+
+  /**
+    * Internal ADTs
+    */
+  // Represents a scatter in the execution store. Index is here in prevision for nested scatters
+  case class ScatterKey(scope: Scatter, index: ExecutionIndex) extends JobKey {
```
mcovarr: does `index` need to be here?  it seems like this could lead to confusion

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesExpressionFunctions.scala
```diff
@@ -0,0 +1,32 @@
...
+    path.toAbsolutePath(fileSystems).asDirectory.glob("**/*") map { _.path.fullPath } filterNot { _.toString == path } toSeq
+  }
+
+  override def postMapping(path: Path): Path = if (!path.isAbsolute) context.root.resolve(path) else path
+
+  override def stdout(params: Seq[Try[WdlValue]]) = Success(WdlFile(context.stdout))
+  override def stderr(params: Seq[Try[WdlValue]]) = Success(WdlFile(context.stderr))
+
+  override val writeDirectory: Path = context.root
+}
```
mcovarr: \n

engine/src/main/scala/cromwell/engine/EngineELF.scala
Horneth: Because they don't really have anything to do with the engine in particular, not more than with the backend ? :/

geoffjentry: What about just `WdlFunctions`?

Horneth: Yes that was my first thought

geoffjentry: As I always say, @Horneth's first thought is always correct ;)

Horneth: If only xD

backend/src/main/scala/cromwell/backend/wdl/ReadLikeFunctions.scala
```diff
@@ -1,27 +1,16 @@
...
-trait WdlStandardLibraryImpl extends WdlStandardLibraryFunctions {
-
-  def toPath(str: String): Path
 
-  override final def fileContentsToString(path: String): String = toPath(path).toAbsolutePath.contentAsString
-
-  private def writeContent(baseName: String, content: String): Try[WdlFile] = {
-    Try(WdlFile(writeTempFile(tempFilePath, s"$baseName.", ".tmp", content)))
-  }
+trait ReadLikeFunctions extends FileSystems { this: WdlStandardLibraryFunctions =>
```
gauravs90: What's the difference between the `PureFunctions` trait above and this trait? As in we have two similar implementations of the `sub` function defined in both traits

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/JobStarterActor.scala
```diff
@@ -0,0 +1,122 @@
...
+                           jobKey: BackendJobDescriptorKey,
+                           factory: BackendLifecycleActorFactory,
+                           configDescriptor: BackendConfigurationDescriptor) extends LoggingFSM[JobStarterActorState, JobStarterActorStateData] with WdlLookup {
+
+  override val workflowDescriptor: EngineWorkflowDescriptor = executionData.workflowDescriptor
+  override val executionStore: ExecutionStore = executionData.executionStore
+  override val outputStore: OutputStore = executionData.outputStore
+  override val expressionLanguageFunctions = executionData.expressionLanguageFunctions
+
+  startWith(Idle, new JobStarterActorStateData())
```
gauravs90: The StateData can probably just be `Unit` if we don't intend on using `JobStarterActorStateData`

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/JobStarterActor.scala
```diff
@@ -0,0 +1,122 @@
...
+          context stop self
+      }
+      goto(EvaluatingInputs)
+  }
+
+  when(EvaluatingInputs) {
+    case Event(EvaluatedInputs(inputs), _) =>
+      val jobDescriptor = BackendJobDescriptor(workflowDescriptor.backendDescriptor, jobKey, inputs)
+      val props = factory.jobExecutionActorProps(jobDescriptor, BackendConfigurationDescriptor(configDescriptor.backendConfig, configDescriptor.globalConfig))
+      context.parent ! BackendJobStartSucceeded(jobDescriptor, props)
```
gauravs90: This name `BackendJobStartSucceeded` (and consequently the name of this class too) is a little confusing IMO.. since it mostly deals with preparing the Job rather than starting anything 😕

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -0,0 +1,312 @@
...
+    case e => throw new RuntimeException("Could not instantiate backend factories", e)
+  } get
+
+  private val configs = TryUtil.sequenceMap(workflowDescriptor.backendAssignments.values.toSeq map { backendName =>
+    backendName -> BackendConfiguration.backendConfigurationDescriptor(backendName)
+  } toMap) recover {
+    case e => throw new RuntimeException("Could not instantiate backend configurations", e)
+  } get
+
+  private val expressionLanguageFunctions = new EngineELF(workflowDescriptor.backendDescriptor.workflowOptions)
```
gauravs90: Unused?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -0,0 +1,312 @@
...
+final case class WorkflowExecutionActor(workflowId: WorkflowId, workflowDescriptor: EngineWorkflowDescriptor) extends LoggingFSM[WorkflowExecutionActorState, WorkflowExecutionActorData] {
+
+  import WorkflowExecutionActor._
+  import lenthall.config.ScalaConfig._
+
+  val tag = self.path.name
+  private lazy val DefaultMaxRetriesFallbackValue = 10
+
+  // TODO: We should probably create a trait which loads all the configuration (once per application), and let classes mix it in
+  // to avoid doing ConfigFactory.load() at multiple places
```
scottfrazer: `ConfigFactory.load` is cached anyway, so we don't need to create a separate trait for this

backend/src/main/scala/cromwell/backend/wdl/ReadLikeFunctions.scala
Horneth: I forgot to remove sub from `ReadLikeFunctions`, it's done now.`PureFunctions` are functions that don't need IO.

mcovarr:
:+1:

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -0,0 +1,310 @@
...
+  }
+
+  /**
+    * Attempt to start all runnable jobs and return updated state data.  This will create a new copy
+    * of the state data including new pending persists.
+    */
+  @tailrec
+  private def startRunnableScopes(data: WorkflowExecutionActorData): WorkflowExecutionActorData = {
+    val runnableScopes = data.executionStore.runnableScopes.toList
+    val runnableCalls = runnableScopes.view collect { case k if k.scope.isInstanceOf[Call] => k } sortBy { _.index.getOrElse(-1) } map { _.tag }
```
scottfrazer: Can this be:

```scala
val runnableCalls = runnableScopes.view collect { case c: Call => c }
```

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -0,0 +1,310 @@
...
+    if (runnableCalls.nonEmpty) log.info(s"Starting calls: " + runnableCalls.mkString(", "))
+
+    // Each process*** returns a Try[WorkflowExecutionDiff], which, upon success, contains potential changes to be made to the execution store.
```
scottfrazer: `process***`?

scottfrazer: maybe `// Each process method ...`?

mcovarr: processRunnableThingmajig

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
Horneth: `runnableScopes` are actually `runnableKeys` (they're `JobKey`s), which I should rename.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
Horneth: `Each processRunnable... method` ?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -0,0 +1,310 @@
...
+    WorkflowExecutionActorData(
+      workflowDescriptor,
+      executionStore = buildExecutionStore(),
+      outputStore = OutputStore.empty))
+
+  private def buildExecutionStore(): ExecutionStore = {
+    val workflow = workflowDescriptor.backendDescriptor.workflowNamespace.workflow
+    // Only add direct children to the store, the rest is dynamically created when necessary
+    val callExecutions = Scope.collectCalls(workflow.children) map { BackendJobDescriptorKey(_, None, 1) -> NotStarted }
+    val scatterExecutions = Scope.collectScatters(workflow.children) map { ScatterKey(_) -> NotStarted }
```
scottfrazer: I feel like this algorithm can be a little better (in the category of "thinking out loud").  I'm not a fan of the `collectCalls` and `collectScatters` methods, I think they're mostly a relic from when the `Scope` class was first created.

Maybe something like

```scala
val keys = workflow.children map {
  case call: Call => BackendJobDescriptorKey(call, None, 1)
  case scatter: Scatter => ScatterKey(scatter)
}

ExecutionStore(keys.map(_ -> NotStarted).toMap)
```

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -0,0 +1,310 @@
...
+
+  // Initialize the StateData with ExecutionStore (all calls as NotStarted) and SymbolStore
+  startWith(
+    WorkflowExecutionPendingState,
+    WorkflowExecutionActorData(
+      workflowDescriptor,
+      executionStore = buildExecutionStore(),
+      outputStore = OutputStore.empty))
+
+  private def buildExecutionStore(): ExecutionStore = {
```
scottfrazer: Again just a thinking out loud nitpick, maybe `buildInitialExecutionStore`?

Horneth: ok

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActorData.scala
```diff
@@ -0,0 +1,72 @@
...
+}
+
+case class WorkflowExecutionActorData(workflowDescriptor: EngineWorkflowDescriptor,
+                                      executionStore: ExecutionStore,
+                                      outputStore: OutputStore) extends WdlLookup {
+
+  override val expressionLanguageFunctions = new EngineELF(workflowDescriptor.backendDescriptor.workflowOptions)
+
+  def jobExecutionSuccess(jobKey: JobKey,
+                          outputs: CallOutputs) = this.copy(
```
scottfrazer: these parameters don't need to be on different lines

engine/src/main/scala/cromwell/engine/EngineELF.scala
scottfrazer: My vote is for **not** EngineELF... all other suggestions seem fine to me!

mcovarr: EngineOrc

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/JobPreparationActor.scala
```diff
@@ -0,0 +1,137 @@
...
+
+  when(EvaluatingInputs) {
+    case Event(EvaluatedInputs(inputs), _) =>
+      val jobDescriptor = BackendJobDescriptor(workflowDescriptor.backendDescriptor, jobKey, inputs)
+      val config = BackendConfigurationDescriptor(configDescriptor.backendConfig, configDescriptor.globalConfig)
+
+      Future(factory.jobExecutionActorProps(jobDescriptor, config)) onComplete {
+        case Success(props) =>
+          context.parent ! BackendJobPreparationSucceeded(jobDescriptor, props)
+          context stop self
```
scottfrazer: I can't remember... but is there some issue with using `context ` within a `onComplete` handler?  @mcovarr?

mcovarr: `context` is fine, `sender()` is the landmine.  😄

geoffjentry: OTOH having Futures inside an actor is also a landmine

geoffjentry:
@Horneth While there are places we can't help it, I"d prefer we avoid creating Futures inside of actors, instead preferring worker actors for async behavior. That helps to preserve the actor model/abstraction and removes a huge potential pitfall

Horneth:
@geoffjentry That's what this particular actor is, a worker actor. The work inside was futurized so that the actor could change state after some work is started (as suggested by @mcovarr). It's either that or the actor itself blocks until the work is done (which is how it was originally). I honestly don't have a strong opinion either way. Unless there is a third option I'm missing.

scottfrazer:
👍

geoffjentry:
@horneth - so I see 2 Futures in here. One the little bit changing the state (around the Props) and the other was resolveAndEvaluate.

As background for my statement I'll say the following:

- We've already seen firsthand the havoc which can erupt from having Futures rolling around inside an Actor. They break the Actor Model's abstraction that the internals of an actor are single threaded, meaning you now have to reason about shared mutable state, etc. We *can* do that, but there are easier paths than actors to deal with that. We've been better about this recently but my concern is that it's too easy for stuff like that to sneak into what were previously pure Futures. Mixing Futures & Actors is not really a great idea.
- There are two async operations in the actor, which means that it is certainly doing two different things (I'll admit that the creation of an actor is a fairly lame 'thing'), disrupting Akka's mantra that actors should do one thing only

What I was suggesting was that the work being performed by these Futures be themselves pushed to their own actors. When they complete they can message back to this one, and those messages could be use to manage state transitions and such.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -0,0 +1,310 @@
...
+  private def processRunnableJob(jobKey: BackendJobDescriptorKey, data: WorkflowExecutionActorData): Try[WorkflowExecutionDiff] = {
+    workflowDescriptor.backendAssignments.get(jobKey.call) match {
+      case None =>
+        val message = s"Could not start call ${jobKey.tag} because it was not assigned a backend"
+        log.error(s"$tag $message")
+        throw new IllegalStateException(s"$tag $message")
+      case Some(backendName) =>
+        (configs.get(backendName), factories.get(backendName)) match {
+          case (Some(configDescriptor), Some(factory)) =>
+            val jobStarterActor = context.actorOf(JobPreparationActor.props(data, jobKey, factory, configDescriptor))
```
geoffjentry: I might be mis-parsing this line but it doesn't look like the actor is named

Horneth:
@geoffjentry I think I understand what you mean. From my perspective it's a matter of where to stop.
We could ban entirely `Future`s from the codebase and decide that every asynchronous task deserves its own actor. That seems a bit extreme to me though.
I see your point about future being dangerous inside of actors. However I believe that small actors with 2 states and 3 internal messages are a small enough unit to be understood well enough to avoid the kind of problems we encountered before. This actor doesn't even have state data, there's no state to be shared or mutated.

geoffjentry:
Futures are fine, just not in the code which the actor itself directly controls - e.g. if an actor calls a function that lives elsewhere in our code, and internally that's using a Future, that's fine as it can't directly manipulate the actor's state (sort of - there's still an issue that the DB is a giant bit of shared mutable state, but that's harder to get around).

I agree with your statement. My point is by making that claim that the burden is now on all future developers/reviewers/etc to notice if that actor somehow modifies beyond the "oh, it's okay for now" point. That's a lot easier said than done, better to prevent the pain from the get-go, it's not like the medicine is all that much extra work.

Horneth:
> Futures are fine, just not in the code which the actor itself directly controls

Hmm pretty much all our relevant code lives in actors so if the only place where we can use `Future`s is outside of actors, then I'd say `Future`'s days are counted xD

The *burden* of verifying that new code doesn't break anything is part of the review process anyway IMHO, and adding N more actors instead of `Future`s with twice as many more messages, states, transitions, and classes doesn't reduce the burden of checking that everything is wired correctly, as far as I'm concerned. 

Also, about futures being dangerous in actors because they can mutate state, this can only happen in a true `Actor` where your state pretty much has to be `var`s if you want to be able to mutate it, or in an `FSM` where you would also store some state as mutable `var`s inside the actor, instead of using the FSM data.
If you're using an FSM and all your mutable data is contained in the FSM data, then I don't see how creating futures would ever lead to mutating your state (and by state I mean mutable data, not FSM state) asynchronously. The FSM data is only updatable when the actor receives a message and decides to change state or stay in the same. A future `onComplete` could never force the FSM to mutate its data, which is why all we do is always send a message to someone when the future completes. And I don't see what difference it makes, from an actor perspective, if a message comes from a `Future` or another actor.

mcovarr:
Remember, @Horneth's first thought is always correct.  😄 

I think the code is fine as it is, plus the author gets the edge anyway.  If there gets to be more complexity here this could certainly be refactored to use a separate actor, but I don't think a separate actor is warranted given the current situation.

geoffjentry:
@mcovarr And originally @Horneth had one fewer Future, listen to your own (well, my own) advice ;)

Ok - so now that I have a chance to look at this more closely, it's unclear why there are any futures at all going on here. If I'm reading things right (as always, a big if) only one thing ever messages it (once) and then waits for a response. That response is either a success or failure. Why not just do the stuff it needs to do in the event loop (as nothing should be messaging it anyways)? The one argument I can come up with is that this would tie tie up one of the actorsystem dispatcher's threads but that's just as easily handled by giving this actor class its own dispatcher - that gets you the same effect as putting the Future in the global EC without all of the state changing and such.

If you all want to make the claim that reasoning about the actor with futures kicking around inside is easier to reason about than multiple actors (a claim I vehemently disagree with), I'd put forth that the futures/states are themselves far more difficult to reason about than just simply doing the work straight up considering how simple this is.

mcovarr:
Hmm yeah if we bulkhead the thread pool for this actor class, blocking might be the simplest thing to do.

geoffjentry:
@mcovarr I'm always super suspicious that I'm handwaving away some complexity I don't spot but from what I saw the effect would be the same but with fewer moving parts.

PR #822, opened 2016-05-12
Engine pushes to the MetadataService closes #668

### LINKED ISSUE #668: Update Engine to call Metadata Service API

gauravs90:
Okay. So initially, I was passing the `ServiceRegistryActor` reference via props down the chain of actors, and each of the actors needed to follow a pattern of steps to insert into the metadata (e.g. create a `MetadataEvent` with a `MetadataPutAction` within which the `Metadata[Key, Value]` resided, send this message over to the the service registry, and handle failures, if any. It did not look good by any stretch IMO. All the metadata generating entities needed to be aware of the `MetadataService` (the erstwhile dataAcess)

Currently, what I've done below is create a single instance of the `ServiceRegistryActor` in the `WorkflowManagerActor`and then create a `WorkflowProfilerActor` (one per workflow) which is supposed to handle all the metadata information coming from the engine side for a particular workflow. The way it happens is based on the presumption that almost all the information that we needed was present in the `StateName` and `StateData` of our FSMs. Unfortunately, with Akka's `SubscribeTransitionalCallback` we can only monitor the FSM states, and not the data. So I've created a trait (which the Engine's FSMs can extend from) which provide the semantics of wrapping up the state and data of the FSM in a message, and publish it into Akka's event stream. The ProfilerActor is the listener of these events and handles them appropriately. With this, I was able to make the FSMs unaware of the MetadataServices, and simply publish it's state and data in the event stream while performing any transitions. 

Let me know if you guys have any (other?) ideas / suggestions.

Edit: Names of new actors might be pretty bad IMO. Please suggest better ones if you have any.

*P.S. : Currently based out of the Chris's branch since his metadata changes were needed.*
*P.P.S. : Still a WIP for improving some stuff.*


engine/src/main/scala/cromwell/services/ServiceRegistryActor.scala
```diff
@@ -40,7 +40,7 @@ case class ServiceRegistryActor(globalConfig: Config) extends Actor {
   import ServiceRegistryActor._
 
   val services: Map[String, ActorRef] = serviceNameToPropsMap(globalConfig) map {
-    case (name, props) => name -> context.actorOf(props, "ServiceRegistryActor")
```
gauravs90: This did not handle multiple services, as all were getting instantiated with the same name. This is one of the WIP stuff, I'll name each of the actors according to their name as defined in the config.

engine/src/main/scala/cromwell/engine/workflow/profiler/WorkflowProfiler.scala
```diff
@@ -0,0 +1,94 @@
...
+          // Add the end time of the workflow in the MetadataService
+          val metadataEventMsg = MetadataEvent(MetadataKey(workflowId, None, WorkflowMetadataKeys.EndTime), MetadataValue(DateTime.now.toString))
+          serviceRegistryActor.publishToMetadataService(metadataEventMsg)
+          // As the state is terminal, this is expected to be the last message related to a particular Workflow. This actor's job here seems to be done
+          self ! PoisonPill
+        }
+    }
+  }
+
+  private def handleMessageFromWorkflowExecutionActor(workflowExecutionActorStateAndData: WorkflowExecutionActorStateAndData) = Unit
```
gauravs90: Will be adding stuff here to add the outputs to the metadata.

engine/src/main/scala/cromwell/engine/workflow/profiler/WorkflowProfiler.scala
```diff
@@ -0,0 +1,94 @@
...
+  private val tag = s"WorkflowProfiler-${workflowId.toString}"
+  implicit val actorSystem = context.system
+  implicit val ec = context.dispatcher
+
+  override def preStart(): Unit = {
+    context.system.eventStream.subscribe(self, classOf[CromwellFsmStateAndData])
+  }
+
+  override def receive = LoggingReceive {
+    case workflowManagerActorMessage: WorkflowManagerMetadataMessage => handleMessageFromWorkflowManagerActor(workflowManagerActorMessage)
```
gauravs90: I might remove this message I think, and simply add the required info (which really is just the submission time in the constructor)

engine/src/main/scala/cromwell/engine/workflow/profiler/WorkflowProfiler.scala
```diff
@@ -0,0 +1,95 @@
...
+        serviceRegistryActor.publishToMetadataService(metadataEventMsg)
+      case WorkflowActorStateAndData(state, data) =>
+        // This updates the workflow status
+        publishCurrentStateToMetadataService(state.workflowState)
+        if (state.terminal) {
+          // Add the end time of the workflow in the MetadataService
+          val metadataEventMsg = MetadataEvent(MetadataKey(workflowId, None, WorkflowMetadataKeys.EndTime), MetadataValue(DateTime.now.toString))
+          serviceRegistryActor.publishToMetadataService(metadataEventMsg)
+          // As the state is terminal, this is expected to be the last message related to a particular Workflow. This actor's job here seems to be done
+          self ! PoisonPill
```
gauravs90: This mechanism to kill itself seems a little fishy to me. It should wait for in-flight updates happening to metadata, isn't it? 😕

cjllanwarne:
@gauravs90 I just merged my branch so this probably needs to be rebased on develop

geoffjentry:
@gauravs90 I've generally seen the akka folks recommend directly passing references to actors which need to be used. That has multiple benefits (e.g. makes it easy to switch out and/or dep injection, etc). My off the cuff reaction is that that seems simpler to just pass the required reference around, although I'll admit I'm basing that purely on your description and not having looked at the changes yet

PR #823, opened 2016-05-12
configure restarting workflow closes #755

### LINKED ISSUE #755: Make restarting old workflows configurable
When Cromwell comes online it detects if there are any unfinished workflows in the system and will pick those back up.

ruchim:
Made a small change to the preStart method of WFMA, it only tries to re-start incomplete workflows when the config "workflow-restart" is set to true. I don't know if I loved how I named that field, and I'm open to suggestions. 

Since we don't have restarts configured right now, I didn't have a chance to test this change. I didn't update the Readme yet either, since this is not a change users can use yet, I'm realizing now we do need some way to keep track of all the documentation changes we'll need to make.

### DIFF
--- a/engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
@@ -78,9 +78,11 @@ class WorkflowManagerActor(config: Config)
 
   private val donePromise = Promise[Unit]()
 
+  private val conf = ConfigFactory.load
+
   override def preStart() {
     addShutdownHook()
-    restartIncompleteWorkflows()
+    if (conf.getBoolean("system.workflow-restart")) { restartIncompleteWorkflows() }
   }
 
   private def addShutdownHook(): Unit = {
--- a/engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
@@ -78,11 +78,9 @@ class WorkflowManagerActor(config: Config)
 
   private val donePromise = Promise[Unit]()
 
-  private val conf = ConfigFactory.load
-
   override def preStart() {
     addShutdownHook()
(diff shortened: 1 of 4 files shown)


ruchim:
@geoffjentry @scottfrazer

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -102,9 +102,11 @@ class WorkflowManagerActor(config: Config)
 
   private val donePromise = Promise[Unit]()
 
+  private val conf = ConfigFactory.load
+
   override def preStart() {
     addShutdownHook()
-    restartIncompleteWorkflows()
+    if (conf.getBoolean("system.workflow-restart")) { restartIncompleteWorkflows() }
```
scottfrazer: Perhaps this can be something like `if (server_mode && conf.getBoolean("system.workflow-restart"))`

scottfrazer: That's a bit of feature creep though

geoffjentry: I've always viewed the 'run' mode picking up old workflows as bug, not feature (and thus would agree w/ your suggestion) although in the back of my mind I remember someone making a strong argument the other way so I don't think that's a universally held opinion.

geoffjentry:
@ruchim 

I think it's fine to update the README even though it's not usable - the develop branch is known to be unstable at the moment, and isn't the default branch on github.

👍

scottfrazer:
👍 I thought Kristian mentioned this is important for Firecloud... does this also need to go into hotfix?

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
ruchim: I'll just bring it up after standup and we can hash it out. And yes @scottfrazer I'll start making the change to 0.19_hotfix.

PR #824, opened 2016-05-12
How can I tell if I'm having fun?  Closes #803

### LINKED ISSUE #803: "Read" Workflow Level Engine Functions (GCS Filesystem)
Support reading a GCS file from workflow-level expression:

```
workflow w {
  String x = read_string("gs://my-bucket/some-object.txt")
}
```

scottfrazer:
This allows this workflow to run:

```
task t {
  Array[String] s
  command {echo ${sep=',' s}}
  output {String o = read_string(stdout())}
}

workflow w {
  Array[String] x = read_lines("gs://sfrazer-dev/array.txt")
  call t {input: s=x}
}

```

So this is the minimum implementation I could come up with... looking for design feedback too, this could be simple or become full-blown pluggable filesystems

### DIFF
--- a/engine/src/main/resources/application.conf
+++ b/engine/src/main/resources/application.conf
@@ -96,6 +96,14 @@ google {
   ]
 }
 
+engine {
+  filesystems {
+    gcs {
+      auth = "application-default"
+    }
+  }
+}
+
 backend {
   default = "Local"
   providers {
--- a/engine/src/main/scala/cromwell/engine/WdlFunctions.scala
+++ b/engine/src/main/scala/cromwell/engine/WdlFunctions.scala
@@ -2,21 +2,31 @@ package cromwell.engine
 
 import java.nio.file.{FileSystem, FileSystems}
 
+import com.typesafe.config.ConfigFactory
 import cromwell.backend.wdl.{PureFunctions, ReadLikeFunctions}
 import cromwell.core.WorkflowOptions
+import cromwell.engine.backend.EnhancedWorkflowOptions._
+import cromwell.filesystems.gcs.{GcsFileSystem, GcsFileSystemProvider, GoogleConfiguration}
+import lenthall.config.ScalaConfig._
+import spray.json.JsObject
 import wdl4s.expression.WdlStandardLibraryFunctions
 import wdl4s.values.{WdlFile, WdlValue}
 
 import scala.util.{Failure, Try}
 
 class WdlFunctions(workflowOptions: WorkflowOptions) extends WdlStandardLibraryFunctions with ReadLikeFunctions with PureFunctions {
+  def gcsFileSystem: Option[GcsFileSystem] = {
+    for {
+      authModeString <- ConfigFactory.load.getStringOption("engine.filesystems.gcs.auth")
+      authMode <- GoogleConfiguration.Instance.auth(authModeString).toOption
+      fs = GcsFileSystem(GcsFileSystemProvider(authMode.buildStorage(workflowOptions.toGoogleAuthOptions)))
+    } yield fs
+  }
 
-  // TODO: GCS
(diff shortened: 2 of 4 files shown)


engine/src/main/scala/cromwell/engine/EngineELF.scala
```diff
@@ -2,21 +2,31 @@ package cromwell.engine
...
 import wdl4s.expression.WdlStandardLibraryFunctions
 import wdl4s.values.{WdlFile, WdlValue}
 
 import scala.util.{Failure, Try}
 
 class EngineELF(workflowOptions: WorkflowOptions) extends WdlStandardLibraryFunctions with ReadLikeFunctions with PureFunctions {
+  def gcsFileSystem: Option[GcsFileSystem] = {
+    for {
+      authModeString <- ConfigFactory.load.getStringOption("engine.filesystems.gcs.auth")
+      authMode <- GoogleConfiguration.Instance.auth(authModeString).toOption
```
Horneth: Maybe if this fails, we should throw an exception ? If you don't specify gcs in your config then that's fine, but if you do and we can't build a FS from it, we should probably not silently ignore it ?

Horneth:
👍

scottfrazer:
Wheel chose @Horneth and @cjllanwarne

engine/src/main/scala/cromwell/engine/EngineELF.scala
scottfrazer: This is why I was complaining about logging the other day... what I really wanted to do was log a warning and then continue on.  My reasoning for semi-silently ignoring this is because somebody might be using only the local backend and the config by default uses `application-default` credentials, but they might not have any GCS stuff setup at all, but that shouldn't necessarily stop the workflow from running.  Up for debate thought, just one idea.

cjllanwarne: Unrelated comment: Can you not use `authMode =` and not have the `toOption`

scottfrazer: Maybe I'm not understanding... `GoogleConfiguration.Instance.auth(authModeString)` is an `ErrorOr`

cjllanwarne: Oh! Ignore that then!

cjllanwarne: My comment on the above would be, I'd probably agree with Thibault and avoid allowing invalid configuration, even if that section isn't necessarily used.

But... that's not a strong opinion and it's probably more a PO question <= @kcibul

cjllanwarne:
Maybe the default application.conf should start with Local FS instead of GCS by default - since we want it to be a pick-up-and-play config?

cjllanwarne:
Other than that, 👍

engine/src/main/scala/cromwell/engine/EngineELF.scala
scottfrazer: Oh yeah, I agree that we should not allow *invalid* config... sorry I misunderstood.  Like if they specify `application-foobar` then yeah, some kind of exception should be thrown... let me see what I can do about that.

scottfrazer:
@cjllanwarne yeah so I think you're right that we do want that functionality.  I'm just not sure if that belongs here or not... I heard pluggable filesystems are coming and if I implemented something just short of pluggable filesystems then that code would be gutted anyway (just like most of this code will be) when pluggable filesystems come along.  So this PR was really in the spirit of "everything's going to change anyway, so don't bother making it really good, just do what the ticket says".  I'll definitely make it really good if I'm not going to be wasting my time.... anybody else have thoughts?  @kcibul?

PR #829, opened 2016-05-12
Engine pushes to the MetadataService closes #668

### LINKED ISSUE #668: Update Engine to call Metadata Service API

gauravs90:
Okay. So initially, I was passing the `ServiceRegistryActor` reference via props down the chain of actors, and each of the actors needed to follow a pattern of steps to insert into the metadata (e.g. create a `MetadataEvent` with a `MetadataPutAction` within which the `Metadata[Key, Value]` resided, send this message over to the the service registry, and handle failures, if any. It did not look good by any stretch IMO. All the metadata generating entities needed to be aware of the `MetadataService` (the erstwhile dataAcess)

**Edit:** Using the above design in-fact now.

~~Currently, what I've done below is create a single instance of the `ServiceRegistryActor` in the `WorkflowManagerActor`and then create a `WorkflowProfilerActor` (one per workflow) which is supposed to handle all the metadata information coming from the engine side for a particular workflow. The way it happens is based on the presumption that almost all the information that we needed was present in the `StateName` and `StateData` of our FSMs. Unfortunately, with Akka's `SubscribeTransitionalCallback` we can only monitor the FSM states, and not the data. So I've created a trait (which the Engine's FSMs can extend from) which provide the semantics of wrapping up the state and data of the FSM in a message, and publish it into Akka's event stream. The ProfilerActor is the listener of these events and handles them appropriately. With this, I was able to make the FSMs unaware of the MetadataServices, and simply publish it's state and data in the event stream while performing any transitions.~~

~~Let me know if you guys have any (other?) ideas / suggestions.~~

Contents added to metadata with this PR:

- [x] workflowName
- [ ]  calls (To come from the backends)
- [x] outputs 
- [x] id
- [x] inputs
- [x] submission
- [x] status
- [x] end
- [x] start

### DIFF
--- a/engine/src/main/scala/cromwell/engine/EngineWorkflowDescriptor.scala
+++ b/engine/src/main/scala/cromwell/engine/EngineWorkflowDescriptor.scala
@@ -1,0 +1,13 @@
+package cromwell.engine
+
+import cromwell.backend.BackendWorkflowDescriptor
+import wdl4s._
+
+final case class EngineWorkflowDescriptor(backendDescriptor: BackendWorkflowDescriptor,
+                                          declarations: WorkflowCoercedInputs,
+                                          backendAssignments: Map[Call, String],
+                                          failureMode: WorkflowFailureMode) {
+  def id = backendDescriptor.id
+  def namespace = backendDescriptor.workflowNamespace
+  def name = namespace.workflow.unqualifiedName
+}
--- a/engine/src/main/scala/cromwell/engine/package.scala
+++ b/engine/src/main/scala/cromwell/engine/package.scala
@@ -81,11 +81,4 @@ package object engine {
     }
   }
 
-  final case class EngineWorkflowDescriptor(backendDescriptor: BackendWorkflowDescriptor,
-                                            declarations: WorkflowCoercedInputs,
-                                            backendAssignments: Map[Call, String],
-                                            failureMode: WorkflowFailureMode) {
-    def id = backendDescriptor.id
-    def namespace = backendDescriptor.workflowNamespace
-  }
 }
--- a/engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
@@ -4,6 +4,7 @@ import akka.actor.SupervisorStrategy.Escalate
 import akka.actor._
 import com.typesafe.config.Config
 import cromwell.core.WorkflowId
+import cromwell.engine._
 import cromwell.engine.workflow.WorkflowActor._
 import cromwell.engine.workflow.lifecycle.MaterializeWorkflowDescriptorActor.{MaterializeWorkflowDescriptorCommand, MaterializeWorkflowDescriptorFailureResponse, MaterializeWorkflowDescriptorSuccessResponse}
 import cromwell.engine.workflow.lifecycle.WorkflowFinalizationActor.{StartFinalizationCommand, WorkflowFinalizationFailedResponse, WorkflowFinalizationSucceededResponse}
@@ -11,7 +12,8 @@ import cromwell.engine.workflow.lifecycle.WorkflowInitializationActor.{StartInit
 import cromwell.engine.workflow.lifecycle._
 import cromwell.engine.workflow.lifecycle.execution.WorkflowExecutionActor
 import cromwell.engine.workflow.lifecycle.execution.WorkflowExecutionActor.{RestartExecutingWorkflowCommand, StartExecutingWorkflowCommand, WorkflowExecutionFailedResponse, WorkflowExecutionSucceededResponse}
-import cromwell.engine.{EngineWorkflowDescriptor, WorkflowSourceFiles}
+import cromwell.services.MetadataServiceActor._
+import org.joda.time.DateTime
 
 import scala.language.postfixOps
 
@@ -35,54 +37,69 @@ object WorkflowActor {
   /**
     * States for the WorkflowActor FSM
     */
-  sealed trait WorkflowActorState { def terminal = false }
-  sealed trait WorkflowActorTerminalState extends WorkflowActorState { override def terminal = true}
+  sealed trait WorkflowActorState {
+    def terminal = false
+    // Each state in the FSM maps to a state in the `WorkflowState` which is used for Metadata reporting purposes
+    def workflowState: WorkflowState
+  }
+  sealed trait WorkflowActorTerminalState extends WorkflowActorState { override val terminal = true }
+  sealed trait WorkflowActorRunningState extends WorkflowActorState { override val workflowState = WorkflowRunning }
 
   /**
     * Waiting for a Start or Restart command.
     */
-  case object WorkflowUnstartedState extends WorkflowActorState
+  case object WorkflowUnstartedState extends WorkflowActorState {
+    override val workflowState = WorkflowSubmitted
+  }
 
   /**
     * The WorkflowActor is created. It now needs to validate and create its WorkflowDescriptor
     */
-  case object MaterializingWorkflowDescriptorState extends WorkflowActorState
+  case object MaterializingWorkflowDescriptorState extends WorkflowActorRunningState
 
   /**
     * The WorkflowActor has a descriptor. It now needs to create its set of WorkflowBackendActors
     */
-  case object InitializingWorkflowState extends WorkflowActorState
+  case object InitializingWorkflowState extends WorkflowActorRunningState
 
   /**
     * The WorkflowActor has a descriptor and a set of backends. It now needs to execute the hell out of its jobs.
     */
-  case object ExecutingWorkflowState extends WorkflowActorState
+  case object ExecutingWorkflowState extends WorkflowActorRunningState
 
   /**
     * The WorkflowActor has completed. So we're now finalizing whatever needs to be finalized on the backends
     */
-  case object FinalizingWorkflowState extends WorkflowActorState
+  case object FinalizingWorkflowState extends WorkflowActorRunningState
 
   /**
     * The WorkflowActor is aborting. We're just waiting for everything to finish up then we'll respond back to the
     * manager.
     */
-  case object AbortingWorkflowState extends WorkflowActorState
+  case object AbortingWorkflowState extends WorkflowActorState {
+    override val workflowState = WorkflowAborting
+  }
 
   /**
     * We're done.
     */
-  case object WorkflowSucceededState extends WorkflowActorTerminalState
+  case object WorkflowSucceededState extends WorkflowActorTerminalState {
+    override val workflowState = WorkflowSucceeded
+  }
 
   /**
     * We're done. But we were aborted in the middle of the lifecycle.
     */
-  case object WorkflowAbortedState extends WorkflowActorTerminalState
+  case object WorkflowAbortedState extends WorkflowActorTerminalState {
+    override val workflowState = WorkflowAborted
+  }
 
   /**
     * We're done. But the workflow has failed.
     */
-  case object WorkflowFailedState extends WorkflowActorTerminalState
+  case object WorkflowFailedState extends WorkflowActorTerminalState {
+    override val workflowState = WorkflowFailed
+  }
 
   /**
     * @param currentLifecycleStateActor The current lifecycle stage, represented by an ActorRef.
@@ -101,7 +118,11 @@ object WorkflowActor {
   case object StartNewWorkflow extends StartMode
   case object RestartExistingWorkflow extends StartMode
 
-  def props(workflowId: WorkflowId, startMode: StartMode, wdlSource: WorkflowSourceFiles, conf: Config): Props = Props(new WorkflowActor(workflowId, startMode, wdlSource, conf))
+  def props(workflowId: WorkflowId,
+            startMode: StartMode,
+            wdlSource: WorkflowSourceFiles,
+            conf: Config,
+            serviceRegistryActor: ActorRef): Props = Props(new WorkflowActor(workflowId, startMode, wdlSource, conf, serviceRegistryActor))
 }
 
 /**
@@ -110,9 +131,13 @@ object WorkflowActor {
 class WorkflowActor(workflowId: WorkflowId,
                     startMode: StartMode,
                     workflowSources: WorkflowSourceFiles,
-                    conf: Config) extends LoggingFSM[WorkflowActorState, WorkflowActorData] with ActorLogging{
+                    conf: Config,
+                    serviceRegistryActor: ActorRef) extends LoggingFSM[WorkflowActorState, WorkflowActorData] with ActorLogging {
 
   val tag = self.path.name
+
+  implicit val actorSystem = context.system
+  implicit val ec = context.dispatcher
 
   startWith(WorkflowUnstartedState, WorkflowActorData.empty)
 
@@ -141,7 +166,10 @@ class WorkflowActor(workflowId: WorkflowId,
 
   when(InitializingWorkflowState) {
     case Event(WorkflowInitializationSucceededResponse, WorkflowActorData(_, Some(workflowDescriptor))) =>
-      val executionActor = context.actorOf(WorkflowExecutionActor.props(workflowId, workflowDescriptor), name = s"WorkflowExecutionActor-$workflowId")
+      // Add Workflow name and inputs to the metadata
+      pushWfNameAndInputsToMetadataService(workflowDescriptor)
+
+      val executionActor = context.actorOf(WorkflowExecutionActor.props(workflowId, workflowDescriptor, serviceRegistryActor), name = s"WorkflowExecutionActor-$workflowId")
       val commandToSend = startMode match {
         case StartNewWorkflow => StartExecutingWorkflowCommand
         case RestartExistingWorkflow => RestartExecutingWorkflowCommand
@@ -184,15 +212,12 @@ class WorkflowActor(workflowId: WorkflowId,
   when(WorkflowFailedState) { FSM.NullFunction }
   when(WorkflowSucceededState) { FSM.NullFunction }
 
-  onTransition {
-    case oldState -> terminalState if terminalState.terminal =>
-      log.info(s"$tag transition from $oldState to $terminalState: shutting down")
-      context.stop(self)
-    case fromState -> toState =>
-      log.info(s"$tag transitioning from $fromState to $toState")
-  }
-
   whenUnhandled {
+    case Event(MetadataPutFailed(action, error), _) =>
+      // Do something useful here??
+      log.warning(s"$tag Put failed for Metadata action $action : ${error.getMessage}")
+      stay
+    case Event(MetadataPutAcknowledgement(_), _) => stay()
     case Event(AbortWorkflowCommand, WorkflowActorData(Some(actor), _)) =>
       actor ! EngineLifecycleActorAbortCommand
       goto(AbortingWorkflowState)
@@ -200,4 +225,43 @@ class WorkflowActor(workflowId: WorkflowId,
       log.warning(s"$tag received an unhandled message $unhandledMessage in state $stateName")
       stay
   }
+
+  onTransition {
+    case fromState -> toState =>
+      log.info(s"$tag transitioning from $fromState to $toState")
+      // This updates the workflow status
+      pushCurrentStateToMetadataService(toState.workflowState)
+  }
+
+  onTransition {
+    case oldState -> terminalState if terminalState.terminal =>
+      log.info(s"$tag transition from $oldState to $terminalState: shutting down")
+      // Add the end time of the workflow in the MetadataService
+      val metadataEventMsg = MetadataEvent(MetadataKey(workflowId, None, WorkflowMetadataKeys.EndTime), MetadataValue(DateTime.now.toString))
+      serviceRegistryActor ! PutMetadataAction(metadataEventMsg)
+      terminalState match {
+        case WorkflowSucceededState =>
+          context.parent ! WorkflowSucceededResponse(workflowId)
+        case WorkflowFailedState =>
+          context.parent ! WorkflowFailedResponse(workflowId, oldState, Seq.empty[Throwable])
+        case WorkflowAbortedState =>
+          context.parent ! WorkflowAbortedResponse(workflowId)
+        case unknownState => log.warning(s"$tag $unknownState is an unhandled terminal state!")
+      }
+  }
+
+  private def pushWfNameAndInputsToMetadataService(workflowDescriptor: EngineWorkflowDescriptor): Unit = {
+    val inputMetadataEvents = workflowDescriptor.backendDescriptor.inputs.map { case (k, v) =>
+      MetadataEvent(MetadataKey(workflowId, None, s"${WorkflowMetadataKeys.Inputs}:$k"), MetadataValue(v.toWdlString))
+    }
+    val metadataEventMsgs = List(MetadataEvent(MetadataKey(workflowId, None, WorkflowMetadataKeys.Name), MetadataValue(workflowDescriptor.name))) ++ inputMetadataEvents
+    metadataEventMsgs foreach ( serviceRegistryActor ! PutMetadataAction(_) )
+  }
+
+  // Update the current State of the Workflow (corresponding to the FSM state) in the Metadata service
+  private def pushCurrentStateToMetadataService(workflowState: WorkflowState): Unit = {
+    val metadataEventMsg = MetadataEvent(MetadataKey(workflowId, None, WorkflowMetadataKeys.Status), MetadataValue(workflowState.toString))
+    serviceRegistryActor ! PutMetadataAction(metadataEventMsg)
+  }
+
 }
--- a/engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
@@ -10,8 +10,11 @@ import cromwell.engine.backend._
 import cromwell.engine.db.DataAccess._
 import cromwell.engine.workflow.WorkflowActor._
 import cromwell.engine.workflow.WorkflowManagerActor.{AbortWorkflowCommand, _}
+import cromwell.services.MetadataServiceActor._
+import cromwell.services.ServiceRegistryActor
 import cromwell.webservice.CromwellApiHandler._
 import lenthall.config.ScalaConfig.EnhancedScalaConfig
+import org.joda.time.DateTime
 
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.duration._
@@ -78,9 +81,11 @@ class WorkflowManagerActor(config: Config)
 
   private val donePromise = Promise[Unit]()
 
+  val serviceRegistryActor = context.actorOf(ServiceRegistryActor.props(config), "ServiceRegistoryActor")
+
   override def preStart() {
     addShutdownHook()
-    if (config.getBoolean("system.workflow-restart")) { restartIncompleteWorkflows() }
+    restartIncompleteWorkflows()
   }
 
   private def addShutdownHook(): Unit = {
@@ -164,6 +169,10 @@ class WorkflowManagerActor(config: Config)
   when (Done) { FSM.NullFunction }
 
   whenUnhandled {
+    case Event(MetadataPutFailed(action, error), _) =>
+      log.warning(s"$tag Put failed for Metadata action $action : ${error.getMessage}")
+      stay()
+    case Event(MetadataPutAcknowledgement(_), _) => stay()
     // Uninteresting transition and current state notifications.
     case Event((Transition(_, _, _) | CurrentState(_, _)), _) => stay()
     // Anything else certainly IS interesting:
@@ -180,6 +189,16 @@ class WorkflowManagerActor(config: Config)
       logger.info(s"$tag transitioning from $fromState to $toState")
   }
 
+  private def pushToMetadataService(workflowId: WorkflowId): Unit = {
+    val curTime = DateTime.now.toString
+    val metadataEventMsgs = List(
+      MetadataEvent(MetadataKey(workflowId, None, WorkflowMetadataKeys.Id), MetadataValue(workflowId.toString)),
+      MetadataEvent(MetadataKey(workflowId, None, WorkflowMetadataKeys.SubmissionTime), MetadataValue(curTime)),
+      // Currently, submission time is the same as start time
+      MetadataEvent(MetadataKey(workflowId, None, WorkflowMetadataKeys.StartTime), MetadataValue(curTime))
+    )
+    metadataEventMsgs foreach (serviceRegistryActor ! PutMetadataAction(_))
+  }
 
   /** Submit the workflow and return an updated copy of the state data reflecting the addition of a
     * Workflow ID -> WorkflowActorRef entry.
@@ -190,7 +209,11 @@ class WorkflowManagerActor(config: Config)
     val isRestart = id.isDefined
 
     val startMode = if (isRestart) RestartExistingWorkflow else StartNewWorkflow
-    val wfActor = context.actorOf(WorkflowActor.props(workflowId, startMode, source, config), name = s"WorkflowActor-$workflowId")
+    val wfActor = context.actorOf(WorkflowActor.props(workflowId, startMode, source, config, serviceRegistryActor), name = s"WorkflowActor-$workflowId")
+
+    // We have a valid workflowId for the workflow, send it over to the metadata service
+    pushToMetadataService(workflowId)
+
     replyTo.foreach { _ ! WorkflowManagerSubmitSuccess(id = workflowId) }
     wfActor ! SubscribeTransitionCallBack(self)
     wfActor ! StartWorkflowCommand
--- a/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
@@ -1,6 +1,6 @@
 package cromwell.engine.workflow.lifecycle.execution
 
-import akka.actor.{FSM, LoggingFSM, Props}
+import akka.actor.{ActorRef, FSM, LoggingFSM, Props}
 import com.typesafe.config.ConfigFactory
 import cromwell.backend.BackendJobExecutionActor._
 import cromwell.backend.{BackendJobDescriptor, BackendJobDescriptorKey, JobKey}
@@ -10,7 +10,8 @@ import cromwell.engine.ExecutionStatus.NotStarted
 import cromwell.engine.backend.{BackendConfiguration, CromwellBackends}
 import cromwell.engine.workflow.lifecycle.execution.JobPreparationActor.{BackendJobPreparationFailed, BackendJobPreparationSucceeded}
 import cromwell.engine.workflow.lifecycle.execution.WorkflowExecutionActor.WorkflowExecutionActorState
-import cromwell.engine.{EngineWorkflowDescriptor, ExecutionStatus}
+import cromwell.engine.{EngineWorkflowDescriptor, ExecutionStatus, workflow}
+import cromwell.services.MetadataServiceActor._
 import lenthall.exception.ThrowableAggregation
 import wdl4s._
 import wdl4s.util.TryUtil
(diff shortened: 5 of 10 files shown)


gauravs90:
I messed up some things I think while trying to change the base of that old PR to develop. Hence created this one out of develop.

gauravs90:
Quoting @geoffjentry from earlier comment so that it's not lost:

> @gauravs90 I've generally seen the akka folks recommend directly passing references to actors which need to be used. That has multiple benefits (e.g. makes it easy to switch out and/or dep injection, etc). My off the cuff reaction is that that seems simpler to just pass the required reference around, although I'll admit I'm basing that purely on your description and not having looked at the changes yet

gauravs90:
@geoffjentry I did implement that before, and after comparing it with this, the latter looked much better. But that's my opinion. I wouldn't mind to change anything that doesn't look right here.

geoffjentry:
@gauravs90 I don't doubt that passing the actorref around everywhere looked ugly. But in terms of moving parts and other such things my bet is that it's worth it. Your description of this scheme sounds a lot more complex (publishing, assuming everything going to metadata will always track w/ state change, having to shoehorn the data side in, etc) than simply passing the metadata down as is typically suggested.

gauravs90:
@geoffjentry I might have become a little too confusing with that description. Please take a look at the PR, and see if it looks' complicated. I'm very much opening to modifying the design to best suit the requirements here.

gauravs90:
Currently, the structure for outputs and inputs is flattened out (eg. output.worfklow.task.variable1, output.worfklow.task.variable2 etc.), not sure how the current metadata scheme supports nested json like structure, or if this is fine.

engine/src/main/scala/cromwell/engine/package.scala
```diff
@@ -87,5 +87,6 @@ package object engine {
                                             failureMode: WorkflowFailureMode) {
     def id = backendDescriptor.id
     def namespace = backendDescriptor.workflowNamespace
+    def name = namespace.workflow.unqualifiedName
```
geoffjentry: I realize that this class still isn't the worst offender in the package object but can you pull it out into its own thing? It looks like we need to clean out our package objects again.

geoffjentry:
@gauravs90 I'm still having trouble buying that this is less complicated than passing the desired actorRef in. Was it just verbose on the constructors or was there actually some other problem with it?

gauravs90:
@geoffjentry Verbosity is just a small part of that. Making different parts of the code aware of the each other, where avoidable, doesn't seem like a very good idea. If I'm not mistaken, what you're saying (and what I initially implemented) is something like the erstwhile `DataAccess` (which would be the `ServiceRegistoryActor` in the current world). With this, each step (actor) of the workflow had a reference to it and pushed to it independently. Any change to the dataAccess might have required changes to all the classes which were accessing it. Alternatively, if there was just one entity which handled the responsibility of collecting the metadata, by sniffing around the actors without their knowledge, and then pushed to the database, we need only change this entity for any modifications if there were to happen to the data access stuff.

I'll try to explain with a very simple (and silly) analogy: (Honestly, couldn't come up with anything better.)
Consider a ginormous Octopus (= `ServiceRegistry`) with a black ink on the tips of it's tentacles, with each of it's legs touching upon different rooms (= classes) in a house. If someday we decide to replace that octopus with something else, we'll be needing to wipe that ink from all the rooms upon which it was standing. On the other hand, if it were to sit and cuddle up just in a single room, there's simply less and comparatively easy work to do to wipe that up.

It's simply the same idea here. The Metadata producing entities in the engine can just go about minding their own business, while a third party (those classes with some weird names currently [CromwellProfilerFsm and WorkflowProfilerActor]) handle what they are meant to do: Profile a given workflow execution. (all the while without explicitly telling those execution engine entities that it's reading it's state and data information)

If the intentions are still not clear, let's talk about it tomorrow in the meeting to make progress with this?

Horneth:
Hey, so I'm going to rephrase to see if I understand what's happening.
Actors (`WorkflowExecutionActor`, `WorkflowActor`, `BackendActor`s ...), when mixing in this `CromwellProfilerFsm` trait, publish their state and data to the world via the akka event stream, on every state transition. Another actor, akka (pun intended) `TheProfiler`, reads those messages, and depending on where they come from, creates custom MetadataEvents, which it pushes to the metadata service. Does that sound close from what you're doing ?

geoffjentry:
@gauravs90 To be clear my primary objection is the assumption will always be tied to state changes and/or all actors involved will be FSM.

The analogy to DataAccess isn't the same - the issue with DataAccess was that you were handing the keys to shared mutable state (the DB) to async processes. In this case the metadata service was designed to not work like that. 

In my mind there already *is* an abstraction for metadata, the metadata service. While I'm not opposed to putting something between the engine and it, I still fall back to the point that when the akka people talk about setting up your system so that one actor can share info with another they tend to recommend passing the reference directly and it's only particular situations where things start listening on eventstreams and such. So even if the proposal was to have an intermediary I'd still say to pass the actorref around ... and I still don't see the value in that intermediary if one were to do away with the notion that metadata always equaled state changes.

Also I think it'll be far easier to transition *to* the direction you're talking about than *from*, if things get more complicated we can always do osmething like you're proposing. It's fully in the engine, it's trivial to change.

Horneth:
Assuming I'm not too far from the reality, my opinion on this is that 
- I like that the actors don't have to know about the actual reference of the Metadata service. Meaning I'm ok with the push to the event stream approach.
- I like less that metadata is pushed only when actors change state. Their data can be changed (and is changed) when they receive a message but don't necessarily transition. And if I'm understanding this correctly this is not enough to capture that.
For example, because the `WorkflowExecutionActor` which are basically `NotRunning`, `Running`, `Done`, we can only update metadata between those states. So there's no update of call status in real-time, outputs etc... Now from your comment I understand you want to do this by having the backend actors publishing metadata, but 
1) nothing guarantees that they will, or on which format
2) there are state that could be only known by then engine and that we might want to publish (`JobInitializing`, `Finalizing` etc...). Even more generally, I think the engine is the ultimate decider of what the state of a call is in the workflow, and relying on the backend to get this information seems a bit wrong.

cjllanwarne:
@gauravs90 I'm slightly confused here!

The model I thought we agreed on was:
- Everything has access to the `ServiceRegistry`
- Anyone (including the Engine) can send a `MetadataPut` message to the `ServiceRegistry`, which will forward it automagically to the `MetadataService`.
- ... That's it!

So, just send a bunch of `MetadataPut` messages to `ServiceRegistry` from the engine, and you're done, no need for all the extra stuff

gauravs90:
@cjllanwarne :
> Everything has access to the ServiceRegistry

This premise still holds. Why do you think otherwise?

>Anyone (including the Engine) can send a MetadataPut message to the ServiceRegistry, which will forward it automagically to the MetadataService.

This is how the magic is currently happening as well. The engine *is* sending MetadataPut messages to the ServiceRegistry.

gauravs90:
Hey @Horneth .  

For the first part, yes nothing guarantees what the backends might send (the intermediate statuses), but I think we all were fine with that right? The `Call` specific information is sent only by the backend.

For the second point, the engine doesn't know the state of a particular Call right? It just sends the Call (= Job) for execution, and waits for a completion. So anyways we don't know the intermediate status of a call. If you meant about WorkflowInitialization, Finalization etc., I am not adding those states currently because we didn't have those states earlier too. Currently, all those states map to `Running`, but that's certainly extensible too.

gauravs90:
Officially, @geoffjentry @Horneth. Others pleases feel free to chip in!

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -35,54 +38,69 @@ object WorkflowActor {
...
     * States for the WorkflowActor FSM
     */
-  sealed trait WorkflowActorState { def terminal = false }
-  sealed trait WorkflowActorTerminalState extends WorkflowActorState { override def terminal = true}
+  sealed trait WorkflowActorState {
+    def terminal = false
+    // Each state in the FSM maps to a state in the `WorkflowState` which is used for Metadata reporting purposes
+    def workflowState: WorkflowState
+  }
+  sealed trait WorkflowActorTerminalState extends WorkflowActorState { override def terminal = true }
```
geoffjentry: override val

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -184,20 +215,75 @@ class WorkflowActor(workflowId: WorkflowId,
...
+            case WorkflowFailedState =>
+              context.parent ! WorkflowFailedResponse(workflowId, oldState, Seq.empty[Throwable])
+            case WorkflowAbortedState =>
+              context.parent ! WorkflowAbortedResponse(workflowId)
+            case unknownState => log.warning(s"tag $unknownState is an unhandled terminal state!")
+          }
+      }
+  }
+
+  private def pushWfNameAndInputsToMetadataService(workflowDescriptor: EngineWorkflowDescriptor): Future[Unit] = {
```
geoffjentry: In general I'd prefer that we model async behavior inside an actor via sub-actors as much as possible instead of mixing abstractions and using Futures. 

Particularly with these longer functions as then everything needs to be inspected every time there's a change to make sure nothing is captured

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -184,20 +215,75 @@ class WorkflowActor(workflowId: WorkflowId,
...
+  }
+
+  onTransition {
+    case oldState -> terminalState if terminalState.terminal =>
+      log.info(s"$tag transition from $oldState to $terminalState: shutting down")
+      // Add the end time of the workflow in the MetadataService
+      val metadataEventMsg = MetadataEvent(MetadataKey(workflowId, None, WorkflowMetadataKeys.EndTime), MetadataValue(DateTime.now.toString))
+      serviceRegistryActor.publishToMetadataService(metadataEventMsg) map handleMetadataServiceResponse recover {
+        case reason: Throwable =>
+          self ! MetadataServiceFailureResponse(reason)
```
geoffjentry: as an example to my actor-not-future statement, I think you just closed over self here

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -184,20 +215,75 @@ class WorkflowActor(workflowId: WorkflowId,
...
+    serviceRegistryActor.publishToMetadataService(metadataEventMsg) map handleMetadataServiceResponse
+  }
+
+    // Sends a message to self in case the response failed
+  private def handleMetadataServiceResponse(response: MetadataServiceResponse): Unit = {
+    response match {
+      case result@MetadataPutFailed(action, reason) =>
+        log.warning(s"$tag Failed to update action $action in Metadata Service: ${reason.getMessage}")
+        // Notify self of the failed event
+        self ! MetadataServiceFailureResponse(reason, Option(action))
```
geoffjentry: And indirectly I believed you closed over self here as well

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -190,7 +206,13 @@ class WorkflowManagerActor(config: Config)
     val isRestart = id.isDefined
 
     val startMode = if (isRestart) RestartExistingWorkflow else StartNewWorkflow
-    val wfActor = context.actorOf(WorkflowActor.props(workflowId, startMode, source, config), name = s"WorkflowActor-$workflowId")
+    val wfActor = context.actorOf(WorkflowActor.props(workflowId, startMode, source, config, serviceRegistryActor), name = s"WorkflowActor-$workflowId")
+
+    // We have a valid workflowId for the workflow, send it over to the metadata service
+    pushToMetadataService(workflowId) onFailure {
+      case reason: Throwable => self ! MetadataServiceFailureResponse(reason)
```
geoffjentry: another place where I think you closed over self

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
geoffjentry: err duh, I was thinking `sender` - i'm stupid

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -310,4 +335,17 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId, workflowDescript
         Success(WorkflowExecutionDiff(Map(collector -> ExecutionStatus.Starting)))
     }
   }
+
+  private def pushOutputsToMetadataService(data: WorkflowExecutionActorData): Future[Unit] = {
+    import workflow._
+    val keyValues = data.outputStore.store.flatMap{
```
geoffjentry: dont mix the infix and the .flatMap( HOF styles like this

engine/src/main/scala/cromwell/engine/workflow/package.scala
```diff
@@ -0,0 +1,40 @@
...
+
+import akka.actor.{ActorSystem, ActorRef}
+import com.typesafe.scalalogging.StrictLogging
+import cromwell.services.MetadataServiceActor.{MetadataServiceAction, PutMetadataAction, MetadataServiceResponse, MetadataEvent}
+
+import scala.concurrent.{Future, ExecutionContext}
+
+package object workflow {
+  // Sent when there is a failure to push to the metadata service
+  private[workflow] case class MetadataServiceFailureResponse(reason: Throwable, action: Option[MetadataServiceAction] = None)
```
geoffjentry: Make this an ADT w/ implementations with and without action

engine/src/main/scala/cromwell/engine/workflow/package.scala
```diff
@@ -0,0 +1,40 @@
...
+    val StartTime = "start"
+    val EndTime = "end"
+    val Status = "status"
+    val Outputs = "outputs"
+    val Inputs = "inputs"
+    val Calls = "calls"
+  }
+
+  implicit class EnhancedServiceRegistry(val serviceRegistryActor: ActorRef) extends StrictLogging {
+    def publishToMetadataService(message: MetadataEvent)(implicit actorSystem: ActorSystem, ec: ExecutionContext): Future[MetadataServiceResponse] = {
```
geoffjentry: I'm really not a fan of using enhanced classes to send messages to actors like this.

engine/src/main/scala/cromwell/engine/workflow/package.scala
```diff
@@ -0,0 +1,40 @@
...
+      import cromwell.util.PromiseActor._
+
+      // Since this will always publish, we expect the callee to just provide an event,
+      // and we wrap it into a `PutMetadataAction` before sending it to the MetadataService
+      serviceRegistryActor.askNoTimeout(PutMetadataAction(message)).mapTo[MetadataServiceResponse].recover {
```
geoffjentry: Also if you really want to use an ask inside of another actor's receive loop, you might want to check out this library: https://github.com/jducoeur/Requester

engine/src/main/scala/cromwell/engine/workflow/package.scala
```diff
@@ -0,0 +1,40 @@
...
+        case reason: Throwable =>
+          // TODO: Do we still proceed with the execution? What do we do with the exception here?
```
geoffjentry: This is a good question. We should probably discuss among the group

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
gauravs90: So if I create another actor to delegate these metadata related futures to, we would essentially be creating an abstraction between the engine and the service like before isn't it?

engine/src/main/scala/cromwell/engine/workflow/package.scala
gauravs90: I thought we used to do it like this before 😕 
Alternatives? Just a normal object.method?

engine/src/main/scala/cromwell/engine/workflow/package.scala
gauravs90: Mind if I move all this stuff to something like a MetadataUtils? Or any other place? It seems like this package object is Metadata heavy currently.

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
geoffjentry: It's not quite the same situation. In that case there was a dedicated service to understanding the structure of these metadata events. In this case there's already async behavior being defined in your actor, and every async operation can also be represented as an actor.

However looking at it again, I think the Futures are only because of your ask in that utility implicit, is that correct? If so, that's what I think the real issue is here.

engine/src/main/scala/cromwell/engine/workflow/package.scala
geoffjentry: I'm always plus-one on moving stuff out of package objects

engine/src/main/scala/cromwell/engine/workflow/package.scala
```diff
@@ -0,0 +1,40 @@
...
+    val SubmissionTime = "submission"
+    val StartTime = "start"
+    val EndTime = "end"
+    val Status = "status"
+    val Outputs = "outputs"
+    val Inputs = "inputs"
+    val Calls = "calls"
+  }
+
+  implicit class EnhancedServiceRegistry(val serviceRegistryActor: ActorRef) extends StrictLogging {
```
geoffjentry: You should also have these enhanced classes extend AnyVal

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -184,20 +215,79 @@ class WorkflowActor(workflowId: WorkflowId,
...
+        case reason: Throwable => self ! MetadataServiceFailureResponse(reason)
+      }
+  }
+
+  onTransition {
+    case oldState -> terminalState if terminalState.terminal =>
+      log.info(s"$tag transition from $oldState to $terminalState: shutting down")
+      // Add the end time of the workflow in the MetadataService
+      val metadataEventMsg = MetadataEvent(MetadataKey(workflowId, None, WorkflowMetadataKeys.EndTime), MetadataValue(DateTime.now.toString))
+      serviceRegistryActor.publishToMetadataService(metadataEventMsg) map handleMetadataServiceResponse recover {
```
jainh: How this recover is different from recover in publishToMetadataService()... ? I think you were using that recover only for logging which can be done here as well.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -1,22 +1,25 @@
...
 import cromwell.engine.backend.{BackendConfiguration, CromwellBackends}
+import cromwell.engine.workflow.{MetadataActionFailureResponse, MetadataServiceFailureResponse}
 import cromwell.engine.workflow.lifecycle.execution.JobPreparationActor.{BackendJobPreparationFailed, BackendJobPreparationSucceeded}
 import cromwell.engine.workflow.lifecycle.execution.WorkflowExecutionActor.WorkflowExecutionActorState
-import cromwell.engine.{EngineWorkflowDescriptor, ExecutionStatus}
+import cromwell.engine.{EngineWorkflowDescriptor, ExecutionStatus, workflow}
+import cromwell.services.MetadataServiceActor._
 import lenthall.exception.ThrowableAggregation
-import wdl4s._
+import wdl4s.{Scope, _}
```
mcovarr: this is strange

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -184,20 +215,79 @@ class WorkflowActor(workflowId: WorkflowId,
...
+      MetadataEvent(MetadataKey(workflowId, None, s"${WorkflowMetadataKeys.Inputs}.$k"), MetadataValue(v.toWdlString))
+    }
+    val metadataEventMsgs = List(MetadataEvent(MetadataKey(workflowId, None, WorkflowMetadataKeys.Name), MetadataValue(workflowDescriptor.name))) ++ inputMetadataEvents
+    Future.sequence(metadataEventMsgs map serviceRegistryActor.publishToMetadataService).map {
+      _ foreach handleMetadataServiceResponse
+    }
+  }
+
+  // Update the current State of the Workflow (corresponding to the FSM state) in the Metadata service
+  private def pushCurrentStateToMetadataService(workflowState: WorkflowState): Future[Unit] = {
```
jainh: I think instead of defining private methods inside dedicated workflow actor, meta data related behaviors can be moved to metadata actor, that would be more appropriate than growing workflow actor, However follow SRP (Single Responsibility principle), it looks like it is breaking it right now.

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -180,6 +186,16 @@ class WorkflowManagerActor(config: Config)
       logger.info(s"$tag transitioning from $fromState to $toState")
   }
 
+  private def pushToMetadataService(workflowId: WorkflowId): Future[List[MetadataServiceResponse]] = {
```
jainh: Same as above.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -210,7 +227,18 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId, workflowDescript
   onTransition {
     case _ -> toState if toState.terminal =>
```
jainh: Combine inside conditional statement with external if statement  if toState.Terminal &&  toState == WorkflowExecutionSuccessfulState, even when u need to case statements , that would not require if else ...

geoffjentry: @jainh Can you rephrase that? I'm trying to understand what you're suggesting

jainh: @geoffjentry what I was trying to say what if above case can be broken in onTransition  { 
case _ -> toState if toState == WorkflowExecutionSuccessfulState => ....
case _  -> toState if toState == WorkflowExecutionAbortedState => .... 
case fromState -> toState  => ....
}

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
gauravs90: `cmd` + `alt` + `L` thing from intelliJ I think 😓

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -190,7 +206,13 @@ class WorkflowManagerActor(config: Config)
     val isRestart = id.isDefined
 
     val startMode = if (isRestart) RestartExistingWorkflow else StartNewWorkflow
-    val wfActor = context.actorOf(WorkflowActor.props(workflowId, startMode, source, config), name = s"WorkflowActor-$workflowId")
+    val wfActor = context.actorOf(WorkflowActor.props(workflowId, startMode, source, config, serviceRegistryActor), name = s"WorkflowActor-$workflowId")
+
+    // We have a valid workflowId for the workflow, send it over to the metadata service
+    pushToMetadataService(workflowId) onFailure {
```
mcovarr: `onFailure` is going to be deprecated in 2.12, try `recover` instead

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
mcovarr: If you're going to do that you shouldn't need `toState`

```scala
onTransition { 
  case _ -> WorkflowExecutionSuccessfulState => ....
  case _ -> WorkflowExecutionAbortedState => .... 
  case fromState -> toState => ....
}
```

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
geoffjentry: FWIW Klang suggests failed.foreach: https://github.com/viktorklang/blog/blob/master/Futures-in-Scala-2.12-part-5.md

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
gauravs90: We discussed it and let's not go there again. This can be extended to a better design later on I guess.

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
gauravs90: `failed.foreach` it is!

engine/src/main/scala/cromwell/engine/package.scala
gauravs90: @geoffjentry Just the `EngineWorkflowDescriptor`? Most of the stuff there is from the oldestyle and might be removed eventually though.

geoffjentry: For this, yeah. I'm just trying to watch for increased bloat on these POs

gauravs90: A-okay!

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -184,20 +215,79 @@ class WorkflowActor(workflowId: WorkflowId,
...
+  onTransition {
+    case fromState -> toState =>
+      log.info(s"$tag transitioning from $fromState to $toState")
+      // This updates the workflow status
+      pushCurrentStateToMetadataService(toState.workflowState) onFailure {
+        case reason: Throwable => self ! MetadataServiceFailureResponse(reason)
+      }
+  }
+
+  onTransition {
```
Horneth: Why is there 2 `onTransition` blocks ?

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -78,6 +81,8 @@ class WorkflowManagerActor(config: Config)
 
   private val donePromise = Promise[Unit]()
 
+  val serviceRegistryActor = context.actorOf(ServiceRegistryActor.props(config), "ServiceRegistoryActor")
```
Horneth: Just a heads up, I moved this to a `ServiceRegistry` trait in my PR so it can be used anywhere (typically the REST endpoint needs it too).

engine/src/main/scala/cromwell/engine/workflow/package.scala
```diff
@@ -0,0 +1,35 @@
...
+    val SubmissionTime = "submission"
+    val StartTime = "start"
+    val EndTime = "end"
+    val Status = "status"
+    val Outputs = "outputs"
+    val Inputs = "inputs"
+    val Calls = "calls"
+  }
+
+  implicit class EnhancedServiceRegistry(val serviceRegistryActor: ActorRef) extends AnyVal {
```
Horneth: I'm probably one of the most `implicit` friendly people in red team, but on this one, you're basically adding a `publishToMetadataService` method to any potential `ActorRef` in scope, which is a bit strange.

Maybe make it a trait instead ?

```scala
trait MetadataServiceUtil extends StrictLogging {
    def serviceRegistryActor: ActorRef

    def publishToMetadataService(message: MetadataEvent)(implicit actorSystem: ActorSystem, ec: ExecutionContext): Future[MetadataServiceResponse] = {
      import cromwell.util.PromiseActor._

      // Since this will always publish, we expect the callee to just provide an event,
      // and we wrap it into a `PutMetadataAction` before sending it to the MetadataService
      serviceRegistryActor.askNoTimeout(PutMetadataAction(message)).mapTo[MetadataServiceResponse].recover {
        case reason: Throwable =>
          // TODO: Do we still proceed with the execution? What do we do with the exception here?
          logger.warn(s"Failed to push event $message in the MetadataService: ${reason.getMessage}", reason)
          throw reason
      }
    }
  }
```

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
Horneth: It looks like you don't really care about the response anyway (which is good IMO). Why not make the push to metadata a fire and forget ? This way we don't need a `Future` nor an actor. We can always catch the `MetadataFailure` cases in the `whenUnhandled` and log it or do something more clever later, and for the rest we don't really care do we ?

geoffjentry: @Horneth I'm glad you said this. I was thinking this on the bus & meant to look into it and then completely forgot :) So yeah, what thibault said

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -310,4 +342,17 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId, workflowDescript
         Success(WorkflowExecutionDiff(Map(collector -> ExecutionStatus.Starting)))
     }
   }
+
+  private def pushOutputsToMetadataService(data: WorkflowExecutionActorData): Future[Unit] = {
+    import workflow._
+    val keyValues = data.outputStore.store.flatMap {
+      case (key, value) => value map (entry => s"${key.call.fullyQualifiedName}.${entry.name}" -> entry.wdlValue.map(_.toWdlString).getOrElse("NA"))
+    }
+    val metadataEventMsgs = keyValues map{ case (k,v) => MetadataEvent(MetadataKey(workflowId, None, s"${WorkflowMetadataKeys.Outputs}.$k"), MetadataValue(v))}
```
Horneth: This does need to be changed (`s"${WorkflowMetadataKeys.Outputs}.$k"`).
We need to pick a different key separator than `.` What about `:` ?

engine/src/main/scala/cromwell/engine/workflow/package.scala
cjllanwarne: This enhancer will enhance every actorRef in scope as though it's a service registry. Which is probably not what was wanted?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
mcovarr: are we really going to migrate all the historical data to use `:`?

Horneth: No, I'm talking about the separator for the metadata `keys` that are used to infer json structure

mcovarr: phew

engine/src/main/scala/cromwell/engine/EngineWorkflowDescriptor.scala
```diff
@@ -0,0 +1,13 @@
...
+import wdl4s._
+
+final case class EngineWorkflowDescriptor(backendDescriptor: BackendWorkflowDescriptor,
+                                          declarations: WorkflowCoercedInputs,
+                                          backendAssignments: Map[Call, String],
+                                          failureMode: WorkflowFailureMode) {
+  def id = backendDescriptor.id
+  def namespace = backendDescriptor.workflowNamespace
+  def name = namespace.workflow.unqualifiedName
+}
```
mcovarr: \n

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
mcovarr: All `onTransition` blocks will run.

engine/src/main/scala/cromwell/engine/workflow/package.scala
jainh: @Horneth , But making it util sounds like a common behavior for other Actors/Service, is that the intention?

Horneth: Hmm I don't see why making it accessible to other actors / services would be a problem. If we really want to restrict it to a package or something we can use some `private [...]` magic but I don't see the use case.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
gauravs90: `:` or `$` or `#` anybody has any opinions?

Horneth: not `$` I'd say because we've used for scatter placeholder I think or logs. I'm pretty sure something uses it anyway.

engine/src/main/scala/cromwell/engine/EngineWorkflowDescriptor.scala
geoffjentry: Didn't I just see you giving thumbs downs to whitespace comments earlier today? :)

mcovarr: yeah but missing newlines have the ability to actually break stuff

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
gauravs90: Actually, yes. I was thinking on the same lines as @geoffjentry, but that'd be pretty great.

engine/src/main/scala/cromwell/engine/EngineWorkflowDescriptor.scala
gauravs90: @mcovarr Just curious, but that happens? How?

mcovarr: I know @kshakir has seen these scenarios

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -184,20 +212,56 @@ class WorkflowActor(workflowId: WorkflowId,
...
+      // This updates the workflow status
+      pushCurrentStateToMetadataService(toState.workflowState)
+  }
+
+  onTransition {
+    case oldState -> terminalState if terminalState.terminal =>
+      log.info(s"$tag transition from $oldState to $terminalState: shutting down")
+      // Add the end time of the workflow in the MetadataService
+      val metadataEventMsg = MetadataEvent(MetadataKey(workflowId, None, WorkflowMetadataKeys.EndTime), MetadataValue(DateTime.now.toString))
+      serviceRegistryActor ! PutMetadataAction(metadataEventMsg)
```
gauravs90: @Horneth With fire and forget, we no longer have futures which is in resonance with what Jeff wanted. However, we can no longer impose an order on the messages, i.e. earlier it would have received the failure message andThen send a workflow completion message to parent.  With that, if it had received a PoisonPill it would have done the appropriate action before being killed. 
We are not doing anything useful with the failures currently anyway, but it might be the case that the workflow completed and the parent sent a message to kill itself, and then it receives the Metadata failure message which will probably go to the dead letter mailbox (Sigh. Side effects of giving the metadata responsibility to this actor I suppose)

gauravs90:
@geoffjentry @Horneth I believe I have addressed all the comments. Can you guys take a look again?

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
Horneth: Yeah that's a good point, I guess when we want to actually handle the failure case we'll re-evaluate. Maybe a "Retry Actor" could be enough.

cjllanwarne: Surely with fire and forget, you'd never do anything with a `Failure` message anyway?

I guess if you wanted to, you could fake the `MetadataPut` message's `sender` to be something which can make sure metadata persists. A `MetadataPersistanceEnforcerActor` or something?

Horneth:
👍

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
geoffjentry: It was the "not doing anything useful with the failures currently" part which sparked my thought this morning :)

IIRC didn't the futures trigger messages to self? Wouldn't that have the same dead letter issue?

Anyhoo, my take was that that nothing was really being done to handle that case anyways so why handwring over futures, actors, etc.

geoffjentry:
👍

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
gauravs90: Yes, but I was chaining the futures with andThen ..  so I guess the failure message would have been processed first before the potential PoisonPill from the parent. But I think this way with fire and forget is much better, and I'm totally with you that, where possible, futures should be avoided within an actor.

PR #830, opened 2016-05-13
Added server side pagination for the workflow query using link headers.

jacarey:
This is modelled after github's pagination https://developer.github.com/v3/#pagination and Link Header RFC https://tools.ietf.org/html/rfc5988#section-5

### DIFF
--- a/README.md
+++ b/README.md
@@ -1533,6 +1533,8 @@ This endpoint allows for querying workflows based on the following criteria:
 * `status`
 * `start` (start datetime)
 * `end` (end datetime)
+* `page` (page of results)
+* `pagesize` (# or results per pa
 
 Names, ids, and statuses can be given multiple times to include
 workflows with any of the specified names, ids, or statuses. When
@@ -1540,7 +1542,7 @@ multiple names are specified, any workflow matching one of the names
 will be returned. The same is true for multiple ids or statuses. When
 different types of criteria are specified, for example names and
 statuses, the results must match both the one of the specified names and
-one of the statuses.
+one of the statuses. Using page and pagesize will enable server side pagination.
 
 Valid statuses are `Submitted`, `Running`, `Aborting`, `Aborted`, `Failed`, and `Succeeded`.  `start` and `end` should
 be in [ISO8601 datetime](https://en.wikipedia.org/wiki/ISO_8601) format and `start` cannot be after `end`.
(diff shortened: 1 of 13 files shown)


mcovarr:
README.md should be updated

mcovarr:
Did you mean for this PR to be against develop?  We're currently using this branch for pluggable backend work which won't be usable for GOTC for quite a while.

jacarey:
It was suggested that we create both a develop PR as well as for a hotfix. Had I been aware of the differences between the two branches I probably would not have created this PR... but we had already finished. Similar work will be done for a PR on the hotfix branch.

I'm happy to do whatever your team prefers. I suspect that this work will need to be duplicated in either case. :(

mcovarr:
I'd be surprised if these changes can't be migrated fairly easily to 0.19-hotfix.  Metadata is an area under heavy construction on develop, and I suspect you're right this will need to be completely reimplemented once the dust settles.  😦 

I'd certainly thumbs up a 0.19-hotfix version of this with updates to the README.md.  😄

jacarey:
Thanks Miguel. Sounds like it makes sense to close this PR. We'll create a hotfix PR with this functionality.

mcovarr:
Actually now I'm realizing that the data retrieved by this API is not really of the "meta" variety, so it's possible this might continue to work in the PBE world as is.  You might want to leave this open, the worst that could happen is it gets disabled in a refactoring.

But PBE is probably a couple of months away, so if you'd like to use this functionality before then you'll still need to make a 0.19_hotfix version of this PR as well.

jacarey:
Sounds good I'll leave this open for now and also create a new PR to add this functionaity to 0.19_hotfix

jacarey:
This is the hotfix PR:
https://github.com/broadinstitute/cromwell/pull/842

mcovarr:
@jacarey are the changes here consistent with what was merged to 0.19_hotfix?  If so I'd like to get this merged so it can be hooked up with the work I'm doing in #855.

jacarey:
We made changes to avoid changing the model that was returned (opting for metadata instead) those changes aren't reflected here.

I will update this PR shortly.

jacarey:
Ok I've updated the PR.

mcovarr:
:+1:

cjllanwarne:
Merge away! 👍

PR #833, opened 2016-05-14
Added dispatcher for slow/blocking actors. Refactored JobPreparationA…

geoffjentry:
…ctor to use this instead of multiple Futures/FSM states

### DIFF
--- a/engine/src/main/resources/application.conf
+++ b/engine/src/main/resources/application.conf
@@ -17,6 +17,16 @@ akka {
       }
     }
   }
+
+  dispatchers {
+    # A dispatcher for actors with slow/blocking behavior
+    slow-actor-dispatcher {
+      type = Dispatcher
+      executor = "fork-join-executor"
+      # Using the forkjoin defaults, this can be tuned if we wish
+    }
+  }
+
 }
 
 spray.can {
@@ -236,8 +246,3 @@ database {
     }
   }
 }
-
-
-
-
-
--- a/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/JobPreparationActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/JobPreparationActor.scala
@@ -1,6 +1,6 @@
 package cromwell.engine.workflow.lifecycle.execution
 
-import akka.actor.{FSM, LoggingFSM, Props}
+import akka.actor.{Actor, ActorLogging, Props}
 import cromwell.backend._
 import cromwell.engine.EngineWorkflowDescriptor
 import cromwell.engine.workflow.lifecycle.execution.JobPreparationActor._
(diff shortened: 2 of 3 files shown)


geoffjentry:
@Horneth @mcovarr

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -261,7 +261,7 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId, workflowDescript
       case k => Failure(new UnsupportedOperationException(s"Unknown entry in execution store: ${k.tag}"))
     }
 
-    TryUtil.sequence(executionDiffs.toSeq) match {
+    TryUtil.sequence(executionDiffs) match {
```
geoffjentry: this was just an intellij suggestion i noticed while poking around

engine/src/test/scala/cromwell/CromwellTestkitSpec.scala
```diff
@@ -60,6 +60,12 @@ object CromwellTestkitSpec {
       |       receive = on
       |    }
       |  }
+      |  dispatchers {
+      |    slow-actor-dispatcher {
+      |      type = Dispatcher
+      |      executor = "fork-join-executor"
+      |    }
+      |  }
```
geoffjentry: This is kinda irritating but the alternative had its own issues

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/JobPreparationActor.scala
```diff
@@ -9,85 +9,56 @@ import wdl4s.expression.WdlStandardLibraryFunctions
...
+  sealed trait JobPreparationActorResponse
+  case class BackendJobPreparationSucceeded(jobDescriptor: BackendJobDescriptor, props: Props) extends JobPreparationActorResponse
+  case class BackendJobPreparationFailed(jobKey: JobKey, throwable: Throwable) extends JobPreparationActorResponse
 
   def props(executionData: WorkflowExecutionActorData,
             jobKey: BackendJobDescriptorKey,
             factory: BackendLifecycleActorFactory,
-            configDescriptor: BackendConfigurationDescriptor) = Props(new JobPreparationActor(executionData, jobKey, factory, configDescriptor))
+            configDescriptor: BackendConfigurationDescriptor) = {
+    Props(new JobPreparationActor(executionData, jobKey, factory, configDescriptor)).withDispatcher("akka.dispatchers.slow-actor-dispatcher")
```
mcovarr: should the association of this actor class with the dispatcher be done in compiled code or out in the config file?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/JobPreparationActor.scala
```diff
@@ -9,85 +9,56 @@ import wdl4s.expression.WdlStandardLibraryFunctions
...
-          context.parent ! BackendJobPreparationSucceeded(jobDescriptor, props)
-          context stop self
-        case Failure(f) =>
-          context.parent ! BackendJobPreparationFailed(jobKey, f)
-          context stop self
-      }
-      goto(BuildingProps)
+      context.parent ! response
+      context stop self
   }
```
mcovarr: no more warning for unhandled messages?  or just rolling with a default warning?

engine/src/main/resources/application.conf
```diff
@@ -233,3 +243,5 @@ database {
 
 
 
+
+
```
mcovarr: oh here are all the newlines

mcovarr:
:+1:

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/JobPreparationActor.scala
Horneth: Or if not in the conf, can we have a trait or something to reuse this dispatcher without having to type the conf path every time ? I'm thinking we could also use it for the `MetadataBuilderActor` which could be slow-ish depending on the number of events to parse/aggregate.

Horneth:
👍

engine/src/main/resources/application.conf
geoffjentry: Heh. Artifact of bouncing between handling dispatcher via conf vs on the props

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/JobPreparationActor.scala
geoffjentry: @Horneth What would the trait do specifically? The `withDispatcher` call is a method of the `Props` object. I could see an enhanced `Props` class with a `useOurSlowpokeDispatcher()` but I don't think that's what you mean here. Also I don't think it's much better, it doesn't scale well to multiple dispatchers (or actually on second thought, perhaps it *is* more clear?)

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/JobPreparationActor.scala
geoffjentry: I don't remember if this was intentional or not. I'm guessing that it's because I lost the builtin log and kinda doubt we'll ever use these unhandled log messages anyways - but I'll add it back in.

geoffjentry:
Comments from @mcovarr addressed
@Horneth - I had one followup Q

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/JobPreparationActor.scala
Horneth: I was just picturing something like

```scala
trait SlowDispatcher {
  private val slowDispatcherPath = "akka.dispatchers.slow-actor-dispatcher"
  def withSlowDispatcher(props: Props) = props.withDispatcher(slowDispatcherPath)
}
```
but it's really not a blocker, and I'm not sure we even have a use for it yet

geoffjentry: That's pretty much what I was on about w/ the enhanced class on Props. Since you don't seem to care a lot and my opinion is waffling back and forth the final outcome will probably depend on what I happen to think when I go to merge this in ;)

geoffjentry: I'm going to go with the "let's wait until n is at least 2" approach (aka the lazy approach)

Issue #808, opened 2016-05-09
Add retries around command script and JES run creation


mcovarr:
See #797

Issue #806, opened 2016-05-09
Gcloud Auth Upload + Remove


ruchim:
AC: Test that the FireCloud style auth upload works

mcovarr:
This was definitely *not* implemented as part of #797

mcovarr:
Also implement deletion of this file.

mcovarr:
In the initialization and finalization lifecycle actors respectively.

PR #834, opened 2016-05-15
Add actor based retry functionality. Closes #754

### LINKED ISSUE #754: Make GCS auth uploads async

geoffjentry:
Strictly speaking does not technically close the original issue although at the moment the gcs auth file upload retrying doesn't seem to exist in develop (@mcovarr can you verify this?).

- Replaced the `withRetry` in `DataAccess` with the new functionality
- Did not remove nor replace the uses of `TryUtil.withRetry` as they're both OldeStyle and it didn't seem worth the hassle. If reviewers would rather they be switched over I can do that.

### DIFF
--- a/core/src/main/scala/cromwell/core/retry/Retry.scala
+++ b/core/src/main/scala/cromwell/core/retry/Retry.scala
@@ -26,8 +26,8 @@ object Retry {
   def withRetry[A](f: => Future[A],
                    maxRetries: Option[Int] = Option(10),
                    backoff: SimpleExponentialBackoff = SimpleExponentialBackoff(5 seconds, 10 seconds, 1.1D),
-                   isTransient: Option[Throwable => Boolean] = None,
-                   isFatal: Option[Throwable => Boolean] = None)
+                   isTransient: Throwable => Boolean = throwableToFalse,
+                   isFatal: Throwable => Boolean = throwableToFalse)
                    (implicit actorSystem: ActorSystem): Future[A] = {
     // In the future we might want EC passed in separately but at the moment it caused more issues than it solved to do so
     implicit val ec: ExecutionContext = actorSystem.dispatcher
@@ -35,9 +35,9 @@ object Retry {
 
     if (maxRetries.forall(_ > 0)) {
       f recoverWith {
-        case throwable if isFatal.evaluate(throwable) => Future.failed(new CromwellFatalException(throwable))
-        case throwable if !isFatal.evaluate(throwable) =>
-          val retriesLeft = if (isTransient.evaluate(throwable)) maxRetries else maxRetries map { _ - 1 }
+        case throwable if isFatal(throwable) => Future.failed(new CromwellFatalException(throwable))
+        case throwable if !isFatal(throwable) =>
+          val retriesLeft = if (isTransient(throwable)) maxRetries else maxRetries map { _ - 1 }
           after(delay, actorSystem.scheduler)(withRetry(f, backoff = backoff, maxRetries = retriesLeft))
       }
     } else f recoverWith {
@@ -45,9 +45,6 @@ object Retry {
     }
   }
 
-  implicit class EnhancedOptionThrowableToBoolean(val f: Option[Throwable => Boolean]) extends AnyVal {
-    /** If f is defined, run the function using the passed throwable otherwise return false */
-    def evaluate(throwable: Throwable) = f.fold(false) { _(throwable) }
-  }
+  def throwableToFalse(t: Throwable) = false
 }
 
--- a/core/src/main/scala/cromwell/core/retry/Retry.scala
+++ b/core/src/main/scala/cromwell/core/retry/Retry.scala
@@ -1,0 +1,53 @@
+package cromwell.core.retry
+
+import akka.actor.ActorSystem
+import cromwell.core.CromwellFatalException
+
+import scala.concurrent.duration._
+import scala.language.postfixOps
+import scala.concurrent.{ExecutionContext, Future}
+import akka.pattern.after
+
+object Retry {
+  /**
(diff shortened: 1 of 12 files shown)


geoffjentry:
@cjllanwarne @kshakir

core/src/main/scala/cromwell/core/retry/Retry.scala
```diff
@@ -0,0 +1,53 @@
...
+    * @param isFatal An optional function of Throwable => Boolean. If provided and it returns true, withRetry will halt
+    * @param actorSystem An ActorSystem, used for its scheduler and dispatcher.
+    * @tparam A The return type of the thunk
+    * @return The final completed Future
+    */
+  def withRetry[A](f: => Future[A],
+                   maxRetries: Option[Int] = Option(10),
+                   backoff: SimpleExponentialBackoff = SimpleExponentialBackoff(5 seconds, 10 seconds, 1.1D),
+                   isTransient: Option[Throwable => Boolean] = None,
+                   isFatal: Option[Throwable => Boolean] = None)
```
kshakir: ToL: The `Option` wrapper plus the implicit utility below seem like overkill. Can you cite precedent in any project / codebase for a parameter that is an option-of-a-function? I'd even accept an overloaded java method with similar semantics. ;) I know that null is a "billion dollar mistake", but is it so bad to just have these parameters return `false` by default instead of `None`?

core/src/main/scala/cromwell/core/retry/Retry.scala
```diff
@@ -0,0 +1,53 @@
...
+                   (implicit actorSystem: ActorSystem): Future[A] = {
+    // In the future we might want EC passed in separately but at the moment it caused more issues than it solved to do so
+    implicit val ec: ExecutionContext = actorSystem.dispatcher
```
kshakir: ToL: Just like we had a `slick` expert come in, more-and-more I'd love to chat with an akka/threading expert.

kshakir:
👍

core/src/main/scala/cromwell/core/retry/Retry.scala
geoffjentry: That's a fair point. So (x: Throwable) => false?

core/src/main/scala/cromwell/core/retry/Retry.scala
geoffjentry: Agreed.

In the meantime I feel I've been making a lot of headway on that front over the last week or two just by reading a ton of crap. I've found trolling through their mailing list (in particular Klang's answers) has led to a lot of helpful tidbits

core/src/main/scala/cromwell/core/retry/Retry.scala
```diff
@@ -0,0 +1,50 @@
...
+import scala.language.postfixOps
+import scala.concurrent.{ExecutionContext, Future}
+import akka.pattern.after
+
+object Retry {
+  /**
+    * Retries a Future on a designated backoff strategy until either a designated number of retries or a fatal error
+    * is reached.
+    *
+    * @param f A Future thunk which will be executed once per cycle
```
cjllanwarne: Is this really what you meant? https://en.wikipedia.org/wiki/Thunk

I've never seen that word before!

geoffjentry: http://stackoverflow.com/a/22670623

cjllanwarne:
I can't fault the code, but as you say, at the moment it doesn't complete the ticket. Can we add another ticket to actually wire this up?

Otherwise, 👍

geoffjentry:
@cjllanwarne I was hoping @mcovarr would chime in here. I *think* that it's his current PR which has his own retry scheme and that the path forward would be that I pull the switcharoo when his PR is merged but I was waiting for confirmation on exactly what needs to be done prior to fiddling with it.

core/src/main/scala/cromwell/core/retry/Retry.scala
mcovarr: hmm the red book uses that word a *lot*...

geoffjentry: it's no orange book

core/src/test/scala/cromwell/core/retry/RetrySpec.scala
```diff
@@ -0,0 +1,92 @@
...
+        Future.failed(ex)
+      }
+    }
+  }
+
+  implicit val defaultPatience = PatienceConfig(timeout = Span(30, Seconds), interval = Span(100, Millis))
+
+  private def runRetry(retries: Int,
+                       work: MockWork,
+                       isTransient: Throwable => Boolean = Retry.throwableToFalse,
```
mcovarr: huh?  why not just lambda this here rather than add the method to production code?

mcovarr:
@geoffjentry I removed retry in two spots which hopefully could be added back per this scheme as part of #808.  Note that neither of those spots corresponds to GCS auth upload, which should be happening in the initialization actor and is the subject of #806.

I actually thought this ticket was meant for the hotfix branch to deal with problems the Greenies had seen, but there doesn't seem to be any more detail or labeling to confirm or refute that.

core/src/test/scala/cromwell/core/retry/RetrySpec.scala
geoffjentry: It's used in 2 places in the real code as well. One of those things where I'm expecting code reviewers to say "it's used more than once, put it in a const"

mcovarr: Sounds reasonable, a const in the test would be less strange than adding it to the production code only for the test.

PR #841, opened 2016-05-17
Hotfix -- configuring restarts

ruchim:
When the workflow-restart config option is set to false, incomplete workflows will not be restarted, and instead all incomplete workflows are aborted. 

Note: Still need to update ReadMe

Questions: The sys.addShutdownHook takes about a minute to start startup (there's a minute long wait between the logs) which seems too long?

### DIFF
--- a/engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
@@ -1,6 +1,5 @@
 package cromwell.engine.workflow
 
-import akka.Main
 import akka.actor.FSM.{CurrentState, SubscribeTransitionCallBack, Transition}
 import akka.actor._
 import akka.event.Logging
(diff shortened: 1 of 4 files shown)


engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -102,9 +103,13 @@ class WorkflowManagerActor(config: Config)
 
   private val donePromise = Promise[Unit]()
 
+  val isRestartable = config.getConfig("system").getBooleanOr("workflow-restart", default = true)
+  val serverMode = CromwellServer.isServerMode
```
Horneth: How can this be `false` ?

ruchim: great question, it definitely isn't. I'm not sure where to add the hook of running in server mode. And I chose this because there was a note about the CromwellServer object: 
// Note that as per the language specification, this is instantiated lazily and only used when necessary (i.e. server mode)

geoffjentry: This will actually instantiate that CromwellServer object. What that comment means is that it won't actually exist until the first thing reaches in and touches it.

mcovarr: Is the `system` section mandatory?  It looks like this would throw if it's missing.

mcovarr:
Looks like this addresses #845 that I ran into yesterday?

geoffjentry:
I was working with @ruchim to refactor this a bit, got distracted and now she's away for a few days. @kcibul am I correct in remembering that this isn't particularly high priority? If I'm right, I can work with her on Monday on this.

PR #842, opened 2016-05-17
Added server side pagination for the workflow query using link headers.

jacarey:
This is model after github's pagination https://developer.github.com/v3/#pagination and Link Header RFC https://tools.ietf.org/html/rfc5988#section-5

Update README.md and made review changes from https://github.com/broadinstitute/cromwell/pull/830

### DIFF
--- a/engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
+++ b/engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
@@ -17,7 +17,7 @@ import slick.driver.JdbcProfile
 import cromwell.engine.db._
 import cromwell.engine.finalcall.FinalCall
 import cromwell.engine.workflow._
-import cromwell.webservice.{CallCachingParameters, WorkflowQueryParameters, WorkflowQueryResponse}
+import cromwell.webservice.{CallCachingParameters, QueryMetadata, WorkflowQueryParameters, WorkflowQueryResponse}
 import lenthall.config.ScalaConfig._
 import org.apache.commons.lang3.StringUtils
 import org.joda.time.DateTime
@@ -872,18 +872,35 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
   }
 
   override def queryWorkflows(queryParameters: WorkflowQueryParameters)
-                             (implicit ec: ExecutionContext): Future[WorkflowQueryResponse] = {
-    val action = dataAccess.queryWorkflowExecutions(queryParameters).result
-    runTransaction(action) map { workflows =>
-      WorkflowQueryResponse(workflows map { workflow =>
-        WorkflowQueryResult(
+                             (implicit ec: ExecutionContext): Future[(WorkflowQueryResponse, Option[QueryMetadata])] = {
+    case class CountAndWorkflowExecutions(count: Int, workflowExecutions: Seq[WorkflowExecution])
+    val action = for {
+      workflowExecutions <- dataAccess.queryWorkflowExecutions(queryParameters).result
+      count <- dataAccess.countWorkflowExecutions(queryParameters).result
+    } yield CountAndWorkflowExecutions(count, workflowExecutions)
+
+    runTransaction(action) map {
+      case c => (WorkflowQueryResponse(
+        for {
+          workflow <- c.workflowExecutions
+        } yield WorkflowQueryResult(
           id = workflow.workflowExecutionUuid,
           name = workflow.name,
           status = workflow.status,
           start = new DateTime(workflow.startDt),
-          end = workflow.endDt map { new DateTime(_) })
-      })
+          end = workflow.endDt map {
+            new DateTime(_)
+          })),
+        //only return metadata if page is defined
+        queryParameters.page map { _ => QueryMetadata(queryParameters.page, queryParameters.pageSize, Option(c.count)) })
     }
+  }
+
+  override def countWorkflows(queryParameters: WorkflowQueryParameters)
+                                      (implicit ec: ExecutionContext): Future[Int] = {
+    val action = dataAccess.countWorkflowExecutions(queryParameters).result
+
+    runTransaction(action)
   }
 
   override def updateCallCaching(parameters: CallCachingParameters)(implicit ec: ExecutionContext): Future[Int] = {
--- a/engine/src/main/scala/cromwell/webservice/ApiDataModels.scala
+++ b/engine/src/main/scala/cromwell/webservice/ApiDataModels.scala
@@ -38,6 +38,8 @@ case class WorkflowMetadataResponse(id: String,
 
 case class WorkflowQueryResponse(results: Seq[WorkflowQueryResult])
 
+case class QueryMetadata(page: Option[Int], pageSize: Option[Int], totalRecords: Option[Int])
+
 final case class CallCachingResponse(updateCount: Int)
 
 case class WorkflowFailuresResponse(id: String,
--- a/engine/src/main/scala/cromwell/webservice/CromwellApiHandler.scala
+++ b/engine/src/main/scala/cromwell/webservice/CromwellApiHandler.scala
@@ -9,9 +9,10 @@ import cromwell.engine.backend.{Backend, CallLogs}
 import cromwell.engine.workflow.MaterializeWorkflowDescriptorActor.{MaterializationFailure, MaterializationSuccess, MaterializeWorkflow}
 import cromwell.engine.workflow.WorkflowManagerActor
 import cromwell.engine.workflow.WorkflowManagerActor.{CallNotFoundException, WorkflowNotFoundException}
-import cromwell.webservice.PerRequest.RequestComplete
+import cromwell.webservice.PerRequest.{RequestComplete, RequestCompleteWithHeaders}
 import cromwell.{core, engine}
-import spray.http.StatusCodes
+import spray.http.HttpHeaders.Link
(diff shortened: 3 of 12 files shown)


engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -873,17 +873,31 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
...
-      })
-    }
+    val workflowExecutionsQuery = dataAccess.queryWorkflowExecutions(queryParameters).result
+
+    val workflowExecutionCountQuery = dataAccess.countWorkflowExecutions(queryParameters).result
+
+    for {
+      count <- runTransaction(workflowExecutionCountQuery)
+      workflowExecutions <- runTransaction(workflowExecutionsQuery)
+    } yield WorkflowQueryResponse(
```
mcovarr: The two separate transactions might yield inconsistent results.  Perhaps compose in the DBIO monad and wrap the composite action with a transaction?

```scala

    case class CountAndWorkflowExecutions(count: Int, workflowExecutions: Seq[WorkflowExecution])

    val action = for {
      workflowExecutions <- dataAccess.queryWorkflowExecutions(queryParameters).result
      count <- dataAccess.countWorkflowExecutions(queryParameters).result
    } yield CountAndWorkflowExecutions(count, workflowExecutions)
    
    runTransaction(action) map { case c => 
...
```

engine/src/main/scala/cromwell/webservice/CromwellApiHandler.scala
```diff
@@ -83,6 +84,34 @@ class CromwellApiHandler(requestHandlerActor: ActorRef) extends Actor {
 
   private def error(t: Throwable)(f: Throwable => RequestComplete[_]): Unit = context.parent ! f(t)
 
+  private def generatePaginationParams(page: Int, pageSize: Int): String = {
+    //uri encode?
```
mcovarr: hmm yeah not sure if the url encoding / decoding here and below should be robustified

engine/src/main/scala/cromwell/webservice/ApiDataModels.scala
```diff
@@ -36,7 +36,7 @@ case class WorkflowMetadataResponse(id: String,
                                     calls: Map[String, Seq[CallMetadata]],
                                     failures: Option[Seq[FailureEventEntry]])
 
-case class WorkflowQueryResponse(results: Seq[WorkflowQueryResult])
+case class WorkflowQueryResponse(results: Seq[WorkflowQueryResult], page: Option[Int], pageSize: Option[Int], totalRecords: Int)
```
vivster7: @jacarey, I think by not making `totalRecords` an `Option[Int]`, can be a potentially breaking change.

At least it broke our very brittle front end :/. -- Just wanted to draw attention that you now always return totalRecords, even if pagination info is not requested.

engine/src/main/scala/cromwell/webservice/CromwellApiHandler.scala
```diff
@@ -83,6 +84,35 @@ class CromwellApiHandler(requestHandlerActor: ActorRef) extends Actor {
...
+        val firstLink = Link(Uri(baseUrl).withQuery(generatePaginationParams(1, ps)), Link.first)
+
+        val prevPage = math.max(p-1, 1)
+        val prevLink = Link(Uri(baseUrl).withQuery(generatePaginationParams(prevPage, ps)), Link.prev)
+
+        val lastPage = math.ceil(response.totalRecords.toDouble / ps.toDouble).toInt
+        val lastLink = Link(Uri(baseUrl).withQuery(generatePaginationParams(lastPage, ps)), Link.last)
+
+        val nextPage = math.min(p+1, lastPage)
+        val nextLink = Link(Uri(baseUrl).withQuery(generatePaginationParams(nextPage, ps)), Link.last)
```
vivster7: should be `Link.next` not `Link.last`

engine/src/main/scala/cromwell/webservice/CromwellApiHandler.scala
```diff
@@ -83,6 +84,35 @@ class CromwellApiHandler(requestHandlerActor: ActorRef) extends Actor {
 
   private def error(t: Throwable)(f: Throwable => RequestComplete[_]): Unit = context.parent ! f(t)
 
+  //Use Uri.withQuery to ensure proper uri encoding.
+  private def generatePaginationParams(page: Int, pageSize: Int): String = {
+    s"?page=$page&pagesize=$pageSize"
```
vivster7: I think the `?` is not necessary here. I suspect that the `withQuery` function injects the `?`. The links are currently coming back as `Link:<http://localhost:8080/api/workflows/v1/query??page=1&pagesize=5>; rel=first`

engine/src/main/scala/cromwell/webservice/ApiDataModels.scala
jacarey: yeah query workflows always gets the count. Are you suggesting that we only run the count query if we are requesting pagination?

engine/src/test/scala/cromwell/engine/db/slick/SlickDataAccessSpec.scala
```diff
@@ -247,6 +247,31 @@ class SlickDataAccessSpec extends FlatSpec with Matchers with ScalaFutures with
...
     }
 
+    it should "return pagination metadata only when page and pagesize query params are specified" taggedAs DbmsTest in {
+      assume(canConnect || testRequired)
+      val workflowInfo = materializeWorkflowDescriptorFromSources(workflowSources = testSources)
+      (for {
+        _ <- dataAccess.createWorkflow(workflowInfo, Nil, Nil, localBackend)
+        //get metadata when page and pagesize are specified
+        _ <- dataAccess.queryWorkflows(
+          WorkflowQueryParameters(Seq(WorkflowQueryKey.Page.name -> "1", WorkflowQueryKey.PageSize.name -> "50"))) map { response =>
```
vivster7: rather than using response._1 and response._2, can you match the tuple as `(response, meta)`?

Something like `map { case (response, meta) => ...}` So you can avoid all the tuple accesses

jacarey: yeah I can do that. Since it was just a test class I went with the lazy solution.

vivster7:
@mcovarr, could you and your red team breathren give this another look

engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala
```diff
@@ -872,20 +872,37 @@ class SlickDataAccess(databaseConfig: Config) extends DataAccess {
...
           name = workflow.name,
           status = workflow.status,
           start = new DateTime(workflow.startDt),
-          end = workflow.endDt map { new DateTime(_) })
-      })
+          end = workflow.endDt map {
+            new DateTime(_)
+          })),
+        //only return metadata if page is defined
+        if(queryParameters.page.isDefined) Option(QueryMetadata(queryParameters.page, queryParameters.pageSize, Option(c.count))) else None)
```
mcovarr: isn't this just a `map`?

```scala
queryParameters.page map { _ => QueryMetadata(queryParameters.page, queryParameters.pageSize, Option(c.count)) })
```

engine/src/main/scala/cromwell/webservice/CromwellApiHandler.scala
```diff
@@ -83,6 +84,39 @@ class CromwellApiHandler(requestHandlerActor: ActorRef) extends Actor {
 
   private def error(t: Throwable)(f: Throwable => RequestComplete[_]): Unit = context.parent ! f(t)
 
+  //Use Uri.withQuery to ensure proper uri encoding.
```
mcovarr: did you mean to do this?

mcovarr:
:+1:

mcovarr:
@geoffjentry

engine/src/main/scala/cromwell/webservice/CromwellApiHandler.scala
jacarey: yup done below.. the comment is in an awkward place.

engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
```diff
@@ -133,22 +133,22 @@ class MockWorkflowManagerActor extends Actor {
...
             Seq(
               WorkflowQueryResult(
                 id = UUID.randomUUID().toString,
                 name = "w",
                 status = "Succeeded",
                 start = new DateTime("2015-11-01T12:12:11"),
-                end = Option(new DateTime("2015-11-01T12:12:12")))
-            )))
+                end = Option(new DateTime("2015-11-01T12:12:12"))))),
+            if(rawParameters.map(_._1).contains("page")) Option(QueryMetadata(Option(1), Option(5), Option(1))) else None)
```
mcovarr: perhaps a tad prettier, though this is just a test:

```scala
rawParameters.collectFirst { case (p, _) if p.contains("page") => QueryMetadata(Option(1), Option(5), Option(1)) })
```

geoffjentry:
@jacarey @vivster7 Just as a sanity check, I'm assuming you all actually tested this for realz, right? i.e. not just unit tests? 

👍

vivster7:
Yup -- consumed it with our viewscreen client.

PR #846, opened 2016-05-17
Don't compute task hash when Call Caching is turned of via Workflow Options Closes #837

### LINKED ISSUE #837: Call caching unbearably slow even when off
Relates to #836

With call caching on, but turned off at the workflow label, the behavior in #836 still occurs. This shouldn't be the case since we're *not call caching*

Horneth:
- [ ] Test it


mcovarr:
:+1:

mcovarr:
should really go against hotfix though

Horneth:
Against hotfix instead https://github.com/broadinstitute/cromwell/pull/847

PR #847, opened 2016-05-17
Turn call caching off for good when it needs to be Closes #837

### LINKED ISSUE #837: Call caching unbearably slow even when off

Horneth:


mcovarr:
:+1:

geoffjentry:
images

(and for pullapprove - 👍 )

PR #851, opened 2016-05-18
Metadata persistence.  Closes #788

### LINKED ISSUE #788: Implement Persistent Store for Metadata Service
depends on #784

mcovarr:


mcovarr:
@cjllanwarne and @geoffjentry

engine/src/main/scala/cromwell/services/MetadataServiceActor.scala
```diff
@@ -64,38 +44,26 @@ object MetadataServiceActor {
 
 case class MetadataServiceActor(serviceConfig: Config, globalConfig: Config) extends Actor {
 
-  private var journal = List.empty[MetadataEvent]
+  private val dataAccess = DataAccess.globalDataAccess
```
gauravs90: Maybe this should be injected via the constructor? Might facilitate a little better testing.

database/src/main/resources/changesets/metadata_journal.xml
```diff
@@ -0,0 +1,67 @@
...
+    <changeSet id="metadata_key_index" author="mcovarr">
+        <createIndex indexName="METADATA_KEY_IDX"
+                     tableName="METADATA_JOURNAL" unique="false">
+            <column name="WORKFLOW_EXECUTION_UUID"/>
+            <column name="METADATA_KEY"/>
+            <column name="METADATA_CALL_FQN"/>
+            <column name="METADATA_CALL_INDEX"/>
+            <column name="METADATA_CALL_ATTEMPT"/>
+        </createIndex>
+    </changeSet>
```
kshakir: Spec assertions for indexes differences are turned [off](https://github.com/broadinstitute/cromwell/blob/0.19/engine/src/main/scala/cromwell/engine/db/DiffResultFilter.scala#L31-L51) for the moment, until I have more time to investigate some issues when comparing slick vs liquibase schemas.

Still, please add the defined index above to the slick schema in `MetadataComponent`. See the `WorkflowExecutionComponent` for an example: [liquibase](https://github.com/broadinstitute/cromwell/blob/0.19/engine/src/main/resources/changesets/db_schema.xml#L135-L138) & then [slick](https://github.com/broadinstitute/cromwell/blob/0.19/engine/src/main/scala/cromwell/engine/db/slick/WorkflowExecutionComponent.scala#L41)

database/src/main/resources/changesets/metadata_journal.xml
```diff
@@ -0,0 +1,67 @@
...
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.3.xsd">
+
+    <changeSet author="mcovarr" id="metadata_journal">
+        <createTable tableName="METADATA_JOURNAL">
+            <column autoIncrement="true" name="METADATA_JOURNAL_ID" type="INT">
+                <constraints primaryKey="true" primaryKeyName="PK_METADATA_JOURNAL"/>
+            </column>
+            <column name="WORKFLOW_EXECUTION_UUID" type="VARCHAR(255)">
+                <constraints nullable="false"/>
+            </column>
```
kshakir: Sorry if I'm Out of the Loop: Why is this table referring to external string/UUID, instead of a standard foreign key `WORKFLOW_EXECTUION.WORKFLOW_EXECUTION_ID`? Kris mentioned that putting strings/UUIDs as primary keys sometimes causes performance problems. I don't know if it affects other index optimization, but it "smells funny" as we should already have the UUID normalized into an internal WORKFLOW_EXECUTION_ID.

As it's also not listed as a FK, I can only assume it's because for the metadata service we explicitly _do not_ want referential integrity checking? (We'll take ANY UUID, even one not for a workflow!) And/Or, it's also some sort of performance boost by explicitly not reaching into another table for denormalized querying?

mcovarr: Yup that's pretty much the idea, we explicitly don't want referential integrity checking.  Yesterday this branch was using WORKFLOW_EXECUTION_ID as a regular FK, but per discussion with @geoffjentry and @cjllanwarne this was refactored to the current model.  The calling APIs will at least enforce a WorkflowId is passed, but I'm not sure the construction of those is restricted in any meaningful way.

database/src/main/resources/changesets/metadata_journal.xml
mcovarr: will do

database/src/main/scala/cromwell/database/DbNullable.scala
```diff
@@ -0,0 +1,34 @@
+package cromwell.database
+
+/**
+  * Trait and typeclasses for converting to and from non-null representations of data in the database that is
+  * actually null in the domain model.
+  * This is used to work around https://bugs.mysql.com/bug.php?id=8173
+  */
+object DbNullable {
```
mcovarr: This is now almost entirely vestigial from an earlier stage of the PR where there was a unique index on the (workflowId + callFqn + index + attempt + key + timestamp) columns.  I ended up removing this unique index due to complications in dealing with constraint violations plus the [MySQL ugliness](https://bugs.mysql.com/bug.php?id=8173) involved with nullable columns in unique constraints.  I'd like to tech talk this tomorrow if possible.

database/src/main/scala/cromwell/database/slick/MetadataComponent.scala
```diff
@@ -0,0 +1,62 @@
...
+      metadatum <- metadata
+      if metadatum.workflowExecutionUuid === workflowExecutionUuid
+    } yield metadatum)
+
+  val metadataByWorkflowUuidAndKey = Compiled(
+    (workflowExecutionUuid: Rep[String], key: Rep[String]) => for {
+      metadatum <- metadata
+      if metadatum.workflowExecutionUuid === workflowExecutionUuid
+      if metadatum.key === key
+    } yield metadatum)
```
mcovarr: Now that I re-read this I'm not sure it's doing the right thing.  Should this be returning only workflow-scoped KVs (i.e. KVs with null callFqn, index, and scope) or every KV in this workflow having this key?  It'd doing the latter, but I'm not sure the former wasn't the intent of the API.

database/src/main/resources/changesets/metadata_journal.xml
geoffjentry: Right. The point behind these services was that they were a transitionary stage between our tightly coupled systems and a world where these could all be separated out completely. So the first pass was that we were going to use the same physical DB but pretend as if we weren't.

database/src/main/scala/cromwell/database/slick/SlickDatabase.scala
```diff
@@ -667,4 +668,61 @@ class SlickDatabase(databaseConfig: Config) extends SqlDatabase {
 
     runTransaction(action)
   }
+
+  override def addMetadataEvent(workflowUuid: String,
+                                key: String,
+                                value: String,
+                                timestamp: Timestamp)(implicit ec: ExecutionContext): Future[Unit] = {
+
+    val action = DBIO.seq(dataAccess.metadataAutoInc += Metadatum(workflowUuid, key, callFqn = None, index = None, attempt = None, Option(value), timestamp))
```
kshakir: `DBIO.seq` is an abomination, that reads to me as "[just run already!](http://slick.lightbend.com/doc/3.1.1/api/index.html#slick.dbio.DBIOAction$@seq[E<:slick.dbio.Effect](actions:slick.dbio.DBIOAction[_,slick.dbio.NoStream,E]*):slick.dbio.DBIOAction[Unit,slick.dbio.NoStream,E])" It consumes practically anything, and then returns `DBIO[Unit]`.<sup>1</sup>

When I've used it in the past, it was because I couldn't get anything else to work for a while. I've since associated it with code smell. I'm not sure if you ran into a similar issues. Perhaps the method return was argumentative, as you wanted to throw away the result of the insert and return `Unit`.

If that's the case, try mapping the result outside the transaction via:

```scala
val action = dataAccess.metadataAutoInc += Metadatum(workflowUuid, key,
  callFqn = None, index = None, attempt = None, Option(value), timestamp)
runTransaction(action).map(_ => ())
```

or map to a `Unit` using a for comprehension (easier to read IMHO, but `map`s inside the transaction):

```scala
val action = for {
  _ <- dataAccess.metadataAutoInc += Metadatum(workflowUuid, key,
    callFqn = None, index = None, attempt = None, Option(value), timestamp)
} yield ()
runTransaction(action)
```

---
<sup>1</sup> FYI, for _actually_ running multiple operations, `DBIO.seq`'s sibling [`DBIO.sequence`](http://slick.lightbend.com/doc/3.1.1/api/index.html#slick.dbio.DBIOAction$@sequence[R,M[+_]<:TraversableOnce[_],E<:slick.dbio.Effect](in:M[slick.dbio.DBIOAction[R,slick.dbio.NoStream,E]])(implicitcbf:scala.collection.generic.CanBuildFrom[M[slick.dbio.DBIOAction[R,slick.dbio.NoStream,E]],R,M[R]]):slick.dbio.DBIOAction[M[R],slick.dbio.NoStream,E]) is cool with me, as it behaves more like [`Future.sequence`](http://www.scala-lang.org/api/2.11.8/index.html#scala.concurrent.Future$@sequence[A,M[X]<:TraversableOnce[X]](in:M[scala.concurrent.Future[A]])(implicitcbf:scala.collection.generic.CanBuildFrom[M[scala.concurrent.Future[A]],A,M[A]],implicitexecutor:scala.concurrent.ExecutionContext):scala.concurrent.Future[M[A]]) and doesn't erase types by returning a wrapped `Unit`.

database/src/main/scala/cromwell/database/slick/MetadataComponent.scala
```diff
@@ -0,0 +1,62 @@
...
+
+  val metadataByWorkflowUuidAndCallFqnAndIndexAndAttempt = Compiled(
+    (workflowExecutionUuid: Rep[String], callFqn: Rep[String], index: Rep[Option[Int]], attempt: Rep[Int]) => for {
+      metadatum <- metadata
+      if metadatum.workflowExecutionUuid === workflowExecutionUuid
+      if metadatum.callFqn === callFqn
+      if (metadatum.index === index) || (metadatum.index.isEmpty && index.isEmpty)
```
kshakir: ToL

Interesting. Does this fail tests if the ` || (metadatum.index.isEmpty && index.isEmpty)` is removed?
I'm still learning slick, but I'm surprised one has do the extra explicit check for effectively `Rep[None] === Rep[None]` to avoid generating the busted SQL 'WHERE NULL = NULL'

database/src/main/scala/cromwell/database/slick/SlickDatabase.scala
geoffjentry: I think this might be the first time I've ever seen an actual footnote in a code review comment. Bravo @kshakir - you win the internets today.

core/src/main/scala/cromwell/core/KnowsWhatTimeItIs.scala
```diff
@@ -0,0 +1,9 @@
+package cromwell.core
+
+import java.sql.Timestamp
+import java.util.Date
+
+
```
geoffjentry: While this made me laugh I'd prefer something a little less tomfooleryish. OTOH I didn't really have a better suggestion in the more serious category. At least you didn't call it FlavaFlav I suppose :)

flavaflav_main1

database/src/main/scala/cromwell/database/slick/MetadataComponent.scala
geoffjentry: Without thinking about it I would have said this was right but that's a good point. Perhaps something else to tech talk tomorrow?

engine/src/main/scala/cromwell/services/package.scala
```diff
@@ -0,0 +1,30 @@
+package cromwell
+
+import java.sql.Timestamp
+
+import cromwell.backend.BackendJobDescriptorKey
+import cromwell.core.WorkflowId
+import cromwell.engine.ExecutionIndex._
+
+
+package object services {
```
geoffjentry: I'd prefer to keep the package objects extremely slim. Perhaps put this elsewhere? Since this is scala and not java they could even just live as-is in a single file. 

I don't really care that much about this specific case though, so feel free to disregard. It's more of a broken windows thing at the moment.

geoffjentry:
modulo some tech talkin' - 👍 

cat-thumbs-up

core/src/main/scala/cromwell/core/KnowsWhatTimeItIs.scala
mcovarr: This is *exactly* what I was thinking

geoffjentry: Too bad we can't have images as identifiers

engine/src/main/scala/cromwell/services/package.scala
mcovarr: I'm okay with whatever organization scheme is agreeable. Why don't we chat about what exactly we want and I'll refactor accordingly.

database/src/main/scala/cromwell/database/slick/SlickDatabase.scala
mcovarr: Yes this was fumbling on my part to get syntax that would work.  And wow on the comment, it seems a shame to fix the code and hide it.  😄

database/src/main/scala/cromwell/database/slick/MetadataComponent.scala
mcovarr: Yes, MetadataActorSpec fails with these removed (there's another instance below).  It took me a while last night to figure out what the problem was, and maybe there's a better way of handling this than what I have now, but something more than just `===`

database/src/main/resources/changesets/metadata_journal.xml
```diff
@@ -0,0 +1,67 @@
...
+            <append value=" ENGINE=INNODB"/>
+        </modifySql>
+    </changeSet>
+
+    <changeSet id="metadata_workflow_index" author="mcovarr">
+        <createIndex indexName="METADATA_WORKFLOW_IDX"
+                     tableName="METADATA_JOURNAL" unique="false">
+            <column name="WORKFLOW_EXECUTION_UUID"/>
+        </createIndex>
+    </changeSet>
```
mcovarr: missed an index for workflowId + key, assuming we still want that per the comments below

engine/src/main/scala/cromwell/services/MetadataServiceActor.scala
mcovarr: I can make the change, but the same globalDataAccess is going to be passed in at least for the short term.

engine/src/main/scala/cromwell/services/package.scala
```diff
@@ -0,0 +1,30 @@
...
+
+  case class MetadataKey(workflowId: WorkflowId, jobKey: Option[MetadataJobKey], key: String)
+  case class MetadataValue(value: String)
+  case class MetadataEvent(key: MetadataKey, value: MetadataValue, timestamp: Timestamp)
+
+  case class MetadataQueryJobKey(callFqn: String, index: ExecutionIndex, attempt: Int)
+  object MetadataQueryJobKey {
+    def forMetadataJobKey(jobKey: MetadataJobKey) = MetadataQueryJobKey(jobKey.callFqn, jobKey.index, jobKey.attempt)
+  }
+  case class MetadataQuery(workflowId: WorkflowId, jobKey: Option[MetadataQueryJobKey], key: Option[String])
```
mcovarr: Note!  I changed MetadataQuery from having a parameter `Option[WorkflowId]` to just `WorkflowId` since I didn't think there was a legitimate use case for a workflow ID-less query.  All the MetadataBuilder* changes here are a consequence of this.

database/src/main/scala/cromwell/database/slick/MetadataComponent.scala
cjllanwarne: Since they'd eventually show up in totally different parts of the metadata JSON, I would guess the former is probably most right. 

But OTOH I suspect the actual number of cases where keys are the same at workflow vs call scope are pretty minimal

database/src/main/scala/cromwell/database/slick/SlickDatabase.scala
```diff
@@ -15,6 +15,7 @@ import slick.driver.JdbcProfile
 import scala.concurrent.duration._
 import scala.concurrent.{Await, ExecutionContext, Future}
 
+
```
cjllanwarne: Superfluous newline?

cjllanwarne:
To keep ourselves honest, and convince ourselves that this is a swappable service rather than a core part of cromwell, I would like to see all of the metadata database access stuff in the `services.metadata` package rather than the engine's `dataAccess` parts

engine/src/main/scala/cromwell/services/MetadataServiceActor.scala
```diff
@@ -57,45 +37,31 @@ object MetadataServiceActor {
...
 
-  def props(serviceConfig: Config, globalConfig: Config) = {
-    Props(MetadataServiceActor(serviceConfig, globalConfig))
+  def props(dataAccess: DataAccess, serviceConfig: Config, globalConfig: Config) = {
+    Props(MetadataServiceActor(dataAccess, serviceConfig, globalConfig))
   }
 }
 
-case class MetadataServiceActor(serviceConfig: Config, globalConfig: Config) extends Actor {
+case class MetadataServiceActor(dataAccess: DataAccess, serviceConfig: Config, globalConfig: Config) extends Actor {
```
cjllanwarne: As in my global comment, I personally would prefer this service's `dataAccess` to be distinct from the engine's `dataAccess`

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
```diff
@@ -60,7 +60,7 @@ trait CromwellApiService extends HttpService with PerRequestCreator with Service
           case Success(w) =>
             version match {
               case "v2" =>
```
cjllanwarne: I suggest we bravely make `v2` the only option.

engine/src/test/scala/cromwell/webservice/MetadataBuilderActorSpec.scala
```diff
@@ -1,26 +1,34 @@
...
+  with FlatSpecLike with Matchers with TableDrivenPropertyChecks with ImplicitSender with BeforeAndAfterAll {
 
   behavior of "MetadataParser"
 
   val defaultTimeout = 100 millis
   val mockServiceRegistry = TestProbe()
   val parentProbe = TestProbe()
 
+  // TestKit has its own `now` that prevents extending KnowsWhatTimeItIs on this class directly.
+  val knowsWhatTimeItIs = new Object with KnowsWhatTimeItIs
```
cjllanwarne: That's unfortunate! Maybe rename the def from from `now` to instead be `timeThatIKnowItIs`?

engine/src/test/scala/cromwell/webservice/MetadataBuilderActorSpec.scala
```diff
@@ -255,11 +249,11 @@ class MetadataBuilderActorSpec extends TestKit(ActorSystem("Metadata")) with Fla
...
     val workflow = WorkflowId.randomId()
-    val query = MetadataQuery(Option(workflow), None, Option("status"))
+    val query = MetadataQuery(workflow, None, Option("status"))
     val events = List(
-      MetadataEvent(MetadataKey(workflow, None, "status"), MetadataValue("Running"), DateTime.parse("2000-01-02")),
-      MetadataEvent(MetadataKey(workflow, None, "status"), MetadataValue("Done"), DateTime.parse("2000-01-03")),
-      MetadataEvent(MetadataKey(workflow, None, "status"), MetadataValue("Submitted"), DateTime.parse("2000-01-01"))
+      MetadataEvent(MetadataKey(workflow, None, "status"), MetadataValue("Running"), Timestamp.valueOf("2000-01-02 12:00:00")),
+      MetadataEvent(MetadataKey(workflow, None, "status"), MetadataValue("Done"), Timestamp.valueOf("2000-01-03 12:00:00")),
+      MetadataEvent(MetadataKey(workflow, None, "status"), MetadataValue("Submitted"), Timestamp.valueOf("2000-01-01 12:00:00"))
```
cjllanwarne: By setting these timestamps to be the same, are you deliberately making this non-deterministic?

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
geoffjentry: I agree however I think this should be a PO decision

engine/src/test/scala/cromwell/webservice/MetadataBuilderActorSpec.scala
mcovarr: My changes here were only to switch from Joda DateTimes to Timestamps, I'm not sure the original intent.  @Horneth may be able to answer your question.

engine/src/main/scala/cromwell/services/MetadataServiceActor.scala
mcovarr: This used to be a field in the MetadataServiceActor and I was asked to make it a constructor parameter.  The engine currently does not use a DataAccess itself.  Not sure what I should do at this point.  😦

database/src/main/scala/cromwell/database/slick/SlickDatabase.scala
mcovarr: This is how I roll, putting two spaces between the import preamble and the meat of the code that follows.  There are probably dozens of files like this.

engine/src/test/scala/cromwell/webservice/MetadataBuilderActorSpec.scala
Horneth: The dates are different

engine/src/main/scala/cromwell/services/MetadataServiceActor.scala
```diff
@@ -57,45 +37,31 @@ object MetadataServiceActor {
   case class MetadataLookupResponse(query: MetadataQuery, eventList: Seq[MetadataEvent]) extends MetadataServiceResponse
   case class MetadataServiceKeyLookupFailed(query: MetadataQuery, reason: Throwable) extends MetadataServiceResponse
 
-  def props(serviceConfig: Config, globalConfig: Config) = {
-    Props(MetadataServiceActor(serviceConfig, globalConfig))
+  def props(dataAccess: DataAccess, serviceConfig: Config, globalConfig: Config) = {
```
Horneth: TOL, is there no possible use case where we would want to actually have an in-memory MetadataService ?
Since they're swappable via the conf file anyway, could we have both an  `InMemoryMetadataActor` and a `PersistenceMetadataActor` living side by side ?

mcovarr: HSQLDB as we use it is in-memory, or did you mean something else?

geoffjentry: I was picturing it as in 'not a database' when he said that, FWIW

Horneth: Yeah I just meant leave the implementation Chris wrote as it is, and put this one beside it, instead of overwriting it. But since HSQLDB is in memory they're pretty much equivalent I guess

mcovarr: If we have different dataAccesses for engine and this service, this could be in-memory while the engine might not be.  Then again it seems like almost anything interesting would be stored via this service.

Is there a change I should make here for this?

Horneth: No it's ok for me like this 👍

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
mcovarr: He might think that's a brave proposal

engine/src/test/scala/cromwell/webservice/MetadataBuilderActorSpec.scala
mcovarr: Yeah I need to do something with this, mixing into a class that already has TestKit is not going to be a weird edge case.

mcovarr:
@cjllanwarne does this perchance rate a thumb now?

geoffjentry:
I'm picturing @cjllanwarne now ...

200_s

cjllanwarne:
👍

PR #853, opened 2016-05-18
Reuse output hashes for scatter collection Closes #836

### LINKED ISSUE #836: Call caching unbearably slow
When you launch a wf for the first time w/ caching on, it can take forever for the cromwell final stuff.

Horneth:


engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -272,12 +273,22 @@ case class WorkflowActor(workflow: WorkflowDescriptor)
...
-      val arrayOfValues = new WdlArray(WdlArrayType(taskOutput.wdlType), wdlValues)
-      taskOutput.name -> CallOutput(arrayOfValues, workflow.hash(arrayOfValues))
+      val callOutputs = shardsOutputs.map(s => s.getOrElse(taskOutput.name, throw new RuntimeException(s"Could not retrieve output ${taskOutput.name}")))
+      val arrayOfValues = new WdlArray(WdlArrayType(taskOutput.wdlType), callOutputs map { _.wdlValue })
+      val hashes = callOutputs map { output => {
+        output.hash orElse workflow.hash(output.wdlValue)
+      } }
+      val hash = if (hashes forall { _.isDefined }) {
+        val concatenatedHashes = hashes map { _.get.value } mkString ""
+        val hash = SymbolHash((WdlArray.getClass.getCanonicalName + concatenatedHashes).md5Sum)
```
Horneth: This is very dirty, it emulates the aggregation logic in wdl4s, which is private and can't be re-used. A cleaner way would be to patch wdl4s 0.4 to factor out this logic and update the dependency here in 0.19_hotfix.

Horneth:
@jsotobroad I haven't run this at scale on JES, but with some small local tests I'm pretty sure that's the reason why scatter collection is so slow with call caching on.
We basically re-compute hashes for all the elements in the array when the shards are collected...
I ran without and with this fix and scatter collection went from ~75ms to ~2ms.

Horneth:
@cjllanwarne @mcovarr

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
geoffjentry: Do you see the same need in ye newe worlde? If so make the wld4s change, if not no

cjllanwarne:
👍

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -272,12 +273,22 @@ case class WorkflowActor(workflow: WorkflowDescriptor)
...
+      fetchCallOutputs(e) getOrElse(throw new RuntimeException(s"Could not retrieve output for shard ${e.scope} #${e.index}"))
     }
     collector.scope.task.outputs map { taskOutput =>
-      val wdlValues = shardsOutputs.map(s => s.getOrElse(taskOutput.name, throw new RuntimeException(s"Could not retrieve output ${taskOutput.name}")))
-      val arrayOfValues = new WdlArray(WdlArrayType(taskOutput.wdlType), wdlValues)
-      taskOutput.name -> CallOutput(arrayOfValues, workflow.hash(arrayOfValues))
+      val callOutputs = shardsOutputs.map(s => s.getOrElse(taskOutput.name, throw new RuntimeException(s"Could not retrieve output ${taskOutput.name}")))
+      val arrayOfValues = new WdlArray(WdlArrayType(taskOutput.wdlType), callOutputs map { _.wdlValue })
+      val hashes = callOutputs map { output => {
+        output.hash orElse workflow.hash(output.wdlValue)
```
mcovarr: If the outputs weren't hashed for the shards before, is there a circumstance where `workflow.hash` will cause that to happen now?

mcovarr: maybe drop a comment in here as to reusing the hashes to avoid needless work?

mcovarr:
:+1:

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
Horneth: Yeah I don't think that should ever happen, if there's no hash it should be because `writeToCache` is false, which means `workflow.hash(output.wdlValue)` will not compute the hash either this time.

Horneth: I'm just going to remove the orElse I think as it's confusing and might lead to weird behaviour if `workflow.hash` gets changed

Issue #852, opened 2016-05-18
Merge Local and HTCondor backend common functionality


francares:
Merge Local and HTCondor backend common functionality.
Move also WorkflowPaths to Backend component.

PR #861, opened 2016-05-19
Adding initial version of HtCondor. Closes #660

### LINKED ISSUE #660: Implement HT Backend

gauravs90:
Also creates #852.

Following things need to be done (separate stories?) :

- [ ] Add docker support (#884)
- [ ] Add recovery and abort (#885)
- [x] Add continueOnErrorCode support
- [ ] Find a way to add condor specific runtime attributes to Condor ClassAds (#886)

Anything more to support?


gauravs90:
@francares and @Horneth

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/CondorJobExecutionActor.scala
```diff
@@ -0,0 +1,195 @@
...
+  val runtimeAttributes = {
+    val evaluateAttrs = call.task.runtimeAttributes.attrs mapValues (_.evaluate(lookup, callEngineFunction))
+    // Fail the call if runtime attributes can't be evaluated
+    TryUtil.sequenceMap(evaluateAttrs, "Runtime attributes evaluation").get.mapValues(_.valueString)
+  }
+
+  /**
+    * Restart or resume a previously-started job.
+    */
+  override def recover: Future[BackendJobExecutionResponse] = Future.failed(new Exception("HtCondorBackend currently doesn't support recovering jobs."))
```
mcovarr: I thought if a job couldn't be recovered it would just be run anew.

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/CondorJobExecutionActor.scala
```diff
@@ -0,0 +1,195 @@
...
+  /**
+    * Execute a new job.
+    */
+  override def execute: Future[BackendJobExecutionResponse] = Future(executeTask())
+
+  /**
+    * Abort a running job.
+    */
+  override def abortJob(): Unit = Future.failed(new Exception("HtCondorBackend currently doesn't support aborting jobs."))
+
```
geoffjentry: UnsupportedOperationException?

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/CondorJobExecutionActor.scala
```diff
@@ -0,0 +1,195 @@
...
+  override def preStart(): Unit = {
+    log.debug(s"Creating execution folder: $executionDir")
+    executionDir.toString.toFile.createIfNotExists(true)
```
geoffjentry: As per The Zen Of Akka, don't use string interp for akka logging (I just learned this last week)

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/CondorJobExecutionActor.scala
```diff
@@ -0,0 +1,195 @@
...
+    try {
+      val localizedInputs = localizeInputs(jobPaths, docker = false, fileSystems, jobDescriptor.inputs)
```
geoffjentry: Don't put unnamed args after named args

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/CondorJobExecutionActor.scala
gauravs90: So you mean doing a `val msg = s"..."` and then `log.debug(msg)` ? Although [documentation](http://doc.akka.io/api/akka/2.0/akka/event/LoggingAdapter.html) does say they support upto 4 simple string interpolation stuff.

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/CondorJobExecutionActor.scala
```diff
@@ -0,0 +1,195 @@
...
+
+    List(stdoutWriter.writer, stderrWriter.writer).foreach(_.flushAndClose())
+    log.debug(s"done flushing")
+    val stderrFileLength = Try(Files.size(stderrPath)).getOrElse(0L)
+    log.debug(s"stderr file length : $stderrFileLength ")
+
+    val rc = Try(returnCodePath.contentAsString.stripLineEnd.toInt)
+    log.debug(s"received return code: $rc ")
+
+    if (processReturnCode != 0) {
```
gauravs90: This if else block is really bad. I will try to find a better way to rewrite this. Functionality first 😓

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/CondorJobExecutionActor.scala
geoffjentry: It's just a slight performance difference between `log.debug(s"blah blah $foo")` and `log.debug("blah blah " + foo)`. In the first case it's evaluated no matter if the log message is emitted or not and in the latter case it's not. The way you described it would do away with both. It's minor.

geoffjentry: Slide #31: http://www.slideshare.net/ktoso/zen-of-akka
or around 10:40 at: https://www.youtube.com/watch?v=tC-joPMPJLs

Horneth: They actually say to use `log.debug("blah blah {}", foo)`
I feel like `log.debug("blah blah " + foo)` would be evaluated too no ?

geoffjentry: Good catch. You can see how much I pay attention.

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/CondorJobExecutionActor.scala
```diff
@@ -0,0 +1,186 @@
...
+                              override val configurationDescriptor: BackendConfigurationDescriptor) extends BackendJobExecutionActor with SharedFileSystem {
+
+
+  import CondorJobExecutionActor._
+  import cromwell.core.PathFactory._
+  import better.files._
+
+  lazy val cmds = new HtCondorCommands {}
+  lazy val extProcess = new HtCondorProcess {}
+  lazy val parser = new HtCondorClassAdParser {}
```
Horneth: Why not make those (`HtCondorCommands`...) classes instead of traits if you want to instantiate them directly ?

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/CondorWrapper.scala
```diff
@@ -0,0 +1,142 @@
...
+  def statusCommand(): String = s"${HtCondorCommands.condor_queue} -xml"
+}
+
+trait HtCondorClassAdParser extends StrictLogging {
+
+  def xmlParser(classAd: String, condorJobId: CondorJobId): Int = {
+    //This is required to turn off external validation of dtd in xml
+    val factory = SAXParserFactory.newInstance()
+    factory.setValidating(false)
+    factory.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false)
```
Horneth: You can probably extract that out and re-use the same factory every time

gauravs90: I actually removed the whole thing.

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/CondorJobExecutionActor.scala
gauravs90: Yep..just did that in my latest refactor.

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/CondorJobExecutionActor.scala
gauravs90: Thanks! Didn't know about that 😄

engine/src/main/resources/application.conf
```diff
@@ -139,7 +139,39 @@ backend {
           }
         }
       }
-      //    },
+    }
+//    ,
+//    HtCondor {
```
francares: Why is this commented?

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/CondorBackendFactory.scala
```diff
@@ -0,0 +1,38 @@
...
+import cromwell.backend._
+import cromwell.core.CallContext
+import wdl4s.Call
+import wdl4s.expression.WdlStandardLibraryFunctions
+
+case class CondorBackendFactory(config: Config) extends BackendLifecycleActorFactory {
+  override def workflowInitializationActorProps(workflowDescriptor: BackendWorkflowDescriptor,
+                                                calls: Seq[Call],
+                                                configurationDescriptor: BackendConfigurationDescriptor): Option[Props] = {
+    Option(CondorInitializationActor.props(workflowDescriptor, calls, configurationDescriptor))
```
francares: I think in the meantime this should be None. Later, when there is a need we can add that.

francares:
I think we should create USs for the other missing features. I think this code is OK as a first iteration for HtCondor backend.

jainh:
👍

Horneth:
👍

engine/src/main/resources/application.conf
gauravs90: For some reason the tests fail complaining that `ClassNotFound`if any of the other backends are uncommented. I have not gone into the details of that yet.

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/CondorBackendFactory.scala
gauravs90: Just saw it..it's doing some validations for the runtime attributes.. I suggest we leave this as is.

PR #862, opened 2016-05-20
Someone doesn't understand how Futures work. Hint, it's me

geoffjentry:
@mcovarr @cjllanwarne - given that you all are hip to the topic, can you take a look and see if looks right?

### DIFF
--- a/engine/src/main/scala/cromwell/engine/backend/jes/JesBackend.scala
+++ b/engine/src/main/scala/cromwell/engine/backend/jes/JesBackend.scala
@@ -129,7 +129,7 @@ object JesBackend {
     *
     * Expect a more unifying solution in the non-hotfix branch
     */
-  def retryAsync[A](f: => Future[A],
+  def retryAsync[A](f: () => Future[A],
                     logger: WorkflowLogger,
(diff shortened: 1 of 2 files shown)


engine/src/main/scala/cromwell/engine/backend/jes/JesBackend.scala
```diff
@@ -141,15 +141,15 @@ object JesBackend {
     val delay = backoff.backoffMillis.millis
 
     if (retries > 0) {
-      f recoverWith {
+      f() recoverWith {
```
geoffjentry: I suppose I could pull the `f()` into a single spot. I'll leave it up to reviewers.

mcovarr: I think it's fine, particularly since I did it this way on develop.  😛

mcovarr:
:+1: assuming Travis approves

geoffjentry:
@mcovarr Can't assume anything like that these days, at least when it comes to me.

cjllanwarne:
👍

PR #864, opened 2016-05-20
Remove Pipeline and use ephemeral Pipelines in JES. Closes #856

### LINKED ISSUE #856: Switch JES to use ephemeral pipelines
From Dion, regarding how we construct a Pipeline for every Run:

Could you convert that into the ephemeral Run version?   RunPipelineRequest.ephemeral_pipeline instead of pipeline_id. Will save you one RPC call too.

geoffjentry:


geoffjentry:
@mcovarr @Horneth - the code is slightly different than the hotfix to account for slightly different realities (plus a moving-stuff-around refactor) but more or less the same.

Horneth:
👍

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/RunStatus.scala
```diff
@@ -0,0 +1,29 @@
+package cromwell.backend.impl.jes
+
+sealed trait RunStatus {
+  import RunStatus._
+
+  // Could be defined as false for Initializing and true otherwise, but this is more defensive.
+  def isRunningOrComplete = this match {
+    case Running | _: TerminalRunStatus => true
```
mcovarr: that `_:` looks suspiciously superfluous

mcovarr:
:+1:

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/RunStatus.scala
geoffjentry: Indeed. Someone named @mcovarr did it, we should ask him 😝

geoffjentry: Appears to simply look suspicious. The original culprit is vindicated. :shipit:

PR #868, opened 2016-05-21
First pass at re-enabling tests post-PBE.  This will be iterative but…

kcibul:
… all moving in the right direction.  Test Infrastructure updated to handle the test re-enabled here, including ThreeStep

### DIFF
--- a/engine/src/test/scala/cromwell/ArrayAndMapIndexingWorkflowSpec.scala
+++ b/engine/src/test/scala/cromwell/ArrayAndMapIndexingWorkflowSpec.scala
@@ -10,10 +10,10 @@ import scala.language.postfixOps
 
 class ArrayAndMapIndexingWorkflowSpec extends CromwellTestkitSpec {
   "A workflow with array/map indexes in expressions" should {
-    "run locally" ignore {
+    "run locally" in {
       runWdlAndAssertOutputs(
         sampleWdl = SampleWdl.ArrayAndMapIndexingWdl,
-        eventFilter = EventFilter.info(pattern = s"starting calls: test.echo_int, test.echo_str", occurrences = 1),
+        eventFilter = EventFilter.info(pattern = "Workflow complete", occurrences = 1),
         expectedOutputs = Map(
           "test.echo_str.o" -> WdlString("bar"),
           "test.echo_int.o" -> WdlInteger(200)
--- a/engine/src/test/scala/cromwell/CromwellTestkitSpec.scala
+++ b/engine/src/test/scala/cromwell/CromwellTestkitSpec.scala
@@ -1,6 +1,6 @@
 package cromwell
 
-import akka.actor.{ActorRef, ActorSystem, Props}
+import akka.actor.{Actor, ActorRef, ActorSystem, Props}
 import akka.pattern.ask
 import akka.testkit._
 import akka.util.Timeout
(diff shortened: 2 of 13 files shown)


engine/src/test/scala/cromwell/ArrayAndMapIndexingWorkflowSpec.scala
```diff
@@ -10,10 +10,10 @@ import scala.language.postfixOps
 
 class ArrayAndMapIndexingWorkflowSpec extends CromwellTestkitSpec {
   "A workflow with array/map indexes in expressions" should {
-    "run locally" ignore {
+    "run locally" in {
       runWdlAndAssertOutputs(
         sampleWdl = SampleWdl.ArrayAndMapIndexingWdl,
-        eventFilter = EventFilter.info(pattern = s"starting calls: test.echo_int, test.echo_str", occurrences = 1),
+        eventFilter = EventFilter.info(pattern = "Workflow complete", occurrences = 1),
```
kcibul: previously this was checking some, but not all of the calls starting (generally) in all the workflow tests.  I don't think it was adding much value so now it is just checking for the log message of Workflow complete.  As we invest more in the logging of the call actors, etc we can beef these up if we want

engine/src/test/scala/cromwell/CromwellTestkitSpec.scala
```diff
@@ -445,22 +440,44 @@ abstract class CromwellTestkitSpec extends TestKit(new CromwellTestkitSpec.TestW
...
+          originalSender ! m
+        case m: GetMetadataQueryAction =>
+          originalSender = sender
+          val mdba = TestActorRef(MetadataBuilderActor.props(serviceRegistryActor), self, "mdba" )
+          mdba ! m
+      }
+    })
+
+    val message = GetMetadataQueryAction(MetadataQuery(Option(workflowId), None, key))
+    Await.result(supervisor.ask(message).mapTo[RequestComplete[(String,JsObject)]], Duration.Inf).response._2
```
kcibul: idiomatic scala comments welcome

engine/src/test/scala/cromwell/CromwellTestkitSpec.scala
```diff
@@ -445,22 +440,44 @@ abstract class CromwellTestkitSpec extends TestKit(new CromwellTestkitSpec.TestW
...
     // Continuously check the state of the workflow until it is in a terminal state
-    awaitCond(pollWorkflowState(wma, workflowId).isTerminal)
+    awaitCond(getWorkflowState(workflowId).isTerminal)
     // Now that it's complete verify that we ended up in the state we're expecting
-    pollWorkflowState(wma, workflowId) should equal (expectedState)
+    getWorkflowState(workflowId) should equal (expectedState)
+  }
+
+  def getWorkflowState(workflowId: WorkflowId)(implicit ec: ExecutionContext): WorkflowState = {
+    WorkflowState.fromString(getWorkflowMetadata(workflowId, Option(WorkflowMetadataKeys.Status)).getFields(WorkflowMetadataKeys.Status).headOption.map( _.asInstanceOf[JsString].value).getOrElse("Submitted"))
```
kcibul: idiomatic scala comments welcome

engine/src/test/scala/cromwell/CromwellTestkitSpec.scala
```diff
@@ -445,22 +440,44 @@ abstract class CromwellTestkitSpec extends TestKit(new CromwellTestkitSpec.TestW
...
   }
 
-  private def pollWorkflowState(wma: ActorRef, workflowId: WorkflowId)(implicit ec: ExecutionContext): WorkflowState = {
-    val futureResponse = wma.ask(OldStyleWorkflowManagerActor.WorkflowStatus(workflowId)).mapTo[WorkflowManagerStatusSuccess]
-    Await.result(futureResponse map { _.state }, timeoutDuration)
+  def getWorkflowOutputs(id: WorkflowId): Map[FullyQualifiedName, String] = {
+    getWorkflowMetadata(id).getFields(WorkflowMetadataKeys.Outputs).head.asInstanceOf[JsObject].fields.map( x => (x._1,x._2.asInstanceOf[JsString].value))
```
kcibul: idiomatic scala comments welcome

kcibul:
Rebased went awry,  will fix and repush this afternoon

kcibul:
rebasing to develop is causing all the tests to timeout :(

engine/src/test/scala/cromwell/ArrayAndMapIndexingWorkflowSpec.scala
mcovarr: That assertion was in place to make sure the "find runnable calls" logic was finding the expected calls (in the very early days of Cromwell it wasn't a given that that logic was working correctly 😄 ).  Hopefully there's a more truly "unit" test for that functionality now and tests like this only need to concern themselves with asserting expected status and outputs.

PR #873, opened 2016-05-23
Only publish relevant state transition to metadata

Horneth:

PR #874, opened 2016-05-23
CRDT WorkflowState

mcovarr:
This works on Kristian's testing branch without Thibault's #873 fixes and will also be useful for the forthcoming Query API work.

### DIFF
--- a/engine/src/main/scala/cromwell/engine/WorkflowState.scala
+++ b/engine/src/main/scala/cromwell/engine/WorkflowState.scala
@@ -1,7 +1,10 @@
 package cromwell.engine
 
+import scalaz.Semigroup
+
 sealed trait WorkflowState {
   def isTerminal: Boolean
+  def append(that: WorkflowState): WorkflowState
 }
 
 object WorkflowState {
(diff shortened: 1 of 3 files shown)


mcovarr:
@geoffjentry and @Horneth

engine/src/test/scala/cromwell/webservice/MetadataBuilderActorSpec.scala
```diff
@@ -248,13 +248,13 @@ class MetadataBuilderActorSpec extends TestKit(ActorSystem("Metadata"))
     val query = MetadataQuery(workflow, None, Option("status"))
     val events = List(
       MetadataEvent(MetadataKey(workflow, None, "status"), MetadataValue("Running"), Timestamp.valueOf("2000-01-02 12:00:00")),
-      MetadataEvent(MetadataKey(workflow, None, "status"), MetadataValue("Done"), Timestamp.valueOf("2000-01-03 12:00:00")),
+      MetadataEvent(MetadataKey(workflow, None, "status"), MetadataValue("Succeeded"), Timestamp.valueOf("2000-01-03 12:00:00")),
```
Horneth: Maybe use identical timestamps for Running and Succeeded ?

mcovarr: good idea

engine/src/main/scala/cromwell/engine/WorkflowState.scala
```diff
@@ -1,42 +1,55 @@
...
 case object WorkflowFailed extends WorkflowState {
   override val toString: String = "Failed"
   override val isTerminal = true
+  override def append(that: WorkflowState): WorkflowState = this
 }
 
 case object WorkflowSucceeded extends WorkflowState {
   override val toString: String = "Succeeded"
   override val isTerminal = true
+  override def append(that: WorkflowState): WorkflowState = if (that == WorkflowFailed) that else this
```
Horneth: Succeeded to Failed is allowed ?

engine/src/main/scala/cromwell/engine/WorkflowState.scala
```diff
@@ -1,42 +1,55 @@
...
 }
 
 case object WorkflowAborted extends WorkflowState {
   override val toString: String = "Aborted"
   override val isTerminal = true
+  override def append(that: WorkflowState): WorkflowState = if (that.isTerminal) that else this
```
Horneth: Aborted to Succeeded or Failed also allowed ?

engine/src/main/scala/cromwell/engine/WorkflowState.scala
mcovarr: Really anything terminal to anything terminal should never happen.  I thought it would be better to call attention to this situation by using the most alarming terminal state actually seen, but I could be convinced otherwise.  😄

Horneth:
👍

engine/src/main/scala/cromwell/engine/WorkflowState.scala
```diff
@@ -1,42 +1,55 @@
...
+
+  implicit val WorkflowStateSemigroup = new Semigroup[WorkflowState] {
+    override def append(f1: WorkflowState, f2: => WorkflowState): WorkflowState = f1.append(f2)
+  }
 }
 
 case object WorkflowSubmitted extends WorkflowState {
   override val toString: String = "Submitted"
   override val isTerminal = false
+  override def append(that: WorkflowState): WorkflowState = that
```
geoffjentry: aside (and it was already like this and I believe you disagree anyways) but cases like this where there are multiple things to override and they're all trivial is exactly where I prefer the fp/pattern match style of ADT-ing vs all of the overrides and the logic in multiple places

engine/src/main/scala/cromwell/engine/WorkflowState.scala
geoffjentry: @Horneth If everything actually succeeded prior to the abort being registered perhaps?

engine/src/main/scala/cromwell/engine/WorkflowState.scala
geoffjentry: Thinking about this a bit I'm wondering if this might make sense to return some sort of error. Dunno.

geoffjentry:
👍

mcovarr:
I'll hold off merging this until we tech-talk the Failed + Succeeded scenario.

PR #879, opened 2016-05-24
Added Call-Metadata MetadataService calls from the engine

cjllanwarne:
Merging into Branch ks_jes_return_of_the_metadata because I originally branched from it.

We can either review this here, or as part of @kshakir's later commit of his branch (or both)?


kshakir:
Follow changes/reviews on #883

cjllanwarne:
Will move this PR against develop when #883 merges

PR #893, opened 2016-05-25
Now I *know* I'm having fun!! Closes #753

### LINKED ISSUE #753: Abort support for JES PBE

scottfrazer:
Don't flog me yet, I will clean this up.  There's about a week of cruft from trying various things in here...

*Diagram Disclaimer: the below diagrams are now obsolete!*

FAQ:

Q: Why did you change stuff unrelated to aborts?
A: Probably because I had a hard time debugging and it facilitated debugging.
A: Bug fixes
A: Probably no good reason, and it might just be cruft changes

This is how workflow aborts work for the LOCAL backend!

workflow aborts local - new page

This is how workflow aborts work for the JES backend!

workflow aborts jes - new page

:-D

### DIFF
--- a/backend/src/main/scala/cromwell/backend/package.scala
+++ b/backend/src/main/scala/cromwell/backend/package.scala
@@ -44,7 +44,9 @@ package object backend {
   case class BackendWorkflowDescriptor(id: WorkflowId,
                                        workflowNamespace: NamespaceWithWorkflow,
                                        inputs: Map[FullyQualifiedName, WdlValue],
-                                       workflowOptions: WorkflowOptions)
+                                       workflowOptions: WorkflowOptions) {
+    override def toString: String = s"[BackendWorkflowDescriptor id=${id.shortString} workflowName=${workflowNamespace.workflow.unqualifiedName}]"
+  }
 
   /**
     * For passing to a BackendActor construction time
--- a/engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
@@ -76,7 +76,7 @@ object WorkflowActor {
     * The WorkflowActor is aborting. We're just waiting for everything to finish up then we'll respond back to the
     * manager.
     */
-  case object AbortingWorkflowState extends WorkflowActorState {
+  case object WorkflowAbortingState extends WorkflowActorState {
     override val workflowState = WorkflowAborting
   }
 
@@ -135,7 +135,7 @@ class WorkflowActor(workflowId: WorkflowId,
                     serviceRegistryActor: ActorRef)
   extends LoggingFSM[WorkflowActorState, WorkflowActorData] with ActorLogging with KnowsWhatTimeItIs {
 
-  val tag = self.path.name
+  val tag = s"WorkflowActor [UUID(${workflowId.shortString})]"
 
   implicit val actorSystem = context.system
   implicit val ec = context.dispatcher
@@ -146,20 +146,17 @@ class WorkflowActor(workflowId: WorkflowId,
 
   when(WorkflowUnstartedState) {
     case Event(StartWorkflowCommand, _) =>
-      val actor = context.actorOf(MaterializeWorkflowDescriptorActor.props())
+      val actor = context.actorOf(MaterializeWorkflowDescriptorActor.props(), s"MaterializeWorkflowDescriptorActor-$workflowId")
       actor ! MaterializeWorkflowDescriptorCommand(workflowId, workflowSources, conf)
       goto(MaterializingWorkflowDescriptorState) using stateData.copy(currentLifecycleStateActor = Option(actor))
-    case Event(AbortWorkflowCommand, stateData) =>
-      // No lifecycle sub-actors exist yet, so no indirection via WorkflowAbortingState is necessary:
-      sender ! WorkflowAbortedResponse(workflowId)
-      goto(WorkflowAbortedState)
+    case Event(AbortWorkflowCommand, stateData) => goto(WorkflowAbortedState)
   }
 
   when(MaterializingWorkflowDescriptorState) {
     case Event(MaterializeWorkflowDescriptorSuccessResponse(workflowDescriptor), stateData) =>
       val initializerActor = context.actorOf(WorkflowInitializationActor.props(workflowId, workflowDescriptor), name = s"WorkflowInitializationActor-$workflowId")
       initializerActor ! StartInitializationCommand
-      goto(InitializingWorkflowState) using WorkflowActorData(initializerActor, workflowDescriptor)
+      goto(InitializingWorkflowState) using WorkflowActorData(Option(initializerActor), Option(workflowDescriptor))
     case Event(MaterializeWorkflowDescriptorFailureResponse(reason: Throwable), _) =>
       context.parent ! WorkflowFailedResponse(workflowId, MaterializingWorkflowDescriptorState, Seq(reason))
       goto(WorkflowFailedState)
@@ -201,10 +198,8 @@ class WorkflowActor(workflowId: WorkflowId,
       goto(WorkflowFailedState)
   }
 
-  when(AbortingWorkflowState) {
-    case Event(x: EngineLifecycleStateCompleteResponse, _) =>
-      context.parent ! WorkflowAbortedResponse(workflowId)
-      goto(WorkflowAbortedState)
+  when(WorkflowAbortingState) {
+    case Event(x: EngineLifecycleStateCompleteResponse, _) => goto(WorkflowAbortedState)
     case _ => stay()
   }
 
@@ -221,19 +216,21 @@ class WorkflowActor(workflowId: WorkflowId,
     case Event(MetadataPutAcknowledgement(_), _) => stay()
     case Event(AbortWorkflowCommand, WorkflowActorData(Some(actor), _)) =>
       actor ! EngineLifecycleActorAbortCommand
-      goto(AbortingWorkflowState)
+      goto(WorkflowAbortingState)
     case unhandledMessage =>
       log.warning(s"$tag received an unhandled message $unhandledMessage in state $stateName")
       stay
   }
 
   onTransition {
-    // Only publish "External" state to metadata service
-    // workflowState maps a state to an "external" state (e.g all states extending WorkflowActorRunningState map to WorkflowRunning)
-    case fromState -> toState if fromState.workflowState != toState.workflowState =>
+    case fromState -> toState =>
       log.info(s"$tag transitioning from $fromState to $toState")
       // This updates the workflow status
-      pushCurrentStateToMetadataService(toState.workflowState)
+      // Only publish "External" state to metadata service
+      // workflowState maps a state to an "external" state (e.g all states extending WorkflowActorRunningState map to WorkflowRunning)
+      if (fromState.workflowState != toState.workflowState) {
+        pushCurrentStateToMetadataService(toState.workflowState)
+      }
   }
 
   onTransition {
--- a/engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
@@ -9,9 +9,9 @@ import cromwell.engine._
 import cromwell.engine.backend._
 import cromwell.engine.db.DataAccess._
 import cromwell.engine.workflow.WorkflowActor._
-import cromwell.engine.workflow.WorkflowManagerActor.{AbortWorkflowCommand, _}
 import cromwell.services.MetadataServiceActor._
 import cromwell.services.{MetadataEvent, MetadataKey, MetadataValue, ServiceRegistryClient}
+import cromwell.engine.workflow.WorkflowManagerActor._
 import cromwell.webservice.CromwellApiHandler._
 import lenthall.config.ScalaConfig.EnhancedScalaConfig
 
@@ -34,6 +34,7 @@ object WorkflowManagerActor {
   case class SubmitWorkflowCommand(source: WorkflowSourceFiles) extends WorkflowManagerActorCommand
   case class AbortWorkflowCommand(id: WorkflowId) extends WorkflowManagerActorCommand
   case object AbortAllWorkflowsCommand extends WorkflowManagerActorCommand
+  case class SubscribeToWorkflowCommand(id: WorkflowId) extends WorkflowManagerActorCommand
   private[WorkflowManagerActor] final case class RestartWorkflowsCommand(workflows: Seq[OldStyleWorkflowDescriptor]) extends WorkflowManagerActorCommand
 
   /**
@@ -89,7 +90,7 @@ class WorkflowManagerActor(config: Config)
   private def addShutdownHook(): Unit = {
     // Only abort jobs on SIGINT if the config explicitly sets backend.abortJobsOnTerminate = true.
     val abortJobsOnTerminate =
-      config.getConfig("backend").getBooleanOr("abortJobsOnTerminate", default = false)
+      config.getConfig("system").getBooleanOr("abort-jobs-on-terminate", default = false)
 
     if (abortJobsOnTerminate) {
       sys.addShutdownHook {
@@ -108,14 +109,15 @@ class WorkflowManagerActor(config: Config)
      */
     case Event(SubmitWorkflowCommand(source), stateData) =>
       val updatedEntry = submitWorkflow(source, replyTo = Option(sender), id = None)
-      // Submit success brought in from the Cromwell API handler
-      sender ! WorkflowManagerSubmitSuccess(updatedEntry.workflowId)
       stay() using stateData.withAddition(updatedEntry)
-    case Event(AbortWorkflowCommand(id), stateData) =>
+    case Event(SubscribeToWorkflowCommand(id), data) =>
+      data.workflows.get(id) foreach {_ ! SubscribeTransitionCallBack(sender())}
+      stay()
+    case Event(WorkflowManagerActor.AbortWorkflowCommand(id), stateData) =>
       val workflowActor = stateData.workflows.get(id)
       workflowActor match {
         case Some(actor) =>
-          actor ! AbortWorkflowCommand
+          actor ! WorkflowActor.AbortWorkflowCommand
           stay()
         case None =>
           sender ! WorkflowManagerAbortFailure(id, new Exception(s"Couldn't abort $id because no workflow with that ID is in progress"))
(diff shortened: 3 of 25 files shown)


backend/src/main/scala/cromwell/backend/package.scala
```diff
@@ -35,6 +35,9 @@ package object backend {
                                   key: BackendJobDescriptorKey,
                                   inputs: Map[LocallyQualifiedName, WdlValue]) {
     val call = key.call
+
+    /** The abort functionality potentially becomes available after the job descriptor is created */
+    var abortFunction: Option[() => Unit] = None
```
scottfrazer: I'm a terrible person and I feel bad!

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -108,10 +109,11 @@ class WorkflowManagerActor(config: Config)
...
     case Event(SubmitWorkflowCommand(source), stateData) =>
       val updatedEntry = submitWorkflow(source, replyTo = Option(sender), id = None)
-      // Submit success brought in from the Cromwell API handler
-      sender ! WorkflowManagerSubmitSuccess(updatedEntry.workflowId)
       stay() using stateData.withAddition(updatedEntry)
-    case Event(AbortWorkflowCommand(id), stateData) =>
+    case Event(SubscribeToWorkflowCommand(id), data) =>
+      data.workflows.get(id) foreach {_ ! SubscribeTransitionCallBack(sender())}
+      stay()
+    case Event(AbortWorkflowWMACommand(id), stateData) =>
```
scottfrazer: `AbortWorkflowWMACommand`: oh my!  What a terrible name!  I'll switch it to `WorkflowManagerActor.AbortWorkflowCommand`.  There were also 2 other `AbortWorkflowCommand`s in the code and I just didn't want to change them all at the beginning because I didn't know which ones would survive after I was done.

backend/src/main/scala/cromwell/backend/package.scala
geoffjentry: Assuming it is only supposed to be set once, a common pattern for cases like this is to have the var be private, and then expose a function which will complain if one tries to overwrite it when the value is Some. There's not really a totally clean solution to the problem, but sometimes you can't avoid being in this situation.

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalJobExecutionActor.scala
```diff
@@ -146,7 +147,16 @@ class LocalJobExecutionActor(override val jobDescriptor: BackendJobDescriptor,
     stderrTailed.writer.flushAndClose()
 
     processReturnCode match {
-      case ProcessKilledCode => AbortedResponse(jobDescriptor.key) // Special case to check for SIGTERM exit code - implying abort
+      /**
+        * SIGTERM is sent to the job process via Process.destroy() when performing an abort
+        * SIGINT  is sent to the job process on CTRL-C.  Presumably, when the Cromwell receives a
+        *         CTRL-C, it *appears* that UNIX is also sending that signal to the child processes
+        *         so sometimes they die with a SIGINT instead of SIGTERM sent from Cromwell
```
geoffjentry: good catch

backend/src/main/scala/cromwell/backend/package.scala
mcovarr: With the caveat that I haven't looked at this nearly as deeply as you have, would it be possible to have the JesJobExecutionActor send an abort to the JesAsyncBackendJobExecutionActor?  It looks like the latter could capture a reference to a `Future[Run]`, on which it could then just call `abort()`.

scottfrazer:
Wheel chose @cjllanwarne and @kshakir (sorry!)

backend/src/main/scala/cromwell/backend/package.scala
scottfrazer: I'll see what I can do...  this particular bit of code was written toward the beginning of this endeavor and has some roots in the way we used to do it in the olde world, so perhaps there's a better way

scottfrazer: So the `receive` method for `JesAsyncBackendJobExecutionActor` is in `AsyncBackendJobExecutionActor`, so if we were to send the abort message there, it couldn't do JES specific abort functionality because `AsyncBackendJobExecutionActor` is in the backend package.

Also, the real reason for the `var` is to know record the RunID when it becomes available to know if I can even abort in the first place... If I don't have this `var`, I'd have to go to the metadata service, I guess?

kshakir: ToL:

I'm not saying change this, just responding to the comment:

> if we were to send the abort message there, it couldn't do JES specific abort functionality

Could one handle the message in `AsyncBackendJobExecutionActor.receive`, while still allowing `JesAsyncBackendJobExecutionActor` to override a handler with specific JES code?

```scala
def receive = {
  // ...
  case Abort => handleAbort()
  // ...
}

protected def handleAbort(): Unit = {}
```

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/Run.scala
```diff
@@ -132,14 +132,22 @@ case class Run(runId: String,  jobDescriptor: BackendJobDescriptor, genomicsInte
...
+      if (currentStatus.isRunningOrComplete && (previousStatus.isEmpty || !previousStatus.get.isRunningOrComplete)) {
+        jobDescriptor.abortFunction = Some(() => abort())
+      }
     }
+
     currentStatus
   }
 
   def abort(): Unit = {
+    println(s"Aborting JES Run: $runId")
```
kshakir: Remove this `println`. Call `logger.info(…)` over in `JesJobExecutionActor.abort`, inside the `foreach` block calling this function.

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/Run.scala
```diff
@@ -132,14 +132,22 @@ case class Run(runId: String,  jobDescriptor: BackendJobDescriptor, genomicsInte
     val currentStatus = status()
 
     if (!(previousStatus contains currentStatus)) {
-      // DONE: PBE: Moved state transition logging and "db" writes to our actor.
-      // Only thing "left" here via another ticket is:
-      // PBE deleted abort function registration
+      val prevStateName = previousStatus map { _.toString } getOrElse "-"
+      // TODO PBE: change println in this file to log statements
+      println(s"Status change from $prevStateName to $currentStatus")
```
kshakir: Remove this line. `JesAsyncBackendJobExecutionActor.poll` already prints out the line using a `log`.

kshakir:
Minor updates, plus (yet another) rebase. 👍

backend/src/main/scala/cromwell/backend/package.scala
mcovarr: +1 to Khalid's suggestion.

I was thinking a `var` would be used to capture the run ID (or a `Run` that encapsulates the run ID) when that becomes available, but yeah I guess you could use the metadata service for this.  Or was capturing things like run IDs supposed to be the responsibility of the KV store?

cjllanwarne: Looks like everyone else has better answers than me... I'll just add if we do keep the `var` in this case class we might want to make `copy` hard to use, to avoid accidentally losing the `var`'d info

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -136,7 +136,7 @@ class WorkflowActor(workflowId: WorkflowId,
                     serviceRegistryActor: ActorRef)
   extends LoggingFSM[WorkflowActorState, WorkflowActorData] with ActorLogging with KnowsWhatTimeItIs {
 
-  val tag = self.path.name
+  val tag = s"WorkflowActor [UUID(${workflowId.shortString})]"
```
cjllanwarne: ToL: It would be really nice if the query APIs also accepted these short strings. I might add a ticket for that

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -202,10 +199,8 @@ class WorkflowActor(workflowId: WorkflowId,
       goto(WorkflowFailedState)
   }
 
-  when(AbortingWorkflowState) {
-    case Event(x: EngineLifecycleStateCompleteResponse, _) =>
-      context.parent ! WorkflowAbortedResponse(workflowId)
-      goto(WorkflowAbortedState)
+  when(WorkflowAbortingState) {
+    case Event(x: EngineLifecycleActorAbortedResponse, _) => goto(WorkflowAbortedState)
```
cjllanwarne: I think we did want to allow any `EngineLifecycleStateCompleteResponse` here - otherwise there'd be a race condition if e.g. the initializer "succeeds" but the WorkflowActor has already entered `WorkflowAbortingState`

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
cjllanwarne: Also if the WorkflowActor were created with an appropriate name, this change before and after should actually have been equivalent

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -211,19 +213,40 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId,
   when(WorkflowExecutionAbortedState) {
     FSM.NullFunction
   }
+  when(WorkflowExecutionAbortingState) {
+    case Event(AbortedResponse(jobKey), stateData) =>
```
cjllanwarne: As above, it's probably good to also catch a badly-timed success or failure here too

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -278,7 +301,7 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId,
     val runnableCalls = runnableScopes.view collect { case k if k.scope.isInstanceOf[Call] => k } sortBy { _.index.getOrElse(-1) } map { _.tag }
     if (runnableCalls.nonEmpty) log.info(s"Starting calls: " + runnableCalls.mkString(", "))
 
-    // Each process*** returns a Try[WorkflowExecutionDiff], which, upon success, contains potential changes to be made to the execution store.
+    // Each processRunnable* method returns a Try[WorkflowExecutionDiff], which, upon success, contains potential changes to be made to the execution store.
```
cjllanwarne: Are the `*`s important here? I must admit I don't know their meanings

engine/src/main/scala/cromwell/webservice/CromwellApiHandler.scala
```diff
@@ -135,7 +135,7 @@ class CromwellApiHandler(requestHandlerActor: ActorRef) extends Actor {
         case _ => RequestComplete(StatusCodes.InternalServerError, APIResponse.error(e))
       }
 
-    case ApiHandlerWorkflowAbort(id) => requestHandlerActor ! OldStyleWorkflowManagerActor.WorkflowAbort(id)
+    case ApiHandlerWorkflowAbort(id) => requestHandlerActor ! WorkflowManagerActor.AbortWorkflowWMACommand(id)
```
cjllanwarne: Does abort not reply to the sender? You could maybe reply `OK` or `204`?

backend/src/main/scala/cromwell/backend/package.scala
Horneth: or make it a ~~case~~ `class`

engine/src/test/scala/cromwell/engine/backend/mock/DefaultBackendJobExecutionActor.scala
```diff
@@ -19,7 +19,7 @@ case class DefaultBackendJobExecutionActor(override val jobDescriptor: BackendJo
   }
   override def recover = execute
 
-  override def abortJob: Unit = ()
+  override def abort: Unit = ()
```
cjllanwarne: ToL: I can't remember why this is `Unit` instead of `Future[Unit]`... I assume that this doesn't block any actors' `receive` methods if we have a slow-acting `abort` method?

cjllanwarne:
Assuming due diligence: 👍

Horneth:
TOL - Shouldn't we run the finalization actor anyway if we're aborting during execution ?

mcovarr:
@Horneth Yes I think finalization should run anyway, e.g. the Firecloud use case and cleaning up auths

geoffjentry:
@Horneth @mcovarr Agreed that finalization actor should always run at the end. I think I saw a PR for exactly that in there so perhaps this comment is now superfluous :)

Horneth:
@geoffjentry The PR is to make it run if execution fails. I didn't want to conflict with Scott's PR so I didn't do the abort case

backend/src/main/scala/cromwell/backend/package.scala
scottfrazer: I don't know guys, this is getting kinda complicated.. I need to override `receive` on two actors and it's still not clear to me how to actually get the Run ID from the metadata service... it might take me a while to figure that out :-/  #notsmartenoughforcromwell

scottfrazer: Alright, I figured it out... but it's not pretty, will likely take another round of code reviews from you guys.  And I committed a mortal sin by using an `ask` so somehow I gotta fix that first...

scottfrazer:
Alright, @cjllanwarne and @kshakir I updated the code again... it's changed quite significantly to address all comments so probably needs another look!

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
```diff
@@ -166,6 +164,27 @@ class JesAsyncBackendJobExecutionActor(override val jobDescriptor: BackendJobDes
     MetadataJobKey(jobDescriptorKey.call.fullyQualifiedName, jobDescriptorKey.index, jobDescriptorKey.attempt)
   }
 
+  override def receive: Receive = LoggingReceive {
+    case AbortJobCommand =>
+      serviceRegistryActor ! GetMetadataQueryAction(MetadataQuery.forKey(metadataKey("jobId")))
+    case MetadataLookupResponse(query, events) if query.key.contains("jobId") =>
+      /** Currently this actor can only get a MetadataLookupResponse when requesting a
+        * job ID for abort.  If this actor becomes more complex in the future, then this
+        * should probably be received in an FSM state.
```
cjllanwarne: I think the `if` covers you for that?

cjllanwarne:
👍

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -203,24 +212,29 @@ class WorkflowManagerActor(config: Config)
...
     wfActor ! SubscribeTransitionCallBack(self)
     wfActor ! StartWorkflowCommand
-    logger.info(s"Successfuly started ${wfActor.path} for Workflow $workflowId")
+    logger.info(s"$tag Successfully started ${wfActor.path.name}")
     WorkflowIdToActorRef(workflowId, wfActor)
   }
 
   private def restartWorkflow(restartableWorkflow: OldStyleWorkflowDescriptor): WorkflowIdToActorRef = {
-    logger.info("Invoking restartableWorkflow on " + restartableWorkflow.id.shortString)
+    logger.info(s"$tag Invoking restartableWorkflow on " + restartableWorkflow.id.shortString)
```
kshakir: `s"$tag Invoking restartableWorkflow on ${restartableWorkflow.id.shortString}"`

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -173,18 +177,17 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId,
...
     case Event(BackendJobPreparationSucceeded(jobDescriptor, actorProps), stateData) =>
       pushPreparedJobMetadata(jobDescriptor.key, jobDescriptor.inputs)
-      context.actorOf(actorProps, buildJobExecutionActorName(jobDescriptor)) ! ExecuteJobCommand
+      val backendExecutionActor = context.actorOf(actorProps, buildJobExecutionActorName(jobDescriptor))
+      backendExecutionActor ! ExecuteJobCommand
       stay() using mergeExecutionDiff(stateData, WorkflowExecutionDiff(Map(jobDescriptor.key -> ExecutionStatus.Running)))
+        .addBackendJobExecutionActor(jobDescriptor.key, backendExecutionActor)
     case Event(BackendJobPreparationFailed(jobKey, t), stateData) =>
-      log.error(s"Failed to start job $jobKey", t)
+      log.error(s"Failed to start job $jobKey: ${t.getMessage}")
```
kshakir: ToL, for Akka Logging, we should eventually switch to using "[{}](http://doc.akka.io/docs/akka/current/java/logging.html)".

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
scottfrazer: Yes, partially, though I just added this comment so it was clear why this actor was receiving a Job ID from metadata... though I guess the code below is self-explanatory enough... next person to modify this code can remove the comment if they want :)

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
```diff
@@ -166,6 +164,27 @@ class JesAsyncBackendJobExecutionActor(override val jobDescriptor: BackendJobDes
...
+        val jobId = event.value.value
+        log.info(s"$tag Aborting $jobId")
+        Try(Run(jobId, jobDescriptor, genomicsInterface).abort()) match {
+          case Success(_) => log.info(s"$tag Aborted $jobId")
+          case Failure(ex) => log.warning(s"$tag Failed to abort $jobId: ${ex.getMessage}")
+        }
+      }
+      context.parent ! AbortedResponse(jobDescriptor.key)
+      context.stop(self)
+    case message => super.receive(message)
```
kshakir: Don't fix now, but Miguel pointed out we should probably use a slightly different syntax in the future, using `PartialFunction.orElse`. Please drop a `PBE` todo here, and we can (maybe) fix it later if it's an issue.

kshakir:
Review complete. 👍  after one new "PBE" comment in the code, and a string interpolation patch.

PR #63, opened 2015-06-24
add INNODB for MySQL

mcovarr:

PR #897, opened 2016-05-25
Typed metadata Closes #866

### LINKED ISSUE #866: All metadata values are returned as strings
With the new metadata service, all values are returned as JSON strings.  However, the WDL spec says that numbers should be returned where appropriate.  For example, in the 3-step WDL we should (and used to have):
```
outputs": {
  "three_step.cgrep.count": 3
}
```
but we currently have
```
outputs": {
  "three_step.cgrep.count": "3"
}
```
This may require changes to how the metadata is stored which currently persists everything as a string.

Horneth:


Horneth:
@cjllanwarne @mcovarr

database/src/main/resources/changesets/metadata_typed_values.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.3.xsd">
+
+    <changeSet author="tjeandet" id="metadata_typed_values">
+        <addColumn
+                tableName="METADATA_JOURNAL">
+            <column name="METADATA_VALUE_TYPE" type="VARCHAR(255)" />
```
geoffjentry: Since the fields in here are finite can you shrink the size of the varchar?

database/src/main/resources/changesets/metadata_typed_values.xml
```diff
@@ -0,0 +1,16 @@
...
+        </addColumn>
+
+        <modifySql dbms="mysql">
+            <append value=" ENGINE=INNODB"/>
+        </modifySql>
```
Horneth: Does anyone know why this is needed ? It was in the previous metadata changelog so I kept it but I'm not sure why. @mcovarr ?

kshakir: Yay git blame! Old discussion [here](https://github.com/broadinstitute/cromwell/pull/63). Not a heavy mysql user, so I'm not sure why we use it. But as the old docs for [5.5.x](https://dev.mysql.com/doc/refman/5.5/en/storage-engine-setting.html) say:

> The default engine is InnoDB as of MySQL 5.5.5 (MyISAM before 5.5.5).

mcovarr: I just copied this from earlier Liquibase work which was probably motivated by the references Khalid links above.

database/src/main/scala/cromwell/database/obj/Metadatum.scala
```diff
@@ -10,6 +10,7 @@ case class Metadatum
   index: Option[Int],
   attempt: Option[Int],
   value: Option[String],
+  valueType: Option[String],
```
mcovarr: why is this optional?

engine/src/main/scala/cromwell/engine/db/DataAccess.scala
```diff
@@ -370,9 +370,12 @@ trait DataAccess extends AutoCloseable {
     val key = metadataEvent.key
     val workflowUuid = key.workflowId.id.toString
     val timestamp = metadataEvent.timestamp
+    val value = metadataEvent.value.value
+    val valueType = if (metadataEvent.value.valueType == null) null else metadataEvent.value.valueType.typeName
```
mcovarr: Didn't actually try it but I think this might work and would be DRYer:

`Option(metadataEvent.value.valueType) map { _.typeName } orNull`

database/src/main/scala/cromwell/database/obj/Metadatum.scala
Horneth: value is optional so what would be the type of a None value ? :/

mcovarr: It could still have a type even if it's (conceptually) null, but that's not necessarily right...

cjllanwarne: Ideally value would be an ADT with an associated `TypedType` which would allow us to convert into two or more database columns...

Ok never mind.

cjllanwarne:
Who is responsible for converting the current messages into appropriate MetadataTypes?

cjllanwarne:
This looks good to me. 👍

mcovarr:
:+1:

Horneth:
@cjllanwarne 
https://github.com/broadinstitute/cromwell/pull/897/files#diff-2560e1a982cebc32659e2745177dd11cR30
This converts anything to everything

services/src/main/scala/cromwell/services/MetadataQuery.scala
```diff
@@ -20,6 +19,41 @@ object MetadataEvent {
...
+case object MetadataInt extends MetadataType { override val typeName = "int" }
+case object MetadataNumber extends MetadataType { override val typeName = "number" }
+case object MetadataBoolean extends MetadataType { override val typeName = "boolean" }
+
+object MetadataValue {
+  def apply(value: Any) = {
+    value match {
+      case null => new MetadataValue(null, null)
+      case _: Int => new MetadataValue(value.toString, MetadataInt)
+      case _: Double | Long | Float => new MetadataValue(value.toString, MetadataNumber)
```
cjllanwarne: Should `Long` still be a `MetadataInt`?

Horneth: Yep

PR #917, opened 2016-06-01
Test fixing

Horneth:


kcibul:
👍

kcibul:
as per discussion -- one thumb is good for test fixin

PR #920, opened 2016-06-01
Retry output assertion in tests

Horneth:


engine/src/test/scala/cromwell/CromwellTestkitSpec.scala
```diff
@@ -341,25 +344,40 @@ abstract class CromwellTestkitSpec extends TestKit(new CromwellTestkitSpec.TestW
...
+          workflowId
         }
-        workflowId
+
+        // Retry because we have no guarantee that all the metadata is there when the workflow finishes...
+        Await.result(Retry.withRetry(
+          checkOutputs,
+          maxRetries = Some(maxRetries),
+          isFatal = isFatal,
+          backoff = SimpleExponentialBackoff(0.5 seconds, 0.5 second, 1D)), timeoutDuration)
```
mcovarr: This seems like it would be a generally useful bit of code for our brave new eventually consistent world. Specifically, I'm going to need something like this on my query API branch soon. 😄

mcovarr:
:+1:

engine/src/test/scala/cromwell/CromwellTestkitSpec.scala
kshakir: @mcovarr You'll need this in tests... or in main code? If in tests, one could definitely refactor this into a method inside `object CromwellTestkitSpec`. (ToL)

kshakir:
👍

engine/src/test/scala/cromwell/CromwellTestkitSpec.scala
mcovarr: I was thinking for test code.  I can definitely do this on my branch as I'll be the N > 1 case.

PR #921, opened 2016-06-01
Updates to re-enable `WriteTsvSpec`.

kshakir:
`JobPreparationActor` using factory `expressionLanguageFunctions` instead of the unimplemented ones in executionData.


geoffjentry:
👍

PR #924, opened 2016-06-01
fixed comparison for Array[File], fixed up tests since the new world …

kcibul:
…verifies workflow outputs exist (and the test didn't produce them before)

### DIFF
--- a/engine/src/test/scala/cromwell/SubFunctionWorkflowSpec.scala
+++ b/engine/src/test/scala/cromwell/SubFunctionWorkflowSpec.scala
@@ -16,10 +16,12 @@ class SubFunctionWorkflowSpec extends CromwellTestkitSpec {
         |  String myBamString = "myfilename.bam"
         |  File myBamFile
         |  String swappedStr = sub(myBamString, ".bam$", ".txt")
-        |  String swappedFile = sub(myBamFile, ".bam$", ".txt")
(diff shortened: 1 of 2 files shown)


engine/src/test/scala/cromwell/SubFunctionWorkflowSpec.scala
```diff
@@ -16,10 +16,12 @@ class SubFunctionWorkflowSpec extends CromwellTestkitSpec {
         |  String myBamString = "myfilename.bam"
         |  File myBamFile
         |  String swappedStr = sub(myBamString, ".bam$", ".txt")
-        |  String swappedFile = sub(myBamFile, ".bam$", ".txt")
+        |  # at this point myBamFile is not localized so path must be removed
+        |  String swappedFile = sub(sub(myBamFile,"/.*/",""), ".bam$", ".txt")
```
Horneth: Isn't this basically `swappedStr` ?

Horneth:
👍

engine/src/test/scala/cromwell/SubFunctionWorkflowSpec.scala
kcibul: yeah, it's not materially different except for the coercion, but also wanted to test out the exact use case that GOTC is using

PR #926, opened 2016-06-01
Fix InputLocalizationWorkflowSpec and WdlValue / WdlType equality checking

Horneth:


engine/src/test/scala/cromwell/CromwellTestkitSpec.scala
```diff
@@ -258,6 +259,43 @@ abstract class CromwellTestkitSpec extends TestKit(new CromwellTestkitSpec.TestW
...
+    override def areEqual(a: WdlValue, b: Any): Boolean = {
+      val typeEquality = b match {
+        case v: WdlValue => wdlTypeSoftEquality.areEqual(a.wdlType, v.wdlType)
+        case _ => false
+      }
+
+      val valueEquality = (a, b) match {
+        case (_: WdlFile, expectedFile: WdlFile) => fileEquality(a.valueString, expectedFile.valueString)
+        case (_: WdlString, expectedFile: WdlFile) => fileEquality(a.valueString, expectedFile.valueString)
+        case (array: WdlArray, expectedArray: WdlArray) => array.value.zip(expectedArray.value).map(Function.tupled(areEqual)).fold(true)(_ && _)
```
kshakir: Should `WdlArray`s check that their `size` are the same before the `zip`?

engine/src/test/scala/cromwell/CromwellTestkitSpec.scala
```diff
@@ -258,6 +259,43 @@ abstract class CromwellTestkitSpec extends TestKit(new CromwellTestkitSpec.TestW
...
+        case (map: WdlMap, expectedMap: WdlMap) =>
+          val mapped = map.value.map {
```
kshakir: Similarly, `WdlMap`s may want to check that the keys contain the same size/elements. Right now `expectedMap` can actually be a superset of `map` and I think this method will return `true`.

Horneth: Good catch 👍

kshakir:
Go for merge, based on our single 👍  for the hackathon.

Horneth:
Yep just waiting to see if Travis is happy

Issue #863, opened 2016-05-20
Split out Metadata DataAccess from Engine


mcovarr:

mcovarr:
This will likely necessitate different schemas for Liquibase / sanity reasons.

kcibul:
Can you provide a few sentences of detail about what this is and the benefits?

cjllanwarne:
Closing as out-of-date

PR #934, opened 2016-06-02
Query API.  Closes #855

### LINKED ISSUE #855: Implement Query API against Metadata Service
Similar to what we've done with Status and Metadata, the Query API should follow the same pattern

mcovarr:


engine/src/test/scala/cromwell/engine/db/slick/SlickDataAccessSpec.scala
```diff
@@ -132,15 +134,18 @@ class SlickDataAccessSpec extends FlatSpec with Matchers with ScalaFutures with
...
-      if (databaseConfig == SlickDatabase.defaultDatabaseConfig)
-        new SlickDatabase() with DataAccess // Test the no-args constructor
-      else
-        new SlickDatabase(databaseConfig) with DataAccess
+    // PBE I needed to comment this out to prevent what appears to be another attempt to Liquibase in the schema.
+//    lazy val dataAccess =
+//      if (databaseConfig == SlickDatabase.defaultDatabaseConfig)
+//        new SlickDatabase() with DataAccess // Test the no-args constructor
+//      else
+//        new SlickDatabase(databaseConfig) with DataAccess
```
mcovarr: Completely punting on this for now to get the PR moving

engine/src/test/scala/cromwell/engine/db/slick/SlickDataAccessSpec.scala
```diff
@@ -3,7 +3,6 @@ package cromwell.engine.db.slick
 import java.nio.file.{FileSystem, FileSystems}
```
mcovarr: I'm going to look at restoring the status tests in this file.

database/src/main/scala/cromwell/database/SqlDatabase.scala
```diff
@@ -213,4 +216,17 @@ trait SqlDatabase extends AutoCloseable {
...
+    * Retrieves all summarizable metadata satisfying the specified criteria.
+    *
+    * @param startMetadataId        The minimum ID an entry in `METADATA_JOURNAL` must have to be examined for summary.
+    * @param startMetadataTimestamp An optional timestamp.  If specified, a metadatum must have a timestamp greater than or equal to this value.
+    * @return A `Future` with the maximum ID value of the metadata summarized by the invocation of this method.
+    */
+  protected def refreshMetadataSummaries(startMetadataId: Long, startMetadataTimestamp: Option[OffsetDateTime])
+                                        (implicit ec: ExecutionContext): Future[Long]
+
+  protected def getStatus(id: WorkflowId)
```
kshakir: Remove the `core` dependency, and pass in the `workflowUuid: String`. IMHO there's too much other stuff in `core` that will leak deep into the `database`, again.

engine/src/main/scala/cromwell/engine/db/DataAccess.scala
```diff
@@ -649,4 +650,14 @@ trait DataAccess extends AutoCloseable {
...
   }
+
+  def refreshWorkflowMetadataSummaries(startMetadataId: Long, startMetadataTimestamp: Option[OffsetDateTime])
+                                      (implicit ec: ExecutionContext): Future[Long] = {
+    self.refreshMetadataSummaries(startMetadataId, startMetadataTimestamp)
+  }
+
+  def getWorkflowStatus(id: WorkflowId)
+                       (implicit ec: ExecutionContext): Future[Option[WorkflowState]] = {
+    self.getStatus(id) map { _ map WorkflowState.fromString }
```
kshakir: Here is where, IMHO, the translation boundary between the database and engine should exist, _not_ in the database code. Effectively, everything outside of only `case class WorkflowMetadataSummary(…)` should be up in the `engine`.

engine/src/main/scala/cromwell/webservice/ApiDataModels.scala
```diff
@@ -63,6 +65,8 @@ object APIResponse {
 
   /** When a request completes successfully. */
   def success(message: String, data: Option[JsValue] = None) = SuccessResponse("success", message, data)
+
+  def workflowNotFound(id: WorkflowId) = RequestComplete(StatusCodes.NotFound, APIResponse.error(new Throwable(s"Workflow '$id' not found.")))
```
kshakir: Instead of using the base class `new Throwbable`, use `new RuntimeException` if you're feeling java-esque, or even `new Exception`.

engine/src/main/scala/cromwell/webservice/WorkflowQueryPagination.scala
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  //Generates link headers for pagination navigation https://tools.ietf.org/html/rfc5988#page-6
+  protected def generateLinkHeaders(uri: Uri, metadata: Option[QueryMetadata]): Seq[HttpHeader] = {
+    //strip off the query params
+    val baseUrl = uri.scheme + ":" + uri.authority + uri.path
+    metadata match {
+      case Some(meta) =>
+        (meta.page, meta.pageSize) match {
+          case (Some(p), Some(ps)) =>
```
kshakir: Can these get more meaningful names?

engine/src/main/scala/cromwell/webservice/WorkflowQueryPagination.scala
```diff
@@ -0,0 +1,41 @@
...
+
+            val firstLink = Link(Uri(baseUrl).withQuery(generatePaginationParams(1, ps)), Link.first)
```
kshakir: No tests seem to be checking this, so I'm not sure.. but... if one uses a URL like in the README, what is the value of headers `firstLink`, etc. after one uses a URL like these?

- `curl "http://localhost:8000/api/workflows/v1/query?start=2015-11-01&end=2015-11-03&status=Failed&status=Succeeded&page=1&pagesize=10"`<sup>1</sup>

- `curl -X POST --header "Content-Type: application/json" -d "[{\"start\": \"2015-11-01\"}, {\"end\": \"2015-11-03\"}, {\"status\": \"Failed\"}, {\"status\": \"Succeeded\"}, {\"page\": "1"}, {\"pagesize\": "10"}]" "http://localhost:8000/api/workflows/v1/query"`

Looking at the code flow, for GET it seems to probably drop parameters like `start` and `status`. For POST, it seems to add parameters that were supposed to be in the POST content, not the query parameters.

<sup>1</sup> There's an extra " in the README after the cURL and HTTPie examples for GET /api/workflows/:version/query.

supportedBackends/jes/src/test/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActorSpec.scala
```diff
@@ -126,7 +128,7 @@ class JesAsyncBackendJobExecutionActorSpec extends TestKit(ActorSystem("JesAsync
 
       override def executeOrRecover(mode: ExecutionMode)(implicit ec: ExecutionContext): Future[ExecutionHandle] = Future.successful(handle)
     }
-    system.actorOf(Props(new TestableJesJobExecutionActor), "TestableJesJobExecutionActor")
+    system.actorOf(Props(new TestableJesJobExecutionActor), "TestableJesJobExecutionActor-" + UUID.randomUUID().toString)
```
kshakir: Don't need the `.toString`

kshakir: Nor the `()`

engine/src/main/scala/cromwell/webservice/WorkflowQueryPagination.scala
geoffjentry: Disagree

engine/src/main/scala/cromwell/webservice/ApiDataModels.scala
geoffjentry: Agree

engine/src/main/scala/cromwell/engine/db/DataAccess.scala
kshakir: I created #935 that has the previously mentioned change in it.

The `Semigroup` code embedded inside the `refreshMetadataSummaries` transaction is too complicated for me to ramp up on at the moment. So I'm just going to leave the scalaz code in there alone and give that a LGTM. 😉

engine/src/main/scala/cromwell/webservice/ApiDataModels.scala
mcovarr: This is straight copied out of CromwellApiHandler, but sure.

engine/src/main/scala/cromwell/webservice/WorkflowQueryPagination.scala
mcovarr: This entire file is a refactoring of the pagination code from CromwellApiHandler.  This code I believe was adapted from the version that was already merged to 0.19_hotfix and tested by our Green friends.  I haven't tested any of this so I can't personally vouch for its correctness.

supportedBackends/jes/src/test/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActorSpec.scala
mcovarr: ok

engine/src/main/scala/cromwell/webservice/WorkflowQueryPagination.scala
kshakir: Are they actually using this header/code? If so, I do not think it does what they think it does. If not, it's code cruft.

README.md
```diff
@@ -1534,7 +1534,7 @@ This endpoint allows for querying workflows based on the following criteria:
 * `start` (start datetime)
 * `end` (end datetime)
 * `page` (page of results)
-* `pagesize` (# or results per pa
```
geoffjentry: did it really say `pa` before or is that just a git diff artifact? :)

README.md
```diff
@@ -1594,6 +1594,22 @@ Server: spray-can/1.3.3
 }
 ```
 
+Query data is refreshed from raw data periodically according to the configuration value `services.MetadataService.metadata-summary-refresh-interval`.
+If not specified the refresh interval will default to 2 seconds.
```
geoffjentry: Can you specify if that's every n seconds, or n seconds after the previous one? 

I was going to ask that question, said I could look further down into the diff, and then figured if I was curious that I might be the only one who wonders that in the future :)

engine/src/main/scala/cromwell/services/EngineMetadataServiceActor.scala
```diff
@@ -23,6 +45,34 @@ case class EngineMetadataServiceActor(serviceConfig: Config, globalConfig: Confi
     }
   }
 
+  private def queryStatusAndRespond(id: WorkflowId): Unit = {
+    val sndr = sender()
+    dataAccess.getWorkflowStatus(id) onComplete {
```
geoffjentry: Am I following things correctly that this status request is hitting the main DB and not the metadata service? is there any potential mismatch problem there?

engine/src/main/scala/cromwell/webservice/WorkflowQueryPagination.scala
geoffjentry: I believe they're actively using the stuff in hotfix, I don't know to what extent this is the same or different. @vivster7 @jacarey - can one of you two chime in?

services/src/main/scala/cromwell/services/ServiceRegistryActor.scala
```diff
@@ -45,7 +45,12 @@ case class ServiceRegistryActor(globalConfig: Config) extends Actor {
     case msg: ServiceRegistryMessage =>
       services.get(msg.serviceName) match {
         case Some(ref) => ref.tell(msg, sender)
-        case None => sender ! ServiceRegistryFailure(msg.serviceName)
+        case None =>
+          log.error("Received ServiceRegistryMessage requesting service '{}' for which no service is configured.  Message: {}", msg.serviceName, msg)
+          sender ! ServiceRegistryFailure(msg.serviceName)
       }
+    case fool =>
```
geoffjentry: Is a hollow voice saying this?

geoffjentry:
:+1:

engine/src/main/scala/cromwell/webservice/WorkflowQueryPagination.scala
vivster7: the code looks the same as what Jay/I wrote in the hotfix branch. Its working for us:

```curl -I "https://cromwell.gotc-dev.broadinstitute.org/api/workflows/v1/query?status=Failed&page=1&pagesize=10"```

will return the proper link headers that we are using to paginate the responses.

Comment (vivster7) [inline engine/src/main/scala/cromwell/webservice/WorkflowQueryPagination.scala]: The response from the above curl command:

```
HTTP/1.1 200 OK
Date: Fri, 03 Jun 2016 14:47:03 GMT
Server: spray-can/1.3.2
X-Frame-Options: SAMEORIGIN
Access-Control-Allow-Origin: *
Access-Control-Allow-Headers: authorization,content-type,accept,origin
Access-Control-Allow-Methods: GET,POST,PUT,PATCH,DELETE,OPTIONS,HEAD
Access-Control-Max-Age: 1728000
Content-Type: application/json; charset=UTF-8
Content-Length: 2036
Access-Control-Expose-Headers: Link
Link: <https://cromwell.gotc-dev.broadinstitute.org/api/workflows/v1/query?page=1&pagesize=10>; rel=first
Link: <https://cromwell.gotc-dev.broadinstitute.org/api/workflows/v1/query?page=1&pagesize=10>; rel=prev
Link: <https://cromwell.gotc-dev.broadinstitute.org/api/workflows/v1/query?page=2&pagesize=10>; rel=next
Link: <https://cromwell.gotc-dev.broadinstitute.org/api/workflows/v1/query?page=87&pagesize=10>; rel=last
```

README.md
mcovarr: That's not an artifact.  It's ok on hotfix, not sure what happened here.

README.md
mcovarr: sure

engine/src/main/scala/cromwell/services/EngineMetadataServiceActor.scala
mcovarr: We only have one DB/schema. #863 is the ticket to change that.  This is hitting the summary table.

geoffjentry: That was bad wording on my part, but your last sentence answered my question anyways :)

mcovarr: I'm not 100% sure I love this arrangement actually.  Now workflows might be submitted and a UUID returned, but a  status poll will return a 404 until the first summary sweep.

Horneth: Can we force a sweep ?

geoffjentry: that's a good point. my best off the cuff thought is that we should be allowed to know if an id at least exists in the system, but that implies gleaning that knowledge from outside metadata

engine/src/main/scala/cromwell/webservice/WorkflowQueryPagination.scala
kshakir: Uhm... those links don't appear to go to the next page for the original query of `status=Failed`. They go to the next page of _any_ workflow.

vivster7: whelp. you are right.

kshakir: Also, what headers does it return for the POST version of the same endpoint, used by Firecloud (I believe)?

The POST endpoint reads parameters in the body, not the query parameters, so I'm not exactly sure what the proper link headers are in that case, as the expected header responses for GET nor POST aren't listed in the `CromwellApiServiceSpec` nor the README.

If the headers are working for users in hotfix, I don't think we need to change anything there. But the generated headers here in develop appear to need some help.

engine/src/main/scala/cromwell/services/EngineMetadataServiceActor.scala
mcovarr: @Horneth AFAICT that's fraught with race conditions

@geoffjentry Personally I don't care as much about the ideological purity of separating this stuff as much as UX / not having a stream of tickets filed for this behavior.  😄

engine/src/main/scala/cromwell/webservice/WorkflowQueryPagination.scala
mcovarr: Just a heads up, I wouldn't recommend trying to fix this on develop until this branch is merged

vivster7: This won't work perfectly for us either on hotfix -- we may have to submit more patches there. First, we have to deploy viewscreen (the cromwell consumer) so we can get a better idea of what we need to make it work.

I'm also not sure what headers should return from a POST request - personally, I don't think Link headers should be in the response to POST requests. I think a metadata object should be included in the response which could be used by the client to construct paginated POST requests. 

And if you're doing that for POST requests, it could make sense to remove Link header pagination and just support this metadata object for GET and POST requests. This implementation probably requires a bit more discussion though.

kshakir:
Good to go for merge as is, assuming the headers will be fixed+tested+documented+used at some future date. 👍

Issue #877, opened 2016-05-24
WorkflowActorTest fails intermittently


kcibul:
I noticed that WorkflowActorSpec is a bit flakey, specifically "transition to InitializingWorkflowState with correct call assignments given workflow options".  It fails with "WorkflowFailedState was not equal to InitializingWorkflowState".  I think the problem is that the test set the FSM state, fires a message and then checks the new state.  The FSM when it gets that message transitions to the right state (Initializing) and then transitions to the next state (failed) and sometimes the test misses the initializing.

but all that might be because the runtime options aren't actually valid?  (sge backend) but I'm missing the purpose of that specific setting so I don't want to change it.

scottfrazer:
This test changed a lot since this ticket was initially filed.  I cannot reproduce it.

PR #935, opened 2016-06-03
Separated `core` out of the `database` for mlc_query_api

kshakir:


database/src/main/scala/cromwell/database/obj/WorkflowMetadataSummary.scala
```diff
@@ -42,9 +39,9 @@ case class WorkflowMetadataSummary
     // Resolve the status if both `this` and `that` have defined statuses.  This will evaluate to `None`
     // if one or both of the statuses is not defined.
     val resolvedStatus = for {
-      a <- this.status map WorkflowState.fromString
-      b <- that.status map WorkflowState.fromString
-    } yield (a |+| b).toString
+      thisStatus <- this.status
+      thatStatus <- that.status
+    } yield (thisStatus |+| thatStatus).toString
```
mcovarr: I don't think you need the `.toString` or the parens around your fancy `Semigroup` append 😛

database/src/main/scala/cromwell/database/slick/SlickDatabase.scala
```diff
@@ -793,10 +792,8 @@ class SlickDatabase(databaseConfig: Config) extends SqlDatabase {
     runTransaction(action)
   }
 
-  def getStatus(id: WorkflowId)
-               (implicit ec: ExecutionContext): Future[Option[String]] = {
-
-    val action = dataAccess.workflowStatusByUuid(id.id.toString).result.headOption
+  def getStatus(workflowUuid: String)(implicit ec: ExecutionContext): Future[Option[String]] = {
+    val action = dataAccess.workflowStatusByUuid(workflowUuid.toString).result.headOption
```
mcovarr: that `.toString` looks superfluous

mcovarr:
Hrm looks like Travis is legit unhappy.  😦   The Semigroup `|+|` on WorkflowStates was resolving states via CRDT, but the default definition of Semigroup on `String` is just concatenating.

kshakir:
Le sigh. Let me take a look.

kshakir:
Moved more of the biz logic out of the slick specific code. There's still a bit in there that could be abstracted out (for example: calculating a max row id, specifying what metadata keys to return, etc.), but it at least passes Travis now.

engine/src/main/scala/cromwell/engine/db/DataAccess.scala
```diff
@@ -42,6 +44,50 @@ object DataAccess {
...
+        // If not both statuses are defined, take whichever is defined.
+        status = resolvedStatus orElse summary1.status orElse summary2.status,
+        name = summary1.name orElse summary2.name,
+        startDate = summary1.startDate orElse summary2.startDate,
+        endDate = summary1.endDate orElse summary2.endDate
+      )
+    }
+  }
+
+  private def getUpdatedSummary(existingSummary: Option[WorkflowMetadataSummary],
```
mcovarr: `get` doesn't seem like the right term here as this always builds a new summary

engine/src/main/scala/cromwell/engine/db/DataAccess.scala
```diff
@@ -42,6 +44,50 @@ object DataAccess {
...
+                                metadataForUuid: Seq[Metadatum]): WorkflowMetadataSummary = {
+    implicit val wmss = WorkflowMetadataSummarySemigroup
+
+    val baseSummary = existingSummary.getOrElse(WorkflowMetadataSummary(metadataForUuid.head.workflowUuid))
+    metadataForUuid.foldLeft(baseSummary) {
+      case (metadataSummary, metadatum) => metadataSummary |+| toSummary(metadatum)
+    }
+  }
+
+  private def toSummary(metadatum: Metadatum): WorkflowMetadataSummary = {
```
mcovarr: This was a method on an implicit class, is there a reason it was refactored away from that?

engine/src/main/scala/cromwell/engine/db/DataAccess.scala
mcovarr: perhaps 'build' instead?  In hindsight my original choice of `fold` seems not so great.  😄

mcovarr:
:+1: apart from some nits.  The Travis unhappiness looks to be an instance of #877 and not related to your changes.  Glad to see the reduction in complexity in `refreshMetadataSummaries`.  😄

engine/src/main/scala/cromwell/engine/db/DataAccess.scala
kshakir: It was only used in one place, so I changed it from a implicit class with a method to just a method. I'll put it back as the original, as this PR is targeting your branch.

geoffjentry: I don't see enhanced methods as being an artifact of how many times something is used anyways, IMO it's just more natural to read stuff left to right, despite what the reverse polish people would claim

database/src/main/scala/cromwell/database/SqlDatabase.scala
```diff
@@ -222,11 +220,15 @@ trait SqlDatabase extends AutoCloseable {
     *
     * @param startMetadataId        The minimum ID an entry in `METADATA_JOURNAL` must have to be examined for summary.
     * @param startMetadataTimestamp An optional timestamp.  If specified, a metadatum must have a timestamp greater than or equal to this value.
+    * @param getUpdatedSummary      Takes in the optional existing summary and the metadata, returns the new summary.
+    *
     * @return A `Future` with the maximum ID value of the metadata summarized by the invocation of this method.
     */
-  protected def refreshMetadataSummaries(startMetadataId: Long, startMetadataTimestamp: Option[OffsetDateTime])
+  protected def refreshMetadataSummaries(startMetadataId: Long, startMetadataTimestamp: Option[Timestamp],
```
geoffjentry: Just curious as I'm still trying to figure out what we should/shouldn't be doing with the time stuff, and you're both the person who made the change and the one who knows best - why the change from OffsetDateTime to Timestamp?

engine/src/main/scala/cromwell/engine/db/DataAccess.scala
geoffjentry: I was just thinking to myself that it's semigroups everywhere these days :P

geoffjentry:
Potential tech talk subject: WTH is `core` for? When I first made it, I intended it to be a grab bag of stuff that everything else sat on top. This is not the first time someone else had a differing view (e.g. someone had said at one point that they thought it was a place to put stuff that everything *did* depend on, not just *could* depend on).

I don't really care what the answer is, although I'm thinking the nomenclature is no longer accurate and perhaps should be changed.

Something to talk about after standup?

kshakir:
@mcovarr I believe all issues are addressed. I'll let you cherry-pick / merge / squash or even reject this PR at your will.

Pre-tech talk with @geoffjentry (including a TLA extravaganza!):

**TL;DR The database/slick is for CRUD, not for the T in ETL.**

IMHO, the database code started to evolve well beyond CRUD. Very often in slick specific code, one saw multiple lines of code like: "before inserting rows in slick, quickly (E)xtract some other rows, (T)ransform them into core objects, filter, re-transform them into new core objects, then (L)oad the new objects into slick." That ETL embedded in slick could never be used DRY-ly, and to me smelled as violating separation of concerns.

With the current SoC, the slick code is _mostly_ concerned with marshaling data to and from the database via slick. If I wanted to, I could very trivially create a different layer that marshaled data using hibernate, a thin layer of prepared statements, mocks, etc., _without_ duplicating a lot of the ETL code.

Another way of visualizing the issue: Below is the current project dependency diagram. The services need to access data from the database. Currently that's implemented as the services depend on engine that depends on the database. The database used to have a similar same circular dependency. Gun-shy of folks (including myself) re-introducing a similar dependency loop, I've kept core as far away as possible from the database/slick, because the slick specific code _should not_ need core for basic CRUD.

As for the rest of the system, I see core as a base of objects for backends and the engine to communicate.

cromwell project dependency diagram

database/src/main/scala/cromwell/database/SqlDatabase.scala
kshakir: Gotta run, so I can't summarize the links below at the moment. In short, even if a specific database can store a datetime+zone in a single column, jdbc only supports the SQL `TIMESTAMP`, not `TIMESTAMPZ`. There's some work to get slick to marshal the type, but when it's not supported by the server, the current draft is marshaling the values as strings.

- https://mariadb.com/kb/en/sql-99/08-temporal-values/
- https://en.wikipedia.org/wiki/SQL#Data_types
- https://java.net/jira/browse/JPA_SPEC-63
- https://github.com/slick/slick/issues/1026
- http://stackoverflow.com/questions/34608650/scala-slick-3-0-implicit-mapping-between-java8-offsetdatetime-and-timestamp
- https://github.com/xavier-fernandez/slick/blob/53683920c553a476b4a87c1ca296ce0c76504bc7/slick/src/main/scala/slick/jdbc/JdbcTypesComponent.scala#L310-L320

mcovarr:
@kshakir this needs a rebase as I incorporated some PR feedback and a rebase on develop

kshakir:
Latest rebase complete

PR #941, opened 2016-06-03
Fixes to `SlickDataAccessSpec` plus some other patches.

kshakir:
Fixes to `SlickDataAccessSpec` plus some other patches.
`SlickDataAccessSpec` should talk to the currently tested `SlickDatabase with DataAccess`, NOT the `service` layer, that in turn talks to the global singleton.
The `root` project was not aggregating `database`.
The `root` project with `Main` only actually depends on `engine` and `core` tests.
More tests fixed that were waiting for the wrong message.
Changed waiting for `start` to `Starting`.
Changed waiting for `call.name` to `call.name:NA:1`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -63,13 +63,5 @@ lazy val root = (project in file("."))
   .settings(rootSettings: _*)
   .dependsOn(core % "test->test;compile->compile")
   .dependsOn(engine % "test->test;compile->compile")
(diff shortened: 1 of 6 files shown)


mcovarr:
:+1: :+1: :+1:

Horneth:
👍

PR #942, opened 2016-06-04
Enable all tests in SlickDataAccessSpec

mcovarr:
This enables tests on a lot of Slick code that's not actually used yet in the New Worlde (nothing writes to the core engine tables since only Recover needs that and Recover hasn't been implemented).  So these changes are valuable iff the New Worlde ultimately uses an engine Slick API that looks a lot like that in the Olde Worlde.  My guess is that will end up being true, but at this point that's only a guess.  Some things here will certainly be nixed (ExecutionEvents) or are likely to be heavily modified (anything caching-related).


kshakir:
👍

PR #948, opened 2016-06-06
Support for transactional insert of several metadata events. Closes #930

### LINKED ISSUE #930: Atomic publishing of multiple metadata events
During the testing hackathon, we discovered a number of problems caused by the eventual consistency of the metadata service.  One specific case of this is the granularity of the events.  

On one side, we have a publisher who has a whole collection of events that they would like to push. They push them one event at a time to the MD service.  Because even things like array elements are pushed one update at a time because of MD format, we run into the situation where a consumer can see half of an array.  Taken to the extreme, we could push every char of a string as a separate event.

The fundamental problem with these partial updates is that a downstream consumer can not tell if an update is complete.  Do they wait?  How long?  Can they check if the data is done?

While this touches on a larger problem in distributed computing, I think we are shooting ourselves in the foot by making every piece of a single update an async, isolated event.  Taken to the extreme, we could push every char of a string as a separate event.

The proposal is to extend the PutMetadataAction to take in a Seq/Varargs of MetadataEvents with the contract that these will be made available atomically (e.g. in a single Slick transaction for our implementation).  Then in places where we basically unrolling a bundle of events to publish, we should use this API (e.g. WorkflowExecutionActor) to do that atomically.  

In theory, this should also help with the scalability as the MD service can persist things with batchinserts in single transaction.  For larger workflows, currently this would be hundreds or thousands of transactions.

Horneth:


database/src/main/scala/cromwell/database/SqlDatabase.scala
```diff
@@ -178,22 +178,12 @@ trait SqlDatabase extends AutoCloseable {
...
-                                 value: String,
-                                 valueType: String,
-                                 timestamp: Timestamp)
-                                (implicit ec: ExecutionContext): Future[Unit]
+  /**
+    * Add metadata events to the database transactionally. normalized type structure is as follows:
+    * (WorkflowId, MetadataKey, Option[CallFqn, CallIndex, CallAttempt], MetadataValue, MetadataValueType, Timestamp)
+    */
+  protected def addMetadata(events: Iterable[(String, String, Option[(String, Option[Int], Int)], String, String, Timestamp)])
+                           (implicit ec: ExecutionContext): Future[Unit]
```
mcovarr: can we have a type for this in `database` land?

geoffjentry: And not a `type` but a `case class` if you want a type

services/src/main/scala/cromwell/services/MetadataServiceActor.scala
```diff
@@ -28,7 +27,11 @@ object MetadataServiceActor {
   trait MetadataServiceAction extends MetadataServiceMessage with ServiceRegistryMessage {
     def serviceName = MetadataServiceName
   }
-  case class PutMetadataAction(event: MetadataEvent) extends MetadataServiceAction
+  object PutMetadataAction {
+    def apply(event: MetadataEvent) = new PutMetadataAction(List(event))
+    def apply(events: MetadataEvent*) = new PutMetadataAction(events)
```
mcovarr: is the first necessary?  what if there are no events?

kshakir: Yeah. While I don't have a citation for all of the benefits, I feel like I've seen the overload pattern as:

```scala
def apply(event: MetadataEvent) = new PutMetadataAction(List(event))
def apply(firstEvent: MetadataEvent, otherEvents: MetadataEvent*) = //…
```

Something like that would at minimum stop one from passing in "no events".

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
```diff
@@ -482,16 +483,17 @@ class JesAsyncBackendJobExecutionActor(override val jobDescriptor: BackendJobDes
     tellMetadata("preemptible", preemptible)
 
     // TODO: PBE: Trace callers of "new CallContext()". Seems to be multiple places in JES, etc. For now:
-    tellJesFileMetadata("stdout", jesCallPaths.stdoutFilename)
-    tellJesFileMetadata("stderr", jesCallPaths.stderrFilename)
-    tellJesFileMetadata("backendLogs:log", jesCallPaths.jesLogFilename)
+    tellMetadata("stdout", jesCallPaths.stdoutPath.toAbsolutePath)
+    tellMetadata("stderr", jesCallPaths.stderrPath.toAbsolutePath)
+    tellMetadata("backendLogs:log", jesCallPaths.jesLogPath.toAbsolutePath)
```
kshakir: While touching these lines, it should be changed to use the "transactional insert of several metadata events" implemented elsewhere in this PR.

kshakir:
Some comments that should hopefully be straightforward enough to address, assuming no objections from the other reviewer. Generally 👍

database/src/main/scala/cromwell/database/SqlDatabase.scala
Horneth: I thought `SqlDatabase` would only handle basic scala types ? If that's not the case why not also a `JobKey`-like type, `WorkflowId` type, etc...

geoffjentry: Good point, I think you're right.

At this point I'm pretty sour on the type aliases in general except for a couple of specific circumstances (mostly related to type wizardry from scalaz/cats). I'd rather stick with primitives for most situations if a case class can't be used.

kshakir: Very-ToL:

a)

While I'm ok with all the discussions above, my preference here is that there be more than one overloaded `addMetadata()`, instead of the slick implementation having to peek so much at the `Some` or `None` passed in. One would only work on calls, the other on the workflow level. The names may need to be changed to `addWorkflowMetadata(Iterable[_], …)` and `addCallMetadata(Iterable[_], …)` due to the persistently annoying java erasure.

b)

I don't mind additional case classes in `cromwell.database.obj` for creating or returning data from the database. I do prefer the above mentioned overloads with basic scala types, but a case class isn't what I'm really worried about bleeding back into the slick code. Pre-`cromwell.database`, what I feel like I saw was: case classes with all sorts of conversion or business logic methods added, especially implicit converters in companion objects. Then all those methods would be invoked inside slick specific code, instead of basic CRUD, making it really hard to separate out what was actually slick FRM, and what was logic.

My latest 2¢

geoffjentry: @kshakir You do touch on the one issue i see w/ case class type wrappers, the attractive nuisance they provide for jamming all sorts of logic in there instead of just being a typed wrapper

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalJobExecutionActor.scala
```diff
@@ -32,7 +33,7 @@ object LocalJobExecutionActor {
 
 class LocalJobExecutionActor(override val jobDescriptor: BackendJobDescriptor,
                              override val configurationDescriptor: BackendConfigurationDescriptor)
-  extends BackendJobExecutionActor with SharedFileSystem {
+  extends BackendJobExecutionActor with SharedFileSystem with ServiceRegistryClient {
```
mcovarr: why this change?

mcovarr:
:+1: delta some nits

supportedBackends/local/src/main/scala/cromwell/backend/impl/local/LocalJobExecutionActor.scala
Horneth: I started adding runtime attributes metadata for local backend and then rollbacked but this was left behind

database/src/main/scala/cromwell/database/SqlDatabase.scala
Horneth: @kshakir While I agree that overloaded `addMetadata` methods would be cleaner, in practice it's only going to result in more "useless" work to do since we'll have to partition the `MetadataEvent`s into "workflow events" and "call events", to choose the right `addMetadata` to call, which will both anyway call the same slick query in the end, only one with `Option(..)` and the other with `None` for some parameters. Or otherwise we could split Metadata into `WorkflowEvents` and `CallEvents` all the way up to the metadata service so the caller would have to choose what kind of event they're sending.

Issue #789, opened 2016-05-05
Data Migration from existing metadata schema into Metadata Service schema


kcibul:
depends on #788 

This is really important for our users so they don't lose all their metadata!

mcovarr:
Depends on #785 (which it is currently prioritized ahead of in Ready for Development), #788 (which is currently in development), and #668 (in review)

mcovarr:
This ticket really can't be taken until the dust completely settles on metadata, which is still a ways off.

Horneth:
@kcibul A few questions about this:
- Should it support in-place DB migration, or DB A to DB B, or both ?
- Is it a separate program that runs independently ? If yes does it still live in the Cromwell repo ? Or a new flag in cromwell (like server, run, migration), ? Is there any "automagic" detection that my DB needs an update when I run cromwell 0.20 on a pre-0.20 DB ?

kcibul:
It should be in-place (e.g. the source and destination tables are in the same database).  

It would be nice to be automagic (e.g. liquibase), but just having a script in the repository to do the upgrade would be fine as well.  It's really just for existing production customers who need to bridge the gap (GOTC/FC)

ruchim:
Working with Henry tomorrow to get these migration changes tested.

geoffjentry:
It'd be fantastic if this could also handle #1336. If not, could you let me know when this is done and I'll do it.

PR #957, opened 2016-06-06
Nix ExecutionEvent

mcovarr:


Horneth:
If I understand this gets rid of ExecutionEvents in the old world, but we're still planning on re-introducing them in the new world right ?

geoffjentry:
To go along w/ the comment from @Horneth - some explanation on what this is all about would be helpful :)

mcovarr:
This gets rid of the ExecutionEvent table and special types for execution events from the Olde Worlde.  We are recording these events as metadata in the New Worlde, but aren't yet consuming them as a richer type.

database/src/main/scala/cromwell/database/liquibase/DiffResultFilter.scala
```diff
@@ -159,5 +171,7 @@ object DiffResultFilter {
       * @return The updated diff result.
       */
     def filterChangedObjects(filters: Seq[DiffFilter]) = filter(diffResult, filters, (_, _) => false)
+
+    def filterOldeObjects = filter(diffResult, Seq.empty, isOldeObject)
```
mcovarr: I will `TODO PBE` this as something that should be deleted after the offending Olde tables are deleted.

Horneth:
👍

database/src/main/scala/cromwell/database/liquibase/DiffResultFilter.scala
kcibul: if we delete the old tables... are we going to transfer the data into the
new structure?  should that be a new ticket?

Kristian Cibulskis
kcibul@broadinstitute.org

mcovarr: I thought that was covered by #789 but if not then there should certainly be a ticket for it.

geoffjentry: I popped in to say "I think there's already a ticket for it", @mcovarr beat me to it

PR #963, opened 2016-06-06
Change default version on status API.  Closes #937

### LINKED ISSUE #937: Status REST API should be backwards compatible
Currently the new status API is on V2, it should be V1 and be made to have the same format as V1 (ie no V2)

If there are other differences between the current and former REST api metadata, let's quickly techtalk those differences and the PO will decide

mcovarr:
The API itself ignores version and always returns data in the classic format, so remove mention of "v2".


cjllanwarne:
👍

PR #964, opened 2016-06-06
Nix FailureEvent

mcovarr:
This is like #957 but for Failure Events.  Failure Events are published as metadata but no longer rate their own separate table.


geoffjentry:
👍

PR #965, opened 2016-06-06
Metadata service workflow existence cache.  Closes #962

### LINKED ISSUE #962: Cache existence of workflows in MetadataServiceActor
Given the eventually consistent scheme in place for deriving workflow status, it's likely that a status check soon after workflow submission will fail with a 404 response.  From a UX perspective a bogus 404 error due to eventual consistency feels wrong; a lag in status update is one thing, but Cromwell probably shouldn't disavow the existence of the workflow.

mcovarr:


engine/src/main/scala/cromwell/services/EngineMetadataServiceActor.scala
```diff
@@ -49,7 +54,9 @@ case class EngineMetadataServiceActor(serviceConfig: Config, globalConfig: Confi
     val sndr = sender()
     dataAccess.getWorkflowStatus(id) onComplete {
       case Success(Some(s)) => sndr ! StatusLookupResponse(id, s)
-      case Success(None) => sndr ! StatusLookupNotFound(id)
+      case Success(None) =>
+        val message = if (workflowExistenceCache.contains(id)) StatusLookupResponse(id, WorkflowSubmitted) else StatusLookupNotFound(id)
```
mcovarr: This is reading mutable state in a callback which is usually dodgy for an actor, but it is a var/immutable which maybe is ok?  I can add a layer of messaging indirection to remove that layer of dodgy.

engine/src/main/scala/cromwell/services/EngineMetadataServiceActor.scala
```diff
@@ -88,9 +95,14 @@ case class EngineMetadataServiceActor(serviceConfig: Config, globalConfig: Confi
     case GetStatus(workflowId) => queryStatusAndRespond(workflowId)
     case WorkflowQuery(uri, parameters) => queryWorkflowsAndRespond(uri, parameters)
     case RefreshSummary => summaryActor ! SummarizeMetadata
-    case MetadataSummarySuccess => scheduleSummary
+    case MetadataSummarySuccess =>
+      workflowExistenceCache = workflowExistenceCache mapValues { _ - 1 } filter { case (_, v) => v > 0}
```
geoffjentry: I'm not 100% sure how the whole `mapValues`/view thing works (except that one shouldn't put non-RT functions in there, which you're not) but I *think* this could be a giant performance issue - if it works the way I think it does, every one of these calls will store the latest changes which will be reapplied on every read.

disclaimer: there's a 82.81222% change it doesn't work the way I think it does

engine/src/main/scala/cromwell/services/EngineMetadataServiceActor.scala
```diff
@@ -28,6 +28,9 @@ object EngineMetadataServiceActor {
 
   val MetadataSummaryTimestampMinimum =
     ConfigFactory.load().getStringOption("services.MetadataService.metadata-summary-timestamp-minimum") map OffsetDateTime.parse
+
+  // Number of runs of the workflow summary actor to allow a workflow ID to sit in the cache before being expired out.
```
geoffjentry: Combining it w/ the code that uses it it makes more sense but I didn't really find this message to be super helpful. I'm not sure if it's the wording here, the lack of comment below, or some combination of the two ... I realize that's super helpful.

geoffjentry:
👍

engine/src/main/scala/cromwell/services/EngineMetadataServiceActor.scala
mcovarr: Wow yikes, did not know about that.  I'm a gonna go ahead and change that.

cjllanwarne:
ToL: I'm a little afraid of the arbitrary number 5... Is it at all easy to make the `WorkflowSummaryActor` send a `ExciseFromCache` message when it sees a status message for a workflow?

Otherwise, (and especially since this makes Centaur great again), 👍

PR #968, opened 2016-06-07
Added a seq function to Cromwell's WDL interpretation.

cjllanwarne:
From dog-fooding array creation in WDL.

### DIFF
--- a/backend/src/main/scala/cromwell/backend/wdl/PureFunctions.scala
+++ b/backend/src/main/scala/cromwell/backend/wdl/PureFunctions.scala
@@ -1,12 +1,30 @@
 package cromwell.backend.wdl
 
 import wdl4s.expression.WdlStandardLibraryFunctions
-import wdl4s.types.WdlStringType
-import wdl4s.values.{WdlString, WdlValue}
+import wdl4s.types.{WdlArrayType, WdlIntegerType, WdlStringType}
+import wdl4s.values.{WdlArray, WdlInteger, WdlString, WdlValue}
 
 import scala.util.{Failure, Success, Try}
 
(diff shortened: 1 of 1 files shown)


cjllanwarne:
Would appreciate any instructions on where to document this (if at all!)

cjllanwarne:
Allows WDL code like:
```
  Array[Int] indexing_list = seq(5)

  scatter (idx in indexing_list) {
    call FileSpam { input: index = idx, width = 5 }
  }
```

it's also annoying that you can't also put the seq(5) inline, but that's a separate, existing ticket

Horneth:
I'd say here https://github.com/broadinstitute/wdl/blob/develop/SPEC.md
And also add it here https://github.com/broadinstitute/wdl4s/blob/d7e19c9f4dfbc5ad912cf641af9c640eb8a9a9c7/src/main/scala/wdl4s/expression/WdlStandardLibraryFunctions.scala

cjllanwarne:
@Horneth - after talking to @scottfrazer I was hoping to avoid this requiring a WDL4S change (i.e. and therefore let it be a Cromwell custom function rather than a core feature of the WDL spec)

scottfrazer:
I'm certainly not opposed to it being an official function in WDL if you want it to be!  If not, it can be a hidden feature :)

Horneth:
Ok for me either way !

scottfrazer:
👍

Horneth:
👍

geoffjentry:
@cjllanwarne So is this like python's range? The simple form at least. If so I wonder if that'd be a more obvious name for it in terms of the primary wdl userbase

PR #970, opened 2016-06-07
Return of metadata Closes #936

### LINKED ISSUE #936: Metadata REST API should be backwards compatible
Currently the new metadata API is on V2, it should be V1 and be made to have the same format as V1 (ie no V2)

One specific change is that the current implementation (V2) returns a map of workflow id -> metadata.  This should be flattened (no workflow id) because it is a single workflow REST API.

If there are other differences between the current and former REST api metadata, let's quickly techtalk those differences and the PO will decide

Horneth:


cjllanwarne:
Just OOI when reviewing, what's the change here actually doing? (and is this `Closes #xyz` anything?)

engine/src/main/scala/cromwell/webservice/MetadataBuilderActor.scala
```diff
@@ -128,41 +128,60 @@ object MetadataBuilderActor {
       case s => Option(s.toInt)
     }
   }
+
+  private case object IndexedEmptyJson extends IndexedJsonValue {
+    override def index: Option[Int] = None
+    override def withIndex(index: String): IndexedJsonValue = this
+    override def toJson: JsValue = throw new IllegalStateException("IndexedEmptyJson cannot be converted to Json")
+  }
```
mcovarr: Has the `Semigroup[IndexedJsonValue]` evolved into a `Monoid`?

cjllanwarne: The fact that this sentence exists makes me sad

Horneth: Hmm yeah actually it might be simpler with a `Monoid`
Also - now I'm going to go home and cry for having said that

Horneth:
@cjllanwarne Yes it closes https://github.com/broadinstitute/cromwell/issues/936

engine/src/main/scala/cromwell/webservice/MetadataBuilderActor.scala
mcovarr: Here's Wikipedia's menu of Group-like category theory [structures](https://en.wikipedia.org/wiki/Category_(mathematics)).  I thought this `IndexedEmptyJson` looked kind of like an identity element for `IndexedJsonValue`s.  An identity element is the only difference between the `Semigroup` currently used here and a `Monoid`.

geoffjentry: I remember thinking a while back that since this stuff has a natural zero that perhaps one could combine monoid and foldable and replace some of the appends + such w/ folds. I think it's changed a fair amount since then though so YMMV

Horneth: We can actually have a small tech talk about it, but I'm not sure the zero here ought to have the exact same behaviour a zero has in a Monoid.
There also might be something missing/needing to be clarified in the mini language to push an empty list.

geoffjentry: I don't think it's really worth worrying about one way or the other TBH

engine/src/main/scala/cromwell/webservice/metadata/IndexedJsonValue.scala
```diff
@@ -0,0 +1,79 @@
+package cromwell.webservice.metadata
+
+import spray.json._
+
+import scalaz.{Monoid, Semigroup}
+import scalaz.Scalaz._
+
+private object IndexedJsonValue {
+   implicit val IndexedJsonSemigroup: Semigroup[IndexedJsonValue] = {
```
mcovarr: The naming is now inconsistent, plus I'm not sure how this is working given that it's declared as a semigroup?

engine/src/main/scala/cromwell/webservice/metadata/IndexedJsonValue.scala
```diff
@@ -0,0 +1,79 @@
...
+
+private case class IndexedJsonObject(v: Map[String, IndexedJsonValue], index: Option[Int] = None) extends IndexedJsonValue {
+  override val toJson = JsObject(v mapValues { _.toJson })
+  override def withIndex(index: String) = this.copy(index = toIndex(index))
+}
+
+private case class IndexedPrimitiveJson(v: JsValue, index: Option[Int] = None) extends IndexedJsonValue {
+  override val toJson = v
+  override def withIndex(index: String) = this.copy(index = toIndex(index))
+}
```
mcovarr: \n

mcovarr:
:+1:

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -153,6 +153,10 @@ class WorkflowActor(workflowId: WorkflowId,
 
   when(WorkflowUnstartedState) {
     case Event(StartWorkflowCommand, _) =>
+      // Is this the right place for startTime ?
```
cjllanwarne: I think so...!

cjllanwarne:
LGTM! 👍

Horneth:
So I got the metadata processing from 30s to ~ 300ms, and instead of rebasing this, merging it, and then rebasing the performance branch again I'm gonna close this one and open a new PR with the better algorithm + those changes. Sorry for the reviewers :/

Horneth:
https://github.com/broadinstitute/cromwell/pull/981

PR #971, opened 2016-06-07
Metadata journal timestamp to microseconds.

mcovarr:


database/src/main/resources/changesets/metadata_journal_subsecond_timestamp.xml
```diff
@@ -0,0 +1,19 @@
...
+                columnName="METADATA_TIMESTAMP"
+                newDataType="DATETIME(6)"
+                tableName="METADATA_JOURNAL"/>
+    </changeSet>
+
+    <changeSet id="metadata_journal_timestamp_not_null" author="mcovarr">
+        <addNotNullConstraint
+                tableName="METADATA_JOURNAL"
+                columnName="METADATA_TIMESTAMP"
+                columnDataType="DATETIME(6)"/>
```
cjllanwarne: Are both of these really necessary? Presumably if the second query requires the data type, it allows it to be modified? (otherwise what's the point!?)

cjllanwarne:
👍

geoffjentry:
👍  modulo resolving things w/ Chris

database/src/main/resources/changesets/metadata_journal_subsecond_timestamp.xml
mcovarr: Yes they are both required for MySQL, strange but true.

Issue #814, opened 2016-05-09
Implement workflow logging


mcovarr:

mcovarr:
Is this 0.20-worthy?  It seems that workflow logs are kind of a big deal and currently we aren't even trying to support them in the New Worlde.

geoffjentry:
@mcovarr I can't imagine this being useful to GOTC w/o workflow logs

cjllanwarne:
Is this ticket related to these workflow options?
```
{
    "workflow_log_dir": "gs://cloud-cromwell-dev/centaur-tests/direct_wf_logs"
    "call_logs_dir": "gs://cloud-cromwell-dev/centaur-tests/direct_call_logs"
    "outputs_path": "gs://cloud-cromwell-dev/centaur-tests/direct_outputs"
} 
```

PR #973, opened 2016-06-08
Reimplemented final calls in the shadow world. Closes #773

### LINKED ISSUE #773: Simplify FinalCalls to functions, instead of tracking in ExecutionStore
Find a way to perform FinalCalls as functions with a WorkflowActor State 

WorkflowActor uses some state to run various final file copies: 
- Possibly use the existing Finalizing state
- Possibly create another state that means the workflow is in this phase of running this specific group operations

If the process dies during this state, upon restart, the workflow would retry copying all of the files, meaning the final actions: 
- Do not need to be a CallKey or other JobKey/ExecutionStoreKey
- Do not need to be tracked in the WorkflowActor ExecutionStore
- Do not need to be individually tracked with state in the database’s Execution table
- Will possibly use more IO than strictly necessary on restart, as we don’t know what copies succeeded and what failed

kshakir:
Concrete implementation of `BackendLifecycleActorFactory.workflowFinalizationActorProps` returning `None`, instead of every subclass needing to do it.
Added a `BackendLifecycleActorFactory.getExecutionRootPath: Path` for use in converting output path strings to `java.nio.Path`.
`WdlFunctions.gcsFileSystem` returns `None` instead of throwing an exception.
`WorkflowExecutionActor.pushWorkflowOutputMetadata` only pushes workflow/reportable outputs, not all outputs.
`WorkflowLifecycleActorData.actors` also tracks the new `EngineWorkflowFinalizationActor`s in addition to the `BackendWorkflowFinalizationActor`s.
`JesFinalizationActor` runs on the `slow-actor-dispatcher` instead of the main akka system dispatcher.

Moved several classes to `core`:
- `JobKey`
- `ExecutionIndex`
- `ExecutionStatus`
- `ExecutionStore`
- `OutputStore`


kshakir:
@mcovarr and @geoffjentry wheel-chosen

backend/src/main/scala/cromwell/backend/io/WorkflowPaths.scala
```diff
@@ -3,15 +3,15 @@ package cromwell.backend.io
...
 import com.typesafe.config.Config
-import cromwell.backend.{BackendJobDescriptorKey, BackendWorkflowDescriptor}
+import cromwell.backend.BackendWorkflowDescriptor
 
 object WorkflowPaths{
   val DockerRoot = Paths.get("/root")
 }
 
 class WorkflowPaths(workflowDescriptor: BackendWorkflowDescriptor, config: Config) {
   // TODO will need update when configuration is rearranged
```
mcovarr: Has this TODO outlived its usefulness? I'm not sure what it could be asking for that we don't already have.

kshakir: Not sure myself... pinging @gauravs90 for comment on whether it should be saved or not. I'll defer to deleting the line unless we get clarification on keeping it.

gauravs90: I think @mcovarr is right. Although a comment would be good that the `config` that it expects is the backend config and not the global one. @Horneth might also have some opinions, just in case.

Horneth: Yes I think I wrote this comment, it was at the time when @mcovarr was re-arranging the config and the value for the base directory attribute (which is now "root") wasn't fully defined or whatever. It can be removed.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/CopyWorkflowLogsActor.scala
```diff
@@ -0,0 +1,34 @@
...
+  extends EngineWorkflowCopyFinalizationActor {
+
+  override def copyFiles(): Unit = {
+    for {
+      workflowLogDirString <- getWorkflowOption("workflow_log_dir")
+      tempLogFilePath <- getTempLogFilePathOption
+    } yield copyWorkflowLog(tempLogFilePath, workflowLogDirString)
+  }
+
+  // TODO: PBE: Where are the workflow logs?
```
mcovarr: hiding in #814

backend/src/main/scala/cromwell/backend/BackendLifecycleActorFactory.scala
```diff
@@ -1,15 +1,25 @@
...
   def workflowFinalizationActorProps(workflowDescriptor: BackendWorkflowDescriptor,
-                                     calls: Seq[Call]): Option[Props]
+                                     calls: Seq[Call],
+                                     executionStore: ExecutionStore,
+                                     outputStore: OutputStore): Option[Props] = None
+
   def expressionLanguageFunctions(workflowDescriptor: BackendWorkflowDescriptor,
                                   jobKey: BackendJobDescriptorKey): WdlStandardLibraryFunctions
+
+  def getExecutionRootPath(workflowDescriptor: BackendWorkflowDescriptor): Option[Path] = None
```
mcovarr: In what circumstances would a `None` return be ok?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/CopyWorkflowOutputsActor.scala
```diff
@@ -0,0 +1,63 @@
...
+      if outputCallKey.call == call
+      outputEntry <- outputEntries
+      if reportableOutput.fullyQualifiedName == s"${call.fullyQualifiedName}.${outputEntry.name}"
+      wdlValue <- outputEntry.wdlValue.toSeq
+      collected = wdlValue collectAsSeq { case f: WdlSingleFile => f }
+      wdlFile <- collected
+      wdlPath = rootPath.getFileSystem.getPath(wdlFile.value)
+    } yield (rootPath, wdlPath)
+
+    // Somewhere, up there, we're duplicating rows in scatter results
```
mcovarr: uh that's kind of scary and perhaps worth investigating

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesFinalizationActor.scala
```diff
@@ -1,36 +1,90 @@
...
+  }
+
+  private def copyCallOutputs(): Future[Unit] = {
+    /*
+    NOTE: Only using one thread pool slot here to upload all the files for all the calls.
+    One could also use Future.sequence to flood either the context.system, potentially starving out the actors, or
+    even create a separate jes final call specific thread pool for parallel uploads.
+
+    Measure and optimize as necessary. Will likely need retry code at some level as well.
+     */
```
mcovarr: Some very good points here!

engine/src/main/scala/cromwell/engine/workflow/lifecycle/CopyWorkflowOutputsActor.scala
mcovarr: that's a TOL and not meant for this PR

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -207,7 +209,9 @@ class WorkflowActor(workflowId: WorkflowId,
 
   when(WorkflowAbortingState) {
     case Event(x: EngineLifecycleStateCompleteResponse, data @ WorkflowActorData(_, Some(workflowDescriptor), _)) =>
-      finalizeWorkflow(data, workflowDescriptor, failures = None)
+      // TODO: PBE: some of the x-es have an actually execution & output stores.
+      // But do we want the finalization to operate on that data during state aborting?
+      finalizeWorkflow(data, workflowDescriptor, ExecutionStore.empty, OutputStore.empty, failures = None)
```
mcovarr: Not sure what the expectations are here currently but we should ask

mcovarr:
:+1: let's merge this before the world changes again!  😄

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -207,7 +209,9 @@ class WorkflowActor(workflowId: WorkflowId,
 
   when(WorkflowAbortingState) {
     case Event(x: EngineLifecycleStateCompleteResponse, data @ WorkflowActorData(_, Some(workflowDescriptor), _)) =>
-      finalizeWorkflow(data, workflowDescriptor, failures = None)
+      // TODO: PBE: some of the x-es have an actually execution & output stores.
+      // But do we want the finalization to operate on that data during state aborting?
```
geoffjentry: I don't understand what you mean by this todo

engine/src/main/scala/cromwell/engine/workflow/lifecycle/CopyWorkflowOutputsActor.scala
```diff
@@ -0,0 +1,64 @@
...
+    }
+  }
+
+  private def getOutputFilePaths(reportableOutputs: Seq[ReportableSymbol]): Seq[(Path, Path)] = {
+    val results = for {
+      reportableOutput <- reportableOutputs
+      // NOTE: Without .toSeq, output in arrays only yield the last output
+      (backend, calls) <- workflowDescriptor.backendAssignments.groupBy(_._2).mapValues(_.keys.toSeq).toSeq
+      config <- BackendConfiguration.backendConfigurationDescriptor(backend).toOption.toSeq
+      rootPath <- CromwellBackends.shadowBackendLifecycleFactory(backend).toOption. // Try/Option/Seq soup
```
geoffjentry: Would it be possible to pull the rhs of this one into a function w/ a descriptive name? I'm thinking that it's a 2 line rhs which requires a comment, it makes it tough when parsing all the steps of the forcomp

engine/src/main/scala/cromwell/engine/workflow/lifecycle/CopyWorkflowOutputsActor.scala
geoffjentry: I'd feel better if it was figured out prior to merging but a ticket would suffice :)

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowFinalizationActor.scala
```diff
@@ -58,21 +64,36 @@ case class WorkflowFinalizationActor(workflowId: WorkflowId, workflowDescriptor:
...
         case Failure(ex) =>
           sender ! WorkflowFinalizationFailedResponse(Seq(ex))
           goto(WorkflowFinalizationFailedState)
         case Success(actors) if actors.isEmpty =>
           sender ! WorkflowFinalizationSucceededResponse
           goto(FinalizationSucceededState)
         case Success(actors) =>
-          actors.keys.foreach(_ ! Finalize)
-          goto(FinalizationInProgressState) using stateData.withBackendActors(actors)
+          val actorSet = actors.toSet
```
geoffjentry: Where does the potential duplication come in?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowInitializationActor.scala
```diff
@@ -63,12 +63,11 @@ case class WorkflowInitializationActor(workflowId: WorkflowId, workflowDescripto
       val backendInitializationActors = Try {
         for {
           (backend, calls) <- workflowDescriptor.backendAssignments.groupBy(_._2).mapValues(_.keys.toSeq)
-          backendConfiguration = BackendConfiguration.backendConfigurationDescriptor(backend).get
           props <- CromwellBackends.shadowBackendLifecycleFactory(backend).map(factory =>
             factory.workflowInitializationActorProps(workflowDescriptor.backendDescriptor, calls)
           ).get
           actor = context.actorOf(props)
-        } yield (actor, backend)
```
geoffjentry: I saw this up above too but couldn't figure out why it was no longer necessary. Can you explain for my own edification (i.e. this isn't an "Explain yourself!" comment :) )

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesFinalizationActor.scala
geoffjentry: You might want to put this on the slow stuff dispatcher (or whatever it is called) instead of the main dispatcher so as not to tie up a thread in the main system while all of this is going on

geoffjentry:
👍

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowFinalizationActor.scala
kshakir: Actor ref tracking used to be a `Map[ActorRef, String]`, where the values were never used. When getting rid of the values, that only left the map's keys, a `Set[ActorRef]`.

I'm guessing the original author was just tracking the `ActorRef`/`String` pair as a sequence of tuples, perhaps for debugging purposes?, since I couldn't find any use the backends in the values. As I needed to track completion of engine level finalization also, I could either add a separate collection for the engine actor refs along side the collection for the backend actor refs, or just merge the two into this single collection.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowInitializationActor.scala
kshakir: TL;DR of the previous comment-- the backend in the map values was never used, and I need to store engine actor refs in the map keys, now just a set of keys.

geoffjentry: Yeah, as soon as I started reading the preivous one I realized they were tied together. The "I couldn't find any use" also explains my confusion I believe :)

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
kshakir: During this aborting state, we could potentially finalize based on already completed executions and outputs. On the next line though, we just finalize as if no execution nor outputs were generated.

Unlike the other classes extending `WorkflowExecutionActorResponse` (WEActorR) `WorkflowExecutionAbortedResponse` (WEAbortedR) is both a `EngineLifecycleStateCompleteResponse` (ELSCR) and a WEActorR. Any WEActorR ~~will~~ could contain an `ExecutionStore` and `OutputStore` up until the point the workflow was aborted. However, the ELSCR trait is the only one we `case` match on though, for now. I don't really understand the current separation between the different classes implementing WEActorR and ELSCR, so this note is a hint to future implementor that one could get the execution and output stores if they change the matching type of `x`.

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesFinalizationActor.scala
kshakir: I understand the concept of creating a different dispatcher, but after googling I'm not sure if I'm missing an already instantiated one in the akka ecosystem. Let's chat IRL.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowFinalizationActor.scala
kshakir: Found and fixed. It was that scatters are in the store as calls with and without indexes.

PR #976, opened 2016-06-08
Metadata entries for early failures. Closes #966

### LINKED ISSUE #966: Add metadata for WorkflowMaterialiserActor Failure
If there is a failure in the `WorkflowMaterialiserActor` phase, there is no external way to debug the error.

Allow the `WorkflowMaterialiserActor` to send failure messages to the metadata service in this instance, so that it's possible to debug errors!

cjllanwarne:


mcovarr:
:+1:

engine/src/test/scala/cromwell/engine/backend/WorkflowDescriptorSpec.scala
```diff
@@ -53,7 +53,8 @@ trait WorkflowDescriptorBuilder {
     implicit val timeout = akka.util.Timeout(awaitTimeout)
     implicit val ec = actorSystem.dispatcher
 
-    val actor = actorSystem.actorOf(MaterializeWorkflowDescriptorActor.props(), "MaterializeWorkflowDescriptorActor-" + id.id)
+    val serviceRegistryIgnorer = actorSystem.actorOf(DevNullActor.props)
```
geoffjentry: I was super scared that we had `DevNullActor` until I noticed the name of this file

kshakir: [`Props.empty`](https://github.com/akka/akka/blob/v2.3.12/akka-actor/src/main/scala/akka/actor/Props.scala#L49) does the same as your `DevNullActor.props`:

```scala
val serviceRegistryIgnorer = actorSystem.actorOf(Props.empty)
```

Horneth: You can also just do `val serviceRegistryIgnorer = TestProbe().ref`

scottfrazer:
👍

PR #977, opened 2016-06-08
resolve workflow state for the metadata endpoint

mcovarr:
The metadata endpoint was previously not applying any business logic to workflow states so it simply returned the last state written to the journal.  This adds specific resolution logic similar to that used for workflow summary.


geoffjentry:
👍 for now

I do think we'll need to merge this with the summary table but that doesn't have to happen now

kshakir:
👍

PR #979, opened 2016-06-09
Force ServiceRegistryActor initialization.  Closes #871

### LINKED ISSUE #871: Missing Service Configuration should fail Cromwell initialisation
If your service configuration is missing or invalid, Cromwell should fail to start up.

Currently, nothing is seen until workflows are submitted and immediately fail

mcovarr:


cjllanwarne:
So this obviously prevents the system from "swapping out" the `ServiceRegistryActor` half way through... but I think that's completely fine!

👍

scottfrazer:
👍

PR #985, opened 2016-06-10
Workflow outputs API.  Closes #945

### LINKED ISSUE #945: Make /workflows/{version}/{id}/outputs great again
This is just a subset of the metadata REST API data... as a first pass, reuse as much as possible.  As a future path this could all be flavors of query (flexible filtering and column selection)

mcovarr:


geoffjentry:
👍

engine/src/main/scala/cromwell/webservice/metadata/MetadataBuilderActor.scala
```diff
@@ -214,6 +214,7 @@ class MetadataBuilderActor(serviceRegistryActor: ActorRef) extends LoggingFSM[Me
       context.parent ! RequestComplete(StatusCodes.OK, processMetadataResponse(query, metadata))
       allDone
     case Event(StatusLookupResponse(w, status), _) =>
+      // TODO PBE refactor this to use workflowMetadataResponse with includeCallsIfEmpty = false
```
mcovarr: er actually don't do this, status goes against the summary table and not the journal

database/src/main/scala/cromwell/database/SqlDatabase.scala
```diff
@@ -180,6 +180,10 @@ trait SqlDatabase extends AutoCloseable {
                                     attempt: Int)
                                    (implicit ec: ExecutionContext): Future[Seq[Metadatum]]
 
+  protected def queryMetadataEventsWithKeyLike(workflowUuid: String,
```
cjllanwarne: Or in other words, `queryMetadata`?

cjllanwarne: `queryMetadataWithWildcardKey`?

database/src/main/scala/cromwell/database/slick/MetadataComponent.scala
```diff
@@ -50,6 +50,16 @@ trait MetadataComponent {
       if metadatum.attempt.isEmpty
     } yield metadatum)
 
+  val metadataByWorkflowUuidWithKeyLike = Compiled(
```
cjllanwarne: `...WithWildcardKey`

engine/src/main/scala/cromwell/services/EngineMetadataServiceActor.scala
```diff
@@ -83,6 +83,14 @@ case class EngineMetadataServiceActor(serviceConfig: Config, globalConfig: Confi
     }
   }
 
+  private def queryWorkflowOutputsAndRespond(id: WorkflowId): Unit = {
+    val replyTo = sender()
+    dataAccess.queryWorkflowOutputs(id) onComplete {
+      case Success(o) => replyTo ! WorkflowOutputsResponse(id, o)
+      case Failure(t) => replyTo ! WorkflowOutputsFailure(id, t)
```
cjllanwarne: Is there logic anywhere which can tell the difference between no-such-id and no-outputs-for-this-workflow?

engine/src/main/scala/cromwell/webservice/metadata/MetadataBuilderActor.scala
```diff
@@ -172,17 +172,17 @@ object MetadataBuilderActor {
     } mapValues { md => JsArray(md.toVector.sortBy(_.index) flatMap { _.metadata }) }
 
     val wrappedCalls = JsObject(Map(WorkflowMetadataKeys.Calls -> JsObject(callsMap)))
-
-    JsObject(foldedWorkflowValues.fields ++ wrappedCalls.fields)
+    val callData = if (callsMap.isEmpty && !includeCallsIfEmpty) Nil else wrappedCalls.fields
```
cjllanwarne: Does this clash with @Horneth's empty objects/lists stuff?

cjllanwarne:
A few comments but I'll trust you with the due diligence, and then 👍

database/src/main/scala/cromwell/database/SqlDatabase.scala
geoffjentry: I'll add in that I found the `KeyLike` thing unwieldy but not enough to chime in. Since it has been brought up by @cjllanwarne I'll add a +0.5

engine/src/main/scala/cromwell/services/EngineMetadataServiceActor.scala
mcovarr: no there isn't and I need to fix that 😦

database/src/main/scala/cromwell/database/SqlDatabase.scala
mcovarr: So much for my plans for a Valleyspeak API  😢

database/src/main/scala/cromwell/database/slick/MetadataComponent.scala
mcovarr: totally

database/src/main/scala/cromwell/database/slick/SlickDatabase.scala
```diff
@@ -648,6 +648,13 @@ class SlickDatabase(databaseConfig: Config) extends SqlDatabase {
     runTransaction(action)
   }
 
+  protected def queryMetadataEventsWithKeyLike(workflowUuid: String,
```
mcovarr: for sure

engine/src/main/scala/cromwell/webservice/metadata/MetadataBuilderActor.scala
mcovarr: No it shouldn't, the only way the behavior changes is if the `includeCallsIfEmpty` flag is set and it's only set on the workflow outputs code path.

Horneth: I don't think so, this prevents from having `"calls": {}` in some cases if I understand which doesn't affect the empty stuff

database/src/main/scala/cromwell/database/SqlDatabase.scala
geoffjentry: groovy to the max

database/src/main/scala/cromwell/database/slick/MetadataComponent.scala
geoffjentry: gag me with a spoon

engine/src/main/scala/cromwell/services/EngineMetadataServiceActor.scala
mcovarr: Also broken on the [metadata endpoint](https://github.com/broadinstitute/cromwell/issues/988).

mcovarr: Actually after further thought and discussion with @cjllanwarne I'll just ticket this rather than fix it as part of this PR.  I'm not sure we currently have enough information in the PBE Cromwell DB to definitively know whether a workflow exists or not.  It's not necessarily correct to infer from the absence of workflow outputs that a workflow doesn't exist; it's possible the outputs haven't been written yet, or a given workflow is all side effects and has no formally declared outputs.  Perhaps we might infer that the presence of *any* metadata for a workflow ID should imply the existence of the workflow, but given the fully fire-and-forget nature of metadata writes I don't think that's quite the contract we have on the metadata journal at the moment.

mcovarr: Ticket for [this endpoint](https://github.com/broadinstitute/cromwell/issues/990).

mcovarr: There is a workflow existence cache in the MetadataActor that probably figures into all this.

PR #986, opened 2016-06-10
HOTFIX: Forcibly abort workflows stuck in aborting for longer than 10 minutes…

### LINKED ISSUE #940: Add stuck workflow monitor to 0.19_hotfix
@abaumann said that they're having a serious issue (due to BBQ) with workflows which have become stuck in non-terminal states. Set up some monitor which runs at a configurable period to look through all non-terminal workflows/calls and make sure that things are actually non-terminal, switching to terminal states as appropriate.

This solution is viewed as a relatively painless (vs fixing it For Real) way of solving the problem, but if one sees an easier way of doing it that's even better.

geoffjentry:
…. Closes #940 

Note that the tests for aborting were already commented out, and as this would be tricky to test anyways I took that as a sign from above. I've manually verified it, however.

@abaumann - this will not clean up previously stuck workflows, I'm going to provide a script or something like that to clean up the old crap, fixing those from within Cromwell was building up too many tendrils for a hack fix to something which is going to be replaced in the near future.

### DIFF
--- a/engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
@@ -30,12 +30,14 @@ import wdl4s.values.{SymbolHash, WdlArray, WdlCallOutputsObject, WdlValue}
 
 import scala.annotation.tailrec
 import scala.concurrent.ExecutionContext.Implicits.global
-import scala.concurrent.Future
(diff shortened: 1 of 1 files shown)


mcovarr:
:+1:

kshakir:
:+1:

geoffjentry:
Will coordinate with @ndbolliger on how to proceed from here

PR #987, opened 2016-06-10
Run a round of Centaur during travis. Closes a long national nightmare

geoffjentry:


travis_run_centaur.sh
```diff
@@ -0,0 +1,31 @@
...
+# Build and run centaur
+echo "Cloning Centaur"
+git clone https://github.com/broadinstitute/centaur.git
+
+cd centaur
+
+echo "Running Centaur with ${PARALLELISM_FACTOR}-way parallelism"
+./run_tests_parallel.sh "${PARALLELISM_FACTOR}"
+
+exit 0
```
geoffjentry: If anyone looks before I fix this, this kinda ruins the point :)

geoffjentry: In fact the exit trap is gonna make that not so good altogether

geoffjentry:
Note that the travis failure is a good thing (in that the change is working, not that it's an awesome circumstance), as it was due to not all of the centaur tests passing

geoffjentry:
And I still need to run the script through that linter that @kshakir made me use the last time

cjllanwarne:
The centaur tests actually failed because `Futures timed out after 10000ms`. Since we're trying to check correctness rather than Local backend scalability performance here, can we run this using the script?

geoffjentry:
It is using the script, but with 3-way parallel. Trying it with 1 but can't go any lower than that.

The travis machines are jankier than what we normally run on, but IMO that's a good thing here.

I'd also question (worth chatting about, as I don't even know what I think) if correctness vs parallel is really the right thing to be testing here.

centaur.json
```diff
@@ -0,0 +1,5 @@
+{
+  "centaur.centaur.cromwell_branch":"BRANCH",
+  "centaur.centaur.conf":"gs://cloud-cromwell-dev/travis-centaur/jes.conf",
+  "centaur.centaur.pem":"gs://cloud-cromwell-dev/travis-centaur/cromwell-account.pem"
```
cjllanwarne: I take it this is appropriately ACL'd?

I was able to gsutil cat it, FWIW

travis.wdl
```diff
@@ -0,0 +1,31 @@
...
+        cpu: "32"
+        memory: "10 GB"
+        zones: "us-central1-b"
+        failOnStderr: false
+    }
+}
+
+workflow centaur {
+    call centaur
+}
```
cjllanwarne: Insufficient newline

travis.wdl
```diff
@@ -0,0 +1,31 @@
...
+    >>>
+
+    output {
+        File cromwell_log = "/cromwell_root/logs/cromwell.log"
+        File centaur_log_file = "/cromwell_root/logs/centaur.log"
+    }
+
+    runtime {
+        docker: "geoffjentry/centaur-cromwell:latest"
+        cpu: "32"
```
cjllanwarne: Really necessary? That's mildly concerning

PR #992, opened 2016-06-13
Rewire SingleWorkflowRunnerActor for the New Worlde.  Closes #960

### LINKED ISSUE #960: single workflow mode doesn't exit
When I run

`java -jar -jar target/scala-2.11/cromwell-0.20.jar run 3step.original.wdl inputs.json`

the pipeline completes, but cromwell just sits there.  Previously, it would exit on completion

mcovarr:


engine/src/main/scala/cromwell/engine/workflow/SingleWorkflowRunnerActor.scala
```diff
@@ -97,19 +100,15 @@ case class OldStyleSingleWorkflowRunnerActor(source: WorkflowSourceFiles,
   }
 
   when (RequestingOutputs) {
-    case Event(WorkflowManagerWorkflowOutputsSuccess(id, outputs), data) =>
-      // Outputs go to stdout
+    case Event(RequestComplete((StatusCodes.OK, outputs: JsObject)), _) =>
```
cjllanwarne: I think we should be careful about making API depend on engine, and not vice versa. Unless we consider `SingleRunnerActor` an API-package actor?

cjllanwarne:
👍

Horneth:
👍  modulo the MetadataBuilder / API discussion

mcovarr:
whoops forgot to fix up the test... 😦

engine/src/main/scala/cromwell/engine/workflow/SingleWorkflowRunnerActor.scala
```diff
@@ -87,7 +89,8 @@ case class OldStyleSingleWorkflowRunnerActor(source: WorkflowSourceFiles,
       workflowManager ! SubscribeToWorkflowCommand(id)
       stay() using data.copy(id = Option(id))
     case Event(Transition(_, _, WorkflowSucceededState), data) =>
-      workflowManager ! WorkflowOutputs(data.id.get)
+      val metadataBuilder = context.actorOf(MetadataBuilderActor.props(serviceRegistryActor))
```
kshakir: Pass `actorOf` an actor `name`.

engine/src/main/scala/cromwell/util/FileUtil.scala
```diff
@@ -8,15 +8,14 @@ import wdl4s.values.Hashable
...
-@deprecated(message = "This class will not be part of the PBE universe", since = "May 2nd 2016")
 object FileUtil {
   def swapExt(filePath: String, oldExt: String, newExt: String): String = {
     filePath.stripSuffix(oldExt) + newExt
   }
 
   def parseTsv(tsv: String): Try[Array[Array[String]]] = {
     val table = tsv.split("\n").map(_.split("\t"))
-    table.map(_.size).toSet match {
+    table.map(_.length).toSet match {
```
geoffjentry: FWIW I'm pretty sure the performance gotcha is no longer true on this

geoffjentry: hrm, perhaps not

mcovarr:
@geoffjentry @Horneth this could use a re-review since I updated it a bunch

engine/src/main/scala/cromwell/util/FileUtil.scala
mcovarr: IntelliJ still whines about it, which is what prompted the change

engine/src/main/scala/cromwell/engine/workflow/SingleWorkflowRunnerActor.scala
```diff
@@ -1,32 +1,41 @@
...
 import spray.json._
 
-import scala.util._
-@deprecated(message = "This class will not be part of the PBE universe", since = "May 2nd 2016")
-object OldStyleSingleWorkflowRunnerActor {
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.duration._
+import scala.language.postfixOps
+import scala.util.{Failure, Try}
+
```
geoffjentry: I think it's worth putting a description in somewhere about what all is going on with the polling, why it's there, etc

engine/src/main/scala/cromwell/engine/workflow/SingleWorkflowRunnerActor.scala
```diff
@@ -45,29 +54,45 @@ object OldStyleSingleWorkflowRunnerActor {
...
+case class SingleWorkflowRunnerActor(source: WorkflowSourceFiles,
+                                     metadataOutputPath: Option[Path],
+                                     workflowManager: ActorRef)
+  extends LoggingFSM[RunnerState, RunnerData] with CromwellActor with ServiceRegistryClient {
 
-  import OldStyleSingleWorkflowRunnerActor._
+  import SingleWorkflowRunnerActor._
 
-  val tag = "SingleWorkflowRunnerActor"
+  private val tag = "SingleWorkflowRunnerActor"
```
geoffjentry: put this in the companion object?

geoffjentry:
LGTM

engine/src/main/scala/cromwell/engine/workflow/SingleWorkflowRunnerActor.scala
mcovarr: ok

engine/src/main/scala/cromwell/engine/workflow/SingleWorkflowRunnerActor.scala
mcovarr: sure

engine/src/main/scala/cromwell/engine/workflow/SingleWorkflowRunnerActor.scala
```diff
@@ -84,32 +109,35 @@ case class OldStyleSingleWorkflowRunnerActor(source: WorkflowSourceFiles,
...
       log.info(s"$tag: Workflow submitted UUID($id)")
-      workflowManager ! SubscribeToWorkflowCommand(id)
+      schedulePollRequest()
       stay() using data.copy(id = Option(id))
-    case Event(Transition(_, _, WorkflowSucceededState), data) =>
-      workflowManager ! WorkflowOutputs(data.id.get)
+    case Event(IssuePollRequest, _) =>
+      requestStatus()
+      stay()
+    case Event(RequestComplete((StatusCodes.OK, jsObject: JsObject)), data) if !jsObject.state.isTerminal =>
```
Horneth: Should we have a handler somewhere for `StatusCode != OK` ?

engine/src/main/scala/cromwell/webservice/metadata/MetadataBuilderActor.scala
```diff
@@ -131,7 +133,9 @@ object MetadataBuilderActor {
 
   /** Sort events by timestamp, transform them into TimestampedJsValues, and merge them together. */
   private def eventsToIndexedJson(events: Seq[MetadataEvent]): TimestampedJsValue = {
-     events map { e => keyValueToIndexedJson(e.key.key, e.value, e.offsetDateTime) } reduce(_ |+| _)
+    // The `List` has a `Foldable` instance defined in scope, and because the `List`'s elements have a `Monoid` instance
+    // defined in scope, `suml` can derive a sane `TimestampedJsValue` value even if the `List` of events is empty.
+    events.toList map { e => keyValueToIndexedJson(e.key.key, e.value, e.offsetDateTime) } suml
```
Horneth: TOL and not a hold back for this PR at all. I think `suml` is a foldLeft behind the hood. The advantage of `reduce` (even if we don't really use it here - but we could with a parallel collection or something else) is that the "reducing" can be done in parallel because the `TimestampedJsValue` Monoid is now commutative. Whereas the `foldLeft` will be done sequentially.

mcovarr: Yes you're right that `suml` is implemented as `foldLeft` using the `Monoid.zero` for the initial value.  I made this change because the current `reduce` code crashes when the list is empty (not just because moar category theory 😄 ).  We can certainly revisit this later if it turns out to warrant a parallel implementation.

engine/src/main/scala/cromwell/engine/workflow/SingleWorkflowRunnerActor.scala
mcovarr: Yes I think you're right.  The generic `whenUnhandled` case doesn't really capture the essence of "not okay" in `StatusCode != OK`.  😄

PR #1008, opened 2016-06-15
Workflow logs are great again.  Closes #946

### LINKED ISSUE #946: Make workflows/{version}/{id}/logs great again
This is just a subset of the metadata REST API data... as a first pass, reuse as much as possible.  As a future path this could all be flavors of query (flexible filtering and column selection)

mcovarr:


engine/src/main/scala/cromwell/services/EngineMetadataServiceActor.scala
```diff
@@ -91,6 +91,14 @@ case class EngineMetadataServiceActor(serviceConfig: Config, globalConfig: Confi
     }
   }
 
+  private def queryLogsAndRespond(id: WorkflowId): Unit = {
+    val replyTo = sender()
```
geoffjentry: Mixing it up? :)

I like this better than `sndr` fwiw

mcovarr: Yeah me too, I saw someone else doing this and decided to mend my ways.

geoffjentry: I was never quite sure what you had against vowels

mcovarr: maybe a good name for an Akka-based startup

Horneth:
👍

geoffjentry:
👍

PR #1009, opened 2016-06-15
Tagged Caching and Restart as PostMVP and re-enabled a few tests Closes #996

### LINKED ISSUE #996: Tag Call Caching & Restart tests as 'PostMVP' and exclude in 'sbt test'
Currently, we have a number of ignored tests that are part of the default 'sbt test' configuration which we are not planning on re-enabling for MVP.  By excluding these we (a) don't lose them and (b) should be able to drive the ignored tests to 0 (yay!)

Horneth:


mcovarr:
:+1:

geoffjentry:
👍

PR #1010, opened 2016-06-16
Cleanup retry.  Closes #1004

### LINKED ISSUE #1004: AsyncBackendJobExecutionActor has its own retry
yet we have a generic retry system

mcovarr:


kshakir:
👍

cjllanwarne:
Codewise this is fine. 👍 

But... I'm still not convinced that these floating futures are the best way to do this and I don't really follow what's going on in this class. Why not a sub-actor in place of retrying Futures?

kshakir:
(ToL, of course)

I _still_ don't always know where to draw the line between creating a new `akka..Actor` and using a `scala..Future`. Something does "smell" funny though about the way we:

- Queue of things-to-do is on the `ec: ExecutionContext`
- A mailbox `message: AnyRef` is received off the `ec` by the dispatcher and passed to our `actor: Actor`.
- Instead of running a `runnable: Runnable` bit of code immediately, the `actor` chooses to throws the `runnable` onto the back of the `ec` queue and then say "done processing `message`".

That said, [this blog](https://www.chrisstucchio.com/blog/2013/actors_vs_futures.html) seems to say that `Actor` and `Future` can work together, but maybe something is off about how we're composing them in our `BackendLifecycleActor` interfaces.

geoffjentry:
@cjllanwarne There's nothing stopping you from submitting a future PR proposing the changes you describe

@kshakir Don't forget about akka streams, which sit in between futures and actors on the generalized concurrency spectrum. 

I recognized the URL of that blog post, I'll say that it has surprisingly useful comments at the end as well.

Personally I find it a lot easier to reason about actors & messages than futures but that's not true for everyone.

mcovarr:
I'm totally fine with using an actor rather than a floating Future here, that should allow for better concurrency management.  But the structure of the code prior to this PR was the lowest-cost refactoring of the Olde Worlde JES backend to the New World, while this PR addressed the issue raised in #1004 that there was effectively a duplication of `Retry.withRetry`.

If somebody wants to ticket replacing all the usages of `Retry.withRetry` with an Actor-based approach and lobby for that to be prioritized that's great, but making changes that broad seemed like more work than had been called for in this ticket.  Or you could just submit a PR as @geoffjentry suggests.  😄

PR #1014, opened 2016-06-17
Added options to reduce /metadata content. Closes #972

### LINKED ISSUE #972: This metadata is gonna be YUUUUUGE!
The new `super_massive_array_output` test in centaur has found a problem in metadata reading. It turns out that it generates quite a bit of metadata:
```
mysql> select count(*) from METADATA_JOURNAL where WORKFLOW_EXECUTION_UUID="31861832-66ee-45e6-ae10-891f55d176df";
+----------+
| count(*) |
+----------+
|  2278541 |
+----------+
```

Luckily, most of it is either inputs or outputs:
```
mysql> select count(*) from METADATA_JOURNAL where WORKFLOW_EXECUTION_UUID="31861832-66ee-45e6-ae10-891f55d176df" and `metadata_key` like "%output%" or `metadata_key` like "%input%";
+----------+
| count(*) |
+----------+
|  2268240 |
+----------+
```

The cromwell service is simply unable to muster 2.3 million rows of metadata into JSON in time or in the memory it has.

On scaling up the scatter from 500 to 20,000, the metadata rows required will almost certainly scale proportionally.

Luckily, this is a read-time problem, Cromwell was able to perform the 500x scatter effortlessly and completed successfully. Only the post-facto attempt to see metadata caused problems.

kshakir:
Replaced undocumented `timings` and `outputs` flags on `/metadata` endpoint with `includeKey`.
Added a `excludeKey` mirror of `includeKey`.
DRY-ed out the `SqlDatabase.queryMetadataEventsWithWildcardKey` by using the very similar `SqlDatabase.queryMetadataEventsWithWildcardKeys`.
Fixed compiler warnings in `BackendWorkflowInitializationActor` and `jes.Run`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1792,6 +1792,19 @@ Workflow metadata includes submission, start, and end datetimes, as well as stat
 Call-level metadata includes inputs, outputs, start and end datetime, backend-specific job id,
 return code, stdout and stderr.  Date formats are ISO with milliseconds.
 
+Accepted parameters are:
+
+* `includeKey` Optional repeated string value, specifies what metadata
+  keys to include in the output, matched as a prefix string. Keys that
+  are not specified are filtered out. The call keys `attempt` and
+  `shardIndex` will always be included. May not be used with
+  `excludeKey`.
+
+* `excludeKey` Optional repeated string value, specifies what metadata
+  keys to exclude from the output, matched as a prefix string. Keys that
+  are specified are filtered out. The call keys `attempt` and
+  `shardIndex` will always be included. May not be used with
+  `includeKey`.
 
 cURL:
 
@@ -1812,7 +1825,7 @@ Server spray-can/1.3.3 is not blacklisted
 Server: spray-can/1.3.3
 Date: Thu, 01 Oct 2015 22:18:07 GMT
 Content-Type: application/json; charset=UTF-8
-Content-Length: 8286
+Content-Length: 7286
 {
   "workflowName": "sc_test",
   "calls": {
@@ -1824,8 +1837,6 @@ Content-Length: 8286
         "outputs": {
           "split_files": [
             "/home/jdoe/cromwell/cromwell-test-executions/sc_test/8e592ed8-ebe5-4be0-8dcb-4073a41fe180/call-do_prepare/temp_aa",
-            "/home/jdoe/cromwell/cromwell-test-executions/sc_test/8e592ed8-ebe5-4be0-8dcb-4073a41fe180/call-do_prepare/temp_ab",
-            "/home/jdoe/cromwell/cromwell-test-executions/sc_test/8e592ed8-ebe5-4be0-8dcb-4073a41fe180/call-do_prepare/temp_ac",
             "/home/jdoe/cromwell/cromwell-test-executions/sc_test/8e592ed8-ebe5-4be0-8dcb-4073a41fe180/call-do_prepare/temp_ad"
           ]
         },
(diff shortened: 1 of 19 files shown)


kshakir:
Hmm... the interweb wheel, I believe for the second PR in a row, has chosen @geoffjentry and @mcovarr. I may need to switch to our fancy new wheel next time.

cjllanwarne:
ToL: It wasn't obvious until I found the README that this only applies to call level keys. Is there any way of also (en/dis)abling the workflow-level outputs and inputs (which can also be pretty large)?

ToL (only if trivial): any chance of an `excludeKey` too? When I call this I often am more interested in hiding a specific key rather than whitelisting a specific one

geoffjentry:
To echo @cjllanwarne - my interpretation of #972 isn't "call level only" but the whole enchilada

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
```diff
@@ -149,11 +145,15 @@ trait CromwellApiService extends HttpService with PerRequestCreator with Service
 
   def metadataRoute =
     path("workflows" / Segment / Segment / "metadata") { (version, workflowId) =>
-      parameters('outputs ? true, 'timings ? true).as(WorkflowMetadataQueryParameters) { parameters =>
+      parameterMultiMap { parameters =>
         Try(WorkflowId.fromString(workflowId)) match {
           case Success(w) =>
+            // import scalaz_ & Scalaz._ add too many slow implicits, on top of the spray and json implicits
+            import scalaz.syntax.std.list._
```
mcovarr: Yeah I should get in the habit of making these imports more targeted.

engine/src/main/scala/cromwell/webservice/metadata/MetadataBuilderActor.scala
```diff
@@ -163,7 +163,12 @@ object MetadataBuilderActor {
   private def parseWorkflowEventsToTimestampedJsValue(events: Seq[MetadataEvent], includeCallsIfEmpty: Boolean): JsObject = {
     // Partition if sequence of events in a pair of (Workflow level events, Call level events)
     val (workflowLevel, callLevel) = events partition { _.key.jobKey.isEmpty }
-    val foldedWorkflowValues = eventsToIndexedJson(reduceWorkflowEvents(workflowLevel)).toJson.asJsObject
+    val reducedEvents = reduceWorkflowEvents(workflowLevel)
+    val reducedFields = if (reducedEvents.isEmpty) {
+      Map.empty
+    } else {
```
mcovarr: Is this `if` actually required?  I would have thought the `suml` magic in `eventsToIndexedJson` would handle the case of an empty `reducedEvents` list correctly.

engine/src/main/scala/cromwell/engine/db/DataAccess.scala
```diff
@@ -373,19 +373,32 @@ trait DataAccess extends AutoCloseable {
...
     }
+
     futureMetadata map metadataToMetadataEvents(query.workflowId)
   }
 
   def queryWorkflowOutputs(id: WorkflowId)
                           (implicit ec: ExecutionContext): Future[Seq[MetadataEvent]] = {
     val uuid = id.id.toString
-    queryMetadataEventsWithWildcardKey(uuid, WorkflowMetadataKeys.Outputs + ":%", requireEmptyJobKey = true) map metadataToMetadataEvents(id)
+    queryMetadataEventsWithWildcardKeys(uuid, s"${WorkflowMetadataKeys.Outputs}:".wrapNel, requireEmptyJobKey = true).
```
mcovarr: ooh `wrapNel`, nice

mcovarr:
It's not really clear from reading #972 what the scope of this ticket should be, but "exclude" seems like it would be useful and hopefully not difficult to add?

mcovarr:
:+1:

mcovarr:
@cjllanwarne @geoffjentry the filtering appears to apply to both call and workflow level keys, it's just the docs that need adjusting

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/Run.scala
```diff
@@ -134,6 +134,8 @@ case class Run(runId: String,  jobDescriptor: BackendJobDescriptor, genomicsInte
           if (jesEndTime.offsetDateTime isAfter finalEventTime) jesEndTime else jesEndTime.copy(offsetDateTime = finalEventTime)
         case (Some(jesEndTime), None) => jesEndTime
         case (None, Some(finalEventTime)) => EventStartTime("cromwell poll interval", finalEventTime)
+        case (None, None) =>
+          throw new IllegalArgumentException("Both jesReportedEndTime and finalEventsListTime were None.")
```
geoffjentry: If you didn't already it'd be good to figure out what actually happens if an exception is thrown in here

geoffjentry:
👍

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/Run.scala
kshakir: Let's tech talk any extra action we want here. Effectively all I've done with this is replace a `scala.MatchError`<sup>1</sup> with a `java.lang.IllegalArgumentException`. I saw a compiler warning and removed it with this "fix". I haven't actually changed the behavior/expectation for the callers who might trip over either the exception. I suspect either would get swallowed somewhere, and maybe logged if we're lucky.

<sup>1</sup>Why on earth does `MatchError` not extend `java.lang.Error`, and [instead](http://www.scala-lang.org/api/2.11.7/index.html#scala.MatchError) extend `java.lang.Exception`??

geoffjentry: As I mentioned earier, these comments are just a "did you think through what happens if an exception gets thrown" and not some Rx. As long as it doesn't blow up cromwell i dont care so much ATM

kshakir:
Now with `excludeKey`™

mcovarr:
:+1: sweet

README.md
```diff
@@ -2011,6 +2024,142 @@ Content-Length: 8286
...
+
+Response:
+```
+HTTP/1.1 200 OK
+Server spray-can/1.3.3 is not blacklisted
+Server: spray-can/1.3.3
+Date: Thu, 01 Oct 2015 22:19:07 GMT
+Content-Type: application/json; charset=UTF-8
+Content-Length: 8286
+{
```
geoffjentry: Could the example in the README be condensed, perhaps scattered 2 ways or something like that?

PR #1018, opened 2016-06-17
Workflow logging. Closes #814

### LINKED ISSUE #814: Implement workflow logging

Horneth:
- [x] Clarify the options / configuration names


core/src/main/scala/cromwell/core/logging/LoggingTest.scala
```diff
@@ -0,0 +1,7 @@
+package cromwell.core.logging
+
+import java.util.regex.Pattern
+
+object LoggingTest {
+  def escapePattern(pattern: String) = Pattern.quote(pattern)
```
cjllanwarne: `LoggingTest`? Should this be in `src/test/...`

engine/src/main/scala/cromwell/engine/backend/jes/Pipeline.scala
```diff
@@ -22,7 +22,7 @@ object Pipeline {
     lazy val command = jesJobDescriptor.jesCommandLine
     lazy val runtimeAttributes = jobDescriptor.callRuntimeAttributes
 
-    val logger = WorkflowLogger(
+    val logger = OldStyleWorkflowLogger(
```
cjllanwarne: Oh... This might have disappeared now...

core/src/main/scala/cromwell/core/logging/LoggingTest.scala
Horneth: Yes but I couldn't import it properly from anywhere when it was in the test folder..

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -134,14 +135,12 @@ object WorkflowActor {
 /**
   * Class that orchestrates a single workflow.
   */
-class WorkflowActor(workflowId: WorkflowId,
+class WorkflowActor(val workflowId: WorkflowId,
```
cjllanwarne: Scala Learnings! What does this `val` do?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/CopyWorkflowLogsActor.scala
```diff
@@ -2,33 +2,44 @@ package cromwell.engine.workflow.lifecycle
...
   }
 
-  // TODO: PBE: https://github.com/broadinstitute/cromwell/issues/814
-  private def getTempLogFilePathOption: Option[Path] = ???
+  def copyAndClean(dest: Path) = {
+    workflowLogger.workflowLogPath foreach { src =>
+      val destPath = dest.toAbsolutePath.resolve(src.getFileName)
+      workflowLogger.info(s"Copying workflow logs from ${src.toAbsolutePath} to $destPath")
+
+      PathCopier.copy(src.toAbsolutePath, destPath)
```
cjllanwarne: Is this resilient to IOExceptions? We don't want an error in some off-to-the-side actor bringing down the actor system

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
geoffjentry: It's `private val` by default. This makes it `val`, basically the way a case class is

cjllanwarne:
Do the final log locations need to be metadata'd?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/CopyWorkflowLogsActor.scala
Horneth: Good point - now that you say that I realize it probably shouldn't be at the root of the actor system : / CopyServiceActor ? 😄

cjllanwarne:
Apart from a couple of OldStyleWorkflowLoggers which may or may not have been nuked... this looks awesome! 👍

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
Horneth: With `case class`es the constructor parameters automatically become attributes of the class, and hence override / implement potentially unimplemented variables in super classes/traits.
With normal classes this is not true, adding the val makes the workflowId an attribute of the class.

```scala
scala> class A(yo: String)
defined class A

scala> val a = new A("hey")
a: A = A@51016012

scala> a.yo
<console>:13: error: value yo is not a member of A
       a.yo
         ^

scala> class A(val yo: String)
defined class A

scala> val a = new A("hey")
a: A = A@7d9d1a19

scala> a.yo
res2: String = hey 
```

core/src/main/scala/cromwell/core/WorkflowOptions.scala
```diff
@@ -42,10 +42,26 @@ import scala.util.{Failure, Success, Try}
  */
 
 object WorkflowOptions {
+
+  sealed trait WorkflowOption { def name: String }
```
mcovarr: making this an abstract class that takes a constructor parameter `name` might be less noisy

```scala
case object WriteToCache extends WorkflowOption("write_to_cache")
```

engine/src/main/scala/cromwell/engine/workflow/lifecycle/CopyWorkflowLogsActor.scala
Horneth: @geoffjentry what do you think ? ^^

core/src/main/scala/cromwell/core/logging/LoggerWrapper.scala
```diff
@@ -0,0 +1,220 @@
...
+    */
+  private def varargsAkkaLog(pattern: String, arguments: Any*): Unit = {
+    lazy val formatted: String = format(pattern)
+
+    arguments.length match {
+      case 1 => akkaLogger.foreach(_.error(formatted, arguments.head))
+      case 2 => akkaLogger.foreach(_.error(formatted, arguments.head, arguments(1)))
+      case 3 => akkaLogger.foreach(_.error(formatted, arguments.head, arguments(1), arguments(2)))
+      case 4 => akkaLogger.foreach(_.error(formatted, arguments.head, arguments(1), arguments(2), arguments(3)))
+      case _ =>
```
mcovarr: is not logging preferable to doing at least 4 substitutions?

core/src/main/scala/cromwell/core/WorkflowOptions.scala
geoffjentry: There's an argument that `sealed abstract class` is better than `sealed trait` regardless. I've not yet decided if I agree w/ that and certainly don't care enough to comment in a PR but since there's already a suggestion for abstract class, there ya go

geoffjentry: Specifically: http://stackoverflow.com/questions/35249901/scala-client-composition-with-traits-vs-implementing-an-abstract-class/35251513#35251513

Horneth: Will change, I prefer abstract class in this case too

core/src/main/scala/cromwell/core/logging/LoggerWrapper.scala
```diff
@@ -0,0 +1,220 @@
...
+    // The :_* indicates the scala compiler to use the (String, Object...) method and not the (String, Object)
+    slf4jLoggers.foreach(_.warn(formatted, arguments:_*))
+  }
+
+  override def warn(pattern: String, arg1: Any, arg2: Any): Unit = {
+    lazy val formatted: String = format(pattern)
+
+    akkaLogger.foreach(_.warning(formatted, arg1, arg2))
+    // The :Any for arg2 indicates the scala compiler to use the (String, Object, Object) method and not the (String, Object...)
+    // See https://github.com/typesafehub/scalalogging/issues/16
```
mcovarr: wow this is a PITA, thanks for documenting it so well!  😄

core/src/main/scala/cromwell/core/logging/JobLogger.scala
```diff
@@ -0,0 +1,26 @@
...
+  def workflowId: WorkflowId
+  def jobTag: String
+
+  lazy val jobLogger: Logger = new JobLogger(self.path.name, workflowId, jobTag, Option(log))
+}
+
+/**
+  * Similar to WorkflowLogger, with the addition that the log tag includes the call tag.
+  */
+class JobLogger(name: String,
```
geoffjentry: is this instantiated (or intended to have the possibility) anywhere other than the trait above? If so, would it make sense to enforce that? making up possibilities that might not even work but e.g. sealing it, sticking it in the trait, etc

README.md
```diff
@@ -846,7 +846,7 @@ And this set of workflow options:
 
 ```json
 {
-  "defaultRuntimeOptions": {
+  "default_runtime_options": {
```
mcovarr: Glad this change is documented, we'll need to make sure to communicate it to our Green friends.

README.md
```diff
@@ -1057,11 +1057,11 @@ Valid keys and their meanings:
...
-    * **workflowFailureMode** - What happens after a task fails. Choose from:
+    * **final_workflow_log_dir** - Specifies a path where per-workflow logs will be written.  If this is not specified, per-workflow logs will not be copied out of the Cromwell workflow log temporary directory/path before they are deleted.
+    * **final_workflow_outputs_dir** - Specifies a path where final workflow outputs will be written.  If this is not specified, workflow outputs will not be copied out of the Cromwell workflow execution directory/path.
+    * **final_call_logs_dir** - Specifies a path where final call logs will be written.  If this is not specified, call logs will not be copied out of the Cromwell workflow execution directory/path.
+    * **default_runtime_options** - A JSON object where the keys are [runtime attributes](#runtime-attributes) and the values are defaults that will be used through the workflow invocation.  Individual tasks can choose to override these values.  See the [runtime attributes](#specifying-default-values) section for more information.
+    * **workflow_failure_mode** - What happens after a task fails. Choose from:
```
mcovarr: ...and all these other changes too

engine/src/main/scala/cromwell/engine/workflow/lifecycle/CopyWorkflowLogsActor.scala
geoffjentry: Discussed in person

core/src/main/scala/cromwell/core/logging/WorkflowLogger.scala
```diff
@@ -0,0 +1,101 @@
...
+
+  lazy val workflowLogger = new WorkflowLogger(self.path.name, workflowId, Option(log))
+}
+
+object WorkflowLogger {
+  def makeFileLogger(path: Path, level: Level): Logger = {
+    val ctx = LoggerFactory.getILoggerFactory.asInstanceOf[LoggerContext]
+    val name = path.getFileName.toString
+
+    Option(ctx.exists(name)) match {
```
mcovarr: is this stuff thread-safe?

core/src/main/scala/cromwell/core/logging/LoggerWrapper.scala
Horneth: Possibly yes, I mostly wanted to avoid cases where one would use one of those varargs methods and not see anything in the log and wonder wth is happening.

core/src/main/scala/cromwell/core/logging/JobLogger.scala
Horneth: I believe it's not instantiated anywhere else. The only use case I can see where you would want to create one manually is to log something from outside an actor where the `ActorLogging` trait wouldn't work. But we have no such case yet.

core/src/main/scala/cromwell/core/logging/WorkflowLogger.scala
Horneth: I copied this from the old world but I'll look into it yes it could be nasty if it wasn't...

Horneth: I think it's fine it's using a `ConcurrentHashMap` to cache the loggers.

mcovarr: It's good that there's a `ConcurrentHashMap` to keep the logger factory's internals from getting out of whack, but it still seems possible that multiple threads could create different `FileAppender`s for the same underlying log file and hand them back to their callers. (i.e. the get/set for the logger by name is not atomic). I can't remember ever seeing any issues like this in the Olde Worlde so maybe I'm just being paranoid.

mcovarr:
:+1:

core/src/main/scala/cromwell/core/logging/WorkflowLogger.scala
Horneth: The appender list is a CopyOnWriteArrayList which provides some thread-safety as well. I will drop a comment here in case we see something break as a hint for future debugging.

README.md
cjllanwarne: If we're taking the chance to change this, the choice of the word `options` has always annoyed me. Should probably be `default_runtime_attributes` instead.

(ps it was me who originally named this, mea culpa)

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -256,20 +255,30 @@ class WorkflowActor(workflowId: WorkflowId,
...
-        case unknownState => log.warning(s"$tag $unknownState is an unhandled terminal state!")
+        case unknownState => workflowLogger.warn(s"$unknownState is an unhandled terminal state!")
       }
+
+      // Copy/Delete workflow logs
+      if (WorkflowLogger.isEnabled) {
+        stateData.workflowDescriptor foreach { wd =>
+          // Create the actor in the system context because this WorkflowActor is going to die right after
+          val copyWorkflowLogsActor: ActorRef = context.system.actorOf(CopyWorkflowLogsActor.props(wd), s"CopyWorkflowLogsActor-$workflowId")
+          copyWorkflowLogsActor ! CopyWorkflowLogsActor.Start
```
kshakir: I'm not a wheeled reviewer, so feel free to ignore, but--

I am _not_ a fan of:

- Create an actor
- Send the actor a message
- Kill yourself

If the actor decides to restart, and there is no supervisor, then the actor will sit there idly waiting for someone, anyone, to tell it what to do... forever alone... 😢 

Instead, I prefer actors like the CopyWorkflowLogsActor know what their purpose in life is, via a value or message the constructor if necessary, and starting running in the `preStart()`.

Horneth: @kshakir That is true of all our actors though I believe, the "pattern" we use is
create an actor
send it a "Start" message

Even if this actor has a parent that can detect that it restarted, I don't think we have a strategy in place to send "Start" again.

Also per discussion with @geoffjentry the last commit modifies this to use an akka router so this time the copy actors will have an alive parent.

Horneth:
@cjllanwarne @mcovarr Could use another review (there's only 1 new commit).

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
geoffjentry: tl;dr feel free to do it how you want

FWIW I also am not a huge fan of what @kshakir is talking about, although more just because it seems superfluous vs his more reasoned approach. I don't really care enough to tell people to do it otherwise. 

I think that tactic started due to some of the timing dances we needed to do. If I'm not just making that statement up that seems like a bit of a smell that the timing dances are overly complicated

core/src/main/scala/cromwell/core/logging/JobLogger.scala
Horneth: FYI I actually stumbled upon 2 use cases (in the tests, and in Run.scala) where we need to use the constructor.

mcovarr:
Still :+1:

core/src/main/scala/cromwell/core/logging/LoggingTest.scala
kshakir: **TL;DR To use `LoggingTest` as a core test utility, add a dependency on `test->test` in build.sbt**

Yay more special maven/sbt syntax.

I'm assuming sbt/ivy is the same as maven, in that "[test … scope is not transitive](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Scope)". Only "compile [artifacts] are propagated to dependent projects".

In build.sbt, by default only main/`compile` code is added to child `compile` dependency trees. You need to explicitly tell the child artifact that it [depends on](http://www.scala-sbt.org/0.13/docs/Multi-Project.html#Per-configuration+classpath+dependencies) core's test code.

```scala
lazy val someArtifact = (project in file("someRoot"))
  .dependsOn(otherArtifact)
  .dependsOn(core % "test->test") // I don't need core's main code -or- I got core's main from otherArtifact transitively
```

or, if you also need the code from core's main folder in your child artifact's main code,

```scala
lazy val someArtifact = (project in file("someFolder"))
  .dependsOn(core % "compile->compile;test->test")  // "I want it all…" ♫
```

Looking through build.sbt, I believe what's happened is that we've correctly setup the dependency tree for our transitive main/`compile` code, but do _not_ have enough explicit setup for the non-transitive code living in test folders.

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
Horneth: I would love to see more "let it crash" behaviours with proper supervision strategy etc... It could actually save us up some work and more importantly a lot of messages going back and forth. I honestly felt like this being the "last" ticket before 0.20, spending time doing a perfect let it crash model for this particular feature wasn't worth doing now. But I'm happy to create a ticket for it or even make a subsequent PR later.

core/src/main/scala/cromwell/core/logging/LoggingTest.scala
Horneth: Thanks for the detailed explanation ! Looks like backends have the dependency on core test transitively through the "backend" project. So it actually works now.

PR #1023, opened 2016-06-21
Remove MetadataAck from dead letter logging

Horneth:
Until we decide what to do with those Acks, the flood of dead letters is rather polluting and useless at the moment.


cjllanwarne:
👍

mcovarr:
:+1: sweet!

PR #1044, opened 2016-06-21
Rewire batch submission.  Closes #1017

### LINKED ISSUE #1017: Make Batch Submit API great again

mcovarr:
The README and Swagger were already great, so no changes there.  The /batch name of the endpoint is maybe not so great; is this perhaps how this was overlooked?  The code was so close to greatness already.

### DIFF
--- a/engine/src/main/scala/cromwell/webservice/WorkflowSubmitBatchActor.scala
+++ b/engine/src/main/scala/cromwell/webservice/WorkflowSubmitBatchActor.scala
@@ -2,6 +2,7 @@ package cromwell.webservice
 
 import akka.actor.{Actor, ActorRef, Props}
 import cromwell.engine.WorkflowSourceFiles
+import cromwell.engine.workflow.WorkflowManagerActor
 import cromwell.webservice.CromwellApiHandler.{WorkflowManagerBatchSubmitResponse, WorkflowManagerResponse}
 
 /** Submits a sequence of sources, then messages back the respective sequence of responses. */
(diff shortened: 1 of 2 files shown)


cjllanwarne:
Somebody Else's Code: I would prefer an immutable var instead of a mutable val

geoffjentry:
@cjllanwarne There are tradeoffs on that conunudrum. Can you explain your preference?

engine/src/main/scala/cromwell/webservice/WorkflowSubmitBatchActor.scala
```diff
@@ -34,7 +35,9 @@ case class WorkflowManagerBatchIndexedResponse(response: WorkflowManagerResponse
 class WorkflowSubmitIndexedActor(batchSubmitActor: ActorRef, requestHandlerActor: ActorRef,
```
cjllanwarne: At the risk of jumping onto the Thibault bandwagon... `Router`?

OTOH... nah :)

mcovarr: IMHO that's a fine suggestion for an enhancement ticket, but beyond the scope of this "make it great again" PR.  😄

cjllanwarne:
@geoffjentry I want to resort to authority and say "Zen of Akka"... 

Reasons for my gut feeling: A mutable val makes it look and act more like a state machine, and reduces the risk of accidentally leaking the variable pointer to other threads which may update it out of band. Obviously not likely in this case, but as a muscle memory thing a-la `Some(constant)`

cjllanwarne:
👍

Horneth:
👍

Issue #1047, opened 2016-06-22
Add tests in ServiceRegistryActorSpec


Horneth:
The service registry actor internals (message flow, service lookup by name...) should be tested. An empty `ServiceRegistryActorSpec` file already exists for that purpose.

mcovarr:
It probably doesn't matter but I'm not finding a `ServiceRegistryActorSpec`.

PR #1048, opened 2016-06-22
Remove KV test in ServiceRegistryActorSpec Closes #1039

### LINKED ISSUE #1039: Assess/Fix/Implement Ignored Tests in ServiceRegistryActorSpec

Horneth:
This test was actually testing the KV service (which already has a test spec) through the ServiceRegistry.
The service registry should have its own tests to assert that the message flow / service lookup is working, but not to test all possible services it knows about. See https://github.com/broadinstitute/cromwell/issues/1047


cjllanwarne:
Are there any useful tests of the ServiceRegistry itself? Otherwise, might as well just delete the whole file

cjllanwarne:
But, I agree with the concept of not testing KVP service in `ServiceRegistrySpec` 👍

Horneth:
There aren't but there could be. But until then we might as well remove the file yes.

mcovarr:
top-flight removing :+1:

PR #1049, opened 2016-06-22
Unignore WorkflowOutputsSpec Closes #1033

### LINKED ISSUE #1033: Assess/Fix/Implement Ignored Tests in WorkflowOutputsSpec

Horneth:


mcovarr:
:+1: assuming you have better luck with Travis than I do

engine/src/test/scala/cromwell/WorkflowOutputsSpec.scala
```diff
@@ -63,7 +63,7 @@ class WorkflowOutputsSpec extends CromwellTestkitSpec {
     "Allow explicitly named inputs in the output section" ignore {
```
Horneth: This test needs to be discussed. It verifies that one can declare a task input as a workflow output, which is not possible in 0.20, because we don't store task inputs anymore (we didn't need to until now).
@kcibul Question: is this a required feature ? If yes it deserves its own ticket IMO.

geoffjentry: Do we know the provenance of this feature in the first place? Was it actually something that anyone used?

geoffjentry:
👍  modulo that open question you had

engine/src/test/scala/cromwell/WorkflowOutputsSpec.scala
Horneth: That's what I'm asking yes, I don't know if anyone actually uses / wants this.

cjllanwarne: Pretty sure it's there because it's in the WDL spec. I'd wager fairly confidently that nobody's using it currently

mcovarr: I would ask the Greens

PR #1050, opened 2016-06-22
Run a round of Centaur during travis. It's cromwell's all the way dow…

### LINKED ISSUE #1022: Centaur in travis

geoffjentry:
…n. Closes #1022 

Known sharp corners I'm inclined to fix later in the interest of getting this rolling:

- The docker image shouldn't be pointing at my own organization
- 32 CPU seems like overkill. OTOH I've seen futures time out at 16 cpu. One could argue (and I would) that this is BS, however for this testing I want to make sure futures never time out. We can/should address the "why" separately
- It's painful to dig into why a test failed on the centaur side. I don't know yet a way to make that easier, but this at least makes it possible


src/main/resources/travis/travis.wdl
```diff
@@ -0,0 +1,31 @@
...
+    File conf
+    File pem
+    File cromwell_jar
+
+    command<<<
+        mkdir -p /cromwell_root/tmp/ivy2
+        export SBT_OPTS=-Dsbt.ivy.home=/cromwell_root/tmp/.ivy2
+        git clone https://github.com/broadinstitute/centaur.git
+        cd centaur
+        git checkout jg_allow_cromwell_jar
```
geoffjentry: This needs to be changed, but waiting on other slackasses to review the centaur PR (just officially assigned @cjllanwarne and @Horneth to that)

src/main/resources/travis/travisCentaur.sh
```diff
@@ -0,0 +1,42 @@
...
+curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key add -
+sudo apt-get install google-cloud-sdk
+export PYTHONPATH="/usr/lib/python2.7/site-packages:/usr/local/lib/python2.7/site-packages:/usr/local/lib/python2.7/dist-packages"
+export CONFIGURE_OPTS="--enable-unicode=ucs4"
+pyenv install 2.7.10
+pyenv global 2.7.10
+sudo pip install --upgrade pip
+sudo pip install pyopenssl ndg-httpsclient pyasn1 --upgrade
+export CLOUDSDK_PYTHON_SITEPACKAGES=1
+gcloud auth activate-service-account --key-file=broad-dsde-cromwell-dev-d9c443bb4a94.json cromwell@broad-dsde-cromwell-dev.iam.gserviceaccount.com
```
geoffjentry: If people are leery about our email being public I can embed that in the credentials file as well

src/main/resources/travis/travis.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
kshakir: Build binaries go under `src/bin`, not `src/main/resources`. The contents of the latter folder end up as embedded resources in the artifact jars.

geoffjentry: You really just mean executables, right? I mean this in a non-cheeky way but e.g. jesConf.tar.enc *is* a binary file but doesn't need to be in an artifact

kshakir: Whoops yeah. :) _Executables_ go under src/bin.

Speaking of binary files that shouldn't ~~bloat~~ go in the artifact, I'm not really sure where jesConf.tar.enc should go.

Can't find the discussion on the topic at the moment, but these two links show "the" conventional directory structure, plus how maven itself organizes under the src directory (including src/bin):

- https://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html
- https://github.com/apache/maven/tree/maven-3.3.9/apache-maven/src

geoffjentry: In theory the enc file could go in private centaur, depends on how much ppl care right now

.travis.yml
```diff
@@ -2,18 +2,11 @@ sudo: required
...
-before_script:
-  - mysql -u root -e "SET GLOBAL sql_mode = 'STRICT_ALL_TABLES';" 
-  - mysql -u root -e "CREATE DATABASE IF NOT EXISTS cromwell_test;"
-  - mysql -u root -e "CREATE USER 'travis'@'localhost' IDENTIFIED BY '';"
-  - mysql -u root -e "GRANT ALL PRIVILEGES ON cromwell_test . * TO 'travis'@'localhost';"
-script: sbt -Dbackend.providers.Local.config.filesystems.local.localization.0=copy clean coverage nointegration:test coverageReport && sbt coverageAggregate && sbt assembly
-after_success: sbt coveralls
+env:
+  - BUILD_TYPE=test
+  - BUILD_TYPE=centaur
```
mcovarr: setting the same variable twice sets up parallel builds?

geoffjentry: images

src/main/resources/travis/travisCentaur.sh
mcovarr: seems like a good idea or we might end up on the same lists as rawls

src/main/resources/travis/travis.wdl
```diff
@@ -0,0 +1,31 @@
+task centaur {
```
mcovarr: `centaur.wdl` seems like a better name for this

mcovarr: ditto the inputs file above

.travis.yml
mcovarr: Indeed.  Perhaps add a comment as to the transcendental nature of what's happening here.

mcovarr:
💯  :+1:

src/main/resources/travis/travisCentaur.sh
geoffjentry: Strictly speaking I don't think the email is valuable for anything but it's easy to remove. I realized the right thing to do would be to add it in as an env var

geoffjentry:
@kshakir the 'where should stuff go' convo got clipped. Looking at your links, one could make an argument that the wdl should go in `src/it` (which still doesn't answer the question of the encrypted travis file). I don't really care where they go, so if you do just tell me where to stick 'em otherwise i'll leave them in src/main/resources

kshakir:
Move the directory in `src/main/resources/travis` to `src/bin/travis/resources`. Then 👍

PR #1053, opened 2016-06-22
SharedFileSystemSpec Closes #1027

### LINKED ISSUE #1027: Assess/Fix/Implement Ignored Tests in SharedFileSystemBackendSpec

Horneth:


backend/src/main/scala/cromwell/backend/BackendJobExecutionActor.scala
```diff
@@ -76,5 +77,7 @@ trait BackendJobExecutionActor extends BackendJobLifecycleActor with ActorLoggin
       outputMap + jobOutput
 
     })
+
+    TryUtil.sequenceMap(evaluatedOutputs, s"Workflow $jobDescriptor.descriptor.id post processing failed")
```
geoffjentry: disappointed-dog-face-04

Horneth: Is `TryUtil.sequenceMap` the reason why I'm looking at a puppy now ?

backend/src/main/scala/cromwell/backend/io/SharedFileSystem.scala
```diff
@@ -159,26 +154,25 @@ trait SharedFileSystem extends PathFactory { this: BackendJobExecutionActor =>
...
+      val tryAdjust = inputArray map localizeWdlValue(toDestPath, strategies)
 
       TryUtil.sequence(tryAdjust, s"Failed to localize files in input Array ${wdlValue.valueString}") map { adjusted =>
         new WdlArray(t, adjusted)
       }
     }
 
     def adjustMap(t: WdlMapType, inputMap: Map[WdlValue, WdlValue]): Try[WdlMap] = {
-      val tryAdjust = inputMap mapValues { localizeWdlValue(toDestPath, strategies, postProcessor) }
+      val tryAdjust = inputMap mapValues { localizeWdlValue(toDestPath, strategies) }
```
geoffjentry: Something to be careful of here, although I think I'm just being ridiculously paranoid (and it's test code anyways and unlikely to matter, although we *do* have weirdness in our test code!)

mapValues returns a View and not a new Map, which isn't a big deal here but I'm not sure what happens when you pass it into the WdlMap. Does Scala convert it to a true Map then? If not that could lead to pain.

geoffjentry:
👍

backend/src/main/scala/cromwell/backend/io/SharedFileSystem.scala
Horneth: Fair point, although we use it so many times throughout Cromwell that if it really was a problem we very likely would have seen it already IMO

kshakir:
👍

PR #1054, opened 2016-06-22
Replace ANSI codes with a handy dandy library

geoffjentry:


engine/src/main/scala/cromwell/logging/TerminalLayout.scala
```diff
@@ -22,25 +24,23 @@ object TerminalLayout {
   implicit class ColorString(msg: String) {
     def colorizeUuids: String = {
       "UUID\\((.*?)\\)".r.findAllMatchIn(msg).foldLeft(msg) {
-        case (l, r) =>
-          val color = if (Option(System.getProperty("RAINBOW_UUID")).isDefined)
-            Math.abs(17 * r.group(1).substring(0,8).map(_.toInt).product) % 209 + 22
-          else 2
```
mcovarr: what happened with `RAINBOW_UUID`?

geoffjentry: It's an undocumented feature not used anywhere which relies on multiple magic numbers

kshakir: I liked it when it used to be there, but we shut off colors in server logging. Maybe useful in tests sometime in the future?

Basically, it generated a different color for each UUID. When viewing a crap ton of workflows going by, the different UUIDs would be clearly highlighted with different colors.

But yes, the code is vestigial at the moment. RIP

geoffjentry: Oh I was just too lazy to close this. 

The background was the other day I stumbled across this and said "WTF do these numbers mean" and around the same time saw Li put out that fansi library and it seemed like a good 5 min spare time project. It turns out that a lot of the way we're highlighting doesn't fit iwth all of the sugar fansi gives you anyways. In this particular case i'm not inclined to figure out how to shoehorn RAINBOW_UUID into fansi.

I had no idea what RAINBOW_UUID was until miguel told me it was something @scottfrazer used. It wasn't documented anywhere and I even googled the term thinking perhaps it was a standard thing.

geoffjentry: also now that you mention it, I'm pretty sure that was the thing which gave me headaches looking at our color-ified logs :)

PR #1057, opened 2016-06-23
Restore MainSpec.  Closes #1041

### LINKED ISSUE #1041: Assess/Fix/Implement Ignored Tests in MainSpec

mcovarr:


mcovarr:
It was taking 12-14 seconds between the time the GetStatus message was received by the ServiceRegistryActor and the time it was received by the EngineMetadataServiceActor.  All sorts of other test stuff was executing in the interim, but the timeout for status queries previously defaulted to 10 seconds so that wasn't going to work.  Added an explicit 30 second dilated timeout as in many other spots.

Horneth:
Should we also look into a way not to have the same ServiceRegistryActor for all tests ?

mcovarr:
I'm not actually sure we do have the same ServiceRegistryActor for all tests, so we'd have to start looking into that first.  😄

kshakir:
👍

Horneth:
👍

PR #1060, opened 2016-06-23
RetryableCallSpec Closes #1037

### LINKED ISSUE #1037: Assess/Fix/Implement Ignored Tests in RetryableCallsSpec

Horneth:
There was already a test for this in WorkflowExecutionActor. Just added some more metadata verification.
This also attempts to provide a different ServiceRegistryActor for every WorkflowManagerActor being build in the tests (which might help travis slowness ? @mcovarr).


mcovarr:
Hmm 

concurrent tests =>
concurrent ServiceRegistryActors =>
concurrent EngineMetadataServiceActors =>
concurrent MetadataSummaryRefreshActors

The system was meant to have only a single MetadataSummaryRefreshActor running, so this might introduce weirdness.

mcovarr:
To be clear I'm not 💩 💩 ing the multiple ServiceRegistryActor idea, just pointing out one potential gotcha I can see.  I have reservations about my hacky "crank up the timeout" "solution" in the #1057 PR and would definitely like to look at other options like this.

Horneth:
Hmm ok, I was just curious to see what it would do honestly more than advocating for it, but I didn't think about the MetadataSummaryRefreshActor. I still think it's weird to build different WMA for every test but have them all pointing to the same ServiceActor.

mcovarr:
:+1:

engine/src/test/scala/cromwell/CromwellTestkitSpec.scala
```diff
@@ -442,33 +447,28 @@ abstract class CromwellTestkitSpec extends TestKit(new CromwellTestkitSpec.TestW
     }
   }
 
-  def getWorkflowMetadata(workflowId: WorkflowId, key: Option[String] = None)(implicit ec: ExecutionContext): JsObject = {
+  def getWorkflowMetadata(workflowId: WorkflowId, key: Option[String] = None, serviceRegistryActor: ActorRef)(implicit ec: ExecutionContext): JsObject = {
```
cjllanwarne: I thought you couldn't have a default in the middle of a parameter chain? Is this why you've needed to make the None explicit below?

engine/src/test/scala/cromwell/CromwellTestkitSpec.scala
```diff
@@ -442,33 +447,28 @@ abstract class CromwellTestkitSpec extends TestKit(new CromwellTestkitSpec.TestW
...
-      }
-    })
-
     val message = GetMetadataQueryAction(MetadataQuery(workflowId, None, key, None, None))
-    Await.result(supervisor.ask(message).mapTo[RequestComplete[(String,JsObject)]], Duration.Inf).response._2
+    val parentProbe = TestProbe()
+
+    TestActorRef(MetadataBuilderActor.props(serviceRegistryActor), parentProbe.ref, s"MetadataActor-${UUID.randomUUID()}") ! message
+    val metadata = parentProbe.expectMsgPF(TimeoutDuration) {
+      case response: RequestComplete[(StatusCode, JsObject)] @unchecked => response.response._2
```
cjllanwarne: Just for the benefit of understanding this easily in the future , could you add a quick comment about the `@unchecked`?

cjllanwarne:
👍 

[Approved with PullApprove](https://pullapprove.com/broadinstitute/cromwell/pull-request/1060/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=broadinstitute/cromwell)

PR #1061, opened 2016-06-23
WorkflowDescriptorSpec Closes #1030

### LINKED ISSUE #1030: Assess/Fix/Implement Ignored Tests in WorkflowDescriptorSpec

Horneth:
Removed `WorkflowDescriptorSpec`. Most tests were about copying logs and outputs which lives in a different (better) place now.
The one test regarding caching config that might be useful later was moved to the backend project and left ignored, with some pre-work to prepare it for whenever it will be prioritized.


scottfrazer:
👍

kshakir:
👍

PR #1069, opened 2016-06-24
SimpleWorkflowActorSpec.  Closes #1032

### LINKED ISSUE #1032: Assess/Fix/Implement Ignored Tests in SimpleWorkflowActorSpec

mcovarr:


engine/src/test/scala/cromwell/SimpleWorkflowActorSpec.scala
```diff
@@ -1,80 +1,144 @@
...
+
+  private def buildWorkflowActor(sampleWdl: SampleWdl, rawInputsOverride: String,
+                                 workflowId: WorkflowId = WorkflowId.randomId(),
+                                 matcher: Option[FailureMatcher] = None): WorkflowActorandMetadataPromise = {
+    val workflowSources = WorkflowSourceFiles(sampleWdl.wdlSource(), rawInputsOverride, "{}")
+    val promise = Promise[Unit]()
+    val watchActor = system.actorOf(MetadataWatchActor.props(matcher, promise), s"service-registry-$workflowId-${UUID.randomUUID()}")
+    val workflowActor = TestFSMRef(new WorkflowActor(workflowId, StartNewWorkflow, workflowSources, ConfigFactory.load(),
+      watchActor,
+      system.actorOf(Props.empty, s"workflow-copy-log-router-$workflowId-${UUID.randomUUID()}")),
```
kshakir: ToL: I never know in scala if one should use `UUID.randomUUID()` or `UUID.randomUUID`.

kshakir:
👍

engine/src/test/scala/cromwell/SimpleWorkflowActorSpec.scala
geoffjentry: id have assumed the latter. Why would it be the former? Guessing that there's actually some internal state being manipulated?

mcovarr: well there's gotta be a PRNG underpinning that, so yeah...

I just took the autocompletion IntelliJ offered, I can change this if you think it shouldn't have parens.

geoffjentry: at that point you get into "what's a side effect", technically anything that throws should then have a `()`, or if one wanted to be super pedantic anything which used memory.

I don't care either way and only chimed in because @kshakir commented

cjllanwarne: It's not about side effects its about referential integrity. You can't replace separate occurrences of randomUuid with a single assignment and get the same result. So, leave in the `()`

geoffjentry: Yeah, that was the conclusion I came to last night.

Also I feel extra sad admitting that I spent time pondering this last night

engine/src/test/scala/cromwell/SimpleWorkflowActorSpec.scala
```diff
@@ -1,80 +1,144 @@
...
-//    val descriptor = materializeWorkflowDescriptorFromSources(workflowSources = workflowSources)
-//    TestFSMRef(new OldStyleWorkflowActor(descriptor))
-//  }
+  case class MetadataWatchActor(matcher: Option[FailureMatcher], promise: Promise[Unit]) extends Actor {
+    override def receive = matcher match {
+      case None =>
+        promise.trySuccess(())
+        Actor.emptyBehavior
+      case Some(m) => {
+        case PutMetadataAction(events) if m.matches(events) =>
```
cjllanwarne: Is it safe to assume that `m.matches(events)` is always true? Otherwise, is another `case` req'd?

engine/src/test/scala/cromwell/SimpleWorkflowActorSpec.scala
```diff
@@ -1,80 +1,144 @@
...
+  }
+
+  case class WorkflowActorandMetadataPromise(workflowActor: ActorRef, promise: Promise[Unit])
+}
+
+class SimpleWorkflowActorSpec extends CromwellTestkitSpec {
+
+  private def buildWorkflowActor(sampleWdl: SampleWdl, rawInputsOverride: String,
+                                 workflowId: WorkflowId = WorkflowId.randomId(),
+                                 matcher: Option[FailureMatcher] = None): WorkflowActorandMetadataPromise = {
```
cjllanwarne: typo: capitalise `And`

mcovarr:
@geoffjentry this PR could be refreshed with the thumb of patriots

engine/src/test/scala/cromwell/SimpleWorkflowActorSpec.scala
```diff
@@ -1,80 +1,144 @@
...
+        }
+        probe.expectTerminated(workflowActor, 10.seconds.dilated)
+      }
     }
 
-    "fail to construct with missing inputs" ignore {
-//      intercept[IllegalArgumentException] {
-//        buildWorkflowFSMRef(SampleWdl.HelloWorld, rawInputsOverride = "{}")
-//      }
+    "fail to construct with missing inputs" in {
```
cjllanwarne: This is already a `MaterializeWorkflowDescriptorActorSpec` test

engine/src/test/scala/cromwell/SimpleWorkflowActorSpec.scala
```diff
@@ -1,80 +1,144 @@
...
+        }
+      }
+      probe.expectTerminated(workflowActor, 10.seconds.dilated)
     }
 
-    "fail to construct with inputs of the wrong type" ignore {
-//      intercept[IllegalArgumentException] {
-//        buildWorkflowFSMRef(SampleWdl.HelloWorld, rawInputsOverride = s""" { "$Addressee" : 3} """)
-//      }
+    "fail to construct with inputs of the wrong type" in {
```
cjllanwarne: I'm not sure whether it is already... but this should be a `MaterializeWorkflowDescriptorActorSpec` test

engine/src/test/scala/cromwell/SimpleWorkflowActorSpec.scala
```diff
@@ -1,80 +1,144 @@
...
     }
 
-    "gracefully handle malformed WDL" ignore {
-//      within(TestExecutionTimeout) {
-//        val fsm = buildWorkflowFSMRef(SampleWdl.CoercionNotDefined, SampleWdl.CoercionNotDefined.wdlJson)
-//        waitForPattern("transitioning from Running to Failed") {
-//          fsm ! Start()
-//        }
-//      }
+    "gracefully handle malformed WDL" in {
```
cjllanwarne: Already a `MaterializeWorkflowDescriptorActorSpec` test

engine/src/test/scala/cromwell/SimpleWorkflowActorSpec.scala
mcovarr: If the matcher doesn't match (which is likely) this code doesn't care.  Do you think there should be an explicit case for that?

engine/src/test/scala/cromwell/SimpleWorkflowActorSpec.scala
mcovarr: ok

engine/src/test/scala/cromwell/SimpleWorkflowActorSpec.scala
mcovarr: I think you mean referential transparency and yeah that's right.  () FTW

engine/src/test/scala/cromwell/SimpleWorkflowActorSpec.scala
cjllanwarne: TBH I'm not completely clear on what this whole construct is trying to achieve... but if that's a case that safely ignored then LGTM

engine/src/test/scala/cromwell/SimpleWorkflowActorSpec.scala
cjllanwarne: Haha yes... But how infeasible is it to claim that I did indeed mean integrity after all?

```
Referential integrity is a property of data which, when satisfied, requires every value of one attribute (column) of a relation (table) to exist as a value of another attribute (column) in a different (or the same) relation (table).
```

engine/src/test/scala/cromwell/SimpleWorkflowActorSpec.scala
mcovarr: Ah okay, I should add comments then.  This is making sure that the right metadata messages have been emitted by the WorkflowActor.  Normally there's a WorkflowManagerActor involved which spins up a real service registry, but this is testing at a lower level and I was only interested in seeing that the right messages make it to the service registry.

cjllanwarne:
I'll leave the existential dilemmas of some of the tests up to you. Otherwise 👍

engine/src/test/scala/cromwell/SimpleWorkflowActorSpec.scala
geoffjentry: I just figured there was a class `Actorand`

engine/src/test/scala/cromwell/SimpleWorkflowActorSpec.scala
geoffjentry: Glad it took 5 days to figure out if we should put parens after a function

geoffjentry:
I got yer patriot right here 👍  🇺🇸

PR #1077, opened 2016-06-27
WorkflowManagerActorSpec fixup. Closes #1029

### LINKED ISSUE #1029: Assess/Fix/Implement Ignored Tests in MaterializeWorkflowDescriptorActorSpec

cjllanwarne:
Ok, ok. This is less "fixup" and more "start again and try to preserve the spirit of the test cases"...

### DIFF
--- a/core/src/main/scala/cromwell/core/WorkflowOptions.scala
+++ b/core/src/main/scala/cromwell/core/WorkflowOptions.scala
@@ -56,7 +56,6 @@ object WorkflowOptions {
   // Misc.
   case object DefaultRuntimeOptions extends WorkflowOption("default_runtime_attributes")
   case object WorkflowFailureMode extends WorkflowOption("workflow_failure_mode")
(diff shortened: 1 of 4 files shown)


geoffjentry:
👍

core/src/main/scala/cromwell/core/WorkflowOptions.scala
```diff
@@ -56,7 +56,6 @@ object WorkflowOptions {
   // Misc.
   case object DefaultRuntimeOptions extends WorkflowOption("default_runtime_attributes")
   case object WorkflowFailureMode extends WorkflowOption("workflow_failure_mode")
-  case object Backend extends WorkflowOption("backend")
```
ruchim: Just a heads up, this will break a Centaur test. It will probably have to be re-enabled in the future when we can assign a backend through runtime attributes instead of workflow options.

ruchim:
👍 all good!

PR #1080, opened 2016-06-28
Fixup of SlickDataAccessSpec. Fixes #1040

### LINKED ISSUE #1040: Assess/Fix/Implement Ignored Tests in SlickDataAccessSpec

cjllanwarne:
Alternative title: kicking the SDAS can down the road until 0.21 (when these methods will make sense)

### DIFF
--- a/engine/src/test/scala/cromwell/engine/db/slick/SlickDataAccessSpec.scala
+++ b/engine/src/test/scala/cromwell/engine/db/slick/SlickDataAccessSpec.scala
@@ -7,7 +7,7 @@ import java.time.OffsetDateTime
 import better.files._
 import com.google.api.client.util.ExponentialBackOff
 import com.typesafe.config.ConfigFactory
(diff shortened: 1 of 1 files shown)


engine/src/test/scala/cromwell/engine/db/slick/SlickDataAccessSpec.scala
```diff
@@ -162,7 +162,8 @@ class SlickDataAccessSpec extends FlatSpec with Matchers with ScalaFutures with
       } yield ()).futureValue
     }
 
-    it should "create and retrieve the workflow for just reading" taggedAs DbmsTest ignore {
+    // TODO: PBE: May not be relevant in the new database
```
mcovarr: ooh foreshadowing, I like it

mcovarr:
:+1:

kshakir:
👍

PR #1084, opened 2016-06-28
Handle ActorInitializationException in supervisors Closes #1079

### LINKED ISSUE #1079: Workflow gets stuck in Initialization if BackendWorkflowInitializationActor crashe during instantiation
If `BackendWorkflowInitializationActor` fails to instantiate, it never receives the `Initialize` message and the WorkflowActor never gets a response. So the workflow hangs forever.

From a more general perspective, if any actor in the lifecycle craches/dies unexpectedly, the workfow will hang at whichever state it was in..

Horneth:


supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesFinalizationActor.scala
```diff
@@ -39,9 +38,7 @@ class JesFinalizationActor (override val workflowDescriptor: BackendWorkflowDesc
 
   private def deleteAuthenticationFile(): Future[Unit] = {
     if (jesConfiguration.needAuthFileUpload) {
-      val delete = () => Future(workflowPaths.gcsAuthFilePath.delete(false))
-
-      Retry.withRetry(delete, isFatal = isFatalJesException, isTransient = isTransientJesException)(context.system) map { _ => () }
+      Future(workflowPaths.gcsAuthFilePath.delete(false)) map { _ => () }
```
cjllanwarne: No more retry?

cjllanwarne:
👍 

[Approved with PullApprove](https://pullapprove.com/broadinstitute/cromwell/pull-request/1084/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=broadinstitute/cromwell)

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesFinalizationActor.scala
Horneth: there's already a retry 3 times in the gcs filesystem

mcovarr:
This seems to fix the specific issue that was causing the workflow to get stuck in #1079, but does it address the more general problem of the workflow getting stuck if the initialization/finalization actors fail to instantiate for some other reason?

Horneth:
@mcovarr correct, the more general problem exists throughout cromwell though and I didn't want to get into a rabbit hole with this but after digging a bit more it's actually not that bad. I'll update the PR with some better crash handling instead

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowLifecycleActor.scala
```diff
@@ -65,6 +66,12 @@ trait WorkflowLifecycleActor[S <: WorkflowLifecycleActorState] extends LoggingFS
   def successResponse: WorkflowLifecycleSuccessResponse
   def failureResponse(reasons: Seq[Throwable]): WorkflowLifecycleFailureResponse
 
+  override def supervisorStrategy = OneForOneStrategy() {
+    case ex: ActorInitializationException =>
+      context.parent ! failureResponse(Seq(ex))
+      Stop
```
cjllanwarne: This is a little confusing: 

I think we also want to stop this actor, and any other supervisees other than the one which failed?

Or will that happen naturally when the parent is stopped?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowLifecycleActor.scala
```diff
@@ -65,6 +66,12 @@ trait WorkflowLifecycleActor[S <: WorkflowLifecycleActorState] extends LoggingFS
...
+    case t => super.supervisorStrategy.decider.applyOrElse(t, (_: Any) => Escalate)
```
cjllanwarne: What does this get escalated to? I can't see a supervisor strategy in WorkflowActor?

Horneth: Every actor has a default supervisor strategy http://doc.akka.io/docs/akka/2.4.7/scala/fault-tolerance.html#Default_Supervisor_Strategy

mcovarr:
:+1: assuming you can convince Travis

cjllanwarne:
re-👍, pending test passes

cjllanwarne:
Going to merge. Centaur has been updated to address the test failures.

PR #1088, opened 2016-06-28
Set the JES polling backoff to be the same as 0.19

geoffjentry:
We had previously capped the backoff at 10m and now at 60s. This doesn't resolve the QPS storm but does seem (he says unscientifically) to make it a little better. At the very least this wasn't helping.


mcovarr:
:+1:

Horneth:
👍

Issue #988, opened 2016-06-13
Metadata endpoint should return 404 for unknown workflow ID


mcovarr:
Currently returns 200 with empty metadata

PR #1092, opened 2016-06-29
Address ignored CromwellApiServiceSpec test closes #1024

### LINKED ISSUE #1024: Assess/Fix/Implement Ignored Tests in CromwellApiServiceSpec

kcibul:


engine/src/main/scala/cromwell/engine/db/DataAccess.scala
```diff
@@ -148,7 +148,7 @@ trait DataAccess extends AutoCloseable {
   def queryWorkflowOutputs(id: WorkflowId)
                           (implicit ec: ExecutionContext): Future[Seq[MetadataEvent]] = {
     val uuid = id.id.toString
-    queryMetadataEventsWithWildcardKeys(uuid, s"${WorkflowMetadataKeys.Outputs}:".wrapNel, requireEmptyJobKey = true).
+    queryMetadataEventsWithWildcardKeys(uuid, s"${WorkflowMetadataKeys.Outputs}:%".wrapNel, requireEmptyJobKey = true).
```
mcovarr: hmm how was this working before?

engine/src/main/scala/cromwell/webservice/CromwellApiHandler.scala
```diff
@@ -33,8 +33,8 @@ object CromwellApiHandler {
   final case class ApiHandlerCallStdoutStderr(id: WorkflowId, callFqn: String) extends ApiHandlerMessage
   final case class ApiHandlerWorkflowStdoutStderr(id: WorkflowId) extends ApiHandlerMessage
   final case class ApiHandlerCallCaching(id: WorkflowId, parameters: QueryParameters, callName: Option[String]) extends ApiHandlerMessage
-  final case class ApiHandlerWorkflowMetadata(id: WorkflowId,
-                                              parameters: WorkflowMetadataQueryParameters) extends ApiHandlerMessage
+//  final case class ApiHandlerWorkflowMetadata(id: WorkflowId,
+//                                              parameters: WorkflowMetadataQueryParameters) extends ApiHandlerMessage
```
mcovarr: can this be deleted?

engine/src/main/scala/cromwell/webservice/CromwellApiHandler.scala
```diff
@@ -89,8 +89,8 @@ class CromwellApiHandler(requestHandlerActor: ActorRef) extends Actor with Workf
       context.parent ! RequestComplete(StatusCodes.OK, WorkflowAbortResponse(id.toString, WorkflowAborted.toString))
     case WorkflowManagerAbortFailure(_, e) =>
       error(e) {
-        //case _: WorkflowNotFoundException => RequestComplete(StatusCodes.NotFound, APIResponse.error(e))
         case _: IllegalStateException => RequestComplete(StatusCodes.Forbidden, APIResponse.error(e))
+        case _: Exception if e.getMessage.endsWith("because no workflow with that ID is in progress") => RequestComplete(StatusCodes.NotFound, APIResponse.error(e))
```
mcovarr: Any reason the abort code couldn't generate a `WorkflowNotFoundException` and this code could stay the same?

engine/src/main/scala/cromwell/services/EngineMetadataServiceActor.scala
```diff
@@ -86,15 +86,26 @@ case class EngineMetadataServiceActor(serviceConfig: Config, globalConfig: Confi
...
     val replyTo = sender()
     dataAccess.queryLogs(id) onComplete {
-      case Success(l) => replyTo ! LogsResponse(id, l)
+      case Success(l) =>
+        if (l.isEmpty) {
+          self ! HandleNotFound(id, replyTo)
+        } else {
+          replyTo ! LogsResponse(id, l)
+        }
+
```
mcovarr: these would probably be better as separate cases with guards

mcovarr: i.e.

```scala
case Success(l) if l.isEmpty => ...
case Success(l) => ...
```

engine/src/main/scala/cromwell/services/EngineMetadataServiceActor.scala
```diff
@@ -86,15 +86,26 @@ case class EngineMetadataServiceActor(serviceConfig: Config, globalConfig: Confi
   private def queryWorkflowOutputsAndRespond(id: WorkflowId): Unit = {
     val replyTo = sender()
     dataAccess.queryWorkflowOutputs(id) onComplete {
-      case Success(o) => replyTo ! WorkflowOutputsResponse(id, o)
+      case Success(o) =>
+        if (o.isEmpty) {
+          self ! HandleNotFound(id, replyTo)
+        } else {
+          replyTo ! WorkflowOutputsResponse(id, o)
+        }
```
mcovarr: separate cases would be good for this too

engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
```diff
@@ -1,133 +1,55 @@
...
   val unknownId = WorkflowId(UUID.randomUUID())
   val submittedWorkflowId = WorkflowId(UUID.randomUUID())
   val submittedScatterWorkflowId = WorkflowId(UUID.randomUUID())
   val abortedWorkflowId = WorkflowId(UUID.randomUUID())
-
-  def props: Props = Props(classOf[MockWorkflowManagerActor])
 }
 
 class MockWorkflowManagerActor extends Actor {
-  this: WorkflowDescriptorBuilder =>
```
mcovarr: love all this red on this class!  😄

engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
```diff
@@ -140,26 +62,55 @@ class CromwellApiServiceSpec extends FlatSpec with CromwellApiService with Scala
...
+    val summaryResult = summaryActor.ask(SummarizeMetadata)(timeout)
+
+    val askResult = summaryResult.futureValue(patienceConfig)
+    askResult match {
+      case MetadataSummarySuccess =>
+      case _ =>
+        fail()
+    }
+
+  }
```
mcovarr: These two methods seem like they might be more generally useful than just this test class, and actually I wonder if they might already exist somewhere

engine/src/main/scala/cromwell/engine/db/DataAccess.scala
cjllanwarne: Is it definitely running?

Or maybe it genuinely wasn't working before, and this has fixed it?

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
```diff
@@ -60,7 +61,7 @@ trait CromwellApiService extends HttpService with PerRequestCreator with Service
       get {
         Try(WorkflowId.fromString(workflowId)) match {
           case Success(w) =>
-            requestContext => perRequest(requestContext, MetadataBuilderActor.props(serviceRegistryActor), GetStatus(w))
+            requestContext => perRequest(requestContext, metadataBuilderProps, GetStatus(w))
```
cjllanwarne: Since they're all basically the same you could collapse these even further with
```
def handleMetadataRequest(req) = requestContext => perRequest(requestContext, metadataBuilderProps, req)
```
and then in the API calls, e.g:
```
case Success(w) => handleMetadataRequest(GetStatus(w))
```

engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
cjllanwarne: "better red than dead"

engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
```diff
@@ -1,133 +1,55 @@
...
-          WorkflowManagerWorkflowOutputsSuccess(id, Map(
-            "three_step.cgrep.count" -> JobOutput(int8, Option(int8hash)),
-            "three_step.ps.procs" -> JobOutput(file, Option(fileHash)),
-            "three_step.wc.count" -> JobOutput(int8, Option(int8hash))))
-        //case w => WorkflowManagerWorkflowOutputsFailure(id, new WorkflowNotFoundException(s"Workflow '$w' not found"))
+        case MockWorkflowManagerActor.runningWorkflowId => WorkflowManagerAbortSuccess(id)
+        case MockWorkflowManagerActor.abortedWorkflowId =>
+          WorkflowManagerAbortFailure(id, new IllegalStateException(s"Workflow ID '$id' is in terminal state 'Aborted' and cannot be aborted."))
+        case x =>
+          WorkflowManagerAbortFailure(id, new Exception(s"Couldn't abort $id because no workflow with that ID is in progress"))
```
cjllanwarne: ToL: Is there a value in all this logic? Are we in danger of making tests which just test that we've mocked out the WorkflowManagerActor correctly?

I'd rather have different flavours of `MockWorkflowManagerActor` which, for example, always return a `WorkflowManagerAbortFailure`, or which always return `WorkflowManagerAbortSuccess` so we can reliably test what happens when we receive the various messages

engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
```diff
@@ -140,26 +62,55 @@ class CromwellApiServiceSpec extends FlatSpec with CromwellApiService with Scala
   import scala.concurrent.duration._
 
   // The submit route takes a bit longer than the default 1 s while things initialize, when this spec is run by itself
-  implicit val defaultTimeout = RouteTestTimeout(5.seconds.dilated)
+  implicit val defaultTimeout = RouteTestTimeout(30.seconds.dilated)
```
cjllanwarne: ToL: This is a sad development!

engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
```diff
@@ -140,26 +62,55 @@ class CromwellApiServiceSpec extends FlatSpec with CromwellApiService with Scala
...
-  // s"CromwellApiService $version" should "return 404 for get of unknown workflow" in {
-  ignore should "return 404 for get of unknown workflow" in {
-    Get(s"/workflows/$version/${MockWorkflowManagerActor.unknownId}") ~>
-      sealRoute(statusRoute) ~>
+  def publishMetadata(events: Seq[MetadataEvent]): Unit = {
+    val timeout: Timeout = 5.seconds.dilated
+    val patienceConfig = PatienceConfig(timeout.duration)
+
+    import akka.pattern.ask
+    val putResult = serviceRegistryActor.ask(PutMetadataAction(events))(timeout)
```
cjllanwarne: ToL: This is `ask`ing for ask timeouts!

engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
```diff
@@ -140,26 +62,55 @@ class CromwellApiServiceSpec extends FlatSpec with CromwellApiService with Scala
...
+      case MetadataSummarySuccess =>
+      case _ =>
+        fail()
+    }
+
+  }
+
+  behavior of "REST API /status endpoint"
+
+  it should "return 404 for get of unknown workflow" in {
```
cjllanwarne: Are #988 and #990 still relevant or closed by this PR?

engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
```diff
@@ -589,36 +468,55 @@ class CromwellApiServiceSpec extends FlatSpec with CromwellApiService with Scala
...
+  it should "return good results for a good query" in {
+    val workflowId = WorkflowId.randomId()
+    val runningId = WorkflowId.randomId()
+    val events = Seq(
+      MetadataEvent(MetadataKey(workflowId, None, WorkflowMetadataKeys.Status), MetadataValue("Succeeded")),
+      MetadataEvent(MetadataKey(runningId, None, WorkflowMetadataKeys.Status), MetadataValue("Running"))
+    )
+
+    publishMetadata(events)
+    forceSummary()
```
cjllanwarne: ToL: At this point, are we really testing the API, or testing the underlying `MetadataService` architecture? I ask because we're now requiring a lot of knowledge of its inner workings in a logically unrelated test

It feels like to be intellectually honest we should mock out the `ServiceRegistryActor` to provide what we want to the REST API, and check that the REST API is acting correctly in various conditions.

cjllanwarne:
Sorry about all the comments regarding the philosophy of tests! I personally think they're important but they are ToL so I won't hold back the 👍

mcovarr:
:+1: with suggested cleanups

engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
kcibul: not sure if I agree this isn't "intellectually honest" but I'd be happy to participate in a discussion led by the team about which kinds of tests the teams feels should go where (unit, mocks, integration, etc).  

The way I think about it here is that... these tests don't take very long to run and they definitely test things are aren't currently tested elsewhere, so that's a "good thing" as opposed to not having them right now because we don't yet have a team agreement on testing philosophy (which I do agree would be great to have).

engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
kcibul: I think they are also closed by this PR, just assigned them to myself

engine/src/main/scala/cromwell/engine/db/DataAccess.scala
kcibul: I don't think it was working before.  I just ran a workflow on a cromwell server off develop and the /outputs endpoint just returns the workflow id and no outputs (even though /metadata has it)

PR #1093, opened 2016-06-29
Oops, these PostMVPs should all be ignored

cjllanwarne:


cjllanwarne:
For benefit @kshakir and @kcibul, please review!

Horneth:
I though the advantage of tagging them was that we could exclude them from "sbt test" while still having them "unignored" so they don't count against the "ignored" count.

kshakir:
Assuming `PostMVP` tagged tests are filtered out of `sbt test`, then they should never be _added_ to the ignored count. This patch therefore only adds to the ignored count for `sbt alltests:test`.

kshakir:
👍

mcovarr:
:+1:

kcibul:
I think you've got your thumbs, but you've got my 👍 as well!

PR #1098, opened 2016-06-29
Updated MaterializeWorkflowDescriptorActor to use its conf when assigning Backends

cjllanwarne:


cjllanwarne:
The wheel chooses @scottfrazer and @ruchim

engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -61,20 +63,20 @@ object MaterializeWorkflowDescriptorActor {
   /*
   Data
    */
-  case class ShadowMaterializeWorkflowDescriptorActorData()
+  case class MaterializeWorkflowDescriptorActorData()
 
-  import lenthall.config.ScalaConfig._
+  //import lenthall.config.ScalaConfig._
```
scottfrazer: Remove this?

scottfrazer:
👍

engine/src/main/scala/cromwell/engine/backend/CromwellBackends.scala
```diff
@@ -9,48 +9,37 @@ import scala.util.{Success, Try, Failure}
...
 
 object CromwellBackends {
 
   var instance: Option[CromwellBackends] = None
 
 
-  def isValidBackendName(name: String): Boolean = evaluateOrThrow(_.isValidBackendName(name))
-  def shadowDefaultBackend = evaluateOrThrow(_.shadowDefaultBackend)
-  def shadowBackendLifecycleFactory(backendName: String) = evaluateOrThrow(_.shadowBackendLifecycleFactory(backendName))
+  def isValidBackendName(name: String): Boolean = evaluateIfInitialised(_.isValidBackendName(name))
```
geoffjentry: :us:

engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -29,7 +29,9 @@ object MaterializeWorkflowDescriptorActor {
 
   val RuntimeBackendKey: String = "backend"
 
-  def props(serviceRegistryActor: ActorRef, workflowId: WorkflowId): Props = Props(new MaterializeWorkflowDescriptorActor(serviceRegistryActor, workflowId))
+  def cromwellBackends = CromwellBackends.instance.get
```
geoffjentry: If this throws, what happens?

geoffjentry:
👍

engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
cjllanwarne: In short, if this throws then `CromwellBackends.evaluateIfInitialised` would also have thrown. I'm relying here on standing on the shoulders of giants who would never let us get into such a sorry state... 🙏

geoffjentry: It might be hard to get the emoji in there but can you put a comment to that effect in here? 

Since complexity and cognitive load are hot topics these days I find the cognitive load of having to figure out why the author determined the raw gets to be safe and/or not worth worrying about to be pretty high

engine/src/main/scala/cromwell/engine/backend/CromwellBackends.scala
cjllanwarne: Fine... but only because it's 🇺🇸's birthday...

geoffjentry: I figured this was your way of getting back at us young upstarts

PR #1099, opened 2016-06-29
Create GCS Filesystems and Genomics once.  Closes #813.  Closes #1065.

### LINKED ISSUE #813: Create GCS filesystem(s) in JES initialization actor
Create these filesystems in the initialization actor so they might be shared with the other JES actors in the workflow.

### LINKED ISSUE #1065: Scaling Problems above ~15 workflows
People are reporting significant slowdown when running more than about 15 workflows concurrently

mcovarr:


backend/src/main/scala/cromwell/backend/package.scala
```diff
@@ -59,4 +59,16 @@ package object backend {
   }
 
   case class PreemptedException(msg: String) extends Exception(msg)
+
```
geoffjentry: What's going on here? Are these supposed to extend BID? Is there a reason why BID isn't sealed? Or does this just suspiciously look ADT-ish but ain't?

mcovarr: All the implementations of BID are implementation specific and will live elsewhere, so this can't be sealed and isn't really an ADT anyway.

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -114,9 +115,14 @@ object WorkflowActor {
...
                                workflowDescriptor: Option[EngineWorkflowDescriptor],
+                               initializationData: AllBackendInitializationData,
                                lastStateReached: StateCheckpoint)
   object WorkflowActorData {
-    def empty = WorkflowActorData(currentLifecycleStateActor = None, workflowDescriptor = None, StateCheckpoint(WorkflowUnstartedState))
+    def empty = WorkflowActorData(
+      currentLifecycleStateActor = None,
+      workflowDescriptor = None,
+      initializationData = AllBackendInitializationData.empty,
+      StateCheckpoint(WorkflowUnstartedState))
```
cjllanwarne: I can see this isn't really your code, but it'd be nice to put `lastStateReached = StateCheckpoint(WorkflowUnstartedState)` to match all the others

cjllanwarne:
Just to be clear, I think in April we chose Option 1 with a sprinkling of Option 3 in https://docs.google.com/a/broadinstitute.com/document/d/1feRDusWXQQ2pJ03sNHTNmrrnnwL3y-vtyF1fv_RdogU/edit?usp=sharing whereas this is clearly Option 2...

Not necessarily wrong, just pointing it out :)

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
geoffjentry: Technically @mcovarr did insert a named parameter before the unnamed parameter so now it *is* his code. Bwahahaha.

@mcovarr Could you make that not be the case plz? :)

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowInitializationActor.scala
```diff
@@ -54,10 +57,17 @@ case class WorkflowInitializationActor(workflowId: WorkflowId, workflowDescripto
   override val failureState = InitializationFailedState
   override val abortedState = InitializationsAbortedState
 
-  override val successResponse = WorkflowInitializationSucceededResponse
+  override def successResponse(data: WorkflowLifecycleActorData) = {
```
geoffjentry: Can you explain what all is going on here?

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesBackendLifecycleActorFactory.scala
```diff
@@ -4,13 +4,21 @@ import java.nio.file.Path
...
+import cromwell.backend._
 import cromwell.core.{ExecutionStore, OutputStore}
 import wdl4s.Call
 import wdl4s.expression.WdlStandardLibraryFunctions
 
+import scala.language.postfixOps
+
+object JesBackendLifecycleActorFactory {
+  implicit class Jessify(val genericInitializationData: Option[BackendInitializationData]) {
+    def toJes = genericInitializationData collectFirst { case d: JesBackendInitializationData => d } get
```
geoffjentry: Is danger your middle name, is that get actually safe, or did you just throw your hands up in the air like you, well, you know

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesInitializationActor.scala
```diff
@@ -52,22 +61,41 @@ class JesInitializationActor(override val workflowDescriptor: BackendWorkflowDes
   /**
     * Abort all initializations.
     */
+  // TODO PBE ??? is not cool
   override def abortInitialization(): Unit = ???
```
geoffjentry: Since we believe abort to be feature complete, could you file a quick ticket on this?

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesInitializationActor.scala
```diff
@@ -52,22 +61,41 @@ class JesInitializationActor(override val workflowDescriptor: BackendWorkflowDes
...
+
+    def buildGcsFileSystem: GcsFileSystem = {
+      val storage = jesConfiguration.jesAttributes.gcsFilesystemAuth.buildStorage(
+        workflowDescriptor.workflowOptions.toGoogleAuthOptions, jesConfiguration.googleConfig)
+      GcsFileSystem(GcsFileSystemProvider(storage))
+    }
+
+    for {
+      // generate single filesystem and genomics instances
+      gcsFileSystem <- Future(buildGcsFileSystem)
```
geoffjentry: Why not set these two Futures up outside of the forcomp since they don't seem to rely on each other?

geoffjentry:
Tentative 👍 - I asked for clarification on one part and since that looked to be kind of an important block I'd like to understand it better first.

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
mcovarr: yeah great  😦

geoffjentry:
Just saw the comments @cjllanwarne made about the old google doc and a potential TT. I'd concur, just to make sure this is thought through. In particular on the google doc the side discussion about implementing such behavior as back channels in specific backends instead of abstracting it out to general behavior.

mcovarr:
@geoffjentry already discussed with @cjllanwarne.  I actually had forgotten that doc existed or I would have TT'd this approach first, but we agreed it would be best to just proceed with the approach in this PR basically as is.  If you disagree and/or want to TT we can definitely do that.

geoffjentry:
I don't want to block this PR, although in the unlikely event that I actually remember come Tuesday I might want to chat w/ you two briefly just to get my own thinking straight

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesBackendLifecycleActorFactory.scala
mcovarr: all of the above

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesInitializationActor.scala
mcovarr: sure

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowInitializationActor.scala
mcovarr: Actually, I'd rather not.  😛 

After further discussion with @cjllanwarne and lingering remorse from so flagrantly having flouted the resolutions of our esteemed Google Doc, I'd like to have a TTT (Tuesday Tech Talk) to redo this a la Option 3, back-channel-service style.  The requirements of the service appear to be distinct from those in our existing service menagerie, so we'd either need to enhance an existing service or add a new one.

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesInitializationActor.scala
mcovarr: you betcha

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowInitializationActor.scala
geoffjentry: Okee doke. Could we do it after the afternoon google sync (3:30+)?

Truth be told even when I first looked at the PR something was bugging me but I didn't know what it was and it was too low a level to bother commenting and I think it ties into the whole google doc options which I'd completely forgotten about. So while I was happy to go w/ the flow I'm 👍 on chatting about it more too

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesInitializationActor.scala
cjllanwarne: Aha! Could this explain my non-terminating abort ticket?

mcovarr: Possibly.  This one is #1110

geoffjentry: If it does I find it scary that we're just silently eating exceptions - it means we're back in the world where we're managing to both not let it crash but also not doing alerting anything to misbehavior.

mcovarr: We do seem to eat exceptions in some cases.  In the process of writing this I had done something boneheaded which caused one of these lifecycle actors to throw, but there was not a whiff of it in the logs.

PR #1102, opened 2016-06-30
REDO: Handle ActorInitializationException in supervisors Closes #1079 (#1084)

### LINKED ISSUE #1079: Workflow gets stuck in Initialization if BackendWorkflowInitializationActor crashe during instantiation

cjllanwarne:
With apologies to @Horneth, I merged too soon! This unfortunately caused  `sbt "test-only cromwellSimpleWorkflowActorSpec"` to hang indefinitely.

* Lazy-fiy workflowPaths to prevent actor from throwing during instantiation

* Handle ActorInitFailure with supervision

* with default'

* cleanup Workflow finalization

* clean imports

* PR commnet

### DIFF
--- a/engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
@@ -137,6 +137,8 @@ class WorkflowManagerActor(config: Config)
     /*
      Responses from services
      */
+      // TODO Restart: to be verified after restart is implemented but these WorkflowSucceededResponse/WorkflowFailedResponse seem useless
+      // Watching the transition should be enough for the WMA to do what it needs to
     case Event(WorkflowSucceededResponse(workflowId), data) =>
       log.info(s"$tag Workflow $workflowId succeeded!")
       stay()
--- a/engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
@@ -137,7 +137,7 @@ class WorkflowManagerActor(config: Config)
     /*
      Responses from services
      */
-      // TODO Restart: to be verified after restart is implemented but these WorkflowSucceededResponse/WorkflowFailedResponse seem useless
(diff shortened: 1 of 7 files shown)


engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -137,6 +137,8 @@ class WorkflowManagerActor(config: Config)
     /*
      Responses from services
      */
+      // TODO Restart: to be verified after restart is implemented but these WorkflowSucceededResponse/WorkflowFailedResponse seem useless
```
mcovarr: TODO PBE will make this easier to find in a future sweep

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesFinalizationActor.scala
```diff
@@ -40,9 +39,7 @@ class JesFinalizationActor (override val workflowDescriptor: BackendWorkflowDesc
 
   private def deleteAuthenticationFile(): Future[Unit] = {
     if (jesConfiguration.needAuthFileUpload) {
-      val delete = () => Future(workflowPaths.gcsAuthFilePath.delete(false))
-
-      Retry.withRetry(delete, isFatal = isFatalJesException, isTransient = isTransientJesException)(context.system) map { _ => () }
+      Future(workflowPaths.gcsAuthFilePath.delete(false)) map { _ => () }
```
mcovarr: I'm going to guess there's already a sufficient layer of retries in the GCS filesystem code?

Horneth: Yes - every GCS operation is retried 3 times.

mcovarr:
:+1:

cjllanwarne:
👍 I'd also lobby for this to be part of the "Miguel explains big/exciting changes" thing on Friday, except maybe not force @mcovarr to explain it... This looks like a pattern we should use more often

PR #1106, opened 2016-07-01
Add PBS backend

delocalizer:
I offer for your consideration a PBSPro/Torque backend that we are using here at QIMR Berghofer (based on SGE one), if you think it might be useful to encourage external user uptake in the short term. I was going to hold off until you had finished the PBE work to refactor this functionality to work with that but I see that you've accepted a PR for LSF backend using the current structure and I don't know what your broader priorities are. If you don't think it's worth your time to review then that's ok too :)


geoffjentry:
Hey Conrad - Thanks, this is awesome.

To give some insight on how things are playing out in the hopefully-not-too-long-term, we're planning on cutting an alpha release of the PBE stuff imminently (perhaps today?) which is something we feel is stable enough to start poking at but is missing a few features we need in our production use cases (restart/recovery, call caching), backends other than local/JES, and with some known warts we need to hammer out. I'm guessing you're looking at roughly a month for something more stable than that, although I'm famous for my "about a month" predictions. However, since you're already pretty up to speed with what's going on, I'd say that the 0.20 alpha should be stable enough to work up a backend. It'd at least be a good test case as someone who *did* figure out how to make one in the old system if the new system is inscrutable or not.

In terms of what to do with this PR, I'll somewhat leave it up to you. We're hoping to close the 0.19 books as much as possible once the alpha thing is out, but if you feel like it'll provide value to folks over the next month or so I'm happy to take some time to review it.

engine/src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
```diff
@@ -142,29 +152,67 @@ object CromwellRuntimeAttributes {
...
-    (docker |@| zones |@| failOnStderr |@| continueOnReturnCode |@| cpu |@| preemptible |@| disks |@| memory |@| bootDiskSize) {
-      new CromwellRuntimeAttributes(attributes, _, _, _, _, _, _, _, _, _)
-    } match {
+    
+    /*
+     * scalaz.syntax.ApplicativeBuilder |@| (as of version 7.3, anyhow) runs out of steam at 12
+     * arguments so we can't just keep extending what Broad had here to add queue, walltime, etc.
+     * The chained <*> below basically does the same thing.
+     */
+    val curriedCtor = (CromwellRuntimeAttributes(_: Map[String, WdlValue], _: Option[String], _: Seq[String], _: Boolean, _: ContinueOnReturnCode, _: Int, _: Int, _: Seq[JesAttachedDisk], _: Double, _: String, _: Option[String], _: Int)).curried
```
mcovarr: If you're willing to rename the `apply` on the companion object (which is more than ok with this reviewer), this could simplify to:

```scala
    val curriedCtor = (CromwellRuntimeAttributes.apply _).curried
```

engine/src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
```diff
@@ -142,29 +152,67 @@ object CromwellRuntimeAttributes {
...
+    bootDiskSize <*> (queue <*> ( walltime <*> (memory <*> ( disks <*> ( preemptible <*> ( cpu <*> (continueOnReturnCode <*> ( failOnStderr <*> ( zones <*> (docker map curriedCtor(attributes))))))))))) match {
```
mcovarr: Not at all required for this PR, but if you enjoy looking at mind-bending type-level code for boilerplate removal, some of this might be useful here someday:

http://meta.plasm.us/posts/2013/06/05/applicative-validation-syntax/

delocalizer:
Thanks for the insight Jeff.Let's just leave it then if the PBE stuff is happening so soon, I don't want to waste your time. I'd be very happy to be a guinea pig for making a backend using an early release of the new design. Cromwell rocks b.t.w.; and thanks for making it open!
cheers,C
Date: Fri, 1 Jul 2016 08:33:49 -0700

engine/src/main/scala/cromwell/engine/backend/runtimeattributes/CromwellRuntimeAttributes.scala
delocalizer: Haha I do enjoy looking at such things!  But also scared... I'm kind of leery of things like shapeless and even much of scalaz that might make my code look like magic to six-months-later-me, let alone poor non-me saps who have to maintain it. I'll leave that gear to the hard-core typeonauts... in production, anyway :)

Date: Fri, 1 Jul 2016 14:44:43 -0700

geoffjentry:
Sounds good to me, and you're welcome! Thanks for contributing!

ALTree:
[sorry for resurrecting this old thread]

Is Cromwell now stable enough for this kind of contributions? I see that the big backend changes were included in the 0.21 release ~1 month ago.

delocalizer:
@ALTree if you're referring specifically to PBS, that can now be supported with a little kludging purely through configuration. The only feature of PBS cluster that needs special treatment is the handling of stderr and stdout, which by default on PBS are copied to the execution directory only _after_  the job completes. [The config file in this gist](https://gist.github.com/delocalizer/fa29139675fb4118e908a4c80249dffb) works for me. Note that it requires that PBS_JOBDIR (the user's home directory by default, but can be a custom value) is shared across compute nodes.

ALTree:
@delocalizer ah, thanks. I didn't realize the new backend system was this flexible (the `README` is a little opaque).

Thank you for sharing your PBS configuration file! I confirm it works for me, too.

PR #1107, opened 2016-07-01
PostMVPified (and/or just deleted) some tests

cjllanwarne:


cjllanwarne:
Sorry @scottfrazer and @mcovarr, you're the only people here to review this! It should be very quick though!

mcovarr:
yes @scottfrazer and I have distinguished ourselves by our non-absence.  😄

engine/src/test/scala/cromwell/WorkflowOutputsSpec.scala
```diff
@@ -59,17 +59,5 @@ class WorkflowOutputsSpec extends CromwellTestkitSpec {
...
-      runWdlAndAssertOutputs(
-        sampleWdl = SampleWdl.ThreeStepWithInputsInTheOutputsSection,
-        eventFilter = EventFilter.info(pattern = s"is in a terminal state: WorkflowSucceededState", occurrences = 1),
-        runtime = "",
-        expectedOutputs = Map(
-          "three_step.cgrep.pattern" -> AnyValueIsFine
-        ),
-        allowOtherOutputs = false
-      )
-    }
```
mcovarr: I have a nagging feeling we supported this for a legit reason but I just don't remember what it was...

mcovarr:
:+1:

engine/src/test/scala/cromwell/WorkflowOutputsSpec.scala
cjllanwarne: I can sort of imagine a future world where workflows are composable, and you want to just "forward" an input through to the next stage in a pipeline of workflows... But for now it doesn't really serve any purpose

scottfrazer:
👍

PR #1108, opened 2016-07-01
Being honest about the state of call caching

cjllanwarne:


cjllanwarne:
Let's reply with honesty rather than sending an unrecognised message to an actor which probably has nothing to do with call caching anymore

mcovarr:
:+1: it really is the best policy

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
```diff
@@ -182,12 +182,8 @@ trait CromwellApiService extends HttpService with PerRequestCreator with Service
...
         val queryParameters = parameters map { case (k, v) => QueryParameter(k, v) }
         post {
-          Try(WorkflowId.fromString(workflowId)) match {
-            case Success(w) =>
-              requestContext =>
-                perRequest(requestContext, CromwellApiHandler.props(workflowManager), CromwellApiHandler.ApiHandlerCallCaching(w, queryParameters, callFqn))
-            case Failure(_) => invalidWorkflowId(workflowId)
-          }
+          // TODO: PBE: Certainly want to do something for this! But probably not to the WMA
+          failBadRequest(new RuntimeException(s"Call caching is currently unsupported."))
```
geoffjentry: `UnsupportedOperationException`?

geoffjentry:
👍

PR #1113, opened 2016-07-01
Added DockerTest tag for HtCondor test

cjllanwarne:


mcovarr:
:+1: spectagular

francares:
I don't know the scope of DockerTest but this test the only thing it does is to check the docker command generated for the job. It does not execute anything.

mcovarr:
@francares @cjllanwarne I think the actual issue here is that the "/tmp" assertions on lines 260 and 261 always fail on Mac, regardless of whether Docker is available or not.

cjllanwarne:
@francares @mcovarr I've modified this (fairly bluntly however) so that it passes on mac. If you have a suggestion for a stronger test case let me know

francares:
👍

PR #1124, opened 2016-07-05
JobStore read-side skeleton

cjllanwarne:
This is obviously a long way from being feature-complete, but is hopefully enough of a skeleton/statement of intent for JobStore readers to develop against in the meanwhile...

### DIFF
--- a/engine/src/main/scala/cromwell/jobstore/JobStoreReader.scala
+++ b/engine/src/main/scala/cromwell/jobstore/JobStoreReader.scala
@@ -1,0 +1,28 @@
+package cromwell.jobstore
+
+import akka.actor.{Actor, ActorLogging}
+import akka.event.LoggingReceive
+import cromwell.core.JobKey
+
+class JobStoreReader extends Actor with ActorLogging {
+  override def receive = LoggingReceive {
+    case QueryJobCompletion(jobKey) => sender ! JobNotComplete(jobKey)
+    case unknownMessage => log.error(s"Unexpected message to JobStoreReader: $unknownMessage")
+  }
+}
+
+/**
+  * Message to query the JobStoreReader, asks whether the specified job has already been completed.
+  */
+case class QueryJobCompletion(jobKey: JobKey)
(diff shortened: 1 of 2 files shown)


cjllanwarne:
@Horneth should be relevant for your EJEA stuff. Needs one other reviewer (whoever fancies it, or otherwise I'll wheel for it)

engine/src/main/scala/cromwell/jobstore/JobStoreReader.scala
```diff
@@ -0,0 +1,17 @@
...
+  override def receive = LoggingReceive {
+    case IsJobCompletedYet(jobKey) => sender ! JobNotCompleteYet(jobKey)
+    case unknownMessage => log.error(s"Unexpected message to JobStoreReader: $unknownMessage")
+  }
+}
+
+case class IsJobCompletedYet(jobKey: JobKey)
+
+case class JobAlreadyCompleted(jobKey: JobKey, jobResult: JobResult)
+case class JobNotCompleteYet(jobKey: JobKey)
```
Horneth: Wouldn't `JobComplete` `JobRunning` be descriptive enough ?

Horneth:
LGTM, if we realize something is missing implementing EJEA I'll drop a ticket 👍

engine/src/main/scala/cromwell/jobstore/JobStoreReader.scala
cjllanwarne: I guess I could change it to `JobComplete` and `JobNotComplete`? I think there's no reason to distinguish running vs unstarted here, that would be a backend issue at restore-time

mcovarr: Yeah I like `JobComplete` and `JobNotComplete`.  `JobRunning` seems to imply a more specific knowledge of state that I don't think this code will know?

mcovarr:
:+1:

Issue #1118, opened 2016-07-05
Use WorkflowStore to support restart


cjllanwarne:
Depends on #1114

Update the WorkflowStore to resubmit non-complete jobs as restarts when the system comes up.

PR #1127, opened 2016-07-06
Make workflow submission async. Closes #1114

### LINKED ISSUE #1114: Async workflow submission via WorkflowStore
Does not currently handle restart

geoffjentry:


geoffjentry:
- Still need to add the new config options to the README
- Something in the tests is broken, still working on that

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -57,12 +55,16 @@ object WorkflowManagerActor {
...
   case class WorkflowManagerData(workflows: Map[WorkflowId, ActorRef]) {
-    def idFromActor(actor: ActorRef): Option[WorkflowId] =
-      workflows.collectFirst { case (id, a) if a == actor => id }
+    def idFromActor(actor: ActorRef): Option[WorkflowId] = workflows.collectFirst { case (id, a) if a == actor => id }
     def withAddition(entry: WorkflowIdToActorRef): WorkflowManagerData =
       this.copy(workflows = workflows + (entry.workflowId -> entry.workflowActor))
-    def without(id: WorkflowId): WorkflowManagerData =
-      this.copy(workflows = workflows - id)
+
+    def withAddition(entries: List[WorkflowIdToActorRef]): WorkflowManagerData = {
```
geoffjentry: I waffled on this matter so wanted to point it out. If people don't like the overloaded signature, I'm happy to call it something like `withAdditions`, collapse the single form above into this somehow, etc.

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
```diff
@@ -0,0 +1,145 @@
...
+    val workflowId = WorkflowId(UUID.randomUUID())
+    workflowStore += (workflowId -> SubmittedWorkflow(source, Submitted))
+    workflowId
+  }
+
+  /**
+    * Provides a bulk add functionality. Adds all of the provided WorkflowSourceFiles at once, returning a List
+    * of UUIDs matching the order of the supplied source files
+    */
+  def add(sources: Seq[WorkflowSourceFiles]): immutable.List[WorkflowId] = {
```
geoffjentry: This is another place I waffled on the overloading (actually 99% of the waffling was here, the other one just came along for the ride).

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
```diff
@@ -0,0 +1,145 @@
...
+    * flag to true
+    */
+  def fetchRunnableWorkflows(n: Int): immutable.List[WorkflowToStart] = {
+    val startableWorkflows = workflowStore filter { case (k, v) => v.state.isStartable } take n
+    workflowStore ++= startableWorkflows map { case (k, v) => k -> SubmittedWorkflow(v.sources, Running) }
+    // Because we explicitly filtered out for startable states this is safe
+    startableWorkflows map { case (k, v) => v.toWorkflowToStart(k) } toList
+  }
+
+  // FIXME: Do we care if it doesn't exist? If so to what extent
```
geoffjentry: Question for reviewers. I initially was in the "yes, very much" category but have talked myself down to "meh, who cares". I'm easily talked back upwards here

engine/src/test/scala/cromwell/engine/WorkflowStoreSpec.scala
```diff
@@ -0,0 +1,39 @@
...
+
+  it should "return only the remaining workflows if N is larger than size" in {
+    val origIds = dump.filter({ case (k, v) => v.state.isStartable }).keySet
+    val newIds = fetchRunnableWorkflows(origIds.size + 5) map { _.id }
+    newIds should have size origIds.size
+    val currentIds = dump.filter({ case (k, v) => v.state.isStartable }).keySet
+    currentIds shouldBe empty
+  }
+
+  // remove
```
geoffjentry: Oops, looks like I forgot some stuff

core/src/main/resources/application.conf
```diff
@@ -52,8 +52,14 @@ system {
   // Max number of retries per job that the engine will attempt in case of a retryable failure received from the backend
   max-retries = 10
 
-  //If 'true' then when Cromwell starts up, it tries to restart incomplete workflows
+  // If 'true' then when Cromwell starts up, it tries to restart incomplete workflows
   workflow-restart = true
+
+  // Cromwell will fetch N submitted workflows at a time. That number is tunable here
```
mcovarr: s/will fetch N/will launch up to N/

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
mcovarr: Yeah with `NonEmptyList`

core/src/main/resources/application.conf
geoffjentry: Yeah, artifact of old nomenclature I was using internally. Will change.

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -191,32 +204,24 @@ class WorkflowManagerActor(config: Config)
...
     * Workflow ID -> WorkflowActorRef entry.
     */
-  private def submitWorkflow(source: WorkflowSourceFiles, replyTo: Option[ActorRef], id: Option[WorkflowId]): WorkflowIdToActorRef = {
-    val workflowId: WorkflowId = id.getOrElse(WorkflowId.randomId())
-    val isRestart = id.isDefined
+  private def submitWorkflow(workflow: WorkflowStore.WorkflowToStart): WorkflowIdToActorRef = {
+    val workflowId = workflow.id
 
-    val startMode = if (isRestart) {
+    val startMode = if (workflow.state == WorkflowStore.Restartable) {
```
mcovarr: TOL Restartable seems like an odd name for a state...

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
```diff
@@ -0,0 +1,145 @@
...
+  * changes, revisit.
+  */
+trait WorkflowStore {
+  import WorkflowStore._
+
+  /*
+    The LinkedHashMap preserves insertion order which means that any requests for runnable workflows will
+    be satisfied by taking the oldest available.
+   */
+  private var workflowStore = new mutable.LinkedHashMap[WorkflowId, SubmittedWorkflow]
```
mcovarr: yo dog I heard you liked mutability

mcovarr: maybe a var + immutable `ListMap`?

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
```diff
@@ -0,0 +1,145 @@
...
+
+  /**
+    * Adds a WorkflowSourceFiles to the store and returns a WorkflowId for tracking purposes
+    */
+  def add(source: WorkflowSourceFiles): WorkflowId = {
+    val workflowId = WorkflowId(UUID.randomUUID())
```
mcovarr: `WorkflowId.randomId()`, and yes with parens as it's not referentially transparent

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
mcovarr: yes + `NonEmptyList`

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
mcovarr: Unless there's an eventual consistency issue that would make this unsurprising we should probably at least log it.

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
```diff
@@ -0,0 +1,145 @@
...
+
+  override def receive = {
+    case SubmitWorkflow(source) =>
+      val id = add(source)
+      sendIdToMetadataService(id)
+      sender ! WorkflowAdded(id)
+    case SubmitWorkflows(sources) =>
+      val ids = add(sources)
+      ids foreach sendIdToMetadataService
+      sender ! WorkflowsAdded(add(sources))
```
mcovarr: `add` has been called twice in this block, was that intentional?

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
```diff
@@ -0,0 +1,145 @@
...
+object WorkflowStoreActor {
+  sealed trait WorkflowStoreActorCommand
+  case class SubmitWorkflow(source: WorkflowSourceFiles) extends WorkflowStoreActorCommand
+  case class SubmitWorkflows(sources: Seq[WorkflowSourceFiles]) extends WorkflowStoreActorCommand
+  case class FetchRunnableWorkflows(n: Int) extends WorkflowStoreActorCommand
+  case class RemoveWorkflow(id: WorkflowId) extends WorkflowStoreActorCommand
+
+  sealed trait WorkflowStoreActorResponse
+  case class WorkflowAdded(workflowId: WorkflowId) extends WorkflowStoreActorResponse
+  case class WorkflowsAdded(workflowIds: List[WorkflowId]) extends WorkflowStoreActorResponse
```
mcovarr: is this `List` mutable, or are all the other `immutable.List` qualifications unnecessary?

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
```diff
@@ -0,0 +1,145 @@
...
+    case SubmitWorkflows(sources) =>
+      val ids = add(sources)
+      ids foreach sendIdToMetadataService
+      sender ! WorkflowsAdded(add(sources))
+    case FetchRunnableWorkflows(n) => sender ! NewWorkflows(fetchRunnableWorkflows(n))
+    case RemoveWorkflow(id) => remove(id)
+  }
+
+  /**
+    * Takes the workflow id and sends it over to the metadata service w// default empty values for inputs/outputs
```
mcovarr: s!//!/!

engine/src/main/scala/cromwell/webservice/CromwellApiHandler.scala
```diff
@@ -89,32 +88,15 @@ class CromwellApiHandler(requestHandlerActor: ActorRef) extends Actor with Workf
         case _ => RequestComplete(StatusCodes.InternalServerError, APIResponse.error(e))
       }
 
-    case ApiHandlerWorkflowSubmit(source) =>
-      val submitMsg = WorkflowManagerActor.SubmitWorkflowCommand(source)
-      requestHandlerActor ! submitMsg
-    case WorkflowManagerSubmitSuccess(id) =>
+    case ApiHandlerWorkflowSubmit(source) => requestHandlerActor ! WorkflowStoreActor.SubmitWorkflow(source)
+
+    case WorkflowStoreActor.WorkflowAdded(id) =>
```
mcovarr: Submit != Added, can this be made symmetric?

engine/src/test/scala/cromwell/CromwellTestkitSpec.scala
```diff
@@ -176,7 +177,8 @@ object CromwellTestkitSpec {
   implicit class EnhancedWorkflowManagerActor(val manager: TestActorRef[WorkflowManagerActor]) extends AnyVal {
 
     def submit(sources: WorkflowSourceFiles): WorkflowId = {
-      val submitMessage = WorkflowManagerActor.SubmitWorkflowCommand(sources)
+      val newWorkflow = WorkflowToStart(WorkflowId(UUID.randomUUID()), sources, Submitted)
```
mcovarr: Workflow.randomId() again

engine/src/test/scala/cromwell/engine/WorkflowStoreSpec.scala
```diff
@@ -0,0 +1,39 @@
+package cromwell.engine
+
+import cromwell.engine.workflow.WorkflowStore
+import cromwell.util.SampleWdl.HelloWorld
+import org.scalatest.{FlatSpec, Matchers}
+
+class WorkflowStoreSpec extends FlatSpec with Matchers with WorkflowStore {
```
mcovarr: It's kind of strange that this extends the `WorkflowStore` trait directly, and makes it so these "unit" tests are cross-talking.

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
geoffjentry: It's what we decided on last week. See the doc.

engine/src/test/scala/cromwell/engine/WorkflowStoreSpec.scala
```diff
@@ -0,0 +1,39 @@
...
+    newIds should have size 1
+    val currentIds = dump.filter({ case (k, v) => v.state.isStartable }).keySet
+    currentIds should be (origIds -- newIds)
+  }
+
+  it should "return only the remaining workflows if N is larger than size" in {
+    val origIds = dump.filter({ case (k, v) => v.state.isStartable }).keySet
+    val newIds = fetchRunnableWorkflows(origIds.size + 5) map { _.id }
+    newIds should have size origIds.size
+    val currentIds = dump.filter({ case (k, v) => v.state.isStartable }).keySet
```
mcovarr: underscore `k`

engine/src/test/scala/cromwell/engine/WorkflowStoreSpec.scala
```diff
@@ -0,0 +1,39 @@
...
+  it should "fetch no more than N workflows" in {
+    val origIds = dump.keySet
+    val newIds = fetchRunnableWorkflows(1) map { _.id }
+    newIds should have size 1
+    val currentIds = dump.filter({ case (k, v) => v.state.isStartable }).keySet
+    currentIds should be (origIds -- newIds)
+  }
+
+  it should "return only the remaining workflows if N is larger than size" in {
+    val origIds = dump.filter({ case (k, v) => v.state.isStartable }).keySet
```
mcovarr: underscore `k`

geoffjentry:
As per a convo with @mcovarr I'm going to switch out the mutable map for an immutable list structure. It'll be less performant but a lot cleaner, and it'd be easy to switch out if performance every did become an issue here. To paraphrase Miguel, something else likely will have blown up in Cromwell prior to performance being an issue here.

engine/src/test/scala/cromwell/engine/WorkflowStoreSpec.scala
geoffjentry: It's because I wanted the test to be a little whitebox-y but I didn't want what became the `dump` method to be public. I suppose what ended up here is in the worst of both worlds camp.

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
geoffjentry: Yes to collapsing on these?

IIRC the main issue on this one was how to discern that these requests came from the single submit endpoint vs the batch submit as a batch submit of 1 workflow still is rendered differently on the way out. Not wanting to think that one through is why I ultimately ended up having two different entry points here, FWIW.

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
mcovarr: Yeah I know, it just seems a little weird in code.  I'll probably get used to it.  😄

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
geoffjentry: I was logging it for a while but took it out because both a) there are already spots in the WMA where it also silently soldiers on in cases like this and b) It seems like the sort of thing that no one would ever notice anyways. 

It's easy enough to add a log in there, it's of the This Shouldn't Happen category but not enough to blow the system up IMO

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
geoffjentry: I fiddled with wording a lot in general as a lot of the "obvious" terms for things seemed unwieldy once I started using them in places. So I don't disagree w/ you, just that at some point I decided to stop worrying about it.

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
geoffjentry: See global comment on changing from the map structure

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
geoffjentry: It's almost as if someone thought of this a long time ago

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
geoffjentry: copy/paste error on my part moving things around last minute

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
geoffjentry: This stuff should all change soon but as I had a mutable construct I was trying to be good about explicitly marking everything mutable or immutable. I missed one :) 

It was also why I wanted no part of using a Seq (besides my typical anti-Seq bias)

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
geoffjentry: Along these lines, one thought I had the other day was that we could just define things such that either all submissions had response json as a list or all submissions with a single entry were always just a single - regardless of which API endpoint they hit, but that seemed problematic (and at the very least would require buy in from our illustrious PO @kcibul).

mcovarr: These `add`s could be merged regardless of the decision on the API endpoints.

mcovarr:
:+1:

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
geoffjentry: True. W/ the split of the store funcs & the actor receive i could control it from there

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
geoffjentry: Granted I'm changing this anyways but I finally got your comment. Ugh.

core/src/main/resources/application.conf
```diff
@@ -52,8 +52,14 @@ system {
   // Max number of retries per job that the engine will attempt in case of a retryable failure received from the backend
   max-retries = 10
 
-  //If 'true' then when Cromwell starts up, it tries to restart incomplete workflows
+  // If 'true' then when Cromwell starts up, it tries to restart incomplete workflows
   workflow-restart = true
+
+  // Cromwell will launch up to N submitted workflows at a time. That number is tunable here
```
cjllanwarne: This is a little confusing. I think you mean `will allow up to N submitted workflows to be running concurrently`?

cjllanwarne: Also the `That number is tunable here` feels a little unnecessary :)

engine/src/main/scala/cromwell/engine/workflow/SingleWorkflowRunnerActor.scala
```diff
@@ -104,15 +110,16 @@ case class SingleWorkflowRunnerActor(source: WorkflowSourceFiles,
 
   when (NotStarted) {
     case Event(RunWorkflow, data) =>
-      log.info(s"$Tag: Launching workflow")
-      val submitMessage = SubmitWorkflowCommand(source)
-      workflowManager ! submitMessage
+      log.info(s"$Tag: Submitting workflow")
+      workflowStore ! SubmitWorkflow(source)
```
cjllanwarne: I would think for the SingleWorkflowRunner case, you're actually fine to continue submitting directly to the WMA... 

Unless you want single runner to restart existing workflows, which IMO it shouldn't

geoffjentry: I had it that way first and then changed it. I can't remember why so it's impossible for me to say if the reason was good. If I remember in the near future I'll pipe up

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -23,6 +20,8 @@ import scala.concurrent.{Await, Promise}
 import scala.language.postfixOps
 
 object WorkflowManagerActor {
+  val DefaultMaxWorkflowsToLaunch = 50
+  val DefaultNewWorkflowPollRate = 20
```
cjllanwarne: I actually think I prefer this as a push rather than a pull, since it means the WMA doesn't need to change at all, and you can have different clients behave differently (e.g. the single runner could submit instantly, whereas the server could delay and submit in batches)

geoffjentry: I disagree. Having it as a pull means that we can tune the WMA to accept more/less work for different situations. Granted there's a lot more complexity than "number of workflows in flight" but the reality is that people are already using that number for the applications which they're building external to Cromwell.

Also I don't see why "WMA doesn't have to change" is a big deal, it's a minor change and IMO a bit cleaner anyways.

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
```diff
@@ -0,0 +1,145 @@
...
+      ids foreach sendIdToMetadataService
+      sender ! WorkflowsSubmitted(ids)
+    case FetchRunnableWorkflows(n) => sender ! NewWorkflows(fetchRunnableWorkflows(n))
+    case RemoveWorkflow(id) => remove(id)
+  }
+
+  /**
+    * Takes the workflow id and sends it over to the metadata service w/ default empty values for inputs/outputs
+    */
+  private def sendIdToMetadataService(id: WorkflowId): Unit = {
```
cjllanwarne: TOL: This name is a bit of a misnomer

engine/src/test/scala/cromwell/engine/WorkflowStoreSpec.scala
```diff
@@ -0,0 +1,39 @@
...
+
+  it should "return only the remaining workflows if N is larger than size" in {
+    val origIds = dump.filter({ case (_, v) => v.state.isStartable }).keySet
+    val newIds = fetchRunnableWorkflows(origIds.size + 5) map { _.id }
+    newIds should have size origIds.size
+    val currentIds = dump.filter({ case (_, v) => v.state.isStartable }).keySet
+    currentIds shouldBe empty
+  }
+
+  // remove
```
cjllanwarne: indeed.

engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
```diff
@@ -239,7 +251,7 @@ class CromwellApiServiceSpec extends FlatSpec with CromwellApiService with Scala
   }
 
   behavior of "REST API batch submission endpoint"
-
+  // FIXME and here too
```
cjllanwarne: ?

geoffjentry: oops leftover note to myself

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
cjllanwarne: in any case it'll become a DB access at some undisclosed future point... so (shrug)

geoffjentry: I was picturing it being like most of Ye Olde Worlde where there was a DB backing it but in memory handling for moment to moment stuff. But you're right that it might not end up that way.

mcovarr: I was holding off commenting on this point as I assumed what wasn't here would become part of #1118, but DB persistence of workflows in the old world is synchronous to a reply going back to the submitter.  I'd personally prefer to keep it that way so submissions aren't acknowledged unless we know we've saved the submission.

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
geoffjentry: Turns out the singular one wasn't being used anymore anyways. However, just for you I'll give the multiple one a little dose of 
nel

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
geoffjentry: @mcovarr agreed

However as fascinating as this discussion is, it's a bit moot at the moment :)

core/src/main/resources/application.conf
geoffjentry: @cjllanwarne re your first comment, no that is not the case. This doesn't limit the number of workflows in flight, just the number of workflows it fetches from the store on any one pull.

As an example, at least in the past we've had issues with too many things submitted at once (which e.g. led to bbq in Rawls). This allows a reasonable rate to be determined by Cromwell.

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
geoffjentry: Agreed, although the one I changed it to you might not like any better, but you'll need to wait for my next push to see it

core/src/main/resources/application.conf
mcovarr: Oh I thought the point of this was to limit the number of in-flight workflows, not the rate of dequeue.

geoffjentry: Discussed in person.

engine/src/main/scala/cromwell/engine/workflow/SingleWorkflowRunnerActor.scala
geoffjentry: Discussed in person, leaving as is

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -1,28 +1,29 @@
...
 import cromwell.webservice.CromwellApiHandler._
 import lenthall.config.ScalaConfig.EnhancedScalaConfig
 
 import scala.concurrent.duration._
 import scala.concurrent.{Await, Promise}
 import scala.language.postfixOps
+import scalaz.NonEmptyList
 
 object WorkflowManagerActor {
+  val DefaultMaxWorkflowsToRun = 5000
```
geoffjentry: I didn't quite pull these numbers out of a hat but it wasn't much better than that. If anyone wants to propose different numbers feel free to chime in

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
```diff
@@ -0,0 +1,152 @@
...
+    val startableWorkflows = workflowStore filter { _.state.isStartable } take n
+    val updatedWorkflows = startableWorkflows map { _.copy(state = Running) }
+    workflowStore = (workflowStore diff startableWorkflows) ++ updatedWorkflows
+
+    startableWorkflows map { _.toWorkflowToStart }
+  }
+
+  def remove(id: WorkflowId): Option[Unit] = {
+    val newWorkflowStore = workflowStore filterNot { _.id == id }
+
```
geoffjentry: Looking at this again I'm even less in love w/ it vs when i first did it. But this portion doesn't have access to the logger, yet using a Try seemed pointless since the akka logger doesn't take exceptions. Open to suggestions

core/src/main/resources/application.conf
```diff
@@ -52,8 +52,17 @@ system {
   // Max number of retries per job that the engine will attempt in case of a retryable failure received from the backend
   max-retries = 10
 
-  //If 'true' then when Cromwell starts up, it tries to restart incomplete workflows
+  // If 'true' then when Cromwell starts up, it tries to restart incomplete workflows
   workflow-restart = true
+
+  // Cromwell will cap the number of running workflows at N
```
geoffjentry: I don't love these names, if anyone else has a better one feel free to suggest

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
geoffjentry: also same comment about the names as in application.conf

core/src/main/resources/application.conf
mcovarr: `max-concurrent-workflows`?

core/src/main/resources/application.conf
```diff
@@ -52,8 +52,17 @@ system {
...
+  max-workflows-running-at-once = 5000
+
+  // Cromwell will launch up to N submitted workflows at a time, regardless of how many open workflow slots exist
+  max-workflows-to-launch-at-once = 50
```
mcovarr: `max-workflow-launch-count`?

geoffjentry:
@mcovarr @cjllanwarne This PR is now more-or-less in the state it was this morning, albeit with a lot of stuff changed as per your request. It's a lot different now so worth a re-look (unit tests are still borked, and this time it's sans centaur-ing as well, but eh)

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
Horneth: Can't the `remove` method be in the actor instead ? Or take a logger as a parameter ? Or better yet the `WorkflowStore` abstract class have a logger member that can b overridden in the actor ?

geoffjentry: - the actor can't directly manipulate the state, so it can't remove.
- That's a possibility which I've considered but it rubs me the wrong way as now this class is akka-y. One of the good things about separating all of the akka bits is that it makes the unit testing trivial. OTOH returning an Option/Try also rubs me the wrong way, although my preference was towards a sharp separation of stuff & actor
- This one is even more akka-y

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
mcovarr: I can't imagine a universally sane default for the max number of workflows to run, but I'm not sure if it's better to have the default be too low or too high.  For Firecloud this number might be fine but it's hard to imagine GOTC Cromwell being able to survive 5000 concurrent workflows (though I'd love to be proven wrong about that 😄 ).

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
```diff
@@ -0,0 +1,152 @@
...
+    */
+  def fetchRunnableWorkflows(n: Int): List[WorkflowToStart] = {
+    val startableWorkflows = workflowStore filter { _.state.isStartable } take n
+    val updatedWorkflows = startableWorkflows map { _.copy(state = Running) }
+    workflowStore = (workflowStore diff startableWorkflows) ++ updatedWorkflows
+
+    startableWorkflows map { _.toWorkflowToStart }
+  }
+
+  def remove(id: WorkflowId): Option[Unit] = {
```
mcovarr: Not sure I've ever seen something return `Option[Unit]`; it seems like this is conceptually returning a boolean.

engine/src/main/scala/cromwell/webservice/CromwellApiHandler.scala
```diff
@@ -89,32 +89,15 @@ class CromwellApiHandler(requestHandlerActor: ActorRef) extends Actor with Workf
...
-            case _: IllegalArgumentException => APIResponse.fail(e)
-            case _ => APIResponse.error(e)
-          })
-        case unexpected => Right(FailureResponse("error", s"unexpected message: $unexpected", None))
-      }
-      context.parent ! RequestComplete(StatusCodes.OK, requestResponse)
+    case ApiHandlerWorkflowSubmitBatch(sources) => requestHandlerActor ! WorkflowStoreActor.BatchSubmitWorkflows(sources)
+
+    case WorkflowStoreActor.WorkflowsBatchSubmitted(ids) =>
+      val responses = ids map { id => WorkflowSubmitResponse(id.toString, WorkflowSubmitted.toString) }
```
mcovarr: is `id` underscoreable?

mcovarr:
:+1: reaffirmed though it would be really nice to see green builds before merge 😄

geoffjentry:
@mcovarr you set such a high bar

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
geoffjentry: good point. 

Another possibility I've considered is just throwing an exception and catching it in the actor (followed by taking a shower)

I'm thinking the boolean is the best thing I've seen so far, it's not an unreasonable API and there are oodles of examples in the world that do something like that

engine/src/main/scala/cromwell/webservice/CromwellApiHandler.scala
geoffjentry: no

core/src/main/resources/application.conf
geoffjentry: sure why not

core/src/main/resources/application.conf
geoffjentry: sure why not

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
geoffjentry: I suppose that with them being configurable anyways it doesn't really matter. We can always fiddle with these if it becomes problematic for some reason

PR #1132, opened 2016-07-06
Spark Backend!

jainh:
Spark backend pull request, Which is capable of running dockerized Spark jobs on a standalone Spark cluster (Note: There is not yet any resource manager involve between Spark driver and worker i.e. Mesos or Yarn or etc.). The way we are submitting jobs is like other backend HtCondor ( wrapping commands in script and expect rc code).  Here goal is to show Cromwell able to submit/support containerized Spark  application through wdl.
NOTE: **This PR is just to add Spark backend, so it is not creating Spark command internally right now rather than accepting full command in wdl command, Next PR i will add Spark submit command through backend like in HtCondor.** 
Formal reviewers: @geoffjentry  @francares


geoffjentry:
@jainh Could you provide a higher level description of how this is working, specifically the Spark parts? In particular I'm picturing the target audience as someone who understands the concept of Cromwell & backends but isn't familiar with (nor cares) about particulars like the different actors. I'd like to be able to run the Spark-y parts by some folks here.

jainh:
@geoffjentry 
**Prerequisite**: We have Spark cluster  (3 nodes = 1 master, rest worker including master), Docker on all three nodes, Hdfs file system (3 nodes = 1 Namenode, rest Data nodes including NN) Spark app build locally in Docker repository.  This app (spark_hdfs.jar) has two main entry points 1) Word count 2) Vowel count as main class.  App consumes input available on Hdfs (assumption pre-loaded) and produces output to Hdfs. 

**WDL:**
```
task spark {
  command {
	/opt/spark/spark-1.6.1-bin-hadoop2.6/bin/spark-submit --class com.intel.spark.poc.nfs.SparkVowelLine --master spark://10.0.1.22:7077 /app/spark_hdfs.jar hdfs://10.0.1.22:8020/home/himanshuj/test/kinglear.txt hdfs://10.0.1.22:8020/home/himanshuj/output
  }

  output {
    File empty =  stdout()
  }

  runtime {
    docker: "sparkapp"
  }

}

workflow test {
  call spark
}

```
So the app is available inside the docker container that has base image with Spark environment(i.e. Spark driver + hadoop connector) that will connect to Spark master on host machine within the cluster to submit job, reads input from hdfs file system and turn them into RDDs and distribute the work to workers with the help of master and at the end write output to hdfs. 

Following are the arguments to the app 
`hdfs://10.0.1.22:8020/home/himanshuj/test/kinglear.txt hdfs://10.0.1.22:8020/home/himanshuj/output`

supportedBackends/spark/src/test/scala/cromwell/backend/impl/spark/SparkInitializationActorSpec.scala
```diff
@@ -0,0 +1,72 @@
...
+      inputs,
+      options
+    )
+  }
+
+  override def afterAll {
+    system.shutdown()
+  }
+
+  private def getHtCondorBackend(workflowDescriptor: BackendWorkflowDescriptor, calls: Seq[Call], conf: BackendConfigurationDescriptor) = {
```
mcovarr: naming

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkInitializationActor.scala
```diff
@@ -0,0 +1,54 @@
...
+import cromwell.backend.validation.RuntimeAttributesKeys._
+import cromwell.backend.impl.spark.SparkInitializationActor._
+import cromwell.backend.{BackendConfigurationDescriptor, BackendWorkflowDescriptor, BackendWorkflowInitializationActor}
+import wdl4s.types.{WdlBooleanType, WdlStringType}
+import wdl4s.{WdlExpression, Call}
+
+import scala.concurrent.Future
+
+object SparkInitializationActor {
+  val SupportedKeys = Set(DockerKey, FailOnStderrKey, ContinueOnReturnCodeKey)
```
francares: Do you need all these keys?

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkJobExecutionActor.scala
```diff
@@ -0,0 +1,161 @@
...
+    Try {
+      val inputFiles = localizedInputs.filter { case (k,v) => v.wdlType.equals(WdlFileType) }
+      val dockerInputDataVol: Seq[String] = inputFiles.values.map { value =>
+        val limit = value.valueString.lastIndexOf("/")
+        value.valueString.substring(0, limit)
+      } toSeq
+      val dockerCmd = "docker run -w %s %s %s --rm %s %s"
+      val dockerVolume = "-v %s:%s"
+      val dockerVolumeInputs = s"$dockerVolume:ro"
+      // `v.get` is safe below since we filtered the list earlier with only defined elements
```
francares: Remove this comment.

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkInitializationActor.scala
francares: Is there any key missing here?

kcibul:
hey @jainh -- Building on @geoffjentry 's question, and after reading through the code here... I'm not sure what about this PR is actually Spark'ified.  It seems like a copy and paste of the HtCondor backend (including some variable names!) and then using that to run a command inside a docker to connect to a pre-existing spark cluster with preloaded data.  All this could already be done using any one of the docker-based back ends (e.g. Local or even Google)

Am I missing something here?  If not, it would be great to see more of a roadmap about where you see this going because it's not clear from the code here what direction it could take.  The concept of running/orchestrating Spark jobs via Cromwell is an important one and could take many, many forms so clarifying which approach this is taking (and making that clear in naming/location/etc) is important.

Thanks

jainh:
@kcibul 
I am missing few things in the implementation like the way to construct Spark command inside Spark backend for this PR, Let me add that to make it Spark'ified. It should be same like Htcondor but entertain Runtime attributes like executor-memory, cores and configuration based master to execute Spark job both in Dockerized or non-Dockerized mode. 
So let me close this pull request and create a new one later. My goal is to create Spark backend to run Spark jobs in standalone cluster mode for this PR, then later to add other Resource manager related support.

Issue #794, opened 2016-05-05
Verify that cache hits no longer cause cromwell to seemingly lock up


jsotobroad:
We have a workflow that scatters 95 wide as its first step.  When running with `read_from_cache=false` the workflow chugs along normally.  When we allow the same workflow to read from the cache, Cromwell seems to lock up after the first handful of cache hits(~30).  Cromwell will stop responding to api requests and after some time with logs being written the workflow that was getting the cache hits will hit 503 and timeout errors.  When running the workflow with `read_from_cache=false` we run into none of these errors.

Timeout Error

```
2016-05-05 17:37:02,285 cromwell-system-akka.actor.default-dispatcher-25 WARN  - Configured registration timeout of 1 second expired, stoppingw
```

503 Error

```
Exception occurred while attempting to copy outputs from gs://broad-gotc-dev-storage/cromwell_execution/JointGenotyping/ccba2c79-c998-4f03-b736-af097391db66/call-SplitGvcf/shard-50 to gs://broad-gotc-dev-storage/cromwell_execution/JointGenotyping/7164dc88-af61-4ea6-8a73-f0b79594ae9a/call-SplitGvcf/shard-50

com.google.api.client.googleapis.json.GoogleJsonResponseException: 503 Service Unavailable

{

  "code" : 503,

  "errors" : [ {

    "domain" : "global",

    "message" : "Backend Error",

    "reason" : "backendError"

  } ],

  "message" : "Backend Error"

}

at com.google.api.client.googleapis.json.GoogleJsonResponseException.from(GoogleJsonResponseException.java:145) ~[cromwell.jar:0.19]

at com.google.api.client.googleapis.services.json.AbstractGoogleJsonClientRequest.newExceptionOnError(AbstractGoogleJsonClientRequest.java:113) ~[cromwell.jar:0.19]

at com.google.api.client.googleapis.services.json.AbstractGoogleJsonClientRequest.newExceptionOnError(AbstractGoogleJsonClientRequest.java:40) ~[cromwell.jar:0.19]

at com.google.api.client.googleapis.services.AbstractGoogleClientRequest$1.interceptResponse(AbstractGoogleClientRequest.java:321) ~[cromwell.jar:0.19]

at com.google.api.client.http.HttpRequest.execute(HttpRequest.java:1056) ~[cromwell.jar:0.19]

[...]

PR #1133, opened 2016-07-07
Return Spray timeouts as Json Closes #947

### LINKED ISSUE #947: Response timeout does not return Content-Type: `application/json`
It would be easier to consume errors if they were wrapped in JSON. I believe 400 errors are already represented this way already, but a 500 response timeout returns with Content-Type:`text/plain`.

Horneth:


geoffjentry:
I know green requested this so they'll be happy but should also let blue know as well

kshakir:
I haven't seen the actual JSON response, but code LGTM 👍

mcovarr:
Per @vivster7 the bug report was against 0.19 hotfix, not develop.  Not that this shouldn't be fixed here too. 😄 

@kcibul It seems #794 is also labeled 0.21 but actually applies to 0.19 hotfix.

vivster7:
^ I would add that its not terribly important that it get fixed in hotfix, if thats what you're suggesting. Just that it gets fixed in some future version of cromwell.

mcovarr:
Thanks @vivster7.  Do we think Blue also shares the view that this can wait for PBE?

kcibul:
Let's keep this just on 0.20+ (PBE).  As for #794, I commented on that ticket, thanks for the heads up

mcovarr:
:+1:

PR #1141, opened 2016-07-08
Publishing cromwell artifacts via travis. Closes #1019 Closes #645 Closes #644

### LINKED ISSUE #644: Docker's build fails when cromwell used as submodule
During the cromwell image build, the Dockerfile first `ADD`s the cromwell directory to a new image, then the install.sh invokes `sbt … assembly`. If cromwell was actually a submodule, then this cromwell directory has a .git file instead of a directory, that fails with sbt-git 0.8.5.

sbt-git 0.7.1 is the last version that appears to work with:
- .git as a directory
- .git as a detached file (cromwell as a git submodule, try `docker build .`)
- .git missing

For example, though this appears to be fixed in 0.7.1, when there's no .git at all, `sbt compile` crashes in 0.7.0 with:

> java.lang.IllegalArgumentException: One of setGitDir or setWorkTree must be called.

The latest sbt-git as of this comment, 0.8.5, doesn't like it when the .git file points to a non-existent git config. To test, temporarily change .git to a file with the contents "gitdir: ../.git/modules/cromwell".
`sbt compile` crashes with:

> org.eclipse.jgit.errors.NoWorkTreeException: Bare Repository has neither a working tree, nor an index

Because we're not actually versioning snapshot releases of cromwell currently, the fix recommended to our GreenFriends was to excise the sbt-git plugin via:

```bash
grep -v sbt-git project/plugins.sbt > tmp && mv tmp project/plugins.sbt
```

### LINKED ISSUE #645: Unable to differentiate cromwell snapshots from release jars
We include the sbt-git plugin, but don't use it to generate version strings with the git hash. If we are going to encourage beta testers to use early snapshots of the develop branch, it would seriously help debugging to be able to get the version number from inside the jar file.

NOTE: GreenFriends are checking out cromwell as a .git submodule. By the time docker runs `sbt … assembly` the git hash information may be unavailable. See #644 for a bit more info on what's going on. We may need to fix our docker build to work if git is a detached submodule. Or communicate GreenFriends that this is unsupported?

Ideally `sbt … assembly` should be able to use sbt-git to automatically determine if a build is a release simply by looking for a git tag, say "0.19". This functionality is NOT implemented currently in cromwell libraries, as lenthall, wdl4s, etc. all use Travis flags to label release builds instead of just git. Instead of only using sbt-git in cromwell, we could also create a dynamic Dockerfile and override the version by passing it in externally. For more info on dynamic dockerfiles, ask our GreenFriends about their `build_docker.sh` scripts.

### LINKED ISSUE #1019: Add backend jar to artifactory
Verify that the backend jar is fat, otherwise the components will need to be up there too

kshakir:
Single version across all artifacts in the repo/sbt aggregate project, published together and referring to each other.
Publishing releases and snapshots to Broad artifactory using lenthall's build bash ported to sbt, appropriate for publishing our multi-module, multi-build-type setup.
Stop wasting time assembling fat jars for artifacts _except_ the root.
Renamed Travis' build types from centaur/test to centaur/sbt.
Split up and thinned compile and test dependency statements, artifacts sometimes depend only on another artifacts tests, but not the same artifact's main code.
Just as many (most?) artifacts depend on other artifacts, rearranged dependencies in a similar way. Test dependencies are now embedded in the artifacts that use them. So jars like spray-testkit are no longer added to the classpath of the gcs filesystem, for example.
Centaur build in Travis now prints a heartbeat '…' every 60 seconds, instead of using the log destroying `travis_wait`.
No longer echoing line for line most bash scripts via `set -x`.
Fixed scaladoc in `SlickDatabase`.
Reverted `sbt-git` back to `0.7.1` due to issues with detached git submodules.


geoffjentry:
you are a legend and my life s aim is to _8dec8c339d5974bfd36ddea1261bf802

src/bin/travis/testSbt.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
geoffjentry: FWIW my use of `/bin/bash` previously was intentional as all of the Travis examples do it that way - I just assumed it was a Travis best practice. OTOH presumably this works in Travis so eh

geoffjentry:
How many thumbs can a man give? 

👍

src/bin/travis/testSbt.sh
kshakir: Docs may be inconsistent, but this is the one I was working off:

> When using custom scripts they should be executable (for example, using chmod +x) and contain a valid shebang line such as /usr/bin/env sh, /usr/bin/env ruby, or /usr/bin/env python.

https://docs.travis-ci.com/user/customizing-the-build/#Customizing-the-Installation-Step

geoffjentry: works for me. I know that `/bin/bash` isn't generally considered the best thing to do so I just assumed the docs I saw were doing things for a good reason ... however I realize that generally things are the way they are just because that's how they are :)

mcovarr:
:+1:

kshakir:
Got over aggressive with last minute dependency thinning, removing some reflective dependencies for tests, but the concept is what you see there. Fixing locally, will confirm via travis, then merge.

Horneth:
You already got 2, but 👍 , because I can.

PR #1147, opened 2016-07-11
Added spray-routing-shapeless2 dependency to be able to work with Lens. Closes #1148

### LINKED ISSUE #1148: Add spray-routing-shapeless2 dependency to be able to work with Lens.

francares:
Minor change in order to get Spray and Lens working together.

Formal reviewers: @geoffjentry and @kshakir.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -20,7 +20,7 @@ object Dependencies {
 
   val sprayDependencies = List(
     "io.spray" %% "spray-can" % sprayV,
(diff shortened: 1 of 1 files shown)


kshakir:
Ugh. Good catch. 👍 

For others that are curious: http://spray.io/project-info/current-versions/#shapeless-versions (and the whole page in general regarding important version syncing)

geoffjentry:
👍 

This was talked about a lot at the typelevel summit in Philly. At some point we'll need to ditch Spray but that time isn't now :)

PR #1151, opened 2016-07-12
RefreshTokenModeSpec closes #1081

### LINKED ISSUE #1081: Assess/Fix/Implement JesBackendSpec should "be verified when localizing with Refresh Token"
This is an extension of #1026, that will not fix this specific test.

If mocking is not possible that it may require that an actual refresh token is used?

ruchim:
The only tests remaining in JesBackendSpec were all related to RefreshToken behavior, so just renaming the spec to reflect what it's testing. The tests in this spec check if workflow options are incomplete and that the refresh token is decrypted when retrieving it.

### DIFF
--- a/engine/src/test/scala/cromwell/engine/backend/jes/RefreshTokenModeSpec.scala
+++ b/engine/src/test/scala/cromwell/engine/backend/jes/RefreshTokenModeSpec.scala
@@ -1,24 +1,13 @@
 package cromwell.engine.backend.jes
 
-import cromwell.CromwellTestkitSpec
-import cromwell.backend.impl.jes.io.{DiskType, JesWorkingDisk}
-import cromwell.core.WorkflowOptions
+import cromwell.backend.BackendSpec._
+import cromwell.core.{TestKitSuite, WorkflowOptions}
 import cromwell.engine.backend.EnhancedWorkflowOptions._
-import cromwell.engine.workflow.WorkflowDescriptorBuilder
 import cromwell.filesystems.gcs._
 import cromwell.util.EncryptionSpec
-import org.scalatest.{BeforeAndAfterAll, FlatSpec, Matchers}
-import org.specs2.mock.Mockito
+import org.scalatest.FlatSpecLike
 
-class RefreshTokenModeSpec extends FlatSpec with Matchers with Mockito with BeforeAndAfterAll with WorkflowDescriptorBuilder {
-  val testWorkflowManagerSystem = new CromwellTestkitSpec.TestWorkflowManagerSystem()
-  override implicit val actorSystem = testWorkflowManagerSystem.actorSystem
-  val workingDisk = JesWorkingDisk(DiskType.SSD, 200)
-
-  override protected def afterAll() = {
-    testWorkflowManagerSystem.shutdownTestActorSystem()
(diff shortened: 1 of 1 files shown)


engine/src/test/scala/cromwell/engine/backend/jes/RefreshTokenModeSpec.scala
```diff
@@ -37,4 +35,12 @@ class JesBackendSpec extends FlatSpec with Matchers with Mockito with BeforeAndA
       } should have message s"Missing parameters in workflow options: refresh_token"
     }
   }
+
+  "refresh token value" should "be decrypted when asserting workflow options" in {
+    EncryptionSpec.assumeAes256Cbc()
+
+    val goodOptions = WorkflowOptions.fromMap(Map("refresh_token" -> mockToken)).get
+    refreshToken.assertWorkflowOptions(goodOptions.toGoogleAuthOptions)
+    goodOptions.toGoogleAuthOptions.get("refresh_token").get shouldBe s"$mockToken"
```
Horneth: could just be `shouldBe mockToken` ?

kshakir:
`cromwell.engine.backend.jes` isn't the right package (I have no clue what the other files are still doing in this folder). Please move this to `cromwell.backend.impl.jes`.

engine/src/test/scala/cromwell/engine/backend/jes/RefreshTokenModeSpec.scala
```diff
@@ -10,7 +10,7 @@ import cromwell.util.EncryptionSpec
 import org.scalatest.{BeforeAndAfterAll, FlatSpec, Matchers}
 import org.specs2.mock.Mockito
 
-class JesBackendSpec extends FlatSpec with Matchers with Mockito with BeforeAndAfterAll with WorkflowDescriptorBuilder {
+class RefreshTokenModeSpec extends FlatSpec with Matchers with Mockito with BeforeAndAfterAll with WorkflowDescriptorBuilder {
   val testWorkflowManagerSystem = new CromwellTestkitSpec.TestWorkflowManagerSystem()
```
kshakir: 1) We no longer need to spin up a whole workflow manager system if we just need an actor system. There's now [`TestKitSuite`](https://github.com/broadinstitute/cromwell/blob/6af5611c27c678bda2b55950e1a3e37c1c766462/core/src/test/scala/cromwell/core/TestKitSuite.scala) for that.

2) This spec doesn't even use the thread pool spun up inside the WMS / actor system. So remove the `testWorkflowManagerSystem` and related cruft, delete all usage of the `BeforeAndAfterAll` trait, and while there I don't think `Mockito` and `workingDisk` are used either.

kshakir:
Specs look good. Some vestigial bits from reusing old code, but overall 👍

Horneth:
👍

engine/src/test/scala/cromwell/engine/backend/jes/RefreshTokenModeSpec.scala
ruchim: @kshakir I didn't realize the TestKitSuite was so lightweight and easy to use, thanks. TIL

PR #1152, opened 2016-07-12
Create dispatchers for api / engine / io Closes #1015

### LINKED ISSUE #1015: Submitting/Checking Status/Metadata for workflows shouldn't be starved out by running workflows
We should have better thread pool separation between the running workflows and all the other REST API endpoints.  Specifically submitting, metadata and status shouldn't be starved out because a lot of workflows are running

Horneth:
Create 2 new dispatchers: `api-dispatcher` and `engine-dispatcher`.
`api-dispatcher` isolates all actors / futures to ensure that every request will be using resources in this thread pool.
`engine-dispatcher` isolates engine actors from the rest of the system (specifically the backends that can run an arbitrary amount of blocking/slow code out of the control of the engine). 
`slow-actor-dispatcher` has been renamed to `io-dispatcher` and is used for io / blocking operations (copying outputs / logs / uploading files to gcs etc..)

Other side quests:

- Re-use the WorkflowManagerSystem created in Main in `CromwellServer`, instead of extending  `CromwellServer` with `WorkflowManagerSystem` which duplicates everything a second time.

- Add an `afterAll` on `CromwellTestKitSpec` to shutdown the test actor system at the end of the spec.


engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
```diff
@@ -78,37 +78,35 @@ abstract class WorkflowStore {
...
    */
 
-  private val logger = Logging(context.system, this)
-
   override def receive = {
     case SubmitWorkflow(source) =>
       val id = add(NonEmptyList(source)).head
       registerIdWithMetadataService(id)
       sender ! WorkflowSubmittedToStore(id)
+      log.info(s"Workflow $id submitted.")
```
geoffjentry: These get really spammy which is why I didn't include them. IMO they should be debug if they exist

geoffjentry:
@Horneth In general, how did you test/benchmark this stuff?

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
Horneth: Compared to other things we have I find this pretty useful though. Without this there's no way to know when looking at the logs if the workflow was submitted except waiting for it to be picked up by the next sweep (at best).

Horneth:
I flooded the Local backend with scattered sleeps and tried to access metadata, status, submit workflows etc...
I'd like to try more benchmarky things though

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
geoffjentry: Think about it on firecloud scale though. The cromwell logs in production systems already overwhelm our logging servers and need to be heavily debulked as it is

mcovarr:
Just a heads up that the changes incoming on `mlc_existence` as currently structured move the execution of workflow-ID-dependent APIs (which is most of them) from `CromwellApiService` to `EngineMetadataServiceActor`.

geoffjentry:
@mcovarr I was just thinking that this PR probably warrants careful timing & some coordination to avoid a giant rebase hell for everyone

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
geoffjentry: Discussed in person, will leave as is until one of the GAWBs comes and beats us with the combined printout of their logs or something like that

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/JobPreparationActor.scala
```diff
@@ -25,7 +25,9 @@ object JobPreparationActor {
             jobKey: BackendJobDescriptorKey,
             factory: BackendLifecycleActorFactory,
             initializationData: Option[BackendInitializationData]) = {
-    Props(new JobPreparationActor(executionData, jobKey, factory, initializationData)).withDispatcher("akka.dispatchers.slow-actor-dispatcher")
+    // Note that JobPreparationActor doesn't run on the engine dispatcher as it mostly executes backend-side code
+    // (WDL expression evaluation using Backend's expressionLanguageFunctions)
+    Props(new JobPreparationActor(executionData, jobKey, factory, initializationData))
```
kcibul: what dispatcher is it running on then?

Horneth: The default one (same as backends)
This is an artifact of the way wdl expressions are evaluated, which is not ideal but is another discussion (which I'd actually like to make my developer's choice issue). 
At least it works well so it's not a big issue IMO, it's just not as clean as it could be.

kcibul: sgtm

Horneth:
assigned @mcovarr @geoffjentry for review since you've been looking at this already

core/src/main/resources/application.conf
```diff
@@ -26,12 +26,30 @@ akka {
...
+
+    # A dispatcher for engine actors
+    # Because backends behaviour is unpredictable (potentially blocking, slow) the engine runs
+    # on its own dispatcher to prevent backends from affecting its performance.
+    engine-dispatcher {
+      type = Dispatcher
+      executor = "fork-join-executor"
+    }
+
+    # Note that without further configuration, backend actors run on the default dispatcher
```
geoffjentry: We talked about making something for our own backends, were you going to do that as part of this PR or leave it for a future person?

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
```diff
@@ -76,37 +76,35 @@ abstract class WorkflowStore {
   }
 
   def remove(id: WorkflowId): Boolean = {
-    val newWorkflowStore = workflowStore filterNot { _.id == id }
-
-    if (newWorkflowStore == workflowStore) {
-      false
-    } else {
-      workflowStore = newWorkflowStore
+    if (workflowStore.exists(_.id == id)) {
```
geoffjentry: Don't really agree w/ this change although since this logic will be db-ized imminently I suppose it doesn't matter either way

geoffjentry:
👍

mcovarr:
:+1: with existence cache removal

engine/src/main/scala/cromwell/services/metadata/EngineMetadataServiceActor.scala
```diff
@@ -0,0 +1,85 @@
...
+  implicit val ec = context.dispatcher
+
+  self ! RefreshSummary
+
+  // Status lookups are eventually consistent, so it's possible a db status lookup may fail for a recently submitted
+  // workflow ID.  This cache records workflow IDs for which metadata has recently flowed through this actor.  If a db
+  // status lookup fails, this actor consults the cache to see if the queried ID is known.  If the queried ID is known a
+  // status query will return `Submitted`, otherwise the status lookup will fail.  This cache is only consulted if the
+  // db status lookup fails.
+  private var workflowExistenceCache: Map[WorkflowId, Int] = Map.empty
```
Horneth: Oops

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
Horneth: It just seemed unnecessary to `val newWorkflowStore = workflowStore filterNot { _.id == id }` and compare it to the old one just to check if `id` was indeed inside the workflowStore.

core/src/main/resources/application.conf
Horneth: Oh yeah I can do it here

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
geoffjentry: In the worst case (the element exists but is the last one in the list) this is making two full scans of the list. If it doesn't exist it's the same amount of work (one full scan). In the best case (the match is the first element) it's making a scan + 1. 

Aside from my DB-izing comment this is small potatoes compared to how often we multi-scan collections all over the place which is another reason I don't really care, although that was my thinking when I did it that way in the first place, since the difference was the same LOC anyways

Horneth: Yeah I was more thinking in terms of memory used when the list is duplicated unnecessarily. But we probably already spent too much time writing about this than it's worth xD I'll revert it and it will be modified when the DB comes in anyway like you said.

geoffjentry: don't worry about reverting

PR #1153, opened 2016-07-12
Firecloud Auth

Horneth:
Fix firecloud auth and pass the whole `JesWorkflowPaths` around in the `BackendInitializationData`

### DIFF
--- a/supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesCallPaths.scala
+++ b/supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesCallPaths.scala
@@ -13,7 +13,7 @@ object JesCallPaths {
 }
 
 class JesCallPaths(jobKey: BackendJobDescriptorKey, workflowDescriptor: BackendWorkflowDescriptor, jesConfiguration: JesConfiguration,
-                   val gcsFileSystem: GcsFileSystem) extends JesWorkflowPaths(workflowDescriptor, jesConfiguration, gcsFileSystem) {
+                   gcsFileSystem: GcsFileSystem) extends JesWorkflowPaths(workflowDescriptor, jesConfiguration, gcsFileSystem) {
 
   val CallPrefix = "call"
   val ShardPrefix = "shard"
--- a/supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesWorkflowPaths.scala
+++ b/supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesWorkflowPaths.scala
@@ -2,11 +2,11 @@ package cromwell.backend.impl.jes
 
 import java.nio.file.Path
 
-import cromwell.backend.BackendWorkflowDescriptor
-import cromwell.backend.impl.jes.io._
+import cromwell.backend.{BackendJobDescriptorKey, BackendWorkflowDescriptor}
 import cromwell.core.WorkflowOptions
+import cromwell.core.WorkflowOptions.FinalCallLogsDir
 import cromwell.filesystems.gcs.GoogleAuthMode.GoogleAuthOptions
-import cromwell.filesystems.gcs.{GoogleAuthMode, GcsFileSystemProvider, GcsFileSystem}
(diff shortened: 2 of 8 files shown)


supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesWorkflowPaths.scala
```diff
@@ -13,22 +17,36 @@ object JesWorkflowPaths {
...
 
-class JesWorkflowPaths(workflowDescriptor: BackendWorkflowDescriptor, jesConfiguration: JesConfiguration, gcsFileSystem: GcsFileSystem) {
+  implicit class GoogleAuthWorkflowOptions(val workflowOptions: WorkflowOptions) extends AnyVal {
+    def toGoogleAuthOptions: GoogleAuthMode.GoogleAuthOptions = new GoogleAuthOptions {
+      override def get(key: String): Try[String] = workflowOptions.get(key)
+    }
+  }
+}
 
+class JesWorkflowPaths(workflowDescriptor: BackendWorkflowDescriptor, jesConfiguration: JesConfiguration, val gcsFileSystem: GcsFileSystem) {
```
mcovarr: Maybe `gcsFileSystem` should get a more specific name to prevent the confusion that led to the original bug?  `gcsFileSystemWithUserAuth` or something?

Horneth: Good point

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesWorkflowPaths.scala
```diff
@@ -13,22 +17,36 @@ object JesWorkflowPaths {
   def apply(workflowDescriptor: BackendWorkflowDescriptor, jesConfiguration: JesConfiguration, gcsFileSystem: GcsFileSystem) = {
     new JesWorkflowPaths(workflowDescriptor, jesConfiguration, gcsFileSystem)
   }
-}
 
-class JesWorkflowPaths(workflowDescriptor: BackendWorkflowDescriptor, jesConfiguration: JesConfiguration, gcsFileSystem: GcsFileSystem) {
+  implicit class GoogleAuthWorkflowOptions(val workflowOptions: WorkflowOptions) extends AnyVal {
```
mcovarr: Can't this use the copy of this implicit class in the package object?  We have three copies of this already... 😢

mcovarr:
:+1:

geoffjentry:
@Horneth This is what you had to do to unstuck @helgridly, right?

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesCallPaths.scala
```diff
@@ -13,7 +13,7 @@ object JesCallPaths {
 }
 
 class JesCallPaths(jobKey: BackendJobDescriptorKey, workflowDescriptor: BackendWorkflowDescriptor, jesConfiguration: JesConfiguration,
-                   val gcsFileSystem: GcsFileSystem) extends JesWorkflowPaths(workflowDescriptor, jesConfiguration, gcsFileSystem) {
+                   gcsFileSystem: GcsFileSystem) extends JesWorkflowPaths(workflowDescriptor, jesConfiguration, gcsFileSystem) {
```
geoffjentry: Not a quibble, just curious why the change

geoffjentry:
👍

Horneth:
@geoffjentry correct

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesCallPaths.scala
Horneth: Because I made it a `val` in `JesWorkflowPaths`, which `JesCallPaths` extends from, so there was no need to have it here (I actually think I remember getting an error for not overriding it or something).

PR #1154, opened 2016-07-12
Workflow ID validation + Outputs/Logs non-404ing.  Closes #1138

### LINKED ISSUE #1138: Outputs and logs return 404 if empty for a workflow
These endpoints return a 404 workflow not found if there is no data to return, as opposed to metadata or status which return empty values.

This is a) confusing due to not being self-consistent and b) confusing due to the workflow actually existing

mcovarr:


database/src/main/scala/cromwell/database/slick/MetadataComponent.scala
```diff
@@ -82,6 +82,14 @@ trait MetadataComponent {
     } yield m
   )
 
+  val workflowExists = Compiled(
+    (workflowExecutionUuid: Rep[String]) => (for {
+      m <- metadata
+      if m.workflowExecutionUuid === workflowExecutionUuid
+      if m.key === WorkflowMetadataKeys.SubmissionTime
```
kshakir: This is _ok_, but can you explain why you need to check for a key at all? Shouldn't a row in the database with the UUID (line 88) be enough to verify existence?

engine/src/main/scala/cromwell/services/EngineMetadataServiceActor.scala
```diff
@@ -86,29 +91,21 @@ case class EngineMetadataServiceActor(serviceConfig: Config, globalConfig: Confi
...
 
   private def queryLogsAndRespond(id: WorkflowId): Unit = {
     val replyTo = sender()
     dataAccess.queryLogs(id) onComplete {
-      case Success(l) if l.isEmpty =>
-        self ! HandleNotFound(id, replyTo)
-      case Success(l) =>
-        replyTo ! LogsResponse(id, l)
-
+      case Success(l) => replyTo ! LogsResponse(id, l)
```
kshakir: Not a fan of single letters again. "Is that `Success(1)` or `Success(l)`??"

database/src/main/scala/cromwell/database/slick/MetadataComponent.scala
mcovarr: Oh right.  I didn't originally have the `exists` on this so I was doing a clunky `headOption.isDefined` in the caller, but you're absolutely right this doesn't need to consider the key now.  That might also clean up the `CromwellApiServiceSpec` test a bit.

engine/src/main/scala/cromwell/webservice/PerRequest.scala
```diff
@@ -115,5 +117,5 @@ object PerRequestCreator {
     This is yucky. For lack of a better idea on how to name the individual endpoint actors I've shamelessly stolen
     what Agora is doing. I believe the Monsanto library will clean this up but we're not using it yet
    */
-  def endpointActorName = "Endpoint-" + java.lang.Thread.currentThread.getStackTrace()(1).getMethodName + System.nanoTime()
+  def endpointActorName = List("Endpoint", java.lang.Thread.currentThread.getStackTrace()(1).getMethodName, UUID.randomUUID()).mkString("-")
```
kshakir: If you're abandoning/modifying the Agora approach, please update the comment.

engine/src/main/scala/cromwell/services/EngineMetadataServiceActor.scala
geoffjentry: That's more a complaint about font than single letter.

engine/src/main/scala/cromwell/webservice/PerRequest.scala
mcovarr: Yeah ok.  I actually had to change this as I was getting actor name collisions somehow.  😦

engine/src/main/scala/cromwell/services/EngineMetadataServiceActor.scala
mcovarr: yeah they're almost indistinguishable on El Capitan. 😄

services/src/main/scala/cromwell/services/MetadataServiceActor.scala
```diff
@@ -41,7 +43,11 @@ object MetadataServiceActor {
   case class WorkflowOutputs(workflowId: WorkflowId) extends MetadataServiceAction
   case class GetLogs(workflowId: WorkflowId) extends MetadataServiceAction
   case object RefreshSummary extends MetadataServiceAction
-  final case class HandleNotFound(workflowId: WorkflowId, sndr: ActorRef) extends MetadataServiceAction
+  final case class ValidateWorkflowIdAndExecute(uuid: UUID,
+                                                requestContext: RequestContext,
+                                                onRecognized: WorkflowId => Route,
+                                                onUnrecognized: UUID => Route,
+                                                onFailure: (UUID, Throwable) => Route) extends MetadataServiceAction
```
kshakir: Two things related to loving the distinction of `unrecognized` vs `malformed` back in the web service:

1) Have you considered using a single trait instead of three function parameters? If this is your preference, then that's fine.

2) What about changing `uuid: UUID` to `possibleWorkflowId: String`, and adding to the class a `def workflowId`. This would require a fourth `onMalformed: (String, Throwable) => Route` function, that made me think of "1)" above.

There's something about UUID imports leaking around, and what seems to be even the tiniest biz logic up in the web services that feels funny to me. This `ValidateWorkflowIdAndExecute` is already doing other validation, so I'm wondering if it could also start validation with the string.

kshakir:
Looking good. Going to give this a 👍  overall

services/src/main/scala/cromwell/services/MetadataServiceActor.scala
mcovarr: Yeah I like both of these ideas, I'll give them a try.

engine/src/main/scala/cromwell/webservice/metadata/MetadataBuilderActor.scala
```diff
@@ -233,8 +230,11 @@ class MetadataBuilderActor(serviceRegistryActor: ActorRef) extends LoggingFSM[Me
     case Event(failure: WorkflowQueryFailure, _) =>
       context.parent ! RequestComplete(StatusCodes.BadRequest, APIResponse.fail(failure.reason))
       allDone
-    case Event(WorkflowOutputsResponse(w, metadata), _) =>
-      context.parent ! RequestComplete(StatusCodes.OK, workflowMetadataResponse(w, metadata, includeCallsIfEmpty = false))
+    case Event(WorkflowOutputsResponse(id, events), _) =>
+      // Add in an empty output event if there aren't already any output events.
+      val hasOutputs = events collectFirst { case MetadataEvent(MetadataKey(_, _, key), _, _) if key == WorkflowMetadataKeys.Outputs => () } isDefined
+      val updatedEvents = if (hasOutputs) events else MetadataEvent.empty(MetadataKey(id, None, WorkflowMetadataKeys.Outputs)) +: events
```
Horneth: I thought that the WorkflowStore was publishing an empty output event on workflow submission regardless 👍 https://github.com/broadinstitute/cromwell/blob/develop/engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala#L131

mcovarr: Yes it does, but in case that event isn't in the journal this makes extra sure that outputs are always returned as an empty block if there are no outputs.  The test data I was working with happened to fit this pattern.

Horneth: Ok, if we start having a lot of those "mandatory even if empty" fields in the response we might want to have some kind of template with all of them (either Json or as MetadataEvents), and add the actual events on top of that. But that's not for this PR

Horneth:
👍

engine/src/main/scala/cromwell/webservice/PerRequest.scala
```diff
@@ -106,14 +108,11 @@ trait PerRequestCreator {
...
 
 object PerRequestCreator {
-  /*
-    This is yucky. For lack of a better idea on how to name the individual endpoint actors I've shamelessly stolen
-    what Agora is doing. I believe the Monsanto library will clean this up but we're not using it yet
-   */
-  def endpointActorName = "Endpoint-" + java.lang.Thread.currentThread.getStackTrace()(1).getMethodName + System.nanoTime()
-}
\ No newline at end of file
+  // This scheme was changed away from the Agora System.nanoTime approach due to actor naming collisions (!)
```
geoffjentry: Any idea how that's possible? Or is it perhaps that nanos isn't really as granular as one would like?

Also it's nice to get rid of that comment regarding a library that clearly hasn't happened after almost a year ;)

mcovarr: This matches up with my intuition at least.  tl;dr implementations of API may be less granular than they appear 

http://stackoverflow.com/questions/351565/system-currenttimemillis-vs-system-nanotime

PR #1186, opened 2016-07-19
Filesystem based WorkflowStore which resubmits previously running workflows. Closes #1118

### LINKED ISSUE #1118: Use WorkflowStore to support restart

cjllanwarne:


geoffjentry:
I don't understand why this is fs based, why not wait a day or two and do it with a db?

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
```diff
@@ -48,70 +48,157 @@ object WorkflowStore {
...
+      val work = store.remove(id) map { removed =>
+        if (removed) {
+          log.info("Workflow {} removed from store successfully.", id)
+        } else {
+          log.warning(s"Attempted to remove ID {} from the WorkflowStore but it didn't exist", id)
+        }
+      }
+      addWorkCompletionHooks(cmd, work)
+      goto(Working) using stateData.copy(currentOperation = Option(cmd))
+    case Event(InitializerCommand, _) =>
```
mcovarr: Should `Initializing` be its own state?

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
```diff
@@ -48,70 +48,157 @@ object WorkflowStore {
...
+    case fromState -> toState =>
+      log.info("WorkflowStore moving from {} to {} using {}", fromState, toState, stateData)
+  }
+
+  private def addWorkCompletionHooks[A](command: WorkflowStoreActorCommand, work: Future[A]) = {
+    work.onSuccess { case _ => self ! WorkDone }
+    work.onFailure { case t: Throwable =>
+      log.error("Error occurred during {}: {} because {}", command.getClass.getSimpleName, t.toString, t.getStackTrace.toList.mkString(System.lineSeparator()))
+      self ! WorkDone
+    }
```
mcovarr: `onComplete` as these two methods are going to be deprecated in 2.12

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
```diff
@@ -48,70 +48,157 @@ object WorkflowStore {
...
+      val sndr = sender
+      val work = store.add(NonEmptyList(source)) map { ids =>
+        val id = ids.head
+        registerIdWithMetadataService(id)
+        sndr ! WorkflowSubmittedToStore(id)
+        log.info("Workflow {} submitted.", id)
+      }
+      addWorkCompletionHooks(cmd, work)
+      goto(Working) using stateData.copy(currentOperation = Option(cmd))
+    case Event(cmd @ BatchSubmitWorkflows(sources), _) =>
```
mcovarr: this case and the preceding feel like they could be collapsed with a judiciously chosen common trait

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
```diff
@@ -48,70 +48,157 @@ object WorkflowStore {
...
+      val sndr = sender
+      val work = store.add(sources) map { ids =>
+        val id = ids.head
+        ids foreach registerIdWithMetadataService
+        sndr ! WorkflowsBatchSubmittedToStore(ids)
+        log.info("Workflows {} submitted.", ids.list.mkString(", "))
+      }
```
mcovarr: it looks like the unhappy paths on the store interactions are generally not being handled

engine/src/main/scala/cromwell/engine/workflow/FileSystemWorkflowStoreDatabase.scala
```diff
@@ -0,0 +1,95 @@
...
+  val fsroot= Paths.get("WorkflowStore").toAbsolutePath
+  val submittedPath = Paths.get(s"$fsroot${separator}submitted").toAbsolutePath
+  val restartablePath = Paths.get(s"$fsroot${separator}restartable").toAbsolutePath
+  val runningPath = Paths.get(s"$fsroot${separator}running").toAbsolutePath
+
+  def workflowPath(base: Path, id: WorkflowId): Path = Paths.get(s"$base$separator$id")
+
+  def wdlFilePath(base: Path) = Paths.get(s"${base.toAbsolutePath}${separator}workflow.wdl")
+  def inputsFilePath(base: Path) = Paths.get(s"${base.toAbsolutePath}${separator}inputs.json")
+  def optionsFilePath(base: Path) = Paths.get(s"${base.toAbsolutePath}${separator}options.json")
```
mcovarr: a lot of things could be DRYer here, and IntelliJ is going to have some yellowing opinions about those `${separator}`s

engine/src/main/scala/cromwell/engine/workflow/FileSystemWorkflowStoreDatabase.scala
```diff
@@ -0,0 +1,95 @@
...
+
+      // Write out the files:
+      FileUtils.writeStringToFile(wdlFilePath(newWorkflowPath).toFile, source.wdlSource.toString)
+      FileUtils.writeStringToFile(inputsFilePath(newWorkflowPath).toFile, source.inputsJson.toString)
+      FileUtils.writeStringToFile(optionsFilePath(newWorkflowPath).toFile, source.workflowOptionsJson.toString)
+      id
+    }}
+
+    // This is horrible because Future.sequence doesn't work with NonEmptyList. Shrug. I assure you it's ok because there's a Nel on the input
+    Future.sequence(nelOfFutures.list) map { list => NonEmptyList.nel(list.head, list.tail) }
```
mcovarr: yeah too bad this doesn't magically work 😢

mcovarr:
:+1:

geoffjentry:
Disregard my previous comment, can't delete on phone. I would say to make sure zero time is spent reviewing the fs pet though

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
cjllanwarne: The `work` is passed to the `addWorkCompletionHooks` method which adds the success and failure paths.

engine/src/main/scala/cromwell/engine/workflow/FileSystemWorkflowStoreDatabase.scala
cjllanwarne: Those are needed, otherwise it would be `$separatoroptions.json`. IntelliJ is cool about it, I believe

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
cjllanwarne: I don't think it would be any different from Working?

engine/src/main/scala/cromwell/engine/workflow/FileSystemWorkflowStoreDatabase.scala
geoffjentry: This will be notafilesystemworkflowstoredatabase within a week, I wouldn't worry about that stuff here

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -223,10 +223,10 @@ class WorkflowManagerActor(config: Config, val workflowStore: ActorRef)
     * Submit the workflow and return an updated copy of the state data reflecting the addition of a
     * Workflow ID -> WorkflowActorRef entry.
     */
-  private def submitWorkflow(workflow: WorkflowStore.WorkflowToStart): WorkflowIdToActorRef = {
+  private def submitWorkflow(workflow: WorkflowStoreDatabase.WorkflowToStart): WorkflowIdToActorRef = {
     val workflowId = workflow.id
 
-    val startMode = if (workflow.state == WorkflowStore.Restartable) {
+    val startMode = if (workflow.state == WorkflowStoreDatabase.Restartable) {
```
geoffjentry: Shouldn't these be an aspect of the WorkflowStore? The fact that it's backed by a DB seems like an implementation detail

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
```diff
@@ -2,19 +2,19 @@ package cromwell.engine.workflow
...
+import cromwell.engine.workflow.WorkflowStoreActor.{WorkflowsBatchSubmittedToStore, _}
+import cromwell.services.MetadataServiceActor.{MetadataPutAcknowledgement, PutMetadataAction}
 import cromwell.services.{MetadataEvent, MetadataKey, MetadataValue, ServiceRegistryClient}
 
+import scala.concurrent.{ExecutionContext, Future}
 import scala.language.postfixOps
 import scalaz.NonEmptyList
 
-object WorkflowStore {
+object WorkflowStoreDatabase {
```
geoffjentry: Same comment as above. 

One could argue that database is being used generically and not to mean an RDBMS but I'd counter that if that's the case then store and database are synonymous here

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
```diff
@@ -48,70 +48,157 @@ object WorkflowStore {
...
+  implicit val ec: ExecutionContext = context.dispatcher
+
+  startWith(Idle, WorkflowStoreActorData(None, List.empty))
+  self ! InitializerCommand
+
+  when(Idle) {
+    case Event(cmd @ SubmitWorkflow(source), _) =>
+      val sndr = sender
+      val work = store.add(NonEmptyList(source)) map { ids =>
+        val id = ids.head
```
geoffjentry: it feels so weird not needing to worry about `.head` :)

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
```diff
@@ -48,70 +48,157 @@ object WorkflowStore {
...
-      val ids = add(sources)
-      ids foreach registerIdWithMetadataService
-      sender ! WorkflowsBatchSubmittedToStore(ids)
-      log.info(s"Workflows ${ids.list.mkString(", ")} submitted.")
-    case FetchRunnableWorkflows(n) => sender ! newWorkflowMessage(n)
-    case RemoveWorkflow(id) =>
-      if (!remove(id)) log.info(s"Attempted to remove ID $id from the WorkflowStore but it already exists!")
+case class WorkflowStoreActor(store: WorkflowStoreDatabase) extends LoggingFSM[WorkflowStoreActorState, WorkflowStoreActorData] with ActorLogging with ServiceRegistryClient {
+
+  implicit val ec: ExecutionContext = context.dispatcher
```
geoffjentry: Anything going to the DB (or FSDB) should be on the IO dispatcher

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
cjllanwarne: Isn't this your code...? :) 

But, I do agree

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
cjllanwarne: How do I get one of those?

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
geoffjentry: Hah, you barely dodged my original scathing reply. My machine rebooted as I was about to hit 'comment'

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
mcovarr: Hmm it just feels like the FSM is in a distinct uninitialized state in which it can't actually be used until this work completes.

engine/src/main/scala/cromwell/engine/workflow/WorkflowStoreActor.scala
cjllanwarne: Also, relevant: `def props(workflowStoreDatabase: WorkflowStoreDatabase = new InMemoryWorkflowStoreDatabase) = Props(WorkflowStoreActor(workflowStoreDatabase)).withDispatcher("akka.dispatchers.api-dispatcher")`

geoffjentry: You can do something like `implicit val ec = system.dispatchers.lookup("akka.dispatchers.io-dispatcher")` or whatever it is called.

The actor itself and it's normal stuff shouldn't be on the io-dispatcher, but any of the DB interaction should be. I'm guessing the only need for an EC is the DB stuff though

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreActor.scala
```diff
@@ -0,0 +1,200 @@
...
+    case fromState -> toState =>
+      log.debug("WorkflowStore moving from {} (using {}) to {} (using {})", fromState, stateData, toState, nextStateData)
+  }
+
+  private def addWorkCompletionHooks[A](command: WorkflowStoreActorCommand, work: Future[A]) = {
+    work.onComplete {
+      case Success(_) =>
+        self ! WorkDone
+      case Failure(t) =>
+        log.error("Error occurred during {}: {} because {}", command.getClass.getSimpleName, t.toString, t.getStackTrace.toList.mkString(System.lineSeparator()))
```
mcovarr: `ExceptionUtils.getStackTrace`

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreActor.scala
```diff
@@ -0,0 +1,200 @@
...
+    if (workflowStore.exists(_.id == id)) {
+      workflowStore = workflowStore filterNot { _.id == id }
+      Future.successful(true)
+    } else {
+      Future.successful(false)
+    }
+  }
+
+  override def initialize(implicit ec: ExecutionContext): Future[Unit] = Future.successful(())
+}
```
mcovarr: feels like this class should be in a separate file now?

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreActor.scala
```diff
@@ -0,0 +1,200 @@
...
+    workflowStore = workflowStore ++ submittedWorkflows.list
+    Future.successful(submittedWorkflows map { _.id })
+  }
+
+  /**
+    * Retrieves up to n workflows which have not already been pulled into the engine and sets their pickedUp
+    * flag to true
+    */
+  override def fetchRunnableWorkflows(n: Int)(implicit ec: ExecutionContext): Future[List[WorkflowToStart]] = {
+    val startableWorkflows = workflowStore filter { _.state.isStartable } take n
```
mcovarr: should this be giving precedence to restartable workflows over new submissions?

geoffjentry: It should be giving precedence to older submissions (ie FIFO behavior), which on startup *should* be `restartable` > `submitted`

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -223,10 +224,10 @@ class WorkflowManagerActor(config: Config, val workflowStore: ActorRef)
     * Submit the workflow and return an updated copy of the state data reflecting the addition of a
     * Workflow ID -> WorkflowActorRef entry.
     */
-  private def submitWorkflow(workflow: WorkflowStore.WorkflowToStart): WorkflowIdToActorRef = {
+  private def submitWorkflow(workflow: WorkflowToStart): WorkflowIdToActorRef = {
```
geoffjentry: Personally I prefer the explicitness (in this case `WorkflowStore.`) when referencing foreign message classes, as sometimes the names overlap

engine/src/main/scala/cromwell/engine/workflow/workflowstore/InMemoryWorkflowStoreDatabase.scala
```diff
@@ -0,0 +1,44 @@
+package cromwell.engine.workflow.workflowstore
+
+import cromwell.core.{WorkflowId, WorkflowSourceFiles}
+
+import scala.concurrent.{ExecutionContext, Future}
+import scalaz.NonEmptyList
+
+class InMemoryWorkflowStoreDatabase extends WorkflowStoreDatabase {
```
geoffjentry: Is this intended to stand the test of time, or does this also go away when the DB form shows up?

engine/src/main/scala/cromwell/engine/workflow/workflowstore/package.scala
```diff
@@ -0,0 +1,59 @@
...
+
+  final case class WorkflowToStart(id: WorkflowId, sources: WorkflowSourceFiles, state: StartableState)
+
+  /**
+    * Stores any submitted workflows which have not completed.
+    *
+    * This is not thread-safe, although it should not be being used in a multithreaded fashion. If that
+    * changes, revisit.
+    */
+  trait WorkflowStoreDatabase {
```
geoffjentry: Don't put this in the package object

geoffjentry: Also I thought you agreed re the redundancy of  `WorkflowStoreDatabase`?

kshakir:
LGTM for what it is 👍

engine/src/main/scala/cromwell/engine/workflow/workflowstore/package.scala
cjllanwarne: I agreed that the workflow state definitions shouldn't belong to the Database trait. But the trait itself is very useful as an interface

engine/src/main/scala/cromwell/engine/workflow/workflowstore/InMemoryWorkflowStoreDatabase.scala
cjllanwarne: It's mainly a test thing now. I'll move the `test/`

Issue #1178, opened 2016-07-18
Move services into services package


geoffjentry:
For more details see the DB meeting notes

At the moment the service actors live in the engine in order to take advantage of globalDataAccess. Once #1177 is complete move these over to the services package

PR #1193, opened 2016-07-20
Database Split Closes #1177 #1178

### LINKED ISSUE #1177: De-DataAccess SlickDatabase
For more specific details see the DB meeting notes

Remove the biz logic-y portions of the DataAccess trait and move them into something alongside/in the appropriate locations (metadata store, kv store, etc) and have that code take a SlickDatabase.

Move globalDataAccess to be a globally (in core package? Something like that) accessible singleton SlickDatabase instance, which the individual ex-DataAccess stuff can use.

The DataAccess trait also had a withRetry function which does seem inappropriate for SlickDatabase, at the moment it's just a wrapper around the main withRetry w/ some defaults set - perhaps that lives w/ the global SlickDatabase? Perhaps allow individual ex-DataAccess stuff set their own? Up to you.

Horneth:
Just to be clear on this - The SqlDatabase and its Slick implementation are not split out. Only the business logic code calling these SQL methods with primitive types has been scattered across sub-projects.


core/src/main/scala/cromwell/util/DatabaseUtil.scala
```diff
@@ -0,0 +1,22 @@
...
+import scala.concurrent.duration._
+import scala.language.postfixOps
+
+object DatabaseUtil {
+  private def isTransient(throwable: Throwable): Boolean = throwable match {
+    case _: SQLTransientException => true
+    case _ => false
+  }
+
+  def withRetry[A](f: => Future[A])(implicit actorSystem: ActorSystem): Future[A] = {
```
mcovarr: I don't think this should be a thunk, the Future can't be re-run in the `withRetry`.

Horneth: Hmm not sure I understand what you mean

mcovarr: In this example the `Future` actually runs only once, despite it being passed as a thunk:

```scala
scala> val f = Future {
     |   println("aw crap")
     |   throw new RuntimeException
     | }
aw crap
f: scala.concurrent.Future[Nothing] = scala.concurrent.impl.Promise$DefaultPromise@490dcafc

scala> def retry(r: => Future[Nothing]) {
     |   r
     |   r
     | }
retry: (r: => scala.concurrent.Future[Nothing])Unit

scala> retry(f)

scala> 
```

build.sbt
```diff
@@ -11,6 +11,7 @@ lazy val gcsfilesystem = (project in file("filesystems/gcs"))
 
 lazy val database = (project in file("database"))
   .settings(databaseSettings:_*)
+  .dependsOn(core % "test->test")
```
mcovarr: @kshakir may have some thoughts on this

Horneth: I thought the same thing. The only reason is to avoid duplicating Tags (eg PostMVP that can apply to any of the subprojects)

services/src/test/scala/cromwell/services/metadata/impl/MetadataDatabasAccessSpec.scala
```diff
@@ -1,210 +1,68 @@
...
-class SlickDataAccessSpec extends FlatSpec with Matchers with ScalaFutures with BeforeAndAfterAll with Mockito
-  with WorkflowDescriptorBuilder {
-  import SlickDataAccessSpec._
-
-  behavior of "SlickDataAccess"
+class MetadataDatabasAccessSpec extends FlatSpec with Matchers with ScalaFutures with BeforeAndAfterAll with Mockito {
+  import MetadataDatabasAccessSpec._
 
-  import TableDrivenPropertyChecks._
+  "MetadataDatabasAccess (main.hsqldb)" should behave like testWith("main.hsqldb")
```
mcovarr: Databas -> Database

services/src/test/scala/cromwell/services/metadata/impl/MetadataServiceActorSpec.scala
```diff
@@ -0,0 +1,76 @@
...
+import akka.pattern._
+import com.typesafe.config.ConfigFactory
+import cromwell.core.WorkflowId
+import cromwell.services.ServicesSpec
+import cromwell.services.metadata.MetadataService._
+import cromwell.services.metadata._
+
+class MetadataServiceActorSpec extends ServicesSpec("Metadata") {
+
+  val config = ConfigFactory.load("{}")
```
mcovarr: if this is intending to parse that as a String I think you mean `ConfigFactory.parseString`

engine/src/main/scala/cromwell/engine/backend/io/package.scala
```diff
@@ -1,15 +0,0 @@
...
-import cromwell.filesystems.gcs.{NioGcsPath, GcsFileSystemProvider}
-import cromwell.filesystems.gcs.{ContentTypeOption, GcsFileSystem}
-
-import scala.util.Try
-
-package object io {
-  val defaultGCSFileSystem = GcsFileSystem.defaultGcsFileSystem
-  val defaultFileSystem = FileSystems.getDefault
-  val defaultFileSystems = List(defaultGCSFileSystem, defaultFileSystem)
-}
```
mcovarr: whoops thought this was long gone 😦

services/src/main/scala/cromwell/services/keyvalue/KeyValueService.scala
```diff
@@ -0,0 +1,26 @@
...
+
+  case class ScopedKey(workflowId: WorkflowId, jobKey: JobKey, key: String)
+  case class KvPut(pair: KvPair) extends KvAction
+  case class KvGet(key: ScopedKey) extends KvAction
+
+  case class KvPair(key: ScopedKey, value: Option[String]) extends KvResponse
+  case class KvFailure(action: KvAction, failure: Throwable) extends KvResponse
+  case class KvKeyLookupFailed(action: KvGet) extends KvResponse
+  case class KvPutSuccess(action: KvPut) extends KvResponse
+}
```
mcovarr: \n

services/src/main/scala/cromwell/services/keyvalue/impl/KeyValueServiceActor.scala
```diff
@@ -0,0 +1,54 @@
...
+      get.key.jobKey.attempt,
+      get.key.key
+    )
+
+    executionInfo map {
+      case Some(maybeValue) => KvPair(get.key, maybeValue)
+      case None => KvKeyLookupFailed(get)
+    }
+  }
+}
```
mcovarr: \n

mcovarr:
:+1:

core/src/main/scala/cromwell/util/DatabaseUtil.scala
Horneth: Even if I pass `() => f` to the `Retry.withRetry` on line 20 ?
Anyway I can change it back to a full function I just find it more practical to use this way, but if doesn't work then probably not worth it :p

core/src/main/scala/cromwell/util/DatabaseUtil.scala
```diff
@@ -0,0 +1,22 @@
...
+import scala.concurrent.duration._
+import scala.language.postfixOps
+
+object DatabaseUtil {
+  private def isTransient(throwable: Throwable): Boolean = throwable match {
+    case _: SQLTransientException => true
+    case _ => false
+  }
+
+  def withRetry[A](f: () => Future[A])(implicit actorSystem: ActorSystem): Future[A] = {
```
cjllanwarne: Don't we already have one of these? And does it belong in `DatabaseUtil`?

core/src/main/scala/cromwell/util/DatabaseUtil.scala
```diff
@@ -0,0 +1,22 @@
...
+import java.sql.SQLTransientException
+
+import akka.actor.ActorSystem
+import cromwell.core.retry.{Retry, SimpleExponentialBackoff}
+
+import scala.concurrent.Future
+import scala.concurrent.duration._
+import scala.language.postfixOps
+
+object DatabaseUtil {
```
cjllanwarne: Does `DatabaseUtil` belong in `core` or should it be in `database`?

core/src/main/scala/cromwell/util/DatabaseUtil.scala
```diff
@@ -0,0 +1,22 @@
...
+  private def isTransient(throwable: Throwable): Boolean = throwable match {
```
cjllanwarne: This could be an implicit method on Throwable!

database/src/main/scala/cromwell/database/SqlDatabase.scala
```diff
@@ -10,184 +9,177 @@ import scalaz.NonEmptyList
...
 trait SqlDatabase extends AutoCloseable {
-  protected def isTransient(throwable: Throwable): Boolean = {
-    throwable match {
-      case e: SQLTransientException => true
-      case _ => false
-    }
-  }
-
-  protected def createWorkflow(workflowExecution: WorkflowExecution,
+  def createWorkflow(workflowExecution: WorkflowExecution,
```
cjllanwarne: This (and the other methods in this file) are actually no longer required, I believe

core/src/main/scala/cromwell/util/DatabaseUtil.scala
Horneth: The problem is our `Retry.withRetry` mechanism uses `ActorSystem`s to re-schehdule futures. 
Database sub-project doesn't have akka as a dependency, which seems good to me. Other alternative would be to re-implement Retries with something else than ActorSystems using only built-in java schedulers but that seemed a lot of work for only one method using this retry anyway.

database/src/main/scala/cromwell/database/slick/SlickDatabase.scala
```diff
@@ -602,7 +602,7 @@ class SlickDatabase(databaseConfig: Config) extends SqlDatabase {
     runTransaction(action)
   }
 
-  protected def addMetadata(events: Iterable[Metadatum])
+  override def addMetadata(events: Iterable[Metadatum])
```
cjllanwarne: I don't understand the new database structure at all so feel free to push back... But I thought these would be moved to a DataAccess in the services package, and `database` would only contain very generic things?

core/src/main/scala/cromwell/util/DatabaseUtil.scala
Horneth: Wait - are you suggesting adding an implicit ? Oo

Horneth: In this case I don't know if it would add a lot of value though because this method's only purpose is to be passed to the withRetry

database/src/main/scala/cromwell/database/slick/SlickDatabase.scala
Horneth: Yeah so I asked myself the same question - but the ticket said move out the "business logic" code to their respective sub-projects. Which I interpreted as all the raw Database query implementations can stay in the same SlickDataAccess file in the database, and the things with custom complex types on top of this have to move. But I might have misunderstood maybe we want to split SlickDataAccess too..

services/src/main/scala/cromwell/services/keyvalue/impl/KeyValueDatabaseAccess.scala
```diff
@@ -0,0 +1,34 @@
...
+import cromwell.database.Database
+import wdl4s.Scope
+
+import scala.concurrent.{ExecutionContext, Future}
+
+trait KeyValueDatabaseAccess { this: Database =>
+
+  def getExecutionInfoByKey(workflowId: WorkflowId, call: Scope, attempt: Int, key: String)
+                           (implicit ec: ExecutionContext): Future[Option[Option[String]]] = {
+    databaseInterface.getExecutionInfoByKey(workflowId.toString, call.fullyQualifiedName, attempt, key)
```
cjllanwarne: Is this really separating things out if it's just redirecting back to a shared `databaseInterface` method?

services/src/test/scala/cromwell/services/metadata/impl/MetadataDatabaseAccessSpec.scala
```diff
@@ -1,210 +1,68 @@
...
-//          val workflowOutput = results.headOption
-//          workflowOutput shouldNot be(empty)
-//          val workflowResult = workflowOutput.get
-//          workflowResult.wdlValue shouldNot be(empty)
-//          val wdlValue = workflowResult.wdlValue
-//          wdlValue.get.valueString shouldBe ""
-//        }
-//      } yield ()).futureValue
-    }
+    //    def assertTerminalExecution(id: WorkflowId, expectedCount: Int, expectedTerminal: ExpectedTerminal): Future[Unit] = {
```
cjllanwarne: Probably safe to just remove this actually

core/src/main/scala/cromwell/util/DatabaseUtil.scala
mcovarr: This seems to work as expected:

```scala
scala> retry (Future { 
     |   println ("aw crap")
     |   throw new RuntimeException
     | })
aw crap
aw crap

scala> 
```

I guess it's possible for the caller to screw this up with or without the `()` but maybe it's harder to screw it up with `()`?

core/src/main/scala/cromwell/util/DatabaseUtil.scala
cjllanwarne: Haha, yup, I surprised myself too!

Fair enough comment though. Looks fine to me.

database/src/main/scala/cromwell/database/slick/SlickDatabase.scala
cjllanwarne: Following the tech talk, does `Metadatum` counts as a simple type or a business logic type?

services/src/main/scala/cromwell/services/keyvalue/impl/KeyValueDatabaseAccess.scala
cjllanwarne: NM, I see the `toString`s now. LGTM

cjllanwarne:
Mine was a fairly shallow review but as far as I'm qualified to say so, 👍

database/src/main/scala/cromwell/database/slick/SlickDatabase.scala
Horneth: It's the Slick type that matches the Table columns - and it only contains scala primitives. So I'd say it's a simple type.

database/src/main/scala/cromwell/database/SqlDatabase.scala
Horneth: Yes a lot of them are useless now. Although I fee like deleting those implies deleting their Slick implementation which implies deleting the unused tables etc...
I don't think it's in the scope of this ticket. I'll make one.

database/src/main/scala/cromwell/database/slick/SlickDatabase.scala
kshakir: To be clear, `Metadatum` is _not_ a "Slick type"-- it is a database object.

The exclusive list of database types currently lives under [`cromwell.database.obj`](https://github.com/broadinstitute/cromwell/tree/adbae9e33267f83afd6203f1a223bc6e59d24d42/database/src/main/scala/cromwell/database/obj).

If one ever wanted to replace slick, there would still be a database object for each table, and within each database object the fields would still map to the database columns one-for-one.

project/Dependencies.scala
```diff
@@ -62,6 +64,7 @@ object Dependencies {
     "ch.qos.logback" % "logback-classic" % "1.1.3",
     "org.apache.commons" % "commons-lang3" % "3.4",
     "com.typesafe" % "config" % "1.3.0",
+    "org.codehaus.janino" % "janino" % "2.7.8",
```
kshakir: ToL (aka we can clean this up later)-- curious what is this? It popped up in the database dependencies also, but didn't use a version variable to keep the versions in sync. Also curious about the akka dependency in the database as well.

Horneth: I'll remove akka dependency, I was trying to add the Retry in the database layer itself but since it needs ActorSystem and I didn't want to add akka dependency, I moved it to core.
This `"org.codehaus.janino" % "janino" % "2.7.8`" is to allow "if" statements in logback.xml, which I copied over to the database project to avoid all the noise from slick. This particular one in core is again a breadcrumb from previous attempts which I'll nix. I'll harmonize the version.

PR #1197, opened 2016-07-22
JES Recover Closes #751

### LINKED ISSUE #751: Recovery support for JES PBE

mcovarr:
Doesn't close #751 but stabs in its general direction.  There is no persistent DB here, I replaced the DB-based approach in `KeyValueServiceActor` with a simple `Map` since the DB-based approach always failed to write due to referential integrity constraints that aren't going to be fixed (no `EXECUTION` or `WORKFLOW_EXECUTION` data for FK constraints).  Ruchi and/or I will have a follow-on PR to put a DB table behind this.

Also fixed "abort" to message the KV service instead of metadata service, removed now-unused `EXECUTION_INFO`-related methods, other assorted cleanup.

### DIFF
--- a/backend/src/main/scala/cromwell/backend/async/AsyncBackendJobExecutionActor.scala
+++ b/backend/src/main/scala/cromwell/backend/async/AsyncBackendJobExecutionActor.scala
@@ -21,9 +21,15 @@ object AsyncBackendJobExecutionActor {
   private final case class FailAndStop(reason: Throwable) extends AsyncBackendJobExecutionActorMessage
   private final case class Finish(executionHandle: ExecutionHandle) extends AsyncBackendJobExecutionActorMessage
 
-  sealed trait ExecutionMode extends AsyncBackendJobExecutionActorMessage
+  trait JobId
+
+  sealed trait ExecutionMode extends AsyncBackendJobExecutionActorMessage {
+    def jobId: Option[JobId] = None
+  }
   case object Execute extends ExecutionMode
-  final case class Resume(executionInfos: Map[String, Option[String]]) extends ExecutionMode
+  final case class Recover(recoveryId: JobId) extends ExecutionMode {
+    override def jobId = Option(recoveryId)
+  }
   final case class UseCachedCall(cachedBackendCall: BackendJobDescriptor) extends ExecutionMode
 }
 
--- a/services/src/main/scala/cromwell/services/keyvalue/impl/KeyValueDatabaseAccess.scala
+++ b/services/src/main/scala/cromwell/services/keyvalue/impl/KeyValueDatabaseAccess.scala
@@ -1,34 +1,7 @@
 package cromwell.services.keyvalue.impl
 
-import akka.actor.ActorSystem
-import cromwell.core.ExecutionIndex._
-import cromwell.util.DatabaseUtil._
-import cromwell.core.{JobKey, WorkflowId}
 import cromwell.database.Database
-import wdl4s.Scope
-
-import scala.concurrent.{ExecutionContext, Future}
 
 trait KeyValueDatabaseAccess { this: Database =>
 
-  def getExecutionInfoByKey(workflowId: WorkflowId, call: Scope, attempt: Int, key: String)
-                           (implicit ec: ExecutionContext): Future[Option[Option[String]]] = {
-    databaseInterface.getExecutionInfoByKey(workflowId.toString, call.fullyQualifiedName, attempt, key)
-  }
-
-  /**
-    * TODO: the interface for retrying SQL commands that might fail because
-    * of transient reasons (e.g. upsertExecutionInfo and upsertRuntimeAttributes)
-    * could be made better.  Also it'd be nice if it were transparently
-    * turned on/off for all methods in dataAccess.
-    *
-    * https://github.com/broadinstitute/cromwell/issues/693
-    */
-  def upsertExecutionInfo(workflowId: WorkflowId,
-                          callKey: JobKey,
-                          keyValues: Map[String, Option[String]])(implicit ec: ExecutionContext, actorSystem: ActorSystem): Future[Unit] = {
-    withRetry (() =>
-      databaseInterface.upsertExecutionInfo(workflowId.toString, callKey.scope.fullyQualifiedName, callKey.index.fromIndex, callKey.attempt, keyValues)
-    )
-  }
 }
--- a/services/src/main/scala/cromwell/services/keyvalue/impl/KeyValueServiceActor.scala
+++ b/services/src/main/scala/cromwell/services/keyvalue/impl/KeyValueServiceActor.scala
@@ -17,6 +17,7 @@ object KeyValueServiceActor {
 case class KeyValueServiceActor(serviceConfig: Config, globalConfig: Config) extends Actor with KeyValueDatabaseAccess with CromwellDatabase {
   private implicit val ec = context.dispatcher
   private implicit val system = context.system
+  private var store: Map[ScopedKey, Option[String]] = Map.empty
 
   def receive = {
     case action: KvGet => respond(sender(), action, doGet(action))
@@ -30,23 +31,13 @@ case class KeyValueServiceActor(serviceConfig: Config, globalConfig: Config) ext
     }
   }
 
-  private def doPut(put: KvPut): Future[KvResponse] = {
-    upsertExecutionInfo(
-      put.pair.key.workflowId,
-      put.pair.key.jobKey,
-      Map(put.pair.key.key -> put.pair.value)
-    ).map(_ => KvPutSuccess(put))
+  private def doPut(put: KvPut): Future[KvResponse] = Future.successful {
+    store += put.pair.key -> put.pair.value
+    KvPutSuccess(put)
   }
 
-  private def doGet(get: KvGet): Future[KvResponse] = {
-    val executionInfo = getExecutionInfoByKey(
-      get.key.workflowId,
-      get.key.jobKey.scope,
-      get.key.jobKey.attempt,
-      get.key.key
-    )
-
-    executionInfo map {
+  private def doGet(get: KvGet): Future[KvResponse] = Future.successful {
+    store.get(get.key) match {
       case Some(maybeValue) => KvPair(get.key, maybeValue)
       case None => KvKeyLookupFailed(get)
     }
--- a/supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
+++ b/supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
@@ -11,9 +11,10 @@ import better.files._
 import com.google.api.client.googleapis.json.GoogleJsonResponseException
 import cromwell.backend.BackendJobExecutionActor.{AbortedResponse, BackendJobExecutionResponse}
 import cromwell.backend.BackendLifecycleActor.AbortJobCommand
-import cromwell.backend.async.AsyncBackendJobExecutionActor.ExecutionMode
+import cromwell.backend.async.AsyncBackendJobExecutionActor.{ExecutionMode, JobId}
 import cromwell.backend.async.{AbortedExecutionHandle, AsyncBackendJobExecutionActor, ExecutionHandle, FailedNonRetryableExecutionHandle, FailedRetryableExecutionHandle, NonRetryableExecution, SuccessfulExecutionHandle}
 import cromwell.backend.impl.jes.JesImplicits.PathString
+import cromwell.backend.impl.jes.JesJobExecutionActor.JesOperationIdKey
 import cromwell.backend.impl.jes.RunStatus.TerminalRunStatus
 import cromwell.backend.impl.jes.io._
 import cromwell.backend.{AttemptedLookupResult, BackendJobDescriptor, BackendJobDescriptorKey, BackendWorkflowDescriptor, ExecutionHash, PreemptedException}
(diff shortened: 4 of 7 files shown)


geoffjentry:
Why does the PR say "closes #751" and the text says "Doesn't close #751"? 😕

mcovarr:
@geoffjentry I was hoping the PR would get rolled up under #751 in ZenHub, but apparently this only bought me an "in review" label.  Oh well.

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesJobExecutionActor.scala
```diff
@@ -28,40 +32,58 @@ case class JesJobExecutionActor(override val jobDescriptor: BackendJobDescriptor
...
-    // PBE TODO: use PartialFunction.orElse instead of this catch-all case, because Akka might use isDefinedAt over this partial function
-    case message => super.receive(message)
+    case KvPair(key, id @ Some(operationId)) if key.key == JesOperationIdKey =>
+      // Successful operation ID lookup during recover.
+      executor foreach { _ ! Recover(id)}
+    case KvKeyLookupFailed(_) =>
+      // Missed operation ID lookup during recover, fall back to execute.
+      executor foreach { _ ! Execute }
+    case KvFailure(_, e) =>
+      // Failed operation ID lookup during recover, crash and let the supervisor deal with it.
```
mcovarr: complete the Promise with a failure

backend/src/main/scala/cromwell/backend/async/AsyncBackendJobExecutionActor.scala
```diff
@@ -21,9 +21,11 @@ object AsyncBackendJobExecutionActor {
   private final case class FailAndStop(reason: Throwable) extends AsyncBackendJobExecutionActorMessage
   private final case class Finish(executionHandle: ExecutionHandle) extends AsyncBackendJobExecutionActorMessage
 
-  sealed trait ExecutionMode extends AsyncBackendJobExecutionActorMessage
+  sealed trait ExecutionMode extends AsyncBackendJobExecutionActorMessage {
+    def jesOperationId: Option[String] = None
+  }
   case object Execute extends ExecutionMode
-  final case class Resume(executionInfos: Map[String, Option[String]]) extends ExecutionMode
+  final case class Recover(override val jesOperationId: Option[String]) extends ExecutionMode
```
mcovarr: shouldn't be an `Option`

geoffjentry: obvs

backend/src/main/scala/cromwell/backend/async/AsyncBackendJobExecutionActor.scala
```diff
@@ -21,9 +21,11 @@ object AsyncBackendJobExecutionActor {
   private final case class FailAndStop(reason: Throwable) extends AsyncBackendJobExecutionActorMessage
   private final case class Finish(executionHandle: ExecutionHandle) extends AsyncBackendJobExecutionActorMessage
 
-  sealed trait ExecutionMode extends AsyncBackendJobExecutionActorMessage
+  sealed trait ExecutionMode extends AsyncBackendJobExecutionActorMessage {
+    def jesOperationId: Option[String] = None
```
geoffjentry: Why is there a `jesOperationId` in `AsyncBackendJobExecutionActor`? Seems awfully specific

backend/src/main/scala/cromwell/backend/async/AsyncBackendJobExecutionActor.scala
geoffjentry: Same comment as above, more or less

services/src/main/scala/cromwell/services/keyvalue/impl/KeyValueDatabaseAccess.scala
```diff
@@ -1,34 +1,7 @@
 package cromwell.services.keyvalue.impl
```
geoffjentry: Since you're making hefty changes to the KV, can you go ahead and rename it to something more apropos as per the discussion last week?

services/src/main/scala/cromwell/services/keyvalue/impl/KeyValueDatabaseAccess.scala
```diff
@@ -1,34 +1,7 @@
...
-import akka.actor.ActorSystem
-import cromwell.core.ExecutionIndex._
-import cromwell.util.DatabaseUtil._
-import cromwell.core.{JobKey, WorkflowId}
 import cromwell.database.Database
-import wdl4s.Scope
-
-import scala.concurrent.{ExecutionContext, Future}
 
 trait KeyValueDatabaseAccess { this: Database =>
```
geoffjentry: I'm assuming this is empty for future expansion?

geoffjentry:
👍

services/src/main/scala/cromwell/services/keyvalue/impl/KeyValueDatabaseAccess.scala
mcovarr: Happy to TechTalk™️ this on Monday.  The choice of name may depend on whether this same store is used for call caching.

backend/src/main/scala/cromwell/backend/async/AsyncBackendJobExecutionActor.scala
mcovarr: Yeah that's not right, I need to generify that.

services/src/main/scala/cromwell/services/keyvalue/impl/KeyValueDatabaseAccess.scala
mcovarr: yes

services/src/main/scala/cromwell/services/keyvalue/impl/KeyValueDatabaseAccess.scala
geoffjentry: FWIW I've got the tall day training tomorrow. 

I've been thinking about CC for most of the morning but that's had the effect of increasing the ways I could see it working, not decreasing. However using the KV seems among the less likely of the reasonable possibilities but who knows

geoffjentry: I take that last statement back. Minus the "but who knows" portion

services/src/main/scala/cromwell/services/keyvalue/impl/KeyValueServiceActor.scala
```diff
@@ -30,23 +31,13 @@ case class KeyValueServiceActor(serviceConfig: Config, globalConfig: Config) ext
     }
   }
 
-  private def doPut(put: KvPut): Future[KvResponse] = {
-    upsertExecutionInfo(
-      put.pair.key.workflowId,
-      put.pair.key.jobKey,
-      Map(put.pair.key.key -> put.pair.value)
-    ).map(_ => KvPutSuccess(put))
+  private def doPut(put: KvPut): Future[KvResponse] = Future.successful {
```
cjllanwarne: I don't like the way the Futures are structured in this Actor.

We have a receive that spins up a Future which modifies the state out-of-bound of the receive block. All sorts of race conditions are possible there...

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
```diff
@@ -145,33 +147,26 @@ class JesAsyncBackendJobExecutionActor(override val jobDescriptor: BackendJobDes
...
-              case Success(_) => jobLogger.info(s"$tag Aborted $jobId")
-              case Failure(ex) => jobLogger.warn(s"$tag Failed to abort $jobId: ${ex.getMessage}")
-            }
-          case None => jobLogger.warn(s"$tag Failed to abort ${jobDescriptor.call.fullyQualifiedName}. Run Id metadata field was empty.")
-        }
+      serviceRegistryActor ! KvGet(ScopedKey(jobDescriptor.descriptor.id, jobDescriptor.key, JesOperationIdKey))
+    case KvPair(scopedKey, operationId) if scopedKey.key == JesOperationIdKey =>
+      operationId match {
+        case Some(id) =>
+          jobLogger.info(s"$tag Aborting $id")
```
cjllanwarne: I thought we were using the operation ID to Restore an operation... not abort it?

services/src/main/scala/cromwell/services/keyvalue/impl/KeyValueServiceActor.scala
Horneth: I don't think that a `Future.successful` executes asynchronously. It just wraps whatever code is inside into a successful Future and returns.

geoffjentry: @Horneth speaks the truth

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesJobExecutionActor.scala
```diff
@@ -28,40 +33,59 @@ case class JesJobExecutionActor(override val jobDescriptor: BackendJobDescriptor
...
+  }
+
+  override def recover: Future[BackendJobExecutionResponse] = {
+    import JesJobExecutionActor._
+
+    for {
+      _ <- launchExecutor
+      _ = serviceRegistryActor ! KvGet(ScopedKey(jobDescriptor.descriptor.id, jobDescriptor.key, JesOperationIdKey))
+      c <- completionPromise.future
+    } yield c
```
cjllanwarne: I would appreciate a tech talk (sort of like Khalid's database explanapaloosa) to explain what on earth is going on with these promises in this actor at some point.

E.g. why does a KvFailure above complete the promise with a failure but a KvSuccess does not, wouldn't the promise already have completed here? What even IS a promise? Much questions, many confusing.

services/src/main/scala/cromwell/services/keyvalue/impl/KeyValueServiceActor.scala
cjllanwarne: Yup, I know how that works but look at the `receive` logic: Ignoring the specific implemention of `doPut`, the `receive` should not farm out state manipulation to something that returns a `Future`

geoffjentry: @cjllanwarne also speaks the truth

Horneth: Right - although I imagine this will not be a problem when it switches to a Database implementation.

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
mcovarr: We use it for both actually.  The above is a somewhat unrelated refactoring of the abort logic to use the KV store and not the metadata service.

cjllanwarne: As a side note, I don't know why abort even needs to look remotely for operation ID - the actor already uses the operation ID every time it queries status, so why is abort a special case?

services/src/main/scala/cromwell/services/keyvalue/impl/KeyValueServiceActor.scala
cjllanwarne: Perhaps, since this is currently CR~~UD~~. But things change and I'm very wary of "it's fine to have race conditions because they're unlikely"

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesJobExecutionActor.scala
mcovarr: I'll point out first that @kshakir seemed to be suggesting a different approach over in his PR though I'm not sure what that approach would be.  Anyway here's how the current scheme works:

Both `recover` above and `execute` below launch an executor actor and hand back a `Future` based on the completion promise.  `execute` just tells the executor to run with an `Execute` message.  `recover` is a bit more complicated, it must first try to fetch the operation ID based on the job key before messaging the executor.  So the first thing `recover` does is query the KV store for the operation ID.

The KV store operation ID query can be answered in one of three ways: 

* `KvPair(...)` which provides an operation ID and allows for true recovery.
* `KvKeyLookupFailed(...)` where there is no operation ID found.  This is less than ideal but not a disaster, the logic just falls back to running the job anew.
* `KvFailure` where something went horribly wrong.

The `Promise` is used here as a proxy to the `Future` that is required to be returned by the `execute` and `recovery` method signatures.  Once the outcome of the recovery or execution is known, that `Promise` is completed either successfully with the response, or unsuccessfully with the reason for failure.  

In the `KvFailure` case something has gone horribly wrong, so the `Future` is completed as failed.  In the other cases the execution is proceeding normally and the existing code will complete the `Future` at a later point as it did prior to these changes.

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
mcovarr: Uh hmm that does sound right.  @scottfrazer?

services/src/main/scala/cromwell/services/keyvalue/impl/KeyValueServiceActor.scala
geoffjentry: In general I agree w/ the sentiment that @cjllanwarne has here. I get skeeved out by stuff which are AOK now but are taking advantage of technicalities and could easily go horribly wrong in a manner which is easy to escape code review.

At the very least it would be helpful to have a comment in an obvious location stating that if this ever *was* async the whole thing should be changed.

mcovarr: This is just a super short term placeholder for the PR after this one and #1200

cjllanwarne:
As a halfway house (which I guess is what this is?): 👍

PR #1198, opened 2016-07-22
Workflow Store as a database table

cjllanwarne:

PR #1199, opened 2016-07-25
Patches to Workflow Store as a database table

kshakir:


kshakir:
Overall, a code-based start of discussion. See also individual commits messages. Some or all could be integrated into #1198.

database/src/main/scala/cromwell/database/sql/tables/WorkflowStoreEntry.scala
```diff
@@ -0,0 +1,14 @@
...
+
+import java.sql.Timestamp
+
+case class WorkflowStoreEntry
+(
+  workflowUuid: String,
+  workflowSource: String,
+  workflowInputs: Option[String],
+  workflowOptions: Option[String],
+  state: String, // TODONE: This is a database column, and **the column** has no restrictions on the string stored.
```
cjllanwarne: In that case, I think I ought to replace the rather vague `initialize` method to the DB with a specific state update invocation and push knowledge of `Submitted`, `Running` and `Restartable` entirely up into the engine.

PR #1202, opened 2016-07-25
Return of the SGE (& friends!)

kshakir:
Added a new Shared File System (SFS) Backend package as a base for any/most backends wanting to execute cromwell calls on an SFS.
Added SGE and LSF configs (commented out) to the main `application.conf`, plus other small tweaks.
Non-main `application.conf` files only set the keys that they need to add or override from the main.
More of the config is defaulted across the board.
A number of classes moved/refactored from other backends for use by the SFS backend.
New trait `RuntimeAttributesValidation[ValidatedType]` for proccessing runtime attributes.
First pass at a `WorkflowFileSystemProvider` that combines a config and workflow options to create a filesystem. Only implemented for local and GCS files.
Added a Config backend that implements the SFS backend by reading wdl-lite command strings and `job-id-regex`.
Currently extending SFS backend to create the SGE backend until one can specify runtime attributes via the Config backend.
Added a "shadow" local backend, extending the SFS backend, that will be evaluated for a while before hopefully usurping the original.
Fixed integration tests that were attempting to mock the gcs file system.
Gave `MetadataDatabaseAccessSpec` a little bit more time the non-split up "create and query a workflow".
Fixed the `SprayDockerRegistryApiClientSpec` for v1 registries by switching from the `ubuntu` repo to the `registry` repo (we query the registry for a repo hosting images of the docker registry-- inception).


backend/src/main/scala/cromwell/backend/validation/ContinueOnReturnCodeValidation.scala
```diff
@@ -0,0 +1,46 @@
...
+    case WdlInteger(value) => ContinueOnReturnCodeSet(Set(value)).successNel
+    case WdlArray(wdlType, seq) =>
+      val nels: Seq[ErrorOr[Int]] = seq map validateInt
+      val numFailures: Int = nels.count(_.isFailure)
+      if (numFailures == 0) {
+        val defaultReturnCodeNel = Set.empty[Int].successNel[String]
+        nels.foldLeft(defaultReturnCodeNel)((acc, v) => (acc |@| v) { (a, v) => a + v }) map ContinueOnReturnCodeSet
+      } else {
+        failureWithMessage
+      }
```
mcovarr: `sequence` might help here:

```scala
    case WdlArray(wdlType, seq) =>
      val errorOrInts = (seq.toList map validateInt).sequence[ErrorOr, Int]
      errorOrInts match {
        case scalaz.Success(ints) => ContinueOnReturnCodeSet(ints.toSet).successNel
        case scalaz.Failure(_) => failureWithMessage
      }
```

kshakir: Awesome, thanks! Though tomorrow I'm going to need you to translate this into english, or at least scala. (Seriously, where are you learning this stuff? And I still bristle at blogs with "starting from an understanding of haskell/applied set based theory...")

pgrosu: @kshakir Sorry to ask, but what part do you find magical?  This is just regular functional programming.  You can learn most of the key parts of Scalaz (functional programming for Scala), from either the following website or the PDF - or the source code:

* Website (Manual): http://eed3si9n.com/learning-scalaz/
* PDF (Manual): http://eed3si9n.com/learning-scalaz/learning-scalaz.pdf
* Github (Source Code): https://github.com/scalaz/scalaz

Hope it helps,
Paul

geoffjentry: @kshakir FWIW the best textual overview I've seen was the underscore advanced scala book, although it's now cats based. Most of what I know (which is admittedly much less than @mcovarr, but he's the one who has been learning Haskell) comes from watching talks and reading code from the notable scalawags (who also have generally morphed over to cats)

pgrosu: Jeff, you mean like this :)

programming cat

backend/src/main/scala/cromwell/backend/validation/RuntimeAttributesValidation.scala
```diff
@@ -47,8 +46,8 @@ object RuntimeAttributesValidation {
       case Some(WdlInteger(i)) => ContinueOnReturnCodeSet(Set(i)).successNel
       case Some(WdlArray(wdlType, seq)) =>
         val nels: Seq[ErrorOr[Int]] = seq map validateInt
-        val nrFailures: Int = nels count { case p => p.isInstanceOf[Failure[NonEmptyList[String]]] }
-        if (nrFailures == 0) {
+        val numFailures: Int = nels.count(_.isFailure)
+        if (numFailures == 0) {
           val defaultReturnCodeNel = Set.empty[Int].successNel[String]
           nels.foldLeft(defaultReturnCodeNel)((acc, v) => (acc |@| v) { (a, v) => a + v }) map ContinueOnReturnCodeSet
```
mcovarr: This looks awfully familiar...

backend/src/main/scala/cromwell/backend/validation/ContinueOnReturnCodeValidation.scala
pgrosu: @kshakir Please don't hesitate to ask anything.  You're among friends, and we would be more than glad to bring you up to speed on anything that you would like to learn.

kshakir: @pgrosu No worries on trying to help, I'm really interested in scalaz, but haven't found a resource that works for my learning style, yet.

For an example of what I still find magical-- I'm used to the function to convert  `Seq[Future[A]]` to `Future[Seq[A]]` being `Future.sequence`. Similarly in Slick `Seq[DBIO[A]]` to `DBIO[Seq[A]]` is `DBIO.sequence`. I actually learned about the first via coursera, and then stumbled upon the other while browsing the companion object for `DBIO`.

I would have _never_ guessed there existed a different syntax, an implicit, "magic", method to convert an `ErrorOr[List[A]]` to an `List[ErrorOr[A]]`. Nor would I have guessed it didn't work on abstract types like `Seq`, only the concrete implementation `List`. And I have no idea where to look currently, besides clicking through the scaladoc or wishing for a hit on stack overflow. With hindsight, I'll add this one method to my slowly growing toolbox, along with semigroup's `|+|`, and the other parts of validation that Miguel and Jeff have been showing us.

As for my preferences, I'm originally a java developer. Although I appreciate Eugene's well cited work, I am _not_ the audience he's writing for. On links that open from google, instead of citing scala examples, or maybe even using java as a reference, he frequently falls back to "Haskell wiki says: ..." or "LYAHFGG: ..." Great starting point for others, not for me personally.

When learning scala, I didn't start with the scaladocs, nor even the EPFL website. Instead books like [_Scala for Java Developers_](https://www.safaribooksonline.com/library/view/scala-for-java/9781783283637/) and the free [_Programming Scala (1st Ed.)_](http://www.artima.com/pins1ed/) were bookmarked, and I could search through them for examples (and often limitations). I haven't found a resource for java developers yet who are interested in scalaz. Jeff's referenced book is paywalled by underscore, and not available on the Broad's safari books subscription, as far as I know. Still, I may have to just pray-and-pay that the book is actually written for non-math majors, and hope that it might be my rosetta stone.

Other resources welcome => kshakir |@| broadinstitute . org

**TL;DR Still searching for a complete java-to-scalaz dictionary**

backend/src/main/scala/cromwell/backend/validation/RuntimeAttributesValidation.scala
kshakir: Only touched this method in the `object RuntimeAttributesValidation` because of my _"must. get. green. intellij. checkmark."_ Otherwise, the functions were generally left behind for the other backends to use as is, while the SFS is now using these methods encapsulated in subclasses of the `trait RuntimeAttributesValidation`.

backend/src/main/scala/cromwell/backend/validation/ContinueOnReturnCodeValidation.scala
geoffjentry: Fwiw I'm happy to give you a try before you buy version of the _ book if you want to get a feel for it. But it's about the same tone as the slick stuff was

backend/src/main/scala/cromwell/backend/BackendInitializationData.scala
```diff
@@ -0,0 +1,13 @@
+package cromwell.backend
```
geoffjentry: Thanks. I was just saying to myself the other day, "I wish this was in its own file"

backend/src/main/scala/cromwell/backend/ExecutionHash.scala
```diff
@@ -0,0 +1,13 @@
+package cromwell.backend
+
+import java.util.UUID
+
+import scala.concurrent.{ExecutionContext, Future}
+
+object ExecutionHash {
+  // TODO: PBE: ideally hashes should be deterministic
```
geoffjentry: At the very least, can we nix the `PBE: `, the PBE TODOs should either be TODONE, TODO or removed at this point

backend/src/main/scala/cromwell/backend/io/WorkflowPathsBackendInitializationData.scala
```diff
@@ -0,0 +1,25 @@
+package cromwell.backend.io
+
+import java.nio.file.FileSystem
+
+import cromwell.backend.BackendInitializationData
+
+trait WorkflowPathsBackendInitializationData extends BackendInitializationData {
```
geoffjentry: can you throw in some comments on what all of this is?

core/src/main/resources/application.conf
```diff
@@ -188,7 +188,21 @@ backend {
...
+    }
+
+    //Sge {
+    //  actor-factory = "cromwell.backend.impl.sge.SgeBackendLifecycleActorFactory"
+    //}
+    //
+    //Lsf {
+    //  actor-factory = "cromwell.backend.impl.sfs.ConfigBackendLifecycleActorFactory"
+    //  config {
+    //    submit = "bsub -J ${job_name} -cwd ${cwd} -o ${out} -e ${err} ${script}"
```
Horneth: That's deep config xD

supportedBackends/local/src/main/scala/cromwell/backend/impl/shadowlocal/ShadowLocalBackendLifecycleActorFactory.scala
```diff
@@ -0,0 +1,68 @@
...
+    val dockerRun = DockerValidation.optional.extract(validatedRuntimeAttributes) map buildDockerRunCommand getOrElse ""
+    new SharedFileSystemCommand(Seq("/bin/bash", "-c", s"cat ${jobPaths.script} | $dockerRun /bin/bash <&0"))
+  }
+
+  override lazy val stdoutSubmit = jobPaths.stdout.untailed
+
+  override lazy val stderrSubmit = jobPaths.stderr.untailed
+
+  override def getJob(process: Process, stdoutWriter: PathWriter, stderrWriter: PathWriter) = {
+    val proc = process
```
Horneth: What is the purpose of creating another val here ?

backend/src/main/scala/cromwell/backend/ExecutionHash.scala
kshakir: I understand the theory, but I have been leaving non-PBE `TODO` comments around "integration" points for code that's probably fine... but if not provides a hint to future-me as to a fix.

supportedBackends/local/src/main/scala/cromwell/backend/impl/shadowlocal/ShadowLocalBackendLifecycleActorFactory.scala
kshakir: Whoops, copypasta from my scratch pad. Really, all the reflection hacks should be cached to. Will tighten it up.

backend/src/main/scala/cromwell/backend/validation/ContinueOnReturnCodeValidation.scala
mcovarr: I first learned of `sequence` through the same Coursera class you mention in the context of transforming `List[Future[A]]` to `Future[List[A]]`.  And as you also mention Slick offers a similar facility for `Seq[DBIO[A]]` to `DBIO[Seq[A]]`.

In picking up Haskell I learned the concept of transforming a `F[G[A]]` to `G[F[A]]` is captured by the `Traversable` typeclass.  This is called `Traverse` in both Scalaz and Cats as the `Traversable` name is already taken by a trait in the standard library.

Scalaz appears to put its "batteries included" typeclasses for standard library classes in scalaz.std, and there's a `ListInstances` there in `List.scala` that provides a `Traverse` instance for `List`.  The signature of `sequence` looks like:

```scala
  /** Traverse with the identity function. */
  def sequence[G[_]:Applicative,A](fga: F[G[A]]): G[F[A]] =
    traversal[G].run[G[A], A](fga)(ga => ga)
```

The role of `F` here is played by `List`, so as long as `G` has an `Applicative` instance it will be possible to transform `F[G[A]]` to `G[F[A]]`.  The role of `G` here is played by `NonEmptyList`, which in its source file has a `Monad` instance, and `Monad` extends `Applicative`.

I haven't found a really effective java to scalaz translation either.  Instead I've found it a lot easier to learn concepts in Haskell and then figure out how they might be applied to Scala using Scalaz or Cats.

mcovarr: Edit: oops `Validation` is actually the `G` here, and that has an `Applicative` instance in `Validation.scala`.

backend/src/main/scala/cromwell/backend/io/WorkflowPathsBackendInitializationData.scala
mcovarr: plz?  😄

backend/src/main/scala/cromwell/backend/validation/RuntimeAttributesValidation.scala
```diff
@@ -105,4 +95,179 @@ object RuntimeAttributesValidation {
...
+    * Returns true if the value can be validated.
+    *
+    * @return true if the value can be validated.
+    */
+  protected def validateExpression: PartialFunction[WdlValue, Boolean] = {
+    case wdlValue => coercion.exists(_ == wdlValue.wdlType)
+  }
+
+  /**
+    * Used to during initialization to validate if an expression could be validated.
```
mcovarr: That sentence isn't quite right.

backend/src/main/scala/cromwell/backend/validation/RuntimeAttributesValidation.scala
```diff
@@ -105,4 +95,179 @@ object RuntimeAttributesValidation {
...
+    */
+  def validate(values: Map[String, WdlValue]): ErrorOr[ValidatedType] = {
+    values.get(key) match {
+      case Some(value) => validateValue.applyOrElse(value, (_: Any) => failureWithMessage)
+      case None => validateNone
+    }
+  }
+
+  /**
+    * Returns true if the value can be validated.
```
mcovarr: That doc doesn't seem quite right, this is checking if a coercion exists.

backend/src/main/scala/cromwell/backend/validation/RuntimeAttributesValidation.scala
```diff
@@ -105,4 +95,179 @@ object RuntimeAttributesValidation {
...
+    * @param wdlValue The default wdl value.
+    * @return The new version of this validation.
+    */
+  final def withDefault(wdlValue: WdlValue) = RuntimeAttributesValidation.withDefault(this, wdlValue)
+
+  private[validation] lazy val validateValueInternal: PartialFunction[WdlValue, ErrorOr[ValidatedType]] = validateValue
+
+  private[validation] lazy val validateExpressionInternal: PartialFunction[WdlValue, Boolean] = validateExpression
+
+  private[validation] lazy val failureMessageInternal: String = failureMessage
```
mcovarr: I'm not sure I get what's going on with these "Internal"s... 😦

backend/src/main/scala/cromwell/backend/validation/RuntimeAttributesValidation.scala
```diff
@@ -105,4 +95,179 @@ object RuntimeAttributesValidation {
...
+    *
+    * @param attrs The values to search.
+    * @return The value matching the key.
+    */
+  def extract(attrs: ValidatedRuntimeAttributes): ValidatedType = {
+    attrs.attributes.apply(key).asInstanceOf[ValidatedType]
+  }
+
+  /**
+    * Returns Some(value) from the attributes matching the validation key, or None.
```
mcovarr: docs don't seem to match code or I'm not getting how they do... 😄

mcovarr: i.e. I suspect the cast and return type are bad

backend/src/main/scala/cromwell/backend/validation/ValidatedRuntimeAttributesBuilder.scala
```diff
@@ -0,0 +1,90 @@
...
+  private def validate(values: Map[String, WdlValue]): ErrorOr[ValidatedRuntimeAttributes] = {
+    val validationsForValues = validations ++ validationsInternal
+    val errorsOrValuesMap = validationsForValues.map(validation => validation.key -> validation.validate(values))
+
+    import scalaz.Scalaz._
+    import scalaz._
+
+    implicit val semigroupForMap = new Semigroup[Map[String, Any]] {
+      override def append(f1: Map[String, Any], f2: => Map[String, Any]) = f1 ++ f2
+    }
```
mcovarr: This is actually not a `Semigroup`.  No `Semigroup` is asserted for the value type so it's not clear how merges should be handled.  More info at

http://stackoverflow.com/questions/7076128/best-way-to-merge-two-maps-and-sum-the-values-of-same-key

backend/src/main/scala/cromwell/backend/validation/ValidatedRuntimeAttributesBuilder.scala
```diff
@@ -0,0 +1,90 @@
...
+
+final case class ValidatedRuntimeAttributes(attributes: Map[String, Any])
+
+trait ValidatedRuntimeAttributesBuilder {
+
+  /**
+    * Returns the validations that should be used and returned during evaluation.
+    *
+    * These validations will be checked on backend initialization against all the calls. If a validation value will not
+    * be supported, but should be checked anyway, supply it in `valuesOnlyValidations`, not here..
```
mcovarr: ..

build.sbt
```diff
@@ -31,33 +31,39 @@ lazy val backend = (project in file("backend"))
   .dependsOn(services)
   .dependsOn(core % "test->test")
 
+lazy val sfsBackend = (project in backendRoot / "sfs")
```
mcovarr: I'm wondering if we could pick a name for this project that makes it clear this is an abstract supertype for building shared filesystem backends, without wrongly implying it is itself a usable backend.  `sfsBackend` and`lsfBackend` look so similar but represent such different ideas... 😄

core/src/main/resources/application.conf
```diff
@@ -306,21 +296,23 @@ database {
...
     hsqldb {
-      db.url = "jdbc:hsqldb:mem:${slick.uniqueSchema};shutdown=false;hsqldb.tx=mvcc"
-      db.driver = "org.hsqldb.jdbcDriver"
-      db.connectionTimeout = 1000
       driver = "slick.driver.HsqldbDriver$"
+      db {
+        driver = "org.hsqldb.jdbcDriver"
+        url = "jdbc:hsqldb:mem:${slick.uniqueSchema};shutdown=false;hsqldb.tx=mvcc"
+        connectionTimeout = 3000
+      }
```
mcovarr: yay

core/src/test/resources/application.conf
```diff
@@ -1,196 +1,12 @@
...
-
-system {
-  // Use the new route of execution starting ShadowWorkflowActor
-  // Once all features related to pluggable backends are in place, this will be deprecated
-  shadowExecutionEnabled = false
-
-  // If 'true', a SIGINT will trigger Cromwell to attempt to abort all currently running jobs before exiting
-  abort-jobs-on-terminate = false
-}
+// Additions/updates to main's application.conf
```
mcovarr: Are we sure the configs are loaded with that precedence?

backend/src/main/scala/cromwell/backend/validation/ContinueOnReturnCodeValidation.scala
pgrosu: @kshakir No problem, but I hope you won't mind if we start with some fundamentals as most of the functional programming languages basically use the same logic in their approach, and understanding those usually clears up most the confusion.  The rest is just syntactic sugar :)

So as the Computer Science department in the 1960s usually was part of the Mathematics department, a lot of this is based on certain foundations of mathematics which you already noticed.  So I will start with some definitions which I will slowly build up:

#### Symbols and Numbers (Atoms)

For simplicity, the smallest form of representation in any language would a symbol or number.  Numbers usually reside in a space (i.e. set of real number).

#### Operators

An operator is a symbol that maps from one space to another, and to simplify we map from the real number system to the real number system.  For example

Plus operator `(+)` maps the list `(1 2)` the number `3`.

Operators come from a branch of mathematics called [Operator theory](https://en.wikipedia.org/wiki/Operator_theory).

#### Functions

A function is a relation of inputs operated on, to provide an output.

For example:

  `x -> f(x)` is a function, where `x` is the input and `f(x)` is the function operating on it.  The inputs and outputs would be symbols.

#### Expressions

So an expression is a combination of symbols that has a syntax that can be used evaluation by an interpreter.  

Or more precisely:

```
   <expression> = <symbol> | <function>`
```
The `|` would be denote either (OR).

#### S-Expressions

These are symbolic expressions (s-exprs), and we can take the simplest form in languages such as LISP where we define as two types:

 - atom (i.e. a number, a character, a string, etc.)
 - list (a list of numbers, list of strings, list of combinations, etc.)

 Example:

```
  (1 2)
  (a b c)
  ( 1, "fluffy", "cat", "tail")
```
You can think of lists as sets from mathematics.

#### Functional Languages

In functional languages everything has a value.  Thus a function is constructed with operators (in prefix form) acting on values to produce values.  For example:

````
  (+ 1 2)
```
The operator + (prefixing the rest of the list) runs across the rest of the list to produce the value 3.  The reason why it is like this is because it naturally follows the mathematical interpretation of saying:

  ```
Add 1 with 2
```

#### [Lambda Calculus](https://en.wikipedia.org/wiki/Lambda_calculus)

[Lambda Calculus](https://en.wikipedia.org/wiki/Lambda_calculus) was created by [Alonzo Church](https://en.wikipedia.org/wiki/Alonzo_Church) in the 1930s to define the simplest way of formalizing a function with which computation can be performed - using mathematical logic - through the following form

```
   λ<symbolic_identifier>.<expression>
```

So let's take a simple function

   `λx.f[x]` which means a function of the form `x → f[x]` (i.e. process dependent variable `x` using the expression `f[x]`)

Below an example of such a function:

```
   λx.x^2
```

Now let's provide an input to it:

```
   (λx.x^2)4
```

The above would equal `16`.

Now we can even define a lambda expression as follows

```
  (lambda (parameters) body)
```

This is a list with the operator `lambda` processing the elements of the the list `(parameters)` using `body`.

Example:

```
  ((lambda (x y) (+ x y)) 2 3)
```

which will equal `5`.

In Scheme, which is a dialect of LISP you can use an operator called `define` to alias a function via a name to use later with custom values.

```
  (define square (lambda (x) (* x x)))
```

But we can then run a function over a list of number using another operator called `map`:

```
  (map square (list 1 2 3 4 5))
```

This would give me the following result:

```
  (1 4 9 16 25)
```

We can even reduce (by accumulating) through `foldl` (fold-left) operator:

```
  (foldl + 0 (list 1 2 3 4))
```

It takes the value 0 and then adds 1 and proceeds until the end of the list, which eventually gives us `(0 + (1 + (2 + (3 + (4 + 5)))))  = 10`.

  Thus now you can see how you can move functions over data, and next I'll cover how to provide context to functions (think of them as methods in class objects that would be reflective of the class - basically the inspection of the type of class).  This way you can reference functions, and operate functions with functions.

#### Functors, Applicative Functors (Applicatives) and Monads

So functors come from a branch of mathematics called [Category Theory](https://en.wikipedia.org/wiki/Category_theory), where if you let's say have two categories, A and B, a functor F maps objects in C to objects in D.  In terms of programming languages you can think of it as a type.  You have a function that can take an object of type A and map it as an object type B.

Applicative functors would be functions that are aware of the context (types) they are able to process.

Monads basically change the context (type) of the return type.  Think of Scala having Option where it can be None (i.e. empty) or Some (having some value).

#### Abstract vs Concrete

Now if a type is abstract it requires definitions or would not be able to be processed.  It can be referenced but not used like a concrete type.  The applicative functors require the concrete context of how to act.  If there is no definition, then they are lost on how to go forward.

Hope this overview helps in elucidating any worries/confusions/etc.  Feel free to ask if anything requires more clarification, as I would be glad to provide them.

My philosophy is don't be afraid to ask anything of anyone, no matter how insignificant it might seem to anyone.  I'm always in favor of as many folks to know as much as possible, and to have them feel they have a grasp of the subject if they want to.  I'm always pleasantly surprised at what folks can do once they know more ;)

Cheers,
Paul (pgrosu@gmail.com)

File: supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystemAsyncJobExecutionActor.scala [hunk 17]
```diff
@@ -0,0 +1,299 @@
...
+    jobLogger.info(s"`$script`")
+    jobPaths.callRoot.createDirectories()
+    writeScript(script, scriptDir.toAbsolutePath)
+    jobLogger.info(s"command: $processArgs")
+    val stdoutWriter = stdoutSubmit
+    val stderrWriter = stderrSubmit
+    val argv = processArgs.argv.map(_.toString)
+    val proc = argv.run(ProcessLogger(stdoutWriter.writeWithNewline, stderrWriter.writeWithNewline))
+    val run = getJob(proc, stdoutWriter, stderrWriter)
+    // TODO: If returning a future is it ok to change the context.become in a future?
```
mcovarr: not sure what you mean here

supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystemJobExecutionActor.scala
```diff
@@ -0,0 +1,43 @@
...
+    def sharedRunning(executor: ActorRef): Receive = {
+      case AbortJobCommand =>
+        executor ! AbortJobCommand
+      case abortResponse: AbortedResponse =>
+        context.parent ! abortResponse
+        context.stop(self)
+    }
+
+    override def execute: Future[BackendJobExecutionResponse] = {
+      // Still not sure why the AsyncJobExecutionActor doesn't wait for an Akka message instead of using Scala promises
```
mcovarr: Could you elaborate on this?

build.sbt
kshakir: I'm going to need a suggestion. It is both a abstract supertype _and_ a usable backend. The config backend is in there. And AFAIK, we will not have an LSF backend as a artifact. I hope to remove the SGE backend artifact in a commit soon, and just have the memory values be fed into a WDL `instantiateCommand`.

core/src/test/resources/application.conf
kshakir: If they aren't, then I believe the test would have failed already, as the main conf would have overwritten the test conf values with original, wrong, values.

supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystemAsyncJobExecutionActor.scala
kshakir: I'm aware that one can send messages inside a `Future`, especially via the `pipeTo`. I'm also aware that one cannot access the `sender()` method from inside a `Future`.

I do not know the rules about messing with `context.become` inside a `Future`.

I'm assuming here that it's aok, but if something funky starts happening or throwing an exception, I'm going to look at this line.

supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystemJobExecutionActor.scala
kshakir: AsyncBackendJobExecutionActor (ABJEA) (I missed "Backend" in my comment) uses a `completionPromise: Promise` to tell its parent actor, the BackendJobExecutionActor (BJEA), when the async job is done.

I doesn't seem obvious why we're passing around references to `scala.concurrent.Promise`s when akka has a perfectly good message system already.

Worst case, if one needed two different states and didn't want the framework to use an FSM, the BJEA could `context.become(initializing)`, start the async job, then `context.become(asyncRunning)`, and wait for the message from the ABJEA.

I didn't go down this path because the ABJEA requires the `Promise` for messaging, and I didn't want to start modifying JES, et al.

backend/src/main/scala/cromwell/backend/io/WorkflowPathsBackendInitializationData.scala
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+  * Extension of backend initialization data that also provides a `WorkflowPaths`, and by proxy its `List[FileSystem]`.
+  *
+  * This class consolidates code used by the legacy Local backend and the SFS backend, that need to localize data
+  * from a collection of different file systems (local and GCS for now).
+  *
+  * NOTE: The JES backend uses a fork of the `WorkflowPaths`. As `JesWorkflowPaths` is not a subclass, the
+  * initialization data for JES does not use this trait nor the companion object. It is not clear if JES needs the
+  * functionality, as it cannot localize from generic file systems, only GCS with different authentication modes.
```
mcovarr: Can we ticket or TODO looking into unifying `WorkflowPath` with `JesWorkflowPaths`?

backend/src/main/scala/cromwell/backend/io/WorkflowPathsBackendInitializationData.scala
```diff
@@ -0,0 +1,51 @@
...
+  * type downcast to an `Option[BackendInitializationData]`.
+  *
+  * The factory then sends a copy of the intialization data down to each execution actor created for every `wdl4s.Call`.
+  *
+  * Each instance of the execution actor will unpack the initialization data, obtaining the workflow paths using
+  * `WorkflowPathsBackendInitializationData.workflowPaths(Option[BackendInitializationData]): WorkflowPaths`
+  *
+  * However, the execution instances are most interested in actually getting to the file systems embedded within the
+  * workflow paths. That functionality is provided via `WorkflowPathsBackendInitializationData.workflowPaths`
+  */
```
mcovarr: great docs!

backend/src/main/scala/cromwell/backend/validation/FailOnStderrValidation.scala
```diff
@@ -0,0 +1,36 @@
...
+
+  override def key = RuntimeAttributesKeys.FailOnStderrKey
+
+  override def coercion = Seq(WdlBooleanType, WdlStringType)
+
+  override protected def validateValue = {
+    case WdlBoolean(value) => value.successNel
+    case WdlString(value) if value.toLowerCase == "true" => true.successNel
+    case WdlString(value) if value.toLowerCase == "false" => false.successNel
+  }
```
mcovarr: hmm looks like we could have a (composable) PF for boolean or Stringified boolean validation

backend/src/main/scala/cromwell/backend/validation/ContinueOnReturnCodeValidation.scala
```diff
@@ -0,0 +1,53 @@
...
+
+  lazy val default = instance.withDefault(WdlInteger(0))
+
+  lazy val optional = default.optional
+
+  val missingMessage =
+    s"Expecting $key runtime attribute to be either a Boolean, a String 'true' or 'false', or an Array[Int]"
+}
+
+trait ContinueOnReturnCodeValidation extends RuntimeAttributesValidation[ContinueOnReturnCode] {
```
mcovarr: just curious, why is this a trait?

kshakir: Saw the concept somewhere else and copied it. I didn't need them to be a full class, nor even an abstract, so I left it a trait.

backend/src/main/scala/cromwell/backend/BackendJobExecutionActor.scala
```diff
@@ -56,28 +54,15 @@ trait BackendJobExecutionActor extends BackendJobLifecycleActor with ActorLoggin
   /**
     * Restart or resume a previously-started job.
     */
-  def recover: Future[BackendJobExecutionResponse]
+  def recover: Future[BackendJobExecutionResponse] = execute
```
geoffjentry: Nice. I said in standup today that I was going to add a ticket for exactly this and you beat me to it

geoffjentry: Oops, hit the button too quickly. I do think this would be better the way HtCondor does it though (assuming it's not a PITA to do so) where it's logging that it doesn't have recover functionality and is executing instead

backend/src/main/scala/cromwell/backend/BackendJobExecutionActor.scala
```diff
@@ -56,28 +54,15 @@ trait BackendJobExecutionActor extends BackendJobLifecycleActor with ActorLoggin
...
 
   /**
     * Abort a running job.
     */
-  def abort: Unit
+  def abort(): Unit = {}
```
geoffjentry: Same comment as above, it'd be good IMO to log what's happening here

backend/src/main/scala/cromwell/backend/ExecutionHash.scala
geoffjentry: I'm just picturing a year from now when people see all of these "TODO PBE" and wonder WTF it means. "Who were you denvercoder9? What did you see?!?"

backend/src/main/scala/cromwell/backend/io/WorkflowPathsBackendInitializationData.scala
```diff
@@ -0,0 +1,51 @@
+package cromwell.backend.io
+
+import java.nio.file.FileSystem
+
+import cromwell.backend.BackendInitializationData
+
+/**
+  * Extension of backend initialization data that also provides a `WorkflowPaths`, and by proxy its `List[FileSystem]`.
+  *
+  * This class consolidates code used by the legacy Local backend and the SFS backend, that need to localize data
```
geoffjentry: Can you spell out shared file system (assuming that's what it is, if not, it helps make my point) somewhere in here instead of the acronym?

backend/src/main/scala/cromwell/backend/io/WorkflowPathsBackendInitializationData.scala
geoffjentry: Indeed. Speaking as the chief offender I think this is a direction we should be trying to guide ourselves

build.sbt
mcovarr: Oops nm then, suggestion withdrawn.  I hadn't realized SFS is directly usable.

core/src/main/resources/application.conf
mcovarr: yo dog

backend/src/main/scala/cromwell/backend/validation/ContinueOnReturnCodeValidation.scala
geoffjentry: super duper TOL, even for me, but the current trend is towards preferring abstract class unless you need a trait - e.g. http://stackoverflow.com/a/35251513

I'm trying to go this way myself, although I've said I'd never pick on someone in a code review. The only reason I'm making this comment was due to the way you phrased your response (which seemed to imply that abstract class was higher, instead of lower - or whichever direction one wants to go). I don't care if you change it or not.

backend/src/main/scala/cromwell/backend/validation/FailOnStderrValidation.scala
```diff
@@ -0,0 +1,36 @@
...
+
+  lazy val default = instance.withDefault(WdlBoolean(false))
+
+  lazy val optional = default.optional
+
+  val missingMessage =
+    s"Expecting $key runtime attribute to be a Boolean or a String with values of 'true' or 'false'"
+}
+
+trait FailOnStderrValidation extends RuntimeAttributesValidation[Boolean] {
```
geoffjentry: If these are intended to be plugged in for new backend implementations, could you put a little bit (doesn't need to be nearly as robust as the other comments I asked for) on how to use them?

supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystemAsyncJobExecutionActor.scala
mcovarr: Ah ok.  Yeah that seems fishy doesn't it.

build.sbt
```diff
@@ -5,13 +5,13 @@ lazy val core = (project in file("core"))
...
 
-lazy val gcsfilesystem = (project in file("filesystems/gcs"))
+lazy val gcsFileSystem = (project in file("filesystems/gcs"))
   .settings(gcsFileSystemSettings:_*)
   .withTestSettings
 
 lazy val database = (project in file("database"))
   .settings(databaseSettings:_*)
-  .dependsOn(core % "test->test")
+  .dependsOn(core % "test->test") // TODO: PBE: Remove once PostMVP tag removed from test
```
geoffjentry: Similar comment which I already know you disagree with, I don't see the point anymore for having todo's tagged as PBE - IMO 'PBE' is a completed project

mcovarr: it is??

core/src/test/resources/application.conf
geoffjentry: I believe it's base on the classpath order itself, such that earlier in the classpath has precedence. 

FWIW I believe they also have merge strategies but I've never played around w/ it

engine/src/main/scala/cromwell/engine/backend/BackendConfiguration.scala
```diff
@@ -28,12 +29,13 @@ object BackendConfiguration {
...
       entry.getString("actor-factory"),
-      entry.getConfig("config")
+      entry.getConfigOr("config")
     )
   }
 
   val DefaultBackendEntry: BackendConfigurationEntry = AllBackendEntries.find(_.name == DefaultBackendName) getOrElse {
-    throw new IllegalArgumentException(s"Could not find specified default backend name '$DefaultBackendName'.")
+    throw new IllegalArgumentException(s"Could not find specified default backend name '$DefaultBackendName' " +
+      s"in '${AllBackendEntries.mkString("', '")}'.")
```
geoffjentry: wtf is up w/ that red comma?

engine/src/test/scala/cromwell/util/docker/SprayDockerRegistryApiClientSpec.scala
```diff
@@ -43,12 +43,14 @@ class SprayDockerRegistryApiClientSpec extends TestKitSuite("SprayDockerRegistry
   }
 
   it should "resolve docker hub image v1 layer ids" taggedAs IntegrationTest in {
+    val dockerRepo = "registry" // "ubuntu" stopped working... only hub.docker.com knows why
```
geoffjentry: dockerhub giveth & dockerhub taketh

supportedBackends/local/src/main/scala/cromwell/backend/impl/shadowlocal/ShadowLocalBackendLifecycleActorFactory.scala
```diff
@@ -0,0 +1,79 @@
...
+
+import cromwell.backend.BackendConfigurationDescriptor
+import cromwell.backend.sfs._
+import cromwell.backend.validation.DockerValidation
+import cromwell.core.PathFactory._
+import cromwell.core.PathWriter
+
+import scala.sys.process.Process
+
+class ShadowLocalBackendLifecycleActorFactory(override val configurationDescriptor: BackendConfigurationDescriptor)
```
geoffjentry: Another request for some doc/comments

supportedBackends/local/src/main/scala/cromwell/backend/impl/shadowlocal/ShadowLocalBackendLifecycleActorFactory.scala
```diff
@@ -0,0 +1,79 @@
...
+    new SharedFileSystemCommand(Seq("/bin/bash", "-c", s"cat ${jobPaths.script} | $dockerRun /bin/bash <&0"))
+  }
+
+  override lazy val stdoutSubmit = jobPaths.stdout.untailed
+
+  override lazy val stderrSubmit = jobPaths.stderr.untailed
+
+  override def getJob(process: Process, stdoutWriter: PathWriter, stderrWriter: PathWriter) = {
+    val p = ShadowLocalBackendLifecycleActorFactory.simpleProcessFieldP.get(process)
+    val pid = ShadowLocalBackendLifecycleActorFactory.unixProcessFieldPid.get(p)
```
geoffjentry: This reminded me of a Q that came up at standup. These changes are just to prevent blocking a thread and not to enable recover, right?

supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystemAsyncJobExecutionActor.scala
geoffjentry: I'm a giant -1 on that idea. IMO changing the receive method inside a Future is just as bad, if not worse than dealing with an Actor's mutable state from inside a Future

Also in general if people are screwing around w/ become, I'd rather we use the FSM DSL since that's our MO

supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystemJobExecutionActor.scala
geoffjentry: That sounds worthy of a ticket to at least investigate, although I agree it's not something you need to fiddle with here

build.sbt
geoffjentry: I consider it to be. I can plug in a backend and it works.

geoffjentry:
👍

supportedBackends/local/src/main/scala/cromwell/backend/impl/shadowlocal/ShadowLocalBackendLifecycleActorFactory.scala
kshakir: It's both.

Will reply to rest of comments later, but I _do_ dream of a local backend that can recover. 😉

geoffjentry: That's why I asked ... there was dissension at standup as to what the expected behavior would be in the first place :)

engine/src/main/scala/cromwell/engine/backend/BackendConfiguration.scala
kshakir: I suspect github's syntax highlighter [had a stroke](http://stackoverflow.com/questions/30667675/what-does-a-red-text-background-mean-in-github-comparison-view), possibly due to all the quoting: `"'"','"'"`

supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystemAsyncJobExecutionActor.scala
kshakir: Sorry for the false alarm with the comment. I'll remove it.

This is actually calling `context.become` _during_ the `receive`, not asynchronously in another `Runnable`. To meet the contract with our base classes, the result of the execution is mapped to an already completed `Future`.

supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystemJobExecutionActor.scala
cjllanwarne: @kshakir Upvote x 100

supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystemValidatedRuntimeAttributesBuilder.scala
```diff
@@ -0,0 +1,74 @@
...
+  *
+  * The SFS will also _always_ validate using the `DockerValidation`, but may end up warning the user that the attribute
+  * is unsupported by the backend implementation.
+  */
+object SharedFileSystemValidatedRuntimeAttributesBuilder {
+
+  lazy val default: SharedFileSystemValidatedRuntimeAttributesBuilder = {
+    val custom = Seq.empty
+    val required = Seq(ContinueOnReturnCodeValidation.default, FailOnStderrValidation.default)
+    val extra = Seq(DockerValidation.optional)
```
mcovarr: In performing the hackery over on my branch I realized I don't understand the differences among these three groupings of validations.

mcovarr:
Admittedly I don't understand all of it but I'm :+1: whenever you think it's ready.

supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystemValidatedRuntimeAttributesBuilder.scala
kshakir: I can/should add more docs if it's not clear what the goal of the `SFSVRAB.default` is doing here.

I did some docs above regarding `default`, and the ability to add other validations via `withValidation`. Also `Docker`, the sometimes extra validation, may toggle between required & extra, depending on the backend implementation.

backend/src/main/scala/cromwell/backend/ExecutionHash.scala
cjllanwarne: I've added a #1221 to catch any remaining before we officially end 0.21

PR #1213, opened 2016-07-27
Report preempted calls as "Preempted" in metadata Closes #1207

### LINKED ISSUE #1207: Cromwell sets executionStatus to 'Failed' on preemption
Previously (cromwell 0.19 and before), Cromwell would set the executionStatus to 'Preempted' instead of `Failed`

Horneth:


mcovarr:
:+1:

geoffjentry:
👍 

Does this need to be on hotfix as well? @jsotobroad @kcibul

geoffjentry:
Actually I think @vivster7 was the person to tag from green

vivster7:
Nope. Hot fix works as expected. This bug was found while testing .20

geoffjentry:
Ah. Carry on then!

PR #1214, opened 2016-07-28
Cromwell initialization Calls Order. Closes #1215.

### LINKED ISSUE #1215: Calls Map during initialization is not ordered

jainh:
This change ensure during initialization of the Backedn the sequence of calls to are in an order as specified in the WDL file. As we are doing `groupBy()` that is changing the order but some of the backend require calls to be in an order.

Formal reviewer @geoffjentry @francares

### DIFF
--- a/engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
@@ -18,6 +18,7 @@ import wdl4s._
 import wdl4s.expression.NoFunctions
 import wdl4s.values.{WdlString, WdlValue}
 
+import scala.collection.immutable.ListMap
 import scala.language.postfixOps
 import scala.util.{Failure, Success, Try}
 import scalaz.Scalaz._
@@ -179,7 +180,7 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef, val wor
 
   private def validateBackendAssignments(calls: Seq[Call], workflowOptions: WorkflowOptions, defaultBackendName: Option[String]): ErrorOr[Map[Call, String]] = {
     val callToBackendMap = Try {
-      calls map { call =>
+      ListMap(calls map { call =>
         val backendPriorities = Seq(
           workflowOptions.get(RuntimeBackendKey).toOption,
           assignBackendUsingRuntimeAttrs(call),
@@ -191,7 +192,7 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef, val wor
           case Some(backendName) => throw new Exception(s"Backend for call ${call.fullyQualifiedName} ('$backendName') not registered in configuration file")
           case None => throw new Exception(s"No backend could be found for call ${call.fullyQualifiedName}")
         }
-      } toMap
+      }:_*)
     }
 
     callToBackendMap match {


cjllanwarne:
OOI, what's the use-case for this? I'm not saying this code change is bad (as a code change it looks totally fine) but the description is flagging up warning signs in my head.

The WDL spec claims that workflows are robust to calls being specified in **any order** since the DAG is 100% implied by inter-call dependencies rather than list-order. Eg this should be fine:

```
workflow x {
 call b { input: i = a.i }
 call a
```

geoffjentry:
@jainh To echo what @cjllanwarne said, there is *no* implied order in WDL, it's a pure dataflow. Any backend which requires an ordering beyond the dependency graph is implementing things incorrectly.

jainh:
@cjllanwarne  As you mentioned "order is defined in WDL and DAG gets created based on inter-dependencies" is true during execution of the task but not during initialization phase, hence that order is required during initialization as well. 

@geoffjentry If i understand correctly there is no graph being received by the backend during initialization it is `Map[Call, String]` to identifies all calls @ backend and then ultimately `Seq[Call]`, what is expected from engine if user specifies calls as @cjllanwarne mentioned in above wdl example (a then b) that order is preserved and pass through. 

RP backend in CCC depends upon order of calls during initialization as well therefore that change is required.

cjllanwarne:
@jainh I would therefore suggest you want to sort the list yourself to make that guarantee robust, rather than assume it's already correctly ordered in the WDL

jainh:
@cjllanwarne Sorting the `Seq[call]` ? how that would work

geoffjentry:
@jainh Just so I'm clear, the WDL example that @cjllanwarne posted above would still work correctly, right? (I believe that's the case, just being cautious here)

engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -179,7 +180,7 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef, val wor
 
   private def validateBackendAssignments(calls: Seq[Call], workflowOptions: WorkflowOptions, defaultBackendName: Option[String]): ErrorOr[Map[Call, String]] = {
     val callToBackendMap = Try {
-      calls map { call =>
+      ListMap(calls map { call =>
```
geoffjentry: Could you rework this so you're not passing a 10 line expression into the constructor of ListMap?

cjllanwarne:
@jainh Why do the calls need to be in any order at all by your backend? Why does your backend care?

If it's because you need to process ancestor calls before you process their descendant calls then **the order of calls in the WDL is insufficient** because WDL allows calls to be specified out of chronological order

jainh:
@cjllanwarne 
Here are two scenarios 
case a)
 ```
 workflow x {
 call c {input:j= b.i}
 call b { input: i = a.i }
 call a
 }
 ```
Call sequence for above workflow during execution is `a->b->c` but for initialization it could be `a->c->b` from the existing source. But what make sense for any backend to receive calls in an order `a->b->c` irrespective of what stage it is at during life cycle.

case b)
```
 workflow x {
 call c 
 call b 
 call a
}
```
Call sequence for above workflow during execution can be  a:b:c in parallel or in any random order as in we don't anticipate order, as it is in out of chronological order. 

So what is missing is for case a).

engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
jainh: yes.

cjllanwarne:
@jainh that's what I mean - if you just make a ListMap in case (a) then you'll just end up with them in the list in the order c, b, a because that's the WDL order. Instead you want to sort them into the order a, b, c

jainh:
@cjllanwarne Sort does not solve the purpose here because it is not what backend expecting and want,However it is expecting the right dependencies  as specified in the wdl.

cjllanwarne:
@jainh just to be clear. I don't think your code breaks anything but I also don't think it can be comprehensively correct. Here's why- Imagine two nearly the same WDL files:
```
workflow ordered {
 call a
 call b { input: i = a.i }
}
```
and:
```
workflow unordered {
 call b { input: i = a.i }
 call a
}
```

Now, if I understand you correctly, in both cases you want to pass to the initialiser the same list: List(a, b) since a should in initialised before b.

However, if you just create a ListMap as your code does, it will pick the same order from the WDL file. So here's what happens:

| Workflow        | List that you want           | List that your code makes  | Match |
| ------------- |-------------| -----|-----|
| ordered     | `List(a, b)` | `List(a, b)` | 🆗 |
| unordered      | `List(a, b)`      |   `List(b, a)` | 🚫 |

cjllanwarne:
@jainh on the other hand, it's quite possible I'm not understanding your code at all, so if you're happy that it works as you want it, go for it.

cjllanwarne:
Something like this: https://en.wikipedia.org/wiki/Topological_sorting - I would imagine it's non trivial to do this manually but there's probably a library that can do most of the work for you

jainh:
@cjllanwarne I agree now that fix will not work, but what i don't see we can perform Topological sort because we have seq of calls instead dependency graph ( Correct me if i am wrong?).

mcovarr:
You have the dependencies in `Call#prerequisiteCallNames`, so you certainly *could* perform a Topological sort.  But TBH a backend having a requirement for a topologically ordered list of calls seems kind of sketchy.

francares:
@mcovarr, @cjllanwarne agree on your comments, I implemented a PoC on graph topology creation using topological sort algorithm based on DFS calculation at the beginning of this year. The idea was to propose the integration of that concept to Cromwell after PBE.
Coming back to this issue, If we need to recreate a DAG in the backend side it means there is something wrong with that backend (we know it). I explained why we try to do so in the Waffle.io ticket but I think we can explain better in the next meeting we may have. Anyways @jainh will not proceed with this change.

PR #1222, opened 2016-07-29
Changed liquibase xml to name constraint.

kshakir:
Added list of Dos and Don'ts for liquibase schemas.
Synced clob types between slick and liquibase.


ruchim:
👍  Super helpful

cjllanwarne:
👍

PR #1223, opened 2016-07-29
Address DB write failures for WorkflowStore and JobStore

cjllanwarne:
These were already handled as messages so mainly a no-op. I did do some tidying up around the edges though

### DIFF
--- a/engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
@@ -224,7 +224,7 @@ class WorkflowManagerActor(config: Config,
       log.error("Error registering JobStore completion for {}: {}", arg1 = wfId, arg2 = t)
       // This is minorly bad. The JobStore would never rid itself of this workflow's entries. Unlikely to be a big deal
       // but might as well try again in a few minutes...
-      context.system.scheduler.scheduleOnce(10 minutes, sender, failedOperation)
+      context.system.scheduler.scheduleOnce(10 minutes, sender, failedOperation)(context.dispatcher)
       stay()
     // Anything else certainly IS interesting:
     case Event(unhandled, data) =>
--- a/engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
@@ -21,6 +21,10 @@ import scala.concurrent.{Await, Promise}
 import scala.language.postfixOps
 import scalaz.NonEmptyList
 
+import akka.actor.Props;
+import scala.concurrent.duration.Duration;
+import java.util.concurrent.TimeUnit;
+
 object WorkflowManagerActor {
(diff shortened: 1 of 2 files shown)


mcovarr:
:+1: delta getting the code to compile 😄

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -115,6 +115,9 @@ class EngineJobExecutionActor(jobKey: BackendJobDescriptorKey,
       context stop self
       stay()
     case Event(JobStoreWriteFailure(_, t), _) =>
+      // This is moderately bad: If we can't write this result to the database, and then undergo a system restart, we'd need
+      // to re-run this job.
```
geoffjentry: Would this job ever have appeared to be complete to an outside party?

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -215,8 +220,11 @@ class WorkflowManagerActor(config: Config,
     // Uninteresting transition and current state notifications.
     case Event((Transition(_, _, _) | CurrentState(_, _)), _) => stay()
     case Event(JobStoreWriteSuccess(_), _) => stay() // Snoozefest
-    case Event(JobStoreWriteFailure(_, t), _) =>
-      log.error("Error writing to JobStore from WorkflowManagerActor: {}", t)
+    case Event(JobStoreWriteFailure(failedOperation @ RegisterWorkflowCompleted(wfId), t), _) =>
+      log.error("Error registering JobStore completion for {}: {}", arg1 = wfId, arg2 = t)
+      // This is minorly bad. The JobStore would never rid itself of this workflow's entries. Unlikely to be a big deal
+      // but might as well try again in a few minutes...
+      context.system.scheduler.scheduleOnce(10 minutes, sender, failedOperation)(context.dispatcher)
```
geoffjentry: I'm a little concerned about cruft building up for reasons which would actually alarm us but we never notice. OTOH firing a log message is probably not enough for us to notice it is happening either so YMMV

geoffjentry:
👍

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
mcovarr: I suppose it could appear complete in the metadata

geoffjentry: The "need to re-run this job" mildly bothered me if an outside entity would never see it as complete, but if it is viewed as complete and gets re-run I don't think we can let that happen.

cjllanwarne: I guess the alternative is to hold back the `response` until we get the JobStoreAck, and if we fail to write a success to the JobStore, we actually fail the job?

PR #1225, opened 2016-07-29
CallCacheActor skeleton strawman

cjllanwarne:
Something of a strawman commit. Very roughly maps out how I think we said the CallCacheActor would fit into the existing actor model?

### DIFF
--- a/backend/src/main/scala/cromwell/backend/BackendJobExecutionActor.scala
+++ b/backend/src/main/scala/cromwell/backend/BackendJobExecutionActor.scala
@@ -1,9 +1,11 @@
 package cromwell.backend
 
-import akka.actor.ActorLogging
+import akka.actor.{ActorLogging, ActorRef}
 import akka.event.LoggingReceive
 import cromwell.backend.BackendJobExecutionActor._
 import cromwell.backend.BackendLifecycleActor._
+import cromwell.backend.callcaching.{CallCachingActor, SyncCallCachingActor}
+import cromwell.backend.callcaching.CallCachingActor._
 import cromwell.core.JobOutputs
 import wdl4s.expression.WdlStandardLibraryFunctions
 import wdl4s.values.WdlValue
@@ -33,8 +35,17 @@ object BackendJobExecutionActor {
   */
 trait BackendJobExecutionActor extends BackendJobLifecycleActor with ActorLogging {
 
+  val callCacheActor = context.actorOf(SyncCallCachingActor.props(jobDescriptor, _ => "Implement me!", _ => "Implement me!", readFromCache = false, writeToCache = false))
+
+  private var executeClient: Option[ActorRef] = None
+
   def receive: Receive = LoggingReceive {
-    case ExecuteJobCommand => performActionThenRespond(execute, onFailure = executionFailed)
+    case ExecuteJobCommand =>
+      executeClient = Option(sender)
+      callCacheActor ! CheckCache
+    case CallCacheHit(backendJobExecutionResponse: BackendJobExecutionResponse) => executeClient foreach { _ ! backendJobExecutionResponse }
+    case CallCacheMiss => performActionThenRespond(executeWithCacheWriteOn, onFailure = executionFailed, respondTo = executeClient.get)
+
     case RecoverJobCommand => performActionThenRespond(recover, onFailure = executionFailed)
     case AbortJobCommand =>
       abort()
@@ -45,6 +56,13 @@ trait BackendJobExecutionActor extends BackendJobLifecycleActor with ActorLoggin
   // We need this for receive because we can't do `onFailure = ExecutionFailure` directly - because BackendJobDescriptor =/= BackendJobDescriptorKey
   private def executionFailed = (t: Throwable) =>
     FailedNonRetryableResponse(jobKey = jobDescriptor.key, throwable = t, returnCode = None)
+
+  private def executeWithCacheWriteOn = execute map {
+    case x @ SucceededResponse(jobKey: BackendJobDescriptorKey, returnCode: Option[Int], jobOutputs: JobOutputs) =>
+      callCacheActor ! WriteCache(returnCode, jobOutputs)
+      x
+    case other => other
+  }
 
   /**
     * Execute a new job.
--- a/backend/src/main/scala/cromwell/backend/BackendJobExecutionActor.scala
+++ b/backend/src/main/scala/cromwell/backend/BackendJobExecutionActor.scala
@@ -44,7 +44,9 @@ trait BackendJobExecutionActor extends BackendJobLifecycleActor with ActorLoggin
       executeClient = Option(sender)
       callCacheActor ! CheckCache
     case CallCacheHit(backendJobExecutionResponse: BackendJobExecutionResponse) => executeClient foreach { _ ! backendJobExecutionResponse }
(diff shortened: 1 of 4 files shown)


geoffjentry:
I just skimmed this, but it looks like it's more or less what I was talking about when we chatted earlier. 

OTOH the reason I further delayed writing up the ticket was I had some thoughts when I was writing them that I thought were worth talking about, but no one is here at the moment :)

geoffjentry:
I'm guessing that the "people aren't around" thing isn't going to change at 5pm on a Friday, so ....

I'm thinking now that instead, perhaps the thing to do is to also have the actor know if read is on and if write is on. When the actor spins up, it ...

- If read is on, just starts determining the cache hit status w/o being asked
- If write is on, starts hashing everything
- If both are on, starts hashing everything but is checking for cache hit until a miss occurs, at which point it's just generating hashes

The actor could then receive two messages, one is "are you a cache hit" and the other is "please now persist to the store" (if you asked for cache hit status before it was done, perhaps it could send back a "come back later" message - that way you can avoid returning the Futures, although perhaps people like that)

We know for a fact that if read is on that we'll be checking the cache, so might as well start that ASAP. If write is on, we're potentially wasting energy - e.g. if a job fails, but presumably (hopefully!) job successes are more common than job failures and we can get a jumpstart on the process.

francares:
@geoffjentry and @cjllanwarne Could you provide a link to CallCache design documentation?... Thanks.

cjllanwarne:
@geoffjentry I like that - don't layer on more futures, just start working and either say "ready and here's the answer" or "not ready yet, ask me later". It's also much more RESTy (I refuse to say `RESTful`) like that, which is probably a good sign

cjllanwarne:
Just FYI, the `BackendJobExecutionActor` is still a work in progress ATM. Changes incoming

backend/src/main/scala/cromwell/backend/BackendJobExecutionActor.scala
```diff
@@ -33,8 +35,17 @@ object BackendJobExecutionActor {
   */
 trait BackendJobExecutionActor extends BackendJobLifecycleActor with ActorLogging {
 
+  val callCacheActor = context.actorOf(CallCachingActor.props(jobDescriptor, _ => "Implement me!", _ => "Implement me!"))
```
mcovarr: Technically those are implemented, they're just terrible hashing functions.

backend/src/main/scala/cromwell/backend/callcaching/CallCachingActor.scala
```diff
@@ -0,0 +1,202 @@
...
+    case Event(CheckCache, _) =>
+      sender ! CallCacheCheckInProgress
+      stay()
+    case Event(CacheHitInternal(cacheResultId), _) if readFromCache =>
+      readEntryFromDatabase(cacheResultId)
+      goto(FetchingJobResult)
+    case Event(CacheHitInternal(_), _) if !readFromCache =>
+      goto(UnreadCacheHit)
+    case Event(CacheMissInternal(hashes), _) if writeToCache =>
+      goto(AwaitingJobResultAndAllHashes) using WithHashes(hashes)
```
mcovarr: Not clear why this sets the state data to `WithHashes(hashes)`?  It doesn't look like this is considered in `AwaitingJobResultAndAllHashes`.

backend/src/main/scala/cromwell/backend/callcaching/CallCachingActor.scala
```diff
@@ -0,0 +1,202 @@
...
+    case Event(CacheMissInternal(_), _) if !writeToCache =>
+      goto(UnupdatedCacheMiss)
+  }
+
+  when(AwaitingJobResultAndAllHashes) {
+    case Event(CheckCache, _) =>
+      sender ! CallCacheMiss
+      stay()
+    case Event(WriteCache(returnCode, jobOutputs), WithHashes(hashes)) =>
+      writeEntryToDatabase(returnCode, jobOutputs)
```
mcovarr: maybe include the hashes in what gets written?  😄

backend/src/main/scala/cromwell/backend/callcaching/CallCachingActor.scala
```diff
@@ -0,0 +1,202 @@
...
+      stay()
+    case Event(CheckWriteComplete, _) =>
+      sender ! CallCacheWriteInProgress
+      stay()
+    case Event(WriteCache(_, _), _) =>
+      // For the sake of idempotency, respond to this message too:
+      sender ! CallCacheWriteInProgress
+      stay()
+    case Event(CallCacheWriteSuccessInternal, oldData @ WithJobResult(_, _)) =>
+      goto(NewCacheEntryWritten) using oldData
```
mcovarr: what's going on with `oldData`?

backend/src/main/scala/cromwell/backend/BackendJobExecutionActor.scala
```diff
@@ -33,8 +35,19 @@ object BackendJobExecutionActor {
   */
 trait BackendJobExecutionActor extends BackendJobLifecycleActor with ActorLogging {
 
+  val callCacheActor = context.actorOf(SyncCallCachingActor.props(jobDescriptor, _ => "Implement me!", _ => "Implement me!", readFromCache = false, writeToCache = false))
+
+  private var executeClient: Option[ActorRef] = None
```
francares: Why this var is needed?

backend/src/main/scala/cromwell/backend/callcaching/CallCachingActor.scala
```diff
@@ -0,0 +1,202 @@
...
+    case Event(CheckCache, _) =>
+      sender ! CallCacheMiss
+      stay()
+    case Event(WriteCache(_, _), _) =>
+      sender ! CallCacheWriteSuccess
+      stay()
+  }
+
+  private def calculateHash(): Unit = {
+    // TODO: Do the work and then:
```
francares: The way of calculating hash will be the same for all backends?

backend/src/main/scala/cromwell/backend/callcaching/CallCachingActor.scala
```diff
@@ -0,0 +1,202 @@
...
+    // self ! CacheHitInternal(cacheResultId)
+    // or collate all the hashes and then:
+    self ! CacheMissInternal(Map.empty)
+  }
+
+  private def writeEntryToDatabase(returnCode: Option[Int], jobOutputs: JobOutputs): Unit = {
+    // TODO: Do the work and then:
```
francares: Same question here but regarding database.

backend/src/main/scala/cromwell/backend/callcaching/CallCachingActor.scala
geoffjentry: If a backend chooses to use this call caching & except for runtime attributes and files, yes.

backend/src/main/scala/cromwell/backend/callcaching/CallCachingActor.scala
geoffjentry: Not if a backend chooses to use this call caching. The intention was that it was as on rails as possible. A backend can choose to use this call caching out of the box (as much as possible, some things are inherently backend specific) or roll their own.

cjllanwarne:
This strawman was brave... but it wasn't good enough. Since this no longer represents our current thinking I'm going to take this strawman down.

mcovarr:
RIP strawman 😢

PR #1227, opened 2016-07-29
Added support for WDLFileArrays in HtCondor Docker commands. Closes #1226.

### LINKED ISSUE #1226: [HtCondor] Add support for WDLFileArrays in Docker commands

francares:


supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActor.scala
```diff
@@ -247,19 +248,28 @@ class HtCondorJobExecutionActor(override val jobDescriptor: BackendJobDescriptor
...
     Try {
-      val inputFiles = localizedInputs.filter { case (k,v) => v.wdlType.equals(WdlFileType) }
-      val dockerInputDataVol: Seq[String] = inputFiles.values.map { value =>
-        val limit = value.valueString.lastIndexOf("/")
-        value.valueString.substring(0, limit)
+      val inputFiles = localizedInputs.filter {
+        case (k, v) => v.wdlType == WdlFileType || v.wdlType == WdlArrayType(WdlFileType)
+      }
+      val dockerInputDataVol: Seq[String] = inputFiles.values.flatMap {
+        case file if file.wdlType == WdlFileType =>
```
cjllanwarne: You might be able to use `case WdlFile(filename) =>` and `case WdlArray(WdlArrayType(WdlFileType), fileArray) =>` for better case matching?

cjllanwarne: You could also probably merge the `filter` above and this `flatMap` into some combination of `collect` and `flatten`

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActor.scala
```diff
@@ -247,19 +248,28 @@ class HtCondorJobExecutionActor(override val jobDescriptor: BackendJobDescriptor
...
+          val limit = file.valueString.lastIndexOf("/")
```
gauravs90: I am not sure but I think it's quite possible to have folders as inputs in the `localizedInputs`, no? Would this work in that case? (I believe the parent folder would be mounted in that case). Either way, a comment would be good here.

gauravs90:
Chris' comment. And then 👍

cjllanwarne:
👍 after comments addressed

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActor.scala
francares: Pattern matching cannot be done in that way since WdlFile is an interface. Keeping as before and just changing filter/map/flatmap to collect/flatten.

PR #1240, opened 2016-08-01
default_runtime_attributes in workflow initialization.  Closes #1166

### LINKED ISSUE #1166: 'docker' in default_runtime_attributes is not setting docker container
Run this WDL, but remove the docker portion from the WDL file, and keep it in the options file.

https://github.com/broadinstitute/centaur/tree/develop/src/main/resources/standardTestCases/default_runtime_attributes

mcovarr:
ended up just KISS

### DIFF
--- a/backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
+++ b/backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
@@ -5,15 +5,15 @@ import akka.event.LoggingReceive
 import cromwell.backend.BackendLifecycleActor._
 import cromwell.backend.BackendWorkflowInitializationActor._
 import cromwell.backend.wdl.OnlyPureFunctions
-import cromwell.services.metadata.{MetadataEvent, MetadataKey, MetadataService, MetadataValue}
-import MetadataService.PutMetadataAction
-import cromwell.core.WorkflowMetadataKeys
-import wdl4s.{Call, NoLookup, Task, WdlExpression}
+import cromwell.core.{WorkflowMetadataKeys, WorkflowOptions}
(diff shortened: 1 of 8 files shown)


kshakir:
K.I.S.S. indeed. 👍  to the code diff.

Your comments do help, but I'm still only at about 75% in understanding of what initialization actors can and cannot validate currently. I'm fine if folks file issues with example "enhancements" for the future. One could also write a large suite of tests with runtime attribute expressions that should and should not validate, but that's another ticket to be prioritized.

backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
```diff
@@ -81,20 +91,55 @@ trait BackendWorkflowInitializationActor extends BackendWorkflowLifecycleActor w
...
+    *   time if task (not call!) declarations and inputs are considered.
+    *
+    * - Assuming the previous two checks pass, an attempt should be made to evaluate the expression.  It's possible
+    *   the expression could be a function of call inputs or declarations which are unavailable at initialization time,
+    *   so these failed evaluations must continue to be accepted as possibly correct expressions.
+    *
+    * - Assuming the expression evaluation succeeds, it should be possible to do real business logic validation.
+    *
+    * - It would be nice to memoize as much of the work that gets done here as possible so it doesn't have to all be
+    *   repeated when the various `FooRuntimeAttributes` classes are created, in the spirit of #1076.
```
cjllanwarne: Even better than the spirit of 1776!

cjllanwarne:
👍

mcovarr:
@kshakir I think I might be able to make the comments more clear, I'll give that a try.

@cjllanwarne Can't be better than the Spirit of 1776. 🇺🇸  🎆

PR #1241, opened 2016-08-02
Backend Store performing recovery closes #751

### LINKED ISSUE #751: Recovery support for JES PBE

ruchim:
The BackendKVStore is now communicating with the KeyValueServiceActor.


database/src/main/scala/cromwell/database/sql/BackendKVStoreSqlDatabase.scala
```diff
@@ -1,13 +1,12 @@
...
   this: SqlDatabase =>
 
-  def addBackendKVStoreEntry(entry: BackendKVStoreEntry)(implicit ec: ExecutionContext): Future[Unit]
+  def addBackendJobKeyValuePair(workflowUuid: String, callFqn: String, callIndex: Int, callAttempt: Int, storeKey: String, storeValue: String)(implicit ec: ExecutionContext): Future[Unit]
 
-  def queryBackendKVStoreValueByStoreKey(workflowUid: String, callFqn: String, callIndex: Int, callAttempt: Int, jobKey: String)(implicit ec: ExecutionContext): Future[Option[String]]
+  def queryBackendJobValueByJobKey(workflowUid: String, callFqn: String, callIndex: Int, callAttempt: Int, jobKey: String)(implicit ec: ExecutionContext): Future[Option[String]]
 
-}
+}
```
kshakir: Add a newline here to end the file

ruchim: I'll never learn

services/src/main/scala/cromwell/services/keyvalue/impl/KeyValueServiceActor.scala
```diff
@@ -31,14 +29,27 @@ case class KeyValueServiceActor(serviceConfig: Config, globalConfig: Config) ext
...
-  private def doPut(put: KvPut): Future[KvResponse] = Future.successful {
-    store += put.pair.key -> put.pair.value
-    KvPutSuccess(put)
+  private def doPut(put: KvPut): Future[KvResponse] = {
+    put.pair.value match {
+      case Some(backendVal) => updateBackendKeyValuePair(put.pair.key.workflowId,
+        put.pair.key.jobKey,
+        put.pair.key.key,
+        put.pair.value.get).map(_ => KvPutSuccess(put))
+      case None => Future(KvFailure(put, new Throwable(s"Failed to find the value associated to key: ${put.pair.key.key}. This key cannot be added to the BackendKVStore.")))
```
kshakir: Never create a `new Throwable`. (side rant: anyone know why that class isn't abstract?) It's dangerous for handling code to try and catch them, as `java.lang.Error`, the sibling of `Exception`, shouldn't be normally handled.

Use `new RuntimeException` or even `new Exception` if you must save on typing characters. The other instances in cromwell and wdl4s that are doing this should all be fixed, someday.

services/src/main/scala/cromwell/services/keyvalue/impl/KeyValueServiceActor.scala
```diff
@@ -31,14 +29,27 @@ case class KeyValueServiceActor(serviceConfig: Config, globalConfig: Config) ext
...
+    }
   }
 
-  private def doGet(get: KvGet): Future[KvResponse] = Future.successful {
-    store.get(get.key) match {
-      case Some(maybeValue) => KvPair(get.key, maybeValue)
+  private def doGet(get: KvGet): Future[KvResponse] = {
+    val executionInfo = getBackendValueByKey(
```
kshakir: How about something like `val backendValue` instead of `val executionInfo`?

`executionInfo` reminds me of the olde worlde table "EXECUTION_INFO".

database/src/main/scala/cromwell/database/slick/BackendKVStoreSlickDatabase.scala
```diff
@@ -11,12 +11,20 @@ trait BackendKVStoreSlickDatabase extends BackendKVStoreSqlDatabase {
   import dataAccess.driver.api._
 
 
-  override def addBackendKVStoreEntry(entry: BackendKVStoreEntry)(implicit ec: ExecutionContext): Future[Unit] = {
+  override def addBackendJobKeyValuePair(workflowID: String, callFqn: String, callIndex: Int, callAttempt: Int, storeKey: String, storeValue: String)(implicit ec: ExecutionContext): Future[Unit] = {
+
+    val entry = BackendKVStoreEntry(workflowID.toString,
```
kshakir: workflowID is already a `string` here. And perhaps call it `workflowUuid`?

database/src/main/scala/cromwell/database/slick/BackendKVStoreSlickDatabase.scala
```diff
@@ -11,12 +11,20 @@ trait BackendKVStoreSlickDatabase extends BackendKVStoreSqlDatabase {
...
+                              callAttempt,
+                              storeKey,
+                              storeValue)
+
     val action = dataAccess.backendKVStoreAutoInc += entry
     runTransaction(action) map { _ => () }
   }
 
-  override def queryBackendKVStoreValueByStoreKey(workflowUid: String, callFqn: String, callIndex: Int, callAttempt: Int, jobKey: String)(implicit ec: ExecutionContext): Future[Option[String]] = {
+  override def queryBackendJobValueByJobKey(workflowUid: String, callFqn: String, callIndex: Int, callAttempt: Int, jobKey: String)(implicit ec: ExecutionContext): Future[Option[String]] = {
```
kshakir: `JOB_KEY` and `JOB_VALUE` seem to actually be named `STORE_KEY` and `STORE_VALUE` in the database. For consistency, I would changed the method name back, and then change the parameter name to `storeKey`.

database/src/main/scala/cromwell/database/slick/BackendKVStoreSlickDatabase.scala
```diff
@@ -11,12 +11,20 @@ trait BackendKVStoreSlickDatabase extends BackendKVStoreSqlDatabase {
...
+  override def addBackendJobKeyValuePair(workflowID: String, callFqn: String, callIndex: Int, callAttempt: Int, storeKey: String, storeValue: String)(implicit ec: ExecutionContext): Future[Unit] = {
+
+    val entry = BackendKVStoreEntry(workflowID.toString,
+                              callFqn,
+                              callIndex,
+                              callAttempt,
+                              storeKey,
+                              storeValue)
+
     val action = dataAccess.backendKVStoreAutoInc += entry
```
kshakir: While this action is fine for inserting, I believe the key value store should be update-or-inserting, aka upserting.

```scala
val entry = BackendKVStoreEntry(workflowUuid,
  callFqn,
  callIndex,
  callAttempt,
  storeKey,
  storeValue)
val action = if (useSlickUpserts) {
  // When using slick based upserts, as long as a unique constraint has been created on the columns belonging
  // to the composite-unique-key, then the rest of the values will be upserted appropriately.
  for {
    _ <- dataAccess.backendKVStoreAutoInc.insertOrUpdate(entry)
  } yield () // <-- maps the result to a DBIO[Unit]
} else {
  // If slick can't upsert for us, then we must within one transaction try to update the row first, and if no
  // row was found then we should insert one.
  // NOTE: Because this upsert has a non-zero chance of failing due to race conditions, the caller
  // should always call this upsert using a `cromwell.util.DatabaseUtil.withRetry()`.
  for {
    updateCount <- dataAccess.
      backendJobValueByBackendJobKey(workflowUuid, callFqn, callIndex, callAttempt, storeKey).
      update(storeValue)
    _ <- updateCount match {
      case 0 => dataAccess.backendKVStoreAutoInc += entry
      case _ => assertUpdateCount("addBackendJobKeyValuePair", updateCount, 1)
    }
  } yield () // <-- maps the result to a DBIO[Unit]
}
runTransaction(action)
```

database/src/main/scala/cromwell/database/slick/BackendKVStoreSlickDatabase.scala
kshakir: `workflowUid` is missing a second "u": `workflowUuid`

database/src/main/scala/cromwell/database/slick/tables/DataAccessComponent.scala
```diff
@@ -28,6 +28,6 @@ class DataAccessComponent(val driver: JdbcProfile)
       metadata.schema ++
       workflowMetadataSummaries.schema ++
       workflowStore.schema ++
-      jobStore.schema ++
-      backendKVStore.schema
+      backendKVStore.schema ++
+      jobStore.schema
```
kshakir: Assuming this was an unintentional change, unless you really care about the order of the two schemas, then revert this diff.

database/src/main/scala/cromwell/database/slick/WorkflowStoreSlickDatabase.scala
```diff
@@ -24,7 +24,7 @@ trait WorkflowStoreSlickDatabase extends WorkflowStoreSqlDatabase {
                                      (implicit ec: ExecutionContext):
   Future[Seq[WorkflowStoreEntry]] = {
 
-    val action = for {
+    val action: DBIO[Seq[WorkflowStoreEntry]] = for {
```
kshakir: Revert unless you're going to type _all_ the `action`s.

services/src/main/scala/cromwell/services/keyvalue/KeyValueService.scala
```diff
@@ -19,4 +19,3 @@ object KeyValueService {
   case class KvKeyLookupFailed(action: KvGet) extends KvResponse
   case class KvPutSuccess(action: KvPut) extends KvResponse
 }
-
```
kshakir: While I normally appreciate most import optimizations during code edits, I don't like PRs with other changes "cleaning" random files-- causing possible merge conflicts, git blame misattribution, etc. So as a reviewer I'm requesting a revert of this file unless you have actual code to contribute to the diff.

database/src/main/scala/cromwell/database/sql/BackendKVStoreSqlDatabase.scala
```diff
@@ -1,13 +1,12 @@
...
 import scala.concurrent.{ExecutionContext, Future}
 
 trait BackendKVStoreSqlDatabase {
   this: SqlDatabase =>
 
-  def addBackendKVStoreEntry(entry: BackendKVStoreEntry)(implicit ec: ExecutionContext): Future[Unit]
+  def addBackendJobKeyValuePair(workflowUuid: String, callFqn: String, callIndex: Int, callAttempt: Int, storeKey: String, storeValue: String)(implicit ec: ExecutionContext): Future[Unit]
 
-  def queryBackendKVStoreValueByStoreKey(workflowUid: String, callFqn: String, callIndex: Int, callAttempt: Int, jobKey: String)(implicit ec: ExecutionContext): Future[Option[String]]
+  def queryBackendJobValueByJobKey(workflowUid: String, callFqn: String, callIndex: Int, callAttempt: Int, jobKey: String)(implicit ec: ExecutionContext): Future[Option[String]]
```
kshakir: Same notes here as the slick based override.

kshakir:
👍  delta the comments above. Re-assign back if there are changes that you would like re-reviewed.

database/src/main/scala/cromwell/database/slick/BackendKVStoreSlickDatabase.scala
ruchim: thanks for all that documentation! good stuff

services/src/main/scala/cromwell/services/keyvalue/KeyValueService.scala
cjllanwarne: Perhaps a massive side-comment: Import optimisation from IntelliJ is a nightmare waiting to happen.

IntelliJ is awful at working implicit imports backwards and I had to turn off auto-import-correction when it trashed all of the Slick JSON implicit imports and I spent hours trying to determine which combination of imports was required in all the files which were "corrected"

scottfrazer:
👍

PR #1250, opened 2016-08-04
Recovery functionality for HtCondor backend. Closes #1249.

### LINKED ISSUE #1249: [HtCondor] Add recovery functionality

francares:
- Added recovery functionality using KV service.
- In the next iteration will refactor to use a Doc store (Mongo, Couchbase) generic service implementation or continue using KV service but with a refactor in order to support not just SQL DBs as KV store but any other kind of DB. I think the best may be to work on a DAL or if it's not possible just modify the service to support other providers.

Let me know what do you think on this.

### DIFF
--- a/supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActor.scala
+++ b/supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActor.scala
@@ -3,12 +3,13 @@ package cromwell.backend.impl.htcondor
 import java.nio.file.attribute.PosixFilePermission
 import java.nio.file.FileSystems
 
-import akka.actor.Props
+import akka.actor.{ActorRef, Props}
 import cromwell.backend.BackendJobExecutionActor.{BackendJobExecutionResponse, FailedNonRetryableResponse, SucceededResponse}
 import cromwell.backend._
 import cromwell.backend.impl.htcondor.caching.CacheActor._
 import cromwell.backend.io.JobPaths
 import cromwell.backend.sfs.{SharedFileSystem, SharedFileSystemExpressionFunctions}
+import cromwell.services.keyvalue.KeyValueService._
 import org.apache.commons.codec.digest.DigestUtils
 import wdl4s._
 import wdl4s.parser.MemoryUnit
@@ -23,14 +24,17 @@ import scala.util.{Failure, Success, Try}
 import scala.language.postfixOps
 
 object HtCondorJobExecutionActor {
+  val HtCondorJobIdKey = "htCondor_job_id"
+
   val fileSystems = List(FileSystems.getDefault)
 
-  def props(jobDescriptor: BackendJobDescriptor, configurationDescriptor: BackendConfigurationDescriptor, cacheActorProps: Option[Props]): Props =
-    Props(new HtCondorJobExecutionActor(jobDescriptor, configurationDescriptor, cacheActorProps))
+  def props(jobDescriptor: BackendJobDescriptor, configurationDescriptor: BackendConfigurationDescriptor, serviceRegistryActor: ActorRef, cacheActorProps: Option[Props]): Props =
+    Props(new HtCondorJobExecutionActor(jobDescriptor, configurationDescriptor, serviceRegistryActor, cacheActorProps))
 }
 
 class HtCondorJobExecutionActor(override val jobDescriptor: BackendJobDescriptor,
                                 override val configurationDescriptor: BackendConfigurationDescriptor,
+                                serviceRegistryActor: ActorRef,
                                 cacheActorProps: Option[Props]) extends BackendJobExecutionActor with SharedFileSystem {
 
   import HtCondorJobExecutionActor._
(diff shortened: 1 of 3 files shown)


supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActor.scala
```diff
@@ -93,14 +97,29 @@ class HtCondorJobExecutionActor(override val jobDescriptor: BackendJobDescriptor
...
+
+    case KvKeyLookupFailed(_) =>
+      log.debug("{} Job id not found for job {} falling back to execute.", tag, jobDescriptor.key)
+      execute
+
+    case KvFailure(_, e) =>
+      val errMsg = s"$tag Failure attempting to look up HtCondor job id for key ${jobDescriptor.key}. Exception message: ${e.getMessage}."
+      log.error(errMsg)
+      executionResponse.tryFailure(e)
+      throw new RuntimeException(errMsg, e)
```
gauravs90: Why do you want to throw an exception? Just completing the promise should suffice here.

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActor.scala
```diff
@@ -93,14 +97,29 @@ class HtCondorJobExecutionActor(override val jobDescriptor: BackendJobDescriptor
     case ExecutionResultStored(hash) => log.debug("{} Cache entry was stored for Job with hash {}.", tag, hash)
 
     case ExecutionResultAlreadyExist => log.warning("{} Cache entry for hash {} already exist.", tag, jobHash)
+
+    case KvPair(key, Some(jobId)) if key.key == HtCondorJobIdKey =>
+      log.debug("{} Found job id {} for job {} trying to recover job now.", tag, jobId, jobDescriptor.key)
```
gauravs90: Could be a `log.info`

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActor.scala
```diff
@@ -93,14 +97,29 @@ class HtCondorJobExecutionActor(override val jobDescriptor: BackendJobDescriptor
...
+      executionResponse success trackTaskToCompletion(jobId)
+
+    case KvKeyLookupFailed(_) =>
+      log.debug("{} Job id not found for job {} falling back to execute.", tag, jobDescriptor.key)
```
gauravs90: Also, just logging `jobDescriptor.key.tag` should be enough here.. as opposed to the structure that you're logging now (`jobDescriptor.key`), which IMO is a little noisy and redundant.
(Same for logging errors below)

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActor.scala
```diff
@@ -93,14 +97,29 @@ class HtCondorJobExecutionActor(override val jobDescriptor: BackendJobDescriptor
     case ExecutionResultStored(hash) => log.debug("{} Cache entry was stored for Job with hash {}.", tag, hash)
 
     case ExecutionResultAlreadyExist => log.warning("{} Cache entry for hash {} already exist.", tag, jobHash)
+
+    case KvPair(key, Some(jobId)) if key.key == HtCondorJobIdKey =>
```
jainh: do you want to name something else key.key is like key^2

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActor.scala
```diff
@@ -93,14 +97,29 @@ class HtCondorJobExecutionActor(override val jobDescriptor: BackendJobDescriptor
...
+      executionResponse success trackTaskToCompletion(jobId)
+
+    case KvKeyLookupFailed(_) =>
+      log.debug("{} Job id not found for job {} falling back to execute.", tag, jobDescriptor.key)
+      execute
+
+    case KvFailure(_, e) =>
+      val errMsg = s"$tag Failure attempting to look up HtCondor job id for key ${jobDescriptor.key}. Exception message: ${e.getMessage}."
+      log.error(errMsg)
+      executionResponse.tryFailure(e)
```
gauravs90: Do we want to deviate from the earlier default behavior for recover? It used to execute if the recover functionality was not supported. I think we could stick to it and execute here rather than failing altogether (because of some other service failure).

supportedBackends/htcondor/src/test/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActorSpec.scala
```diff
@@ -423,4 +453,11 @@ class HtCondorJobExecutionActorSpec extends TestKitSuite("HtCondorJobExecutionAc
     override def storeExecutionResult(cachedExecutionResult: CachedExecutionResult): Unit = ()
   }
 
+  class KVServiceActor extends Actor {
```
gauravs90: Not that it's much important in the scope of writing tests (at least for me personally), a TestProbe could have been a better candidate for a Test Actor this simple in functionality.

francares: I don't know if Test Probes are intended for standard mock scenarios.

gauravs90: Yep.. they have a fairly standard `.expectMsg(..)` and `.reply(..)` methods to do exactly what you are doing here.

gauravs90:
Looks good 👍

mcovarr:
:+1:

PR #1252, opened 2016-08-05
SFS Config supports arbitrary runtime attributes. Closes #1217

### LINKED ISSUE #1217: Arbitrary SgeLikeBackend command line parameters
> We use lots of different runtime attributes (bsub job submission parameters) in our workflows.  
> 
> With existing Cromwell backends, you need to put all the possible runtime attributes in the code itself.  Would it be possible to let the user specify arbitrary runtime parameters only in the configuration file (not in the source code) that Cromwell just passes to the submit command?  
> 
> We want to avoid asking for Cromwell code changes every time we decide to use a new bsub parameter.

-- Pfizer

kshakir:
Moved parts of the shadow local backend into a new background executor.
Patches to tests based on changed to local: now with job id, fewer processes running in three step, using sfs config.
Replaced SGE backend with a config based version and added LSF example.
Revert: Jes's poll backoff starting at 30s up to 10 minutes is completely inappropriate for shared file system polling.
Docker Hub appears to have busted the v1 enpoint so disabling the test for now.
Sleep a second before letting the `WorkflowExecutionActorSpec` check for metadata status.


backend/src/main/scala/cromwell/backend/BackendInitializationData.scala
```diff
@@ -3,6 +3,23 @@ package cromwell.backend
 // single-backend "marker interface"
 trait BackendInitializationData
 
+object BackendInitializationData {
+
+  /**
+    * Utility for extracting a type of backend data that we know is wrapped in an Option.
+    * @param initializationDataOption The "optional" backend data passed usually from the initialization actor to each
+    *                                 execution actor.
```
mcovarr: is it not really optional any more?

kshakir: ToL: I don't know about _\<insert future backend here\>_, I think most backends should be caching some computed values, then passing it into execution actors. In the case of the SFS this includes things like the GCS file system used for localization.

engine/src/test/scala/cromwell/engine/workflow/SingleWorkflowRunnerActorSpec.scala
```diff
@@ -121,7 +121,7 @@ class SingleWorkflowRunnerActorWithMetadataSpec extends SingleWorkflowRunnerActo
     calls should not be empty
 
     forAll(expectedCalls) { (callName, numInputs, numOutputs) =>
-      val callSeq = calls.get(callName).get.asInstanceOf[JsArray].elements
+      val callSeq = calls(callName).asInstanceOf[JsArray].elements
```
geoffjentry: Someone has the latest version of intellij ;)

geoffjentry:
👍

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActorSpec.scala
```diff
@@ -57,6 +57,9 @@ class WorkflowExecutionActorSpec extends CromwellTestkitSpec with BeforeAndAfter
 
       val isFatal = (_: Throwable) => false
 
+      // Sleep a bit, and let the metadata do its thing in the background.
+      Thread.sleep(1000L)
```
mcovarr: 😢

supportedBackends/sfs/src/main/scala/cromwell/backend/impl/sfs/config/ConfigAsyncJobExecutionActor.scala
```diff
@@ -0,0 +1,150 @@
...
+  * containing the call command.
+  *
+  * There are two types of config executors, 1) those that will require being sent to the unix background, where their
+  * PID will be recorded, and 2) those that are dispatched to another system such as Grid Engine, LSF, PBS, etc.
+  */
+sealed trait ConfigAsyncJobExecutionActor extends SharedFileSystemAsyncJobExecutionActor {
+
+  lazy val configInitializationData = params.backendInitializationDataOption match {
+    case Some(data: ConfigInitializationData) => data
+    case other => throw new RuntimeException("Unable to get config initialization data from $other")
```
mcovarr: missing leading `s`

supportedBackends/sfs/src/main/scala/cromwell/backend/impl/sfs/config/DeclarationValidation.scala
```diff
@@ -0,0 +1,198 @@
...
+    */
+  def fromDeclaration(declaration: Declaration): DeclarationValidation = {
+    declaration.name match {
+      // Docker and CPU are special keys understood by cromwell.
+      case DockerValidation.key => new DeclarationValidation(declaration, DockerValidation.instance)
+      case CpuValidation.key => new DeclarationValidation(declaration, CpuValidation.default)
+      // See MemoryDeclarationValidation for more info
+      case name if MemoryDeclarationValidation.isMemoryDeclaration(name) =>
+        new MemoryDeclarationValidation(declaration)
+      // All other declarations must be a Boolean, Float, Integer, or String.
```
mcovarr: continue on return code?

mcovarr: nm that's really an engine thing isn't it

kshakir: "Engine"-ish. `ContinueOnReturnCodeValidation` and `FailOnStderrValidation` are both required validations by the SFS.

backend/src/main/scala/cromwell/backend/validation/RuntimeAttributesValidation.scala
```diff
@@ -221,7 +275,27 @@ trait RuntimeAttributesValidation[ValidatedType] {
...
+
+        exception:
+          java.lang.RuntimeException: Expression evaluation failed due to java.lang.UnsupportedOperationException:
+          No identifiers should be looked up: fromoptions: WdlExpression(<string:1:1 identifier "ZnJvbW9wdGlvbnM=">)
+
+        Not sure why yet the string "fromoptions" is being converted to an expression and not a WdlString.
+
+        For now, if something tries to "lookup" a value, convert it to a WdlString.
+         */
+        val wdlStringLookup: ScopedLookupFunction = (value: String) => WdlString(value)
```
mcovarr: I can't understand how this would be happening... 😦

mcovarr:
:+1:

[Approved with PullApprove](https://pullapprove.com/broadinstitute/cromwell/pull-request/1252/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=broadinstitute/cromwell)

PR #1254, opened 2016-08-05
Modifications on KV service to make it extensible. Closes #1255.

### LINKED ISSUE #1255: Modifications on KV service to make it extensible

francares:
Minor change in order to support different implementations of KV stores.
I think with this minor change we will be able to use different dbs for this service and not just SQL.
Our business requires to use Doc dbs so that why I'm proposing this change.
Later in the future we can discuss if it's worthy to define a DAL.

FYI... I tried to perform minimal changes to what is there...


services/src/main/scala/cromwell/services/keyvalue/impl/SqlKeyValueServiceActor.scala
```diff
@@ -0,0 +1,43 @@
...
+  override implicit val ec = context.dispatcher
+  private implicit val system = context.system
+
+  override def doPut(put: KvPut): Future[KvResponse] = {
+    put.pair.value match {
+      case Some(backendVal) => updateBackendKeyValuePair(put.pair.key.workflowId,
+        put.pair.key.jobKey,
+        put.pair.key.key,
+        put.pair.value.get).map(_ => KvPutSuccess(put))
+      case None => Future(KvFailure(put, new RuntimeException(s"Failed to find the value associated to key: ${put.pair.key.key}. This key cannot be added to the BackendKVStore.")))
```
mcovarr: `Future.successful`

services/src/main/scala/cromwell/services/keyvalue/KeyValueServiceActor.scala
```diff
@@ -0,0 +1,50 @@
...
+import cromwell.core.{JobKey, WorkflowId}
+import cromwell.services.ServiceRegistryActor.ServiceRegistryMessage
+import cromwell.services.keyvalue.KeyValueServiceActor._
+
+import scala.concurrent.{ExecutionContextExecutor, Future}
+import scala.util.{Failure, Success}
+
+
+object KeyValueServiceActor {
+  trait KvMessage
```
mcovarr: I think all these `trait`s could be `sealed`

mcovarr:
:+1:

cjllanwarne:
IMO this is an example where composition would be better than inheritance. If we just passed a general-purpose `KeyValueServiceActor` a `databaseLike: KeyValueDatabaseInterface` at construction time (for us, an SQL-backed database), this might look a lot nicer

francares:
@cjllanwarne I think that should be the second step to do (I was referring previously to that with a data abstraction layer -DAL- implementation) but It may involve a lot of more refactoring since you will need to modify database package.

cjllanwarne:
@francares oh ok, as long as it's on the long range plan. Incidentally, the other DB components (job store, workflow store) use a composition model without going too overboard on abstraction layers.

cjllanwarne:
LGTM.

francares:
@cjllanwarne Yes, but for example Workflow store has a dependency on WorkflowStoreSqlDatabase trait, so that means you can not use a NoSQLDatabase impl.
In order to allow WorkflowStore to do that you will need to implement a generic interface to work with different specializations of dbs and for that you will need to define a DAL and perform a bigger refactoring.

gauravs90:
👍

kshakir:
Very fast scan and not a proper review, but conceptually LGTM

geoffjentry:
Channeling my inner Joshua Bloch and make a similar comment as @cjllanwarne although I recognize the pain of doing that isn't worth the hassle here 👍

PR #1259, opened 2016-08-06
Reworking "run a wdl" tests, first blast

geoffjentry:
There's a bunch of little crud here and there which is remnants of me poking around before veering off on the Main.scala tangent. 

As I noted in standup I was going to push this stuff iteratively, the only stuff here which is actually complete are the changes related to Main (inc CromwellCommandLine). 

The rest of the changes were notes to myself, minor cleanups to help w/ my sanity as I start peeling CromwellTestKitSpec apart, etc.


src/main/scala/cromwell/CromwellCommandLine.scala
```diff
@@ -0,0 +1,104 @@
...
+        val workflowSourceFiles = WorkflowSourceFiles(w, i, o)
+        metadataPath foreach { m => if (!metadataPathIsWriteable(m)) throw new RuntimeException(s"Unable to write to metadata directory: $m")}
+        RunSingle(wdlPath, workflowSourceFiles, inputsPath, optionsPath, metadataPath)
+    } match {
+      case scalaz.Success(r) => r
+      case scalaz.Failure(nel) => throw new RuntimeException with MessageAggregation {
+        override def exceptionContext: String = "ERROR: Unable to run Cromwell:"
+        override def errorMessages: Traversable[String] = nel.list
+      }
+    }
```
mcovarr: Disclaimer: didn't actually try this outside IntelliJ, but it does keep everything in `ErrorOr` until the final failure check:

```scala
    def writeableMetadataPath(path: Option[Path]): ErrorOr[Unit] = {
      path match {
        case Some(p) if !metadataPathIsWriteable(p) => s"Unable to write to metadata directory: $p".failureNel
        case otherwise => ().successNel
      }
    }

    val sourceFiles: ErrorOr[WorkflowSourceFiles] = (wdl |@| inputsJson |@| optionsJson) {
      WorkflowSourceFiles.apply
    }

    val runSingle = for {
      sources <- sourceFiles
      _ <- writeableMetadataPath(metadataPath)
    } yield RunSingle(wdlPath, sources, inputsPath, optionsPath, metadataPath)

    runSingle match {
      case scalaz.Success(r) => ...
    }
```

src/main/scala/cromwell/CromwellCommandLine.scala
```diff
@@ -0,0 +1,104 @@
...
+      case _ => UsageAndExit
+    }
+  }
+}
+
+object RunSingle {
+  def apply(args: Seq[String]): RunSingle = {
+    val wdlPath = Paths.get(args.head)
+    val inputsPath = argPath(args, 1, Option(".inputs"), checkDefaultExists = false)
+    val optionsPath = argPath(args, 2, Option(".options"), checkDefaultExists = true)
```
mcovarr: I don't understand the difference in `checkDefaultExists` for the inputs and options here

src/main/scala/cromwell/CromwellCommandLine.scala
```diff
@@ -0,0 +1,104 @@
...
+    * Retrieve the arg at index as path, or return some default. Args specified as "-" will be returned as None.
+    *
+    * @param args The run command arguments, with the wdl path at arg.head.
+    * @param index The index of the path we're looking for.
+    * @param defaultExt The default extension to use if the argument was not specified at all.
+    * @param checkDefaultExists If true, verify that our computed default file exists before using it.
+    * @return The argument as a Path resolved as a sibling to the wdl path.
+    */
+  private def argPath(args: Seq[String], index: Int, defaultExt: Option[String],
+                      checkDefaultExists: Boolean = true): Option[Path] = {
```
mcovarr: I was reading over the docs on defaulted inputs and options with Ruchi the other day and it made me wonder if people actually like this functionality.  I wonder if it's more likely to be a source of confusing errors than a convenience.

src/main/scala/cromwell/CromwellCommandLine.scala
geoffjentry: *shrug*, a lot of this code is just c/p from Main and moved around a bit

src/main/scala/cromwell/CromwellCommandLine.scala
geoffjentry: I *don't* like it, FWIW

mcovarr:
:+1: delta some confusion / grousing on inputs and options parsing.

src/main/scala/cromwell/CromwellCommandLine.scala
geoffjentry: Yeah, I changed something around towards the end which would allow something like this, and then promptly forgot to go back and look at it

src/main/scala/cromwell/CromwellCommandLine.scala
mcovarr: Yeah I noticed that.  But AFAICT from the README I would have thought they would be handled identically?

geoffjentry: TBH I raised an eyebrow at it but it was working before so I brought it forward. Feature party and all that. I'll take a quick look to make sure that the code and readme at least agree.

I'd certainly having a confab regarding defaults, command line entry points,  and such at some point in the future, touching on your other point.

geoffjentry: I traced this back a bit. This behavior doesn't contradict anything in README, however it does seem inconsistent. My archeological dig leads me to believe that it is an artifact of a) originally being done differently because two different people did it and then b) the previous major refactor of Main dutifully preserved the behavior difference.

I'll unify these as I don't see why they'd be different. However I'm concerned that there's something subtle going on that we're missing here so we should keep an eye out

src/main/scala/cromwell/CromwellCommandLine.scala
geoffjentry: resolved

src/main/scala/cromwell/CromwellCommandLine.scala
geoffjentry: Thinking about this during lunch (at commonwealth, what is this, a weekday?) I started to get more concerned about the "something subtle going on" aspect. 

Questions like this and your other comment regarding defaults & such make me think the right path forward would be to sit down at some point and decide how we think things *should* work and then make the code/docs more clear. Perhaps the former is a no op once we work it through, but the latter definitely isn't. I'll make a ticket and advocate it as part of the 0.23 milestone

geoffjentry: "resolved" via #1260

src/main/scala/cromwell/CromwellCommandLine.scala
geoffjentry: "resolved" via #1260

src/main/scala/cromwell/CromwellCommandLine.scala
pgrosu: Tackling big problems on a beautiful Sunday afternoon - Jeff, the sun is shining :)

geoffjentry: I was outside in the morning ;)

engine/src/test/scala/cromwell/CromwellTestkitSpec.scala
```diff
@@ -118,6 +119,7 @@ object CromwellTestkitSpec {
 
   val TimeoutDuration = 60 seconds
 
+  // FIXME JG: Directly called here (withTestWorkflowManagerSystem & passed to CTS) and LocalBackendSpec - it's assigning one and not using (looks like a candidate for MainSpec type behavior!)
```
cjllanwarne: ToL: I'm not a big fan of personal "notes to self" being added to Develop

geoffjentry: I explained last week why they'll be there. Since I'm changing nearly all meaningful tests, it's going to be impossible to stay on top of develop while I proceed.

cjllanwarne:
Apart from not liking the Notes to Self, looks fine 👍

PR #1262, opened 2016-08-08
changes to include new metadata from JES closes #1258

### LINKED ISSUE #1258: Add machine type back into metadata (added back in by JES)
JES had removed the machine type from their `gcloud alpha genomics operations describe` json response some time ago but it has been added back in recently.  We would like this to be put back into the cromwell metadata so we can correctly calculate costs for our workflows.  

Example json from the operations describe command:

```
"runtimeMetadata": {
      "@type": "type.googleapis.com/google.genomics.v1alpha2.RuntimeMetadata",
      "computeEngine": {
        "diskNames": [
          "local-disk-13965373557709634529"
        ],
        "instanceName": "ggp-13965373557709634529",
        "machineType": "us-central1-b/n1-standard-2",
        "zone": "us-central1-b"
      }
    }
```

kcibul:


geoffjentry:
👍

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/Run.scala
```diff
@@ -92,11 +92,16 @@ case class Run(runId: String, genomicsInterface: Genomics) {
   def status(): RunStatus = {
     val op = genomicsInterface.operations().get(runId).execute
     if (op.getDone) {
-      // If there's an error, generate a Failed status. Otherwise, we were successful!
       val eventList = getEventList(op)
+      val ceInfo = op.getMetadata.get ("runtimeMetadata").asInstanceOf[GArrayMap[String,Object]].get("computeEngine").asInstanceOf[GArrayMap[String, String]]
```
cjllanwarne: There's a lot of `getOrDefault`s below to cover us if they're missing. Which makes this combination of `get` and `asInstanceOf` look quite a bit more dangerous by comparison!

geoffjentry: FWIW this line bothered me due to the runtime danger, but this whole function is wrapped in a Try so I went back to not worrying

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/Run.scala
```diff
@@ -92,11 +92,16 @@ case class Run(runId: String, genomicsInterface: Genomics) {
...
+      val machineType = ceInfo.getOrDefault("machineType","unknown")
+      val instanceName = ceInfo.getOrDefault("instanceName","unknown")
+      val zone = ceInfo.getOrDefault("zone","unknown")
```
cjllanwarne: I feel like it would be more intellectually honest to just return these as `Option`s

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/RunStatus.scala
```diff
@@ -18,13 +18,16 @@ object RunStatus {
...
+    def instanceName: String
   }
 
-  case class Success(eventList: Seq[EventStartTime]) extends TerminalRunStatus {
+  case class Success(eventList: Seq[EventStartTime], machineType: String, zone: String, instanceName: String) extends TerminalRunStatus {
     override def toString = "Success"
   }
 
-  final case class Failed(errorCode: Int, errorMessage: Option[String], eventList: Seq[EventStartTime])
+  final case class Failed(errorCode: Int, errorMessage: Option[String], eventList: Seq[EventStartTime], machineType: String, zone: String, instanceName: String)
```
cjllanwarne: Unless I missed something, nothing ever reads this triple for Failed so they could just not be here.

OTOH, There might be value in sending them to metadata even in case of a failure, so you can still calculate the cost for failed jobs

cjllanwarne:
Small comments. Otherwise 👍

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/RunStatus.scala
kcibul: yes -- we need them for failed jobs... but good catch that we're not sending them!

kcibul: hmmm... wait does this mean that we don't get the events either for failed (say preempted) tasks?

kcibul: nope -- we do... I misread the Success as being a successful job... that's not true.  It's a successful of a try of getting the status (which could be failed).

PR #1275, opened 2016-08-09
Return workflow outputs.  Closes #1257

### LINKED ISSUE #1257: Outputs empty but not metadata
This was supposedly fixed already but we saw the following report from a user in gitter, he's using HEAD on develop:

Hey guys, One more quick question for you! When using the rest-api locally I submit a workflow and wait for it to finish. Once its finished I call the /outputs endpoint for that specific job, but I consistently get an empty output object, despite the fact that in the /metadata endpoint the outputs object is populated. Is this a bug or the desired behaviour?

mcovarr:
The outputs query was changed such that the key the `MetadataBuilderActor` was looking for in the `WorkflowOutputsResponse` handler was never returned.  `MetadataBuilderActor` quietly (some might say blithely) fell back to returning empty outputs.  

It's alarming that this total breakage was not caught by any tests.  But rather than add to the morass that is `CromwellTestKitSpec` I'm going to look at enhancing Centaur to be able to hit the outputs endpoint.

### DIFF
--- a/engine/src/main/scala/cromwell/webservice/metadata/MetadataBuilderActor.scala
+++ b/engine/src/main/scala/cromwell/webservice/metadata/MetadataBuilderActor.scala
@@ -232,7 +232,7 @@ class MetadataBuilderActor(serviceRegistryActor: ActorRef) extends LoggingFSM[Me
       allDone
     case Event(WorkflowOutputsResponse(id, events), _) =>
       // Add in an empty output event if there aren't already any output events.
-      val hasOutputs = events collectFirst { case MetadataEvent(MetadataKey(_, _, key), _, _) if key == WorkflowMetadataKeys.Outputs => () } isDefined
+      val hasOutputs = events exists { _.key.key.startsWith(WorkflowMetadataKeys.Outputs + ":") }
       val updatedEvents = if (hasOutputs) events else MetadataEvent.empty(MetadataKey(id, None, WorkflowMetadataKeys.Outputs)) +: events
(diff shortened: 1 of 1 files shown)


geoffjentry:
👍 

I thank you for considering your fellow Cromwellian currently stuck in the morass.

engine/src/main/scala/cromwell/webservice/metadata/MetadataBuilderActor.scala
```diff
@@ -232,7 +232,7 @@ class MetadataBuilderActor(serviceRegistryActor: ActorRef) extends LoggingFSM[Me
       allDone
     case Event(WorkflowOutputsResponse(id, events), _) =>
       // Add in an empty output event if there aren't already any output events.
-      val hasOutputs = events collectFirst { case MetadataEvent(MetadataKey(_, _, key), _, _) if key == WorkflowMetadataKeys.Outputs => () } isDefined
+      val hasOutputs = events collectFirst { case MetadataEvent(MetadataKey(_, _, key), _, _) if key.startsWith(WorkflowMetadataKeys.Outputs + ":") => () } isDefined
```
cjllanwarne: Is this rather odd combination of `collectFirst`, `case x if y => ()` and `isDefined` just equivalent to `exists`?

cjllanwarne:
👍

engine/src/main/scala/cromwell/webservice/metadata/MetadataBuilderActor.scala
mcovarr: sure looks that way

PR #1277, opened 2016-08-09
Metadata migration Closes #789

### LINKED ISSUE #789: Data Migration from existing metadata schema into Metadata Service schema

Horneth:
Uses liquibase [custom change](http://www.liquibase.org/documentation/changes/custom_change.html) to write classes that transform pre-0.20 existing data into 0.20 metadata.

- [x] Test at scale


database/src/main/scala/cromwell/database/migration/metadata/MetadataStatement.scala
```diff
@@ -0,0 +1,114 @@
...
+import java.time.{ZoneOffset, OffsetDateTime}
+import java.time.format.DateTimeFormatter
+
+import cromwell.database.SqlConverters._
+import liquibase.database.jvm.JdbcConnection
+import org.slf4j.LoggerFactory
+import wdl4s.values.{WdlBoolean, WdlFloat, WdlInteger, WdlValue}
+
+object MetadataStatement {
+  val workflowIdIdx = 1
```
geoffjentry: workflowIdIdx.capitalize (and for the rest)

database/src/main/resources/changesets/metadata_migration.xml
```diff
@@ -0,0 +1,22 @@
...
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.3.xsd">
+
+    <property name="clob.type" value="LONGTEXT" dbms="mysql"/>
+    <property name="clob.type" value="LONGVARCHAR" dbms="hsqldb"/>
+
+    <changeSet author="tjeandet" id="METADATA_MIGRATION">
+        <comment>
+            Metadata migration from 0.19 to 0.20.
```
mcovarr: .21?

database/src/main/scala/cromwell/database/migration/metadata/MetadataMigration.scala
```diff
@@ -0,0 +1,76 @@
...
+      case (row, idx) =>
+        migrateRow(connection, collectors, metadataInsertStatement, row, idx)
+        if (idx % 100 == 0) {
+          metadataInsertStatement.executeBatch()
+          connection.commit()
+        }
+    }
+
+    metadataInsertStatement.executeBatch()
+    connection.commit()
```
mcovarr: is this going to be cool if the number of executions % 100 == 0?

database/src/main/scala/cromwell/database/migration/metadata/MetadataMigration.scala
```diff
@@ -0,0 +1,76 @@
...
+        |GROUP BY CALL_FQN, ATTEMPT, WORKFLOW_EXECUTION_ID
+        |HAVING COUNT(*) > 1
+      """.stripMargin
+
+    val collectorsRS = new ResultSetIterator(connection.createStatement().executeQuery(collectorsIdQuery))
+    collectorsRS map { _.getInt("EXECUTION_ID") } toSet
+  }
+
+  override def execute(database: Database): Unit = {
+    val dbConn = database.getConnection.asInstanceOf[JdbcConnection]
```
mcovarr: move this inside the `try`?

database/src/main/scala/cromwell/database/migration/metadata/MetadataMigration.scala
Horneth: This was a problem on Hsql until 2.3.4 because they wouldn't execute empty batches, but they added a flag to allow so it should be fine now.

database/src/main/scala/cromwell/database/migration/metadata/MetadataStatement.scala
```diff
@@ -0,0 +1,114 @@
...
+}
+
+class MetadataStatementForCall(preparedStatement: PreparedStatement, workflowId: String, callFqn: String, index: Int, attempt: Int) extends MetadataStatementForWorkflow(preparedStatement, workflowId) {
+  override def setStatement() = {
+    preparedStatement.setString(MetadataStatement.workflowIdIdx, workflowId)
+    preparedStatement.setString(MetadataStatement.callFqnIdx, callFqn)
+    preparedStatement.setInt(MetadataStatement.callIndexIdx, index)
+    preparedStatement.setInt(MetadataStatement.callAttemptIdx, attempt)
+  }
+}
```
mcovarr: \n

database/src/main/scala/cromwell/database/migration/metadata/MetadataMigration.scala
```diff
@@ -0,0 +1,76 @@
...
+    try {
+      dbConn.setAutoCommit(false)
+      migrate(dbConn, findCollectorIds(dbConn))
+    } catch {
+      case t: CustomChangeException => throw t
+      case t: Throwable => throw new CustomChangeException("Could not apply migration script for metadata", t)
+    }
+  }
+
+  override def setUp(): Unit = { }
```
mcovarr: `= ()`

geoffjentry: TOL (i.e. it just reminded me of something I was thinking about this morning) - we might want to consider turning on -Ywarn-value-discard compiler feature. I keep reading about people having bugs due to unit return types and other tomfoolery

mcovarr: Ooh that would make a nice Developer's Choice ticket

geoffjentry: If we start talking that route there are all *kinds* of compiler flags I'd like to turn on, which unfortunately would probably break a bunch of stuff at the moment

mcovarr: I suspect IntelliJ probably does an adequate job warning of these cases, but I find it a weird default behavior for everything to be silently coerced to the Unit value.

geoffjentry: I'm sure @Horneth loves this side conversation ;)

Recently I came across a blog post describing people running into this, which caused them much pain (can only find the SIP, not the blog): https://issues.scala-lang.org/browse/SI-9822

database/src/main/scala/cromwell/database/migration/metadata/MetadataStatement.scala
```diff
@@ -0,0 +1,114 @@
...
+
+  protected def addDataAndBatch(key: String, value: Any) = {
+    preparedStatement.setString(MetadataStatement.keyIdx, key)
+
+    // Set the value and type
+    value match {
+      case null =>
+        preparedStatement.setNull(MetadataStatement.valueIdx, Types.VARCHAR)
+        preparedStatement.setNull(MetadataStatement.valueTypeIdx, Types.VARCHAR) // Null values have null type
+      case v =>
```
mcovarr: _

database/src/main/scala/cromwell/database/migration/metadata/table/ExecutionEventTableMigration.scala
```diff
@@ -0,0 +1,35 @@
...
+
+  override protected def migrateRow(connection: JdbcConnection, collectors: Set[Int],
+                                    statement: PreparedStatement, row: ResultSet, idx: Int): Unit = {
+    if (!collectors.contains(row.getInt("EXECUTION_ID"))) {
+      val metadataStatement = new MetadataStatementForCall(statement,
+        row.getString("WORKFLOW_EXECUTION_UUID"),
+        row.getString("CALL_FQN"),
+        row.getInt("IDX"),
+        row.getInt("ATTEMPT")
+      )
```
mcovarr: The `if` and the `MetadataStatementForCall` construction seems to occur in a lot of these migrations, can it be refactored to the trait?

database/src/main/scala/cromwell/database/migration/metadata/MetadataMigration.scala
```diff
@@ -0,0 +1,76 @@
...
+          metadataInsertStatement.executeBatch()
+          connection.commit()
+        }
+    }
+
+    metadataInsertStatement.executeBatch()
+    connection.commit()
+  }
+
+  private var resourceAccessor: ResourceAccessor = null
```
mcovarr: IntelliJ requests `= _`

database/src/main/scala/cromwell/database/migration/metadata/MetadataMigration.scala
```diff
@@ -0,0 +1,76 @@
...
+
+    val collectorsRS = new ResultSetIterator(connection.createStatement().executeQuery(collectorsIdQuery))
+    collectorsRS map { _.getInt("EXECUTION_ID") } toSet
+  }
+
+  override def execute(database: Database): Unit = {
+    val dbConn = database.getConnection.asInstanceOf[JdbcConnection]
+    try {
+      dbConn.setAutoCommit(false)
+      migrate(dbConn, findCollectorIds(dbConn))
```
mcovarr: Not sure what the execution time is like for running all these migrations on a "mature" database, but might it be worth caching the collector IDs since it doesn't look like the results should change?

mcovarr:
:+1: delta your "test at scale" item.  😄   Very cool!

geoffjentry:
💯  / 💯   ✔️ ➕ 

👍

ruchim:
This is such a cool branch Thibault, I didn't realize Liquibase could do so much!

Horneth:
Yeah I didn't know either but it's pretty useful in this case !

database/src/main/scala/cromwell/database/migration/metadata/table/SymbolTableMigration.scala
```diff
@@ -0,0 +1,69 @@
...
+  override protected def migrateRow(connection: JdbcConnection, statement: PreparedStatement, row: ResultSet, idx: Int): Unit = {
+    // Try to coerce the value to a WdlValue
+    val value = for {
+      wdlTypeValue <- Try(row.getString("WDL_TYPE"))
+      wdlType <- Try(WdlType.fromWdlString(wdlTypeValue))
+      wdlValueValue <- Try(row.getString("WDL_VALUE"))
+      inflated <- inflate(wdlValueValue)
+
+      wdlValue <- wdlType match {
+        case p: WdlPrimitiveType => p.coerceRawValue(inflated)
```
mcovarr: Unfortunately that this distinction in behavior needs to be in the caller; it would be nice if `WdlPrimitive`s overrode `fromWdlString` to do `coerceRawValue`.

database/src/main/scala/cromwell/database/migration/metadata/table/CallOutputSymbolTableMigration.scala
```diff
@@ -0,0 +1,49 @@
...
+                    SELECT CALL_FQN, IDX, WORKFLOW_EXECUTION_ID, MAX(EXECUTION.ATTEMPT) AS MaxAttempt
+                    FROM EXECUTION
+                    GROUP BY EXECUTION.CALL_FQN, EXECUTION.IDX, EXECUTION.WORKFLOW_EXECUTION_ID
+                  ) MaxExecution
+           ON SYMBOL.SCOPE = MaxExecution.CALL_FQN
+              AND SYMBOL.WORKFLOW_EXECUTION_ID = MaxExecution.WORKFLOW_EXECUTION_ID
+              AND SYMBOL.IDX = MaxExecution.IDX
+         LEFT JOIN(SELECT EXECUTION_ID, CALL_FQN, IDX, WORKFLOW_EXECUTION_ID, ATTEMPT
+                   FROM EXECUTION
+                  ) ExecutionId
```
mcovarr: Why is this a subquery, couldn't it just `LEFT JOIN EXECUTION`?

database/src/main/scala/cromwell/database/migration/metadata/MetadataMigration.scala
```diff
@@ -0,0 +1,108 @@
...
+          metadataInsertStatement.executeBatch()
+          connection.commit()
+        }
+    }
+
+    metadataInsertStatement.executeBatch()
+    connection.commit()
+  }
+
+  private var resourceAccessor: ResourceAccessor = _
```
mcovarr: is this thing actually used?

database/src/main/scala/cromwell/database/migration/metadata/table/CallOutputSymbolTableMigration.scala
Horneth: Yes and it should

database/src/main/scala/cromwell/database/migration/metadata/table/CallOutputSymbolTableMigration.scala
```diff
@@ -0,0 +1,49 @@
...
+import cromwell.database.migration.metadata.MetadataStatementForCall
+import wdl4s.values._
+
+class CallOutputSymbolTableMigration extends SymbolTableMigration {
+  override protected def selectQuery: String = """
+     SELECT SYMBOL.SCOPE, SYMBOL.NAME, MaxExecution.IDX, IO, WDL_TYPE,
+         WDL_VALUE, WORKFLOW_EXECUTION_UUID, ExecutionId.EXECUTION_ID, REPORTABLE_RESULT,
+         MaxExecution.MaxAttempt
+       FROM SYMBOL
+         LEFT JOIN WORKFLOW_EXECUTION ON SYMBOL.WORKFLOW_EXECUTION_ID = WORKFLOW_EXECUTION.WORKFLOW_EXECUTION_ID
```
mcovarr: seems like the `LEFT` part of the `LEFT JOIN` should be unnecessary?

Horneth: Yep - I have cleaned this up in all the queries just haven't pushed it yet xD

database/src/main/scala/cromwell/database/migration/metadata/table/CallOutputSymbolTableMigration.scala
```diff
@@ -0,0 +1,49 @@
...
+         LEFT JOIN(
+                    SELECT CALL_FQN, IDX, WORKFLOW_EXECUTION_ID, MAX(EXECUTION.ATTEMPT) AS MaxAttempt
+                    FROM EXECUTION
+                    GROUP BY EXECUTION.CALL_FQN, EXECUTION.IDX, EXECUTION.WORKFLOW_EXECUTION_ID
+                  ) MaxExecution
```
mcovarr: I wonder if it would be faster to `MAX` on the execution ID rather than the attempt?

database/src/main/scala/cromwell/database/migration/package.scala
```diff
@@ -0,0 +1,10 @@
+package cromwell.database
+
+import java.sql.ResultSet
+
+package object migration {
+  class ResultSetIterator(rs: ResultSet) extends Iterator[ResultSet] {
```
kshakir: Move this to `ResultSetIterator.scala`

kshakir:
Re: the `hsqldb compliant` commit: There should be multiple ways to check at runtime the database type, and then customize the code/SQL within the `CustomTaskChange`. For example, matching on the `liquibase.database.Database`, or go further and use the `Database.getDatabaseProductName`. The latter is a wrapper to `java.sql.Connection.getMetadata.getDatabaseProductName`. For HSQL it should return `HsqlDatabaseProperties.PRODUCT_NAME`, or the hard coded string `"MySQL"`.

Horneth:
@kshakir I tagged everything with `dbms="mysql"' after the standup discussion and removed the hsql specific code...

core/src/main/scala/cromwell/core/simpleton/WdlValueSimpleton.scala
```diff
@@ -26,6 +27,7 @@ object WdlValueSimpleton {
   implicit class WdlValueSimplifier(wdlValue: WdlValue) {
     def simplify(name: String): Iterable[WdlValueSimpleton] = wdlValue match {
       case prim: WdlPrimitive => List(WdlValueSimpleton(name, prim))
+      case expr: WdlExpression => List(WdlValueSimpleton(name, WdlString(expr.valueString)))
```
mcovarr: How is this currently being used?  I had thought that `WdlValueSimpleton`s were used only for expressions that had been fully evaluated to `WdlValue`s, in particular the cache and the job store.  It doesn't look like `WdlValueBuilder` could currently handle reversing this simplification.

database/src/main/resources/changesets/migration/metadata/CreateAndLoadTmpCollector.sql
```diff
@@ -0,0 +1,19 @@
...
+  EXECUTION_ID INT NOT NULL
+);
+
+INSERT INTO TMP_COLLECTOR (
+  EXECUTION_ID
+)
+  SELECT
+    EXECUTION_ID
+  FROM EXECUTION e
+  WHERE e.CALL_FQN LIKE '%$%' OR  -- filter out scatters
```
mcovarr: I guess the temp table is slightly misnamed if it's getting scatters too

database/src/main/scala/cromwell/database/migration/metadata/table/ExecutionTableMigration.scala
```diff
@@ -0,0 +1,269 @@
...
+    IDX,
+    ATTEMPT,
+    DATE_FORMAT(e.START_DT, '%Y-%m-%dT%T.%f$Offset'),
+    'string',
+    NOW()
+    FROM EXECUTION e
+    JOIN WORKFLOW_EXECUTION we ON we.WORKFLOW_EXECUTION_ID = e.WORKFLOW_EXECUTION_ID
+    WHERE
+    e.START_DT IS NOT NULL AND
+    e.EXECUTION_ID NOT IN (SELECT EXECUTION_ID FROM TMP_COLLECTOR);""".stripMargin
```
mcovarr: no margins in need of stripping here 😄

database/src/main/resources/changesets/migration/metadata/LoadInputSymbols.sql
```diff
@@ -0,0 +1,30 @@
...
+    JOIN WORKFLOW_EXECUTION we ON
+                                 we.WORKFLOW_EXECUTION_ID = s.WORKFLOW_EXECUTION_ID
+    LEFT JOIN EXECUTION e ON
+                            e.CALL_FQN = s.SCOPE AND
+                            e.WORKFLOW_EXECUTION_ID = s.WORKFLOW_EXECUTION_ID
+                            -- Don't join on index here because inputs have index null (-1), but we want to duplicate them as many times as there are indices for a given execution
+  WHERE
+    s.IO = 'INPUT' AND
+    (
+      e.EXECUTION_ID IS NULL OR -- If it's a workflow output the execution ID will be null but we want to keep it
```
mcovarr: but this is looking at INPUTs?

database/src/main/scala/cromwell/database/migration/metadata/table/executionevent/ExecutionEventTableDescriptionMigration.scala
```diff
@@ -0,0 +1,37 @@
...
+        |  CALL_FQN,
+        |  JOB_SCATTER_INDEX,
+        |  JOB_RETRY_ATTEMPT,
+        |  METADATA_VALUE,
+        |  METADATA_VALUE_TYPE,
+        |  METADATA_TIMESTAMP
+        |)
+        |  SELECT
+        |    WORKFLOW_EXECUTION_UUID,
+        |    CONCAT('executionEvents[', ev.EVENT_ID ,']:description'),
```
mcovarr: The index here is a synthetic PK in the EXECUTION_EVENTS table, does that make sense here?

database/src/main/scala/cromwell/database/migration/metadata/table/FailureEventTableMigration.scala
```diff
@@ -0,0 +1,62 @@
...
+        |  CALL_FQN,
+        |  JOB_SCATTER_INDEX,
+        |  JOB_RETRY_ATTEMPT,
+        |  METADATA_VALUE,
+        |  METADATA_VALUE_TYPE,
+        |  METADATA_TIMESTAMP
+        |)
+        |SELECT
+        |  WORKFLOW_EXECUTION_UUID,
+        |  CONCAT('failures[', fe.FAILURE_EVENT_ID ,']:failure'),
```
mcovarr: same question as below about using a synthetic PK as a subscript for this metadata

database/src/main/scala/cromwell/database/migration/metadata/table/ExecutionTableMigration.scala
```diff
@@ -0,0 +1,269 @@
...
+        METADATA_VALUE_TYPE,
+        METADATA_TIMESTAMP
+      )
+        SELECT
+        WORKFLOW_EXECUTION_UUID,
+      'start',
+    CALL_FQN,
+    IDX,
+    ATTEMPT,
+    DATE_FORMAT(e.START_DT, '%Y-%m-%dT%T.%f$Offset'),
```
mcovarr: Is it correct to apply the `$Offset` here?  I thought the original times were UTC and should be kept that way?

database/src/main/scala/cromwell/database/migration/metadata/table/executionevent/ExecutionEventTableDescriptionMigration.scala
```diff
@@ -0,0 +1,37 @@
+package cromwell.database.migration.metadata.table.executionevent
+
+import cromwell.database.migration.metadata.MetadataCustomSql
+
+class ExecutionEventTableDescriptionMigration extends MetadataCustomSql {
```
mcovarr: not sure why this one is a custom sql migration?

database/src/main/scala/cromwell/database/migration/metadata/table/ExecutionTableMigration.scala
mcovarr: ToL these queries seem more interested in finding out what executions *are not* scatters/collectors than which *are* scatters / collectors.  Perhaps there should be a table that contains these executions and then these queries could just join to that.

core/src/main/scala/cromwell/core/simpleton/WdlValueSimpleton.scala
Horneth: Some symbols are stored as `WdlExpression` so I got a match error when trying to simpletonize them. I can filter them out earlier before calling `simplify` if that's a problem, but `simplify` should at least handle this case even if it's just throwing an exception

database/src/main/resources/changesets/migration/metadata/LoadInputSymbols.sql
Horneth: the comment is wrong - it's workflow input

database/src/main/scala/cromwell/database/migration/metadata/table/executionevent/ExecutionEventTableDescriptionMigration.scala
Horneth: It doesn't have to - I just thought migration queries are already all over the place so I at least wanted to keep tables together in the same "logic area"

database/src/main/scala/cromwell/database/migration/metadata/table/executionevent/ExecutionEventTableDescriptionMigration.scala
Horneth: Hmm I'm not sure I understand what you're asking

database/src/main/scala/cromwell/database/migration/metadata/table/ExecutionTableMigration.scala
Horneth: I double checked in the code, but from what I see we inject
`new Timestamp(new Date().getTime)` ([here](https://github.com/broadinstitute/cromwell/blob/0.19_hotfix/engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala#L740) + [here](https://github.com/broadinstitute/cromwell/blob/0.19_hotfix/engine/src/main/scala/cromwell/engine/db/slick/SlickDataAccess.scala#L46)) as start and end times, which on my machine returns local time:

```scala
scala> import java.sql.Timestamp
import java.sql.Timestamp
scala> import java.util.Date
import java.util.Date
scala> new Timestamp(new Date().getTime)
res0: java.sql.Timestamp = 2016-09-02 18:08:27.173

```

database/src/main/scala/cromwell/database/migration/metadata/table/executionevent/ExecutionEventTableDescriptionMigration.scala
Horneth: Oh I get it now -_- I was just looking for something unique so they wouldn't override each other. The order in the list doesn't matter as much I think as the elements have a startTime and endTime field.

core/src/main/resources/application.conf
```diff
@@ -362,7 +362,7 @@ database {
       driver = "slick.driver.HsqldbDriver$"
       db {
         driver = "org.hsqldb.jdbcDriver"
-        url = "jdbc:hsqldb:mem:${uniqueSchema};shutdown=false;hsqldb.tx=mvcc"
+        url = "jdbc:hsqldb:mem:${uniqueSchema};shutdown=false;hsqldb.tx=mvcc;allow_empty_batch=true"
```
kshakir: ToL: Is this still required for all HSQLDB urls? If so, we may want to start documenting these requirements.

database/migration/src/main/resources/changesets/migration/metadata/CreateAndLoadTmpExecutionMigration.sql
```diff
@@ -0,0 +1,42 @@
...
+    BACKEND_TYPE
+  FROM EXECUTION e
+  WHERE e.CALL_FQN NOT LIKE '%$%' AND  -- filter out scatters
+        NOT (e.IDX = -1 AND EXISTS (   -- filter out collectors
+            SELECT 1 FROM EXECUTION e2 WHERE
+              e2.WORKFLOW_EXECUTION_ID = e.WORKFLOW_EXECUTION_ID AND
+              e2.CALL_FQN = e.CALL_FQN AND
+              e2.IDX != -1)
+        );
+
```
geoffjentry: \n

core/src/main/resources/application.conf
Horneth: I guess not since the migration only applies to mysql. However there will be some flag required for mysql which I thought could be documented in the Changelog

database/migration/src/main/resources/changesets/standardize_column_names.xml
```diff
@@ -0,0 +1,42 @@
...
+                      newColumnName="JOB_SCATTER_INDEX"
+                      oldColumnName="METADATA_CALL_INDEX"
+                      tableName="METADATA_JOURNAL"/>
+        <renameColumn columnDataType="INT"
+                      newColumnName="JOB_RETRY_ATTEMPT"
+                      oldColumnName="METADATA_CALL_ATTEMPT"
+                      tableName="METADATA_JOURNAL"/>
+    </changeSet>
+
+</databaseChangeLog>
```
kshakir: \n, mainly because after this gets merged, the md5 is saved into the database and I can't change it in another PR later.

geoffjentry: Oh now I think that @Horneth should leave it like this just to make @kshakir forever mournful

geoffjentry:
@Horneth woohoo!

PR #1281, opened 2016-08-10
Handle Option[BackendInitializationData] like it's optional.  Closes #1272

### LINKED ISSUE #1272: JES backend blithely assumes non-empty BackendInitializationData on finalization
There seem to be intermittent Centaur failures on `invalid_runtime_attributes`.  I investigated this for a while and found the following:

`WorkflowActor` will always try to run finalization, even if initialization fails.  And if initialization fails there will be no initialization data, which means a `.get` on the optional initialization data will throw.  Unfortunately this is exactly what `toJes` in `JesBackendLifecycleActorFactory` is doing:

```scala
def toJes = genericInitializationData collectFirst { case d: JesBackendInitializationData => d } get
```

My guess is that this doesn't fail 100% of the time because there's already failure metadata generated (intentionally), and there's a race condition as to which failure event shows up first.  If the intentional failure is generated first the test would pass.

mcovarr:


supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesBackendLifecycleActorFactory.scala
```diff
@@ -32,7 +34,7 @@ case class JesBackendLifecycleActorFactory(configurationDescriptor: BackendConfi
   override def jobExecutionActorProps(jobDescriptor: BackendJobDescriptor,
                                       initializationData: Option[BackendInitializationData],
                                       serviceRegistryActor: ActorRef): Props = {
-    JesJobExecutionActor.props(jobDescriptor, jesConfiguration, initializationData.toJes, serviceRegistryActor).withDispatcher(BackendDispatcher)
+    JesJobExecutionActor.props(jobDescriptor, jesConfiguration, initializationData.toJes.get, serviceRegistryActor).withDispatcher(BackendDispatcher)
```
cjllanwarne: If this is calling `.get` unprotected, it would be better to accept a `BackendInitializationData` in the signature instead of an `Option`

mcovarr: I added to the comment above, does that make it more clear what's happening?

cjllanwarne: My point was more that, if we know the `.get` is fine, then it's fine to alter the method signature to reflect that.

I see they're overrides but in that case the thing calling them must be deciding based on the Option whether to call them or not?

mcovarr: The WorkflowActor decides based on whether it receives an `InitializationSuccess` or `InitializationFailure` whether to proceed to workflow execution, but it never actually looks at the initialization data.  Most of the current backends don't actually care about initialization data and leave it `None`.

Horneth:
👍

cjllanwarne:
👍

PR #1291, opened 2016-08-12
Call caching hashing rebased on Call Cache reading

cjllanwarne:
With compliments to @mcovarr for the rebase base.

Creates hashes for the following:

- command
- backend name
- output expression
- non-file inputs (as simpletons)
- file input paths (according to config)

Not included in this PR:

- backend specific hashes (runtime attributes, docker, file contents)

Note that if you want anything to actually be written you'll want the following options (to avoid a hashing failure)

- `lookup-docker-hash=false`
- `hash-docker-names=false`
- `hash-file-paths=true` -- actually you could leave this false but... then you'd always cache hit regardless of what files you're using!
- `hash-file-contents=false`


core/src/main/resources/application.conf
```diff
@@ -118,6 +120,11 @@ call-caching {
   // request to an external service (DockerHub, GCR).  If a call fails to lookup a
   // Docker hash, it will fail.
   lookup-docker-hash = false
+
+  // This option determines whether file contents are hashed to determine cache hit or miss.
+  hash-file-contents = true
+  // This option determines whether file paths are hashed to determine cache hit or miss.
+  hash-file-paths = false
```
mcovarr: Would it ever make sense to have both Docker hashing modes or file hashing modes turned on?  I wonder if these would make more sense as enumerated values:

```
call-caching {
  docker-hashing = "name"  # Could also be 'off' or 'lookup'
  file-hashing = "path" # Could also be 'off' or 'lookup'
}
```

core/src/main/scala/cromwell/core/simpleton/WdlValueSimpleton.scala
```diff
@@ -0,0 +1,21 @@
...
+      case prim: WdlPrimitive => List(WdlValueSimpleton(name, prim))
+      case WdlArray(_, arrayValue) => arrayValue.zipWithIndex flatMap { case (arrayItem, index) => arrayItem.simplify(s"$name[$index]") }
+      case WdlMap(_, mapValue) => mapValue flatMap { case (key, value) => value.simplify(s"$name:${key.valueString}") }
+      case wdlObject: WdlObjectLike => wdlObject.value flatMap { case (key, value) => value.simplify(s"$name:$key") }
+    }
+  }
+
+  implicit class WdlValuesSimplifier(wdlValues: Map[String, WdlValue]) {
+    def simplify: Iterable[WdlValueSimpleton] = wdlValues flatMap { case (name, value) => value.simplify(name) }
+  }
```
mcovarr: pretty cool!  😄

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -21,10 +27,21 @@ class EngineJobExecutionActor(jobKey: BackendJobDescriptorKey,
...
                               jobStoreActor: ActorRef,
+                              backendName: String,
                               callCachingMode: CallCachingMode) extends LoggingFSM[EngineJobExecutionActorState, EJEAData] with WorkflowLogging {
 
   override val workflowId = executionData.workflowDescriptor.id
 
+  val tag = s"EJEA [UUID(${workflowId.shortString}/${jobKey.call.fullyQualifiedName}:${jobKey.index.fromIndex}(attempt ${jobKey.attempt}))]"
+
+  // There's no need to check for a cache hit again if:
+  // - We got preempted
```
mcovarr: Good point about the preemption, but a bullet list with one item?  😛

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -21,10 +27,21 @@ class EngineJobExecutionActor(jobKey: BackendJobDescriptorKey,
...
+  // NB: this can also change (e.g. if we have a HashError we just force this to CallCachingOff)
+  var effectiveCallCachingMode = if (jobKey.attempt > 1) { callCachingMode.withoutRead } else { callCachingMode }
```
mcovarr: decurly

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -59,15 +76,15 @@ class EngineJobExecutionActor(jobKey: BackendJobDescriptorKey,
           stay()
       }
     case Event(f: JobStoreReadFailure, NoData) =>
-      log.error(f.reason, "Error reading from JobStore for " + jobKey)
+      log.error("{}: Error reading from JobStore: {}", tag, f.reason)
```
mcovarr: Shouldn't this be:

```
log.error(f.reason, "{}: Error reading from JobStore", tag)
```
to get the stack trace?  Unless you don't want the stack trace, which maybe you don't.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -84,24 +101,36 @@ class EngineJobExecutionActor(jobKey: BackendJobDescriptorKey,
 
   // When CheckingCallCache, the FSM always has EJEAJobDescriptorData
   when(CheckingCallCache) {
+    case Event(HashError(t), EJEAJobDescriptorData(Some(jobDescriptor), Some(bjeaProps))) =>
+      writeToMetadata(Map("Call caching read result" -> s"Hashing Error: ${t.getMessage}"))
```
mcovarr: The Map key should be a constant

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -84,24 +101,36 @@ class EngineJobExecutionActor(jobKey: BackendJobDescriptorKey,
...
       saveCacheResults(EJEASuccessfulCompletionDataWithHashes(response, hashes))
     case Event(hashes: CallCacheHashes, data @ EJEAPartialCompletionData(None, None)) =>
-      stay using data.copy(hashes = Option(hashes))
+      stay using data.copy(hashes = Option(Success(hashes)))
+    case Event(HashError(t), data @ EJEAPartialCompletionData(Some(response: BackendJobExecutionResponse), _)) =>
+      recordHashError(t)
+      saveJobCompletionToJobStore(response)
+    case Event(HashError(t), data @ EJEAPartialCompletionData(None, _)) =>
+      recordHashError(t)
+      stay using data.copy(hashes = Some(Failure(t)))
```
mcovarr: Option

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
mcovarr: could this go in the state data rather than being a var?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -242,15 +287,19 @@ case class EJEAJobDescriptorData(jobDescriptor: Option[BackendJobDescriptor], bj
   override def toString = s"EJEAJobDescriptorData(backendJobDescriptor: ${jobDescriptor.isDefined}, bjeaActorProps: ${bjeaActorProps.isDefined})"
 }
 
-case class EJEAPartialCompletionData(jobResult: Option[BackendJobExecutionResponse], hashes: Option[CallCacheHashes]) extends EJEAData {
-  override def toString = s"EJEAPartialCompletionData(jobResult: ${jobResult.isDefined}, hashes: ${hashes.isDefined})"
+case class EJEAPartialCompletionData(jobResult: Option[BackendJobExecutionResponse], hashes: Option[Try[CallCacheHashes]]) extends EJEAData {
+  override def toString = s"EJEAPartialCompletionData(jobResult: ${jobResult.isDefined}, hashes: ${hashes.map(_.getClass.getSimpleName).getOrElse("None")})"
```
mcovarr: that second expression is kind of a lot for a string interpolation

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -242,15 +287,19 @@ case class EJEAJobDescriptorData(jobDescriptor: Option[BackendJobDescriptor], bj
...
 
-// Le sigh. Can't do case-to-case inheritance so let's jump through these extractor-pattern hoops...
+// Can't do case-to-case inheritance so let's jump through these extractor-pattern hoops...
 class EJEACompletionData(val jobResult: BackendJobExecutionResponse) extends EJEAData {
   override def toString = s"EJEACompletionData(${jobResult.getClass.getSimpleName})"
 }
-object EJEACompletionData { def apply(jobResult: BackendJobExecutionResponse) = new EJEACompletionData(jobResult); def unapply(completionData: EJEACompletionData) = Some(completionData.jobResult) }
+object EJEACompletionData {
+  def apply(jobResult: BackendJobExecutionResponse) = new EJEACompletionData(jobResult)
+  def unapply(completionData: EJEACompletionData) = Some(completionData.jobResult)
```
mcovarr: Option

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
geoffjentry: and at the very least private. There should never be a non-private var anywhere

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -198,6 +232,15 @@ class EngineJobExecutionActor(jobKey: BackendJobDescriptorKey,
     val jobStoreResult = JobResultFailure(returnCode, reason, retryable)
     jobStoreActor ! RegisterJobCompleted(jobStoreKey, jobStoreResult)
   }
+
+  private def writeToMetadata(keyValue: Map[String, String]) = {
+    val events = keyValue map { case (key, value) =>
+      val metadataKey = MetadataKey(workflowId, Some(MetadataJobKey(jobKey.call.fullyQualifiedName, jobKey.index, jobKey.attempt)), key)
```
mcovarr: Option

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/EngineJobHashingActor.scala
```diff
@@ -83,69 +143,88 @@ case class EngineJobHashingActor(jobDescriptor: BackendJobDescriptor,
...
+  def intersectCacheResults(newCacheResults: CacheResultMatchesForHashes) = {
     val newIds = newCacheResults.cacheResultIds
-    val intersectedIds = if (cacheResultsIntersected.nonEmpty) possibleCacheResults.intersect(newIds) else newIds
+    val intersectedIds = if (keysCheckedAgainstCache.nonEmpty) possibleCacheResults.intersect(newIds) else newIds
     this.copy(
       possibleCacheResults = intersectedIds,
-      cacheResultsIntersected = cacheResultsIntersected ++ newCacheResults.hashResults.map(_.hashKey))
+      keysCheckedAgainstCache = keysCheckedAgainstCache ++ newCacheResults.hashResults.map(_.hashKey))
+  }
+  def withNewKnownHashes(hashResults: Set[HashResult]) = this.copy(hashesKnown = hashesKnown ++ hashResults) // NB doesn't update hashesNeeded which should be considered immutable
```
mcovarr: As you foreshadowed in my PR, `hashesNeeded` is no longer considered immutable.

core/src/main/resources/application.conf
mcovarr: Actually, 'off' doesn't really make sense for either of these, does it?  So perhaps default to "name" and "path" respectively, with the option to upgrade either to "lookup".

geoffjentry: I believe the argument (at least for docker hashing) was that the simple case was so simple we might as well do it anyways. It'd also enable a case where e.g. I write a job's hashes with both and then later am running with only the docker name, I'd still be able to get a cache hit

I could see a similar argument for the file thing

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -148,14 +182,14 @@ class EngineJobExecutionActor(jobKey: BackendJobDescriptorKey,
   }
 
   def initializeJobHashing(jobDescriptor: BackendJobDescriptor, activity: CallCachingActivity) = {
-    val fileHasherActor = context.actorOf(FileHasherActor.props)
-    context.actorOf(EngineJobHashingActor.props(jobDescriptor, fileHasherActor, activity))
+    val fileHasherActor = context.actorOf(BackendSpecificHasherActor.props(effectiveCallCachingMode))
+    context.actorOf(EngineJobHashingActor.props(jobDescriptor, fileHasherActor, backendName, activity))
```
mcovarr: These actors should get names

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -198,6 +232,15 @@ class EngineJobExecutionActor(jobKey: BackendJobDescriptorKey,
...
   }
+
+  private def writeToMetadata(keyValue: Map[String, String]) = {
+    val events = keyValue map { case (key, value) =>
+      val metadataKey = MetadataKey(workflowId, Some(MetadataJobKey(jobKey.call.fullyQualifiedName, jobKey.index, jobKey.attempt)), key)
+      MetadataEvent(metadataKey, MetadataValue(value))
+    }
+
+    serviceRegistryActor ! PutMetadataAction(events)
+  }
```
mcovarr: Is there really not a method in a common place that does this already?

core/src/main/resources/application.conf
mcovarr: Yes exactly, which is why I don't think we should support 'off'.

mcovarr: If you ran once with both Docker hashes on, and then a second time with only the name hashing on, would you really expect a cache hit?  It seems we'd have to record aggregate hashes for the first run with and without the docker image hash included.

geoffjentry: I could see myself wanting that behavior. Im telling it I don't care about the real docker hash, so I shouldn't care

mcovarr: Ok, I can buy that.  So for the Docker case, if we have:

```
call-caching {
  docker-hashing: "lookup"   # Could be "name"
}
```

In Docker "lookup" mode, Cromwell could write one aggregate hash using the Docker image name but *not* the Docker image hash, and a second aggregate hash using the Docker image hash but *not* the Docker image name.  If a subsequent call reads the cache in Docker "lookup" mode, its Docker image hash would be included in its aggregate hash for cache hit determination, but not its name.  In "name" mode, the opposite would be true.

Or do we really want to support both Docker name and image hashes as part of one aggregate hash?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/BackendSpecificHasherActor.scala
```diff
@@ -0,0 +1,50 @@
...
+import cromwell.core.JobKey
+import cromwell.engine.workflow.lifecycle.execution.callcaching.EngineJobHashingActor.{HashingFailedMessage, PlaceholderHashKeyExpansion, SuccessfulHashResultMessage}
+import cromwell.engine.workflow.lifecycle.execution.callcaching.BackendSpecificHasherActor._
+import org.apache.commons.lang3.NotImplementedException
+import wdl4s.values.WdlFile
+
+/**
+  * Used if a backend has not provided a backend hasher actor.
+  *
+  * Backend specific hasher actors should extend and reply to commands in the same way.
```
mcovarr: I think this is what you meant, but just to be clear:

```scala

class FooBackendHasherActor extends BackendSpecificHasherActor... {
  private def fooReceive: Receive = ...

  override def receive = fooReceive orElse super.receive
}
```

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/BackendSpecificHasherActor.scala
```diff
@@ -0,0 +1,50 @@
...
+
+object BackendSpecificHasherActor {
+
+  def props(mode: CallCachingMode): Props = Props(new BackendSpecificHasherActor(mode))
+
+  case class SingleFileHashRequest(hashKey: HashKey, file: WdlFile)
+
+  sealed trait BackendSpecificHasherCommand
+  case class JobFileHashRequests(jobKey: JobKey, files: Iterable[SingleFileHashRequest]) extends BackendSpecificHasherCommand
+  case class RuntimeAttributesHashesRequest(placeholderHashKey: HashKey, jobDescriptor: BackendJobDescriptor)
```
mcovarr: I think this one was meant to extend the trait

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/EngineJobHashingActor.scala
```diff
@@ -1,53 +1,112 @@
...
+    case Event(PlaceholderHashKeyExpansion(placeholderHashKey, newHashKeys), _) => checkWhetherAllHashesAreKnownAndTransition(stateData.replacePlaceholderHashKey(placeholderHashKey, newHashKeys))
     case Event(CacheResultMatchesForHashes(_, _), _) => stay // Don't care; we already know the hit/miss status. Ignore this message
   }
 
+  whenUnhandled {
+    case Event(failure: CacheResultLookupFailure, _) =>
+      // Crash and let the supervisor deal with it.
+      throw new RuntimeException("Failure looking up call cache results", failure.reason)
+    case Event(HashingFailedMessage(hashKey, reason), _) =>
+      receiver ! HashError(new Exception(s"Unable to generate ${hashKey.key} hash. Caused by ${reason.getMessage}", reason))
```
mcovarr: so we're saying this is not exceptional enough to crash?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/package.scala
```diff
@@ -1,28 +1,71 @@
...
+    def readFromCache = false
+    def writeToCache = false
+    def hashDockerNames = false
+    def lookupDockerHashes: Boolean = false
+    def hashFilePaths: Boolean = false
+    def hashFileContents: Boolean = false
+
+    /**
+      * Return an equivalent of this call caching mode with READ disabled.
+      * Especially useful to not re-check the cache again before the second attempt of a job.
```
mcovarr: "before retrying a preempted job" would be more generally correct

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/EngineJobHashingActor.scala
```diff
@@ -83,69 +143,88 @@ case class EngineJobHashingActor(jobDescriptor: BackendJobDescriptor,
...
     this.copy(
       possibleCacheResults = intersectedIds,
-      cacheResultsIntersected = cacheResultsIntersected ++ newCacheResults.hashResults.map(_.hashKey))
+      keysCheckedAgainstCache = keysCheckedAgainstCache ++ newCacheResults.hashResults.map(_.hashKey))
+  }
+  def withNewKnownHashes(hashResults: Set[HashResult]) = this.copy(hashesKnown = hashesKnown ++ hashResults) // NB doesn't update hashesNeeded which should be considered immutable
+
+  def replacePlaceholderHashKey(placeholderHashKey: HashKey, newHashKeys: Iterable[HashKey]) = {
+    val newHashesNeeded = this.hashesNeeded ++ newHashKeys filter {
+      case `placeholderHashKey` => false // We don't want this!
```
mcovarr: sneaky backticks!  😄

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/package.scala
```diff
@@ -1,28 +1,71 @@
...
   // TODO: Find somewhere better for all these?
   case class HashKey(key: String)
   case class HashValue(value: String)
   case class HashResult(hashKey: HashKey, hashValue: HashValue)
 
-  private[callcaching] case class CacheResultMatchesForHashes(hashResults: Iterable[HashResult], cacheResultIds: Set[Int])
+  object HashValue {
+    implicit class StringMd5er(unhashedString: String) {
+      def md5HashValue: HashValue = {
+        val hashBytes = java.security.MessageDigest.getInstance("MD5").digest(unhashedString.getBytes)
```
mcovarr: is it worth caching this instance in a `lazy val` in the `object`?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/EngineJobHashingActor.scala
```diff
@@ -1,53 +1,112 @@
...
+    }
+
+    val initialHashes = calculateInitialHashes(nonFileInputSimpletons, fileInputSimpletons)
+
+    val fileContentHashesNeeded = if (mode.hashFileContents) {
+      JobFileHashRequests(jobDescriptor.key, fileInputSimpletons map { case WdlValueSimpleton(name, x: WdlFile) => SingleFileHashRequest(HashKey(s"input: File(Content) $name"), x) })
+    } else {
+      JobFileHashRequests(jobDescriptor.key, List.empty)
+    }
+    val runtimeAttributeHashesNeededPlaceholder = HashKey("PLACEHOLDER: runtime attributes")
```
mcovarr: Is there value in having an explicit placeholder `HashKey`?  It seems like this just travels from here to the backend specific hasher actor and back.  It gets added to `hashesNeeded` below but then filtered out.

core/src/main/resources/application.conf
cjllanwarne: Let's TT this

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
cjllanwarne: I was hoping I could also add "restarted calls" - but unfortunately the WF only tells us whether we're in a restarted workflow, which isn't good enough.

OTOH, the impact isn't actually too bad since most restarted jobs will be short-circuited already by the JobStore

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/EngineJobHashingActor.scala
cjllanwarne: Let's TT this. I believe so, but may be wrong

core/src/main/resources/application.conf
cjllanwarne: Side note: being able to turn both docker things off is handy if your backend hasher isn't implemented yet!

README.md
```diff
@@ -1209,14 +1209,22 @@ To enable Call Caching, add the following to your Cromwell [configuration](#conf
...
+  hash-docker-names = true
   lookup-docker-hash = false
+  hash-file-contents = true
+  hash-file-paths = false
 }
 ```
 
-When `call-caching.enabled=true`, Cromwell will add completed calls to the cache as well as do a cache lookup before running any call.
+When `call-caching.enabled=true` (default: `false`), Cromwell will be able to to copy results from previously run jobs (when appropriate).
+
```
geoffjentry: @cjllanwarne To poke at your "flexibility" argument, if you want this stuff to be configurable, why stop here? Why not make every component able to be turned off?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
cjllanwarne: 😞

cjllanwarne: I'm on a mission to make people feel bad for calling me out on this (but I will change it): 

I'm having to move against "what I mean" in the code and move towards "arbitrary rule". It makes things harder to read and harder to understand for 0 benefit in the end result.

core/src/main/resources/application.conf
```diff
@@ -103,10 +103,12 @@ workflow-options {
   //workflow-failure-mode: "ContinueWhilePossible"
 }
 
-// Optional call-caching configuration. Disabled by default.
+// Optional call-caching configuration.
 call-caching {
   enabled = false
 
+  // This option determines whether the name of a docker image should be hashed (if docker is specified for a task)
+  hash-docker-names = true
```
geoffjentry: To be clear about what I was advocating for, I think we should *always* hash the docker tuple and file paths. By default we also hash from dockerhub and the file hash themselves, but one can disable that. If we have the dockerhub and/or the file hash we'll only use those, not the "simple" variants (i.e. if i'm checking full file hashes I don't care if the path isn't the same)

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/EngineJobHashingActor.scala
cjllanwarne: We want to continue but treat it like a cache miss and not write anything new to the cache.

I could **probably** do the same with crashes and state-specific recovery options in the EJEA... but this just feels so much neater

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/package.scala
cjllanwarne: until we have other reasons for retries... :)

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/package.scala
cjllanwarne: Looking at the docs, I think you need a new instance per hash (otherwise everything aggregates with the previous hash result)

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/EngineJobHashingActor.scala
cjllanwarne: I'll try to rework this

core/src/main/resources/application.conf
cjllanwarne: We should sort out these details in TTs not PR comments. 

For example, we don't want to always hash file paths because the contents might match even when the paths are different.

geoffjentry: I said that in the message you're replying to.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -243,15 +289,22 @@ case class EJEAJobDescriptorData(jobDescriptor: Option[BackendJobDescriptor], bj
...
-  override def toString = s"EJEAPartialCompletionData(jobResult: ${jobResult.isDefined}, hashes: ${hashes.isDefined})"
+case class EJEAPartialCompletionData(jobResult: Option[BackendJobExecutionResponse], hashes: Option[Try[CallCacheHashes]]) extends EJEAData {
+    override def toString = {
+      val hashesString = hashes.map(_.getClass.getSimpleName).getOrElse("None")
+      s"EJEAPartialCompletionData(jobResult: ${jobResult.isDefined}, hashes: $hashesString)"
+    }
 }
 
-// Le sigh. Can't do case-to-case inheritance so let's jump through these extractor-pattern hoops...
+// Can't do case-to-case inheritance so let's jump through these extractor-pattern hoops...
```
geoffjentry: Nice job removing the 🇫🇷

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/package.scala
```diff
@@ -9,23 +9,62 @@ package object callcaching {
...
-      case (true, true) => ReadAndWriteCache(lookupDockerHashes)
+      case (false, true) => WriteCache(hashDockerNames, lookupDockerHashes, hashFilePaths, hashFileContents)
+      case (true, false) => ReadCache(hashDockerNames, lookupDockerHashes, hashFilePaths, hashFileContents)
+      case (true, true) => ReadAndWriteCache(hashDockerNames, lookupDockerHashes, hashFilePaths, hashFileContents)
     }
   }
 
-  case object CallCachingOff extends CallCachingMode { override def activity = None; }
+  case object CallCachingOff extends CallCachingMode {
+    override def activity = None
```
geoffjentry: I don't care enough to say to change it (feel free not to), but I'm chiming in here just to advocate what I think is a better pattern on ADTs. I find that for ADT hierarchies with simple overrides like this the FP style of implementation is cleaner, e.g.

    sealed trait parent {
      def foo = this match {
        case childA => "foo"
        case childB => "bar"
     }
    }
    case object childA extends parent
    case object childB extends parent

over the OO ADT style here.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/package.scala
```diff
@@ -9,23 +9,62 @@ package object callcaching {
...
-  case class ReadCache(override val lookupDockerHashes: Boolean) extends CallCachingActivity { override def readFromCache = true }
-  case class WriteCache(override val lookupDockerHashes: Boolean) extends CallCachingActivity { override def writeToCache = true }
-  case class ReadAndWriteCache(override val lookupDockerHashes: Boolean) extends CallCachingActivity { override def readFromCache = true; override def writeToCache = true }
+  sealed trait CallCachingActivity extends CallCachingMode {
+    override def activity = Option(this)
+  }
+  case class ReadCache(override val hashDockerNames: Boolean, override val lookupDockerHashes: Boolean, override val hashFilePaths: Boolean, override val hashFileContents: Boolean) extends CallCachingActivity {
+    override def readFromCache = true
+    override def withoutRead = CallCachingOff
+  }
```
geoffjentry: \n

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/package.scala
```diff
@@ -9,23 +9,62 @@ package object callcaching {
...
 
-  sealed trait CallCachingActivity extends CallCachingMode { override def activity = Option(this); }
-  case class ReadCache(override val lookupDockerHashes: Boolean) extends CallCachingActivity { override def readFromCache = true }
-  case class WriteCache(override val lookupDockerHashes: Boolean) extends CallCachingActivity { override def writeToCache = true }
-  case class ReadAndWriteCache(override val lookupDockerHashes: Boolean) extends CallCachingActivity { override def readFromCache = true; override def writeToCache = true }
+  sealed trait CallCachingActivity extends CallCachingMode {
+    override def activity = Option(this)
+  }
+  case class ReadCache(override val hashDockerNames: Boolean, override val lookupDockerHashes: Boolean, override val hashFilePaths: Boolean, override val hashFileContents: Boolean) extends CallCachingActivity {
```
geoffjentry: If you need to put all the junk in the signature, are you getting anything out of it being a case class? If not, just make it a class

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/package.scala
```diff
@@ -9,23 +9,62 @@ package object callcaching {
...
+    override def readFromCache = true
+    override def withoutRead = CallCachingOff
+  }
+  case class WriteCache(override val hashDockerNames: Boolean, override val lookupDockerHashes: Boolean, override val hashFilePaths: Boolean, override val hashFileContents: Boolean) extends CallCachingActivity {
```
geoffjentry: you shouldn't need the `override val` stuff here. 

Also these case classes should be final

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/package.scala
```diff
@@ -9,23 +9,62 @@ package object callcaching {
...
+    override def writeToCache = true
```
geoffjentry: these should be vals

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -22,10 +28,22 @@ class EngineJobExecutionActor(jobKey: BackendJobDescriptorKey,
...
                               serviceRegistryActor: ActorRef,
                               jobStoreActor: ActorRef,
+                              backendName: String,
                               callCachingMode: CallCachingMode) extends LoggingFSM[EngineJobExecutionActorState, EJEAData] with WorkflowLogging {
 
   override val workflowId = executionData.workflowDescriptor.id
 
+  val tag = s"EJEA [UUID(${workflowId.shortString}/${jobKey.call.fullyQualifiedName}:${jobKey.index.fromIndex}(attempt ${jobKey.attempt}))]"
+
+  // There's no need to check for a cache hit again if we got preempted
```
geoffjentry: good catch, i'd never have thought of that

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -22,10 +28,22 @@ class EngineJobExecutionActor(jobKey: BackendJobDescriptorKey,
...
+  // NB: this can also change (e.g. if we have a HashError we just force this to CallCachingOff)
+  private var effectiveCallCachingMode = if (jobKey.attempt > 1) { callCachingMode.withoutRead } else { callCachingMode }
```
geoffjentry: I'm kind of curious why the king of FSM love chose to manage state via a var in this one case

core/src/main/scala/cromwell/core/simpleton/WdlValueSimpleton.scala
cjllanwarne: I presumed that this'll be useful for the JobStore simpletonians ticket too?

mcovarr: I've temporarily copy/pasted it into my branch until this branch gets merged to develop, so yeah.  😄

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -22,10 +24,23 @@ class EngineJobExecutionActor(jobKey: BackendJobDescriptorKey,
...
                               callCachingMode: CallCachingMode) extends LoggingFSM[EngineJobExecutionActorState, EJEAData] with WorkflowLogging {
 
   override val workflowId = executionData.workflowDescriptor.id
 
+  val jobTag = s"${workflowId.shortString}:${jobKey.call.fullyQualifiedName}:${jobKey.index.fromIndex}:${jobKey.attempt}"
+  val tag = s"EJEA_$jobTag"
+
+  // There's no need to check for a cache hit again if we got preempted
+  // NB: this can also change (e.g. if we have a HashError we just force this to CallCachingOff)
+  private var effectiveCallCachingMode = if (jobKey.attempt > 1) { callCachingMode.withoutRead } else { callCachingMode }
```
mcovarr: curlies

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -84,54 +99,69 @@ class EngineJobExecutionActor(jobKey: BackendJobDescriptorKey,
...
+    case Event(HashError(t), data @ PartialCompletionDataWithSucceededResponse(response)) =>
+      recordHashError(t)
+      saveJobCompletionToJobStore(response)
+    case Event(HashError(t), data @ EmptyPartialCompletionData) =>
+      recordHashError(t)
+      stay using PartialCompletionDataWithHashes(Failure(t))
+
+    case Event(response: SucceededResponse, PartialCompletionDataWithHashes(Success(hashes))) if effectiveCallCachingMode.writeToCache =>
+      saveCacheResults(CacheWriteOnCompletionData(response, hashes))
+    case Event(response: SucceededResponse, data @ EmptyPartialCompletionData) if effectiveCallCachingMode.writeToCache =>
```
mcovarr: capturing `data` here and on several other lines above, but then not doing anything with it

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -84,54 +99,69 @@ class EngineJobExecutionActor(jobKey: BackendJobDescriptorKey,
...
       saveJobCompletionToJobStore(response)
   }
 
   // When WritingToCallCache, the FSM always has EJEASuccessfulCompletionDataWithHashes
   when(UpdatingCallCache) {
-    case Event(CallCacheWriteSuccess, data @ EJEASuccessfulCompletionDataWithHashes(response, _)) =>
+    case Event(CallCacheWriteSuccess, data @ CacheWriteOnCompletionData(response, _)) =>
       saveJobCompletionToJobStore(response)
-    case Event(CallCacheWriteFailure(reason), data @ EJEASuccessfulCompletionDataWithHashes(response, _)) =>
+    case Event(CallCacheWriteFailure(reason), data @ CacheWriteOnCompletionData(response, _)) =>
```
mcovarr: more `data` captures here and above

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -233,25 +274,19 @@ object EngineJobExecutionActor {
...
-case class EJEAJobDescriptorData(jobDescriptor: Option[BackendJobDescriptor], bjeaActorProps: Option[Props]) extends EJEAData {
-  override def toString = s"EJEAJobDescriptorData(backendJobDescriptor: ${jobDescriptor.isDefined}, bjeaActorProps: ${bjeaActorProps.isDefined})"
-}
+private[execution] sealed trait EJEAData { override def toString = getClass.getSimpleName }
 
-case class EJEAPartialCompletionData(jobResult: Option[BackendJobExecutionResponse], hashes: Option[CallCacheHashes]) extends EJEAData {
-  override def toString = s"EJEAPartialCompletionData(jobResult: ${jobResult.isDefined}, hashes: ${hashes.isDefined})"
-}
+private[execution] case object NoData extends EJEAData
+private[execution] case class JobDescriptorData(jobDescriptor: BackendJobDescriptor, bjeaActorProps: Props) extends EJEAData
```
geoffjentry: make case classes final.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/BackendSpecificHasherActor.scala
```diff
@@ -0,0 +1,54 @@
...
+    // Since all these methods are sync, we'll never have any work outstanding. So we don't need to do anything.
+    case cancellation: HashesNoLongerRequired => // Nothing
+  }
+}
+
+object BackendSpecificHasherActor {
+
+  def props(mode: CallCachingActivity): Props = Props(new BackendSpecificHasherActor(mode))
+
+  case class SingleFileHashRequest(hashKey: HashKey, file: WdlFile)
```
geoffjentry: more finals

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/package.scala
```diff
@@ -6,26 +6,60 @@ package object callcaching {
 
   // TODO: Find somewhere better for all these?
   case class HashKey(key: String)
+  object RuntimeAttributeHashKeyPlaceholder extends HashKey("PLACEHOLDER: Runtime Attributes")
+
   case class HashValue(value: String)
   case class HashResult(hashKey: HashKey, hashValue: HashValue)
 
+  object HashValue {
```
geoffjentry: If something requires its own object it doesn't belong in the package object

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/EngineJobHashingActor.scala
```diff
@@ -151,7 +207,15 @@ private[callcaching] case class EJHAData(possibleCacheResults: Set[MetaInfoId],
       possibleCacheResults = intersectedIds,
       keysCheckedAgainstCache = keysCheckedAgainstCache ++ newCacheResults.hashResults.map(_.hashKey))
   }
-  def withNewKnownHashes(hashResults: Set[HashResult]) = this.copy(hashesKnown = hashesKnown ++ hashResults) // NB doesn't update hashesNeeded which should be considered immutable
+  def withNewKnownHashes(hashResults: Set[HashResult]) = this.copy(hashesKnown = hashesKnown ++ hashResults)
+
+  def replacePlaceholderHashKey(placeholderHashKey: HashKey, newHashKeys: Iterable[HashKey]) = {
+    val newHashesNeeded = this.hashesNeeded ++ newHashKeys filter {
+      case `placeholderHashKey` => false // We d  on't want this!
```
geoffjentry: "we don't"

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/package.scala
```diff
@@ -6,26 +6,60 @@ package object callcaching {
...
 
-  sealed trait CallCachingMode { def activity: Option[CallCachingActivity]; def readFromCache = false; def writeToCache = false; def lookupDockerHashes: Boolean = false }
-  object CallCachingMode {
-    def apply(read: Boolean, write: Boolean, lookupDockerHashes: Boolean): CallCachingMode = (read, write) match {
-      case (false, false) => CallCachingOff
-      case (false, true) => WriteCache(lookupDockerHashes)
-      case (true, false) => ReadCache(lookupDockerHashes)
-      case (true, true) => ReadAndWriteCache(lookupDockerHashes)
-    }
+  sealed trait CallCachingMode {
```
geoffjentry: 40 line ADT hierarchies don't belong in a package object

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -22,10 +24,23 @@ class EngineJobExecutionActor(jobKey: BackendJobDescriptorKey,
...
   override val workflowId = executionData.workflowDescriptor.id
 
+  val jobTag = s"${workflowId.shortString}:${jobKey.call.fullyQualifiedName}:${jobKey.index.fromIndex}:${jobKey.attempt}"
+  val tag = s"EJEA_$jobTag"
+
+  // There's no need to check for a cache hit again if we got preempted
+  // NB: this can also change (e.g. if we have a HashError we just force this to CallCachingOff)
+  private var effectiveCallCachingMode = if (jobKey.attempt > 1) callCachingMode.withoutRead else callCachingMode
+
+  val effectiveCallCachingKey = "Effective call caching mode"
```
mcovarr: TOL could be private and in a companion object

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/BackendSpecificHasherActor.scala
```diff
@@ -0,0 +1,54 @@
...
+  * Used if a backend has not provided a backend hasher actor.
+  *
+  * Backend specific hasher actors could extend like: ** override def receive = fooReceive orElse super.receive **
+  */
+class BackendSpecificHasherActor(mode: CallCachingActivity) extends Actor with ActorLogging {
+
+  override def receive: Receive = {
+    // We can't do file hashes, so respond with a bunch of failures.
+    case JobFileHashRequests(jobKey, fileHashRequests) =>
+      fileHashRequests map { fileHashRequest => HashingFailedMessage(fileHashRequest.hashKey, new NotImplementedException("This type of hashing is not implemented for this backend")) } foreach { sender ! _ }
```
mcovarr: Is this message going to show up with appropriate job and backend context?

mcovarr:
:+1:

cjllanwarne:
Closing because superceded by later work

PR #1293, opened 2016-08-14
Add an endpoint to report supported backends

geoffjentry:
To enable some of the Centaur changes we discussed on Friday

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,7 @@ A [Workflow Management System](https://en.wikipedia.org/wiki/Workflow_management
   * [POST /api/workflows/:version/:id/abort](#post-apiworkflowsversionidabort)
   * [POST /api/workflows/:version/:id/call-caching](#post-apiworkflowsversionidcall-caching)
   * [POST /api/workflows/:version/:id/call-caching/:call](#post-apiworkflowsversionidcall-cachingcall)
+  * [GET /api/workflows/:version/backends](#get-apiworkflowsversionbackends)
   * [Error handling](#error-handling)
 * [Developer](#developer)
   * [Generating table of contents on Markdown files](#generating-table-of-contents-on-markdown-files)
@@ -2300,6 +2301,35 @@ Server: spray-can/1.3.3
     "updateCount": 1
 }
 
+```
+
+
+## GET /api/workflows/:version/backends
+
+This endpoint returns a list of the backends supported by the server as well as the default backend.
+
+cURL:
+```
+$ curl http://localhost:8000/api/workflows/v1/backends
+```
+
+HTTPie:
+```
+$ http http://localhost:8000/api/workflows/v1/backends
+```
+
+Response:
+```
+HTTP/1.1 200 OK
+Content-Length: 379
+Content-Type: application/json; charset=UTF-8
+Date: Mon, 03 Aug 2015 17:11:28 GMT
+Server: spray-can/1.3.3
+
+{
+  "supportedBackends": ["JES", "LSF", "Local", "SGE"],
+  "defaultBackend": "Local"
+}
 ```
 
 ## Error handling
--- a/engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
+++ b/engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
@@ -4,6 +4,7 @@ import akka.actor._
 import cromwell.core.{WorkflowId, WorkflowSourceFiles}
 import cromwell.services.metadata.MetadataService
 import MetadataService.{ValidateWorkflowIdAndExecute, _}
+import cromwell.engine.backend.BackendConfiguration
 import cromwell.webservice.WorkflowJsonSupport._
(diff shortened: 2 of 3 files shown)


engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
```diff
@@ -184,4 +185,16 @@ trait CromwellApiService extends HttpService with PerRequestCreator {
...
+    path("workflows" / Segment / "backends") { version =>
+      get {
+        val backends = BackendConfiguration.AllBackendEntries map { _.name }
+        complete {
+          JsObject(Map("backends" -> backends.toJson))
+        }
+      }
+    }
+
+  val backends = BackendConfiguration.AllBackendEntries map { _.name } toJson
```
geoffjentry: I considered both sorting the list alphabetically as well as including what the default was. I didn't need this for my purposes so laziness took over, but throwing it out there for reviewers

README.md
```diff
@@ -2302,6 +2303,34 @@ Server: spray-can/1.3.3
...
+```
+
+HTTPie:
+```
+$ http http://localhost:8000/api/workflows/v1/backends
+```
+
+Response:
+```
+HTTP/1.1 200 OK
```
geoffjentry: TOL: Why do we include this header junk in the docs? It doesn't seem useful. Full disclosure: I copy/pasted from elsewhere

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
geoffjentry: I lied. I'm going to do the above, as I'd rather be lazy on the Centaur side for now and the above will enable just that

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
```diff
@@ -184,4 +185,21 @@ trait CromwellApiService extends HttpService with PerRequestCreator {
         }
       }
     }
-}
\ No newline at end of file
+
+  def backendRoute =
+    path("workflows" / Segment / "backends") { version =>
+      get {
+        val backends = BackendConfiguration.AllBackendEntries map { _.name }
```
geoffjentry: This didn't seem worth the pre-request actor

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
```diff
@@ -184,4 +185,20 @@ trait CromwellApiService extends HttpService with PerRequestCreator {
...
       }
     }
-}
\ No newline at end of file
+
+  def backendRoute =
+    path("workflows" / Segment / "backends") { version =>
+      get {
+        complete {
+          backendResponse
```
geoffjentry: This didn't seem worth a per request actor here

README.md
```diff
@@ -2302,6 +2303,35 @@ Server: spray-can/1.3.3
...
+Content-Length: 379
+Content-Type: application/json; charset=UTF-8
+Date: Mon, 03 Aug 2015 17:11:28 GMT
+Server: spray-can/1.3.3
+
+{
+  "supported_backends": ["JES", LSF", "Local", "SGE"],
```
mcovarr: missing the opening quote on LSF

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
```diff
@@ -12,7 +13,7 @@ import spray.http._
 import spray.httpx.SprayJsonSupport._
 import spray.json._
 import spray.routing._
-
+import scala.language.postfixOps
```
mcovarr: not clear why this was added

README.md
geoffjentry: Someone caught me hand editing the output

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
geoffjentry: For somehting which was de-postfixed

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
mcovarr: I'd at least put a comment here if you think it should stay like this.  As you've observed, things tend to get copy/pasted; it would be unfortunate if this was taken as the right way to write these endpoints generally.

mcovarr:
:+1:

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
geoffjentry: Sure.

README.md
ruchim: I wonder if its breaking change to just delete the header stuff.

geoffjentry: Spoke in person but this was just me floating a trial balloon to see if anyone really wanted this stuff around. I'll probably open a PR at some point when i'm poking at the readme again

ruchim:
👍 swagger yaml to be updated

PR #1305, opened 2016-08-18
WdlValues from WdlValueSimpletons.

mcovarr:
The extracted `WdlValueSimpleton` file was thieved from @cjllanwarne's magnum opus hashing branch and some metacharacter handling added to it.

### DIFF
--- a/core/src/main/scala/cromwell/core/simpleton/WdlValueBuilder.scala
+++ b/core/src/main/scala/cromwell/core/simpleton/WdlValueBuilder.scala
@@ -1,0 +1,151 @@
+package cromwell.core.simpleton
+
+import wdl4s.TaskOutput
+import wdl4s.types.{WdlArrayType, WdlMapType, WdlPrimitiveType, WdlType}
+import wdl4s.values.{WdlArray, WdlMap, WdlValue}
+
+import scala.language.postfixOps
+import cromwell.core.{JobOutput, JobOutputs}
+import cromwell.core.simpleton.WdlValueSimpleton._
+
+
+/** Builds arbitrary WdlValues from WdlValueSimpletons. */
+object WdlValueBuilder {
+
+  /**
+    * Looks for a WDL identifier possibly followed by a metacharacter and more stuff.
+    *
+    * Capture groups:
+    *
+    * <ol>
+    * <li>A WDL identifier</li>
+    * <li>Possibly a metacharacter and more stuff after the WDL identifier</li>
+    * </ol>
+    */
+  private val IdentifierAndPathPattern = "^([a-zA-Z][a-zA-Z0-9_]*)(.*)".r
+
+  /**
+    * Looks for an array element reference: square braces surrounding digits, possibly followed by a metacharacter and
+    * more stuff.
+    *
+    * Capture groups:
+    *
+    * <ol>
+    * <li>Square braces plus enclosed digits</li>
+    * <li>Only the enclosed digits</li>
+    * <li>Possibly a metacharacter and more stuff after the square brace enclosed digits</li>
+    * </ol>
+    */
+  private val ArrayElementPattern = raw"^(\[(\d+)\])(.*)".r
+
+  /**
+    * Looks for a map element reference: a colon followed by one or more non-metacharacters, possibly followed by a
+    * metacharacter and more stuff.  Any metacharacters in map keys have been escaped by `WdlValueSimpleton#escapeMeta`,
+    * which is taken account by the regular expression below.
+    *
+    * Capture groups:
+    *
+    * <ol>
+    * <li>A colon followed by one or more non-metacharacters, including `WdlValueSimpleton#escapeMeta` escaped metacharacters.</li>
+    * <li>Same as above except for the initial colon</li>
+    * <li>Possibly a metacharacter and more trailing stuff</li>
+    * </ol>
+    */
+  private val MapElementPattern = """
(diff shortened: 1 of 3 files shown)


core/src/main/scala/cromwell/core/simpleton/WdlValueBuilder.scala
```diff
@@ -0,0 +1,145 @@
...
+import wdl4s.values.{WdlArray, WdlMap, WdlValue}
+
+import scala.language.postfixOps
+import cromwell.core.simpleton.WdlValueSimpleton._
+
+
+/** Builds arbitrary WdlValues from WdlValueSimpletons. */
+object WdlValueBuilder {
+
+  private val SimpletonPattern = """
```
geoffjentry: I appreciate the effort but I'm thinking that I'd read this easier if it was just the regex w/ a comment explaining what was going on. YMMV and all that, if others disagree so be it

geoffjentry:
👍

core/src/main/scala/cromwell/core/simpleton/WdlValueBuilder.scala
mcovarr: I had the reverse comment on the regexp I copied this from.  😄 

Though TBH the first two are probably simple enough I would agree with that; I'm less sure about the third.

geoffjentry: I just found that I neither understood what the regex was nor what it did which wasn't such a good situation. Like I said though, I'm happy to be an outlier

core/src/main/scala/cromwell/core/simpleton/WdlValueBuilder.scala
```diff
@@ -0,0 +1,145 @@
...
+    *
+    * the corresponding `SimpletonComponent` would be
+    *
+    * {{{
+    * SimpletonComponent(":bar[0]", WdlString("baz"))
+    * }}}
+    */
+  private case class SimpletonComponent(path: String, value: WdlValue)
+
+  def build(taskOutputs: Traversable[TaskOutput], simpletons: Traversable[WdlValueSimpleton]): Map[String, WdlValue] = {
```
ruchim: @mcovarr Is there a reason we wouldn't want to convert the taskOutputs and WdlSimpleton values directly into JobOutputs right here? That's what you need for the JobStore as well, yes?

ruchim:
👍 well documented!

core/src/main/scala/cromwell/core/simpleton/WdlValueBuilder.scala
mcovarr: I believe you're right that both the call cache result reader and job store need `JobOutputs`, but I'm not sure I agree that knowledge of what these two callers will do with this data  should necessarily be baked into a `WdlValueSimpleton` -> `WdlValue` converter.  Also the code to do the `JobOutputs` conversion is pretty simple:

```scala
val jobOutputs = WdlValueBuilder.build(taskOutputs, simpletons) mapValues JobOutput.apply
```

ruchim: To me, it was the fact that this object exists for very specific use case of converting simpletons to JobOutputs and thus its should do that last bit, since it would be the remaining transformation. But if Simpleton -> WdlValue conversion could be reused elsewhere, then I see the use of leaving things generalized. I certainly don't feel strongly one way or another, just philosophizing.

core/src/main/scala/cromwell/core/simpleton/WdlValueBuilder.scala
```diff
@@ -0,0 +1,151 @@
...
+  private val MapElementPattern = """
+        (?x)                             # Turn on comments and whitespace insensitivity.
+
+        ^(                               # Begin capture.  This outer capture grabs all the stuff that will
+                                         # be ignored as the simpleton parsing "descends" into the map.
+
+          :                              # colon
+            (                            # Begin nested capture.
+              (?:                        #   Begin nested nested non-capturing group.
+                \\\[                     #     An escaped open square brace;
```
cjllanwarne: ToL: At this point, it's looking more like a lexing/parsing problem than a regex problem...

core/src/test/scala/cromwell/core/simpleton/WdlValueBuilderSpec.scala
```diff
@@ -0,0 +1,51 @@
...
+        WdlString("foo[1]") -> WdlString("foo"),
+        WdlString("bar[[") -> WdlString("bar"),
+        WdlString("baz:qux") -> WdlString("baz:qux")
+      ))
+    )
+
+    val taskOutputs = wdlValues map { case (k, wv) => TaskOutput(k, wv.wdlType, IgnoredExpression) }
+
+    import WdlValueSimpleton._
+    val actual = WdlValueBuilder.toWdlValues(taskOutputs, wdlValues.simplify)
```
cjllanwarne: no implicit `complexificate` method? :(

core/src/main/scala/cromwell/core/simpleton/WdlValueBuilder.scala
mcovarr: It's really not as complicated as it looks, but handling the escaped metacharacters leads to an explosion of backslashes that make the non-`(?x)` version of this unreadable.  😢

PR #1309, opened 2016-08-19
Simpletonize the JobStore.  Closes #1280

### LINKED ISSUE #1280: Job Store should use simpletons just like Call Cache
The Call Cache persists job results as `simpleton`s to prevent large arrays breaking database limits. The JobStore write out job results as a single giant `JSON` blob. It should use `simpleton`s too (preferably in the same way, and using the same code, as the Call Cache)

mcovarr:


geoffjentry:
It is a good thing @cjllanwarne didn't create the complecton

core/src/main/scala/cromwell/core/simpleton/WdlValueBuilder.scala
```diff
@@ -78,6 +78,7 @@ object WdlValueBuilder {
         )                                # End capture.
 
                                   """.trim.r          // The trim is necessary as (?x) must be at the beginning of the regex.
+
```
kshakir: Revert this line.

database/src/main/scala/cromwell/database/slick/tables/JobStoreComponent.scala
```diff
@@ -32,7 +30,7 @@ trait JobStoreComponent {
 
   protected val jobStore = TableQuery[JobStoreEntries]
 
-  val jobStoreAutoInc = jobStore returning jobStore.map(_.jobStoreTableId)
+  val jobStoreAutoInc = jobStore returning jobStore.map(_.jobStoreId) into ((a, id) => a.copy(jobStoreId = Option(id)))
```
kshakir: Get rid of the `into`, and just have `val jobStoreAutoInc = jobStore returning jobStore.map(_.jobStoreId)`.

In `addJobStoreEntry()`, you're only using the `JobStoreEntry.jobStoreId`, not the whole `JobStoreEntry`.

database/src/main/scala/cromwell/database/sql/tables/JobStoreResultSimpletonEntry.scala
```diff
@@ -0,0 +1,11 @@
+package cromwell.database.sql.tables
+
+case class JobStoreResultSimpletonEntry
+(
+  simpletonKey: String,
+  simpletonValue: String,
+  wdlType: String,
+  jobStoreEntryId: Int,
```
kshakir: The column is `JOB_STORE_ID` in the database, so the field name should match, aka `jobStoreId`.

engine/src/test/scala/cromwell/SimpleWorkflowActorSpec.scala
```diff
@@ -11,7 +11,7 @@ import cromwell.engine.workflow.WorkflowActor
 import cromwell.engine.workflow.WorkflowActor.{StartNewWorkflow, StartWorkflowCommand}
 import cromwell.services.metadata.MetadataService
 import MetadataService.PutMetadataAction
-import cromwell.jobstore.{JobStoreActor, WriteCountingJobStoreDatabase}
+import cromwell.jobstore.{JobStoreActor, WriteCountingJobStore$}
```
kshakir: Haven't checked out the code and tested myself-- but is the `$` necessary for the import? Also appears in `SingleWorkflowRunnerActorSpec`.

kshakir:
👍  once comments above addressed.

core/src/main/scala/cromwell/core/simpleton/WdlValueBuilder.scala
mcovarr: Ok, but all this is changed by my other PR

kshakir: Even more reason to revert this one whitespace line change. :) No git merge conflict.

engine/src/test/scala/cromwell/SimpleWorkflowActorSpec.scala
mcovarr: I have no idea what happened here.  That's now an unnecessary import in both spots anyway.  (shrug)

core/src/main/scala/cromwell/core/simpleton/WdlValueBuilder.scala
mcovarr: That is true

database/src/main/scala/cromwell/database/slick/JobStoreSlickDatabase.scala
```diff
@@ -10,23 +10,36 @@ trait JobStoreSlickDatabase extends JobStoreSqlDatabase {
 
   import dataAccess.driver.api._
 
-  override def addJobStoreEntries(entries: Iterable[JobStoreEntry])(implicit ec: ExecutionContext): Future[Unit] = {
-    val action = dataAccess.jobStoreAutoInc ++= entries
+  private def addJobStoreEntry(jobStoreEntry: JobStoreEntry, jobStoreSimpletonEntries: Iterable[JobStoreResultSimpletonEntry])
+                              (implicit ec: ExecutionContext): DBIO[Unit] = {
+    for {
+      insertJobStoreEntry <- dataAccess.jobStoreAutoInc += jobStoreEntry
+      entries = jobStoreSimpletonEntries map { _.copy(jobStoreId = insertJobStoreEntry.self) }
```
mcovarr: @kshakir does this look right to you?  This Slick `SingleInsertResult` stuff is all a bit magical to me.  😦

cjllanwarne:
@geoffjentry Job Store Anti-dis-ensimplificationalism?

database/src/main/scala/cromwell/database/slick/JobStoreSlickDatabase.scala
kshakir: Heh. I feel the same way about scalaz, and I think it's because slick is using monads here too. 

In this case, `SingleInsertResult` is just the auto incremented `jobStoreId`. I also have no clue what/where `.self` is.

This would probably read better as:

```scala
insertedJobStoreId <- dataAccess.jobStoreAutoInc += jobStoreEntry
entries = jobStoreSimpletonEntries map { _.copy(jobStoreId = insertedJobStoreId) }
```

mcovarr: OK I think this may be a dependent typing thing; `Int` appears to be the type of `RU` in `ReturningInsertActionComposer[U, RU]` after the `returning jobStore.map(_.jobStoreId)` business.  

I have no idea how I went down that `self` path, that appears to be impliciting the `Int` into a `RichInt` and then pulling the raw `Int` back out.  I think I'll just go with your version instead.  😄

engine/src/main/scala/cromwell/jobstore/SqlJobStore.scala
```diff
@@ -1,61 +1,79 @@
...
+          key.workflowId.toString,
+          key.callFqn,
+          key.index.fromIndex,
+          key.attempt,
+          jobSuccessful = true,
+          returnCode,
+          None,
+          None)
+        val simpletons = jobOutputs.mapValues(_.wdlValue).simplify.map {
+          s => JobStoreResultSimpletonEntry(s.simpletonKey, s.simpletonValue.valueString, s.simpletonValue.wdlType.toWdlString, -1) }
```
kshakir: ToL: Creating the database row here with a `-1` for the foreign key is very un-scala-esque. Almost like passing in a `null` to be filled in later. For primary keys, we've used `Option[_]` to represent the auto-incremented keys populated by the database. For most fk's though I've jumped through hoops to pass in a function-that-takes-the-fk-and-returns-the-new-rows. Using `-1` is much, much simpler though than even using an `Option` for the fk, and definitely simpler than passing in a function parameter. I'm torn on what's most "correct" though.

mcovarr: Uh whoops, I meant this as only a typechecking placeholder but forgot to circle back to it.  😦 

Though now that you point this out, I'm not sure what the best thing to do here is either.  I could `Option` the FK, but as the FK isn't actually optional I'm not too thrilled with that idea.  Not sure I grok your function idea, can you point to an example of that?

mcovarr: Oh nm I get it.  Let me see if I can fix this up.

Horneth:
👍

PR #1310, opened 2016-08-20
Tighten up regexps

mcovarr:
- Removed one unnecessary capture each from the array and map regexps
- Factored out the backslash on escaped metacharacters in the map regexp
- Transformed the map regexp to standard, non-"free spacing" form
- Added some docs on the `WdlValueSimpleton` encoding protocol

### DIFF
--- a/core/src/main/scala/cromwell/core/simpleton/WdlValueBuilder.scala
+++ b/core/src/main/scala/cromwell/core/simpleton/WdlValueBuilder.scala
@@ -9,7 +9,9 @@ import cromwell.core.{JobOutput, JobOutputs}
 import cromwell.core.simpleton.WdlValueSimpleton._
 
 
-/** Builds arbitrary WdlValues from WdlValueSimpletons. */
+/**
+  * Builds arbitrary `WdlValues` from `WdlValueSimpletons`.
+  **/
 object WdlValueBuilder {
 
   /**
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cromwell/core/simpleton/WdlValueBuilder.scala
```diff
@@ -31,60 +33,35 @@ object WdlValueBuilder {
...
     * <ol>
-    * <li>A colon followed by one or more non-metacharacters, including `WdlValueSimpleton#escapeMeta` escaped metacharacters.</li>
-    * <li>Same as above except for the initial colon</li>
+    * <li>The map key, possibly including `WdlValueSimpleton#escapeMeta` escaped metacharacters.  Does <b>not</b> include
+    *     the leading colon.</li>
     * <li>Possibly a metacharacter and more trailing stuff</li>
     * </ol>
     */
-  private val MapElementPattern = """
-        (?x)                             # Turn on comments and whitespace insensitivity.
```
geoffjentry: I'm going to miss the ee cummings authored code

geoffjentry:
👍 

A+++++ would regex again

cjllanwarne:
Could you add a test case for this fix?

Otherwise 👍

mcovarr:
@cjllanwarne not sure what you mean?  This isn't really a fix so much as a simplification, and there are already tests in `WdlValueBuilderSpec` which thankfully continue to pass.  😄

PR #1334, opened 2016-08-24
remove low hanging fruit from cromwelltestkitspec

geoffjentry:
This is all functionality now being tested by Centaur


geoffjentry:
I'm trying to figure out how to get it to redo the travis build with the local travis check as well. Opening and closing the PR didn't help. I'll wait and see if there's any code review changes which require a new commit, otherwise i'll add something and then remove it

kshakir:
> I'm trying to figure out how to get [travis?] to redo the travis build

Yeah, PRs 1333 and this 1334 currently have the same git hash. Try updating to a new git hash by "touching" the commit, and then force pushing:

```bash
git checkout jg_haircut_for_testkitspec && \
git commit --amend -C HEAD --date=now && \
git push -f origin HEAD
```

NOTE: Add `-q` after each `git <command>` to quiet down the output.

geoffjentry:
Yeah, I was somewhat hoping for code review comment so I could do it au natural

kshakir:
Request denied. 👍

geoffjentry:
Fair enough, although since I didn't realize you could 'touch' like that and this seems likely to be uncontroversial I'll do that now and save myself 30 mins tomorrow :)

kshakir:
Upon closer inspection, I think Travis is doing the right thing by not currently running `centaurLocal` for the "push", but only running it for the "pr".

The `.travis.yml` in the current "push"ed commit doesn't specify to run `centaurLocal`. Meanwhile the develop branch has already been updated, meaning that if-this-"pr"-were-merged `centaurLocal` will run.

Rebase onto develop, picking up the `.travis.yml` changes, and the "push" build should also run `centaurLocal`, probably?

core/src/test/scala/cromwell/util/SampleWdl.scala
```diff
@@ -425,80 +381,6 @@ object SampleWdl {
       """.stripMargin
     override lazy val rawInputs = Map.empty[String, String]
   }
-
-  object NestedScatterWdl extends SampleWdl {
```
ruchim: I should know and already forgot--nested scatters are definitely tested in Centaur?

mcovarr: nested scatters are not currently supported

mcovarr: Scott did do some work in that direction in his wdl4s branch

core/src/test/scala/cromwell/util/SampleWdl.scala
```diff
@@ -1813,512 +1273,6 @@ object SampleWdl {
...
-        |
-        |workflow googly_workflow {
-        |  call googly_task
-        |}
-      """.stripMargin
-
-    val rawInputs = Map.empty[String, Any]
-  }
-
-  object WorkflowWithMessedUpMemory extends SampleWdl {
```
ruchim: I dont think we have all these cases of MessedUpMemory in Centaur---and neither do I feel there's an emergency for it. Especially since its the error's that would be valuable, and that's going to be a loooot of manual work.

core/src/test/scala/cromwell/util/SampleWdl.scala
```diff
@@ -1813,512 +1273,6 @@ object SampleWdl {
...
-        |
-        |workflow w {call x}
-      """.stripMargin
-
-    val rawInputs = Map(
-      "w.x.a" -> 5
-    )
-  }
-
-  object WorkflowWithFourPreemptibleRetries extends SampleWdl {
```
ruchim: TOL: I don't think we do preemptible testing at all in Centaur-- for very understandable reasons. Kind of along the same lines as testing restarts in Centaur, it will require extra magic.

ruchim:
~~spring~~ fall cleaning

The way I reviewed this is use the test names to identify the wdl/cromwell behavior being tested, so I'm not really looking super deep into what was removed (with the understanding that Travis will catch all that). I had some general comments but none of them require immediate action. 👍

core/src/test/scala/cromwell/util/SampleWdl.scala
geoffjentry: Some of these wdls weren't being used anywhere, so these cases you're pointing out likely fall under that umbrella. I'll double check

PR #1339, opened 2016-08-25
Cromwell 1087

jainh:
PR Documentation describing the SparkBackend added to Cromwell

This PR adds support for execution of spark jobs as task in a workflow using the existing wdl format, with restrictions on the environment like having a local file system, a shared file system or a network file system when running a spark job in the spark standalone cluster mode.

This implementation takes a wdl with the backend configuration specified as "Spark" and then generates the appropriate spark commands and monitoring process to ensure the job runs to completion. Meaning, details of the spark internals are completely abstracted from the user provided backends with different configurations containing different flavours of { master and deployMode } combinations are already set.

Internally, we create a bash script containing a spark-submit (depending on the backend flavour selected at runtime) command using all the specified wdl runtime attributes which is then executed by Spark.  

Current deploy modes supported for any spark job:
  a - Client deploy mode using the spark standalone cluster manager
  b - Cluster deploy mode using the spark standalone cluster manager
  c - Client deploy mode using Yarn resource manager
  d - Cluster deploy mode using Yarn resource manager
  
Future PR Plans:
  In this PR, the hadoop file system cannot be used as an input/output for the SBE because the Cromwell engine does not identify the protocol, and this results in the hdfs path being localized (soft-link, hard-link or copied).
  This is not a problem until the SBE tries to evaluate the output after a successful execution, and because it cannot interpret the protocol, it tries to look for an hdfs output locally which results in an error. Note: This is only the case when the spark job writes the output to an hdfs location. Then cromwell cannot find the output file for evaluation.

  In the near **Future**, we plan to provide an hdfs client similar to that of the gcs to add support for the hdfs, primarily because hdfs is spark's natural file system.
  Note that this doesn't actually prevent spark from writing to the hdfs, in order words, the spark application can write or read from the hdfs if given hdfs locations as arguments.

Reason for restriction on environment:
  In spark cluster mode, the assembly jar file containing the application has to exist in all the nodes of the cluster since the driver program can be started on any of the nodes in the cluster.
  Known solution to this is to put the jar file in a shared file system like hdfs or a network file system, or a parallel distributed file system like lustre where all nodes in the cluster can access the file.
  For this reason, we did all our testing using the lustre file system, though it works just fine on a local file system with replication.
  
  Also, note that as pointed out in the future PR plans section, the hadoop file system is not supported on this release.

Supported File Systems:
  Local File System
  Network File System
  Distributed file system

PS: Please find attached read me for examples on How to use Spark Backend
[readMe.md.zip](https://github.com/broadinstitute/cromwell/files/437890/readMe.md.zip)

contributor: @iyanuobidele

Reviewers:  @geoffjentry  @francares


project/Dependencies.scala
```diff
@@ -88,6 +88,8 @@ object Dependencies {
     "org.mongodb" %% "casbah" % "3.0.0"
   )
 
+  val sparkBackendDependencies = List("org.json4s" %% "json4s-jackson" % "3.4.0")
```
geoffjentry: Why not use the spray json support?

jainh: 1) It is easier to use than Spray Json
2) Some boilerplate code needs to be included if change to Spray Json
3) Doesn't work in a straight forward way if case class has Date or some non-primitive data types.

However here we are using primitive data types, it would not matter.

geoffjentry: I don't think "it's easier to use" use is a good argument for doing something completely differently from the rest of the code base

jainh: I can use Spray Json not a problem but Json4s is faster.

geoffjentry: There are a lot of reasons why I might want to use a diff library here and there. The most important reason is "that's what's being used everywhere else". 

Aside: If we were to switch our json handling libraries across the board I'd want it to be circe, which hasn't even been part of this discussion yet.

cjllanwarne:
I haven't looked at the code yet but based on the description, Very Cool! (tm)

Just to check, you're saying we **can** use HDFS inside jobs, but it's just the functions don't exist yet to (e.g.) read_string from an HDFS file?

jainh:
@cjllanwarne Yes read_string from hdfs, It will be same as in Jes backend.

build.sbt
```diff
@@ -78,11 +84,13 @@ lazy val root = (project in file("."))
...
   .aggregate(jesBackend)
   .aggregate(engine)
   // Next level of projects to include in the fat jar (their dependsOn will be transitively included)
   .dependsOn(engine)
   .dependsOn(jesBackend)
   .dependsOn(htCondorBackend)
+  .dependsOn(sparkBackend)
   // Dependencies for tests
-  .dependsOn(engine % "test->test")
+  .dependsOn(engine % "test->test")
```
geoffjentry: \n

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkClusterProcess.scala
```diff
@@ -0,0 +1,173 @@
...
+  val SparkRestApiPort = "6066"
+  val SparkClusterIdentifier = Seq("6066", "spark")
+  val SparkClusterDeployMode = "cluster"
+  val PossibleSparkClusterMaster = Seq("spark", "Spark")
+  val SubmitJobJson = "%s.json"
+  val ReturnCodeFile = "rc"
+  val RunningStatus = "RUNNING"
+  val FinishedStatus = "FINISHED"
+  val FailedStatus = "FAILED"
+  val FAILED = -1
```
geoffjentry: Failed

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkClusterProcess.scala
```diff
@@ -0,0 +1,173 @@
...
+  val FINISHED = 0
```
geoffjentry: Finished

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkClusterProcess.scala
```diff
@@ -0,0 +1,173 @@
...
+  val sleepTime = 1000
+  val submissionIdInJson = "submissionId"
+  val successInJson = "success"
+  val submissionsApiLink = "v1/submissions/status"
+  lazy val SparkClusterMasterHostName = Try(sys.env("HOSTNAME")) match {
+    case Success(s) => Some(s)
```
geoffjentry: Option(s)

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkClusterProcess.scala
```diff
@@ -0,0 +1,173 @@
...
+      case Failure(exception: Throwable) =>
+        logger.error(s"{} Spark Job failed to submit successfully following reason: {}", tag, exception.getMessage)
+        completionPromise success Failed(exception)
+    }
+    completionPromise.future
+  }
+
+  override def monitorSparkClusterJob(subId: String,  rcPath: Path, promise: Promise[Unit]): Unit = {
+    pollForJobStatus(subId) onComplete {
+      case Success(resp: SuccessResponse) if resp.driverState == RunningStatus => Thread.sleep(sleepTime)
```
geoffjentry: Who owns the thread which is being blocked?

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkClusterProcess.scala
```diff
@@ -0,0 +1,173 @@
...
+  case object Finished extends TerminalStatus
+
+  object SparkClusterJsonProtocol extends DefaultJsonProtocol {
+    implicit val sparkStatusResponseFormat = jsonFormat7(SuccessResponse)
+    implicit val successfulParsedResponseFormat = jsonFormat5(SuccessfulParsedResponse)
+  }
+}
+
+object SparkClusterConstants {
+  lazy val SparkClusterMasterDefault = "spark-master"
```
geoffjentry: Are all of these constants being used in multiple places? If anything is being used in only a single place, de-constantize

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkClusterProcess.scala
```diff
@@ -0,0 +1,173 @@
...
+        rcPath.contentAsString.stripLineEnd.toInt match {
+          case FINISHED => completionPromise success Finished
+          case FAILED => completionPromise success Failed(new IllegalStateException("Spark Driver returned failed status"))
+        }
+      case Failure(err) => completionPromise success Failed(err)
+    }
+  }
+
+  override def completeMonitoringProcess(rcPath: Path, status: String, promise: Promise[Unit]) = {
+    rcPath < status
```
geoffjentry: use `write` not `<`

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkInitializationActor.scala
```diff
@@ -0,0 +1,72 @@
...
+
+
+  /**
+    * A call which happens before anything else runs
+    */
+  override def validate(): Future[Unit] = {
+    Future {
+      calls foreach { call =>
+        val runtimeAttributes = call.task.runtimeAttributes.attrs
+        val notSupportedAttributes = runtimeAttributes filterKeys {
```
geoffjentry: one line

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkJobExecutionActor.scala
```diff
@@ -0,0 +1,207 @@
...
+import wdl4s.parser.MemoryUnit
+import wdl4s.util.TryUtil
+
+import scala.concurrent.{Future, Promise}
+import scala.sys.process.ProcessLogger
+import scala.util.{Failure, Success, Try}
+import scala.language.postfixOps
+
+object SparkJobExecutionActor {
+  val fileSystems = List(FileSystems.getDefault)
```
geoffjentry: FileSystems

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkJobExecutionActor.scala
```diff
@@ -0,0 +1,207 @@
...
+
+  private def executeTask(process: SparkProcess, stdoutWriter: UntailedWriter, stderrWriter: TailedWriter): Future[BackendJobExecutionResponse] = {
+    val submitResult = submitSparkScript(process, ProcessLogger(stdoutWriter.writeWithNewline, stderrWriter.writeWithNewline))
+    List(stdoutWriter.writer, stderrWriter.writer).foreach(_.flushAndClose())
+
+    resolveExecutionResult(submitResult, runtimeAttributes.failOnStderr)
+  }
+
+  private def submitSparkScript(sparkProcess: SparkProcess, processLogger: ProcessLogger): Try[Int] = {
+    val argv = Seq(SparkCommands.SHELL_CMD, scriptPath.toString)
```
geoffjentry: Give Seq a concrete type. In this case since you're effectively aking a List, make a List

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkJobExecutionActor.scala
```diff
@@ -0,0 +1,207 @@
...
+        SparkCommands.MASTER -> sparkMaster,
+        SparkCommands.EXECUTOR_CORES -> runtimeAttributes.executorCores,
+        SparkCommands.EXECUTOR_MEMORY -> runtimeAttributes.executorMemory.to(MemoryUnit.GB).amount.toLong,
+        SparkCommands.SPARK_APP_WITH_ARGS -> command.get,
+        SparkCommands.DEPLOY_MODE -> sparkDeployMode
+      )
+
+      val sparkSubmitCmd = cmds.sparkSubmitCommand(attributes)
+      val isSparkClusterMode = checkForSparkClusterMode(sparkDeployMode, sparkMaster)
+      val sparkCommand = if(isSparkClusterMode) {
```
geoffjentry: if (

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkJobExecutionActor.scala
```diff
@@ -0,0 +1,207 @@
...
+      scriptPath.addPermission(PosixFilePermission.OWNER_EXECUTE)
+
+    } catch {
+      case ex: Exception =>
+        log.error(ex, "Failed to prepare task: " + ex.getMessage)
+        executionResponse success FailedNonRetryableResponse(jobDescriptor.key, ex, None)
+    }
+  }
+
+  private def checkForSparkClusterMode(deployMode: String, master: String): Boolean = {
```
geoffjentry: function name should start with `is` or something like that if it returns a Boolean

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkJobExecutionActor.scala
```diff
@@ -0,0 +1,207 @@
...
+    SparkClusterIdentifier.exists(master.toLowerCase.contains) && deployMode.equals(SparkClusterDeployMode)
+  }
+
+  private def taskLauncher = {
+    Try {
+      checkForSparkClusterMode(sparkDeployMode, sparkMaster) match {
+        case true => executionResponse completeWith  executeTask(clusterExtProcess, clusterStdoutWriter, clusterStderrWriter)
+        case _ => executionResponse completeWith  executeTask(extProcess, stdoutWriter, stderrWriter)
```
geoffjentry: why would this not be `false`?

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkJobExecutionActor.scala
```diff
@@ -0,0 +1,207 @@
...
+      }
+    } recover {
+      case exception => executionResponse success FailedNonRetryableResponse(jobDescriptor.key, exception, None)
+    }
+  }
+
+}
```
geoffjentry: \n

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkProcess.scala
```diff
@@ -0,0 +1,93 @@
...
+import cromwell.core.{TailedWriter, UntailedWriter}
+import cromwell.core.PathFactory.EnhancedPath
+
+import scala.sys.process._
+import better.files._
+import scala.language.postfixOps
+import scala.util.{Failure, Success, Try}
+
+object SparkCommands {
+  val SPARK_SUBMIT = "spark-submit"
```
geoffjentry: Conform to standard naming conventions.

Same comment as above. If these are used in a single place, de-constantify them. This isn't enterprise java :)

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkProcess.scala
```diff
@@ -0,0 +1,93 @@
...
+    filePath.write(
+      s"""#!/bin/sh
+          |cd $containerRoot
+          |$instantiatedCommand
+          |echo $$? > rc
+          |""".stripMargin)
+  }
+
+  def sparkSubmitCommand(attributes: Map[String, Any]): String = {
+    val stringBuilder = StringBuilder.newBuilder
```
geoffjentry: Why do you need a mutable construct here?

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkProcess.scala
```diff
@@ -0,0 +1,93 @@
...
+    val commandPlaceHolder = " %s%s %s "
+    val sparkHome = Try(sys.env("SPARK_HOME")) match {
+      case Success(s) => Some(s)
```
geoffjentry: Option(s)

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkProcess.scala
```diff
@@ -0,0 +1,93 @@
...
+      case Failure(ex) => logger.warn(s"Spark home does not exist picking up default command")
```
geoffjentry: put logger on the next line

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkProcess.scala
```diff
@@ -0,0 +1,93 @@
...
+        None
+    }
+
+    val sparkSubmit = sparkHome
+      .map(p => String.format("%s%s", p, s"/bin/$SPARK_SUBMIT"))
+      .getOrElse(SPARK_SUBMIT)
+
+    attributes.foreach {
+      case (key, _) => attributes.get(key) match {
+        case Some (v) => commandPlaceHolder.format (SEPARATOR, key, v)
```
geoffjentry: either use `.format(` or ` format { `. Don't take advantage of the ability to do a bit of both. The latter is preferred by the scala style guide where possible

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkProcess.scala
```diff
@@ -0,0 +1,93 @@
...
+      .append(commandPlaceHolder.format(SEPARATOR, EXECUTOR_CORES, attributes(EXECUTOR_CORES)))
+      .append(commandPlaceHolder.format(SEPARATOR, EXECUTOR_MEMORY, "%s%s".format(attributes(EXECUTOR_MEMORY), MEMORY_UNIT)))
+      .append(attributes(SPARK_APP_WITH_ARGS)).toString
+    logger.debug(s"spark command is : $sparkCmd")
+    sparkCmd
+  }
+}
+
+trait SparkProcess {
+  private val stdout = new StringBuilder
```
geoffjentry: should be def in a trait. 

Why are you using mutable constructs?

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkProcess.scala
```diff
@@ -0,0 +1,93 @@
...
+  def processStderr: String = stderr.toString().trim
+
+  def commandList(command: String): Seq[String] = Seq("/bin/bash", command)
+
+  def untailedWriter(path: Path): UntailedWriter = path.untailed
+
+  def tailedWriter(limit: Int, path: Path): TailedWriter = path.tailed(limit)
+
+  def externalProcess(cmdList: Seq[String], processLogger: ProcessLogger = processLogger): Process = cmdList.run(processLogger)
+}
```
geoffjentry: \n

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkProcess.scala
```diff
@@ -0,0 +1,93 @@
...
+  private val stdout = new StringBuilder
+  private val stderr = new StringBuilder
+
+  def processLogger: ProcessLogger = ProcessLogger(stdout append _, stderr append _)
+
+  def processStdout: String = stdout.toString().trim
+
+  def processStderr: String = stderr.toString().trim
+
+  def commandList(command: String): Seq[String] = Seq("/bin/bash", command)
```
geoffjentry: `List(`

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkRuntimeAttributes.scala
```diff
@@ -0,0 +1,82 @@
...
+        override def exceptionContext: String = "Runtime attribute validation failed"
+
+        override def errorMessages: Traversable[String] = nel.list
+      }
+    }
+  }
+
+  private def validateNumberOfExecutors(numOfExecutors: Option[WdlValue], onMissingKey: => ErrorOr[Option[Int]]): ErrorOr[Option[Int]] = {
+    numOfExecutors match {
+      case Some(i: WdlInteger) => Some(i.value.intValue()).successNel
```
geoffjentry: Option > Some

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkClusterProcess.scala
jainh: Thread is owned by the execution actor but since we are using promise in the execution and returning future[backendresponse] from execute, we will complete that promise with this future once job is finished or failed.

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkProcess.scala
jainh: We can change it to String.Format..

core/src/main/resources/application.conf
```diff
@@ -194,7 +194,39 @@ backend {
...
+    //       //
+    //       // NOTE: soft-link will be skipped for Docker jobs
+    //       localization: [
+    //         "hard-link", "soft-link", "copy"
+    //       ]
+    //     }
+    //     gcs {
+    //       // References an auth scheme defined in the `google` stanza.
+    //       auth = "application-default"
+    //     }
```
mcovarr: Does gcs really make sense here?

project/Settings.scala
```diff
@@ -78,7 +78,22 @@ object Settings {
...
+  ) ++ commonSettings
+
+  val sparkBackendSettings = List(
+    name := "cromwell-spark-backend",
+    libraryDependencies ++= sparkBackendDependencies,
+    version := "0.1",
+    assemblyJarName in assembly := name.value + "-" + version.value + ".jar"
+  ) ++ commonSettings
+
+  val sgeBackendSettings = List(
```
mcovarr: Why does it look like sge was added here?

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkClusterProcess.scala
```diff
@@ -0,0 +1,173 @@
...
+import spray.httpx.unmarshalling._
+
+import scala.util.{Failure, Success, Try}
+
+object SparkClusterProcess {
+  sealed trait SparkClusterResponse
+  case class SuccessResponse(action: String, driverState: String, serverSparkVersion: String,
+                             submissionId: String, success: Boolean, workerHostPort: String, workerId: String) extends SparkClusterResponse
+  sealed trait ParserResponse
+  case class SuccessfulParsedResponse(action: String, message: String, serverSparkVersion: String, submissionId: String, success: Boolean) extends ParserResponse
```
mcovarr: Failure responses aren't possible?

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkClusterProcess.scala
```diff
@@ -0,0 +1,173 @@
...
+  val FINISHED = 0
+  val sleepTime = 1000
+  val submissionIdInJson = "submissionId"
+  val successInJson = "success"
+  val submissionsApiLink = "v1/submissions/status"
+  lazy val SparkClusterMasterHostName = Try(sys.env("HOSTNAME")) match {
+    case Success(s) => Some(s)
+    case Failure(_) => None
+  }
+  val sparkClusterSubmissionResponseRegex = """(?s)\{(.*?)}""".r.unanchored
```
mcovarr: What is the pattern this is trying to match?

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkClusterProcess.scala
```diff
@@ -0,0 +1,173 @@
...
+  lazy val SparkClusterMasterRestUri = SparkClusterMasterHostName
+    .map(p => String.format("%s%s%s", HttpProtocol, p, s":$SparkRestApiPort"))
```
mcovarr: Scala prefers `"%s%s%s".format(...)`

project/Settings.scala
jainh: I need to check on this but i didn't add that.

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkClusterProcess.scala
jainh: Failure response is possible from the parse function but since it is used inside a future container we got that exception and hence onFailure of that future.

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkClusterProcess.scala
jainh: It is trying to match something like  this
```
   {
      "action" : "CreateSubmissionResponse",
       "message" : "Driver successfully submitted as driver-20160806183527-0006",
       "serverSparkVersion" : "1.6.1",
       "submissionId" : "driver-20160806183527-0006",
       "success" : true
      }
```

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkClusterProcess.scala
jainh: ok

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkClusterProcess.scala
mcovarr: It's a bit strange to have what looks like an algebraic data type defined by a sealed trait, but which can be instantiated by only a single "successful" variant.

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkClusterProcess.scala
mcovarr: I think the question mark in `.*?` is superfluous then; the `.*` will already match zero or more characters.

It's also possible that the capturing group should encompass the curly braces if you want to capture those too.

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkJobExecutionActor.scala
```diff
@@ -0,0 +1,207 @@
...
+  }
+
+  /**
+    * Abort a running job.
+    */
+  override def abort(): Unit = Future.failed(new UnsupportedOperationException("SparkBackend currently doesn't support aborting jobs."))
+
+  /**
+    *
+    */
```
mcovarr: vacuous comment

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkProcess.scala
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    val sparkSubmit = sparkHome
+      .map(p => String.format("%s%s", p, s"/bin/$SPARK_SUBMIT"))
+      .getOrElse(SPARK_SUBMIT)
+
+    attributes.foreach {
+      case (key, _) => attributes.get(key) match {
+        case Some (v) => commandPlaceHolder.format (SEPARATOR, key, v)
+        case None => logger.warn (s" key: $key doesn't exit spark should pick it")
```
mcovarr: exit => exist

supportedBackends/spark/src/test/scala/cromwell/backend/impl/spark/SparkJobExecutionActorSpec.scala
```diff
@@ -0,0 +1,560 @@
...
+  "executeCluster method" should {
+    "return finished status when keyword \"spark\" is used as master attribute along with \"cluster\" for deployMode" in {
+      val jobDescriptor = prepareJob(helloWorldClusterWdl, passOnStderr, isCluster = true)
+      val (job, jobPaths, backendConfigDesc) = (jobDescriptor.jobDescriptor, jobDescriptor.jobPaths, jobDescriptor.backendConfigurationDescriptor)
+      val stubProcess = mock[Process]
+      val stubUntailed = new UntailedWriter(jobPaths.stdout) with MockPathWriter
+      val stubTailed = new TailedWriter(jobPaths.stderr, 100) with MockPathWriter
+      val subId = "driver-20160803181054-0000"
+      val status = true
+      val stderr = ""
```
mcovarr: this test is filled with unused variables

supportedBackends/spark/src/test/scala/cromwell/backend/impl/spark/SparkRuntimeAttributesSpec.scala
```diff
@@ -0,0 +1,131 @@
...
+import cromwell.core.{WorkflowId, WorkflowOptions}
+import org.scalatest.{Matchers, WordSpecLike}
+import spray.json.{JsBoolean, JsNumber, JsObject, JsString, JsValue}
+import wdl4s.WdlExpression._
+import wdl4s.expression.NoFunctions
+import wdl4s.util.TryUtil
+import wdl4s.{Call, WdlExpression, _}
+import wdl4s.values.WdlValue
+
+class SparkRuntimeAttributesSpec extends WordSpecLike with Matchers {
```
mcovarr: lots of unused variables here too

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkBackendFactory.scala
```diff
@@ -0,0 +1,32 @@
...
+import wdl4s.expression.WdlStandardLibraryFunctions
+
+case class SparkBackendFactory(configurationDescriptor: BackendConfigurationDescriptor) extends BackendLifecycleActorFactory {
+  override def workflowInitializationActorProps(workflowDescriptor: BackendWorkflowDescriptor, calls: Seq[Call], serviceRegistryActor: ActorRef): Option[Props] = {
+    Option(SparkInitializationActor.props(workflowDescriptor, calls, configurationDescriptor, serviceRegistryActor))
+  }
+
+  override def jobExecutionActorProps(jobDescriptor: BackendJobDescriptor, initializationData: Option[BackendInitializationData],
+                                      serviceRegistryActor: ActorRef): Props = {
+    SparkJobExecutionActor.props(jobDescriptor, configurationDescriptor)
```
mcovarr: Interesting that the `serviceRegistryActor` isn't passed to the execution actor, which means no metadata is being published during execution.

jainh: Yes, right now it is not being used, but we will add it later.

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkJobExecutionActor.scala
```diff
@@ -0,0 +1,207 @@
...
+    val process = sparkProcess.externalProcess(argv, processLogger)
+    val jobReturnCode = Try(process.exitValue()) // blocks until process (i.e. spark submission) finishes
+    log.debug("{} Return code of spark submit command: {}", tag, jobReturnCode)
+    jobReturnCode
+  }
+
+  private def resolveExecutionResult(jobReturnCode: Try[Int], failedOnStderr: Boolean): Future[BackendJobExecutionResponse] = {
+    (jobReturnCode, failedOnStderr)  match {
+      case (Success(0), false) if jobPaths.stderr.lines.toList.isEmpty => resolveExecutionProcess
+      case (Success(0), false) => resolveExecutionProcess
```
mcovarr: line 110 seems unnecessary if it's going to produce the same response as 111

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkProcess.scala
cjllanwarne: Hmm, I don't think this is a functional argument? I think it's just a function call with an accidental space before the `(` - So, `.format(...)`

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkClusterProcess.scala
cjllanwarne: @mcovarr is that preferred over the `s"...${...}..."` notation?

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkInitializationActor.scala
```diff
@@ -0,0 +1,72 @@
...
+                               override val calls: Seq[Call],
+                               override val configurationDescriptor: BackendConfigurationDescriptor,
+                               override val serviceRegistryActor: ActorRef) extends BackendWorkflowInitializationActor {
+
+  override protected def runtimeAttributeValidators: Map[String, (Option[WdlExpression]) => Boolean] = Map(
+    FailOnStderrKey -> wdlTypePredicate(valueRequired = false, WdlBooleanType.isCoerceableFrom),
+    SparkRuntimeAttributes.AppMainClassKey -> wdlTypePredicate(valueRequired = true, WdlBooleanType.isCoerceableFrom),
+    SparkRuntimeAttributes.NumberOfExecutorsKey -> wdlTypePredicate(valueRequired = false, WdlBooleanType.isCoerceableFrom),
+    SparkRuntimeAttributes.ExecutorMemoryKey -> wdlTypePredicate(valueRequired = false, WdlBooleanType.isCoerceableFrom),
+    SparkRuntimeAttributes.ExecutorCoresKey -> wdlTypePredicate(valueRequired = false, WdlBooleanType.isCoerceableFrom)
```
mcovarr: it seems unlikely any of these except the first are actually supposed to be booleans

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkBackendFactory.scala
```diff
@@ -0,0 +1,32 @@
...
+
+import akka.actor.{ActorRef, Props}
+import cromwell.backend._
+import cromwell.backend.sfs.SharedFileSystemExpressionFunctions
+import cromwell.backend.io.JobPaths
+import cromwell.core.CallContext
+import wdl4s.Call
+import wdl4s.expression.WdlStandardLibraryFunctions
+
+case class SparkBackendFactory(configurationDescriptor: BackendConfigurationDescriptor) extends BackendLifecycleActorFactory {
```
cjllanwarne: Just a heads up, when you rebase onto the latest develop, this constructor is also passed an extra `actorSystem: ActorSystem` argument. Of course, you can just do nothing with it if you don't need it

supportedBackends/spark/src/test/scala/cromwell/backend/impl/spark/SparkJobExecutionActorSpec.scala
```diff
@@ -0,0 +1,560 @@
...
+  private val timeout = Timeout(20.seconds)
+
+  after {
+    Mockito.reset(sparkProcess)
+    Mockito.reset(sparkClusterProcess)
+  }
+
+  before {
+    when(sparkClusterProcess.logger).thenReturn(Logger(LoggerFactory.getLogger(getClass.getName)))
+    when(sparkClusterProcess.completeMonitoringProcess(any[Path], any[String], any[Promise[Unit]])).thenCallRealMethod()
```
mcovarr: This is a lot of partial mocking.  If you click through to the docs for `thenCallRealMethod` you'll see the authors' warnings on this pattern, culminating in:

>...I wouldn't use partial mocks for new, test-driven & well-designed code.

jainh: Mocking is powerful tool problem is if we mock particular class or trait we need to use same mocked instance to call actual behavior or mock behavior. So, I think to me it looks fine.

supportedBackends/spark/src/test/scala/cromwell/backend/impl/spark/SparkRuntimeAttributesSpec.scala
```diff
@@ -0,0 +1,131 @@
...
+      val workflowOptions = workflowOptionsWithDefaultRA(Map(FailOnStderrKey -> JsBoolean(true)))
+      assertSparkRuntimeAttributesSuccessfulCreation(runtimeAttributes, workflowOptions, staticDefaults.copy(failOnStderr = true))
+    }
+
+  }
+
+  private def buildWorkflowDescriptor(wdl: WdlSource,
+                                      inputs: Map[String, WdlValue] = Map.empty,
+                                      options: WorkflowOptions = WorkflowOptions(JsObject(Map.empty[String, JsValue])),
+                                      runtime: String = "") = {
```
mcovarr: wouldn't this throw an `IllegalArgumentException` if the default value of `runtime` kicked in?

supportedBackends/spark/src/test/scala/cromwell/backend/impl/spark/SparkRuntimeAttributesSpec.scala
```diff
@@ -0,0 +1,131 @@
...
+    new BackendWorkflowDescriptor(
+      WorkflowId.randomId(),
+      NamespaceWithWorkflow.load(wdl.replaceAll("RUNTIME", runtime.format("appMainClass", "com.test.spark"))),
+      inputs,
+      options
+    )
+  }
+
+  private def createRuntimeAttributes(wdlSource: WdlSource, runtimeAttributes: String = "") = {
```
mcovarr: Same issue with this default arg

supportedBackends/spark/src/test/scala/cromwell/backend/impl/spark/SparkRuntimeAttributesSpec.scala
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    workflowDescriptor.workflowNamespace.workflow.calls map {
+      call =>
+        val ra = call.task.runtimeAttributes.attrs mapValues { _.evaluate(createLookup(call), NoFunctions) }
+        TryUtil.sequenceMap(ra, "Runtime attributes evaluation").get
+    }
+  }
+
+  private def assertSparkRuntimeAttributesSuccessfulCreation(runtimeAttributes: Map[String, WdlValue], workflowOptions: WorkflowOptions, expectedRuntimeAttributes: SparkRuntimeAttributes): Unit = {
```
mcovarr: The name is misleading, this not only asserts successful creation but also equality.

supportedBackends/spark/src/test/scala/cromwell/backend/impl/spark/SparkJobExecutionActorSpec.scala
mcovarr: Is the coupling between `SparkJobExecutionActor` and `SparkClusterProcess` really so tight that the former can't be unit tested without essentially a real version of the latter?

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkClusterProcess.scala
mcovarr: @cjllanwarne that might be an even better choice.  😄

supportedBackends/spark/src/test/scala/cromwell/backend/impl/spark/SparkJobExecutionActorSpec.scala
jainh: `SparkClusterProcess` needs executioncontext from `SparkJobExecutionActor` in order to perform Rest Api call (Future). Therefore, all its behaviors  being brought inside SparkJobExecutionActorSpec. So, coupling between both depends upon rcfile which is a state of execution actor, also context and rest is cohesion (method calls).

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkClusterProcess.scala
mcovarr: the sleep should be on its own line

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkProcess.scala
jainh: @cjllanwarne  Yes.

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkJobExecutionActor.scala
mcovarr: can't this be calculated once and cached?  it's currently called in three different spots

jainh: Yes

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkJobExecutionActor.scala
```diff
@@ -0,0 +1,207 @@
...
+
+  // Files
+  private val executionDir = jobPaths.callRoot
+  private val scriptPath = jobPaths.script
+
+  private lazy val stdoutWriter = extProcess.untailedWriter(jobPaths.stdout)
+  private lazy val stderrWriter = extProcess.tailedWriter(100, jobPaths.stderr)
+
+  private lazy val clusterStdoutWriter = clusterExtProcess.untailedWriter(jobPaths.stdout)
+  private lazy val clusterStderrWriter = clusterExtProcess.tailedWriter(100, jobPaths.stderr)
```
mcovarr: It feels like the deploy mode is an abstraction that needs to be broken out here.  There are parallel groups of writers and processes, and "if cluster mode" logic which would seem to benefit from encapsulation and polymorphism.

jainh: This is what happening in the code right ? same abstraction inside the execution acotr but polymorphic calls based on deploymode.

supportedBackends/spark/src/test/scala/cromwell/backend/impl/spark/SparkRuntimeAttributesSpec.scala
iyanuobidele: It would, the default value will be removed to ensure that the method is called with the proper arguments

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkRuntimeAttributes.scala
```diff
@@ -0,0 +1,82 @@
...
+import cromwell.backend.validation.RuntimeAttributesValidation._
+import cromwell.core._
+import lenthall.exception.MessageAggregation
+import wdl4s.types.{WdlBooleanType, WdlIntegerType, WdlStringType, WdlType}
+import wdl4s.values.{WdlBoolean, WdlInteger, WdlString, WdlValue}
+
+import scalaz.Scalaz._
+import scalaz._
+
+object SparkRuntimeAttributes {
```
cjllanwarne: Another note re the latest develop: We've centralised the RuntimeAttribute evaluation, so all you need to do is override the `RuntimeAttributeDefinitions` on the backend factory, and hey-presto, the engine will deliver your requested runtime attributes as part of the `jobDescriptor`

supportedBackends/spark/src/test/scala/cromwell/backend/impl/spark/SparkJobExecutionActorSpec.scala
iyanuobidele: cleaned

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkClusterProcess.scala
iyanuobidele: Indeed the `?` is not needed.

The matching regex already does that. `...\{` escapes the initial opening brace and the closing brace is `...)}"""`

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkJobExecutionActor.scala
jainh: It is conflicting with right hand FileSystems name , i can change to FileSystemCollection or something.

supportedBackends/spark/src/test/scala/cromwell/backend/impl/spark/SparkJobExecutionActorSpec.scala
jainh: I looked at Mockito.spy(), that is a good option if we need to test the actual class like `SparkClusterProcess` but here `SparkExecutionActor` has dependency on `SparkClusterProcess` therefore we need to mock it in the `SparkExecutionActor`.  So i would like to keep it as it is.

gauravs90:
@jainh Please consider using git's `--author` tag to add @iyanuobidele as a contributor. It's usually  the standard way of crediting someone not having proper rights to a repo.

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkClusterProcess.scala
```diff
@@ -0,0 +1,151 @@
...
+        evaluateMonitoringFuture(rcPath)
+      case Failure(exception: Throwable) =>
+        logger.error(s"{} Spark Job failed to submit successfully following reason: {}", tag, exception.getMessage)
+        completionPromise success Failed(exception)
+    }
+    completionPromise.future
+  }
+
+  override def monitorSparkClusterJob(subId: String,  rcPath: Path, promise: Promise[Unit]): Unit = {
+    pollForJobStatus(subId) onComplete {
```
geoffjentry: You need to provide a fall through Success() state in the match. I realize that it's not practically a thing but the compiler can't enforce that due to the guard statements.

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkClusterProcess.scala
```diff
@@ -0,0 +1,151 @@
...
+      case Failure(error) =>
+        logger.error(s"{} Spark cluster poll for status failed Reason: {} and error: {}", tag, error.getMessage, error)
+        completeMonitoringProcess(rcPath, "-1", promise)
+    }
+  }
+
+  override def evaluateMonitoringFuture(rcPath: Path) = {
+    monitorPromise.future.onComplete {
+      case Success(_) =>
+        rcPath.contentAsString.stripLineEnd.toInt match {
```
geoffjentry: Don't have match statements which are lurking MatchError timebombs

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkInitializationActor.scala
```diff
@@ -0,0 +1,69 @@
...
+  override def beforeAll(): Future[Option[BackendInitializationData]] = Future.successful(None)
+
+
+  /**
+    * A call which happens before anything else runs
+    */
+  override def validate(): Future[Unit] = {
+    Future {
+      calls foreach { call =>
+        val notSupportedAttributes = call.task.runtimeAttributes.attrs filterKeys { !SupportedKeys.contains(_)}
```
geoffjentry: add a space before `}`

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkJobExecutionActor.scala
geoffjentry: perhaps `DefaultFileSystems`? I'm just saying that due to the `getDefault` on the RHS

geoffjentry: You could also explicitly refer to the right, e.g `file.FileSystems` and change the imports

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkJobExecutionActor.scala
geoffjentry: @mcovarr do you have an example in mind here?

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkJobExecutionActor.scala
```diff
@@ -0,0 +1,204 @@
...
+      log.debug("{} Creating execution folder: {}", tag, executionDir)
+      executionDir.toString.toFile.createIfNotExists(true)
+
+      log.debug("{} Resolving job command", tag)
+      val command= localizeInputs(jobPaths.callRoot, docker = false, fileSystems, jobDescriptor.inputs) flatMap {
+        localizedInputs => call.task.instantiateCommand(localizedInputs, callEngineFunction, identity)
+      }
+
+      log.debug("{} Creating bash script for executing command: {}", tag, command)
+      //TODO: we should use shapeless Heterogeneous list here not good to have generic map
```
geoffjentry: You know how to make me smile ;)

OTOH can you add a space after the `//`

supportedBackends/spark/src/test/scala/cromwell/backend/impl/spark/SparkJobExecutionActorSpec.scala
geoffjentry: @jainh I don't understand why the EC needs to come from a particular place?

I can't say that I care all that much but I start raising an eyebrow when people believe they need mocks at all.

jainh: @geoffjentry Are you saying we should use global EC instead of one from actor system dispatcher ? Anyhow `SparkClusterProcess` is the dependency to `SparkExecutionActor` and definitely needs to be mocked either way using Spy or Mock in general, therefore either of the approach would work but in case of Spy we need to supply arguments to the method in order to have real invocation. Let me know which way you guys prefer ? for me existing looks fine so far.

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkJobExecutionActor.scala
jainh: ok

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkClusterProcess.scala
jainh: I see what you mean.

supportedBackends/spark/src/test/scala/cromwell/backend/impl/spark/SparkJobExecutionActorSpec.scala
geoffjentry: What I mean is, if there's a dependency based on an EC, why couldn't just the EC be passed - why does the whole object need to be passed. I was just replying to a single line in one of your comments, which now that I reread was less about the EC and more about something else.

I still say that when one needs a mock they're either a) too tightly coupled, b) performing integration testing in unit tests or c) both a & b. 

However, this is a common complaint I have throughout Cromwell's test code so you're hardly the first nor worst offender.

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkClusterProcess.scala
```diff
@@ -0,0 +1,151 @@
...
+        logger.error(s"{} Spark Job failed to submit successfully following reason: {}", tag, exception.getMessage)
+        completionPromise success Failed(exception)
+    }
+    completionPromise.future
+  }
+
+  override def monitorSparkClusterJob(subId: String,  rcPath: Path, promise: Promise[Unit]): Unit = {
+    pollForJobStatus(subId) onComplete {
+      case Success(resp: SparkDriverStateQueryResponse) if resp.driverState == "RUNNING" =>
+        Thread.sleep(1000)
```
francares: Can this be configurable through config file?

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkJobExecutionActor.scala
```diff
@@ -0,0 +1,204 @@
...
+
+  private def isSparkClusterMode(deployMode: String, master: String): Boolean = {
+    // typical master value is spark://host-ip:6066 and deployMode also is cluster
+    List("6066", "spark").exists(master.toLowerCase.contains) && deployMode.equals("cluster")
+  }
+
+  private def taskLauncher = {
+    Try {
+      isClusterMode match {
+        case true => executionResponse completeWith  executeTask(clusterExtProcess, clusterStdoutWriter, clusterStderrWriter)
```
francares: Check format.

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkJobExecutionActor.scala
```diff
@@ -0,0 +1,204 @@
...
+    } catch {
+      case ex: Exception =>
+        log.error(ex, "Failed to prepare task: " + ex.getMessage)
+        executionResponse success FailedNonRetryableResponse(jobDescriptor.key, ex, None)
+    }
+  }
+
+  private def isSparkClusterMode(deployMode: String, master: String): Boolean = {
+    // typical master value is spark://host-ip:6066 and deployMode also is cluster
+    List("6066", "spark").exists(master.toLowerCase.contains) && deployMode.equals("cluster")
```
francares: Extract these to const.

jainh: @francares as @geoffjentry  suggested if it is not being used more than one place, he prefers to be de-constantized. Earlier it was in constant.

geoffjentry: Yes. @francares please do not make constants for things which are only used in one place - IMO that's an awful practice which should be left behind in the java world.

francares: val names helps with the context description. I don't know at simple sight what 6066 and spark mean there.

geoffjentry: And cluttering up the code base with junk like `val Spark = "spark"` also doesn't do anything and just adds code clutter.

francares:
LGTM 👍

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkJobExecutionActor.scala
francares: Agree with that but that's an example of a bad practice :). 
I think defining SparkUrlPrefix will help or just UrlPrefix.
Anyway this is minor change as you want.

supportedBackends/spark/src/test/scala/cromwell/backend/impl/spark/SparkJobExecutionActorSpec.scala
jainh: Ok, I have changed `SparkExecutionActorSpec` only unit tests and `SparkClusterProcessSpec` as separate unit test. However `SparkClusterProcess` needs not only EC (I forgot to mention before) but the `actorsystem` because of Spray client for making rest calls.

jainh:
@geoffjentery is there anything else left ?

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkClusterProcess.scala
```diff
@@ -0,0 +1,165 @@
...
+        logger.error(s"{} Spark Job failed to submit successfully following reason: {}", tag, exception.getMessage)
+        completionPromise success Failed(exception)
+    }
+    completionPromise.future
+  }
+
+  override def monitorSparkClusterJob(subId: String, rcPath: Path, promise: Promise[Unit]): Unit = {
+    pollForJobStatus(subId) onComplete {
+      case Success(resp: SparkDriverStateQueryResponse) if resp.driverState == "RUNNING" =>
+        Thread.sleep(1000)
```
geoffjentry: You have an actor system, why not use Retry instead of blocking a thread? 

I can't overstate how much time we've speant dealing with fires in production due to Thread.sleep. I'd like to see the usages go down, not up

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkClusterProcess.scala
```diff
@@ -0,0 +1,165 @@
...
+        logger.debug(s"{} Spark Driver is now in :{} state", tag, "RUNNING")
+        monitorSparkClusterJob(subId, rcPath, promise)
+      case Success(resp: SparkDriverStateQueryResponse) if resp.driverState == "FINISHED" =>
+        logger.debug(s"{} Spark Driver is now in :{} state", tag, "FINISHED")
+        completeMonitoringProcess(rcPath, "0", promise)
+      case Success(resp: SparkDriverStateQueryResponse) if resp.driverState == "FAILED" =>
+        logger.debug(s"{} Spark Driver is now in :{} state", tag, "FAILED")
+        completeMonitoringProcess(rcPath, "-1", promise)
+      case Success(_) =>
+        logger.debug(s"{} Spark Driver is now in Unknown state as polling returned success", tag)
```
geoffjentry: Is a success that returns a driver state you're unaware of really a success or should it be an error? I can see arguments for either way but wanted to see what argument you went with.

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkClusterProcess.scala
```diff
@@ -0,0 +1,165 @@
...
+        logger.error(s"{} Spark cluster poll for status failed Reason: {} and error: {}", tag, error.getMessage, error)
+        completeMonitoringProcess(rcPath, "-1", promise)
+    }
+  }
+
+  override def evaluateMonitoringFuture(rcPath: Path) = {
+    monitorPromise.future.onComplete {
+      case Success(_) =>
+        val rc = rcPath.contentAsString.stripLineEnd.toInt
+        if (rc == 0) {
```
geoffjentry: move the if branch into the case statements

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkClusterProcess.scala
jainh: It should be an error.

geoffjentry:
@jainh I'd want to take another scan after rebase, just because there's a lot of changes, and I'd really prefer to find a way to not need the `Thread.sleep`, but 👍

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkClusterProcess.scala
jainh: Ok

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkClusterProcess.scala
jainh: Ok let me see

geoffjentry: If you can't you can't - e.g. we still have one in the engine which is too much of a pain to remove at the moment due to not having an actor system. But I've been bit by these way too many times to not try to avoid them

jainh: I changed `Thread.sleep()` to `system.scheduler.scheduleOnce`

geoffjentry:
@jainh Thanks! :)

jainh:
@iyanuobidele can you please pull it and try one more round of integration testing with the Spark cluster ? 
@geoffjentry : I rebased it with the develop branch let me know how does it look ? I will merge it then. Thank you.

core/src/main/resources/application.conf
```diff
@@ -347,4 +375,4 @@ database {
       }
     }
   }
-}
+}
```
geoffjentry: \n

project/Dependencies.scala
```diff
@@ -107,4 +111,4 @@ object Dependencies {
   ) ++ sprayServerDependencies
 
   val rootDependencies = slf4jBindingDependencies
-}
+}
```
geoffjentry: \n

project/Settings.scala
```diff
@@ -93,4 +98,4 @@ object Settings {
     name := "cromwell",
     libraryDependencies ++= rootDependencies
   ) ++ commonSettings
-}
+}
```
geoffjentry: \n

geoffjentry:
modulo a couple of newlines 👍

PR #1340, opened 2016-08-25
Restart/recover migration.  Closes #1119

### LINKED ISSUE #1119: Data migration for restart/recover
When a user upgrades a production instance of cromwell from 0.19 -> 0.21, we need to be able to preserve the state of running workflows so that when the server restarts their workflows are resumed.

mcovarr:


build.sbt
```diff
@@ -11,6 +11,7 @@ lazy val gcsFileSystem = (project in file("filesystems/gcs"))
 
 lazy val database = (project in file("database"))
   .settings(databaseSettings:_*)
+  .dependsOn(core) // used for migration
```
kshakir: ToL: Going to split migration, and its dependencies such as liquibase, core, etc., out into a separate artifact just after this PR. Fine for now.

database/src/main/resources/changesets/restart_and_recover_migration.xml
```diff
@@ -0,0 +1,138 @@
...
+              REPLACE(WORKFLOW_OPTIONS, 'workflowFailureMode', 'workflow_failure_mode');
+
+            UPDATE WORKFLOW_STORE SET WORKFLOW_OPTIONS =
+              REPLACE(WORKFLOW_OPTIONS, 'workflow_log_dir', 'final_workflow_log_dir');
+
+            UPDATE WORKFLOW_STORE SET WORKFLOW_OPTIONS =
+              REPLACE(WORKFLOW_OPTIONS, 'outputs_path', 'final_workflow_outputs_dir');
+
+            UPDATE WORKFLOW_STORE SET WORKFLOW_OPTIONS =
+              REPLACE(WORKFLOW_OPTIONS, 'call_logs_dir', 'final_call_logs_dir');
```
kshakir: I think there's a bug in all of these update/replace statements.

Let's assume that this _isn't_ provenance, we've just chosen to store the WORKFLOW_OPTIONS in the database as a string. We still should only modify the json keys, not the values.

```
mysql> select replace('{"foo_key":"foo_value"}', 'foo', 'bar');
+--------------------------------------------------+
| replace('{"foo_key":"foo_value"}', 'foo', 'bar') |
+--------------------------------------------------+
| {"bar_key":"bar_value"}                          |
+--------------------------------------------------+
1 row in set (0.03 sec)

mysql> 
```

database/src/main/scala/cromwell/database/migration/restart/table/JobStoreSimpletonMigration.scala
```diff
@@ -0,0 +1,103 @@
...
+      val wdlType = WdlType.fromWdlString(results.getString(4))
+      val rawValue = results.getString(3)
+      val inflated = inflate(rawValue)
+      val wdlValue = wdlType match {
+        case p: WdlPrimitiveType => p.coerceRawValue(inflated).get
+        case o => o.fromWdlString(inflated)
+      }
+      val name = results.getString(2)
+      val simpletons = wdlValue.simplify(name)
+      simpletons foreach { s =>
```
kshakir: `simpletons foreach { simpleton =>`

I'm going to leave the single line variables alone. But if my eye has to leave the line to figure out what `s` , `p`, `o`, etc. are, I'm going to ask you to type a few more characters into the IDE.

geoffjentry: We're never going to agree on this, but that's never stopped me from arguing ... ;)

In this case you don't need to scan down because you see `s` is tied to `simpletons` so when you see `s` you know it's a `simpleton`. 

I strongly prefer single character placeholders, I find it annoying having to see `simpleton` taking up all of that visual space when (IMHO) it's clear what it is. I'd rather spend my time thinking about what's happening to that simpleton

geoffjentry: Also, see the scala style guide:
http://docs.scala-lang.org/style/naming-conventions.html#classestraits

project/Dependencies.scala
```diff
@@ -66,9 +66,10 @@ object Dependencies {
   val databaseDependencies = List(
     "com.typesafe.slick" %% "slick" % slickV,
     "com.typesafe.slick" %% "slick-hikaricp" % slickV,
-    "org.hsqldb" % "hsqldb" % "2.3.2",
+    "org.hsqldb" % "hsqldb" % "2.3.4",
     "mysql" % "mysql-connector-java" % "5.1.36",
-    "com.github.pathikrit" %% "better-files" % betterFilesV % Test
+    "com.github.pathikrit" %% "better-files" % betterFilesV % Test,
+    "org.broadinstitute" %% "wdl4s" % wdl4sV // Used in migration scripts
```
kshakir: ToL: Booooo. Database segregation forever! Btw. I've decide I'm switching us from slick to doobie because I like the name better. _/s_

geoffjentry: If you all want to move towards libraries that are militant about pure FP and typeclasses I'd support that ;)

engine/src/main/scala/cromwell/jobstore/SqlJobStore.scala
```diff
@@ -56,6 +56,7 @@ class SqlJobStore(sqlDatabase: JobStoreSqlDatabase) extends JobStore {
       case "Int" => WdlIntegerType
       case "Float" => WdlFloatType
       case "Boolean" => WdlBooleanType
+      case "File" => WdlFileType
```
kshakir: ToL: We can discuss in person- I'm curious about simpletonization/serialization of `WdlFile`, especially as there isn't currently a one type to one value-type mapping.

`WdlFileType` (currently) has `WdlSingleFile` and `WdlGlobFile`. I'm not sure how knowing one has a `WdlFileType` alone tells you which simpleton type one should use for a round trip.

Premature optimization sucks, but based on our early discussions yesterday, I feel like groups of files might actually be appearing soon, further complicating the one file type to N value-types mapping.

database/src/main/scala/cromwell/database/migration/restart/table/JobStoreSimpletonMigration.scala
kshakir: Well cited. You win this philosophical battle, this time... ;)

 I'll move my eyes so one doesn't have to move their fingers.

geoffjentry: Oh you gave up way too easy. I was hoping you'd cite the followup lines that mention keeping blocks short and sweet.

Or perhaps what really happened is you astutely foresaw the larger philosophical rant I was trying to find an excuse to go on ... ;)

engine/src/main/scala/cromwell/jobstore/SqlJobStore.scala
mcovarr: Yeah @ruchim and I noticed this the other day.  We only use `WdlValueSimpleton`s to represent call outputs in both the job store and call cache result contexts.  So I wouldn't expect these `File`s to represent expressionish `WdlGlobFile`s.  The current code *appears* to be doing the right thing for our use cases in making `WdlSingleFile`s, but it probably wouldn't hurt to give this a bit more scrutiny.

database/src/main/resources/changesets/restart_and_recover_migration.xml
mcovarr: Yeah this was a last minute hack.  I had been iteratively fighting my way through the preceding migrations for quite a while and was taken by surprise when they all worked and Cromwell actually began to restart workflows.  😄   The restarts failed for the mismatch in`defaultRuntimeOptions` vs `default_runtime_attributes` and this was a quick bandaid to get past that.

Anyhoo, I agree this is bad and should be fixed.  I don't see a way of doing this generally in straight SQL, but fortunately the scale is small enough that this could be handled as a custom task change class that actually parses the JSON map and only renames the keys.

engine/src/main/scala/cromwell/jobstore/SqlJobStore.scala
kshakir: Still ToL: If we're starting to discuss "current code" vs. this "migration code", _in theory_ I may want a separation. The migration code shouldn't change. It will always migrate 0.19 to 0.21. Meanwhile current code will continue to change to suit the needs of the application.

That said this may be a non-issue for active users. And if there was someone migrating from <0.19 to a much later version, we did discuss that one could just be required to say run the binary cromwell 0.21-0.x _before_ running say cromwell 1.x.

supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystem.scala
```diff
@@ -147,7 +147,9 @@ trait SharedFileSystem extends PathFactory {
       case (name, value) => localizeFunction(value) map { name -> _ }
     }
 
-    TryUtil.sequence(localizedValues, "Failures during localization").map(_.toMap)
+    TryUtil.sequence(localizedValues, "Failures during localization").map(_.toMap) recover {
+      case e => throw new CromwellFatalException(e)
+    }
```
mcovarr: I added this since otherwise the failed localizations were retried infinitely.

database/src/main/scala/cromwell/database/migration/restart/table/RenameWorkflowOptionKeysMigration.scala
```diff
@@ -0,0 +1,61 @@
...
+       SELECT WORKFLOW_STORE_ID, WORKFLOW_OPTIONS FROM WORKFLOW_STORE
+    """
+
+  private val UpdateWorkflowStore =
+    """
+       UPDATE WORKFLOW_STORE SET WORKFLOW_OPTIONS = ? WHERE WORKFLOW_STORE_ID = ?
+    """
+}
+
+
```
kshakir: IMHO, there's a ton of extra whitespace above, and those aren't even really multi-line strings.

database/src/main/scala/cromwell/database/migration/restart/table/RenameWorkflowOptionKeysMigration.scala
```diff
@@ -0,0 +1,61 @@
...
+    while (results.next()) {
+      val options = results.getString(2)
+      if (!results.wasNull()) {
+        val optionsJson = options.parseJson
+        val newOptionsJson = optionsJson match {
+          case JsObject(fields) => JsObject(fields map renameOptionKeys)
+          case other => other // There really shouldn't be workflow options of other types, but if there are pass them through.
+        }
+
+        insert.setString(1, newOptionsJson.prettyPrint)
```
kshakir: ToL: Someday, if this isn't provenance, we'll switch to a structured, simpleton(?) storage. Today is not that day.

kshakir:
The future enhancements of `WdlFile` affecting this migrator still makes me uneasy, but this code looks good as is. 👍 

Also I'm assuming this won't get merged before the other migration PR?

database/src/main/resources/changesets/restart_and_recover_migration.xml
```diff
@@ -0,0 +1,123 @@
...
+              RETRYABLE_FAILURE            -- 8
+            )
+            SELECT
+              we.WORKFLOW_EXECUTION_UUID,  -- 1
+              e.CALL_FQN,                  -- 2
+              e.IDX,                       -- 3
+              e.ATTEMPT,                   -- 4
+              1,                           -- 5
+              e.RC,                        -- 6
+              null,                        -- 7
```
cjllanwarne: Assuming null is the same as empty?

database/src/main/resources/changesets/restart_and_recover_migration.xml
```diff
@@ -0,0 +1,123 @@
...
+              we.STATUS = 'Running'
+        </sql>
+    </changeSet>
+
+    <changeSet id="JOB_STORE_MIGRATION" author="mcovarr" >
+        <comment>
+            Restart/recover migration from 0.19 to 0.21.
+        </comment>
+        <sql>
+            INSERT INTO JOB_STORE(
```
cjllanwarne: I don't see any migration of the Job Store simpletons, is that deliberate?

database/src/main/scala/cromwell/database/migration/restart/table/RenameWorkflowOptionKeysMigration.scala
mcovarr: That's what I get for copy/pasting the other migration as a baseline.  Will fix.

database/src/main/resources/changesets/restart_and_recover_migration.xml
mcovarr: I should confirm that.

engine/src/main/scala/cromwell/jobstore/SqlJobStore.scala
cjllanwarne: We should tech talk this but long story short: you'll never have a WdlGlobFile here. Always reinstantiate to WdlSingleFile

mcovarr:
@kshakir Right, I'm not planning to merge this before @Horneth's work; he's got enough to worry about on his branch without changelog conflicts and waiting for my migrations to run.  😄

cjllanwarne:
LGTM 👍

engine/src/main/scala/cromwell/jobstore/SqlJobStore.scala
mcovarr: Let me see if I can rework this to be more futureproof.

mcovarr: Actually upon further reflection I'm going to ticket this and revert these changes.  This change is unrelated to migrations but rather to a bug @ruchim discovered in the call cache result code.  I realized I had the same bug in the job store code and was slipping the changes into this PR, but given the discussion here I'm going to back away from that.

mcovarr: Unless I really need it for the migrations to run, which I might.  😦

database/src/main/resources/changesets/restart_and_recover_migration.xml
mcovarr: Yup looks to be the same.

mcovarr:
Just a heads up, after talking with @Horneth and @cjllanwarne I'm planning to merge this before Thibault's PR if/when these builds go green.

PR #1346, opened 2016-08-26
Fix polling

ffinfo:
While polling the method isAlive was not executed.


cjllanwarne:
I like the idea here but I think we're probably now over the boundary of "do it sync and then wrap the result in Future.successful". A hidden cost to `isAlive` is that we have to create a shell process and wait for it to complete every time we query a job's status using it.

@geoffjentry @kshakir if we have to fire up a new shell process for every status query, I will re-suggest my previous idea of putting the whole thing behind a routed actor pool so that we can at least rate-limit it... 😄

kshakir:
Discussed on gitter:

> Whoops, sorry, I may have been too vague. `isAlive` is considered expensive, and will be only used on startup. But, we can consider reducing the verbosity of the rc check if you’d like.

I concur with @cjllanwarne.

cjllanwarne:
Looking at the gitter convo, the problem looks more like a case of overzealous logging - could we just drop it from `info` down to `debug`? I'm curious why it was `info` in the first place tbh

ffinfo:
Well if it's doing it only when the job is finished I would even make it `warn` instead of `info`.

> Right now a workflow is not robust enough to run it on a/our HPC, see also on gitte:
> Peter van 't Hof @ffinfo Aug 26 19:05
> ye I see what you mean but it's the only way. When in SGE your job is killed he never get to the point of $? > rc
> so cromwell can not detect is a job is killed, meaning it will end in a endless loop polling for rc what never will come anymore
> in this case a drmaa connection would be better
> but not so sure if that still works on a start of a server
> I think there it's bound to a session
> 
> Peter van 't Hof @ffinfo Aug 26 19:11
> but only have seen the dmraa implementation inside Gatk Queue
> 
> Peter van 't Hof @ffinfo Aug 26 19:28
> when using qstat i would use it only once for the complete pool instead executing it for each job
> so then you get an output like this:
> `
> job-ID prior name user state submit/start at queue slots ja-task-ID
> 9923549 0.00000 cromwell_1 pjvan_thof qw 08/26/2016 17:23:16 1
> 9923550 0.00000 cromwell_1 pjvan_thof qw 08/26/2016 17:23:16 1
> `
> this is only 2 jobs but having a lot of jobs this will reduce the load a lot
> 
> kshakir @kshakir Aug 26 21:21
> True, Cromwell will end up in an endless loop if someone terminates the SGE job, or if the rc file doesn’t appear in general. One could use isAlive intermittently, but it was introduced mainly for recovering jobs at re-startup, & I would not have isAlive poll as often as we check for the rc file. Btw, GATK Queue actually only checks drmaa every 30 seconds, so that it doesn’t overload dispatchers. Something like isAlive could be checked with similar frequency. All this is a bigger discussion that could be tracked in a git issue.
> Separately, I am hearing from multiple people that the rc poll logs are spam. 
> 
> Peter van 't Hof @ffinfo Aug 26 21:44
> As already suggested in the PR, a actor pool would be better I think but that's not a small change indeed
> mostly jobs are running way longer that 10 or 30 sec does not matter a lot 
> 
> Peter van 't Hof @ffinfo Aug 26 21:50
> On our cluster we need something like retries but if it goes to an endless loop he will never retry.

In it's current state it's for us not yet usable but If you open to it I can think/test things then there is a improvement on this. I can even try to get some time to do some developing but that I can't promise directly.

I can try to look into a lfs.drama backend. If it's possible to check running jobs after restart this would be nice to have, but need to look into this.

What to do with this PR i leave up to you guys, merging it or not does not really matter for me at this point ;).

ffinfo:
Maybe this is a workaround till there is a actor pool in place to replace this? Tested this on SGE and it seems to work. Now isAlive is only executing a shell command once each 20 seconds. This 20 seconds can maybe be inside the config but first want to know what you guys think of this.

supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystemAsyncJobExecutionActor.scala
```diff
@@ -325,12 +328,25 @@ trait SharedFileSystemAsyncJobExecutionActor
...
-    val argv = checkAliveArgs(job).argv
-    val stdout = pathPlusSuffix(jobPaths.stdout, "check")
-    val stderr = pathPlusSuffix(jobPaths.stderr, "check")
-    val checkAlive = new ProcessRunner(argv, stdout, stderr)
-    checkAlive.run() == 0
+    val now = Calendar.getInstance()
+    jobLogger.debug(s"executing isAlive for ${job.jobId}")
+    if (isAliveCash(job).cash && (isAliveCash(job).last.getTimeInMillis + isAliveTimeout) < now.getTimeInMillis) {
+      val argv = checkAliveArgs(job).argv
+      jobLogger.debug(s"Executing shell for ${job.jobId}: " + argv.mkString("'", " ", "'"))
```
ffinfo: While it does work on SGE I see this in my log:
`
[2016-08-29 13:37:07,74] [debug] SharedFileSystemAsyncJobExecutionActor [13f02c3apipe.fastqsync:NA:1]: Executing shell for 28265: 'ps 28265'
`

I would expect here to see `qstat -j 28265`, he must also execute this because it is working. Anyone an idea what goes wrong here?

cjllanwarne:
@ffinfo just to be clear - unless the script explodes in a **really** unusual way, you should *always* eventually get a return code file. So for us, waiting for it to appear is sufficient. (e.g. even if a command fails, we still get the return code file). 

In fact, in previous Cromwell versions (before making these backends "config-file based") our hard-coded local and SGE backends waited for the return code file and never checked `isAlive` apart from at restart-time.

Just so I can understand, is there a specific reason that you don't believe that an `rc` file will eventually be produced for your situation? (and if so, could we maybe try to protect that instead?)

supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystemAsyncJobExecutionActor.scala
cjllanwarne: This feels like a bug... @kshakir?

cjllanwarne:
From my point of view, I'd recommend merging this one as is. If we start seeing issues with too many shells being spawned we can look again at routing the shell requests. 👍

kshakir:
Hi @ffinfo

I originally thought this was going to be a patch to just lower the rc polling verbosity, but you've begun tackling a much bigger issue. Thanks a bunch for your work so far!

We've also got a bunch of other work we're juggling at the moment, so it's unlikely I or others on the team will have time to look at this issue of disappearing SGE jobs in depth for at least a couple weeks.

For now, here's a brain dump of notes.

After a short bit of review, I'd perhaps try a different approach.

- On execute or recover, `scheduleOnce` a message back to `self` to later check if a job is alive.
- When the message is received check if the job is alive.
- If the job is alive `scheduleOnce` a message to check if the job is alive again.
- If the job is not alive write an rc file with `143` (or other code, see notes on configuration below).
- An instance of `cromwell.core.retry.Backoff` should travel inside the scheduled messages. Each time the message is to be scheduled, get the next time.

As for the existing code, here are a few notes.

- Use `java.time` instead of `java.util`. `java.time.Instant` and `java.time.Duration` may be used to calculate the amount of time between two instants.
- `IsAliveCash.cash` should be `.cache`.
- `.map(_.cache).getOrElse(true)` should be `.forall(_.cache)`, however...
- `.cache` always appears to be `true`, and thus not needed.
- `!isAliveCache.contains` followed by `isAliveCache.get` should be `isAliveCache.getOrElseUpdate(job, IsAliveCache(Instant.now))`.
- There should be only one `SharedFileSystemJob` per `SharedFileSystemAsyncJobExecutionActor`. The reason the values are passed around is because the actor is also partially stateless, using `context.become` to track the `SharedFileSystemJob`. This PR adds state more to the actor, outside of the context, but that shouldn't be needed if the `isAlive` check is switched to running due to multiple `scheduleOnce` calls.

The tests are likely timing out because the of the extra checks to `isAlive`. The tests are meant to run as quickly as possible. In general, the order of the job completion checking should always be 1) multiple rc file polls, 2) 30 seconds later `isAlive` checks as necessary. This individual polling per job may overwhelm the SGE scheduler if hundreds or thousands of scatter jobs are running, so more work may need to be done in the future to simplify the qstat process to check jobs in batches.

Notes on configuration:

The initial scheduling should also be configurable. It should be off by default. Also, the error code that is returned may want to be configurable and/or we would want the actor to handle this special case differently and reattempt instead of fail.

Cromwell's SFS implementation assumes jobs will always be writing their rc files. If something else out there is truly killing the jobs, we will have to wire in a way for `poll` to return a `FailedRetryableExecutionHandle`. I'm not sure that writing a value into the rc file is the best way to do this, and not yet sure what a suitable alternative is also.

That's all I've got for now. Thanks again for all your work so far! I'm also game if we move this discussion over to a github issue instead of a PR, as I suspect the final version will look a bit different, and we can discuss and capture any other design there instead.

geoffjentry:
@ffinfo Hi Peter - apologies for taking so long, the release I mentioned ended up taking a while longer than we thought.

I talked to our PO this morning about this pull request and his take was that if this could be hooked up in a way which keeps the tests green (as much as they ever are) and doesn't add noticeable latency in the system for other users (and/or the behavior change is put behind a config option) that he'd be good with this concept. 

It's been a month now so it's entirely possible you've already moved on with life or perhaps you have no interest for other reasons so I'll leave it up to you on how to proceed

geoffjentry:
I'm closing this as there's been no activity for a few months now. We still like the idea - @ffinfo if this is something you're still interested in let me know and we can chat

PR #1350, opened 2016-08-26
Fix JobStore handling of File outputs.  Closes #1349.

### LINKED ISSUE #1349: Job Store doesn't handle File outputs
Currently the Job Store doesn't cope with File outputs at all.  See the discussion at #1340 for thoughts about doing this in a futureproof way.  Also think about the how much logic will be shared with call cache results in #1233, and particularly about factoring out the simpleton nature in the job store and call cache result simpletons.

mcovarr:
* Handle "File" output types for the JobStore
* Create a common `DatabaseSimpleton` trait for Job Store and call cache result code to share
* Centralize logic for conversion of `DatabaseSimpleton`s to `WdlValueSimpleton`s

### DIFF
--- a/database/src/main/scala/cromwell/database/sql/tables/CallCachingResultSimpletonEntry.scala
+++ b/database/src/main/scala/cromwell/database/sql/tables/CallCachingResultSimpletonEntry.scala
@@ -7,4 +7,4 @@ case class CallCachingResultSimpletonEntry
   wdlType: String,
   resultMetaInfoId: Int,
   callCachingResultSimpletonId: Option[Int]
-)
+) extends DatabaseSimpleton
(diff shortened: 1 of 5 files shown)


kshakir:
Would you object to just adding another overload of:
```private def toSimpleton(entry: CallCachingResultSimpletonEntry): WdlValueSimpleton```

If not at this point and time, no worries and I'll submit something for PR review in the future.

mcovarr:
Sorry, I'm not sure what you mean.  For context, Ruchi has some forthcoming work for the call cache result stuff that will want to do exactly the same db simpleton -> WdlValueSimpleton conversions as the job store.

geoffjentry:
👍

kshakir:
No worries. I'll submit a later suggestion inside a code PR. There would be no traits under `package cromwell.database.sql.tables`, because MySQL, HSQLDB, etc. don't have traits.

I'm desperately trying to keep the `database` layer modeling what's-in-the-database, versus modeling what-cromwell-would-like-to-see, ahead of Slick 4, doobie, or any other refactorings.

kshakir:
👍

kshakir:
Btw, the closes "Closes" in the commit message and the PR name are pointing to the wrong id.

PR #1352, opened 2016-08-29
Fixed MockitoSugar import.

kshakir:


cjllanwarne:
So MockitoSugar still exists, it just needs to be imported from elsewhere?

👍

mcovarr:
:+1:

kshakir:
Yep. It just moved.

PR #1353, opened 2016-08-29
Recent change to JES travis test made it blind to failures

geoffjentry:


src/bin/travis/testCentaurJes.sh
```diff
@@ -72,3 +73,5 @@ export CENTAUR_LOG_PATH="gs://cloud-cromwell-dev/cromwell_execution/travis/centa
 gsutil cp ${CENTAUR_LOG_PATH} centaur.log
 cat centaur.log
 echo "More logs for this run are available at https://console.cloud.google.com/storage/browser/cloud-cromwell-dev/cromwell_execution/travis/centaur/${WORKFLOW_ID}/call-centaur/"
+
+exit "${EXIT_CODE}"
```
geoffjentry: add \n

cjllanwarne: The `"`s seem superfluous (if not dangerous?)

cjllanwarne:
I'm assuming coercion to/from string doesn't actually hurt anything though so 👍

src/bin/travis/testCentaurJes.sh
geoffjentry: If I don't put those in, @kshakir tells me to add them

kshakir: How are quotes dangerous? Not testing to see if a variable is set is dangerous, but not quoting to avoid accidental shell expansion.

kshakir:
👍

PR #1361, opened 2016-08-30
filter out local docker tests on jes travis

geoffjentry:


geoffjentry:
Merging as-is. I'll say a few hail marys as penance

PR #1366, opened 2016-09-01
Tracking summary status in the database instead of in memory. Closes #978

### LINKED ISSUE #978: Record last processed summary ID in db
As suggested by @cjllanwarne, record the last processed metadata journal ID in the database.  This will allow Cromwell to not spin its wheels processing metadata pointlessly on restarts, as well as give more explicit control if a full summary rebuild is desired (clear WORKFLOW_METADATA_SUMMARY, reset last processed METADATA_JOURNAL id to 0).

kshakir:
Also:
Splitting the database sql and migration, plus other cleanup.
Bits of call caching business logic refactored out of `database`.
Moved all simpleton conversion out of the `database`.
PK columns that cannot be filled in by business logic are defaulted to `None`.
Renamed `Database` to `ServicesStore`.
Renamed `CromwellDatabase` to `SingletonServicesStore`.


geoffjentry:
Don't read this as anything other than tongue in cheek, but aren't you the one who gripes about potpourri being mixed in with PRs? :)

geoffjentry:
👍

kshakir:
In my defense I split the PR into separate, digestible commits. 😉

build.sbt
```diff
@@ -9,19 +9,22 @@ lazy val gcsFileSystem = (project in file("filesystems/gcs"))
...
+  .dependsOn(core)
   .withTestSettings
 
 lazy val services = (project in file("services"))
   .settings(servicesSettings:_*)
   .withTestSettings
   .dependsOn(core)
-  .dependsOn(database) // Assuming services impl in services
+  .dependsOn(databaseSql)
+  .dependsOn(databaseMigration)
```
Horneth: Out of curiosity, why do the services depend on the migration project ?

kshakir: The service store runs the liquibase migration before using their sql database.

Horneth: Hmm ok - I guess I didn't realize the database was actually now the "services database" and not "cromwell database". If that's the case then it makes sense.

kshakir: Yep, we apparently had created a full services layer under `cromwell-services`. I quite fond of the way the separation of concerns has emerged, in that `cromwell-backend`/`cromwell-engine` talks to the `cromwell-services` that talks to the `cromwell-database` (if necessary).

database/sql/src/main/scala/cromwell/database/slick/MetadataSlickDatabase.scala
```diff
@@ -100,17 +100,21 @@ trait MetadataSlickDatabase extends MetadataSqlDatabase {
...
 
-  def refreshMetadataSummaries(startId: Long, startTimestamp: Option[Timestamp],
+  def refreshMetadataSummaries(startTimestamp: Option[Timestamp],
                                key1: String, key2: String, key3: String, key4: String,
                                buildUpdatedSummary: (Option[WorkflowMetadataSummary], Seq[Metadatum]) =>
                                  WorkflowMetadataSummary)
-                              (implicit ec: ExecutionContext): Future[Seq[Metadatum]] = {
+                              (implicit ec: ExecutionContext): Future[Long] = {
     val action = for {
+      startIdOption <- getSummaryStatusMaximumId("WORKFLOW_METADATA_SUMMARY", "METADATA_JOURNAL")
```
Horneth: I don't know if it's possible, but is there any way we can reference the name of the table through the ***Component class to avoid copy / pasting it ?

Horneth:
👍

database/sql/src/main/scala/cromwell/database/slick/MetadataSlickDatabase.scala
kshakir: I'll double check, but I didn't see an easy way I recognized. It may just be easier to start using constants within the components.

kshakir: I've decided to merge as is. I looked at how one might retrieve the table name from slick, and the direction I [found](http://stackoverflow.com/a/28706363/3320205) went through the `TableQuery.baseTableRow`-- except the inline docs on that method [explicitly](https://github.com/slick/slick/blob/3.1.1/slick/src/main/scala/slick/lifted/Query.scala#L298-L299) hint that it's not for public consumption.

Instead, I _feel_ like the table names (and column names, and index names, etc.) should be defined not by slick, but up in the base `cromwell.database.sql.tables` layer. However I also feel like the slick code adds so much verbosity already, that I hesitate to add yet-another set of boilerplate/wiring to be filled in to get a table up and running, in addition to the liquibase modeling too.

I'm very ok with someone smoothing this out in a future PR.

PR #1370, opened 2016-09-02
Bump up to akka 2.4.7

geoffjentry:


project/Dependencies.scala
```diff
@@ -12,7 +12,7 @@ object Dependencies {
   - http://doc.akka.io/docs/akka/2.4/scala/http/common/json-support.html#akka-http-spray-json
    */
   lazy val sprayJsonV = "1.3.1"
-  lazy val akkaV = "2.3.15"
+  lazy val akkaV = "2.4.7"
```
geoffjentry: To fend off any questions asking "why not 2.4.9", it appears that the lightbend reactive platform is built against 2.4.7. It might not matter, in which case i'll later go to 2.4.9

mcovarr:
:+1:

geoffjentry:
@francares IIRC you all were wanting 2.4.x for something in particular, so NB

geoffjentry:
Actually now that I look at http://doc.akka.io/docs/akka/2.4.0/project/migration-guide-2.3.x-2.4.x.html I might take a spin around our codebase before merging

kshakir:
In addition to the akka migration docs, and the commented out implicit compilation error, there a several new compilation warnings / recommendations that appeared with this change. They each look straightforward enough to fix and quiet, in a following PR or this PR.

While we're in there, perhaps we can also use [`"com.timushev.sbt" % "sbt-updates"`](https://github.com/rtimush/sbt-updates) to update our other dependencies in cromwell, lenthall, and wdl4s, too.

Btw, the explicit form that compiles your implict error:

```scala
val futureAny = actorRef.?(message)(timeout = timeout, sender = actorRef)
```

Switch `timeout` and `actorRef` with whatever `implicit val` you intended, but those were the closest in scope from `trait DefaultTimeout` and the enclosing method parameters, respectively.

geoffjentry:
@kshakir thanks. BTW I'm not sure why there's that implicit actor ref anyways. that whole class needs to be nuked from orbit :'(

kshakir:
👍

Issue #1377, opened 2016-09-07
FileHasherWorkerActor needs some love


geoffjentry:
This is currently being built by all implementers of the BackendLifecycleActorFactory, as opposed to props like the trait normally does. @cjllanwarne explained that this is because we want system-wide throttling on this behavior. 

At the moment it's being built as a direct descendent of the root, ie a sibling of CromwellRootActor, and this doesn't seem correct. Also, it appears that all instances of this router will have the same name which should be problematic.

Make this more sane please

cjllanwarne:
This appears to be resolved

PR #1379, opened 2016-09-08
Multiple updates

kshakir:
See individual commit messages for more info.


kshakir:
Will push a patch updating the lenthall and wdl4s dependencies-- once those PRs get one more thumb.

geoffjentry:
:+1:

backend/src/main/scala/cromwell/backend/BackendLifecycleActorFactory.scala
```diff
@@ -48,7 +44,5 @@ trait BackendLifecycleActorFactory {
   lazy val fileHashingFunction: Option[FileHashingFunction] = None
   lazy val fileHashingWorkerCount: Int = 50
 
-  final lazy val fileContentsHasherActor: ActorRef = actorSystem.actorOf(RoundRobinPool(fileHashingWorkerCount)
-    .props(FileHasherWorkerActor.props(fileHashingFunction)),
-    "FileContentsHasherActor")
+  def fileContentsHasherActorProps: Props = FileHasherWorkerActor.props(fileHashingFunction)
```
cjllanwarne: Not that there might not be a better way, but it wasn't just a crazy accident

cjllanwarne: An alternative scheme I thought about was passing in a `TokenDistributingActor` to the `FileHashingActor` as part of the props. The idea being, you can't start working until you get a `WorkToken` and you'd return in when you're done.

I moved away from that because the built-in `Router`s just seemed like a neater fit.

geoffjentry: On the chance that the ticket I'm going to mention isn't what sparked this, there's #1377

kshakir: It's definitely related, imho. I'll wire in a singleton cache `Map[Factory, Pool]` into the engine side for the moment. But yeah, these pools-of-actors-per-factory needs "love".

kshakir:
wdl4s and lenthall updated. Fixed test failures and re-singletoned the factory hashing pools.

mcovarr:
:+1:

kshakir:
Will merge end of day or after the metadata migration merges. Whatever comes first.

PR #1381, opened 2016-09-08
Look for failure mode option in the right place in the config.  Closes #1380

### LINKED ISSUE #1380: ContinueWhilePossible not working via Config
User @gavvvr  reports:

If i run cromwell with -Dworkflow-options.workflow-failure-mode="ContinueWhilePossible" it does not work. Also using -Dconfig.file=application.conf does not work. Only workflow_failure_mode option in JSON config works.

mcovarr:


Horneth:
👍

geoffjentry:
👍 

Can you leave a note on the issue when this is closed so they know it's fixed?

PR #1382, opened 2016-09-08
Do not become catatonic on abort.

mcovarr:


engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -143,8 +133,7 @@ class EngineJobExecutionActor(replyTo: ActorRef,
   }
 
   when(BackendIsCopyingCachedOutputs) {
-    // It's possible that cache writing was turned off due to a `HashError`, hence the `writeToCache` guard.
-    case Event(response: SucceededResponse, data @ ResponsePendingData(_, _, Some(Success(hashes)))) if effectiveCallCachingMode.writeToCache =>
+    case Event(response: SucceededResponse, data @ ResponsePendingData(_, _, Some(Success(hashes)))) =>
```
cjllanwarne: Did something change to make the guard no longer needed?

cjllanwarne:
Assuming the deguarding is justified, 👍

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
mcovarr: I think if there are hashes in the state data then the hashing must have fully completed successfully and cache writing wouldn't have been turned off.

kshakir:
👍

PR #1385, opened 2016-09-09
Optionalize metadata summarization.  Closes #1378.

### LINKED ISSUE #1378: Disable metadata summary refresh via config
As a user, I would like to be able to disable the metadata summary refresh.  I want to do this because I'd like to be able to stand up a "read only" cromwell that that doesn't write to the database because I can use these to scale out horizontally for read api operations (e.g. status, metadata).  Scaling out these operations is important because it allows my cromwell to remain responsive even under heavy read load.

@geoffjentry -- pls expand/correct

mcovarr:


Horneth:
👍

kshakir:
👍

core/src/main/resources/application.conf
```diff
@@ -84,6 +84,10 @@ system {
 
   // Since the WorkflowLogCopyRouter is initialized in code, this is the number of workers
   number-of-workflow-log-copy-workers = 10
+
+  // Whether to run the metadata summarizer in this Cromwell instance.  There should only be one metadata summarizer
+  // per database, so if this Cromwell instance is being used in read-only mode this value should be `false`.
```
cjllanwarne: Do we want to have a `read-only-cromwell` option which implies this?

cjllanwarne: Although tbh if we have `Cromwell Cluster`s (tasty and nutritious!) I'd nominate the Metadata Reader Cromwell as the summariser so maybe not

mcovarr: I was thinking that myself, but this ticket specifically asked for something much narrower.  @kcibul @geoffjentry ?

geoffjentry: I could see cases where people might want to have fine grained control, but I've also been advocating for giving fewer knobs to be turned. 

OTOH this isn't *truly* a 'read only' mode being discussed because they'll still want to submit workflows. It's more of a "don't actually run anything" mode and the metadata summarization is a side effect we need to take care of

geoffjentry: @mcovarr I had a few lines saying why I thought it was a great idea until I thought of the thing after "OTOH". If the right nomenclature could be found to capture what it's really doing and get over my pedantic streak I'd be for that myself

cjllanwarne: Possible Cromwell roles:
* Executor
* Metadata Reader
* Database Maintainer (i.e. summariser)
* Submitter

kcibul: I'd rather see a few general purposes switches that we can figure out the
right way to take advantage of at this time, than specialized
roles/behaviors that we're not sure if they're useful.

the special flag here is ok -- I had mentioned in the ticket just reusing
the refresh interval since intuitively setting that to infinity/zero says
to me that it won't happen and doesn't introduce a special flag... but
they're close enough in intent that i'm fine with either (esp since the
special flag is already done).



Kristian Cibulskis
kcibul@broadinstitute.org

mcovarr: @kcibul sorry, I didn't read the ticket thoroughly enough and missed the comment about reusing the refresh interval.  I'm happy to change this, I think that could actually simplify things nicely.

kcibul: If you think it's a good idea ;)

Kristian Cibulskis
kcibul@broadinstitute.org

kshakir:
ToL: `Inf` as `None` _feels_ weird, but I've done worse. 😉  Still LGTM

core/src/main/resources/application.conf
```diff
@@ -351,6 +351,8 @@ services {
   }
   MetadataService {
     class = "cromwell.services.metadata.impl.MetadataServiceActor"
+    // Set this value to "Inf" to turn off metadata summary refresh.
+    metadata-summary-refresh-interval = "Inf"
```
mcovarr: oops this should not be the default, I'll comment this line out 😦

Horneth: Maybe say here that the default is 2 seconds ?

Horneth:
re 👍  once not defaulting to infinity

core/src/main/resources/application.conf
geoffjentry: I do like the idea of specifying what the defaults are when they're buried in the scala code

geoffjentry:
@Horneth  
to_infinity_and_beyond -340698

geoffjentry:
Take *that* 🇬🇧 !

PR #1391, opened 2016-09-09
Renamed application.conf to reference.conf. Closes #1389

### LINKED ISSUE #1389: Move (most of?) application.conf to reference.conf
NOTE: I haven't 100% confirmed that the application.conf isn't auto-included, but since we now have separated cromwell into artifacts, based on the HOCON docs it does not appear that we're using an "application.conf" as intended: https://github.com/typesafehub/config/blob/master/HOCON.md#conventional-configuration-files-for-jvm-apps

If someone needs to specify a backend via a new application.conf, they shouldn't need to do one of the two current workarounds:

1) Know to include the _original_ application.conf.
2) Copy/paste all of the previous application.conf, including things like akka dispatcher executors.

The fix is to rename the core/src/main/resources/{application.conf => reference.conf}. When one then adjusts settings while running the fat-jar, one definitely doesn't need to re-include a reference.conf.

The ticket says "most of?" because there may be some elements of the current core/.../application.conf that would be more appropriate for a fat-jar _only_, such as having the main.hsqldb and an included Local backend.

kshakir:


geoffjentry:
@kshakir I don't *think* anyone builds cromwell jar directly into their app anymore (at least the constituents that we know about in DSDE) and I don't *think* it should matter, but it'd be good to check with our blue tinged friends.

geoffjentry:
Otherwise as long as it works 👍

kshakir:
Hm... seems like the workflow id parsing under centaurJes maybe isn't working. Will debug this later:

```
+EXIT_CODE=0
++grep 'SingleWorkflowRunnerActor: Workflow submitted ' log.txt
++perl -pe 's/\e\[?.*?[\@-~]//g'
++cut -f7 '-d '
+export WORKFLOW_ID=Workflow
+WORKFLOW_ID=Workflow
```

geoffjentry:
It *was* working but that doesn't mean it still is.

kshakir:
More googling to do. The assembly merge is resulting in the original akka conf overriding ours, and removing the custom akka logging values.

kshakir:
Issue resolved, ready for review. Overrides moved to application.conf. Defaults remain in reference.conf.

core/src/main/resources/reference.conf
```diff
@@ -370,20 +361,20 @@ database {
...
-        password = "pass"
-        connectionTimeout = 5000
-      }
-
-      migration {
-        read-batch-size = 100000
-        write-batch-size = 100000
-      }
-    }
+    //mysql_example {
```
geoffjentry: While you're in here, can you make the scheme for commenting out blocks uniform? For instance I see `#` above and `//` here

geoffjentry:
Sticking with my "as long as it works" stance :)

core/src/main/resources/reference.conf
```diff
@@ -0,0 +1,380 @@
...
+    #        host = "127.0.0.1"
+    #        port = 27017
+    #        name = "htcondor"
+    #        collection = "cache"
+    #      }
+    #    }
+    #
+    #    filesystems {
+    #      local {
+    #        # Cromwell makes a link to your input files within <root>/<workflow UUID>/workflow-inputs
```
Horneth: I imagine this is copy / paste and wasn't updated in a while, but this is no longer true, we don't copy to `<root>/<workflow UUID>/workflow-inputs` anymore, but rather each input in its call directory.

Horneth:
👍 

[Approved with PullApprove](https://pullapprove.com/broadinstitute/cromwell/pull-request/1391/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=broadinstitute/cromwell)

core/src/main/resources/reference.conf
kshakir: Thanks. Will remove the comment.

PR #1392, opened 2016-09-09
Use subshell and temporary rc file Closes #1383

### LINKED ISSUE #1383: Local/SGE backend may read the (wrong) rc before it's finished writing
The SharedFileSystem (SFS) backend that is the basis for the local and SGE backends currently has a race condition in the script:

```bash
#!/bin/sh
cd $cwd
$instantiatedCommand
echo $$? > rc
```

1. The `echo` needs to write to a `rc.tmp` and then use the atomic `mv rc.tmp rc`. Otherwise cromwell will (very very rarely?) pickup the existence of the file before bash has written, flushed, and closed the rc contents. This is not an issue on JES because the API is waiting for the script to exit, not the appearance of the `rc` file.
2. The `rc` path should be absolute, whether inside or outside of docker. `echo $$? > $cwd/rc` may be enough? Not an issue on JES as it only writes to the dockerized rc path.
3. The SFS (and probably JES for consistency, and backends in general?) should run the WDL command in a subshell, either with a (yet-another) call to bash, or using [bashisms](https://github.com/koalaman/shellcheck/wiki/SC2103#correct-code). This would better protect that the WDL from killing the current shell without writing an `rc`.

A proper centaur test would contain _diabolical_ WDL such as:

```
command {
exit
}
```

```
command {
mkdir newdir
cd newdir
}
```

Horneth:


mcovarr:
:+1:

kshakir:
Centaur should also be updated run the two tests in #1383 before closing the ticket.

kshakir:
👍  For this PR, but not for closing the ticket until this is actually tested.

Horneth:
Centaur: https://github.com/broadinstitute/centaur/pull/105

PR #1403, opened 2016-09-12
Bring abort to HtCondor. Closes #1402

### LINKED ISSUE #1402: Add abort support in HtCondor
HtCondor backend should be responsive to abort requests from the engine, and kill and remove the job from it's internal queue.

gauravs90:
As a side effect to enable abort support in HtCondor, this PR makes the polling (for checking job status) asynchronous, and the polling interval to be configurable.

### DIFF
--- a/supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActor.scala
+++ b/supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActor.scala
@@ -214,12 +214,12 @@ class HtCondorJobExecutionActor(override val jobDescriptor: BackendJobDescriptor
     log.debug("{} Process complete. RC file now exists with value: {}", tag, jobReturnCode)
 
     jobReturnCode match {
-      case Success(rc) if rc == extProcess.IncompleteRcIdentifier =>
+      case Success(None) =>
         import scala.concurrent.duration._
         // Job is still running in HtCondor. Check back again after `pollingInterval` seconds
         context.system.scheduler.scheduleOnce(pollingInterval.seconds, self, TrackTaskStatus(jobIdentifier))
-      case Success(rc) if rc == 0 | runtimeAttributes.continueOnReturnCode.continueFor(rc) => self ! JobExecutionResponse(processSuccess(rc))
-      case Success(rc) => self ! JobExecutionResponse(FailedNonRetryableResponse(jobDescriptor.key,
+      case Success(Some(rc)) if rc == 0 | runtimeAttributes.continueOnReturnCode.continueFor(rc) => self ! JobExecutionResponse(processSuccess(rc))
+      case Success(Some(rc)) => self ! JobExecutionResponse(FailedNonRetryableResponse(jobDescriptor.key,
         new IllegalStateException("Job exited with invalid return code: " + rc), Option(rc)))
       case Failure(error) => self ! JobExecutionResponse(FailedNonRetryableResponse(jobDescriptor.key, error, None))
     }
--- a/supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActor.scala
+++ b/supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActor.scala
@@ -165,7 +165,6 @@ class HtCondorJobExecutionActor(override val jobDescriptor: BackendJobDescriptor
       else
         log.error("{}: Failed to kill / remove job {}. Exit Code: {}, Stderr: {}", tag, id, exitVal, abortProcess.processStderr)
     }
-    self ! JobExecutionResponse(AbortedResponse(jobDescriptor.key))
   }
 
   private def executeTask(): Unit = {
@@ -218,7 +217,7 @@ class HtCondorJobExecutionActor(override val jobDescriptor: BackendJobDescriptor
         import scala.concurrent.duration._
         // Job is still running in HtCondor. Check back again after `pollingInterval` seconds
         context.system.scheduler.scheduleOnce(pollingInterval.seconds, self, TrackTaskStatus(jobIdentifier))
-      case Success(Some(rc)) if rc == 0 | runtimeAttributes.continueOnReturnCode.continueFor(rc) => self ! JobExecutionResponse(processSuccess(rc))
+      case Success(Some(rc)) if runtimeAttributes.continueOnReturnCode.continueFor(rc) => self ! JobExecutionResponse(processSuccess(rc))
       case Success(Some(rc)) => self ! JobExecutionResponse(FailedNonRetryableResponse(jobDescriptor.key,
         new IllegalStateException("Job exited with invalid return code: " + rc), Option(rc)))
       case Failure(error) => self ! JobExecutionResponse(FailedNonRetryableResponse(jobDescriptor.key, error, None))
--- a/supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActor.scala
+++ b/supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActor.scala
@@ -4,7 +4,7 @@ import java.nio.file.attribute.PosixFilePermission
 import java.nio.file.FileSystems
 
 import akka.actor.{ActorRef, Props}
-import cromwell.backend.BackendJobExecutionActor.{BackendJobExecutionResponse, FailedNonRetryableResponse, SucceededResponse}
+import cromwell.backend.BackendJobExecutionActor.{AbortedResponse, BackendJobExecutionResponse, FailedNonRetryableResponse, SucceededResponse}
 import cromwell.backend._
 import cromwell.backend.impl.htcondor.caching.CacheActor._
 import cromwell.backend.io.JobPaths
@@ -41,6 +41,8 @@ class HtCondorJobExecutionActor(override val jobDescriptor: BackendJobDescriptor
   import cromwell.core.PathFactory._
 
   private val tag = s"CondorJobExecutionActor-${jobDescriptor.call.fullyQualifiedName}:"
+
+  implicit val executionContext = context.dispatcher
 
   lazy val cmds = new HtCondorCommands
   lazy val extProcess = new HtCondorProcess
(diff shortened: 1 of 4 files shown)


gauravs90:
@mcovarr / @cjllanwarne / @francares

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActor.scala
```diff
@@ -136,9 +152,22 @@ class HtCondorJobExecutionActor(override val jobDescriptor: BackendJobDescriptor
...
+  override def abort(): Unit = {
+    // Nothing to do in case `condorJobId` is not defined
+    condorJobId foreach { id =>
+      log.info("{}: Aborting job [{}:{}].", tag, jobDescriptor.key.tag, id)
+      val abortProcess = new HtCondorProcess
+      val argv = Seq(HtCondorCommands.Remove, id)
+      val process = abortProcess.externalProcess(argv)
+      val exitVal = process.exitValue()
+      val msgMap = Map(0 -> s"Job $id successfully killed and removed from the queue.",
+        1 -> s"Failed to kill / remove job $id: ${abortProcess.processStderr}")
```
mcovarr: `if`

mcovarr:
:+1: you'll need to rebase on develop to get your builds to go green (zone issue)

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActor.scala
```diff
@@ -88,25 +90,39 @@ class HtCondorJobExecutionActor(override val jobDescriptor: BackendJobDescriptor
 
   private val executionResponse = Promise[BackendJobExecutionResponse]()
 
-  override def receive = super.receive orElse {
-    case ExecutionResultFound(succeededResponse) => executionResponse success succeededResponse.copy(jobKey = jobDescriptor.key)
+  // Message sent (by self, to self) wrapping over the response produced by HtCondor
+  private case class JobExecutionResponse(resp: BackendJobExecutionResponse)
```
geoffjentry: Make case classes final

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActor.scala
```diff
@@ -88,25 +90,39 @@ class HtCondorJobExecutionActor(override val jobDescriptor: BackendJobDescriptor
...
 
-    case ExecutionResultNotFound => prepareAndExecute()
+  // Message sent (by self, to self) to trigger a status check to HtCondor
+  private case class TrackTaskStatus(id: String)
 
-    case ExecutionResultStored(hash) => log.debug("{} Cache entry was stored for Job with hash {}.", tag, hash)
+  private var condorJobId: Option[String] = _
```
geoffjentry: Why `_` and not `None`?

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorWrapper.scala
```diff
@@ -8,20 +8,19 @@ import cromwell.backend.impl.htcondor
...
-    val MapOfstatuses = Map(
-      0 -> Created, // This is actually `unexpanded` in HtCondor, not sure what that actually means
-      1 -> Created, // Idle
-      2 -> Running,
-      3 -> Removed,
-      4 -> Completed,
-      5 -> Failed, // SystemOnHold
-      6 -> SubmissionError // Also the default
-    )
+  val MapOfstatuses = Map(
```
cjllanwarne: Preexisting typo: capital S needed

cjllanwarne:
Just thinking out loud - there seems to be a lot of duplication between this backend and the configurable shared-filesystem backend. 
I don't know whether you've tried expressing this as a configuration file backend?

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActor.scala
```diff
@@ -136,9 +152,23 @@ class HtCondorJobExecutionActor(override val jobDescriptor: BackendJobDescriptor
...
     */
-  override def abort(): Unit = throw new UnsupportedOperationException("HtCondorBackend currently doesn't support aborting jobs.")
+  override def abort(): Unit = {
+    // Nothing to do in case `condorJobId` is not defined
+    condorJobId foreach { id =>
+      log.info("{}: Aborting job [{}:{}].", tag, jobDescriptor.key.tag, id)
+      val abortProcess = new HtCondorProcess
+      val argv = Seq(HtCondorCommands.Remove, id)
+      val process = abortProcess.externalProcess(argv)
+      val exitVal = process.exitValue()
```
cjllanwarne: Ideally the process which is killing the task wouldn't be synchronous inside the receive block

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorWrapper.scala
```diff
@@ -99,23 +101,17 @@ class HtCondorProcess extends StrictLogging {
...
-        case status if JobStatus.isTerminal(status) =>
-          Files.exists(returnCodeFilePath) match {
-            case true => File(returnCodeFilePath).contentAsString.stripLineEnd.toInt
-            case false =>
-              val msg = s"JobStatus from Condor is terminal ($status) and no RC file exists!"
-              logger.debug(msg)
-              throw new IllegalStateException(msg)
-          }
-        case nonTerminalStatus =>
-          Thread.sleep(5000)
```
cjllanwarne: The docs for this method are now out-of-date

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorWrapper.scala
```diff
@@ -99,23 +101,17 @@ class HtCondorProcess extends StrictLogging {
...
+    checkStatus(jobId) match {
+      case status if JobStatus.isTerminal(status) =>
+        Files.exists(returnCodeFilePath) match {
+          case true => File(returnCodeFilePath).contentAsString.stripLineEnd.toInt
+          case false =>
+            val msg = s"JobStatus from Condor is terminal ($status) and no RC file exists!"
+            logger.debug(msg)
+            throw new IllegalStateException(msg)
+        }
+      case nonTerminalStatus => IncompleteRcIdentifier
```
cjllanwarne: Why not just make the return type of the function an `Option[Int]` so that you can return `None` here and handle this case more officially where it gets called from

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActor.scala
```diff
@@ -160,43 +190,47 @@ class HtCondorJobExecutionActor(override val jobDescriptor: BackendJobDescriptor
...
     jobReturnCode match {
-      case Success(rc) if rc == 0 | runtimeAttributes.continueOnReturnCode.continueFor(rc) => processSuccess(rc)
-      case Success(rc) => FailedNonRetryableResponse(jobDescriptor.key,
-        new IllegalStateException("Job exited with invalid return code: " + rc), Option(rc))
-      case Failure(error) => FailedNonRetryableResponse(jobDescriptor.key, error, None)
+      case Success(rc) if rc == extProcess.IncompleteRcIdentifier =>
+        import scala.concurrent.duration._
+        // Job is still running in HtCondor. Check back again after `pollingInterval` seconds
+        context.system.scheduler.scheduleOnce(pollingInterval.seconds, self, TrackTaskStatus(jobIdentifier))
+      case Success(rc) if rc == 0 | runtimeAttributes.continueOnReturnCode.continueFor(rc) => self ! JobExecutionResponse(processSuccess(rc))
```
cjllanwarne: I don't think you want the `if rc ==0` - just in case `continueOnReturnCode` explicitly disables 0 return codes

cjllanwarne: Also is that a binary OR `|` !?

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActor.scala
gauravs90: Didn't get you. It shouldn't? Are you saying this should be wrapped in a Future block to be async?
(I wanted it to be synchronous to adhere to the return type expected by this method)

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActor.scala
gauravs90: Or a non-short circuit OR. Changed it. Thanks! :)
rc == 0 is a normal check that says the job exited normally, I think we do want it there. (`continueOnReturnCode = false` would still mean 0 is a valid RC, right?)

gauravs90:
@cjllanwarne Things in this backend long existed before sfs backend came into being, and we didn't look into it yet. Good point though, I think we might try to adapt this to the sfs backend some time in the future.

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActor.scala
```diff
@@ -160,43 +190,47 @@ class HtCondorJobExecutionActor(override val jobDescriptor: BackendJobDescriptor
...
     jobReturnCode match {
-      case Success(rc) if rc == 0 | runtimeAttributes.continueOnReturnCode.continueFor(rc) => processSuccess(rc)
-      case Success(rc) => FailedNonRetryableResponse(jobDescriptor.key,
-        new IllegalStateException("Job exited with invalid return code: " + rc), Option(rc))
-      case Failure(error) => FailedNonRetryableResponse(jobDescriptor.key, error, None)
+      case Success(None) =>
+        import scala.concurrent.duration._
+        // Job is still running in HtCondor. Check back again after `pollingInterval` seconds
+        context.system.scheduler.scheduleOnce(pollingInterval.seconds, self, TrackTaskStatus(jobIdentifier))
+      case Success(Some(rc)) if rc == 0 | runtimeAttributes.continueOnReturnCode.continueFor(rc) => self ! JobExecutionResponse(processSuccess(rc))
```
cjllanwarne: The other backends allow a WDL author to say "0 is an invalid return code for this task".

TBH this feels more like it should be a centralised engine decision rather than a backend decision but that's where we are: it's up to you but just be aware that you'll have different semantics from the other backends.

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActor.scala
cjllanwarne: IIRC the semantics of `abort()` are:
* This will trigger an abort of the underlying code but not do anything synchronously in the method
* Nothing is returned and no reply messages are sent out directly by this method
* When the `abort()` causes the underlying job to stop, the BJEA will notice this as part of its ongoing monitoring of the job and can then reply to its parent with the `AbortedResponse`

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActor.scala
```diff
@@ -136,9 +152,23 @@ class HtCondorJobExecutionActor(override val jobDescriptor: BackendJobDescriptor
...
+      val abortProcess = new HtCondorProcess
+      val argv = Seq(HtCondorCommands.Remove, id)
+      val process = abortProcess.externalProcess(argv)
+      val exitVal = process.exitValue()
+      if (exitVal == 0)
+        log.info("{}: Job {} successfully killed and removed from the queue.", tag, id)
+      else
+        log.error("{}: Failed to kill / remove job {}. Exit Code: {}, Stderr: {}", tag, id, exitVal, abortProcess.processStderr)
+    }
+    self ! JobExecutionResponse(AbortedResponse(jobDescriptor.key))
```
cjllanwarne: So I don't think you want to send the `AbortedResponse` here, I think you want your usual monitoring to detect that the job has been aborted and at that time reply with the `AbortedResponse`

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActor.scala
gauravs90: I couldn't find out how an author can specify that. Could you point me to it?
Either way, that check (for rc == 0 is redundant, since that function (`continueFor`) seems to be now checking for rc == 0 by default  anyway, I'll get rid of that)
(PS: Was looking at the readme we have [here](https://github.com/broadinstitute/cromwell#continueonreturncode))

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActor.scala
gauravs90: I think I still don't understand what is expected here. (I do get the semantics you explained above, and have removed the `AbortedResponse` message that I was sending directly from here before)

The BackendJobExecutionActor has:
```scala
case AbortJobCommand =>
      abort()
      context.parent ! AbortedResponse(jobDescriptor.key)
      context.stop(self)
```
So it expects `abort()` to be called, which when completes, it sends the `AbortedResponse` to the parent.  For Jes and SFS I think we are overirding the default behavior (Jes is extending AsyncBackendJobExecutionActor while SFS overrides it with it's own `receive` definition), which allows for the 3rd point you mentioned above. Currently for this backend (and by default), there is just an `abort()` method which the contract enforces to implement, synchronously (being devoid of any Future[Unit] or expecting any completion msg). Anything I'm missing?

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActor.scala
cjllanwarne: It was right there:
> When set to an integer, or an array of integers, only those integers will be considered as successful return codes.
```
runtime {
  continueOnReturnCode: 1
}
```
In the above example, 0 is not listed as a valid return code

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActor.scala
cjllanwarne: Hmm, that's interesting. I'd leave as-is for now. I think @mcovarr is looking at how abort is/should/might work since it looks like it's only partially-wired through elsewhere

cjllanwarne:
👍 from me

PR #1404, opened 2016-09-13
switch default gce zone to us-central1-b

Horneth:


supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesRuntimeAttributes.scala
```diff
@@ -45,7 +45,7 @@ object JesRuntimeAttributes {
   private val MemoryDefaultValue = "2 GB"
 
   val ZonesKey = "zones"
-  private val ZoneDefaultValue = "us-central1-a"
+  private val ZoneDefaultValue = "us-central1-b"
```
Horneth: This is the actual change, the rest is just so people that copy/paste the readme don't use a either and updating the tests

PR #1407, opened 2016-09-13
Added a better set of EJEA tests. Closes 1373.

cjllanwarne:
Testing the FSM state-by-state, testing what happens when transitions happen and then making assumptions about the starting conditions of subsequent states.


engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -166,18 +172,18 @@ class EngineJobExecutionActor(replyTo: ActorRef,
 
     case Event(HashError(t), data: SucceededResponseData) =>
       disableCallCaching(t)
-      saveJobCompletionToJobStore(data)
+      saveJobCompletionToJobStore(data.copy(hashes = Some(Failure(t))))
     case Event(HashError(t), data: ResponsePendingData) =>
       disableCallCaching(t)
-      stay
+      stay using data.copy(hashes = Some(Failure(t)))
```
kshakir: `Some.apply` should be `Option.appy`. Same comment goes for other usages of `Some.apply`.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -204,7 +210,7 @@ class EngineJobExecutionActor(replyTo: ActorRef,
 
   whenUnhandled {
     case Event(msg, _) =>
-      log.error(s"Bad message to EngineJobExecutionActor in state $stateName(with data $stateData): $msg")
+      log.error(s"Bad message from ${sender.toString} to EngineJobExecutionActor in state $stateName(with data $stateData): $msg")
```
kshakir: `from $sender`

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -282,23 +271,34 @@ class EngineJobExecutionActor(replyTo: ActorRef,
     *
     * @return a RoundRobinPool of actors based on this.factory.
     */
-  private def fileContentsHasherActor: ActorRef = {
+  private[execution] lazy val fileContentsHasherActor: ActorRef = {
```
kshakir: Revert this. It's busted as the actors will die when the EJEA dies. I've fixed it in #1411.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -403,27 +407,23 @@ object EngineJobExecutionActor {
...
-
-  implicit class EnhancedNotSuccess(val response: BackendJobExecutionResponse) extends AnyVal {
-    def updatedData(data: ResponsePendingData): NotSucceededResponseData = {
-      NotSucceededResponseData(response, data.hashes)
-    }
-  }
-
   private[execution] sealed trait EJEAData {
-    override def toString = getClass.getSimpleName
+    //override def toString = getClass.getSimpleName
```
kshakir: Don't leave behind dead code. Remove the whole block.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/JobPreparationActor.scala
```diff
@@ -20,76 +22,83 @@ case class JobPreparationActor(executionData: WorkflowExecutionActorData,
...
 
   override def receive = {
     case Start =>
-      val response = resolveAndEvaluateInputs(jobKey, expressionLanguageFunctions) flatMap { prepareJobExecutionActor } match {
-        case Success(m) => m
-        case Failure(f) => BackendJobPreparationFailed(jobKey, f)
-      }
+      val response = resolveAndEvaluateInputs(jobKey, expressionLanguageFunctions) map { prepareJobExecutionActor } recover {
+        case f => BackendJobPreparationFailed(jobKey, f)
+      } andThen { case _ => self ! PoisonPill }
```
kshakir: Is there a good reason to use the `PoisonPill` instead of `context stop self` used most other places?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/JobPreparationActor.scala
```diff
@@ -20,76 +22,83 @@ case class JobPreparationActor(executionData: WorkflowExecutionActorData,
...
-      evaluated match {
-        case Some(value) =>
-          val coercedValue = value flatMap declaration.wdlType.coerceRawValue
-          inputs + ((name, coercedValue))
-        case None => inputs
-      }
-    })
-
-    TryUtil.sequenceMap(inputEvaluationAttempt, s"Input evaluation for Call ${call.fullyQualifiedName} failed")
+    Future {
```
kshakir: Why is this actor now running yet another `Future` inside the akka `receive` (that is already inside a `Future` itself)? If conceptually this method should create a second `Future`, then why return a `Future[Try[A]]` instead of just a flattened `Future[A]`.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -130,32 +131,37 @@ class EngineJobExecutionActor(replyTo: ActorRef,
     case Event(HashError(t), data: ResponsePendingData) =>
       disableCacheWrite(t)
       // Can't write hashes for this job, but continue to wait for the lookup response.
-      stay
+      stay using data.copy(hashes = Some(Failure(t)))
```
mcovarr: Option

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -130,32 +131,37 @@ class EngineJobExecutionActor(replyTo: ActorRef,
...
     case Event(HashError(t), data: SucceededResponseData) =>
-      disableCallCaching(t)
-      saveJobCompletionToJobStore(data)
+      disableCacheWrite(t)
+      saveJobCompletionToJobStore(data.copy(hashes = Some(Failure(t))))
     case Event(HashError(t), data: ResponsePendingData) =>
       disableCacheWrite(t)
       // Can't write hashes for this job, but continue to wait for the copy response.
-      stay
+      stay using data.copy(hashes = Some(Failure(t)))
```
mcovarr: Option

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -166,18 +172,18 @@ class EngineJobExecutionActor(replyTo: ActorRef,
 
     case Event(HashError(t), data: SucceededResponseData) =>
       disableCallCaching(t)
-      saveJobCompletionToJobStore(data)
+      saveJobCompletionToJobStore(data.copy(hashes = Some(Failure(t))))
```
mcovarr: Option

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
mcovarr: `{}`ify

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -130,32 +131,37 @@ class EngineJobExecutionActor(replyTo: ActorRef,
...
       saveCacheResults(hashes, data)
     case Event(hashes: CallCacheHashes, data: ResponsePendingData) =>
       addHashesAndStay(data, hashes)
+
+    // Hash error occurs:
     case Event(HashError(t), data: SucceededResponseData) =>
-      disableCallCaching(t)
-      saveJobCompletionToJobStore(data)
+      disableCacheWrite(t)
+      saveJobCompletionToJobStore(data.copy(hashes = Some(Failure(t))))
```
mcovarr: Option

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -36,17 +36,20 @@ class EngineJobExecutionActor(replyTo: ActorRef,
...
+  private var backendCacheHitCopyingActorProps: Props = _
 
-  // There's no need to check for a cache hit again if we got preempted
+  // There's no need to check for a cache hit again if we got preempted, or if there's no result copying actor defined
   // NB: this can also change (e.g. if we have a HashError we just force this to CallCachingOff)
-  private var effectiveCallCachingMode = if (jobKey.attempt > 1) callCachingMode.withoutRead else callCachingMode
+  private var effectiveCallCachingMode = if (factory.cacheHitCopyingActorProps.isEmpty || jobDescriptorKey.attempt > 1) callCachingMode.withoutRead else callCachingMode
+
+  // For tests:
+  private[execution] def checkEffectiveCallCachingMode = effectiveCallCachingMode
```
mcovarr: why not just `private[execution]` the variable?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -238,33 +244,16 @@ class EngineJobExecutionActor(replyTo: ActorRef,
     writeToMetadata(Map(effectiveCallCachingKey -> effectiveCallCachingMode.toString))
   }
 
+  def createJobPreparationActor(jobPrepProps: Props, name: String): ActorRef = context.actorOf(jobPrepProps, name)
```
mcovarr: inline or at least move this inside `prepareJob`

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -282,23 +271,34 @@ class EngineJobExecutionActor(replyTo: ActorRef,
...
-    goto(BackendIsCopyingCachedOutputs)
+    factory.cacheHitCopyingActorProps match {
+      case Some(propsMaker) =>
+        val backendCacheHitCopyingActorProps = propsMaker(data.jobDescriptor, initializationData, serviceRegistryActor)
+        val cacheHitCopyActor = context.actorOf(backendCacheHitCopyingActorProps, buildCacheHitCopyingActorName(data.jobDescriptor))
+        cacheHitCopyActor ! CopyOutputsCommand(cachedJobOutputs)
+        goto(BackendIsCopyingCachedOutputs)
+      case None =>
+        // This should be impossible with the FSM, but luckily, we CAN recover if this happens
+        val errorMessage = "Call caching copying impossible with no copy actor props! (Programmer error!)"
```
mcovarr: The wording might be more clear: if `factory.cacheHitCopyingActorProps` is empty call cache reading should have been turned off and it shouldn't be possible to get here

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
cjllanwarne: I felt uneasy doing that to a var

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/JobPreparationActor.scala
cjllanwarne: I'll have to double check but I think there was still some work to do in the receive method. So I didn't want to abruptly stop it but rather make it stop the next time it's in a stable state

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
cjllanwarne: I agree with this case. A lot of the other `Some`s are in `should be(Some(_))`  and I intend to keep those as-is

cjllanwarne: s/agree/won't bother arguing this time/g

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/ejea/EjeaBackendIsCopyingCachedOutputsSpec.scala
```diff
@@ -0,0 +1,136 @@
...
+import scala.util.{Failure, Success, Try}
+
+class EjeaBackendIsCopyingCachedOutputsSpec extends EngineJobExecutionActorSpec with HasJobSuccessResponse with HasJobFailureResponses with CanExpectJobStoreWrites with CanExpectCacheWrites {
+
+  override implicit val stateUnderTest = BackendIsCopyingCachedOutputs
+
+  "An EJEA in FetchingCachedOutputsFromDatabase state" should {
+
+    val hashErrorCause = new Exception("blah")
+    val hashResultsDataValue = Some(Success(callCacheHashes))
```
mcovarr: please `Option` all of these

cjllanwarne: This is used as an expectation. I actively expect it to be `Some`.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
cjllanwarne: I'll rebase when your branch merges

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
cjllanwarne: Oops, that wasn't meant to be permanent!

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/JobPreparationActor.scala
cjllanwarne: I generally have an aversion to blocking the receive. E.g. If we wired up abort to go to this actor we can't see it until we finish working.

But, (shrug) I'll do it sync on receive.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
cjllanwarne: The whole point was so we could mock this out :)

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -423,7 +410,15 @@ object EngineJobExecutionActor {
 
   private[execution] case class ResponsePendingData(jobDescriptor: BackendJobDescriptor,
                                                     bjeaProps: Props,
-                                                    hashes: Option[Try[CallCacheHashes]] = None) extends EJEAData
+                                                    hashes: Option[Try[CallCacheHashes]] = None) extends EJEAData {
+
+    def withResponse(success: SucceededResponse) = SucceededResponseData(success, hashes)
+
+    def withResponse(response: BackendJobExecutionResponse) = response match {
+      case success: SucceededResponse => SucceededResponseData(success, hashes)
```
mcovarr: I don't think this first case can match, the more derived version above would hit first

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/ejea/EjeaBackendIsCopyingCachedOutputsSpec.scala
```diff
@@ -0,0 +1,136 @@
...
+      }
+    }
+  }
+
+  def standardResponsePendingData(hashes: Option[Try[CallCacheHashes]]) = ResponsePendingData(helper.backendJobDescriptor, helper.bjeaProps, hashes)
+  def ejeaInBackendIsCopyingCachedOutputsState(initialHashes: Option[Try[CallCacheHashes]], callCachingMode: CallCachingMode, restarting: Boolean = false) = helper.buildEJEA(restarting = restarting, callCachingMode = callCachingMode).setStateInline(data = standardResponsePendingData(initialHashes))
+
+}
+
+object EjeaBackendIsCopyingCachedOutputsSpec {
```
mcovarr: vacuous

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
cjllanwarne: I shall let the tests decide whether that's safe!

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/ejea/PerTestHelper.scala
```diff
@@ -0,0 +1,145 @@
...
+                             serviceRegistryActor: ActorRef,
+                             jobStoreActor: ActorRef,
+                             callCacheReadActor: ActorRef,
+                             backendName: String,
+                             callCachingMode: CallCachingMode) extends EngineJobExecutionActor(replyTo, jobDescriptorKey, executionData, factory, initializationData, restarting, serviceRegistryActor, jobStoreActor, callCacheReadActor, backendName, callCachingMode) {
+
+
+  override def makeFetchCachedResultsActor(cacheHit: CacheHit, taskOutputs: Seq[TaskOutput]) = helper.fetchCachedResultsActorCreations :+= (cacheHit, taskOutputs)
+  override def initializeJobHashing(jobDescriptor: BackendJobDescriptor, activity: CallCachingActivity) = helper.jobHashingInitializations :+= (jobDescriptor, activity)
+  override def createSaveCacheResultsActor(hashes: CallCacheHashes, success: SucceededResponse) = helper.callCacheWriteActorCreations :+= (hashes, success)
```
mcovarr: Are these call cache write actor creations conceptually an `Option` rather than a `Seq`?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
cjllanwarne: Turns out this IS needed, but is kinda confusing. I'll rename it so that its obvious which places are calling which method

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/ejea/EjeaBackendIsCopyingCachedOutputsSpec.scala
```diff
@@ -0,0 +1,131 @@
...
+          }
+
+          if (cacheUpdateRequired) {
+            expectCacheWrite(successResponse, finalHashData.get.get)
+          } else {
+            expectJobStoreWrite(SucceededResponseData(successResponse, finalHashData))
+          }
+          // A separate check of the final effective call caching mode:
+          if (ejhaResponse == hashErrorEjhaResponse || initialHashData == hashErrorDataValue) {
+            ejea.underlyingActor.checkEffectiveCallCachingMode should be(mode.withoutWrite)
```
mcovarr: might be a little more readable if it went straight to the `var`... :smile:

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/ejea/EngineJobExecutionActorSpec.scala
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  override def afterAll(): Unit = {
+    system.terminate()
+    super.afterAll()
+  }
+
+  // Some helper lists
+  val CallCachingModes = List(CallCachingOff, CallCachingActivity(ReadCache), CallCachingActivity(WriteCache), CallCachingActivity(ReadAndWriteCache))
+  val RestartOrExecuteCommandTuples = List(("execute", false, BackendJobExecutionActor.ExecuteJobCommand), ("restart", true, BackendJobExecutionActor.RecoverJobCommand) )
```
mcovarr: can haz proper type?

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/ejea/EngineJobExecutionActorSpecUtil.scala
```diff
@@ -0,0 +1,72 @@
...
+        validateJobStoreKey(jobKey)
+        returnCode should be(successRc)
+        jobOutputs should be(successOutputs)
+        ejea.stateName should be(UpdatingJobStore)
+        ejea.stateData should be(expectedData)
+    }
+  }
+}
+
+private[ejea] trait CanExpectHashingIitialization extends Eventually { self: EngineJobExecutionActorSpec =>
```
mcovarr: typo in the name

supportedBackends/spark/src/test/scala/cromwell/backend/impl/spark/SparkJobExecutionActorSpec.scala
```diff
@@ -14,6 +14,7 @@ import cromwell.core.{PathWriter, TailedWriter, TestKitSuite, UntailedWriter, _}
 import org.mockito.Matchers._
 import org.mockito.Mockito
 import org.mockito.Mockito._
+import org.scalatest.mockito.MockitoSugar
```
mcovarr: already imported

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/ejea/EjeaFetchingCachedOutputsFromDatabaseSpec.scala
```diff
@@ -0,0 +1,74 @@
...
+        // Send the response from the "Fetch" actor
+        val cachedResults = successResponse.jobOutputs
+        ejea ! CachedOutputLookupSucceeded(cachedResults, CacheHit(MetaInfoId(75)))
+        helper.callCacheHitCopyingProbe.expectMsg(CopyOutputsCommand(cachedResults))
+
+        // Check we end up in the right state:
+        ejea.stateName should be(BackendIsCopyingCachedOutputs)
+        // Check we end up with the right data:
+        val expectedData = initialData.copy(hashes = ejhaResponse map {
+          case `callCacheHashes` => Success(callCacheHashes)
```
kshakir: ToL: What do you think about finding an object (say `object HasJobSuccessResponse`) to house and uppercase your [constants](http://docs.scala-lang.org/style/naming-conventions.html#constants-values-variable-and-methods), and remove the backtick workaround?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/JobPreparationActor.scala
kshakir: ToL: While the jpa would "see" the abort message sent to receive, I believe the jpa wouldn't/couldn't terminate the runnable block already scheduled via `Future.apply(runnable)(ec)`, as is, since the runnable would be queued up in the ec without a way to kill it. Alternatively, if we _don't_ queue up a runnable on an ec, and use `context stop jpaRef`, then the actor should stop after it finishes the current `receive`, in theory... I haven't fully thought this through and would be interested to hear others thoughts sometime, maybe in a separate ticket.

Blocking the `receive` when an actor has other messages to "simultaneously" process doesn't feel bad yet, but I've been wrong many times before. Everything I've read says that long running tasks should be at bottom of your actor hierarchy, mainly so that if it fails you can restart it, but also with the benefit that the work isn't blocking any important messaging. But- the akka docs seem to hint that if one gets tired of actors, to [just use a Future](http://doc.akka.io/docs/akka/2.4/scala/futures.html#Use_Directly). I'm still looking for best practices with relation to blocking `receive` though.

I'm not sure at the moment what a proper "Abort, stop prepping the job please!" looks like. I did skim these three threads [[1]](https://groups.google.com/d/topic/scala-user/XfT3zsectQI/discussion) [[2]](https://groups.google.com/d/topic/akka-user/TlIbfaC1eb8/discussion) [[3]](https://groups.google.com/d/topic/akka-user/nkD5BN17kVk/discussion), that made it clear that aborting long running akka receives / scala futures should use a hand-coded circuit breaker.

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/ejea/EjeaBackendIsCopyingCachedOutputsSpec.scala
cjllanwarne: Did you say this after you saw my response? If you're that sure, I'll change it...

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/ejea/PerTestHelper.scala
cjllanwarne: Hmm, probably. That'd make it easier to check we didn't accidentally go x2 on actor creation too. I'll change it

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
kshakir: Live by the `var`, die by the `var`. The def by itself adds no protection whatsoever unless you were to start synchronizing.

cjllanwarne: Wait, how do you mean? If it's a def you can't update the var from outside the actor?

kshakir: EDIT: Whoops. Chris was right, I was wrong. The `private[execution] def` allows that tests can only read, not write, the underlying `var`. I'm voting it stays as is.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/JobPreparationActor.scala
```diff
@@ -11,6 +11,7 @@ import wdl4s.expression.WdlStandardLibraryFunctions
 import wdl4s.util.TryUtil
 import wdl4s.values.WdlValue
 
+import scala.concurrent.ExecutionContext
```
kshakir: This class isn't spawning separate `Future`s, so this import and the implicit below shouldn't be needed.

kshakir:
One `ExecutionContext` to be removed, a rebase, then 👍

mcovarr:
:+1:

PR #1411, opened 2016-09-13
Move BLAF.FHA back to child of AS instead of under EJEA.

kshakir:


engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -269,27 +269,15 @@ class EngineJobExecutionActor(replyTo: ActorRef,
       self,
       jobDescriptor,
       initializationData,
-      fileContentsHasherActor,
+      /*
+      Use context.system instead of context as the factory. Otherwise when we die, so will the child actors.
+       */
+      fileContentsHasherSingletonActor(backendName, factory, context.system),
```
cjllanwarne: Can we somehow sneakily use the named `cromwell-root` actor to be the parent? @geoffjentry?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -444,4 +432,22 @@ object EngineJobExecutionActor {
     * More refinement may appear via #1377.
     */
   private val factorySingletonRouters = scala.collection.mutable.Map[BackendLifecycleActorFactory, ActorRef]()
+
```
cjllanwarne: Think I'd prefer `var` of immutable rather than `val` of mutable

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
geoffjentry: Discussed in person, answer is no :'(

cjllanwarne:
Otherwise 👍

geoffjentry:
Why do I have a feeling that you're being a bit rebellious with those acronyms?

geoffjentry:
I do think that in practice this is unlikely to matter for a while and this isn't the long term right solution IMO.

That said I also see the logic as to why it was the way it was (even if it was in an incorrect location) 

tl;dr I don't think any of this is a huge deal at the moment so i'm cool with whatever 👍

kshakir:
ToL: I'm fine if these actors hang as children off the initialization actor or some other _actor_, but not as singletons-in-the-system-jvm-referenced-from-the-non-actor-blaf. Relatedly, the name blaf is a slight misnomer, at it actually creates props, not actor refs, so maybe it should be the blpf? Pinging #1377 to record some of this as food for later thought.

kshakir:
Ready for re-review if anyone wants to take a look. Will merge in a couple hours.

backend/src/main/scala/cromwell/backend/BackendLifecycleActorFactory.scala
```diff
@@ -42,7 +42,7 @@ trait BackendLifecycleActorFactory {
   def runtimeAttributeDefinitions(initializationDataOption: Option[BackendInitializationData]): Set[RuntimeAttributeDefinition] = Set.empty
 
   lazy val fileHashingFunction: Option[FileHashingFunction] = None
-  lazy val fileHashingWorkerCount: Int = 50
+  lazy val fileHashingActorCount: Int = 50
 
-  def fileContentsHasherActorProps: Props = FileHasherWorkerActor.props(fileHashingFunction)
+  def fileHashingActorProps: Props = FileHashingActor.props(fileHashingFunction)
```
cjllanwarne: Info only: 
Oh I remember why `Contents` was in there. I was preparing for other types of file hash (e.g. path, companion `.md5`, etc). I approve of this change

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -269,27 +269,15 @@ class EngineJobExecutionActor(replyTo: ActorRef,
       self,
       jobDescriptor,
       initializationData,
-      fileContentsHasherActor,
+      /*
+      Use context.system instead of context as the factory. Otherwise when we die, so will the child actors.
+       */
```
cjllanwarne: Mid-argument-list multi-line comments FTL :(

geoffjentry:
:+1:

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -443,5 +431,54 @@ object EngineJobExecutionActor {
...
+    *
+    * @param backendName                  Name of the backend.
+    * @param backendLifecycleActorFactory A backend factory.
+    * @param actorRefFactory              An actor factory.
+    * @return a RoundRobinPool of actors based on backend factory.
+    */
+  private def factoryFileHashingRouter(backendName: String,
+                                       backendLifecycleActorFactory: BackendLifecycleActorFactory,
+                                       actorRefFactory: ActorRefFactory): ActorRef = {
+    val (originalOrUpdated, result) = getOrElseUpdated(
```
cjllanwarne: ToL: I applaud the FP pureness of this... but couldn't you just:
```
if (!x.contains(y) { x :+= y -> ... }
x(y)
```

cjllanwarne:
Don't let me nit-picking block you: re-👍

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
kshakir: That does an extra lookup. This is an `var immutable.Map` version of the `val mutable.Map.getOrElseUpdate` without the second pass for lookups.

See http://stackoverflow.com/questions/4385976/idiomatic-get-or-else-update-for-immutable-map#answer-5840119

PR #1418, opened 2016-09-14
Cromwell 1308 Closes #1308 Closes #1321

### LINKED ISSUE #1308: SFS globbing matches too widely
If an input file matches a glob string, it is also included in the glob output.

This is undesirable

### LINKED ISSUE #1321: Glob grabs incorrect input file (for local and sge backend)
The attached wdl and json file have an issue with the second call in the pipeline. It is scattered over the bam files output by the first call (RevertSam). The problem is that shard-5 grabs the wrong bam file to input into the second call (FirstSortSam). All of the other shards grab the correct file except for shard-5 which grabs the bam file that is the input to the RevertSam task (rather than its output). The workaround to fix this was to change the glob command to be more specific: if I change it from *.bam to *.unmapped.bam, then shard-5 is correct.

[BrokenShard.wdl.txt](https://github.com/broadinstitute/cromwell/files/432898/BrokenShard.wdl.txt)
[testBam.json.txt](https://github.com/broadinstitute/cromwell/files/432900/testBam.json.txt)

Note the example files that are attached need access to the Broad file system to run.

Horneth:
Centaur: https://github.com/broadinstitute/centaur/pull/106

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,38 +46,3 @@ passed absolute paths for input `File`s.
 * On the SFS backends, the call directory now contains two sub-directories:
     * `inputs` contains all the input files that have been localized for this task (see next below for more details)
     * `execution` contains all other files (script, logs, rc, potential outputs etc...)
-    
-* On the SFS backends, inputs are now localized as follow:
-    The absolute path of the directory containing the input is hashed.
-    This hash is used to create a directory under `inputs` where the file will be localized
-    e.g:
-    
-    ```
-    ├── execution
-    │   ├── rc
-    │   ├── script
-    │   ├── script.background
-    │   ├── script.submit
-    │   ├── stderr
-    │   ├── stderr.background
-    │   ├── stdout
-    │   └── stdout.background
-    └── inputs
-        ├── 2c142eed1d519eabc0520becff443327
-        │   ├── file3.txt
-        │   └── file4.txt
-        └── 63a9f0ea7bb98050796b649e85481845
-            ├── file1.txt
-            └── file2.txt
-    ```
-    
-    where the file*.txt input files path structure would be similar too:
-    
-    ```
-    root
-    ├── file1.txt       -> will be localized under hash("root")/file1.txt
-    ├── file2.txt       -> will be localized under hash("root")/file2.txt
-    └── mydir
-        ├── file3.txt   -> will be localized under hash("root/mydir")/file3.txt
-        └── file4.txt   -> will be localized under hash("root/mydir")/file4.txt
-    ```
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,3 +42,42 @@ to work with various other command line dispatchers such as LSF. See the
 
 * On the JES and SFS backends, task `command` blocks are now always
 passed absolute paths for input `File`s.
+
+* On the SFS backends, the call directory now contains two sub-directories:
+    * `inputs` contains all the input files that have been localized for this task (see next below for more details)
+    * `execution` contains all other files (script, logs, rc, potential outputs etc...)
+    
+* On the SFS backends, inputs are now localized as follow:
+    The absolute path of the directory containing the input is hashed.
+    This hash is used to create a directory under `inputs` where the file will be localized
+    e.g:
+    
+    ```
+    ├── execution
+    │   ├── rc
+    │   ├── script
+    │   ├── script.background
+    │   ├── script.submit
+    │   ├── stderr
+    │   ├── stderr.background
+    │   ├── stdout
+    │   └── stdout.background
+    └── inputs
+        ├── 2c142eed1d519eabc0520becff443327
+        │   ├── file3.txt
+        │   └── file4.txt
+        └── 63a9f0ea7bb98050796b649e85481845
+            ├── file1.txt
+            └── file2.txt
+    ```
+    
+    where the file*.txt input files path structure would be similar too:
+    
+    ```
+    root
+    ├── file1.txt       -> will be localized under hash("root")/file1.txt
+    ├── file2.txt       -> will be localized under hash("root")/file2.txt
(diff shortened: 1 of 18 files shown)


supportedBackends/htcondor/src/test/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActorSpec.scala
```diff
@@ -386,9 +388,9 @@ class HtCondorJobExecutionActorSpec extends TestKitSuite("HtCondorJobExecutionAc
     val bashScript = Source.fromFile(jobPaths.script.toFile).getLines.mkString
 
     assert(bashScript.contains("docker run -w /workingDir -v"))
-    assert(bashScript.contains(tempDir1.toAbsolutePath.toString))
-    assert(bashScript.contains(tempDir2.toAbsolutePath.toString))
-    assert(bashScript.contains("/call-hello:/outputDir --rm ubuntu/latest echo"))
+    assert(bashScript.contains(tempDir1.toAbsolutePath.toString.md5Sum))
```
cjllanwarne: Oh, ha, I made a new one of these implicits for the EngineJobFileHasher, `EngineJobHashingActor.StringMd5er`... I wonder if we can find a better place for them all

CHANGELOG.md
```diff
@@ -42,3 +42,42 @@ to work with various other command line dispatchers such as LSF. See the
...
+    │   ├── rc
+    │   ├── script
+    │   ├── script.background
+    │   ├── script.submit
+    │   ├── stderr
+    │   ├── stderr.background
+    │   ├── stdout
+    │   └── stdout.background
+    └── inputs
+        ├── 2c142eed1d519eabc0520becff443327
```
cjllanwarne: Probably worth a tech talk and/or PO input on the new structure. I can imagine this could be confusing for any spelunking users... @kcibul any thoughts?

cjllanwarne:
Uh, looks like 👍 doesn't work in "review summary" comments... so: 👍 again

CHANGELOG.md
Horneth: Yeah I put it in there to see what people think and planned to bring it up at stand up. It personally annoys me to have this nesting so I wanted to give it a try.

cjllanwarne: FWIW I suspect the repeated nesting of directories was also relatively difficult to spelunk(sp?)

core/src/main/scala/cromwell/core/PathFactory.scala
```diff
@@ -4,20 +4,38 @@ import java.io.Writer
...
 import scala.collection.immutable.Queue
-import scala.util.{Success, Try}
+import scala.util.{Success, Failure, Try}
+
+class FileSystemNotFound(str: String) extends CromwellFatalException(
+  new IllegalArgumentException(s"Could not find suitable filesystem to parse $str")
+)
 
 trait PathFactory {
+  private val schemeMatcher = """([a-z]*://).*""".r
```
mcovarr: you probably want `+` and not `*` for at least the first quantifier, maybe the second

core/src/main/scala/cromwell/core/PathFactory.scala
```diff
@@ -4,20 +4,38 @@ import java.io.Writer
...
+          Failure(new IllegalArgumentException(s"$rawString scheme doesn't match ${fs.provider.getScheme}"))
+        } else {
+          Try(fs.getPath(rawString))
+        }
     })
   }
+
+  private def hasWrongScheme(rawString: String, fileSystem: FileSystem): Boolean = {
+    schemeMatcher.findFirstMatchIn(rawString) match {
+      case Some(m) if m.groupCount == 1 => m.group(1) != fileSystem.provider().getScheme
```
mcovarr: the `groupCount` check seems unnecessary, there wouldn't be a match without that group being defined

supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystem.scala
```diff
@@ -184,8 +185,9 @@ trait SharedFileSystem extends PathFactory {
     }
 
     def adjustFile(path: String) = {
-      val (src, dst) = toDestPath(path)
-      localize(src, dst) map { Unit => WdlFile(dst.toString) }
+      toDestPath(path) flatMap {
+        case (src, dst) => localize(src, dst) map { Unit => WdlFile(dst.toString) }
```
mcovarr: `_` rather than `Unit`

geoffjentry:
I agree w/ the discussion that this requires broader discussion

mcovarr:
:+1:

supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystem.scala
geoffjentry: Also it's `()` not `Unit`, we've run into issues w/ this before

Horneth:
@geoffjentry All right I'll hold off merging this until we get @kcibul's (also gotc / firecloud ?) opinion.

CHANGELOG.md
kcibul: seems reasonable to me.  The only thing would be to not use hashes in the inputs subdirectory, and use some other form of name-shortening (like take the last element of the path plus an incrementing number after 1 like when you download the same file in chrome twice).  

Generally though, I think people don't spelunk as much on inputs as they do on outputs, so this seems fine

Horneth: Just seeing this now for some reason.. The main constraint I think (from an execution standpoint) is that we want input files that originally are in the same directory to end up in the same directory again once localized. I remember having bugs like this before. Let me try with chrome-style naming see how it looks

PR #1422, opened 2016-09-15
Reduce quantity of scary logging (and don't rely on scary logging in the tests) Closes 1311, Closes 1422

cjllanwarne:


engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -309,14 +309,16 @@ class WorkflowActor(val workflowId: WorkflowId,
     goto(finalState) using data.copy(currentLifecycleStateActor = None)
   }
 
+  private[workflow] def makeFinalizationActor(workflowDescriptor: EngineWorkflowDescriptor, executionStore: ExecutionStore, outputStore: OutputStore) = {
+    context.actorOf(WorkflowFinalizationActor.props(workflowId, workflowDescriptor, executionStore, outputStore, stateData.initializationData), name = s"WorkflowFinalizationActor-$workflowId")
```
mcovarr: will the workflow ID not already be in the this actor's name "path"?

engine/src/test/scala/cromwell/CromwellTestkitSpec.scala
```diff
@@ -325,63 +325,26 @@ abstract class CromwellTestkitSpec(val twms: TestWorkflowManagerSystem = new Cro
...
+    eventually { verifyWorkflowState(rootActor.underlyingActor.serviceRegistryActor, workflowId, terminalState) }
+    val outcome = getWorkflowOutputsFromMetadata(workflowId, rootActor.underlyingActor.serviceRegistryActor)
     system.stop(rootActor)
-    id
+    // And return the outcome:
+    outcome
   }
 
-  def eventually[T](isFatal: Throwable => Boolean, maxRetries: Int = 3)(f: => T): T = {
+  def cromwellTestkitSpecEventually[T](isFatal: Throwable => Boolean, maxRetries: Int = 3)(f: => T): T = {
```
mcovarr: perhaps a name that describes what it does rather than where it is?

mcovarr:
:+1:

engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -136,10 +136,10 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef, val wor
 
   onTransition {
     case oldState -> terminalState if terminalState.terminal =>
-      workflowLogger.info(s"transition from $oldState to $terminalState: shutting down")
+      workflowLogger.debug(s"transition from $oldState to $terminalState: shutting down MWDA")
```
kshakir: The name of the actor should be in the message already.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -404,10 +404,10 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId,
 
   onTransition {
     case fromState -> toState if toState.terminal =>
-      workflowLogger.info(s"$tag transitioning from $fromState to $toState. Shutting down.")
+      workflowLogger.debug(s"$tag transitioning from $fromState to $toState. Shutting down WorkflowExecutionActor.")
```
kshakir: The name of the actor should be in the message already.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -420,7 +420,7 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId,
 
   onTransition {
     case _ -> toState if toState.terminal =>
-      workflowLogger.info(s"$tag done. Shutting down.")
+      workflowLogger.debug(s"$tag done. Shutting down WorkflowExecutionActor.")
```
kshakir: The name of the actor should be in the message already.

kshakir:
Kicking pull approve: 👍

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
cjllanwarne: Dunno. This is a copy/paste from what it was before. Not too keen on doing half of a "rationalise actor naming" ticket in this PR

engine/src/test/scala/cromwell/CromwellTestkitSpec.scala
cjllanwarne: Hmm, I had to rename this to avoid a naming collision with `Eventually`... 
I guess I could give it a synonym, `tryUntilTestPasses` or something...

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
cjllanwarne: Right, but the message was still scary... "why is the WorkflowExecutionActor shutting down Cromwell!?!"

Maybe there's a middle ground? `shutting down this actor`?

kshakir: This has moved to the `debug` level, so normal users won't see it to be scared. They'll only turn on debugging, when requested, to send us a log so that we can debug a bug. In that case we can see the actor name already.

Parallel ToL: IMHO the "shutting down" is the scary part. Maybe comms team can come up better terminology if needed. I've seen them smooth over all sorts of UX issues.

cjllanwarne: Good call. I'll try to find a better form of words (without the actor name). And since this is now debug (hooray!), I won't worry too much about prettiness

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
cjllanwarne: OTOH I guess I might as well change this one since I'm already here...

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
geoffjentry: "giving a poison pill to actor X, i want it to die, perhaps go to the dungeon"?

geoffjentry:
👍 for less forced verbosity

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
cjllanwarne: Done. Github not hiding comments :(

engine/src/test/scala/cromwell/CromwellTestkitSpec.scala
cjllanwarne: Fixed by removing the method entirely

engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
cjllanwarne: Done.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
cjllanwarne: Done

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
cjllanwarne: Done

geoffjentry:
👍

PR #1424, opened 2016-09-15
[HtCondor] Create symlink after cache hit. Closes #1425.

### LINKED ISSUE #1425: [HtCondor] Create symlink after cache hit
Based on new security/permissions requirements there is a need to add extra functionality to HtCondor backend.

Basically the functionality should provide a way to link cached file / array of files outputs to the current workflow execution. All paths should point to current workflow execution dir.

francares:
Functionality added to localize WdlFile / WdlArray[WdlFile] cached results in HtCondor so in that way cached results are link to the new task.

i.e.:
1. w0/task1 execution produces w0/result1.
2. w0/result1 is stored in the cache.
3. w1/task1 (w1/task1 is the same as w0/task1) is executed again.
4. Cache result is hit.
5. File cached results are used to generate new symlinks to point to them in current task.
6. Result is generated using new paths based on symlinks and/or other cached results.

This change is based on new security/permissions requirements.

### DIFF
--- a/supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/caching/localization/CachedResultLocalization.scala
+++ b/supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/caching/localization/CachedResultLocalization.scala
@@ -1,0 +1,39 @@
+package cromwell.backend.impl.htcondor.caching.localization
+
+import java.nio.file.{Files, Path, Paths}
+
+import better.files.File
+import cromwell.core.{JobOutput, _}
+import wdl4s.types.{WdlArrayType, WdlFileType}
+import wdl4s.values.{WdlArray, WdlSingleFile, WdlValue}
+
+trait CachedResultLocalization {
+  private[localization] def localizePathViaSymbolicLink(originalPath: Path, executionPath: Path): Path = {
+    if (File(originalPath).isDirectory) throw new UnsupportedOperationException("Cannot localize directory with symbolic links.")
+    else {
+      File(executionPath).parent.createDirectories()
+      Files.createSymbolicLink(executionPath, originalPath.toAbsolutePath)
+    }
+  }
+
+  private[localization] def localizeCachedFile(executionPath: Path, output: WdlValue): WdlSingleFile = {
+    val origPath = Paths.get(output.valueString)
+    val newPath = executionPath.toAbsolutePath.resolve(origPath.getFileName)
+    val slPath = localizePathViaSymbolicLink(origPath, newPath)
+    WdlSingleFile(slPath.toString)
+  }
+
+  def localizeCachedOutputs(executionPath: Path, outputs: JobOutputs): JobOutputs = {
+    outputs map { case (lqn, jobOutput) =>
+      jobOutput.wdlValue.wdlType match {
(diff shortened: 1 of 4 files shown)


supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/caching/localization/CachedResultLocalization.scala
```diff
@@ -0,0 +1,43 @@
...
+  def localizeCachedFile(executionPath: Path, output: WdlValue): WdlSingleFile = {
+    val origPath = Paths.get(output.valueString)
+    val newPath = executionPath.toAbsolutePath.resolve(origPath.getFileName)
+    val slPath = localizePathViaSymbolicLink(origPath, newPath)
+    WdlSingleFile(slPath.toString)
+  }
+
+  def localizeCachedOutputs(executionPath: Path, outputs: JobOutputs): Try[JobOutputs] = {
+    Try {
+      outputs.map { output =>
```
mcovarr: `case` here to avoid the `_.1` and `_.2` below

mcovarr: also no `.` with the curly variant of HOFs (and you'll want to keep the curly variant to be able to `case` 😄 )

mcovarr:
:+1:

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/caching/localization/CachedResultLocalization.scala
```diff
@@ -0,0 +1,43 @@
...
+
+import better.files.File
+import cromwell.core.{JobOutput, _}
+import wdl4s.types.{WdlArrayType, WdlFileType}
+import wdl4s.values.{WdlArray, WdlSingleFile, WdlValue}
+
+import scala.util.Try
+
+trait CachedResultLocalization {
+  def localizePathViaSymbolicLink(originalPath: Path, executionPath: Path): Path = {
```
gauravs90: The two methods below could be `private` and not be exposed.

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/caching/localization/CachedResultLocalization.scala
```diff
@@ -0,0 +1,43 @@
...
+
+  def localizeCachedFile(executionPath: Path, output: WdlValue): WdlSingleFile = {
+    val origPath = Paths.get(output.valueString)
+    val newPath = executionPath.toAbsolutePath.resolve(origPath.getFileName)
+    val slPath = localizePathViaSymbolicLink(origPath, newPath)
+    WdlSingleFile(slPath.toString)
+  }
+
+  def localizeCachedOutputs(executionPath: Path, outputs: JobOutputs): Try[JobOutputs] = {
+    Try {
```
gauravs90: A pair of curly braces could be avoided by putting `Try` with the outer (function) block itself.

gauravs90:
👍  for me too.

francares:
I don't know why Github conversation screen is not being updated with new commits.

PR #1429, opened 2016-09-16
Avoid unnecessary token refreshing

Horneth:
Only the refresh-token mode requires new credentials to be created/validated for every workflow, all other modes can be validated only from the configuration. This ensures we don't re-create unnecessary auth modes.

### DIFF
--- a/supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/authentication/JesCredentials.scala
+++ b/supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/authentication/JesCredentials.scala
@@ -1,0 +1,5 @@
+package cromwell.backend.impl.jes.authentication
+
+import com.google.api.client.auth.oauth2.Credential
+
+case class JesCredentials(genomicsCredential: Credential, gcsCredential: Credential)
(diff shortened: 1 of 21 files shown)


geoffjentry:
👍

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/authentication/JesCredentials.scala
```diff
@@ -0,0 +1,5 @@
+package cromwell.backend.impl.jes.authentication
+
+import com.google.api.client.auth.oauth2.Credential
+
+case class JesCredentials(genomicsCredential: Credential, gcsCredential: Credential)
```
mcovarr: would `userAuthCredential` be more clear for the latter?

mcovarr:
hey pullapprove :+1:

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/authentication/JesCredentials.scala
Horneth: I called it gcs because it's the one coming from `filesystems.gcs.auth` in the config but if `userAuthCredential` makes it clearer I'm all for that !

cjllanwarne:
The test failure looks like it might be real: should parse all manner of well-formed auths *** FAILED *** (1 second, 53 milliseconds)

PR #1432, opened 2016-09-16
Add configurable docker command and soft links for dockerized jobs. Closes #1433.

### LINKED ISSUE #1433: Add configurable docker command and soft links for dockerized jobs
There is a need to support:
1. Configurable docker command in HtCondor.
2. Allow the use of soft-links in dockerized jobs.

francares:
Minor changes in order to support:
1. Configurable docker command in HtCondor.
2. Allow the use of soft-links in dockerized jobs.

### DIFF
--- a/supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActor.scala
+++ b/supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActor.scala
@@ -318,7 +318,7 @@ class HtCondorJobExecutionActor(override val jobDescriptor: BackendJobDescriptor
       }.flatten.toSeq
 
       log.debug("{} List of input volumes: {}", tag, dockerInputDataVol.mkString(","))
-      val dockerCmd = "docker run -w %s %s %s --rm %s %s"
+      val dockerCmd = configurationDescriptor.backendConfig.getString("docker.cmd")
       val dockerVolume = "-v %s:%s"
       val dockerVolumeInputs = s"$dockerVolume:ro"
       // `v.get` is safe below since we filtered the list earlier with only defined elements
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -226,6 +226,17 @@ backend {
     #    # launches.
     #    root: "cromwell-executions"
     #
+    #    #Placeholders:
+    #    #1. Working directory.
+    #    #2. Inputs volumes.
+    #    #3. Output volume.
+    #    #4. Docker image.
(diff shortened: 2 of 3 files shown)


supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActor.scala
```diff
@@ -269,7 +269,9 @@ class HtCondorJobExecutionActor(override val jobDescriptor: BackendJobDescriptor
       executionDir.toString.toFile.createIfNotExists(asDirectory = true, createParents = true)
 
       log.debug("{} Resolving job command", tag)
-      val command = localizeInputs(jobPaths.callRoot, runtimeAttributes.dockerImage.isDefined, fileSystems, jobDescriptor.inputs) flatMap {
+      val allowSoftLinks = configurationDescriptor.backendConfig.getBoolean("docker.allowSoftLinks")
+      val dockerLocEnabled = runtimeAttributes.dockerImage.isDefined && !allowSoftLinks
+      val command = localizeInputs(jobPaths.callRoot, dockerLocEnabled, fileSystems, jobDescriptor.inputs) flatMap {
```
cjllanwarne: It would be great if you changed the `localizeInputs` method's `docker` parameter to be called just `allowSoftLinks` since that's what it really does.

If I'm reading this right, you're adding a short cut so that you can force it to allow soft-links even in docker mode. Is that what you wanted? A quick google brings up all sorts of "here be dragons" style warnings for doing that

cjllanwarne:
👍 from me (I'm assuming you've read, digested, and feel comfortable with all the "here be dragons" warnings w.r.t. symlinks and docker)

supportedBackends/htcondor/src/main/scala/cromwell/backend/impl/htcondor/HtCondorJobExecutionActor.scala
francares: Yes, it will bring security concerns. That's why I'm discussing with some people in the CCC world why we should avoid use cases which require multiple file systems and big size files and they don't want these to be copied (that's why I added sl flag in dockerized jobs in the first place). I may remove this flag from the PR.

francares: By tomorrow I should have an answer if I need to proceed with this change or not. 
The idea is to merge configurable docker command and try to avoid the merge of allowSoftLinks.

francares:
@cjllanwarne reverted allowSoftLinks flag and keeping configurable docker cmd.

PR #1438, opened 2016-09-18
Add an endpoint w/ some basic workflow/job counts

geoffjentry:
edit: what was previously here is no longer true

### DIFF
--- a/engine/src/main/resources/swagger/cromwell.yaml
+++ b/engine/src/main/resources/swagger/cromwell.yaml
@@ -529,6 +529,26 @@ paths:
       security:
         - google_oauth:
             - openid
+  '/engine/{version}/stats':
+    get:
+      summary: Returns basic statistics from this Cromwell's engine.
+      parameters:
+        - name: version
+          description: API Version
+          required: true
+          type: string
+          in: path
+          default: v1
+      tags:
+        - Engine
+      responses:
+        '200':
+          description: Successful Request
+          schema:
+            $ref: '#/definitions/StatsResponse'
+      security:
+        - google_oauth:
+            - openid
 securityDefinitions:
   google_oauth:
     type: oauth2
@@ -753,3 +773,15 @@ definitions:
       defaultBackend:
         type: string
         description: The default backend of this server
+  StatsResponse:
+    description: 'Provides engine level statistics for things running inside the system'
+    required:
+      - workflows
+      - jobs
+    properties:
+      workflows:
+        type: integer
+        description: The number of currently running workflows
+      jobs:
+        type: integer
+        description: The number of currently running jobs
--- a/engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
@@ -20,6 +20,7 @@ import cromwell.engine.workflow.lifecycle.execution.WorkflowExecutionActor
 import cromwell.engine.workflow.lifecycle.execution.WorkflowExecutionActor._
 import cromwell.services.metadata.MetadataService._
 import cromwell.services.metadata.{MetadataEvent, MetadataKey, MetadataValue}
+import cromwell.webservice.EngineStatsActor
 
 import scala.language.postfixOps
 import scala.util.Random
@@ -220,6 +221,13 @@ class WorkflowActor(val workflowId: WorkflowId,
     case Event(WorkflowExecutionFailedResponse(executionStore, outputStore, failures),
     data @ WorkflowActorData(_, Some(workflowDescriptor), _, _)) =>
       finalizeWorkflow(data, workflowDescriptor, executionStore, outputStore, Option(failures.toList))
+    case Event(msg @ EngineStatsActor.JobCountQuery, data) =>
+      data.currentLifecycleStateActor match {
+        case Some(a) => a forward msg
+        case None => sender ! EngineStatsActor.NoJobs // This should be impossible, but if somehow here it's technically correct
+      }
+
+      stay()
   }
 
   when(FinalizingWorkflowState) {
@@ -249,6 +257,9 @@ class WorkflowActor(val workflowId: WorkflowId,
     case Event(AbortWorkflowCommand, WorkflowActorData(Some(actor), _, _, _)) =>
       actor ! EngineLifecycleActorAbortCommand
       goto(WorkflowAbortingState)
+    case Event(EngineStatsActor.JobCountQuery, _) =>
+      sender ! EngineStatsActor.NoJobs
+      stay()
     case unhandledMessage =>
       workflowLogger.warn(s"received an unhandled message $unhandledMessage in state $stateName")
       stay
--- a/engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
@@ -1,5 +1,7 @@
 package cromwell.engine.workflow
 
+import java.util.UUID
+
 import akka.actor.FSM.{CurrentState, SubscribeTransitionCallBack, Transition}
 import akka.actor._
 import akka.event.Logging
@@ -11,6 +13,7 @@ import cromwell.engine.workflow.WorkflowManagerActor._
 import cromwell.engine.workflow.workflowstore.{WorkflowStoreActor, WorkflowStoreState}
 import cromwell.jobstore.JobStoreActor.{JobStoreWriteFailure, JobStoreWriteSuccess, RegisterWorkflowCompleted}
 import cromwell.services.metadata.MetadataService._
+import cromwell.webservice.EngineStatsActor
 import lenthall.config.ScalaConfig.EnhancedScalaConfig
 
 import scala.concurrent.duration._
@@ -35,6 +38,7 @@ object WorkflowManagerActor {
   final case class AbortWorkflowCommand(id: WorkflowId, replyTo: ActorRef) extends WorkflowManagerActorCommand
   case object AbortAllWorkflowsCommand extends WorkflowManagerActorCommand
   final case class SubscribeToWorkflowCommand(id: WorkflowId) extends WorkflowManagerActorCommand
+  case object EngineStatsCommand extends WorkflowManagerActorCommand
 
   def props(workflowStore: ActorRef,
             serviceRegistryActor: ActorRef,
@@ -221,6 +225,10 @@ class WorkflowManagerActor(config: Config,
     case Event(JobStoreWriteFailure(t), _) =>
       log.error("Error writing to JobStore from WorkflowManagerActor: {}", t)
       // This is minorly bad. The JobStore would never rid itself of this workflow's entries. Unlikely to be a big deal
+      stay()
+    case Event(EngineStatsCommand, data) =>
+      val sndr = sender()
+      context.actorOf(EngineStatsActor.props(data.workflows.values.toList, sndr), s"EngineStatsActor-${sndr.hashCode()}")
       stay()
     // Anything else certainly IS interesting:
     case Event(unhandled, data) =>
--- a/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
@@ -24,6 +24,7 @@ import cromwell.engine.workflow.lifecycle.execution.WorkflowExecutionActor.Workf
 import cromwell.engine.{ContinueWhilePossible, EngineWorkflowDescriptor}
 import cromwell.services.metadata.MetadataService._
 import cromwell.services.metadata._
+import cromwell.webservice.EngineStatsActor
 import lenthall.exception.ThrowableAggregation
 import wdl4s.types.WdlArrayType
 import wdl4s.util.TryUtil
@@ -395,6 +396,9 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId,
       } else {
         goto(WorkflowExecutionAbortedState)
       }
+    case Event(EngineStatsActor.JobCountQuery, data) =>
+      sender ! EngineStatsActor.JobCount(data.backendJobExecutionActors.size)
+      stay()
     case unhandledMessage =>
       workflowLogger.warn(s"$tag received an unhandled message: ${unhandledMessage.event} in state: $stateName")
       stay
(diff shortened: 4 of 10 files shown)


geoffjentry:
Add at least some sort of testing

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -222,6 +226,10 @@ class WorkflowManagerActor(config: Config,
       log.error("Error writing to JobStore from WorkflowManagerActor: {}", t)
       // This is minorly bad. The JobStore would never rid itself of this workflow's entries. Unlikely to be a big deal
       stay()
+    case Event(EngineStatsCommand, data) =>
+      val sndr = sender()
+      context.actorOf(EngineStatsActor.props(data.workflows.values.toList, sndr), s"EngineStatsActor-${sndr.hashCode()}")
```
geoffjentry: Rewire this so the ESA requests the workflow actors from WMA and isn't created by the WMA

geoffjentry: Fitting this into our PerRequest structure ends up being just as hokey IMO. Leaving this as-is for now to see what others think

geoffjentry: new actor might be able to extend perRequest itself

engine/src/main/resources/swagger/cromwell.yaml
```diff
@@ -529,6 +529,26 @@ paths:
       security:
         - google_oauth:
             - openid
+  '/workflows/{version}/stats':
```
geoffjentry: perhaps `engine` instead of `workflows`?

engine/src/test/scala/cromwell/webservice/EngineStatsActorTest.scala
```diff
@@ -0,0 +1,60 @@
...
+import scala.concurrent.duration._
+import scala.language.postfixOps
+
+class EngineStatsActorTest extends TestKitSuite("EngineStatsActor") with FlatSpecLike with Matchers {
+  behavior of "EngineStatsActor"
+
+  val replyTo = TestProbe()
+  val defaultTimeout = 100 millis
+
+  it should "return snakeyes with no WorkflowActors" in {
```
cjllanwarne: I thought `snakeyes` was double-1s? :p

geoffjentry: You raise a good point but technically  we're talking about this: 
snake_eyes_classic

engine/src/test/scala/cromwell/webservice/EngineStatsActorTest.scala
```diff
@@ -0,0 +1,60 @@
...
+    TestActorRef(EngineStatsActor.props(List.empty[ActorRef], replyTo.ref))
+    replyTo.expectMsg(defaultTimeout, EngineStats(0, 0))
+  }
+
+  it should "return a single workflow with one job when that's the world it lives in" in {
```
cjllanwarne: In either case, I'm disappointed we didn't continue with https://en.wikipedia.org/wiki/List_of_British_bingo_nicknames

geoffjentry: 83 it is when i edit this

cjllanwarne: Glad I'm making a positive contribution to the project!

geoffjentry: Gotta do something in this world

cjllanwarne:
ToL: I'd prefer if this query was actioned read-side instead of execution-side. 

Thinking more, why not just use the Query endpoint? Or is this *designed* to be an actor debugging endpoint?

geoffjentry:
I want it more for the latter but it doesn't matter that much since it'll never be 100% accurate anyways.

The query isn't quite right but I suppose this could sit on top of and post process the query endpoint

cjllanwarne:
Hmm, or does akka have a count-by-name operation?

geoffjentry:
Not that I was able to find

geoffjentry:
In the spirit of "everything will be different in a month" I'm less concerned about some of the weird stuff I pointed out above and I realized it'd be good to get this functionality in 0.21 as some downstream constituents would like this behavior

engine/src/main/resources/swagger/cromwell.yaml
```diff
@@ -529,6 +529,26 @@ paths:
...
+      summary: Returns basic statistics from this Cromwell's engine.
+      parameters:
+        - name: version
+          description: API Version
+          required: true
+          type: string
+          in: path
+          default: v1
+      tags:
+        - Workflows
```
mcovarr: change this too?

engine/src/test/scala/cromwell/webservice/EngineStatsActorSpec.scala
```diff
@@ -0,0 +1,60 @@
...
+
+  it should "return an unemployed workflow when that's the world it lives in" in {
+    val workflowActors = List(Props(FakeWorkflowActor(0))) map { TestActorRef(_) }
+    TestActorRef(EngineStatsActor.props(workflowActors, replyTo.ref))
+    replyTo.expectMsg(defaultTimeout, EngineStats(1, 0))
+  }
+
+  it should "return in a timely manner if one of the WorkflowActors is quiet" in {
+    val workflowActors = List(Props(FakeWorkflowActor(0)), Props(new BlockingFakeWorkflowActor)) map { TestActorRef(_) }
+    TestActorRef(EngineStatsActor.props(workflowActors, replyTo.ref, 50 millis))
```
mcovarr: is the timing here not going to be flaky in sleepyhead Travis?

engine/src/test/scala/cromwell/webservice/EngineStatsActorSpec.scala
```diff
@@ -0,0 +1,60 @@
...
+  }
+}
+
+object EngineStatsActorSpec {
+  final case class FakeWorkflowActor(jobs: Int) extends Actor {
+    override def receive = {
+      case JobCountQuery => sender ! JobCount(jobs)
+    }
+  }
+  class BlockingFakeWorkflowActor extends Actor {
```
mcovarr: it's not so much blocking as ignoring

engine/src/main/resources/swagger/cromwell.yaml
```diff
@@ -753,3 +773,15 @@ definitions:
       defaultBackend:
         type: string
         description: The default backend of this server
+  StatsResponse:
+    description: ''
```
mcovarr: defies description?

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
mcovarr: hard to say without seeing the alleged hokiness, but I leave it to your discretion

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -252,6 +260,9 @@ class WorkflowActor(val workflowId: WorkflowId,
     case Event(AbortWorkflowCommand, WorkflowActorData(Some(actor), _, _, _)) =>
       actor ! EngineLifecycleActorAbortCommand
       goto(WorkflowAbortingState)
+    case Event(EngineStatsActor.JobCountQuery, _) =>
+      sender ! EngineStatsActor.NoJobs
+      stay()
```
mcovarr: so we say there are no jobs if we're aborting and not abort*ed*?

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -223,6 +224,13 @@ class WorkflowActor(val workflowId: WorkflowId,
     case Event(WorkflowExecutionFailedResponse(executionStore, outputStore, failures),
     data @ WorkflowActorData(_, Some(workflowDescriptor), _, _)) =>
       finalizeWorkflow(data, workflowDescriptor, executionStore, outputStore, Option(failures.toList))
+    case Event(msg @ EngineStatsActor.JobCountQuery, data) =>
```
cjllanwarne: ToL If you put this in `whenUnhandled` you could respond even if we're no longer in `executing` (and your "should be impossible" will be possible!)

mcovarr:
:+1:

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
cjllanwarne: ToL: I'm pretty sure you won't need the `sndr` here since it's not closing over it

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -397,6 +398,9 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId,
       } else {
         goto(WorkflowExecutionAbortedState)
       }
+    case Event(EngineStatsActor.JobCountQuery, data) =>
+      sender ! EngineStatsActor.JobCount(data.backendJobExecutionActors.size)
```
ruchim: Right now it seems like job count is the # of backendJobExecutionActors, but it's possible that if a user has caching on, they might want to know the # of jobs that are busy caching, otherwise their job count will appear 0 and give this false sense of no work being done. Is it possible to return the count of backend job execution as well as caching actors? though not sure if we have a list of caching actors anywhere right now? If this is outside the scope of the original ticket, we can file it for a future change.

ruchim:
I just have one comment that might be speculation for a change now or the future. Other than that, 👍

engine/src/test/scala/cromwell/webservice/EngineStatsActorSpec.scala
cjllanwarne: Pretty sure this is just `Props.empty`?

cjllanwarne:
👍

ruchim:
@geoffjentry the travis build is failing because some error handling has changed and so 2 refresh token centaur tests are failing--they should go green once you're rebased onto develop.

engine/src/test/scala/cromwell/webservice/EngineStatsActorSpec.scala
geoffjentry: :shrug emoji:

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
geoffjentry: Yeah it got kinda gnarly when I was playing around w/ it. Since I want to replace spray anyways I don't think that it's worth fighting our PerRequest structure

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
geoffjentry: I view this as a different statistic. I'll take a quick look to see if it's easy to grab that as well. If not I'll leave it for future expansion.

engine/src/main/resources/swagger/cromwell.yaml
geoffjentry: Heh the thing I c/p'd apparently was blank as well ;)

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
mcovarr: This would also address the concern in my "aborting versus aborted" comment below

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
geoffjentry: My thinking here was that it was going to be more prone to stuff disappearing, timeouts, etc. However I moved the main block to unhandled so we'll see how it goes

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
mcovarr: he doesn't *need* it, but otherwise he'd be invoking `sender()` twice

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
geoffjentry: Actually I think this might be a case where the value in the FSM data is a misnomer. I remember being a little confused by that the other day. it's actually an EJEA

engine/src/test/scala/cromwell/webservice/EngineStatsActorSpec.scala
mcovarr: or you could set `receive` to `Actor.emptyBehavior` if you're trying to work that "res ipsa loquitur"

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
geoffjentry: What @mcovarr said.

geoffjentry: Not that it matters much either way

engine/src/main/resources/swagger/cromwell.yaml
geoffjentry: This is actually kinda annoying in swagger IMO but I'm willing to accept that it's like the anger I have for a week after facebook changes its UI

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
geoffjentry: FWIW that also causes it to lock up. 

I'm probably not going to have time to dive into 'why' prior to people wanting to release although we'll see how that plays out.

I don't personally care too much about the aborting thing right now

cjllanwarne: Putting this in whenUnhandled causes it to lock up?? That's scary...

geoffjentry: I'm assuming it's something silly/stupid/obvious

PR #1442, opened 2016-09-19
Give FilePassingWorkflowSpec more time!

cjllanwarne:
Replacing `awaitCond` with `eventually` we should also get a better failure message than "timeout expired"


geoffjentry:
👍

Horneth:
👍

PR #1447, opened 2016-09-19
Update Cache Metadata message

ruchim:
Instead of adding the MetaInfoId as a part of the Cache Hit Metadata, add the source worklow id, call name and job index using the Fetch Cached Results Actor.


ruchim:
@cjllanwarne @mcovarr you guys are the lucky reviewers! Making the changes to publish cache hit metadata that includes Workflow ID, call name and job index of the source cache hit call.

cjllanwarne:
Looks like you're still based on @kshakir's PR?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/FetchCachedResultsActor.scala
```diff
@@ -32,7 +33,13 @@ class FetchCachedResultsActor(cacheHit: CacheHit, replyTo: ActorRef, callCache:
         val jobDetritusFiles = result.jobDetritus map { jobDetritusEntry =>
           jobDetritusEntry.detritusKey -> jobDetritusEntry.detritusValue
         }
-        replyTo ! CachedOutputLookupSucceeded(simpletons, jobDetritusFiles.toMap, result.returnCode, cacheHit)
+        val sourceCacheDetails = Seq(result.callCachingEntry.workflowExecutionUuid,
+                                    result.callCachingEntry.callFullyQualifiedName,
+                                    result.callCachingEntry.jobIndex).mkString(".")
```
cjllanwarne: I think it's more canonical to use `:` instead of `.` to avoid FQN confusions

cjllanwarne:
👍

cjllanwarne:
👍 (Assuming you're merging into the right branch!)

kshakir:
I'm (slowly) spinning up a cloud sql instance to sanity check that my branch can handle a gotc db migration in a single transaction.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCache.scala
```diff
@@ -86,7 +86,7 @@ class CallCache(database: CallCachingSqlDatabase) {
   private def cachedResultOption(callCachingJoinOption: Option[CallCachingJoin]): Option[CachedResult] = {
     callCachingJoinOption map { callCachingJoin =>
       CachedResult(
-        callCachingJoin.callCachingEntry.returnCode,
+        callCachingJoin.callCachingEntry,
```
kshakir: ToL: At this point, the classes `engine.….CachedResult` and `database.….CallCachingJoin` are _effectively_ the same. One part of me likes this separation, the other feels like the database class could just be passed to the actor as is, removing the need for this mapping function.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -117,8 +115,10 @@ class EngineJobExecutionActor(replyTo: ActorRef,
   }
 
   when(FetchingCachedOutputsFromDatabase) {
-    case Event(CachedOutputLookupSucceeded(wdlValueSimpletons, jobDetritus, returnCode, cacheHit), data: ResponsePendingData) =>
-      makeBackendCopyCacheHit(cacheHit, wdlValueSimpletons, jobDetritus, returnCode, data)
+    case Event(CachedOutputLookupSucceeded(wdlValueSimpletons, jobDetritus, returnCode, cacheResultId, cacheHitDetails), data: ResponsePendingData) =>
+      writeToMetadata(Map(callCachingReadResultMetadataKey -> s"Cache Hit: $cacheHitDetails"))
+      log.debug("Cache hit for {}! Fetching cached result {}", jobTag, cacheResultId)
```
kshakir: ToL: Are there any other `log.info()` messages to let the user know that call caching has ~actived~ hit? Otherwise, with a long running copy, I feel like we're super verbose about running, then _nothing_ when we cache hit.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/FetchCachedResultsActor.scala
ruchim: will change :)

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
ruchim: When I test caching for a wide scatter, I enable the log for printing during a cache miss--to be clear when caching failed on an expected hit. Either way--I prefer some kind of log printing as well.

mcovarr:
:+1:

ruchim:
@cjllanwarne purposefully making my branch off of Khalid's :)

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/ejea/EjeaFetchingCachedOutputsFromDatabaseSpec.scala
```diff
@@ -34,7 +35,8 @@ class EjeaFetchingCachedOutputsFromDatabaseSpec extends EngineJobExecutionActorS
         val cachedSimpletons = Seq(WdlValueSimpleton("a", WdlString("hullo")), WdlValueSimpleton("b", WdlString("cheerio")))
         val detritusMap = Map("stdout" -> "//somePath")
         val cachedReturnCode = Some(17)
-        ejea ! CachedOutputLookupSucceeded(cachedSimpletons, detritusMap, cachedReturnCode, CacheHit(MetaInfoId(75)))
+        val sourceCacheDetails = Seq(WorkflowId.randomId.toString,"call-someTask",1).mkString(":")
```
kshakir:
```scala
Seq(WorkflowId.randomId.toString,"call-someTask",1).mkString(":")
```

should be

```scala
s"${WorkflowId.randomId}:call-someTask:1"
```

kshakir:
Merge at will. I've left a comment regarding a test value, but it's not a must fix.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
geoffjentry: FWIW our log messages are even spammier than I realized. That's not to say that this one shouldn't hae more but *something* needs to have less

PR #1453, opened 2016-09-20
Add details in failure when failing to upload auth file for JES

Horneth:


supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesInitializationActor.scala
```diff
@@ -94,7 +96,9 @@ class JesInitializationActor(override val workflowDescriptor: BackendWorkflowDes
       val upload = () => Future(path.writeAsJson(content))
 
       workflowLogger.info(s"Creating authentication file for workflow ${workflowDescriptor.id} at \n ${path.toString}")
-      Retry.withRetry(upload, isFatal = isFatalJesException, isTransient = isTransientJesException)(context.system) map { _ => () }
+      Retry.withRetry(upload, isFatal = isFatalJesException, isTransient = isTransientJesException)(context.system) map { _ => () } recover {
+        case failure => throw new IOException("Failed to upload authentication file", failure)
```
cjllanwarne: I always get confused. Is this equivalent to 
```
  recoverWith {
    case failure => Future.failure(new IOException("Failed to upload authentication file", failure))
```
Except that recoverWith doesn't need to do the whole throw/catch dance?

cjllanwarne:
Otherwise looks good to me. 👍

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesInitializationActor.scala
Horneth: Yes it's equivalent, instead of throwing the exception you need to wrap it in a `Future` though, I usually don't use recoverWith unless the recovery work actually needs to be asynchronous but I honestly don't care if people prefer this way I'll change it, this is mainly so I can add a failure check on the centaur test.

geoffjentry: This got longer than I wanted, tl;dr either is ok by me.

I think it's basically just the difference between `map` and `flatMap`. Either is fine although `recoverWith` seems More Right. The `recover` thing above kinda reminds me of a pattern I used to use in scala which now I avoid like the plague, where `x` is an `Option`:

    val z = x.getOrElse(throw new Exception("foo"))

It gets the job done but as my love of typing grew it started to bug me.

geoffjentry:
👍 

[Approved with PullApprove](https://pullapprove.com/broadinstitute/cromwell/pull-request/1453/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=broadinstitute/cromwell)

PR #1456, opened 2016-09-20
Updates to workflow logger creation.

kshakir:
Overriding `WorkflowLogger.getName`.
Lookup loggers without synchronization. Create loggers with synchronization.

### DIFF
--- a/core/src/main/scala/cromwell/core/logging/WorkflowLogger.scala
+++ b/core/src/main/scala/cromwell/core/logging/WorkflowLogger.scala
@@ -4,7 +4,6 @@ import java.nio.file.{Path, Paths}
 
 import akka.actor.{Actor, ActorLogging}
 import akka.event.LoggingAdapter
-import better.files._
 import ch.qos.logback.classic.encoder.PatternLayoutEncoder
 import ch.qos.logback.classic.spi.ILoggingEvent
 import ch.qos.logback.classic.{Level, LoggerContext}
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cromwell/core/logging/WorkflowLogger.scala
```diff
@@ -22,12 +21,29 @@ trait WorkflowLogging extends ActorLogging { this: Actor =>
...
+  logback-core 1.1.6 started iterating over an internal hashmap, and now throws a ConcurrentModificationException
+  if two appenders call start() at the same time.
+
+  So now we synchronize makeFileLogger(). Could also try just synchronizing just the appender.start().
+
+  https://github.com/qos-ch/logback/commit/77128a003a7fd7e8bd7a6ddb12da7a65cf296593#diff-f8cd32379a53986c2e70e2abe86fa0faR145
+   */
+  private def makeSynchronizedFileLogger(path: Path, level: Level, ctx: LoggerContext, name: String): Logger =
+  synchronized {
+    Option(ctx.exists(name)) match {
```
Horneth: do we need to recheck for existence here since it's already done on line 29 in `makeFileLogger` ?

mcovarr: I think this needs to check again in the `synchronized` block to avoid a race?

Horneth: this checks against a logger cache which is implemented with a ConcurrentHashMap so I think it should be fine ? https://github.com/qos-ch/logback/blob/master/logback-classic/src/main/java/ch/qos/logback/classic/LoggerContext.java#L76

mcovarr: I thought the comments were saying that this map is no longer threadsafe in 1.1.6

core/src/main/scala/cromwell/core/logging/WorkflowLogger.scala
```diff
@@ -22,12 +21,29 @@ trait WorkflowLogging extends ActorLogging { this: Actor =>
...
       case Some(existingLogger) => existingLogger
+      case None => makeSynchronizedFileLogger(path, level, ctx, name)
+    }
+  }
+
+  /*
+  logback-core 1.1.6 started iterating over an internal hashmap, and now throws a ConcurrentModificationException
+  if two appenders call start() at the same time.
+
+  So now we synchronize makeFileLogger(). Could also try just synchronizing just the appender.start().
```
mcovarr: I think the comment needs a wee update to say `makeSynchronizedFileLogger`

core/src/main/scala/cromwell/core/logging/WorkflowLogger.scala
Horneth: It's not this one - they have a whole bunch of context maps and collision maps used in the `appender.start()` and those are normal `HashMaps`.

kshakir: While the inner map is the one that throws a concurrent modification exception, I wanted to avoid creating two loggers for the same name, like we used to. So to avoid that case this code now:

- checks the logger doesn't already exist
- waits for the mutex
- re-checks that the logger doesn't already exist
- creates the logger if needed
- releases the mutex

core/src/main/scala/cromwell/core/logging/WorkflowLogger.scala
kshakir: Will fix it during a rebase.

mcovarr:
:+1:

Horneth:
👍

PR #1468, opened 2016-09-21
Avoid hashing Scopes.  Closes #1457

### LINKED ISSUE #1457: WorkflowExecutionActor bogs down with needless Scope equals
There are at least four spots in `WorkflowExecutionActor` where `==` is invoked on `Scope`s.  This is not so good as `==` in Scala does a deep equals, and the `Scope` implementations are case classes with compiler-generated `equals` and `hashCode` implementations.  These implementations invoke `equals`/`hashCode` on their constructor fields, including `Task` which contains a lot of hairy stuff including ASTs.  This is extra bad in `WorkflowExecutionActor` since these `==`s are invoked synchronously with message processing to update the execution store, and all this computation is effectively single-threaded within a workflow.  In practice, for a 20K wide scatter this causes the `WorkflowExecutionActor` to spend upwards of 20 minutes from the time work finishes in (mock) JES to actually succeeding the workflow.

I *think* this might be fixed by chasing down all the spots where `==` is used on `Scope`s and replacing that with call FQN + index + attempt comparisons instead.

mcovarr:


engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -147,8 +145,7 @@ object WorkflowExecutionActor {
   }
 
   private implicit class EnhancedExecutionStore(val executionStore: ExecutionStore) extends AnyVal {
-    /** Find currently runnable scopes */
-    def runnableScopes: Iterable[JobKey] = executionStore.store.filter(isRunnable).keys
+    def runnableScopes = executionStore.store.toList collect { case entry if isRunnable(entry) => entry._1 }
```
cjllanwarne: Could you add a comment to avoid regression?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -437,7 +435,7 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId,
   }
 
   private def handleJobSuccessful(jobKey: JobKey, outputs: JobOutputs, data: WorkflowExecutionActorData) = {
-    workflowLogger.info(s"Job ${jobKey.tag} succeeded!")
+    workflowLogger.debug(s"Job ${jobKey.tag} succeeded!")
```
cjllanwarne: Ooh, this might be controversial?

If you're not going to log this, you'll have a lot of `Starting calls:...` and no closing `Job succeeded!`

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActorData.scala
```diff
@@ -43,19 +43,13 @@ case class WorkflowExecutionActorData(workflowDescriptor: EngineWorkflowDescript
   /** Checks if the workflow is completed by scanning through the executionStore.
     * If complete, this will return Some(finalStatus).  Otherwise, returns None */
   def workflowCompletionStatus: Option[ExecutionStatus] = {
+    def upstream(scope: Scope): List[Scope] = scope.prerequisiteScopes.toList ++ scope.prerequisiteScopes.toList.flatMap(upstream)
```
cjllanwarne: ToL: In this spirit of "this will all be different in a month", I believe this is part of wdl4s++

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActorData.scala
```diff
@@ -43,19 +43,13 @@ case class WorkflowExecutionActorData(workflowDescriptor: EngineWorkflowDescript
...
-        case NotStarted =>
-          // For NotStarted jobs: if any upstream jobs are failed, then assume this job will never run
-          def upstream(scope: Scope): Set[Scope] = scope.prerequisiteScopes ++ scope.prerequisiteScopes.flatMap(upstream)
-          val upstreamFailed = upstream(jobKey.scope) filter { s =>
-            executionStore.store.map({ case (a, b) => a.scope -> b }).get(s).contains(Failed)
-          }
-          if (upstreamFailed.nonEmpty) acc - jobKey else acc
-        case _ => acc
-      }
+    val activeJobs = executionStore.store.toList filter {
```
cjllanwarne: ToL: I dislike this name...
- `unfinishedJobs`?
- `incompleteJobs`?
- `detritusJobs`?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
mcovarr: sure

cjllanwarne:
👍 although I don't understand the low level magic that means this is better than it was before :)

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
geoffjentry: Do people look for those pairings?

I don't, although I could see why a person would

Horneth: Do we need the `Starting calls...` in the first place ?...

geoffjentry:
👍

kshakir:
I concur-- more comments in general to avoid people "fixing" anything you've profiled here. After that 👍

mcovarr:
@Horneth yeah something like that seems like a good idea, but a bit more than what I was trying to accomplish here.

PR #1469, opened 2016-09-21
Added an execution token dispenser to limit concurrent calls

cjllanwarne:
Consider this a strawman. Lots of tidying up still to do but... it works!


supportedBackends/sfs/src/main/scala/cromwell/backend/impl/sfs/config/ConfigBackendLifecycleActorFactory.scala
```diff
@@ -34,4 +35,10 @@ class ConfigBackendLifecycleActorFactory(val configurationDescriptor: BackendCon
 
   override lazy val fileHashingFunction: Option[FileHashingFunction] = Option(FileHashingFunction(ConfigBackendFileHashing.getMd5Result))
   override lazy val fileHashingActorCount: Int = 5
+
+  override val jobExecutionTokenType: JobExecutionTokenType = {
+    val concurrentJobLimit = configurationDescriptor.backendConfig.getIntOption("concurrent-job-limit")
+    System.out.println(s"Operating with a concurrent job limit for $name of ${concurrentJobLimit.getOrElse("None")}")
```
cjllanwarne: cookies!

PR #1471, opened 2016-09-21
[HtCondor] Added configurable docker container default working and output directories.

francares:
Goals of these changes:
1. Support configurable default docker container working dir for cases where dockerWorkingDir is not defined in runtime attributes.
2. Support configurable default docker container output dir for cases where dockerOutputDir is not defined in runtime attributes.
3. Be able to mount working directories for cases where a tool creates intermediate results (big files) in order to produce final output in the node FS and this is not capable to handle those file sizes.

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -228,13 +228,16 @@ backend {
     #
     #    #Placeholders:
     #    #1. Working directory.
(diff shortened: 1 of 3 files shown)


geoffjentry:
👍

gauravs90:
:+1:

PR #1472, opened 2016-09-22
Bug fix for metadata refresh.

kshakir:
When no rows were found to "refresh", was setting the summary status tracking back to zero, instead of the previous value.


Horneth:
👍

mcovarr:
:+1: glad you caught this before release!  😌

cjllanwarne:
😵 <=== 🌴

kshakir:
Centaur issues may be separate. Can be merged before or after resolving those issues.

PR #1482, opened 2016-09-23
 added $CROMWELL_ARGS  to entrypoint closes #1481

### LINKED ISSUE #1481: Can't use sbt-docker output from a docker-compose file
When running `docker-compose up`, I have to way to supply the necessary argument `server` to cromwell.  This makes me sad.

kcibul:


project/Settings.scala
```diff
@@ -57,7 +57,7 @@ object Settings {
         from("openjdk:8")
         expose(8000)
         add(artifact, artifactTargetPath)
-        entryPoint("java", "-jar", artifactTargetPath)
+        entryPoint("/bin/bash", "-c", "java -jar " + artifactTargetPath + " $CROMWELL_ARGS")
```
kcibul: note: the /bin/bash is because otherwise the variable expansion won't happen (thanks Docker!)

geoffjentry:
👍

project/Settings.scala
cjllanwarne: Is there a fancy sbt way of saying "read-from-env" to avoid that?

cjllanwarne:
👍 is this worth updating in the docker docs?

project/Settings.scala
kcibul: i spent a few hours banging against this, and I couldn't find a better way.  Docker is pretty clear that if you want shell variables to be used, you have to launch a shell yourself.

PR #1493, opened 2016-09-27
meow meow meow meow

mcovarr:


backend/src/main/scala/cromwell/backend/validation/ValidatedRuntimeAttributesBuilder.scala
```diff
@@ -68,14 +69,12 @@ trait ValidatedRuntimeAttributesBuilder {
     val errorsOrValuesMap: Seq[(String, ErrorOr[Any])] =
       validationsForValues.map(validation => validation.key -> validation.validate(values))
 
-    import scalaz.Scalaz._
-
-    val emptyResult: ErrorOr[List[(String, Any)]] = List.empty[(String, Any)].success
+    val emptyResult: ErrorOr[List[(String, Any)]] = List.empty[(String, Any)].validNel
     val validationResult = errorsOrValuesMap.foldLeft(emptyResult) { (agg, errorOrValue) =>
-      agg +++ {
+      agg |+| {
```
kshakir: ToL: More a question-- We have to implement this empty plus semigroup combine ourselves? I would have guessed that there would have been a default monoid implemented for validation. (aka fought for literally hours to try and get the syntax/implicits right)

README.md
```diff
@@ -101,7 +101,7 @@ There is a [Cromwell gitter channel](https://gitter.im/broadinstitute/cromwell)
 
 The following is the toolchain used for development of Cromwell.  Other versions may work, but these are recommended.
 
-* [Scala 2.11.7](http://www.scala-lang.org/news/2.11.7/)
+* [Scala 2.11.8](http://www.scala-lang.org/news/2.11.8/)
 * [SBT 0.13.8](https://github.com/sbt/sbt/releases/tag/v0.13.8)
```
mcovarr: will fix

backend/src/main/scala/cromwell/backend/validation/ValidatedRuntimeAttributesBuilder.scala
mcovarr: Interesting, I'll give this a try.

backend/src/main/scala/cromwell/backend/validation/ValidatedRuntimeAttributesBuilder.scala
```diff
@@ -54,32 +53,26 @@ trait ValidatedRuntimeAttributesBuilder {
...
-        }
-      }
+    val listOfErrorOrKeysToAnys: List[ErrorOr[(String, Any)]] = listOfKeysToErrorOrAnys map {
+      case (key, errorOrAny) => errorOrAny map { any => (key, any) }
     }
 
-    validationResult.map(result => ValidatedRuntimeAttributes(result.toMap))
+    import cats.syntax.traverse._
+    val errorOrListOfKeysToAnys: ErrorOr[List[(String, Any)]] = listOfErrorOrKeysToAnys.sequence[ErrorOr, (String, Any)]
+    errorOrListOfKeysToAnys map { listOfKeysToAnys => ValidatedRuntimeAttributes(listOfKeysToAnys.toMap) }
```
mcovarr: @kshakir I looked at this for a while and basically I don't think the types fit for a `Monoid#combineAll`.  While there is a `Monoid` instance for `Map`, it requires the `Map` values to have `Semigroup` instances.  This is because a `Map` semigroup/monoid would have to be able to combine values associatively if two entries have the same key.  But the values here are just `Any`s and for entries with the same key the `toMap` logic throws away one or the other value arbitrarily.

I did manage to clean this up a bit with some `sequence`ing though the variable names are now painfully explicit.  😄

kshakir:
👍

geoffjentry:
👍

PR #1500, opened 2016-09-29
Added an execution token dispenser to limit # of concurrent calls

cjllanwarne:
Suggested reading order:

- JobExecutionTokenDispenserActorSpec (to see how the TokenDispenserActor is intended to operate)
- EJEA (to see how the token dispenser gets invoked)
- Everything else (a lot of the more annoying wiring can probably be refactored once I add per-backend god-actors in my other ticket)

### DIFF
--- a/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
@@ -18,6 +18,7 @@ import cromwell.engine.workflow.lifecycle.execution.JobPreparationActor.{Backend
 import cromwell.engine.workflow.lifecycle.execution.callcaching.EngineJobHashingActor.{CacheHit, CacheMiss, CallCacheHashes, HashError}
 import cromwell.engine.workflow.lifecycle.execution.callcaching.FetchCachedResultsActor.{CachedOutputLookupFailed, CachedOutputLookupSucceeded}
 import cromwell.engine.workflow.lifecycle.execution.callcaching._
+import cromwell.engine.workflow.tokens.JobExecutionTokenDispenserActor.{JobExecutionTokenDenied, JobExecutionTokenDispensed, JobExecutionTokenRequest, JobExecutionTokenReturn}
 import cromwell.jobstore.JobStoreActor._
 import cromwell.jobstore.{Pending => _, _}
 import cromwell.services.SingletonServicesStore
@@ -25,6 +26,7 @@ import cromwell.services.metadata.MetadataService.PutMetadataAction
 import cromwell.services.metadata.{MetadataEvent, MetadataJobKey, MetadataKey, MetadataValue}
 import wdl4s.TaskOutput
 
+import scala.concurrent.ExecutionContext
 import scala.util.{Failure, Success, Try}
 
 class EngineJobExecutionActor(replyTo: ActorRef,
@@ -36,6 +38,7 @@ class EngineJobExecutionActor(replyTo: ActorRef,
                               serviceRegistryActor: ActorRef,
                               jobStoreActor: ActorRef,
                               callCacheReadActor: ActorRef,
+                              jobTokenDispenserActor: ActorRef,
                               backendName: String,
                               callCachingMode: CallCachingMode) extends LoggingFSM[EngineJobExecutionActorState, EJEAData] with WorkflowLogging {
 
@@ -51,7 +54,11 @@ class EngineJobExecutionActor(replyTo: ActorRef,
   // For tests:
   private[execution] def checkEffectiveCallCachingMode = effectiveCallCachingMode
 
+  private[execution] var executionToken: Option[JobExecutionToken] = None
+
   private val effectiveCallCachingKey = "Effective call caching mode"
+
+  implicit val ec: ExecutionContext = context.dispatcher
 
   log.debug(s"$tag: $effectiveCallCachingKey: $effectiveCallCachingMode")
   writeCallCachingModeToMetadata()
@@ -62,6 +69,13 @@ class EngineJobExecutionActor(replyTo: ActorRef,
   // When Pending, the FSM always has NoData
   when(Pending) {
     case Event(Execute, NoData) =>
+      requestExecutionToken()
+      goto(RequestingExecutionToken)
+  }
+
+  when(RequestingExecutionToken) {
+    case Event(JobExecutionTokenDispensed(jobExecutionToken), NoData) =>
+      executionToken = Option(jobExecutionToken)
       if (restarting) {
         val jobStoreKey = jobDescriptorKey.toJobStoreKey(workflowId)
         jobStoreActor ! QueryJobCompletion(jobStoreKey, jobDescriptorKey.call.task.outputs)
@@ -69,6 +83,9 @@ class EngineJobExecutionActor(replyTo: ActorRef,
       } else {
         prepareJob()
       }
+    case Event(JobExecutionTokenDenied(positionInQueue), NoData) =>
+      log.debug("Token denied so cannot start yet. Currently position {} in the queue", positionInQueue)
+      stay()
   }
 
   // When CheckingJobStore, the FSM always has NoData
@@ -146,7 +163,10 @@ class EngineJobExecutionActor(replyTo: ActorRef,
       saveJobCompletionToJobStore(data.withSuccessResponse(response))
     case Event(response: BackendJobExecutionResponse, data: ResponsePendingData) =>
       // This matches all response types other than `SucceededResponse`.
-      log.error("{}: Failed copying cache results, falling back to running job.", jobDescriptorKey)
+      response match {
+        case f: BackendJobFailedResponse =>log.error("{}: Failed copying cache results, falling back to running job: {}", jobDescriptorKey, f.throwable)
+        case _ => //
+      }
       runJob(data)
 
     // Hashes arrive:
@@ -217,8 +237,17 @@ class EngineJobExecutionActor(replyTo: ActorRef,
       stay
   }
 
+  private def requestExecutionToken(): Unit = {
+    jobTokenDispenserActor ! JobExecutionTokenRequest(factory.jobExecutionTokenType)
+  }
+
+  private def returnExecutionToken(): Unit = {
+    executionToken foreach { jobTokenDispenserActor ! JobExecutionTokenReturn(_) }
+  }
+
   private def forwardAndStop(response: Any): State = {
     replyTo forward response
+    returnExecutionToken()
     tellEventMetadata()
     context stop self
     stay()
@@ -226,6 +255,7 @@ class EngineJobExecutionActor(replyTo: ActorRef,
 
   private def respondAndStop(response: Any): State = {
     replyTo ! response
+    returnExecutionToken()
     tellEventMetadata()
     context stop self
     stay()
@@ -393,6 +423,7 @@ object EngineJobExecutionActor {
   /** States */
   sealed trait EngineJobExecutionActorState
   case object Pending extends EngineJobExecutionActorState
+  case object RequestingExecutionToken extends EngineJobExecutionActorState
   case object CheckingJobStore extends EngineJobExecutionActorState
   case object CheckingCallCache extends EngineJobExecutionActorState
   case object FetchingCachedOutputsFromDatabase extends EngineJobExecutionActorState
@@ -417,6 +448,7 @@ object EngineJobExecutionActor {
             serviceRegistryActor: ActorRef,
             jobStoreActor: ActorRef,
             callCacheReadActor: ActorRef,
+            jobTokenDispenserActor: ActorRef,
             backendName: String,
             callCachingMode: CallCachingMode) = {
     Props(new EngineJobExecutionActor(
@@ -429,6 +461,7 @@ object EngineJobExecutionActor {
       serviceRegistryActor = serviceRegistryActor,
       jobStoreActor = jobStoreActor,
       callCacheReadActor = callCacheReadActor,
+      jobTokenDispenserActor = jobTokenDispenserActor,
       backendName = backendName: String,
       callCachingMode = callCachingMode)).withDispatcher(EngineDispatcher)
   }
--- a/engine/src/main/scala/cromwell/engine/workflow/tokens/JobExecutionTokenDispenserActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/tokens/JobExecutionTokenDispenserActor.scala
@@ -1,0 +1,133 @@
+package cromwell.engine.workflow.tokens
+
+import akka.actor.{Actor, ActorLogging, ActorRef, Props, Terminated}
+import cromwell.core.JobExecutionToken
+import JobExecutionToken._
+import cromwell.engine.workflow.tokens.JobExecutionTokenDispenserActor._
+import cromwell.engine.workflow.tokens.TokenPool.TokenPoolPop
+
+import scala.collection.immutable.Queue
+import scala.language.postfixOps
+
+class JobExecutionTokenDispenserActor extends Actor with ActorLogging {
+
+  /**
+    * Lazily created token pool. We only create a pool for a token type when we need it
+    */
+  var tokenPools: Map[JobExecutionTokenType, TokenPool] = Map.empty
+  var tokenAssignments: Map[ActorRef, JobExecutionToken] = Map.empty
+
+  override def receive: Actor.Receive = {
+    case JobExecutionTokenRequest(tokenType) => sendTokenRequestResult(sender, tokenType)
+    case JobExecutionTokenReturn(token) => unassign(sender, token)
+    case Terminated(terminee) => onTerminate(terminee)
+  }
+
+  private def sendTokenRequestResult(sndr: ActorRef, tokenType: JobExecutionTokenType): Unit = {
+    if (tokenAssignments.contains(sndr)) {
+      sndr ! JobExecutionTokenDispensed(tokenAssignments(sndr))
+    } else {
+      context.watch(sndr)
+      val updatedTokenPool = getTokenPool(tokenType).pop() match {
+        case TokenPoolPop(newTokenPool, Some(token)) =>
+          assignAndSendToken(sndr, token)
+          newTokenPool
+        case TokenPoolPop(sizedTokenPoolAndQueue: SizedTokenPoolAndActorQueue, None) =>
+          val (poolWithActorEnqueued, positionInQueue) = sizedTokenPoolAndQueue.enqueue(sndr)
+          sndr ! JobExecutionTokenDenied(positionInQueue)
+          poolWithActorEnqueued
+        case TokenPoolPop(someOtherTokenPool, None) =>
+          //If this has happened, somebody's been playing around in this class and not covered this case:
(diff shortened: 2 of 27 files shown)


core/src/main/scala/cromwell/core/execution.scala
```diff
@@ -0,0 +1,6 @@
+package cromwell.core
+
+
+class execution {
+
```
geoffjentry: ?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -62,13 +70,26 @@ class EngineJobExecutionActor(replyTo: ActorRef,
...
         jobStoreActor ! QueryJobCompletion(jobStoreKey, jobDescriptorKey.call.task.outputs)
         goto(CheckingJobStore)
       } else {
         prepareJob()
       }
+    case Event(JobExecutionTokenDenied, NoData) =>
+      // Wait randomly between 100 and 200 millis, then retry:
+      // TODO: Make this a queue on the dispensor. This way is far too likely to cause starvation!
+      val delay = 1000 + Random.nextInt(1000)
+      context.system.scheduler.scheduleOnce(delay.milliseconds) { requestExecutionToken() }
```
Horneth: maybe use exponential backoff ?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -146,7 +167,11 @@ class EngineJobExecutionActor(replyTo: ActorRef,
       saveJobCompletionToJobStore(data.withSuccessResponse(response))
     case Event(response: BackendJobExecutionResponse, data: ResponsePendingData) =>
       // This matches all response types other than `SucceededResponse`.
-      log.error("{}: Failed copying cache results, falling back to running job.", jobDescriptorKey)
+      response match {
+        case f: FailedNonRetryableResponse =>log.error("{}: Failed copying cache results, falling back to running job: {}", jobDescriptorKey, f.throwable)
+        case f: FailedRetryableResponse =>log.error("{}: Failed copying cache results, falling back to running job: {}", jobDescriptorKey, f.throwable)
+        case _ => //
```
Horneth: what else can it be ?
Also the 2 cases can be merged into 
`case f: FailedNonRetryableResponse | FailedRetryableResponse => ...` I think ?

core/src/main/scala/cromwell/core/execution.scala
cjllanwarne: Interesting...

Looks like an accidental "create Package" => "create class" finger slip. Will remove

cjllanwarne: Done

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -62,13 +69,23 @@ class EngineJobExecutionActor(replyTo: ActorRef,
...
       if (restarting) {
         val jobStoreKey = jobDescriptorKey.toJobStoreKey(workflowId)
         jobStoreActor ! QueryJobCompletion(jobStoreKey, jobDescriptorKey.call.task.outputs)
         goto(CheckingJobStore)
       } else {
         prepareJob()
       }
+    case Event(JobExecutionTokenDenied(positionInQueue), NoData) =>
+      log.debug("Token denied so cannot start yet. Currently position {} in the queue", positionInQueue)
+      stay()
```
Horneth: We don't wait and ask again for a token ?

engine/src/main/scala/cromwell/engine/workflow/tokens/JobExecutionTokenDispenserActor.scala
```diff
@@ -0,0 +1,130 @@
...
+    } else {
+      context.watch(sndr)
+      val updatedTokenPool = getTokenPool(tokenType).pop() match {
+        case TokenPoolPop(newTokenPool, Some(token)) =>
+          assignAndSendToken(sndr, token)
+          newTokenPool
+        case TokenPoolPop(sizedTokenPoolAndQueue: SizedTokenPoolAndActorQueue, None) =>
+          val (poolWithActorEnqueued, positionInQueue) = sizedTokenPoolAndQueue.enqueue(sndr)
+          sndr ! JobExecutionTokenDenied(positionInQueue)
+          poolWithActorEnqueued
```
Horneth: for completeness (even if the other type of pool is Infinite so it won't happen), the 
`case TokenPoolPop(somethingNotASizedTokenPoolAndActorQueue, None)` is not caught here

engine/src/main/scala/cromwell/engine/workflow/tokens/JobExecutionTokenDispenserActor.scala
```diff
@@ -0,0 +1,130 @@
...
+
+      val pool = getTokenPool(token.jobExecutionTokenType) match {
+        case SizedTokenPoolAndActorQueue(innerPool, queue) if queue.nonEmpty =>
+          val (nextInLine, newQueue) = queue.dequeue
+          assignAndSendToken(nextInLine, token)
+          SizedTokenPoolAndActorQueue(innerPool, newQueue)
+        case other => other
+      }
+
+      tokenPools += token.jobExecutionTokenType -> pool.push(token)
```
Horneth: If I understand, above on line 68 you give to the next in line the token that just got unassigned. I don't think that in that case you want to push back the token to to the pool ? (if the queue is empty and you don't re-assign it then yes it makes sense to push it back).

engine/src/main/scala/cromwell/engine/workflow/tokens/JobExecutionTokenDispenserActor.scala
```diff
@@ -0,0 +1,130 @@
...
+    }
+  }
+
+  private def onTerminate(terminee: ActorRef) = {
+    tokenAssignments.get(terminee) match {
+      case Some(token) =>
+        log.error("Actor {} stopped without returning its Job Execution Token ({}). Reclaiming it!", terminee, tokenAssignments(terminee))
+        self.tell(msg = JobExecutionTokenReturn(token), sender = terminee)
+      case None =>
+        log.debug("Actor {} stopped unexpectedly but it doesn't have a token. Removing it from queues if necessary", terminee)
```
Horneth: Does this method only get called if the actor stops unexpectedly ? Or does it also catch when the actor stops normally ?

engine/src/test/scala/cromwell/engine/workflow/tokens/JobExecutionTokenDispenserActorSpec.scala
```diff
@@ -0,0 +1,304 @@
...
+    actorRefUnderTest ! JobExecutionTokenRequest(TestInfiniteTokenType)
+    expectMsgPF(max = MaxWaitTime, hint = "token dispensed message") {
+      case JobExecutionTokenDispensed(token) =>
+        actorRefUnderTest ! JobExecutionTokenReturn(token)
+    }
+  }
+
+  it should "dispense indefinitely for an infinite token type" in {
+    var currentSet: Set[JobExecutionToken] = Set.empty
+    100 indexedTimes { i =>
```
Horneth: 100 = ∞ ? xD

engine/src/main/scala/cromwell/engine/workflow/tokens/TokenPool.scala
```diff
@@ -0,0 +1,49 @@
...
+      SizedTokenPool(pool.toSet, Set.empty)
+    } getOrElse {
+      InfiniteTokenPool(tokenType, Set.empty)
+    }
+  }
+}
+
+final case class SizedTokenPool(pool: Set[JobExecutionToken], override val currentLoans: Set[JobExecutionToken]) extends TokenPool {
+
+  override def pop(): TokenPoolPop = pool.toList match {
```
mcovarr: The `toList`ing / `toSet` ing seems strange, perhaps just `pool.isEmpty` and getting the first element of the pool as an iterator?

engine/src/main/scala/cromwell/engine/workflow/tokens/TokenPool.scala
```diff
@@ -0,0 +1,49 @@
...
+    case head :: tail => TokenPoolPop(SizedTokenPool(tail.toSet, currentLoans + head), Some(head))
```
mcovarr: `Option`

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/ejea/EjeaRequestingExecutionTokenSpec.scala
```diff
@@ -0,0 +1,54 @@
...
+        ejea = helper.buildEJEA(restarting = false)
+        ejea ! JobExecutionTokenDispensed(helper.executionToken)
+
+        helper.jobPreparationProbe.expectMsg(max = awaitTimeout, hint = "Awaiting job preparation", JobPreparationActor.Start)
+        helper.jobStoreProbe.expectNoMsg(awaitAlmostNothing)
+        ejea.stateName should be(PreparingJob)
+      }
+    }
+  }
+}
```
mcovarr: \n

engine/src/test/scala/cromwell/engine/workflow/tokens/JobExecutionTokenDispenserActorSpec.scala
mcovarr: Travis will make it feel like ∞

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
geoffjentry: You could also add a layer in the BJEA response ADT to have a `FailedResponse` which those two extend and then just check for that here.

geoffjentry:
To what extent, if any, does the stuff in ejea move out with your upcoming changes?

I don't know how else it could be structured but I'm a little sad that we're adding to the hairball that is ejea instead of going the other direction

engine/src/main/scala/cromwell/engine/workflow/tokens/JobExecutionTokenDispenserActor.scala
```diff
@@ -0,0 +1,130 @@
...
+  var tokenAssignments: Map[ActorRef, JobExecutionToken] = Map.empty
+
+  override def receive: Actor.Receive = {
+    case JobExecutionTokenRequest(tokenType) => sendTokenRequestResult(sender, tokenType)
+    case JobExecutionTokenReturn(token) => unassign(sender, token)
+    case Terminated(terminee) => onTerminate(terminee)
+  }
+
+  private def sendTokenRequestResult(sndr: ActorRef, tokenType: JobExecutionTokenType): Unit = {
+    if(tokenAssignments.contains(sndr)) {
```
geoffjentry: `if (`

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
cjllanwarne: I originally had it doing this, but realised we could end up with 10s of thousands of actors requesting tokens every second for no purposeful reason since they know that they're already queued up if they get the "Denied" response (and forcing the dispenser to traverse its queue every time to find the index seems wasteful!)

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
cjllanwarne: This was removed by a subsequent commit

engine/src/main/scala/cromwell/engine/workflow/tokens/TokenPool.scala
```diff
@@ -0,0 +1,49 @@
...
+    case Nil => TokenPoolPop(SizedTokenPool(Set.empty, currentLoans), None)
+  }
+
+  override def push(token: JobExecutionToken): SizedTokenPool = {
+    if (currentLoans.contains(token)) { SizedTokenPool(pool + token, currentLoans - token) } else this
+  }
+}
+
+final case class InfiniteTokenPool(tokenType: JobExecutionTokenType, override val currentLoans: Set[JobExecutionToken]) extends TokenPool {
```
geoffjentry: While I heartily applaud the belts & suspenders of the `final` here, why are you doing it here and not everywhere else? :)

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
cjllanwarne: It could also be an AbortedResponse (if/when we wire that through to the cache hit actor...). It also stops scala from whining about incomplete case statements... :)

But I'll definitely merge those failed two cases. That's just lazy copy/pasting...

engine/src/main/scala/cromwell/engine/workflow/tokens/JobExecutionTokenDispenserActor.scala
cjllanwarne: That's a great spot! I'll enhance the test case to make sure I catch this

engine/src/main/scala/cromwell/engine/workflow/tokens/JobExecutionTokenDispenserActor.scala
cjllanwarne: Will rephrase

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
Horneth: Oh right I forgot they were queued up so that they would eventually get one without having to ask

engine/src/test/scala/cromwell/engine/workflow/tokens/JobExecutionTokenDispenserActorSpec.scala
cjllanwarne: That's just how I count... One, Two, Many, Infinite

engine/src/main/scala/cromwell/engine/workflow/tokens/JobExecutionTokenDispenserActor.scala
```diff
@@ -0,0 +1,130 @@
...
+      context.unwatch(actor)
+    } else {
+      log.error("Job execution token returned from incorrect actor: {}", token)
+    }
+  }
+
+  private def onTerminate(terminee: ActorRef) = {
+    tokenAssignments.get(terminee) match {
+      case Some(token) =>
+        log.error("Actor {} stopped without returning its Job Execution Token ({}). Reclaiming it!", terminee, tokenAssignments(terminee))
```
cjllanwarne: Agree. But it wasn't really adding much value so I'll just say "token". Which token was only really useful during debugging

engine/src/main/scala/cromwell/engine/workflow/tokens/JobExecutionTokenDispenserActor.scala
```diff
@@ -0,0 +1,130 @@
...
+      queue.indexOf(actor) match {
+        case -1 =>
+          val newQueue = queue :+ actor
+          (SizedTokenPoolAndActorQueue(sizedPool, newQueue), newQueue.size - 1) // Convert from 1-indexed to 0-indexed
+        case index => (this, index)
+      }
+    }
+  }
+
+}
```
cjllanwarne: Done

engine/src/main/scala/cromwell/engine/workflow/tokens/TokenPool.scala
cjllanwarne: Looks like `Set` has a `.head` and `.tail` method for me to use so it's even easier

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/ejea/EjeaRequestingExecutionTokenSpec.scala
cjllanwarne: ok

engine/src/main/scala/cromwell/engine/workflow/tokens/TokenPool.scala
cjllanwarne: Which do you mean?

cjllanwarne:
@geoffjentry We could certainly move the token dispenser into a per-BE model (and then only have one tokenPool to make the logic simpler. Hooray!)

Also FWIW I don't believe the EJEA is a hairball. It's just getting large enough to merit decomposing.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
cjllanwarne: Done

engine/src/main/scala/cromwell/engine/workflow/tokens/JobExecutionTokenDispenserActor.scala
cjllanwarne: Done

engine/src/main/scala/cromwell/engine/workflow/tokens/TokenPool.scala
cjllanwarne: Done

engine/src/main/scala/cromwell/engine/workflow/tokens/TokenPool.scala
```diff
@@ -30,11 +30,13 @@ object TokenPool {
 
 final case class SizedTokenPool(pool: Set[JobExecutionToken], override val currentLoans: Set[JobExecutionToken]) extends TokenPool {
 
-  override def pop(): TokenPoolPop = pool.toList match {
-    case head :: tail => TokenPoolPop(SizedTokenPool(tail.toSet, currentLoans + head), Some(head))
-    case Nil => TokenPoolPop(SizedTokenPool(Set.empty, currentLoans), None)
+  override def pop(): TokenPoolPop = if (pool.isEmpty) {
+    TokenPoolPop(SizedTokenPool(Set.empty, currentLoans), None)
+  } else {
+    TokenPoolPop(SizedTokenPool(pool.tail, currentLoans + pool.head), pool.headOption)
```
Horneth: Totally ToL: Is it guaranteed that multiple calls to Set.tail,Set.head (and headOption) will always return the same value ? (since there is no ordering)

cjllanwarne: I did some refactoring but I really don't like having to iterate the entire set just to remove the head. Since nothing else looks at it, I'm just going to make it a list so I can use `::` to split it!

Horneth:
👍

cjllanwarne:
@mcovarr I think this is good to go? Any other comments?

mcovarr:
Sorry I had "approved" but forgot the pullapprove-friendly :+1:

cjllanwarne:
@mcovarr can't wait for "require two approvals" mode in github (*shakes fist at sky*)

PR #1509, opened 2016-09-29
adding support for JAVA_OPTS env variable and symlink jar to /app/cromwell.jar

kcibul:
requests (and good ideas in general) from workbench


geoffjentry:
👍

kshakir:
ToL: At some point, I would love to see one of the many travis targets (or yet-another) ensure that `sbt docker` still works.

For now: 👍

PR #1511, opened 2016-10-04
Mutliple Inputs

ruchim:


engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
```diff
@@ -26,6 +27,19 @@ trait CromwellApiService extends HttpService with PerRequestCreator {
   val workflowStoreActor: ActorRef
   val serviceRegistryActor: ActorRef
 
+  def toMap(someInput: Option[String]): Map[String, JsValue] = {
+    import spray.json._
+    someInput match {
+      case Some(inputs: String) => inputs.parseJson match {
+        case JsObject(inputMap) => inputMap
+        case _ =>
+          //throw new RuntimeException(s"Submitted inputs couldn't be processed, please check for syntactical errors.")
```
ruchim: Woops, need to fix that.

kshakir:
**TL;DR Discussed in person with @ruchim. Going to 👍 , and perhaps dev choice a PR to change the syntax, using a secondary route that looks for `workflowInputs[]`.**

This current PR is very swagger spec friendly, using a fixed 2-based list of additional inputs:
- `workflowInputs`
- `workflowInputs_2`
- `workflowInputs_3`
- `workflowInputs_4`
- `workflowInputs_5`

Ideally we could use a PHP compatible syntax on a secondary spray route:
- `workflowInputs[]`
- `workflowInputs[]`
- `workflowInputs[]`
- etc.

This array, using a [custom](https://groups.google.com/d/msg/spray-user/5kSZ87OnfkE/I_A_OcaIticJ) spray marshaller could be programmatically converted into an variable length `workflowInputs: Seq[String]`.

Passing the sequence into the business logic would also allow storing the separated inputs in the metadata. For now the five inputs are merged into a single value in the web service and stored in the database as a merged clob.

At this second I do not know if swagger would allow multiple form data elements with the same name. I'm assuming curl, HTTPie, and jvm clients such as spray-client would, as PHP supports the above syntax.

geoffjentry:
FWIW in Akka HTTP (for now at least) you need to process the multi-part form manually anyways so I wouldn't worry too much about finding some perfect magic solution.

Horneth:
👍

PR #1518, opened 2016-10-05
JES: Manage API requests and poll in batches

cjllanwarne:
In case anyone wants to have an early preview: Singleton backend actors, JES batching and work pulling. Oh my.

### DIFF
--- a/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
@@ -39,6 +39,7 @@ class EngineJobExecutionActor(replyTo: ActorRef,
                               jobStoreActor: ActorRef,
                               callCacheReadActor: ActorRef,
                               jobTokenDispenserActor: ActorRef,
+                              backendSingletonActor: Option[ActorRef],
                               backendName: String,
                               callCachingMode: CallCachingMode) extends LoggingFSM[EngineJobExecutionActorState, EJEAData] with WorkflowLogging {
 
@@ -282,7 +283,7 @@ class EngineJobExecutionActor(replyTo: ActorRef,
   def createJobPreparationActor(jobPrepProps: Props, name: String): ActorRef = context.actorOf(jobPrepProps, name)
   def prepareJob() = {
     val jobPreparationActorName = s"BackendPreparationActor_for_$jobTag"
-    val jobPrepProps = JobPreparationActor.props(executionData, jobDescriptorKey, factory, initializationData, serviceRegistryActor)
+    val jobPrepProps = JobPreparationActor.props(executionData, jobDescriptorKey, factory, initializationData, serviceRegistryActor, backendSingletonActor)
     val jobPreparationActor = createJobPreparationActor(jobPrepProps, jobPreparationActorName)
     jobPreparationActor ! JobPreparationActor.Start
     goto(PreparingJob)
@@ -449,6 +450,7 @@ object EngineJobExecutionActor {
             jobStoreActor: ActorRef,
             callCacheReadActor: ActorRef,
             jobTokenDispenserActor: ActorRef,
+            backendSingletonActor: Option[ActorRef],
             backendName: String,
             callCachingMode: CallCachingMode) = {
     Props(new EngineJobExecutionActor(
@@ -462,6 +464,7 @@ object EngineJobExecutionActor {
       jobStoreActor = jobStoreActor,
       callCacheReadActor = callCacheReadActor,
       jobTokenDispenserActor = jobTokenDispenserActor,
+      backendSingletonActor = backendSingletonActor,
       backendName = backendName: String,
       callCachingMode = callCachingMode)).withDispatcher(EngineDispatcher)
   }
--- a/supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
+++ b/supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
@@ -15,6 +15,7 @@ import cromwell.backend.impl.jes.JesImplicits.PathString
 import cromwell.backend.impl.jes.JesJobExecutionActor.JesOperationIdKey
 import cromwell.backend.impl.jes.RunStatus.TerminalRunStatus
 import cromwell.backend.impl.jes.io._
+import cromwell.backend.impl.jes.statuspolling.JesPollingActorClient
 import cromwell.backend.{AttemptedLookupResult, BackendJobDescriptor, BackendWorkflowDescriptor, PreemptedException}
 import cromwell.core.Dispatcher.BackendDispatcher
 import cromwell.core._
@@ -42,12 +43,14 @@ object JesAsyncBackendJobExecutionActor {
             completionPromise: Promise[BackendJobExecutionResponse],
             jesWorkflowInfo: JesConfiguration,
             initializationData: JesBackendInitializationData,
-            serviceRegistryActor: ActorRef): Props = {
+            serviceRegistryActor: ActorRef,
+            jesBackendSingletonActor: ActorRef): Props = {
     Props(new JesAsyncBackendJobExecutionActor(jobDescriptor,
       completionPromise,
       jesWorkflowInfo,
       initializationData,
-      serviceRegistryActor)).withDispatcher(BackendDispatcher)
+      serviceRegistryActor,
+      jesBackendSingletonActor)).withDispatcher(BackendDispatcher)
   }
 
   object WorkflowOptionKeys {
@@ -78,10 +81,13 @@ class JesAsyncBackendJobExecutionActor(override val jobDescriptor: BackendJobDes
                                        override val completionPromise: Promise[BackendJobExecutionResponse],
                                        override val jesConfiguration: JesConfiguration,
                                        override val initializationData: JesBackendInitializationData,
-                                       override val serviceRegistryActor: ActorRef)
-  extends Actor with ActorLogging with AsyncBackendJobExecutionActor with JesJobCachingActorHelper with JobLogging {
+                                       override val serviceRegistryActor: ActorRef,
+                                       val jesBackendSingletonActor: ActorRef)
+  extends Actor with ActorLogging with AsyncBackendJobExecutionActor with JesJobCachingActorHelper with JobLogging with JesPollingActorClient {
 
   import JesAsyncBackendJobExecutionActor._
+
+  override val pollingActor = jesBackendSingletonActor
 
   override lazy val pollBackOff = SimpleExponentialBackoff(
     initialInterval = 30 seconds, maxInterval = 10 minutes, multiplier = 1.1)
@@ -119,7 +125,7 @@ class JesAsyncBackendJobExecutionActor(override val jobDescriptor: BackendJobDes
     case KvPutSuccess(_) => // expected after the KvPut for the operation ID
   }
 
-  override def receive: Receive = jesReceiveBehavior orElse super.receive
+  override def receive: Receive = pollingActorClientReceive orElse jesReceiveBehavior orElse super.receive
 
   private def globOutputPath(glob: String) = callRootPath.resolve(s"glob-${glob.md5Sum}/")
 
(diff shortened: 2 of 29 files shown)


engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -39,6 +39,7 @@ class EngineJobExecutionActor(replyTo: ActorRef,
                               jobStoreActor: ActorRef,
                               callCacheReadActor: ActorRef,
                               jobTokenDispenserActor: ActorRef,
+                              backendSingletonActor: Option[ActorRef],
```
geoffjentry: This name doesn't really describe why it exists

cjllanwarne: Very happy to hear alternatives? (perhaps something so awful we're certain to change it...? Pinging @mcovarr 😛)

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesApiQueryManager.scala
```diff
@@ -0,0 +1,69 @@
...
+
+  private var workQueue: List[JesStatusPollQuery] = List.empty
+
+  // TODO: Random magic number 5. Dynamically resize this?
+  // TODO: Supervision strategy for pollers should be: stop that one, tell its requesters that it failed, and create a new one to replace it.
+  private val pollers = (0 until 5) map { i => context.actorOf(JesPollingActor.props(self)) }
+  pollers foreach { _ ! NoWorkToDo }
+
+  override def receive = {
+    case DoPoll(run) => workQueue :+= JesStatusPollQuery(sender, run)
```
mcovarr: Appends to a `List` are O(n); this should probably be a different data structure.

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesApiQueryManager.scala
```diff
@@ -0,0 +1,69 @@
...
+    workQueue = beheaded.newWorkQueue
+  }
+
+  private final case class BeheadedWorkQueue(workToDo: Option[NonEmptyList[JesStatusPollQuery]], newWorkQueue: List[JesStatusPollQuery])
+  private def beheadWorkQueue: BeheadedWorkQueue = {
+    workQueue.grouped(100).toList match { // TODO: 100 is a random magic number: too big/too small/just right?
+      case head :: tail => BeheadedWorkQueue(Option(NonEmptyList.fromListUnsafe(head)), tail.flatten) // NB: This IS safe: we know that head is non-empty due to the behavior of .grouped
+      case Nil => BeheadedWorkQueue(None, List.empty)
+    }
+  }
```
mcovarr: This seems kind of complicated.  Couldn't this just be a `take` / `drop` on the underlying `List` (given the current choice of `List` for the implementation)?

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesPollingActor.scala
```diff
@@ -0,0 +1,87 @@
...
+
+    override def onFailure(e: GoogleJsonError, responseHeaders: HttpHeaders): Unit = {
+      log.debug(s"Batch request onFailure callback triggered!")
+      originalRequest.requester ! JesPollFailed(e, responseHeaders)
+      completionPromise.tryFailure(new Exception()) // TODO: We could recover and retry this?
+    }
+  }
+
+  // TODO: FSMify this actor!
+  private def interstitialRecombobulation: PartialFunction[Try[_], Unit] = {
```
mcovarr: 😂

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesPollingActor.scala
```diff
@@ -0,0 +1,87 @@
...
+      val batchResultFutures = handleBatch(workBatch)
+      val overallFuture = Future.sequence(batchResultFutures.map(_._2).toList)
+      overallFuture.andThen(interstitialRecombobulation)
+    case NoWorkToDo =>
+      context.system.scheduler.scheduleOnce(10.seconds) { checkForWork() }
+  }
+
+
+  // TODO: safety first! That .queue(...) can throw an IOException!
+  private def handleBatch(workBatch: NonEmptyList[JesStatusPollQuery]): NonEmptyList[(ActorRef, Future[Unit])] = {
```
mcovarr: the result type seems more complicated than the only caller wants

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesPollingActor.scala
```diff
@@ -0,0 +1,87 @@
...
+class JesPollingActor(pollingManager: ActorRef) extends Actor with ActorLogging {
+
+  implicit val ec: ExecutionContext = context.dispatcher
+
+  override def receive = {
+
+    case JesPollingWorkBatch(workBatch) =>
+      log.debug(s"Got a polling batch with ${workBatch.tail.size + 1} requests.")
+      val batchResultFutures = handleBatch(workBatch)
+      val overallFuture = Future.sequence(batchResultFutures.map(_._2).toList)
```
mcovarr:
```scala
  import cats.syntax.traverse._

  val overallFuture = batchResultFutures.sequence
```

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesPollingActor.scala
```diff
@@ -0,0 +1,87 @@
...
+      overallFuture.andThen(interstitialRecombobulation)
+    case NoWorkToDo =>
+      context.system.scheduler.scheduleOnce(10.seconds) { checkForWork() }
+  }
+
+
+  // TODO: safety first! That .queue(...) can throw an IOException!
+  private def handleBatch(workBatch: NonEmptyList[JesStatusPollQuery]): NonEmptyList[(ActorRef, Future[Unit])] = {
+
+    // Assume that the first element is also
```
mcovarr: unfinished thought

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesPollingActor.scala
```diff
@@ -0,0 +1,87 @@
...
+    override def onSuccess(operation: Operation, responseHeaders: HttpHeaders): Unit = {
+      log.debug(s"Batch result onSuccess callback triggered!")
+      originalRequest.requester ! originalRequest.run.interpretOperationStatus(operation)
+      completionPromise.trySuccess(())
+    }
+
+    override def onFailure(e: GoogleJsonError, responseHeaders: HttpHeaders): Unit = {
+      log.debug(s"Batch request onFailure callback triggered!")
+      originalRequest.requester ! JesPollFailed(e, responseHeaders)
+      completionPromise.tryFailure(new Exception()) // TODO: We could recover and retry this?
```
mcovarr: yeah this will likely need to be more robust

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesPollingActor.scala
```diff
@@ -0,0 +1,87 @@
...
+  private def handleBatch(workBatch: NonEmptyList[JesStatusPollQuery]): NonEmptyList[(ActorRef, Future[Unit])] = {
+
+    // Assume that the first element is also
+    val batch: BatchRequest = workBatch.head.run.genomicsInterface.batch()
+
+    // Create the batch:
+    val requesterMap = workBatch map { pollingRequest =>
+      val completionPromise = Promise[Unit]()
+      pollingRequest.run.getOperationCommand.queue(batch, batchResultHandler(pollingRequest, completionPromise))
+      (pollingRequest.requester, completionPromise.future)
```
mcovarr: as mentioned above, the requester is not of interest to the caller

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesPollingActor.scala
```diff
@@ -0,0 +1,87 @@
...
+    }
+  }
+
+  // TODO: FSMify this actor!
+  private def interstitialRecombobulation: PartialFunction[Try[_], Unit] = {
+    case Success(_) =>
+      log.debug(s"All status polls completed successfully.")
+      checkForWork()
+    case Failure(t) =>
+      log.error("Error fetching JES status: {}. This JES polling actor is stopping itself")
```
mcovarr: missing parameter for template

mcovarr:
Looks good but I think I would still benefit from a high-level description of the changes

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesApiQueryManager.scala
cjllanwarne: I like this... and I can then switch it out for a Queue to improve that O(n) addition time you mentioned

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesApiQueryManager.scala
```diff
@@ -0,0 +1,69 @@
...
+  * Currently, just holds a set of JES status poll requests until a PollingActor pulls the work.
+  * TODO: Could eventually move all of the JES queries into a single work-pulling model.
+  */
+class JesApiQueryManager extends Actor with ActorLogging {
+
+  private var workQueue: List[JesStatusPollQuery] = List.empty
+
+  // TODO: Random magic number 5. Dynamically resize this?
+  // TODO: Supervision strategy for pollers should be: stop that one, tell its requesters that it failed, and create a new one to replace it.
+  private val pollers = (0 until 5) map { i => context.actorOf(JesPollingActor.props(self)) }
```
cjllanwarne: I like the fact that `0 until 5` gives you `5` iterations rather than 6.

cjllanwarne: but as a side-note I don't think this is a good long-term way of generating the workers...

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
```diff
@@ -510,19 +521,19 @@ class JesAsyncBackendJobExecutionActor(override val jobDescriptor: BackendJobDes
     errorMessage.substring(0, errorMessage.indexOf(':')).toInt
   }
 
-  private def preempted(errorCode: Int, errorMessage: Option[String]): Boolean = {
+  private def preempted(errorCode: Int, errorMessage: List[String]): Boolean = {
```
cjllanwarne: The return from the batch API gives you a list of errors instead of an option of one (in the case where the REST query fails, as opposed to the query successfully returning a failure message)

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesPollingActor.scala
cjllanwarne: I **definitely** need to change this awful name

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesPollingActor.scala
cjllanwarne: 😳

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
mcovarr: Dunno, you've set a mighty high bar with `interstitialRecombobulation`.  😛 

And actually I don't think the current name is too far off the mark, but it would perhaps benefit from at least a little documentation to describe its purpose.

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesApiQueryManager.scala
Horneth: Router ?

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesApiQueryManager.scala
```diff
@@ -0,0 +1,69 @@
...
+        log.debug(s"No work for JES poller. Sad.")
+        workPullingJesPollingActor ! NoWorkToDo
+    }
+
+    workQueue = beheaded.newWorkQueue
+  }
+
+  private final case class BeheadedWorkQueue(workToDo: Option[NonEmptyList[JesStatusPollQuery]], newWorkQueue: List[JesStatusPollQuery])
+  private def beheadWorkQueue: BeheadedWorkQueue = {
+    workQueue.grouped(100).toList match { // TODO: 100 is a random magic number: too big/too small/just right?
```
Horneth: Is that relevant ? 

> You're limited to 100 calls in a single batch request. If you need to make more calls than that, use multiple batch requests.

https://cloud.google.com/storage/docs/json_api/v1/how-tos/batch

cjllanwarne: Ha! Magic random number == precisely the right amount!

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
geoffjentry: @cjllanwarne despite what it might seem, i'm actually trying to not opine on these too much right now ;) 

@mcovarr makes a decent point in my mind, that given enough documentation perhaps this is better than what I was thinking. I was going to suggest something like `BackendManagerActor` but while the concept originally came from our convo where I was talking about an "overlord actor" that's not necessarily what this needs to be.

I think the real thing is that we need some serious (as in quality, not quantity) documentation of our backend sitch and somehow this needs to be filed under "don't just explain it in some paragraph that nobody reads"

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesPollingActor.scala
cjllanwarne: What's this one for?

(I just refactored this so that I could get the List of failures by lifting `overallFuture` up into a `Future[List[Try[Unit]]]`. Would this do the same "for free")?

PR #1519, opened 2016-10-05
Gcs Nio

Horneth:


core/src/main/scala/cromwell/core/path/RetryableFileSystemProviderProxy.scala
```diff
@@ -0,0 +1,51 @@
...
+import java.nio.file.attribute.{BasicFileAttributes, FileAttribute, FileAttributeView}
+import java.nio.file.spi.FileSystemProvider
+import java.util
+
+import akka.actor.ActorSystem
+import cromwell.core.retry.Retry
+
+import scala.concurrent.{Await, Future}
+
+class RetryableFileSystemProviderProxy[T <: FileSystemProvider](delegate: T, retryParams: CustomRetryParams = CustomRetryParams.Default)(implicit actorSystem: ActorSystem) extends FileSystemProvider {
```
Horneth: I'm realizing now that this is not going to work - gonna have to find another way to do retries

Horneth: Added a few more `xxxProxy` classes to make it work

core/src/test/scala/cromwell/core/path/RetryableFileSystemProxySpec.scala
```diff
@@ -0,0 +1,278 @@
...
+    verify(mockFs, times(3)).delete(any[Path])
+    verify(mockFs, times(3)).newDirectoryStream(any[Path], any[Filter[Path]]())
+    verify(mockFs, times(3)).setAttribute(any[Path], any[String], any[Object])
+    verify(mockFs, times(3)).readAttributes(any[Path], any[String])
+    verify(mockFs, times(3)).readAttributes(any[Path], any[Class[BasicFileAttributes]])
+    verify(mockFs, times(3)).isSameFile(any[Path], any[Path])
+    verify(mockFs, times(3)).getFileAttributeView(any[Path], any[Class[FileAttributeView]])
+  }
+
+  it should "fail imediately on fatal exceptions" in {
```
mcovarr: imediately => immediately

mcovarr:
:+1:

engine/src/test/scala/cromwell/engine/workflow/SingleWorkflowRunnerActorSpec.scala
```diff
@@ -111,9 +114,9 @@ class SingleWorkflowRunnerActorWithMetadataSpec extends SingleWorkflowRunnerActo
       singleWorkflowActor(
         sampleWdl = wdlFile,
         outputFile = Option(metadataFile.path))
+        TestKit.shutdownActorSystem(system, TimeoutDuration)
     }
-    TestKit.shutdownActorSystem(system, TimeoutDuration)
-
+    println(metadataFile.pathAsString)
```
kshakir: Delete this `println`

project/Dependencies.scala
```diff
@@ -14,7 +14,8 @@ object Dependencies {
   lazy val sprayJsonV = "1.3.2"
   lazy val akkaV = "2.4.9"
   lazy val slickV = "3.1.1"
-  lazy val googleClientApiV = "1.20.0"
+  lazy val googleClientApiV = "1.22.0"
+  lazy val googleApiV = "1.20.0"
```
kshakir: This name isn't specific enough. Instead of `googleApi`, how about `googleGenomicsServicesApi'?

project/Dependencies.scala
```diff
@@ -86,7 +87,9 @@ object Dependencies {
 
   // Sub-project dependencies, added in addition to any dependencies inherited from .dependsOn().
 
-  val gcsFileSystemDependencies = baseDependencies ++ googleApiClientDependencies ++ googleCloudDependencies
+  val gcsFileSystemDependencies = baseDependencies ++ googleApiClientDependencies ++ googleCloudDependencies ++ List (
+    "com.github.pathikrit" %% "better-files" % betterFilesV
```
kshakir: Don't duplicate dependencies. If you're adding better-files to gcs, since engine inherits from gcs, then remove them from the engine.

kshakir:
TL;DR 👍  post rebase and minor comments.

ToL 1: I dislike that one can't `Paths.get(Paths.get("gs://bucket/path").toString)`, and one must insert a `.toUri` for this to work. Hopefully this will be updated over in GCS NIO in the future and we can convert all the `.toUri.toString` back to `.toString`.

ToL 2: Already discussed with Thibault. I concur that I don't like the proxy classes for long term. I'm picturing a custom cromwell wrapper that externally retries `Path.copy(): Unit`. Right now the proxy `.copy()` themselves are retrying internally while blocking the `Unit`. This custom wrapper should have a different behavior. Possibly a method returning a scala `Future`, even better Akka actors using messages and supervisors!

geoffjentry:
Or even a stream ;) at least I believe you can handle retry logic in streams

PR #1520, opened 2016-10-06
Print PID to stdout.background, SGE

### LINKED ISSUE #1512: SGE job number in stdout
Often we call a bash command in a task, read in the stdout, and set a variable to the string read in. When a cromwell job is submitted to SGE, however, cromwell prints the SGE job id to the stdout (don't know when or how often). When we then read from the stdout, the job id gets read in addition to the real output we are interested in. 

The example test wdl can be found in /humgen/gsa-hpprojects/dev/tsato/wdl/testSGE.wdl. The output of `GetBamFileName` starts with the job-id + newline character. We don't want them. The same workflow runs as expected on a GSAx server.

ruchim:
closes #1512

Bug fix -- Currently we have the SGE process ID printed to stdout, and instead this PR redirects it to a stdout.background file instead (mimicking the behavior of Local)

### DIFF
--- a/supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystemAsyncJobExecutionActor.scala
+++ b/supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystemAsyncJobExecutionActor.scala
@@ -231,7 +231,9 @@ trait SharedFileSystemAsyncJobExecutionActor
     * @return A process runner that will relatively quickly submit the script asynchronously.
     */
   def makeProcessRunner(): ProcessRunner = {
-    new ProcessRunner(processArgs.argv, jobPaths.stdout, jobPaths.stderr)
+    val stdout = pathPlusSuffix(jobPaths.stdout, "background")
+    val stderr = pathPlusSuffix(jobPaths.stderr, "background")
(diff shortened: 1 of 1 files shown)


cjllanwarne:
Marvolo! 👍

supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystemAsyncJobExecutionActor.scala
```diff
@@ -231,7 +231,9 @@ trait SharedFileSystemAsyncJobExecutionActor
     * @return A process runner that will relatively quickly submit the script asynchronously.
     */
   def makeProcessRunner(): ProcessRunner = {
-    new ProcessRunner(processArgs.argv, jobPaths.stdout, jobPaths.stderr)
+    val stdout = pathPlusSuffix(jobPaths.stdout, "background")
+    val stderr = pathPlusSuffix(jobPaths.stderr, "background")
```
kshakir: `"submit"` instead of `"background"`

kshakir:
For now 👍  assuming you switch the file extension.

ToL: This code may need some refactoring love (some functionality could be in the base SFS backend, some should migrate to the Config backend). I'm fine leaving it until somehow we need to touch it.

PR #1545, opened 2016-10-07
Wholly unsatisfying monad transformer PR.  Does not close #1212

### LINKED ISSUE #1212: Convert Future[Option[_]]s to use a monad transformer

mcovarr:
tl;dr I tried replacing existing `Future[Option[A]]`s with monad transformers as requested by #1212, but unfortunately found them not to be a good fit for our use cases.  I wouldn't recommend merging this PR in its current state.

Problems:

1) Monad transformer examples seem to break down into a "happy path" or at least "substantive path" (in the case of `Option`).  A value is either mapped/flatMapped out of the fused monadic context on the happy path, or processing short circuits on the unhappy path and eventually results in a one-size-fits-all error.  But this doesn't really fit the patterns of our APIs.  Most of the `Future[Option[A]]`s on database APIs break down into 3 distinct states in engine or API code, while for initialization actors a failed `Future` is the only error case; if the inner `Option` is `None` that's perfectly fine but is handled in deeper business logic.

2) There are a couple of spots included in this PR where I could apply monad transformers.  But IMHO the net effect of these changes is to "Waldo" these two APIs to be inconsistent with the others for reasons that are not obvious or particularly compelling.  Also by the point above, this will restrict the way in which callers can use these APIs going forward.

### DIFF
--- a/database/sql/src/main/scala/cromwell/database/slick/JobStoreSlickDatabase.scala
+++ b/database/sql/src/main/scala/cromwell/database/slick/JobStoreSlickDatabase.scala
@@ -1,5 +1,6 @@
 package cromwell.database.slick
 
+import cats.data.OptionT
 import cats.instances.future._
 import cats.syntax.functor._
 import cromwell.database.sql.JobStoreSqlDatabase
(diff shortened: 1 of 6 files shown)


geoffjentry:
As the person who filed the ticket, if you don't think it's worth doing feel free to ignore the ticket

PR #1554, opened 2016-10-07
Firstly, you must find... another shrubbery!  Closes #1211

### LINKED ISSUE #1211: Switch to using an off the shelf Config enhancer library
Reduce the reliance on lenthall

mcovarr:
https://www.getyarn.io/yarn-clip/ae3a7f44-c9f1-4caa-a83b-b3cbd40d874b

### DIFF
--- a/backend/src/main/scala/cromwell/backend/io/WorkflowPaths.scala
+++ b/backend/src/main/scala/cromwell/backend/io/WorkflowPaths.scala
@@ -5,14 +5,14 @@ import java.nio.file.{FileSystem, FileSystems, Path, Paths}
 import com.typesafe.config.Config
 import cromwell.backend.{BackendJobDescriptorKey, BackendWorkflowDescriptor}
 import cromwell.core.PathFactory
-import lenthall.config.ScalaConfig._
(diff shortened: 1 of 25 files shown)


geoffjentry:
Curious why ficus over configs (or something else)

mcovarr:
I chose ficus since its `as` API returns `A`s rather than config's `configs.Result[A]`s which would have been a lot more disruptive.  I didn't look at other alternatives.

cjllanwarne:
I like the syntax that this supports. It's pretty obvious at a very casual glance what's going on.
:+1: for me.

geoffjentry:
Yeah, it's definitely easier to get rolling. I found pretty quickly in Centaur that I was rolling my own extensions to Ficus (typeclasses IIRC?) which seemed to defeat the point. I switched to configs as it gives you more stuff out of the box.  The monad returns I've both enjoyed and shaken my fist at depending on the situation. I liked these two better than any of the rest I saw. 

I didn't look at the PR but if you didn't need to do any extensions I suppose it doesn't matter for now.

kshakir:
👍

PR #1559, opened 2016-10-11
Call caching output copy failures handling. Closes #1510

### LINKED ISSUE #1510: Invalidate calls that fail to be use as a call cache for future cache lookups.
If a call hits the cache but fails to use the cached call for whatever reason (output / detritus file not found etc...) , the cached call isn't invalidated so every other similar call will always fail to use the cache.

Horneth:
When a backend fails to copy outputs from a cache hit, invalidates that cache hit for the future - and try potential other hits found. Otherwise run the job normally.

### DIFF
--- a/core/src/main/scala/cromwell/core/PathCopier.scala
+++ b/core/src/main/scala/cromwell/core/PathCopier.scala
@@ -1,5 +1,6 @@
 package cromwell.core
 
+import java.io.IOException
 import java.nio.file.Path
 
 import better.files._
@@ -23,7 +24,12 @@ object PathCopier {
     */
   def copy(sourceFilePath: Path, destinationFilePath: Path): Unit = {
     Option(File(destinationFilePath).parent).foreach(_.createDirectories())
-    File(sourceFilePath).copyTo(destinationFilePath, overwrite = true)
-    ()
+    try {
+      File(sourceFilePath).copyTo(destinationFilePath, overwrite = true)
+      ()
+    } catch {
+      case ex: Exception =>
+        throw new IOException(s"Failed to copy ${sourceFilePath.toUri.toString} to ${destinationFilePath.toUri.toString}", ex)
+    }
   }
 }
--- a/core/src/main/scala/cromwell/core/PathCopier.scala
+++ b/core/src/main/scala/cromwell/core/PathCopier.scala
@@ -5,6 +5,8 @@ import java.nio.file.Path
 
 import better.files._
 
+import scala.util.{Failure, Try}
+
 object PathCopier {
   def getDestinationFilePath(sourceContextPath: Path, sourceFilePath: Path, destinationDirPath: Path): Path = {
     val relativeFileString = sourceContextPath.toAbsolutePath.relativize(sourceFilePath.toAbsolutePath).toString
@@ -14,7 +16,7 @@ object PathCopier {
   /**
     * Copies from a relative source to destination dir. NOTE: Copies are not atomic, and may create a partial copy.
     */
-  def copy(sourceContextPath: Path, sourceFilePath: Path, destinationDirPath: Path): Unit = {
+  def copy(sourceContextPath: Path, sourceFilePath: Path, destinationDirPath: Path): Try[Unit] = {
     val destinationFilePath = getDestinationFilePath(sourceContextPath, sourceFilePath, destinationDirPath)
     copy(sourceFilePath, destinationFilePath)
   }
@@ -22,14 +24,14 @@ object PathCopier {
   /**
     * Copies from source to destination. NOTE: Copies are not atomic, and may create a partial copy.
     */
-  def copy(sourceFilePath: Path, destinationFilePath: Path): Unit = {
+  def copy(sourceFilePath: Path, destinationFilePath: Path): Try[Unit] = {
     Option(File(destinationFilePath).parent).foreach(_.createDirectories())
-    try {
+    Try {
       File(sourceFilePath).copyTo(destinationFilePath, overwrite = true)
+
       ()
-    } catch {
-      case ex: Exception =>
-        throw new IOException(s"Failed to copy ${sourceFilePath.toUri.toString} to ${destinationFilePath.toUri.toString}", ex)
+    } recoverWith {
+      case ex => Failure(new IOException(s"Failed to copy ${sourceFilePath.toUri} to ${destinationFilePath.toUri}", ex))
     }
   }
 }
--- a/database/sql/src/main/scala/cromwell/database/slick/CallCachingSlickDatabase.scala
+++ b/database/sql/src/main/scala/cromwell/database/slick/CallCachingSlickDatabase.scala
@@ -51,7 +51,7 @@ trait CallCachingSlickDatabase extends CallCachingSqlDatabase {
                                (implicit ec: ExecutionContext): Future[Unit] = {
     import cats.syntax.functor._
     import cats.instances.future._
-    val action = dataAccess.allowResultReuseForCallCachingEntry(callCachingEntryId).update(false)
+    val action = dataAccess.allowResultReuseForCallCachingEntryId(callCachingEntryId).update(false)
     runTransaction(action) void
   }
 }
--- a/database/sql/src/main/scala/cromwell/database/slick/CallCachingSlickDatabase.scala
+++ b/database/sql/src/main/scala/cromwell/database/slick/CallCachingSlickDatabase.scala
@@ -46,4 +46,12 @@ trait CallCachingSlickDatabase extends CallCachingSqlDatabase {
 
     runTransaction(action)
   }
+
+  override def invalidateCall(callCachingEntryId: Int)
+                               (implicit ec: ExecutionContext): Future[Unit] = {
+    import cats.syntax.functor._
+    import cats.instances.future._
+    val action = dataAccess.allowResultReuseForCallCachingEntry(callCachingEntryId).update(false)
+    runTransaction(action) void
+  }
 }
--- a/database/sql/src/main/scala/cromwell/database/slick/tables/CallCachingEntryComponent.scala
+++ b/database/sql/src/main/scala/cromwell/database/slick/tables/CallCachingEntryComponent.scala
@@ -41,7 +41,7 @@ trait CallCachingEntryComponent {
     } yield callCachingEntry
   )
 
-  val allowResultReuseForCallCachingEntry = Compiled(
+  val allowResultReuseForCallCachingEntryId = Compiled(
     (callCachingEntryId: Rep[Int]) => for {
       callCachingEntry <- callCachingEntries
       if callCachingEntry.callCachingEntryId === callCachingEntryId
--- a/database/sql/src/main/scala/cromwell/database/slick/tables/CallCachingEntryComponent.scala
+++ b/database/sql/src/main/scala/cromwell/database/slick/tables/CallCachingEntryComponent.scala
@@ -40,4 +40,11 @@ trait CallCachingEntryComponent {
       if callCachingEntry.callCachingEntryId === callCachingEntryId
     } yield callCachingEntry
   )
+
+  val allowResultReuseForCallCachingEntry = Compiled(
+    (callCachingEntryId: Rep[Int]) => for {
+      callCachingEntry <- callCachingEntries
+      if callCachingEntry.callCachingEntryId === callCachingEntryId
+    } yield callCachingEntry.allowResultReuse
+  )
 }
--- a/database/sql/src/main/scala/cromwell/database/sql/CallCachingSqlDatabase.scala
+++ b/database/sql/src/main/scala/cromwell/database/sql/CallCachingSqlDatabase.scala
@@ -13,4 +13,7 @@ trait CallCachingSqlDatabase {
 
   def queryCallCaching(callCachingResultMetainfoId: Int)
                       (implicit ec: ExecutionContext): Future[Option[CallCachingJoin]]
+
+  def invalidateCall(callCachingResultMetainfoId: Int)
+                    (implicit ec: ExecutionContext): Future[Unit]
 }
--- a/database/sql/src/main/scala/cromwell/database/sql/CallCachingSqlDatabase.scala
+++ b/database/sql/src/main/scala/cromwell/database/sql/CallCachingSqlDatabase.scala
@@ -11,9 +11,9 @@ trait CallCachingSqlDatabase {
   def queryCallCachingEntryIds(hashKeyHashValues: NonEmptyList[(String, String)])
                               (implicit ec: ExecutionContext): Future[Seq[Int]]
 
-  def queryCallCaching(callCachingResultMetainfoId: Int)
+  def queryCallCaching(callCachingEntryId: Int)
                       (implicit ec: ExecutionContext): Future[Option[CallCachingJoin]]
 
-  def invalidateCall(callCachingResultMetainfoId: Int)
+  def invalidateCall(callCachingEntryId: Int)
                     (implicit ec: ExecutionContext): Future[Unit]
 }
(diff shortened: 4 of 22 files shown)


database/sql/src/main/scala/cromwell/database/slick/tables/CallCachingEntryComponent.scala
```diff
@@ -40,4 +40,11 @@ trait CallCachingEntryComponent {
       if callCachingEntry.callCachingEntryId === callCachingEntryId
     } yield callCachingEntry
   )
+
+  val allowResultReuseForCallCachingEntry = Compiled(
```
kshakir: `allowResultReuseForCallCachingEntry` -> `allowResultReuseForCallCachingEntryId`

core/src/main/scala/cromwell/core/PathCopier.scala
```diff
@@ -23,7 +24,12 @@ object PathCopier {
...
   def copy(sourceFilePath: Path, destinationFilePath: Path): Unit = {
     Option(File(destinationFilePath).parent).foreach(_.createDirectories())
-    File(sourceFilePath).copyTo(destinationFilePath, overwrite = true)
-    ()
+    try {
+      File(sourceFilePath).copyTo(destinationFilePath, overwrite = true)
+      ()
+    } catch {
+      case ex: Exception =>
+        throw new IOException(s"Failed to copy ${sourceFilePath.toUri.toString} to ${destinationFilePath.toUri.toString}", ex)
```
kshakir: The `toString`s aren't needed within `s"…"`

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheInvalidateActor.scala
```diff
@@ -0,0 +1,36 @@
...
+  implicit val ec: ExecutionContext = context.dispatcher
+
+  def receiver = context.parent
+
+  callCache.invalidate(cacheId) onComplete {
+    case Success(_) =>
+      receiver ! CallCacheInvalidatedSuccess
+      context.stop(self)
+    case Failure(t) =>
+      receiver ! CallCacheInvalidatedFailure(t)
```
ruchim: What does this failure imply?

Horneth: it means something went wrong at the DB level when trying to flip the allowResultReuse flag to 0. So the next time a call matches the hash of this call it will still be considered a valid cache hit.
Currently the EJEA doesn't actually do anything with except logging it though.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -337,6 +355,22 @@ class EngineJobExecutionActor(replyTo: ActorRef,
     goto(RunningJob) using data
   }
 
+  private def handleCacheInvalidatedResponse(response: CallCacheInvalidatedResponse, data: ResponsePendingData) = {
+    response match {
+      case CallCacheInvalidatedFailure(failure) => log.error(failure, "Failed to invalidate cache entry for job: {}", jobDescriptorKey)
```
ruchim: It is possible then that we end up with an invalidated cache that we retry to copy with?

Horneth: yes - we could retry the invalidation if it failed I guess. We have this problem with every DB transaction though I think (I don't think we ever retry if a transaction fail do we ?)

ruchim: I think it's fine if it doesn't get retried. Like you're saying, it seems like a retry for failed transactions is something more global and thus might be worth implementing in general instead of for this specific case.

core/src/main/scala/cromwell/core/PathCopier.scala
```diff
@@ -23,7 +24,12 @@ object PathCopier {
     */
   def copy(sourceFilePath: Path, destinationFilePath: Path): Unit = {
     Option(File(destinationFilePath).parent).foreach(_.createDirectories())
-    File(sourceFilePath).copyTo(destinationFilePath, overwrite = true)
-    ()
+    try {
+      File(sourceFilePath).copyTo(destinationFilePath, overwrite = true)
+      ()
+    } catch {
```
cjllanwarne: Is there a reason this can't convert the return type of `def copy` to `Try[Unit]` instead? I don't trust us to pay attention to what exceptions might be thrown by methods otherwise.

database/sql/src/main/scala/cromwell/database/slick/CallCachingSlickDatabase.scala
```diff
@@ -46,4 +46,12 @@ trait CallCachingSlickDatabase extends CallCachingSqlDatabase {
 
     runTransaction(action)
   }
+
+  override def invalidateCall(callCachingEntryId: Int)
+                               (implicit ec: ExecutionContext): Future[Unit] = {
+    import cats.syntax.functor._
+    import cats.instances.future._
+    val action = dataAccess.allowResultReuseForCallCachingEntry(callCachingEntryId).update(false)
+    runTransaction(action) void
```
cjllanwarne: What's this `void` doing?

database/sql/src/main/scala/cromwell/database/sql/CallCachingSqlDatabase.scala
```diff
@@ -13,4 +13,7 @@ trait CallCachingSqlDatabase {
 
   def queryCallCaching(callCachingResultMetainfoId: Int)
                       (implicit ec: ExecutionContext): Future[Option[CallCachingJoin]]
+
+  def invalidateCall(callCachingResultMetainfoId: Int)
```
cjllanwarne: ToL I don't think these are called `metainfo` in the DB any more

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -129,8 +130,8 @@ class EngineJobExecutionActor(replyTo: ActorRef,
       writeToMetadata(Map(callCachingReadResultMetadataKey -> "Cache Miss"))
       log.debug("Cache miss for job {}", jobTag)
       runJob(data)
-    case Event(hit @ CacheHit(cacheResultId), data: ResponsePendingData) =>
-      fetchCachedResults(data, jobDescriptorKey.call.task.outputs, hit)
+    case Event(hit: CacheHit, data: ResponsePendingData) =>
+      fetchCachedResults(jobDescriptorKey.call.task.outputs, hit.cacheResultIds.head) using data.withCacheHit(hit).popCacheHitId._2
```
cjllanwarne: ToL: I personally prefer the `using` to go with the `goto`. I think that's why the `data` was passed in originally.

Also ToL: I'd prefer to have the data have some method that calculates `data.withCacheHit(hit).popCacheHitId._2` with a nice easy-to-understand name

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -155,20 +156,22 @@ class EngineJobExecutionActor(replyTo: ActorRef,
...
       saveCacheResults(hashes, data.withSuccessResponse(response))
-    case Event(response: SucceededResponse, data @ ResponsePendingData(_, _, None)) if effectiveCallCachingMode.writeToCache =>
+    case Event(response: SucceededResponse, data @ ResponsePendingData(_, _, None, _)) if effectiveCallCachingMode.writeToCache =>
       // Wait for the CallCacheHashes
       stay using data.withSuccessResponse(response)
     case Event(response: SucceededResponse, data: ResponsePendingData) => // bad hashes or cache write off
       saveJobCompletionToJobStore(data.withSuccessResponse(response))
-    case Event(response: BackendJobExecutionResponse, data: ResponsePendingData) =>
+    case Event(response: BackendJobExecutionResponse, data @ ResponsePendingData(_, _, _, Some(cacheHit))) =>
       // This matches all response types other than `SucceededResponse`.
```
cjllanwarne: This comment is no longer true, since there's now an assumption that we've cacheHit'd.

We still need the catch-all case?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -186,6 +189,21 @@ class EngineJobExecutionActor(replyTo: ActorRef,
       stay using data.copy(hashes = Option(Failure(t)))
   }
 
+  when(InvalidatingCacheEntry) {
```
cjllanwarne: ToL: All this in the EJEA is making me more and more convinced a separate `DoACacheHit` is needed. Not in this PR though :)

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/ejea/EjeaBackendIsCopyingCachedOutputsSpec.scala
```diff
@@ -96,17 +98,17 @@ class EjeaBackendIsCopyingCachedOutputsSpec extends EngineJobExecutionActorSpec
         }
 
         RestartOrExecuteCommandTuples foreach { case RestartOrExecuteCommandTuple(operationName, restarting, expectedMessage) =>
-          s"$operationName the job immediately when it gets a failure result, and it was going to receive $hashComboName, if call caching is $mode" in {
+          s"invalidate a call for caching if backend coping failed when start mode is $operationName, and it was going to receive $hashComboName, if call caching is $mode" in {
```
cjllanwarne: This no longer seems to depend on the `RestartOrExecuteCommandTuples` - the result is the same either way. (I think that now only the `EjeaInvalidatingCacheEntrySpec ` needs to check that it works right depending on the restart mode)

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/ejea/EjeaBackendIsCopyingCachedOutputsSpec.scala
```diff
@@ -96,17 +98,17 @@ class EjeaBackendIsCopyingCachedOutputsSpec extends EngineJobExecutionActorSpec
...
             ejea = ejeaInBackendIsCopyingCachedOutputsState(initialHashData, mode, restarting = restarting)
             // Send the response from the copying actor
             ejea ! failureNonRetryableResponse
 
-            helper.bjeaProbe.expectMsg(awaitTimeout, expectedMessage)
-            ejea.stateName should be(RunningJob)
-            ejea.stateData should be(ResponsePendingData(helper.backendJobDescriptor, helper. bjeaProps, initialHashData))
+            expectInvalidateCallCacheActor(cacheId)
+            ejea.stateName should be(InvalidatingCacheEntry)
```
cjllanwarne: Is there a race condition where the `expectInvalidateCallCacheActor(cacheId)` might become true before the ejea changes state?
If so, add `eventually { ... }` around the `ejea.stateName should be(InvalidatingCacheEntry)`

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/ejea/EngineJobExecutionActorSpec.scala
```diff
@@ -40,6 +40,7 @@ trait EngineJobExecutionActorSpec extends CromwellTestkitSpec
     List(
       ("FetchCachedResultsActor", helper.fetchCachedResultsActorCreations),
       ("JobHashingActor", helper.jobHashingInitializations),
+      ("JobHashingActor", helper.invalidateCacheActorCreations),
```
cjllanwarne: Copy/paste typo: rename this

database/sql/src/main/scala/cromwell/database/slick/CallCachingSlickDatabase.scala
Horneth: It's turning the return value into a `Unit` since the return type of the method is `Future[Unit]` ( @mcovarr would know better )

database/sql/src/main/scala/cromwell/database/sql/CallCachingSqlDatabase.scala
kshakir: `callCachingEntryId`

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -320,14 +319,14 @@ class EngineJobExecutionActor(replyTo: ActorRef,
...
+    context.actorOf(FetchCachedResultsActor.props(callCachingEntryId, self, new CallCache(SingletonServicesStore.databaseInterface)))
     ()
   }
 
-  private def fetchCachedResults(taskOutputs: Seq[TaskOutput], metaInfoId: MetaInfoId) = {
-    makeFetchCachedResultsActor(metaInfoId, taskOutputs)
-    goto(FetchingCachedOutputsFromDatabase)
+  private def fetchCachedResults(taskOutputs: Seq[TaskOutput], callCachingEntryId: CallCachingEntryId, data: ResponsePendingData) = {
+    makeFetchCachedResultsActor(callCachingEntryId, taskOutputs)
+    goto(FetchingCachedOutputsFromDatabase) using data
```
cjllanwarne: Is this change intended? Just moving the `using` inside this function (from line 134) this would have given `using data.withCacheHit(hit).popCacheHitId._2`

EDIT: Just saw this is called with the NEW data from line 366?

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/ejea/EjeaInvalidatingCacheEntrySpec.scala
```diff
@@ -17,49 +17,36 @@ class EjeaInvalidatingCacheEntrySpec extends EngineJobExecutionActorSpec with Ca
...
       }
     }
 
-    cacheHitCombinations foreach { case combo @ CacheHitAndNextExpected(hitComboName, initialHitData, nextHitToTry, finalHitData) =>
-        List(invalidateSuccess, invalidateFailure) foreach { invalidateActorResponse =>
-          s"try the next available hit when there is a $hitComboName when response is $invalidateActorResponse" in {
-            ejea = ejeaInvalidatingCacheEntryState(Option(initialHitData), restarting = false)
-            // Send the response from the invalidate actor
-            ejea ! invalidateActorResponse
+    List(invalidateSuccess, invalidateFailure) foreach { invalidateActorResponse =>
```
cjllanwarne: If you invert the order of the two `foreach` blocks above, you could avoid having to have this second `foreach` here

cjllanwarne:
For pullapprove: 👍

kshakir:
Not 100% of the logic, but as a PR/step in the right direction: LGTM 👍

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
Horneth: Yes so the idea is that if data.cacheHit is `Some(NonEmptyList)`, then the head is the current cacheHit being processed. We need this because when we get the response from the backend copying cache outputs, it doesn't send back the cache hit ID, so we need to remember it in case it failed and we want to invalidate it.
At first I thought of the data.cacheHit value to mean `if the current cacheHit fails, these are the potential next ones to try`, but by doing this we would lose the information about what the current one actually is, hence the above change.

PR #1562, opened 2016-10-12
Modified the reference docker submit command to avoid bash redirection. Closes #1556

### LINKED ISSUE #1556: Run script.submit without bash input redirection
Cromwell version 0.21
Running on Openstack + ubuntu 16.04 instance
Local backend with a docker container

I was working on creating a simple WDL for this workflow: https://github.com/ICGC-TCGA-PanCancer/pcawg_delly_workflow

Which looks a little like this:
```
workflow PcawgDelly {
  call SeqwareWorkflow
}

task SeqwareWorkflow {

  String run_id
  File reference_gc
  File tumor_bam
  File normal_bam
  File reference_gz
  String delly_id = "embl-delly_1-3-0-preFilter.20150318"

  command {
    perl /usr/bin/run_seqware_workflow.pl \
      --run-id ${run_id} \
      --reference-gc ${reference_gc} \
      --tumor-bam ${tumor_bam} \
      --normal-bam ${normal_bam} \
      --reference-gz ${reference_gz}
  }

  runtime {
    docker: "delly-docker-root"
  }
}
```
and I'm leaving out some details, but you get the idea, it's very simple.

I would frequently get a failed Cromwell workflow, with an error in the logs like:
```
mv: cannot stat/root/PcawgDelly/e173fd52-3c15-4b87-bfec-087c7cf0a4ac/call-SeqwareWorkflow/execution/rc.tmp': No such file or directory`
```

I tried to come up with a minimal WDL that would reproduce the error, but so far I can only get it from this workflow, possibly because of running time, IO, perl, seqware, I don't know.

After much headdesking, I am nearly certain I have fixed the issue by changing the way cromwell executes a call. As you know, Cromwell generates a script.submit file which looks like this (in this case):
```bash
#!/bin/bash
docker run --rm -v /home/ubuntu/projects/debug-small-data/cromwell-executions/PcawgDelly/ebce533d-9f49-4e6d-8512-db7a182d3365/call-SeqwareWorkflow:/root/PcawgDelly/ebce533d-9f49-4e6d-8512-db7a182d3365/call-SeqwareWorkflow -i delly-docker-root /bin/bash < /home/ubuntu/projects/debug-small-data/cromwell-executions/PcawgDelly/ebce533d-9f49-4e6d-8512-db7a182d3365/call-SeqwareWorkflow/execution/script
```

By removing input redirection into bash (i.e. removing the "<" character and changing the paths) I can get this workflow to run consistently without error. The new script.submit looks like:

```bash
#!/bin/bash
docker run --rm -v /home/ubuntu/projects/debug-small-data-alex/cromwell-executions/PcawgDelly/bdc567a9-b1f7-4530-853a-5263b890ca6a/call-SeqwareWorkflow:/root/PcawgDelly/bdc567a9-b1f7-4530-853a-5263b890ca6a/call-SeqwareWorkflow -i delly-docker-root /bin/bash /root/PcawgDelly/bdc567a9-b1f7-4530-853a-5263b890ca6a/call-SeqwareWorkflow/execution/script
```

Basically, it runs the script from inside the docker container.

I cannot, unfortunately, describe why the bug happened in this seemingly rare case, other than pointing at the dangers of shell commands being interpreted from pipes/stdin and waving my hands a lot.

But, I do think avoiding input redirection of commands into bash is probably a good thing, in this case. Am I missing some case where it's necessary?

Luckily I think the fix is simple, and probably just involves updating the core conf. here:
core/src/main/resources/reference.conf

I was able to provide my own config file overrides for now as a workaround.

Thanks! Sorry for the super long bug report.

cjllanwarne:
On behalf of @abuchanan. This modification seems to address an edge case in command redirection.

Would appreciate a review from at least @kshakir and one other.

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -149,7 +149,7 @@ backend {
         run-in-background = true
         runtime-attributes = "String? docker"
         submit = "/bin/bash ${script}"
-        submit-docker = "docker run --rm -v ${cwd}:${docker_cwd} -i ${docker} /bin/bash < ${script}"
+        submit-docker = "docker run --rm -v ${cwd}:${docker_cwd} -i ${docker} /bin/bash ${docker_cwd}/execution/script"
 
         # Root directory where Cromwell writes job results.  This directory must be
         # visible and writeable by the Cromwell process as well as the jobs that Cromwell


cjllanwarne:
cf: #1556

kshakir:
👍

core/src/main/resources/reference.conf
```diff
@@ -149,7 +149,7 @@ backend {
         run-in-background = true
         runtime-attributes = "String? docker"
         submit = "/bin/bash ${script}"
-        submit-docker = "docker run --rm -v ${cwd}:${docker_cwd} -i ${docker} /bin/bash < ${script}"
+        submit-docker = "docker run --rm -v ${cwd}:${docker_cwd} -i ${docker} /bin/bash ${docker_cwd}/execution/script"
```
Horneth: Is there a reason why `${script}` was transformed into `${docker_cwd}/execution/script` ?

Horneth:
👍

core/src/main/resources/reference.conf
kshakir: When the command runs inside the container, the path to the script is the latter.

PR #1563, opened 2016-10-12
Beginnings of the AWS backend!

cjllanwarne:
Intended to avoid multiple people iterating on this having to regenerate the same set of placeholders:
 - Project definition in sbt
 - Basic ecs backend factory
 - Better defaults in the BackendLifecycleActorFactory trait
 - Fixed a few typos in nearby files... 😳


Horneth:
👍

build.sbt
```diff
@@ -43,6 +43,12 @@ lazy val sfsBackend = (project in backendRoot / "sfs")
   .dependsOn(gcsFileSystem)
   .dependsOn(backend % "test->test")
 
+lazy val ecsBackend = (project in backendRoot / "ecs")
```
mcovarr: Maybe `amazonBackend` would be more intuitive?  Unless we expect multiple Amazon-flavored backends.

mcovarr:
:+1:

build.sbt
kcibul: I think ecs is appropriate for now.  it's possible we might have other backends (and even this one might not be called ecs...)  But if we need to change it later, it wouldn't be the end of the world

kcibul: AWESOME to see this started though!!

mcovarr: Thirded, the motion carries.

PR #1580, opened 2016-10-14
Added index and attempt number to failure messages. Closes #1479

### LINKED ISSUE #1479: Failure message for failed workflow should specify the shard number for scattered jobs
Looking at status message in the swagger GUI, I see that my workflow failed.  However, the job it specifies is scattered and I do not know which shard is the issue....

If there is an easy way to find this, that I missed, then apologies.

Example:
```
....
"status": "Failed",
  "failures": [
    {
      "message": "Call case_gatk_acnv_workflow.TumorCalculateTargetCoverage: return code was -1"
    }
  ],
....
```

cjllanwarne:


engine/src/test/scala/cromwell/SimpleWorkflowActorSpec.scala
```diff
@@ -111,7 +111,7 @@ class SimpleWorkflowActorSpec extends CromwellTestkitSpec with BeforeAndAfter {
     }
 
     "fail when a call fails" in {
-      val expectedError = "Call goodbye.goodbye: return code was 1"
+      val expectedError = "Call goodbye.goodbye:-1:1: return code was 1"
```
mcovarr: This brings to the fore that `JobKey#toString` is kind of lame, and this test is dependent on its format.  🙁

mcovarr:
:+1:

geoffjentry:
👍

PR #1583, opened 2016-10-14
Added transpose for Array[Array[X]]. Closes #1463

### LINKED ISSUE #1463: Array Transpose WDL Function
Extremely useful in conjunction with Nested Scatter ( #838 ).  A typical use case is to first scatter by sample, and then for each sample scatter by genomic location (e.g. running haplotype caller).  After that I have a two-dimensional array (Array[Array[]] ) in "sample-major" order (e.g. the first dimension is the sample).

However, next I often want to then do something across all sample for each genomic region (e.g. run GenotypeGVCFs).  That is, scatter by region and do something with all the samples.

To do this I would like a transpose() engine function that transposes the supplied array.  For example, using the notation of  "<sample><region>" -> e.g. sample_A_region_1 -> A1

My first. Array[Array[]] in "sample-major" order after a scatter based on sample, followed by a scatter on genomic location (region) would be

A = [
  [A1, A2, A3, A4],
  [B1, B2, B3, B4],
  [C1, C2, C3, C4],
]

and then after the  transpose to a "region-major" order so that we can scatter by region to combine/process all the data in that region I would like

transpose(A) = [
  [A1, B1, C1, D1],
  [A2, B2, C2, D2],
  [A3, B3, C3, D3]
]

The primary use case is transposing a two dimensional array, and it is fine if it is restricted to that.  Generalizing to n-dimensions is more complex as you have to specify the axis, and is much less common at this point

cjllanwarne:
Relies on https://github.com/broadinstitute/wdl4s/pull/37

### DIFF
--- a/backend/src/main/scala/cromwell/backend/wdl/PureFunctions.scala
+++ b/backend/src/main/scala/cromwell/backend/wdl/PureFunctions.scala
@@ -1,7 +1,7 @@
 package cromwell.backend.wdl
 
 import wdl4s.expression.WdlStandardLibraryFunctions
-import wdl4s.types.{WdlArrayType, WdlIntegerType, WdlStringType}
+import wdl4s.types.{WdlArrayType, WdlIntegerType, WdlStringType, WdlType}
 import wdl4s.values.{WdlArray, WdlFile, WdlFloat, WdlInteger, WdlString, WdlValue}
 
 import scala.util.{Failure, Success, Try}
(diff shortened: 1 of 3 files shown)


geoffjentry:
👍

backend/src/main/scala/cromwell/backend/wdl/PureFunctions.scala
```diff
@@ -20,10 +20,39 @@ case object OnlyPureFunctions extends WdlStandardLibraryFunctions with PureFunct
 
 trait PureFunctions { this: WdlStandardLibraryFunctions =>
 
+  def transpose(params: Seq[Try[WdlValue]]): Try[WdlArray] = {
+    def extractExactlyOneArg: Try[WdlValue] = params.size match {
+      case 1 => params.head
+      case n => Failure(new IllegalArgumentException(s"Invalid number of parameters for engine function transpose: $n. Ensure transpose(x: Array[Array[X]]) takes exactly 1 parameters."))
```
mcovarr: 1 parameters => 1 parameter

mcovarr:
:+1:

kcibul:
This looks great -- two questions

1. what happens if you pass in a >2-dimensional array?

2. can we add some docs somewhere a user can see them (I think https://github.com/broadinstitute/wdl/blob/develop/SPEC.md#standard-library might be the right place)

cjllanwarne:
@kcibul
1) It works as though the higher (lower?) dimensions are just any old value. So it transposes `Array[Array[ <<Array[Int]>> ]]` exactly the same way it transposes `Array[Array[ <<Int>> ]]`

2) Great idea. I'll do that

PR #1586, opened 2016-10-17
Bumped upcoming version. Added changelog entry for 1.0.0

cjllanwarne:


geoffjentry:
Assuming you rooted out all of the `0.` references going forward 👍

mcovarr:
:+1: yeah I math real good

geoffjentry:
:-1: to the 1.0.0 vs 23

geoffjentry:
To explain, "1.0" is a very loaded term in my mind. People have already been ignoring the `0.` colloquially anyways, I'd rather just drop it and move on with life

geoffjentry:
rescinded my thumb and the two thumbs up label until the actual number issue is resolved.

cjllanwarne:
I agree with @geoffjentry. Going to merge as 23.

I think that's the way our versioning has been going anyway (even if we've been timidly adding 0. to the start of our version numbers). And semver is nowhere near the horizon (or even particularly wanted by the sounds of it).

PR #1588, opened 2016-10-17
JABJEA failure catcher. Closes #1435

### LINKED ISSUE #1435: Exceptions thrown in JABJEA <init> not handled, hang workflow
We saw during JG testing that a stray non-`lazy` instance variable in `JesAsyncBackendJobExecutionActor` led to an exception being thrown during `<init>`.  This exception wasn't handled by any supervisor nor by JABJEA itself, so the JABJEA (and possibly `JesJobExecutionActor` and other actors) simply crashed.  The net result was the `WorkflowExecutionActor` stayed `Running` forever, waiting for a terminal job message that would never come.

cjllanwarne:


mcovarr:
:+1: assuming JJEA's supervisor behaves reasonably when JJEA stops.

cjllanwarne:
@mcovarr if I'm reading your comment right, that's a supervisor strategy so it `Stop`s the `JABJEA` rather than the `JJEA`. The progression of the `JJEA` is mostly unchanged by this PR

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesJobExecutionActor.scala
```diff
@@ -95,4 +98,12 @@ case class JesJobExecutionActor(override val jobDescriptor: BackendJobDescriptor
   }
 
   override def abort(): Unit = {}
+
+  // Supervision strategy: if the JABJEA throws an exception, stop the actor and fail the job.
+  def jobFailingDecider: Decider = {
+    case e: Exception =>
+      completionPromise.failure(new RuntimeException("JesAsyncBackendJobExecutionActor failed and didn't catch its exception.", e))
```
kshakir: ToL: `Promise.tryFailure` is more forgiving. I do not know if it applies here for the JES backend Promises/Futures/Actors, but `Promise.failure` [throws exceptions](http://docs.scala-lang.org/overviews/core/futures.html) if the promise is already completed for some unexpected reason (the scala `Promise` is completed in some other thread, then an exception escapes elsewhere in the akka supervised `Actor`).

"ToL" because I don't know off the top of my head what happens if this `Decider` throws an exception.

cjllanwarne: I think `tryFailure` is better, on the off chance it completes and then throws a tidy-up exception somehow.

Changed!

kshakir:
👍

PR #1592, opened 2016-10-18
Added invalidate-bad-cache to call caching options. Closes #1587

### LINKED ISSUE #1587: Option to skip (but don't invalidate) call cache hits that fail for permission errors
As FireCloud, I would like to enable call caching.  Because this is a multi-tenant system, and Cromwell knows nothing about users, I may try to call cache to a call where I don't have permission to read the files.  Currently (due to #1510) this would cause that cache hit to be ejected from the cache, and thus unavailable to users who do have the right file permissions.

I would like to have a option in cromwell which would allow me to disable this invalidation step.  When I attempt to call cache, and the cache hit fails due to a permission problem I simply attempt to cache using the next hit (if there are any).

@abaumann -- please comment with any changes

cjllanwarne:


mcovarr:
Update README, including documenting what the default behavior is if unspecified.

core/src/main/resources/reference.conf
```diff
@@ -97,7 +97,12 @@ workflow-options {
 
 // Optional call-caching configuration.
 call-caching {
+  # Allows re-use of existing results for jobs you've already run
```
mcovarr: Document default values here and below.  The defaults might be a little surprising given that this config explicitly sets them to their default values.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -384,6 +381,22 @@ class EngineJobExecutionActor(replyTo: ActorRef,
     ()
   }
 
+  private def invalidateCacheHitAndTransition(cacheId: CallCachingEntryId, data: ResponsePendingData, reason: Throwable) = {
+    val invalidationRequired = effectiveCallCachingMode match {
+      case CallCachingOff =>
+        log.error("Should not be calling invalidateCacheHit if call caching is off!")
```
mcovarr: include job descriptor key in message

mcovarr:
😡  apparently thumbs in approval messages don't count so here's another :+1:

core/src/main/resources/reference.conf
cjllanwarne: I think that's fairly standard in config files, to have commented-out options being set to their default value?

mcovarr: ok nm then

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
mcovarr: I probably wouldn't bother with the exception since it doesn't look like the stack trace would be informative.  And yeah, this will never happen anyway.  😄

Horneth:
👍

PR #1609, opened 2016-10-21
Robust METADATA_VALUE embiggening.  Closes #1607.

### LINKED ISSUE #1607: embiggen texty columns
Joel attempted a 0.19 -> 0.21 migration on a Firecloud dev database and hit an overflow migrating call outputs:

```
Oct 20 19:11:12 gce-cromwell-dev203 docker/cromwell-app[5085]: Caused by: liquibase.exception.MigrationFailedException: Migration failed for change set changesets/metadata_migration.xml::call_output_symbol_table_migration::tjeandet:
Oct 20 19:11:12 gce-cromwell-dev203 docker/cromwell-app[5085]:     Reason: liquibase.exception.UnexpectedLiquibaseException: liquibase.exception.CustomChangeException: Could not apply migration script for metadata at CallOutputSymbolTableMigration
.
.
.
Caused by: com.mysql.jdbc.MysqlDataTruncation: Data truncation: Data too long for column 'METADATA_VALUE' at row 1

```

This column is currently of type `TEXT` which has a size limit of 2^16 bytes.  Perusing the MySQL docs I see that `MEDIUMTEXT` supports 2^24 bytes and `LONGTEXT` supports 2^32 bytes, and these consume only 1 and 2 additional bytes over `TEXT` respectively.

Looking at our DB schema I'm also noticing that 

`CALL_CACHING_DETRITUS_ENTRY.DETRITUS_VALUE` has type `VARCHAR(255)`

which I could see ending badly with long GCS paths.

mcovarr:
Will need to be picked over to an 0.21ish branch for the Firecloud migration @jmthibault79 is working on.

### DIFF
--- a/database/migration/src/main/resources/changesets/embiggen_metadata_value.xml
+++ b/database/migration/src/main/resources/changesets/embiggen_metadata_value.xml
@@ -1,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.3.xsd">
+
+    <changeSet author="mcovarr" id="entry_or_journal_existence_xor">
+        <preConditions>
+            <or>
+                <and>
+                    <tableExists tableName="METADATA_JOURNAL"/>
+                    <not>
+                        <tableExists tableName="METADATA_ENTRY"/>
+                    </not>
+                </and>
+                <and>
+                    <tableExists tableName="METADATA_ENTRY"/>
+                    <not>
+                        <tableExists tableName="METADATA_JOURNAL"/>
+                    </not>
+                </and>
+            </or>
+        </preConditions>
+        <comment>
+            Either METADATA_JOURNAL or METADATA_ENTRY must exist, but not both (Liquibase doesn't have an xor).
+        </comment>
+    </changeSet>
+
+    <changeSet author="mcovarr" id="embiggen_metadata_entry">
+        <preConditions onFail="MARK_RAN">
+            <tableExists tableName="METADATA_ENTRY"/>
+        </preConditions>
+        <comment>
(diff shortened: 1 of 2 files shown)


database/migration/src/main/resources/changesets/embiggen_metadata_value.xml
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.3.xsd">
+
+    <changeSet author="mcovarr" id="entry_or_journal_existence_xor">
+        <preConditions>
```
Horneth: Shouldn't this be directly in the changelog and not in the changeset ?

mcovarr: The current code appears to work but I think the suggested version would read a lot more intuitively, so I'll try that.

mcovarr: tl;dr it looks like it needs to stay the way it is.

It turns out that if the precondition is put at the changelog level, Liquibase will try to run it before running any changesets.  Since neither METADATA_JOURNAL nor METADATA_ENTRY would exist on a new database, the overall migration would incorrectly fail.

Horneth: Wouldn't a a `onFail="MARK_RAN"` fix that ?
Anyway if it works as is don't bother :)

Horneth:
👍

kshakir:
LGTM if it works for @jmthibault79 👍

mcovarr:
Per discussion in 10/25 standup, waiting to hear how the migration goes on the prod Firecloud snapshot before merging this.

jmthibault79:
OK will keep you updated.  Looping in @abaumann as well.

mcovarr:
Should anyone here present know of any reason that these branches should not be joined in holy mergemony, speak now or forever hold your peace.

jmthibault79:
Dev and Prod migration testing went well.  It will be a while before we do the real Prod migration.

mcovarr:
Cool, proceeding with merge.

PR #1640, opened 2016-11-01
JES waits for glob results. Closes #1395

### LINKED ISSUE #1395: JES Globbing should use new glob count to beat eventual consistency
Currently, to process a glob on JES, Cromwell does an `ls` of the google cloud storage location.  The problem with this is that ls is eventually consistent, which leads to bugs like #843 

JES has added a feature (#28858407) where they now return the number of files that matched the glob as part of their metadata.  These appear as events of the form

`{Description: "copied 3 file(s) to \"gs://my-bucket/out/\"",
           StartTime:   {Seconds: 1470063955,
                         Nanos:   748725437}},`

In Cromwell, when processing these globs, we should poll (with adjustable maximum timeout) for this number of files to appear via the ls.  If they do not appear after the timeout, the task should fail with an appropriate error message.

If we are processing globs on GCS and NOT using JES, the best we can do is just grab and go via the ls (as we are doing currently for JES).

cjllanwarne:
Alas, the directory list code is unbatchable and synchronous...

I guess we could make this a work-pulling glob counter if we need to scale better. In the meantime, I'm blocking the JABJEA's EC with a Future.


geoffjentry:
Instead of creating a future in the actor could you create a separate actor which would perform the blocking operation and pass it back?

cjllanwarne:
@geoffjentry yeah but assuming it's on the same EC that doesn't actually solve anything. 

It'd be much better to have... (wait for it....)... a work pulling system that limits how many of these are going on at once. Now admittedly that's an easier transition if you already have the actors in place but unless you want the full scalability fix here, I suggest we address that as part of the scalability sprint (and I'd be very happy to do it then!)

geoffjentry:
My objection was more about crossing the streams by putting futures in actors

geoffjentry:
nm, now that i actually look at it i see it's due to those blasted Futures in the signatures

I don't think I agree with just throwing up hands and saying "here's a potential explosion, let's just wait until we work on scaling" as this causes the very real possibility of making scalability much worse. On an 8 core system (which is what I think FC uses) it wouldn't take very many of these to wreak massive havoc. Also I don't see how it could be more than a small amount of work to go all the way.

Tagging @kcibul in case he wants to overrule me from a product perspective but the number of times we've wound up in trouble when we just handwave situations like this makes me uneasy.

Horneth:
There is I think a way to make the list queries part of a batch although I'm not sure how nicely it would play with the existing batch poll status.

kcibul:
From a PO perspective -- globs will be used at scale within FireCloud and
GOTC... so they've got to scale at least to those workloads.  I'm not sure
if you're saying this is or isn't a problem... but that's the need

Kristian Cibulskis
kcibul@broadinstitute.org

cjllanwarne:
@geoffjentry @kcibul ok... let's tech talk at stand up :)

cjllanwarne:
Spoke in person; not as bad as Jeff thought it was... Ready for review!

cjllanwarne:
Closing for additional work

PR #1643, opened 2016-11-01
Add logging to SFS localization functions.

francares:
There is a requirement from CCC to be able to identify through logs why localization functionality fallback to an specific strategy.

A simple solution would be to log failures in each strategy invoke but I'm open to any other alternative.

OPEN: I made use of StrictLogging but if you think I should use a different implementation, just let me know and I will change it.

### DIFF
--- a/supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystem.scala
+++ b/supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystem.scala
@@ -5,7 +5,6 @@ import java.nio.file.{Path, Paths}
 import cats.instances.try_._
 import cats.syntax.functor._
 import com.typesafe.config.Config
-import com.typesafe.scalalogging.StrictLogging
 import cromwell.backend.io.JobPaths
 import cromwell.core._
 import cromwell.core.path.PathFactory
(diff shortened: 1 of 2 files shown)


supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystem.scala
```diff
@@ -33,91 +34,24 @@ object SharedFileSystem {
...
 
   def pathPlusSuffix(path: File, suffix: String) = path.sibling(s"${path.name}.$suffix")
 }
 
-trait SharedFileSystem extends PathFactory {
+trait SharedFileSystem extends PathFactory with StrictLogging {
   import SharedFileSystem._
   import better.files._
 
-  def sharedFileSystemConfig: Config
```
geoffjentry: Why are you doing things which amount to just moving stuff around?

francares: I moved localization functions from the object to the trait mostly due to logging trait but also due to trait responsibilities.
For this specific function, I took the opportunity to order based on code convention (move function after trait attributes and move private functions after public functions), nothing else.
If you don't like those changes I can revert them and make the object extends from logging trait but still need an answer on the logging part which is the main point of the PR.

supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystem.scala
```diff
@@ -224,4 +190,45 @@ trait SharedFileSystem extends PathFactory {
...
+    executionPath.parent.createDirectories()
+    val executionTmpPath = pathPlusSuffix(executionPath, ".tmp")
+    val result = Try(originalPath.copyTo(executionTmpPath, overwrite = true).moveTo(executionPath, overwrite = true)).void
+    if (result.isFailure) logger.warn(s"Localization via copy has failed: ${result.failed.get.getMessage}")
+    result
+  }
+
+  private def localizePathViaHardLink(originalPath: File, executionPath: File): Try[Unit] = {
+    executionPath.parent.createDirectories()
+    val result = Try(executionPath.linkTo(originalPath, symbolic = false)).void
```
francares: @geoffjentry please check localize... functions and you will realize about required logging changes.

supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystem.scala
```diff
@@ -224,4 +190,45 @@ trait SharedFileSystem extends PathFactory {
...
+    if (result.isFailure) logger.warn(s"Localization via hard link has failed: ${result.failed.get.getMessage}")
+    result
+  }
+
+  private def localizePathViaSymbolicLink(originalPath: File, executionPath: File): Try[Unit] = {
+    if (originalPath.isDirectory) Failure(new UnsupportedOperationException("Cannot localize directory with symbolic links"))
+    else {
+      executionPath.parent.createDirectories()
+      val result = Try(executionPath.linkTo(originalPath, symbolic = true)).void
+      if (result.isFailure) logger.warn(s"Localization via symbolic link has failed: ${result.failed.get.getMessage}")
```
Horneth: Could you use the log method that takes the exception as an argument as well ? I often find that the exception message only is not super useful.

supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystem.scala
francares: OK, np... I will revert those changes and just keep it to minimal. It will not look good but if this helps, there is no problem from my side.

supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystem.scala
```diff
@@ -45,19 +46,25 @@ object SharedFileSystem {
   private def localizePathViaCopy(originalPath: File, executionPath: File): Try[Unit] = {
     executionPath.parent.createDirectories()
     val executionTmpPath = pathPlusSuffix(executionPath, ".tmp")
-    Try(originalPath.copyTo(executionTmpPath, overwrite = true).moveTo(executionPath, overwrite = true)).void
+    val result = Try(originalPath.copyTo(executionTmpPath, overwrite = true).moveTo(executionPath, overwrite = true)).void
+    if (result.isFailure) logger.warn(s"Localization via copy has failed: ${result.failed.get.getMessage}", result.failed.get)
```
cjllanwarne: I think there's an opportunity to DRY this with a `private def copyOrElseLog` but I'm not too offended with this as-is

cjllanwarne:
👍

gauravs90:
Looks sufficient from the requirements from our side. Since @Horneth's the official reviewer, requesting him to have a final look and a thumbs up.

Horneth:
👍

supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystem.scala
```diff
@@ -45,22 +46,27 @@ object SharedFileSystem {
...
   private def localizePathViaHardLink(originalPath: File, executionPath: File): Try[Unit] = {
     executionPath.parent.createDirectories()
-    Try { executionPath.linkTo(originalPath, symbolic = false) } void
+    logOnFailure(Try(executionPath.linkTo(originalPath, symbolic = false)).void, "hard link")
   }
 
   private def localizePathViaSymbolicLink(originalPath: File, executionPath: File): Try[Unit] = {
       if (originalPath.isDirectory) Failure(new UnsupportedOperationException("Cannot localize directory with symbolic links"))
       else {
         executionPath.parent.createDirectories()
```
francares: Yes, can do that change in all the three functions.

PR #1647, opened 2016-11-03
Encrypting and clearing workflow options.

kshakir:


database/migration/src/main/scala/cromwell/database/migration/custom/BatchedTaskChange.scala
```diff
@@ -0,0 +1,140 @@
+package cromwell.database.migration.custom
```
kshakir: ToL: I'm open to other names for this package. I stopped myself from throwing it in yet-another generic `*.util` package.

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreActor.scala
```diff
@@ -117,6 +118,38 @@ case class WorkflowStoreActor(store: WorkflowStore, serviceRegistryActor: ActorR
...
+
+  private def processSources(sources: NonEmptyList[WorkflowSourceFiles],
+                             processOptions: WorkflowOptions => WorkflowOptionsJson):
+  Try[NonEmptyList[WorkflowSourceFiles]] = {
+    val nelTries: NonEmptyList[Try[WorkflowSourceFiles]] = sources map processSource(processOptions)
+    val seqTries: Seq[Try[WorkflowSourceFiles]] = nelTries.list.toList
+    val trySeqs: Try[Seq[WorkflowSourceFiles]] = TryUtil.sequence(seqTries)
+    import scalaz.Scalaz._
+    val tryNel: Try[NonEmptyList[WorkflowSourceFiles]] = trySeqs.map(_.toList.toNel.get)
+    tryNel
```
kshakir: ToL: Also open to folks helping with this HoF soup. Of course will be a different brand of soup for 0.22_hotfix and develop due to scalaz->cats.

mcovarr:
:+1:

database/migration/src/main/scala/cromwell/database/migration/custom/BatchedTaskChange.scala
```diff
@@ -0,0 +1,140 @@
...
+    */
+  def migrateBatchQuery: String
+
+  /**
+    * Migrate a row.
+    *
+    * Read the values from readRow, update the values, set the updated values on the migrateStatement, and then call
+    * migrateStatement.addBatch(). Return the (estimated) number of rows to be written by this batch.
+    *
+    * @param readRow          The row to migrate
```
kshakir: I thought about changing the name to better match the type (this code is copied from elsewhere). But I like that the name conveys that the implementor _is going to have a bad time_ if they call any other methods on the `ResultSet` besides accessing the current row values.

geoffjentry:
Just to make sure I'm understanding correctly, all the stuff in the migration package is a one time run, correct?

👍

kshakir:
Yes, migration runs once as a changeset, and liquibase records that changeset as done within its tables.

PR #1652, opened 2016-11-07
SingleWorkflowMode support for WDL import directories closes #1525

### LINKED ISSUE #1525: Implement Imports for Single Workflow mode

ruchim:


engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -299,14 +299,18 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef, val wor
...
+      source.importsFile match {
+        case Some(dir) => WdlNamespaceWithWorkflow.load(source.wdlSource, dir).validNel
+        case None => WdlNamespaceWithWorkflow.load(source.wdlSource).validNel
+      }
     } catch {
       case e: Exception => s"Unable to load namespace from workflow: ${e.getMessage}".invalidNel
     }
   }
 
+
```
mcovarr: remove extra line

src/main/scala/cromwell/CromwellCommandLine.scala
```diff
@@ -38,12 +38,14 @@ object RunSingle {
...
     val metadataPath = argPath(args, 3, None)
+    val importPath = argPath(args, 4, None)
 
     val wdl = readContent("WDL file", wdlPath)
     val inputsJson = readJson("Inputs", inputsPath)
     val optionsJson = readJson("Workflow Options", optionsPath)
+    val importsFile = validateImportsDirectory(importPath)
 
-    val sourceFiles = (wdl |@| inputsJson |@| optionsJson) map { WorkflowSourceFiles.apply }
+    val sourceFiles = (wdl |@| inputsJson |@| optionsJson |@| (importsFile) ) map { WorkflowSourceFiles.apply }
```
mcovarr: remove parens around `importsFile`

mcovarr: I don't think the curlies around `WorkflowSourceFiles.apply` are necessary

src/main/scala/cromwell/CromwellCommandLine.scala
```diff
@@ -59,6 +61,32 @@ object RunSingle {
     }
   }
 
+  private def validateImportsDirectory(path: Option[Path]): ErrorOr[Option[File]] = {
+    def isZipped(f: File) = { f.pathAsString.endsWith(".zip") }
```
mcovarr: unnecessary curlies

src/main/scala/cromwell/CromwellCommandLine.scala
```diff
@@ -59,6 +61,32 @@ object RunSingle {
...
+    def unZippedDirectory(f: File): File = {
+      val unzippedFile = f.unzip()
+      //Needs prettification
```
mcovarr: now or never 😛

src/main/scala/cromwell/CromwellCommandLine.scala
```diff
@@ -59,6 +61,32 @@ object RunSingle {
...
+      if (unzippedFile.toJava.listFiles().head.isDirectory) {
+        File(unzippedFile.toJava.listFiles().head.getPath)
+      }
+      else { unzippedFile }
+    }
+
+    val importsDirectory: Option[File] = path match {
+      case Some(p) if isZipped(File(p)) => Option(unZippedDirectory(File(p)))
+      case Some(p) if !isZipped(File(p)) => Option(File(p))
```
mcovarr: the `if !isZipped(...)` is unnecessary

src/main/scala/cromwell/CromwellCommandLine.scala
```diff
@@ -59,6 +61,32 @@ object RunSingle {
...
+
+    importsDirectory match {
+      case Some(file) if !file.isDirectory => s"Unable to import workflows as the given path is not a directory: ${file.pathAsString}".invalidNel
+      case Some(file) if file.isDirectory && file.isEmpty => s"Unable to import workflows as the given path is an empty directory: ${file.pathAsString}".invalidNel
+      case Some(file) => Option(file).validNel
+      case None => None.validNel
+    }
+  }
+
+
```
mcovarr: remove extra newline

mcovarr:
:+1:

ruchim:
@mcovarr Need to update the readme and changelog, thanks for the reminder :)

core/src/main/scala/cromwell/core/WorkflowSourceFiles.scala
```diff
@@ -1,9 +1,11 @@
...
+import better.files.File
 import wdl4s.{WdlJson, WdlSource}
 
 /**
   * Represents the collection of source files that a user submits to run a workflow
   */
 final case class WorkflowSourceFiles(wdlSource: WdlSource, inputsJson: WdlJson,
-                                     workflowOptionsJson: WorkflowOptionsJson)
+                                     workflowOptionsJson: WorkflowOptionsJson,
+                                     importsFile: Option[File] = None)
```
geoffjentry: In situations where there's a case class w/ a single optional field I prefer to see them structured as an ADT sum type where the two implementations are such that one does not have an importsFile and the other has a nonOptional one. 

In situations where there are more than one option the code gets too messy to deal with and I give up :)

geoffjentry:
👍

core/src/main/scala/cromwell/core/WorkflowSourceFiles.scala
ruchim: I made this change and as I was doing it, I realized basically all WorkflowSourceFiles will require the imports path, because eventually they will be available through both server/run mode and thus it might make more sense to keep the optional file parameter, but not default it to None.

core/src/main/scala/cromwell/core/WorkflowSourceFiles.scala
```diff
@@ -1,9 +1,28 @@
...
+
+  def defaultImports: WorkflowSourceFilesWithImports = {
+    WorkflowSourceFilesWithImports(wdlSource, inputsJson, workflowOptionsJson, None)
+  }
+}
+
+final case class WorkflowSourceFilesWithImports(wdlSource: WdlSource,
+                                                inputsJson: WdlJson,
+                                                workflowOptionsJson: WorkflowOptionsJson,
+                                                importsFile: Option[File]) extends WorkflowSourceFilesCollection
```
cjllanwarne: If I understand this change correctly, if we have the `WorkflowSourceFilesWithImports ` type, we should disambiguate based on `WorkflowSourceFilesWithImports ` vs `WorkflowSourceFiles ` rather than `Some` vs `None`. So this doesn't need to be an `Option`

ToL: Personally I wouldn't mind a single `case class` with an `Option`, but if we have both we should use both

geoffjentry: I asked for this but really for what you said Chris - if you set it up as an ADT sum type one variant should have a `File` and one should have nothing at all. 

I don't like seeing case classes with an Option in them. Sadly I've never found a reasonable approach for ones which have multiple Options, but we can at least do the right thing when there's a single Option.

@ruchim I can explain in person more what I meant and mean now

PR #1653, opened 2016-11-08
Doc section on "security by sysadmin"

delocalizer:
As per discussion on gitter https://gitter.im/broadinstitute/cromwell?at=5818fee6806316005de7615c

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,6 @@ A [Workflow Management System](https://en.wikipedia.org/wiki/Workflow_management
   * [Database](#database)
   * [SIGINT abort handler](#sigint-abort-handler)
 * [Security](#security)
- * [Multi-tenant](#multi-tenant)
 * [Backends](#backends)
   * [Backend Filesystems](#backend-filesystems)
     * [Shared Local Filesystem](#shared-local-filesystem)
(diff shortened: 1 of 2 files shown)


kcibul:
Thanks for the contribution!  @katevoss @geoffjentry can we get a quick review?

cjllanwarne:
IMO if we put any security recommendations in our README we should be very careful to disclaimer it. **Heavily**. Maybe with something along the lines of:
```
Warning! 
- Only YOU are responsible for your own security! 
- Cromwell is NOT a security appliance! 
- What follows are ideas and starting points and not necessarily a secure system configuration in your situation"
```

delocalizer:
I'm with @cjllanwarne in favouring some form of disclaimer on this, wanting as little as you to be or feel actually responsible for others' security!

geoffjentry:
@delocalizer Just so it doesn't seem like I'm being lazy, I wanted to chime in on some of the stuff this has brought up (e.g. the disclaimer) but wanted to chat w/ @katevoss first and we just haven't been able to be mutually free yet today. Worst case it'll happen tomorrow and we can get everything sorted.

Thanks for doing this though, now I'll have to actually make good on my side of the deal ;)

geoffjentry:
Discussed with @katevoss - as we don't have our long range documentation plan hashed out yet for now we'll be:

- Making a separate doc (name tbd) with two initial bits. One is the content from @delocalizer describing a user based setup. We'll make it clear that this is their set up and not ours, both from the indemnity angle that @cjllanwarne was concerned about but more importantly because it helps to demonstrate the vibrant community which is building around Cromwell
- I'll add a second section describing Firecloud's security model
- That doc will be linked from the README
- We'll set up a blog post on the main site describing security/auth options in Cromwell and directly referencing this doc. Readers/users will be encouraged to ask questions, provide alternate suggestions, etc. The security doc (for lack of a better word atm) will be more of a living doc

@delocalizer ... I don't want to make extra work for you here. If you wanted to update this PR to reflect the first and third bullet points great, otherwise I can pick up this PR and do first and third while i'm doing the second. If you're going with the former hold off until I confirm the name of the file :)

geoffjentry:
Discussed name of the file with @katevoss and we're going to go with `SecurityRecommendations.md`

geoffjentry:
This is great, thanks @delocalizer !

The downside of you doing those followups is that now I don't have the open PR staring me in the face reminding me I have to do something ;) I'll try to put in some content re Firecloud in the near future.

PR #1660, opened 2016-11-09
Pin centaur to a specific version 0.21 hotfix.

kshakir:


.travis.yml
```diff
@@ -6,10 +6,13 @@ scala:
 jdk:
   - oraclejdk8
 env:
-  # Setting this variable twice will cause the 'script' section to run twice with the respective env var invoked
-  - BUILD_TYPE=sbt
-  - BUILD_TYPE=centaurJes
-  - BUILD_TYPE=centaurLocal
+  global:
+    - CENTAUR_BRANCH=79dc5f9f0f6ca6aedb5b13358d47b0c9c4156d51
```
mcovarr: Isn't this is a commit and not a branch?

mcovarr:
:+1:

.travis.yml
kshakir: It is a commit, but the variable is named after the the `TRAVIS_BRANCH` tag name used for the cromwell commit, that used to be plumbed through as `BRANCH` in the various centaur scripts.

In the future, one could create named branches (like `develop`), and then pin to the branch names instead of the current hashes.

ruchim: Would this mean that we would create tagged branches in Centaur as a part of the overall Cromwell release?

ruchim:
👍

.travis.yml
kshakir: Yes-ish. To be explicit-- one can use a hash, a named-branch, or a tagged commit.

In general across our repos, currently our git tags point to the first hash that released a set of changes. There as fixes are added, the tag is not updated to the new hashes. Meanwhile, the branch name "xyz_hotfix" points to the latest and greatest hash with bug fixes for that release xyz.

geoffjentry: Would it be more or less confusing to call it `CENTAUR_TREEISH` as that's really what it is?

PR #1668, opened 2016-11-15
Publish hotfix branches to docker.

kshakir:


src/bin/travis/afterSuccess.sh
```diff
@@ -0,0 +1,24 @@
...
+echo "BUILD_TYPE='$BUILD_TYPE'"
+echo "TRAVIS_BRANCH='$TRAVIS_BRANCH'"
+echo "TRAVIS_PULL_REQUEST='$TRAVIS_PULL_REQUEST'"
+
+if [ "$BUILD_TYPE" == "sbt" ] && [ "$TRAVIS_PULL_REQUEST" == "false" ]; then
+
+    if [ "$TRAVIS_BRANCH" == "develop" ]; then
+        sbt 'set test in Test := {}' publish
+
+    elif [[ "$TRAVIS_BRANCH" =~ ^[0-9\.]+_hotfix$ ]]; then
```
kshakir: Only docker pushes hotfix branches. Do we have a tagging scheme for pre-release branches?

src/bin/travis/afterSuccess.sh
```diff
@@ -0,0 +1,24 @@
...
+        docker login -u="$DOCKER_USERNAME" -p="$DOCKER_PASSWORD"
```
kshakir: Still awaiting these credentials to be added [to Travis](https://travis-ci.org/broadinstitute/cromwell/settings).

src/bin/travis/afterSuccess.sh
```diff
@@ -0,0 +1,24 @@
...
+        sbt 'set test in Test := {}' -Dproject.isSnapshot=false dockerBuildAndPush
+
+    elif [[ "$TRAVIS_BRANCH" == "ks_docker_push" ]]; then
```
kshakir: DEBUG code to remove.

.travis.yml
```diff
@@ -11,13 +11,13 @@ env:
   matrix:
     # Setting this variable twice will cause the 'script' section to run twice with the respective env var invoked
     - BUILD_TYPE=sbt
-    - BUILD_TYPE=checkPublish
-    - BUILD_TYPE=centaurJes
-    - BUILD_TYPE=centaurLocal
+    #- BUILD_TYPE=checkPublish
+    #- BUILD_TYPE=centaurJes
+    #- BUILD_TYPE=centaurLocal
```
kshakir: DEBUG code to remove, makes the builds faster.

src/bin/travis/afterSuccess.sh
geoffjentry: Tagging @hjfbynara

src/bin/travis/afterSuccess.sh
geoffjentry: I think I don't understand the question. If I do I'd answer with the question "do we want to?"

src/bin/travis/afterSuccess.sh
kshakir: Creds worked great for the test!

kshakir:
Once this review is complete, I'll back port the patch to the actual hotfix branches `0.21_hotfix` and `0.22_hotfix`.

ruchim:
@hjfbynara  would you mind being one of the two reviewers for this PR?

src/bin/travis/afterSuccess.sh
kshakir: Discussed offline. Will only be pushing hotfix branches to docker, for now. Thus to kick off this process, the current operating procedure should be to create a new hotfix git branch whenever we release.

hjfbynara:
Just wanted to confirm - what will the resulting tags look like?  

See that one will be tagged cromwellVersion.  And then there is another tagged with "version.value".  My scala is not all that great but will version.value result in cromwellVersion-7digitGITHASH ?

which is what this code does:

val basePrefix = baseVersion.map(_ + "-").getOrElse("")
def commitVersion: Option[String] = headCommit map (sha => basePrefix + sha.take(7))

geoffjentry:
👍 provided that @hjfbynara is ok w/ the tag structure

kshakir:
They'll look like the two tags currently in https://hub.docker.com/r/broadinstitute/cromwell/tags/ are:

- `23-0c25192`
- `23`

Note to future viewers: I'll be deleting the git hashed tag, but 23 should re-appear when 23 is actually released.

hjfbynara:
👍

ruchim:
👍 [proxy thumb for Henry]

PR #1685, opened 2016-11-17
Better error reporting if an EJEA crashes unexpectedly

cjllanwarne:
Should be impossible for the EJEA to exit without the workflow noticing it now

### DIFF
--- a/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
@@ -2,7 +2,6 @@ package cromwell.engine.workflow.lifecycle.execution
 
 import java.time.OffsetDateTime
 
-import akka.actor.SupervisorStrategy.{Escalate, Stop}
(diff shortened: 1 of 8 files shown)


engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -623,6 +631,7 @@ final case class WorkflowExecutionActor(workflowId: WorkflowId,
               self, jobKey, data, factory, initializationData.get(backendName), restarting, serviceRegistryActor,
               jobStoreActor, callCacheReadActor, jobTokenDispenserActor, backendSingleton, backendName, workflowDescriptor.callCachingMode)
             val ejeaRef = context.actorOf(ejeaProps, ejeaName)
+            context watch ejeaRef
             pushNewJobMetadata(jobKey, backendName)
```
Horneth: why not add the `ejeaRef` to the data here instead of when receiving the `JobStarting` message ?

Horneth:
👍

mcovarr:
:+1:

PR #1692, opened 2016-11-18
added support for a FileRoller logback configuration

hjfbynara:
I labeled "do not merge" since I would like discussion first and there are one or two items that I need to figure out a good way to test.  Once merged and in image one would update JAVA_OPTS similar to below to enable and modify the defaults.

JAVA_OPTS="-DLOG_MODE=FILEROLLER -DFILEROLLER_NAMEPATTERN=%d{yyyyMMddHHmm} -DFILEROLLER_NAME=crom -DFILEROLLER_DIR=/local/cromwell  -DFILEROLLER_MAXHISTORY=3"

To try it out prior a full build I did the following.

1. Copy the logback.xml from this branch into some directory.  make a logs subdirectory
2. from that directory run the following

docker run -it --rm -p 8000:8000  -v ${PWD}:/working -v ${PWD}/logs:/local/cromwell -e JAVA_OPTS="-Dlogback.configurationFile=/working/logback.xml -DLOG_MODE=FILEROLLER -DFILEROLLER_NAMEPATTERN=%d{yyyyMMddHHmm} -DFILEROLLER_NAME=crom -DFILEROLLER_DIR=/local/cromwell -DFILEROLLER_MAXHISTORY=3"  broadinstitute/cromwell:0.22-881e7b7 server

after running for several minutes logs directory should end up looking like:

201611182054-crom
201611182056-crom
crom

NOTE: FileRoller rotation seems to only happen when logs are generated.  So IF your cromwell is not generating any log messages the logfile won't be rotated.  So in my example above the logfile "crom" may have messages from several minutes ago (in my case).  But when the next message comes in - crom will be rotated prior to the new log message being written.  I think the rotated logfile will get the timestamp based on the date of the last log message in that file.  Which is why in the list above there is a gap.


hjfbynara:
I found two logback.xml files in cromwell.   They were the exact same, but my PR only updates the one under engine.  Is because I did not update the other one as well - the reason that travis is failing with the following error?

[error] 1 error was encountered during merge
…java.lang.RuntimeException: deduplicate: different file contents found in the following:
logback.xml
logback.xml

cjllanwarne:
You'll need to create an equivalent (maybe cherry-picked) change under develop or this'll be lost for 23.

This PR looks fine to me. I don't think it affects me and presumably makes your life easier so 👍 :)

cjllanwarne:
Actually, micro concern: does our workflow-logs copier need updating to reflect more than 1 Cromwell log file? (I don't know who might know the answer to this)

Horneth:
Workflows create their own logback appender programmatically so they shouldn't be affected by this ? @kshakir should know I think.

kshakir:
ToL: In a future life, I would like to see if we can fix the issue requiring duplicate logback.xml files. For now, LGTM 👍

geoffjentry:
@hjfbynara What's the status of this from your perspective? You still have the do not merge label on it, is that still the case?

PR #1693, opened 2016-11-18
allow specification of compute service account in JES closes #1656

### LINKED ISSUE #1656: Configurable value for service account Google Pipelines API uses for GCE nodes
Initiated by conversation on http://gatkforums.broadinstitute.org/wdl/discussion/8546/using-different-service-account-on-compute-instance#latest 

Currently, we are supplying 'default' to the RunPipelineArgs service_account email.  This means the Pipelines API node will spin up using the default compute service account for the project.  However (a) that account can be removed and (b) users may want to use a different service account.

This should be configurable at both the cromwell server level (ie applies to all workflows) as well as by the workflow options (so it can vary by workflow).

This new capability should also be documented in the README (or whever @katevoss says!)

kcibul:


supportedBackends/jes/src/test/scala/cromwell/backend/impl/jes/JesAttributesSpec.scala
```diff
@@ -60,17 +69,18 @@ class JesAttributesSpec extends FlatSpec with Matchers {
     errorsList should contain("no protocol: myEndpoint")
   }
 
-  val configString =
-    """
+  def configString(preemptible:String = "", genomics:String = "") =
```
cjllanwarne: Add some spaces between `:` and `String`

cjllanwarne: Add some spaces between `:` and `String`

cjllanwarne:
LGTM. Some sort of comment in the docs about when/why I'd want to use this might be nice. My thinking is, if I just want to "try it out as simple as possible" I probably don't want to be scared off by all these options?

Anyway, that's just a ToL so 👍

mcovarr:
:+1:

PR #1696, opened 2016-11-22
Resolve downstream dependencies 0.21

ruchim:


engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -168,8 +168,12 @@ object WorkflowExecutionActor {
         case _ => Nil
       }
 
+      def isDone(e: JobKey): Boolean = executionStore.store exists {
+        case (k, s) => k.scope == e.scope && k.index == e.index && s == ExecutionStatus.Done
+      }
+
       val dependencies = upstream.flatten ++ downstream
-      val dependenciesResolved = dependencies forall { case (_, s) => s == ExecutionStatus.Done }
+      val dependenciesResolved = dependencies forall { case (k, s) => isDone(k) }
```
mcovarr: underscore the `s`

Horneth: total nit pick: this could be `dependencies.keys forall isDone`

Horneth: As long as @mcovarr says it's fine wrt hashing

Horneth:
👍

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
mcovarr: That would be ok wrt hashing, but the data structure is a List of tuples rather than a Map so it's probably not ok wrt scalac

mcovarr:
:+1: with real world confirmation that the preemption problem is actually fixed and the "takes forever to finish" issue hasn't regressed

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
ruchim: Scala is being fussy about the tuple, wouldn't let me grab the keys from the tuple.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -168,8 +168,12 @@ object WorkflowExecutionActor {
         case _ => Nil
       }
 
+      def isDone(e: JobKey): Boolean = executionStore.store exists {
+        case (k, s) => k.scope.fullyQualifiedName == e.scope.fullyQualifiedName && k.index == e.index && s == ExecutionStatus.Done
+      }
+
       val dependencies = upstream.flatten ++ downstream
-      val dependenciesResolved = dependencies forall { case (_, s) => s == ExecutionStatus.Done }
+      val dependenciesResolved = dependencies forall { case (k,_) => isDone(k) }
```
geoffjentry: `k, _`

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -168,8 +168,12 @@ object WorkflowExecutionActor {
         case _ => Nil
       }
 
+      def isDone(e: JobKey): Boolean = executionStore.store exists {
+        case (k, s) => k.scope == e.scope && k.index == e.index && s == ExecutionStatus.Done
```
ruchim: To be clear, it's been changed to `case (k, s) => k.scope.fullyQualifiedName == e.scope.fullyQualifiedName && k.index == e.index && s == ExecutionStatus.Done`

PR #1697, opened 2016-11-22
Resolve downstream Dependencies 0.22 closes #1690

### LINKED ISSUE #1690: JES Preemption during scatter operation leads to stall after scatter block
When running some of our pipelines with the latest release version (0.22), I am not able to get the pipelines to actually complete if they run into a preemption during their scatter operation. The scatter block will successfully complete for all cases, however the execution stalls and does not progress any further.

To make sure this had nothing to do with a slow process I let the job sit overnight, and still it did not complete, or even move past the "Success" of the final scatter operation to complete. I used the wdl code below to replicate the issue several times locally, without having to run our entire workflow

```
task c {
   Int num
   command {
	echo ${num}
        #Wait 5 minutes, and hope for a preemption. make longer to force preemption
	sleep 300
   }
   runtime {
     docker: "ubuntu"
     preemptible: 4
   }
   output {
     Int out = read_int(stdout())
   }
}
task d {
   Array[Int] num
   command {
	echo ${sep=" " num}
        sleep 300
   }
   runtime {
     docker: "ubuntu"
     preemptible: 4
   }
   output {
     String out=stdout()
   }
}
workflow wf {
 Array[Int] first = [1,2,3,4,5]

 scatter (x in first){
    	call c {input: num=x}
	call c as a {input: num=c.out}
 }

 call d {input: num=a.out}
 output {
    d.*
 }
}
```

ruchim:


engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -170,8 +170,12 @@ object WorkflowExecutionActor {
         case _ => Nil
       }
 
+      def isDone(e: JobKey): Boolean = executionStore.store exists {
+        case (k, s) => k.scope.fullyQualifiedName == e.scope.fullyQualifiedName && k.index == e.index && s == ExecutionStatus.Done
```
mcovarr: ToL this might benefit from a comment as to why the `exists` is required so I don't go in here and break it again in a couple of months.  😄

mcovarr:
:+1:

Horneth:
👍

PR #1707, opened 2016-11-28
Re-enable flag for workflow-restart Closes #1706

### LINKED ISSUE #1706: Config parameter for 'system.workflow-restart' doesn't work
It's in the docs... but not the code!

kcibul:


mcovarr:
need to fix the unused import to unbreak the build 😄

engine/src/main/scala/cromwell/engine/workflow/workflowstore/SqlWorkflowStore.scala
```diff
@@ -13,9 +15,13 @@ import scala.concurrent.{ExecutionContext, Future}
 
 case class SqlWorkflowStore(sqlDatabase: WorkflowStoreSqlDatabase) extends WorkflowStore {
   override def initialize(implicit ec: ExecutionContext): Future[Unit] = {
-    sqlDatabase.updateWorkflowState(
-      WorkflowStoreState.Running.toString,
-      WorkflowStoreState.Restartable.toString)
+          if ( ConfigFactory.load().as[Option[Boolean]]("system.workflow-restart").getOrElse(true)) {
```
mcovarr: extra space after the open paren

mcovarr:
:+1:

engine/src/main/scala/cromwell/engine/workflow/workflowstore/SqlWorkflowStore.scala
```diff
@@ -13,9 +15,13 @@ import scala.concurrent.{ExecutionContext, Future}
...
+            sqlDatabase.updateWorkflowState(
+                WorkflowStoreState.Running.toString,
+                WorkflowStoreState.Restartable.toString)
+          } else {
+            Future.successful(())
```
Horneth: Will they stay forever in the WorkflowStore with a "Running" state in this case ?

geoffjentry: I remember having the discussion here of what *should* happen in a case like this but don't recall the result except to say that it wasn't unanimous :) 

I'd lean towards leaving them there although I probably said the opposite before.

mcovarr: Did we not document what Cromwell does in this case?  If we didn't, we should.

geoffjentry: On that note I don't see this option documented at all

kcibul: it's in the migration.md, as well as in the comments of the conf.reference.  At this point, I think it's a small enough option to leave it there.  However, if/when this becomes a key parameter in standing up several cromwells (e.g. 'reader cromwells' then we should document the suite of parameters required to do that)

Horneth:
👍

engine/src/main/scala/cromwell/engine/workflow/workflowstore/SqlWorkflowStore.scala
cjllanwarne: If you switch from `don't restart` to `do restart`, do we expect all of our old unfinished workflows from months ago to spring back into life again?

kcibul: It's a fair question.  Yes -- I think that's what should happen.  Predominantly because if you've been running one way for a long time and then fiddle some "advanced" setting (ie only documented in the reference.conf), you should know what you're doing ;)  We do have a need to disable this for the use case I mentioned above, so I want to be able to support that, which is impossible now since it's hard coded, but not solve for use cases that we don't have right now because we're unlikely to get it right in absence of a such a use case.  phew!

PR #1710, opened 2016-11-29
added support for a FileRoller logback configuration (#1692)

kcibul:
* added support for a FileRoller logback configuration
* made both logback.xml files in repo the same
* rearrange to single logback.xml


geoffjentry:
LGTM but I don't know that I'd notice if it wasn't right. 👍

kshakir:
That's the right location. 👍

geoffjentry:
👍 👍 👍 👍 👍 👍 👍 

Hopefully one of these wokrs

PR #1714, opened 2016-11-29
Workflow name expanded to 500 chars as consistent with 0.19

mcovarr:


Horneth:
👍

Horneth:
Is that branch name > 100 char 😄

ruchim:
👍

kcibul:
two thumbs, green tests, and it's holding up workbench... so merging!

PR #1718, opened 2016-11-30
Allow for JES QPS to be tunable

geoffjentry:


supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesPollingActor.scala
```diff
@@ -109,13 +108,31 @@ class JesPollingActor(pollingManager: ActorRef) extends Actor with ActorLogging
...
+    * Given the Genomics API queries per 100 seconds and given MaxBatchSize will determine a batch interval which
+    * is at 90% of the quota. The (still crude) delta is to provide some room at the edges for things like new
+    * calls, etc.
+    *
+    * Forcing the minimum value to be 1 second, for now it seems unlikely to matter and it makes testing a bit
+    * easier
+    */
+  def determineBatchInterval(qps: Int): FiniteDuration = {
+    val batchesPerSecond = qps / MaxBatchSize.toDouble // Force this to be floating point in case the value is < 1
+    val maxInterval = 1 / batchesPerSecond
```
Horneth: Should we check somewhere that qps is not 0 ?
Of course it should never be 0 but if for some reason it happens it would be nice not to have a giant crash with `java.lang.ArithmeticException` :p

geoffjentry: You raise a good point

geoffjentry: I'm thinking that 0 would map to the default value, what do you think about that?

mcovarr: I would think nonpositive values would map to bottom

geoffjentry: @mcovarr It seems unlikely that someone would have requested their QPS to be lower than what comes for free (10)

mcovarr: not on purpose

geoffjentry: So then what is bottom QPS? Technically their quota could be 1 in which case QPS is 1/100. That'd be a poll every 10000 seconds. Is that really what you want?

Horneth: Would there be any use case where setting it to 0 would actually be intentional if you want to prevent cromwell from submitting anything to JES ? I can't think of one but who knows.
I think at least a value <= 0 should be logged as a warning, regardless of what is actually used afterwards.

mcovarr: @geoffjentry oh I meant bottom as in ⊥, i.e. throw.  Or at least get really mad and log an error.

geoffjentry: I doubt anyone would want to do that but you could convince me that it'd be feature &not bug. If we go that route I'd be fine w/ zeroing it out. Otherwise my take is that hitting teh QPS limit is more about wasted work instead of things going haywire and that seems superior to making it so that someone's jobs will take a week to register as complete because they did something stupid

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesBackendSingletonActor.scala
```diff
@@ -1,12 +1,12 @@
 package cromwell.backend.impl.jes
 
 import akka.actor.{Actor, ActorLogging, Props}
-import cromwell.backend.impl.jes.statuspolling.{JesApiQueryManager}
+import cromwell.backend.impl.jes.statuspolling.JesApiQueryManager
 import cromwell.backend.impl.jes.statuspolling.JesApiQueryManager.DoPoll
 
-class JesBackendSingletonActor extends Actor with ActorLogging {
+final case class JesBackendSingletonActor(jesConfiguration: JesConfiguration) extends Actor with ActorLogging {
```
cjllanwarne: ToL I've recently gone off the idea of using `case class` for actors. They aren't being used anything like ADTs and this limits their future extensibility.

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesPollingActor.scala
cjllanwarne: Uh, but why not just `val maxInterval = MaxBatchSize.toDouble / qps`?

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesPollingActor.scala
```diff
@@ -109,13 +108,31 @@ class JesPollingActor(pollingManager: ActorRef) extends Actor with ActorLogging
...
   private def scheduleCheckForWork(): Unit = {
-    context.system.scheduler.scheduleOnce(BatchInterval) { pollingManager ! JesApiQueryManager.RequestJesPollingWork(MaxBatchSize) }
+    context.system.scheduler.scheduleOnce(batchInterval) { pollingManager ! JesApiQueryManager.RequestJesPollingWork(MaxBatchSize) }
     ()
   }
 }
 
 object JesPollingActor {
-  def props(pollingManager: ActorRef) = Props(new JesPollingActor(pollingManager))
+  def props(pollingManager: ActorRef, jesConfiguration: JesConfiguration) = Props(new JesPollingActor(pollingManager, jesConfiguration))
```
cjllanwarne: Requiring the entire `JesConfiguration` just so that you can extract the `qps` seems a little OTT

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesBackendSingletonActor.scala
cjllanwarne: A concrete example from further down, you could no longer subtype the actor in tests, if you so wished (like we have for the other actors in this PR)

geoffjentry: My stance is things should be final case class by default unless there's a reason To not do it (as in current not theiretical). To your second point I originally converted everything but reverted when I went to run tests.

IMO extension should be thought through carefully and not done willy nilly so providing a hurdle is a good thing.

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesPollingActor.scala
geoffjentry: The alternative is to grab the Config out of the blue as opposed to the centralized scheme

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesPollingActor.scala
geoffjentry: Probably because this changed several times before it got here and I wasn't paying attention :)

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesBackendSingletonActor.scala
cjllanwarne: Is there a solid reason for that stance though?

I'm happy to be corrected but IMO ADTs should be considered a *data type*, for immutable data storage. This is a mutating, active object. We're instrumenting this actor with a bunch of helper methods for apply, unapply, variable access that we actively don't want and actually go against the grain of how this actor object should be used.

core/src/main/resources/reference.conf
```diff
@@ -319,6 +319,15 @@ backend {
     #    # Base bucket for workflow executions
     #    root = "gs://my-cromwell-workflows-bucket"
     #
+    #    # Set this to the lower of the two values "Queries per 100 seconds" and "Queries per 100 seconds per user" for
+    #    # your project.
+    #    #
+    #    # Used to help determine maximum throughput to the google genomics API. Setting this value too low will
```
mcovarr: capitalize Google Genomics

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesPollingActor.scala
```diff
@@ -19,12 +19,11 @@ import scala.concurrent.duration._
 /**
   * Polls JES for status. Pipes the results back (so expect either a RunStatus or a akka.actor.Status.Failure).
   */
-class JesPollingActor(pollingManager: ActorRef) extends Actor with ActorLogging {
+class JesPollingActor(val pollingManager: ActorRef, val jesConfiguration: JesConfiguration) extends Actor with ActorLogging {
+  // The interval to delay between submitting each batch
+  lazy val batchInterval = determineBatchInterval(jesConfiguration.qps)
+  log.debug("JES batch polling interval is %", batchInterval)
```
mcovarr: why `%` rather than `{}`?

mcovarr:
:+1:

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesPollingActor.scala
cjllanwarne: Because it's 10{} better!

geoffjentry: Because I couldn't remember what the symbol was and this seemed to work ;)

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesPollingActor.scala
geoffjentry: @mcovarr you and your haskell

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesBackendSingletonActor.scala
geoffjentry: But case classes & adts aren't isomorphic. The idiomatic way to implement an adt in scala is via case classes but one can use case classes for other things.

I've long felt that case classes should have been the default class, and class the opt-in. The reason being that they provide a set of reasonable default behavior that one should generally want. public vals for the items in the constructor, a default apply(), inability to extend, a usually not completely stupid equals/toString, etc.

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesPollingActor.scala
geoffjentry: Oh nm, I see what you mean

geoffjentry:
Updated for code review

PR #1721, opened 2016-11-30
Revert workflow name embiggening, 0.22 style.

mcovarr:
Apparently there weren't any "real" workflow names > 100 chars, only one artificially long one created for test purposes, and Cloud SQL did not take kindly to the attempt to have a 500-char field as part of an index.  The test data was downsized and the embiggening commit reverted and the migration ran successfully.


ruchim:
👍 👍 👍

Horneth:
👍

Horneth:
Should we create a ticket to gracefully handle workflow name > 100 char if that ever happens ?

geoffjentry:
@Horneth #1725

PR #1739, opened 2016-12-05
Fix execution events not propagated from the backend

Horneth:


mcovarr:
:+1:

ruchim:
👍

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -407,7 +408,9 @@ class EngineJobExecutionActor(replyTo: ActorRef,
 
   private def saveJobCompletionToJobStore(updatedData: ResponseData) = {
     updatedData.response match {
-      case JobSucceededResponse(jobKey: BackendJobDescriptorKey, returnCode: Option[Int], jobOutputs: CallOutputs, _, _) => saveSuccessfulJobResults(jobKey, returnCode, jobOutputs)
+      case JobSucceededResponse(jobKey: BackendJobDescriptorKey, returnCode: Option[Int], jobOutputs: CallOutputs, _, executionEvents) =>
+        eventList ++= executionEvents
```
cjllanwarne: Is this one needed?

cjllanwarne: Seems like it should have already happened by the time we're writing to the JobStore

cjllanwarne: E.g. if this is called from line 227, won't we get duplicate entries?

Horneth: do you mean `saveJobCompletionToJobStore` can be called several times ? I see it's called in different places but I assumed they were all exclusive. I put it there because that's what the nominal "BackendResponse" calls line 221

Horneth: FWIW I think it would be nice to have the execution events be a little be more stream-y and be updated as we get them (with every poll). Also failed jobs for some reason don't send execution events back to the engine which they could.

Horneth: I double checked, `saveJobCompletionToJobStore` puts the FSM into `UpdatingJobStore` state from which it can only `forwardAndStop` or `forwardAndStop` so it can't be called several times.

PR #1741, opened 2016-12-05
Add dispatchers to actors which were missing them

geoffjentry:


kshakir:
ToL: For our pluggable backends, I don't like that the backend implementor needs to know to specify these dispatchers. I've thought that perhaps the engine should retrieve the various Props from the factory, and then the engine invokes ".withDispatcher()" on the returned value.

That's a bigger change. For now. 👍

geoffjentry:
@kshakir Yeah, when we put these in I remember it came up that a malicious and/or uninformed implementer could really gum up the works the way things are now although for now I suspect the solution is way more work than the size of the actual problem

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/FetchCachedResultsActor.scala
```diff
@@ -10,7 +11,7 @@ import scala.util.{Failure, Success}
 
 object FetchCachedResultsActor {
   def props(callCachingEntryId: CallCachingEntryId, replyTo: ActorRef, callCache: CallCache): Props =
-    Props(new FetchCachedResultsActor(callCachingEntryId, replyTo, callCache))
+    Props(new FetchCachedResultsActor(callCachingEntryId, replyTo, callCache)).withDispatcher(EngineDispatcher)
```
cjllanwarne: I thought we had a special dispatcher for DB operations?

Looking at the list... was it the `IODispatcher`?

Also relevant for a lot of the other DB accessor actors in the PR

services/src/main/scala/cromwell/services/metadata/impl/MetadataSummaryRefreshActor.scala
```diff
@@ -21,7 +22,7 @@ object MetadataSummaryRefreshActor {
   case object MetadataSummarySuccess extends MetadataSummaryActorMessage
   final case class MetadataSummaryFailure(t: Throwable) extends MetadataSummaryActorMessage
 
-  def props() = Props(new MetadataSummaryRefreshActor())
+  def props() = Props(new MetadataSummaryRefreshActor()).withDispatcher(ServiceDispatcher)
```
cjllanwarne: I think this should be something else. We don't want the summarizer gumming up the rapid API responses

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/FetchCachedResultsActor.scala
geoffjentry: Discussed in person

services/src/main/scala/cromwell/services/metadata/impl/MetadataSummaryRefreshActor.scala
geoffjentry: Discussed in person

PR #1742, opened 2016-12-05
Enable centaur JES full parallelization

Horneth:


src/bin/travis/testCentaurLocal.sh
```diff
@@ -36,4 +36,8 @@ git clone https://github.com/broadinstitute/centaur.git
 cd centaur
 git checkout ${CENTAUR_BRANCH}
 cd ..
+# All tests use ubuntu:latest - make sure it's there before starting the tests 
+# because pulling the image during some of the tests would cause them to fail
```
ruchim: Do you think we could build a retry for docker failures in Centaur itself?  Or better to have something like that in Cromwell?

Horneth: The reason this is needed is because when docker pulls the image (because it couldn't find it locally), it writes the log in stderr. It's not an actual failure though. It looks like 
```
Unable to find image 'ubuntu:latest' locally
latest: Pulling from library/ubuntu
af49a5ceb2a5: Pulling fs layer
8f9757b472e7: Pulling fs layer
e931b117db38: Pulling fs layer
...
```

However one of the tests writes something to stderr and centaur expects this something to be the value, but it finds all this docker logs on top of it.

geoffjentry:
👍

src/bin/travis/testCentaurLocal.sh
ruchim: ahhhh I see, makes sense

ruchim:
👍

PR #1749, opened 2016-12-06
add the QPS to the jes config whitelist

geoffjentry:


kshakir:
No idea where this is used, but speeling checks out. 👍

geoffjentry:
@kshakir That list is used to provide a warning in the log if you specify a key that isn't used. I'd prefer to either do away w/ the warning or find some way to force the config keys we use and this list to be the same but I figured my fellow cromwellians would balk at the former and I couldn't think of a non-annoying way to do the latter.

mcovarr:
:+1:

PR #1752, opened 2016-12-06
Create cromwell-version.conf

ruchim:
Happy to remove Wdl4s versioning if that's not anything us or a user would find helpful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ A [Workflow Management System](https://en.wikipedia.org/wiki/Workflow_management
 * [Command Line Usage](#command-line-usage)
   * [run](#run)
   * [server](#server)
+  * [version](#version)
 * [Getting Started with WDL](#getting-started-with-wdl)
 * [Configuring Cromwell](#configuring-cromwell)
   * [Workflow Submission](#workflow-submission)
@@ -159,6 +160,12 @@ run <WDL file> [<JSON inputs file>] [<JSON workflow options>]
 
   Starts a web server on port 8000.  See the web server
   documentation for more details about the API endpoints.
+
+  -version
+
+  Returns the version of Cromwell engine being used and the
+  version of Wdl4S being utilized.
+
 ```
 
 ## run
@@ -283,6 +290,10 @@ $ java -jar cromwell.jar run threestep.wdl - - - /path/to/my_WDLs.zip
 ## server
 
 Start a server on port 8000, the API for the server is described in the [REST API](#rest-api) section.
+
+## version
+
+Returns the version of Cromwell and the version of Wdl4S being consumed.
 
 # Getting Started with WDL
 
@@ -3138,7 +3149,7 @@ Response:
 "content-type": "application/json; charset=UTF-8"
 
 {
-  "cromwell": 23,
+  "cromwell": 23-8be799a-SNAP,
   "wdl4s": 0.7
 }
 ```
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -3115,6 +3116,33 @@ Response:
...
+
+Response:
+```
+"date": "Sun, 18 Sep 2016 14:38:11 GMT",
+"server": "spray-can/1.3.3",
+"content-length": "33",
+"content-type": "application/json; charset=UTF-8"
+
+{
+  "cromwell": 23,
```
ruchim: Need to fix this, it will actually be 23-xxxxxx-SNAP

engine/src/main/resources/swagger/cromwell.yaml
```diff
@@ -822,3 +842,12 @@ definitions:
...
         description: The number of currently running jobs
+  VersionResponse:
+    description: 'Provides the version of the engine running inside the system'
+    properties:
+      cromwell:
+        type: string
+        description: The version of the Cromwell Engine
+      wdl4s:
+        type: string
+        description: The version of the Wdl4S beign used by Cromwell
```
Horneth: Do we really want the wdl4s version ? wdl4s is just a dependency of cromwell as many others IMO.
The wdl version (when well defined) would be better.

Horneth: Also beign -> being

engine/src/main/resources/swagger/cromwell.yaml
```diff
@@ -822,3 +842,12 @@ definitions:
       jobs:
         type: integer
         description: The number of currently running jobs
+  VersionResponse:
+    description: 'Provides the version of the engine running inside the system'
```
mcovarr: I think the outer quotes are superfluous?  Also maybe "Returns the version of Cromwell"

project/Version.scala
```diff
@@ -19,6 +19,20 @@ object Version {
...
+      s"""|version {
+          |  cromwell: "${version.value}"
+          |  wdl4s: "${Dependencies.wdl4sV}"
+          |}
+          |""".stripMargin
+    IO.write(file, contents)
+    Seq(file)
+  }
+
+  val versionConfCompileSettings = List(resourceGenerators in Compile <+= writeVersionConf)
```
mcovarr: careful, you could put an eye out with that operator

engine/src/main/resources/swagger/cromwell.yaml
mcovarr: lowercase `wdl4s`

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
```diff
@@ -271,6 +271,21 @@ trait CromwellApiService extends HttpService with PerRequestCreator {
       }
     }
 
+  def versionRoute =
+    path("engine" / Segment / "version") { version =>
+      get {
+        complete {
+          val versionConf = ConfigFactory.load("cromwell-version.conf").getConfig("version")
```
mcovarr: maybe `lazy val` this as a class member so it doesn't have to be loaded every time somebody hits the endpoint

mcovarr:
:+1:

project/Version.scala
ruchim: 🙈

Horneth:
👍

PR #1758, opened 2016-12-08
Publish develop to docker hub.

kshakir:


src/bin/travis/afterSuccess.sh
```diff
@@ -9,7 +9,12 @@ echo "TRAVIS_PULL_REQUEST='$TRAVIS_PULL_REQUEST'"
 if [ "$BUILD_TYPE" == "sbt" ] && [ "$TRAVIS_PULL_REQUEST" == "false" ]; then
 
     if [ "$TRAVIS_BRANCH" == "develop" ]; then
-        sbt 'set test in Test := {}' publish
+        docker login -u="$DOCKER_USERNAME" -p="$DOCKER_PASSWORD"
+        sbt \
+          'set test in Test := {}' \
```
ruchim: Why does Test need to be explicitly stated as an empty set? Ive seen it in multiple places now but I'm unclear of the why

ruchim:
👍

src/bin/travis/afterSuccess.sh
kshakir: This is a way of stopping all tests from running, even the defaults.

Sometimes it's not needed and may be extra copypasta. But I believe `sbt publish` will try to re-run the test suites.

Since Travis has made it successfully to our `afterSuccess.sh` we know that `sbt test` has just run, and they succeeded.

Horneth:
👍

PR #1760, opened 2016-12-08
With thanks to @delocalizer

cjllanwarne:


ruchim:
Great catch @delocalizer  👍

kshakir:
👍

PR #1772, opened 2016-12-09
Escalate service instantiation failures Closes #1266

### LINKED ISSUE #1266: Startup should fail if services are invalid
To reproduce: Add an invalid actor for KeyValueService in configuration.

Startup should fail, but instead the service continues running

Horneth:


kshakir:
`ServiceRegistryActorSpec` needs to be patched also, but otherwise LGTM 👍

geoffjentry:
👍

PR #1778, opened 2016-12-12
Rework stats actor Closes #1593

### LINKED ISSUE #1593: Some logging messages about completed jobs and (if possible) remaining jobs would be useful
When looking at the output of cromwell (0.22) single workflow mode (local backend), it would be useful to have output that listed the number of completed jobs and remaining jobs.

Currently, it is not possible to determine whether the jobs are finishing, hanging, etc in single workflow mode.

The behavior in GATK 3.x Queue can serve as a model.

Horneth:
Reworks the stats actor to passively receive information about workflows / jobs instead of asking for them.
Job status changes are published every time the execution store is modified.
Workflow status changes are published every time the WA changes state.

When a workflow execution completes (ie the WEA sends back a terminal response), jobs run by this workflow are removed from the count.
When a workflow completes (ie the WA dies) it's removed from the count.

Also logs periodically (in single workflow mode only) the number of jobs in each status.


engine/src/main/scala/cromwell/engine/EngineStatsActor.scala
```diff
@@ -0,0 +1,85 @@
...
+    Props(new EngineStatsActor(statusReport)).withDispatcher(Dispatcher.ApiDispatcher)
+  }
+
+  sealed abstract class EngineStatsActorMessage
+  case object StatsQuery extends EngineStatsActorMessage
+  case object LogStats extends EngineStatsActorMessage
+
+  final case class EngineStats(workflows: Map[String, Int], jobs: Map[String, Int])
+
+  val MaxTimeToWait = 3 seconds
```
Horneth: this can go

engine/src/main/scala/cromwell/engine/EngineStatsActor.scala
```diff
@@ -0,0 +1,85 @@
...
+
+final class EngineStatsActor(statusReportInterval: FiniteDuration) extends Actor with ActorLogging {
+  implicit val ec = context.dispatcher
+
+  private var workflows: Map[WorkflowId, WorkflowState] = Map.empty
+  private var jobs: Map[JobKey, ExecutionStatus] = Map.empty
+  
+  private lazy val logHeader = List("Status", "Count")
+  
+  if (statusReportInterval != Duration.Zero) context.system.scheduler.scheduleOnce(statusReportInterval, self, LogStats)
```
Horneth: should this be in preStart ?

geoffjentry:
In the back of my mind I have this feeling that I purposefully didn't do it this way but for the life of me can't think of why that'd be or why I'd think it was a good idea. Was hoping writing this comment would help jog the memory but apparently not

geoffjentry:
And my hope that hitting `comment` so everyone could see my feeble attempt at jogging my memory would jog my memory also failed.

geoffjentry:
Following up on my mysterious internal discussion I think the reason was that originally I wanted to directly hit the DB for everything to remove any possible ambiguity* and then that turned out to be unpossible anyways but by then I forgot about this debate

* It turns out the DB can lag reality a bit anyways

README.md
```diff
@@ -3102,7 +3102,7 @@ Server: spray-can/1.3.3
 
 ## GET /api/engine/:version/stats
 
-This endpoint returns some basic statistics on the current state of the engine. At the moment that includes the number of running workflows and the number of active jobs. 
+This endpoint returns some basic statistics on the current state of the engine. At the moment that includes the number of running workflows and the number of jobs belonging to running workflows with their respective status.
```
mcovarr: Should this include the jobs in aborting workflows too?  Either folded into the one histogram or broken out into their own histogram?

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -225,19 +228,14 @@ class WorkflowActor(val workflowId: WorkflowId,
   }
 
   when(ExecutingWorkflowState) {
-    case Event(WorkflowExecutionSucceededResponse(jobKeys, outputs),
+    case Event(response @ WorkflowExecutionSucceededResponse(jobKeys, outputs),
     data @ WorkflowActorData(_, Some(workflowDescriptor), _, _)) =>
+      statsActor ! response
```
mcovarr: ToL The pattern here and elsewhere seems to be to send the stats actor a notification before the WA has really done the processing for the state, in particular before the finalization.  Is there a danger that the stats actor will think the workflow has finished in a happy place but in reality a failed finalization will fail the workflow?

mcovarr:
:+1:

Horneth:
I'm closing this and will open a new PR with a hopefully better way to collect the statuses

PR #1781, opened 2016-12-13
A version of JES using the Standard backend traits.

kshakir:
Moved supervision from JES down into Standard trait.
Fixed some vals in the Standard trait.
Fixed non-Travised integration test `JesAttributesSpec`.


geoffjentry:
Except for the fact that every file ends in a \n :+1:

mcovarr:
:+1:

PR #1784, opened 2016-12-14
don't log yuuuuuuuge output values closes #1770

### LINKED ISSUE #1770: Do not log outputs when they are massive enough to have a gravitational pull
`handleWorkflowSuccess` in `WorkflowExecutionActor` attempts to print out a log message which include the outputs. This doesn't go so well when the outputs are ginormous.

geoffjentry:
It turns out that it is really freaking annoying when you have millions of filepaths in a workflow output, who would have guessed?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Cromwell Change Log
 
+## 24
+
+* When emitting workflow outputs to the Cromwell log only the first 1000 characters per output will be printed
+
 ## 23
 
 * The `meta` and `parameter_meta` blocks are now valid within `workflow` blocks, not just `task`
--- a/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
@@ -22,7 +22,8 @@ import cromwell.webservice.EngineStatsActor
 import lenthall.exception.ThrowableAggregation
 import lenthall.util.TryUtil
 import net.ceedubs.ficus.Ficus._
-import wdl4s.values.{WdlArray, WdlValue}
+import org.apache.commons.lang3.StringUtils
+import wdl4s.values.{WdlArray, WdlString, WdlValue}
 import wdl4s.{Scope, _}
 
 import scala.annotation.tailrec
@@ -254,9 +255,10 @@ case class WorkflowExecutionActor(workflowDescriptor: EngineWorkflowDescriptor,
   }
   
   private def handleWorkflowSuccessful(data: WorkflowExecutionActorData) = {
+    import WorkflowExecutionActor.EnhancedWorkflowOutputs
     import cromwell.util.JsonFormatting.WdlValueJsonFormatter._
     import spray.json._
-    
+
      val (response, finalState) = workflowDescriptor.workflow.evaluateOutputs(
       workflowDescriptor.knownValues,
       data.expressionLanguageFunctions,
@@ -264,7 +266,7 @@ case class WorkflowExecutionActor(workflowDescriptor: EngineWorkflowDescriptor,
     ) map { workflowOutputs =>
        workflowLogger.info(
          s"""Workflow ${workflowDescriptor.workflow.unqualifiedName} complete. Final Outputs:
-             |${workflowOutputs.toJson.prettyPrint}""".stripMargin
+             |${workflowOutputs.stripLarge.toJson.prettyPrint}""".stripMargin
        )
        pushWorkflowOutputMetadata(workflowOutputs)
        (WorkflowExecutionSucceededResponse(data.jobExecutionMap, workflowOutputs mapValues JobOutput.apply), WorkflowExecutionSuccessfulState)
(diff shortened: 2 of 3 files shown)


engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActorData.scala
```diff
@@ -134,19 +133,6 @@ case class WorkflowExecutionActorData(workflowDescriptor: EngineWorkflowDescript
     this.copy(downstreamExecutionMap = downstreamExecutionMap ++ jobExecutionMap)
   }
 
-  def outputsJson(): String = {
```
geoffjentry: this was unused

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -623,4 +624,11 @@ object WorkflowExecutionActor {
     Props(WorkflowExecutionActor(workflowDescriptor, serviceRegistryActor, jobStoreActor, subWorkflowStoreActor,
       callCacheReadActor, jobTokenDispenserActor, backendSingletonCollection, initializationData, restarting)).withDispatcher(EngineDispatcher)
   }
-}
\ No newline at end of file
+
+  implicit class EnhancedWorkflowOutputs(val outputs: Map[LocallyQualifiedName, WdlValue]) extends AnyVal {
+    def stripLarge = outputs map { case (k, v) =>
+      if (v.toWdlString.length > 1024) (k, WdlString("Output too long, elided"))
```
mcovarr: Eliding would be cool, but this is more like censoring as none of the actual output value makes it out

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActorData.scala
mcovarr: Surprising the compiler didn't balk

mcovarr:
:+1:

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
geoffjentry: censorship is uncool!

elided in the context that among the other outputs this one was omitted

geoffjentry: i don't really care what the string says so long as it isn't gigabytes long though, was 'censored' what you really want it to say?

geoffjentry:
@mcovarr I come across unused stuff now and then, I don't know why they're not always caught

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
mcovarr: how about a ~1000 char preview that ends in a message like this?

geoffjentry: Sure

Horneth:
If they're public they don't get flagged as unused I think

geoffjentry:
Added note in changelog

geoffjentry:
@Horneth That makes sense although I think it was either changed at some point along the way and/or it wasn't like that in my java days. Eh, whatever :)

cjllanwarne:
@mcovarr I think the "request changes" review requires another "looking good" review to dismiss it. Assuming the requested changes now look good

geoffjentry:
@cjllanwarne that assumes i pay attention to such things. also in the time you spent pointing that out you could've been reviewing the code :)

Horneth:
👍

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Cromwell Change Log
 
+## 24
+
+* When emitting workflow outputs to the Cromwell log only the first 1000 characters will be printed
```
mcovarr: 1000 characters max per output

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -624,4 +625,15 @@ object WorkflowExecutionActor {
...
\ No newline at end of file
+
+  implicit class EnhancedWorkflowOutputs(val outputs: Map[LocallyQualifiedName, WdlValue]) extends AnyVal {
+    def maxStringLength = 1000
+
+    def stripLarge = outputs map { case (k, v) =>
+      val wdlString = v.toWdlString
+
+      if (wdlString.length > maxStringLength) (k, WdlString(wdlString.take(maxStringLength) + " ... output too long, elided"))
+      else (k, v)
```
mcovarr: Are you aware of `StringUtils.abbreviate`?  It doesn't do exactly what you have here but it might be a decent alternative.

geoffjentry: I didn't realize we had commons-lang, but there it is

PR #1786, opened 2016-12-15
Clean up workflow descriptor inputs Closes #1785

### LINKED ISSUE #1785: Aliased Sub-workflows Broken
As per the test case in https://github.com/broadinstitute/centaur/pull/123.

If you assign an alias to a subworkflow call (e.g. `call subwf as subwfA`) variable resolution is not working for providing the inputs.

Horneth:


kshakir:
👍  except for naming question

engine/src/main/scala/cromwell/engine/EngineWorkflowDescriptor.scala
```diff
@@ -20,9 +19,10 @@ final case class EngineWorkflowDescriptor(namespace: WdlNamespaceWithWorkflow,
     case None => this
   }
   
-  val id = backendDescriptor.id
+  lazy val id = backendDescriptor.id
   lazy val workflow = backendDescriptor.workflow
   lazy val name = workflow.unqualifiedName
-  val inputs = backendDescriptor.inputs
+  lazy val knownDeclarations = backendDescriptor.knownValues
```
Horneth: yeah I wasn't very happy about the name either. I just didn't want it to be inputs anymore because it's confusing. 
I picked declarations because this literally contains workflow declarations + workflow inputs (from the input file), which are all declarations in wdl4s terms. 
The other reason is that there are other values we know but that are not in this map. For example some call output values are known but are provided by the outputResolver. 
If people think knownValues is less confusing I'm happy to switch it back.

mcovarr:
:+1:

PR #1787, opened 2016-12-15
Randomize execution events index so they don't get overridden

Horneth:


kshakir:
I loathe coded non-determinism (ex: `Random` is ok when using a static seed), but I don't have a better solution off the top of my head. 👍

Horneth:
The only other solution I can think of would be to query metadata to get the index but we don't want to do that :/

mcovarr:
How does this play with the timing diagram?

mcovarr:
:+1:

Horneth:
@mcovarr timing diagram explicitly sorts the events by timestamp so this doesn't affect it.

PR #1789, opened 2016-12-15
fix some optional values bugs

Horneth:


kshakir:
Not sure what this fixes, but code LGTM 👍

core/src/main/scala/cromwell/util/JsonFormatting/WdlValueJsonFormatter.scala
```diff
@@ -18,6 +18,8 @@ object WdlValueJsonFormatter extends DefaultJsonProtocol {
       case m: WdlMap => new JsObject(m.value map {case(k,v) => k.valueString -> write(v)})
       case e: WdlExpression => JsString(e.toWdlString)
       case q: WdlPair => new JsObject(Map("left" -> write(q.left), "right" -> write(q.right)))
+      case WdlOptionalValue(_, Some(opt)) => write(opt)
+      case WdlOptionalValue(_, None) => JsString("null")
```
Horneth: No this one specifically is when the workflow outputs are printed to the console at the end of the workflow.

Horneth:
It makes sure optional values are unboxed before being processed and be sent to the metadata.
Without this a defined optional array `[1, 2, 3]` would end up as a JsString `"[1, 2, 3]"` instead of JsArray in the metadata.

core/src/main/scala/cromwell/util/JsonFormatting/WdlValueJsonFormatter.scala
mcovarr: @kshakir that's a very good point.  Aesthetically I prefer no k/v, but I bet an explicit k/"null" will probably reduce support questions.  So I'm not sure which to vote for...

mcovarr:
:+1:

core/src/main/scala/cromwell/util/JsonFormatting/WdlValueJsonFormatter.scala
geoffjentry: I could see support Qs either way. I think I'd prefer to see nothing

PR #1796, opened 2016-12-17
Added docs about GCE zone for clarification

seandavi:
Just temporary, I am sure.  For myself if nothing else.


geoffjentry:
@seandavi I just opened #1797 which should take care of this more permanently. Let's consider this an easier default in case it is soundly rejected ;)

geoffjentry:
@seandavi Closing this as I believe #1797 supersedes it. If you disagree we can revisit

PR #1797, opened 2016-12-18
 Allow default zones to be set in the config. Closes #1795

### LINKED ISSUE #1795: Possible to specify runtime config (for example, zones) in application config?
I haven't been able to discern from the documentation if specifying GCE zones in the application config is possible.  I see that it is listed as runtime config.  Sorry if I am misunderstanding.

geoffjentry:


supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesBackendLifecycleActorFactory.scala
```diff
@@ -101,7 +101,7 @@ object JesBackendLifecycleActorFactory {
       RuntimeAttributeDefinition(FailOnStderrKey, Option(staticDefaults(FailOnStderrKey)), usedInCallCaching = true),
       RuntimeAttributeDefinition(MemoryKey, Option(staticDefaults(MemoryKey)), usedInCallCaching = false),
       RuntimeAttributeDefinition(DisksKey, Option(staticDefaults(DisksKey)), usedInCallCaching = false),
-      RuntimeAttributeDefinition(ZonesKey, Option(staticDefaults(ZonesKey)), usedInCallCaching = false),
+      RuntimeAttributeDefinition(ZonesKey, None, usedInCallCaching = false),
```
geoffjentry: Wasn't sure if this was the right thing to do. W/ this PR there is a hardwired default to us-central1-b if no config option is supplied at all but if they do supply a config option that won't be accurate and these point to the static defaults

kcibul:
Hey @geoffjentry this is a great change to make, didn't make sense to have a non-changeable hardcoded default!

ToL: What struck me in this PR was that it took changing of 11 files to wire a default from the config file to where it is used.  I struggled with this same thing when adding in the alternate compute service account.  Felt like a lot of boilerplate.  As a future refactoring, it would be nice if this was simplified overall

geoffjentry:
@kcibul next time I won't update tests or docs to keep the file count down ;)

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesRuntimeAttributes.scala
```diff
@@ -110,11 +109,11 @@ object JesRuntimeAttributes {
     }
   }
 
-  private def validateZone(zoneValue: WdlValue): ErrorOr[Vector[String]] = {
+  private def validateZone(zoneValue: Option[WdlValue], defaultZones: NonEmptyList[String]): ErrorOr[Vector[String]] = {
```
ruchim: TOF (Thinking of Future): maybe we'll want to validate the zones available to a google project as well? but it seems outside the scope of this ticket.

ruchim:
👍 LGTM

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesBackendLifecycleActorFactory.scala
geoffjentry: Yeah. Part of the problem is that there's an assumption that runtime attributes aren't coming from the config (which somewhat ties into @kcibul 's comment). If we want to start going down this path more frequently we'll probably need to rethink RAs in general to keep our sanity intact

supportedBackends/jes/src/test/scala/cromwell/backend/impl/jes/JesConfigurationSpec.scala
```diff
@@ -103,6 +104,11 @@ class JesConfigurationSpec extends FlatSpec with Matchers with TableDrivenProper
     new JesConfiguration(BackendConfigurationDescriptor(backendConfig, globalConfig)).root shouldBe "gs://my-cromwell-workflows-bucket"
   }
 
+  it should "have the correct default zones" in {
+    new JesConfiguration(BackendConfigurationDescriptor(backendConfig, globalConfig)).defaultZones.toList shouldBe List("us-central1-a", "us-central1-b")
+
```
Horneth: extra `\n`

supportedBackends/jes/src/test/scala/cromwell/backend/impl/jes/JesRuntimeAttributesSpec.scala
```diff
@@ -200,7 +204,7 @@ class JesRuntimeAttributesSpec extends WordSpecLike with Matchers with Mockito {
   private def assertJesRuntimeAttributesFailedCreation(runtimeAttributes: Map[String, WdlValue], exMsg: String, workflowOptions: WorkflowOptions = emptyWorkflowOptions): Unit = {
     val withDefaults = RuntimeAttributeDefinition.addDefaultsToAttributes(JesBackendLifecycleActorFactory.staticRuntimeAttributeDefinitions, workflowOptions) _
     try {
-      JesRuntimeAttributes(withDefaults(runtimeAttributes), NOPLogger.NOP_LOGGER)
+      JesRuntimeAttributes(withDefaults(runtimeAttributes), NOPLogger.NOP_LOGGER, defaultZones)
```
Horneth: I'm not sure why the other attributes defaults can be handle with `withDefaults` but zones need theirs separately, would have been nice if they were not a special case

Horneth:
Few minor comments but 👍

supportedBackends/jes/src/test/scala/cromwell/backend/impl/jes/JesRuntimeAttributesSpec.scala
geoffjentry: JesRuntimeAttributes now explicitly takes zones. It ties in w/ what I said to @ruchim that there's heavy expectation that default values aren't coming from the config and that if we want to do this more often we should rethink the whole enchilada. 

i'll take another look to see if something less hokey is doable for now.

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesBackendLifecycleActorFactory.scala
ruchim: Sounds like @kshakir is cooking up some changes to the runtime attributes anyways...

README.md
```diff
@@ -1586,7 +1586,7 @@ runtime {
 }
 ```
 
-Defaults to "us-central1-b"
+Defaults to the configuration setting `genomics.default-zones` in the JES configuration block which in turn defaults to using `us-central1-b`
```
cjllanwarne: Suggest you add this to line 1705 too (i.e. the place that lists all of the JES runtime attributes (seen in https://github.com/broadinstitute/cromwell/pull/1796)

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesBackendLifecycleActorFactory.scala
geoffjentry: Ok so the issue here is that we *could* reach into the config and grab the value that we want directly in `JesRuntimeAttributes` which would actually make the code on this PR a lot more streamlined (to the point @kcibul made about how many files were touched). However that violates the model that everything else is using in the JES backend which is that `JesAttributes` is passed the config for the JES backend and it is deconstructing everything and passing it along. In this case centralization of logic causes the issue.

I'm going to leave as-is and leave it as another point in favor of my stance that runtime attrs need to be rethought in general

README.md
geoffjentry: That's not listing the runtime attrs, that's listing workflow options. Which now that I look at it again I think it'd be weird to make a special case for this one setting (although I understand why he made that PR) considering that there are many things settable via RA and/or configuration that aren't tweakable in the workflow options

kshakir:
When rebasing and resolving conflicts this PR will need to be rewired, starting with:

```scala
object JesRuntimeAttributes {
  val runtimeAttributesBuilder: StandardValidatedRuntimeAttributesBuilder = …

  private val zonesValidation: RuntimeAttributesValidation[Vector[String]] =
    ZonesValidation.withDefault(WdlString(ZoneDefaultValue))

  def apply(validatedRuntimeAttributes: ValidatedRuntimeAttributes): JesRuntimeAttributes = …
```

converted to:

```scala
object JesRuntimeAttributes {
  def runtimeAttributesBuilder(jesConfiguration: JesConfiguration): StandardValidatedRuntimeAttributesBuilder = 

  private def zonesValidation(defaultZones: NonEmptyList[String]): RuntimeAttributesValidation[Vector[String]] =
    ZonesValidation.withDefault(WdlString(defaultZones.toList.mkString(",")))

  def apply(validatedRuntimeAttributes: ValidatedRuntimeAttributes,
            jesConfiguration: JesConfiguration): JesRuntimeAttributes = …
```

This should be fine for plumbing from the main code, as each caller has a jesConfiguration.

The test code may need some refactoring and/or mocks-- or the JRA object can add overloads, that in addition to receiving jesConfigurations, can take a nel of default zones.

geoffjentry:
@kshakir Thanks, that sounds easier than I was expecting. I was getting ready to do the old "mental rebase" on a fresh branch, I'll give this a try first

PR #1805, opened 2016-12-22
Conditionals changelog

cjllanwarne:


CHANGELOG.md
```diff
@@ -3,6 +3,47 @@
...
+  }
+
+  # Because it was declared inside the scatter and the if-block, the type of x_out is different here:
+  Array[Int?] x_out_maybes = x_out
+
+  # We can select only the valid elements with select_all:
+  Array[Int] x_out_valids = select_all(x_out_maybes)
+
+  # Or we can select the first valid element:
+  Int x_out_first = select_first(x_out_maybes)
```
Horneth: Should we say what happens if none of the elements are valid ?

geoffjentry:
Why not just reference the appropriate doc in the wdl spec (which also needs to be updated)?

I don't think of a changelog as a place for in depth documentation but rather a thing that tells me what happened that I might be interested in

PR #1812, opened 2017-01-03
Updates for Standard backends.

kshakir:
See commit message for more info.


backend/src/main/scala/cromwell/backend/standard/StandardInitializationActor.scala
```diff
@@ -0,0 +1,70 @@
...
+    Future.fromTry(Try {
+      calls foreach { call =>
+        val runtimeAttributeKeys = call.task.runtimeAttributes.attrs.keys.toList
+        val notSupportedAttributes = runtimeAttributesBuilder.unsupportedKeys(runtimeAttributeKeys).toList
+
+        if (notSupportedAttributes.nonEmpty) {
+          val notSupportedAttrString = notSupportedAttributes mkString ", "
+          workflowLogger.warn(
+            s"Key/s [$notSupportedAttrString] is/are not supported by $backendType. " +
+              s"Unsupported attributes will not be part of jobs executions.")
```
mcovarr: jobs => job

backend/src/main/scala/cromwell/backend/standard/StandardLifecycleActorFactory.scala
```diff
@@ -18,11 +23,40 @@ trait StandardLifecycleActorFactory extends BackendLifecycleActorFactory {
...
   /**
-    * Returns the main engine for async execution.
+    * Returns the initialization class.
+    *
+    * @return the initialization class.
+    */
+  def initializationActorClass: Class[_ <: StandardInitializationActor]
+
+  /**
+    * Returns the synchronous executor class. By default using the standard sync executor that should be sufficient
```
mcovarr: that => _nothing_

backend/src/main/scala/cromwell/backend/validation/RuntimeAttributesValidation.scala
```diff
@@ -78,46 +75,62 @@ object RuntimeAttributesValidation {
...
   OptionalRuntimeAttributesValidation[ValidatedType] = {
     new OptionalRuntimeAttributesValidation[ValidatedType] {
-      override def key = validation.key
+      override def key: String = validation.key
+
+      override def coercion: Traversable[WdlType] = validation.coercion
 
-      override def coercion = validation.coercion
+      override protected def validateOption: PartialFunction[WdlValue, ErrorOr[ValidatedType]] =
+        validation.validateValuePackagePrivate
```
mcovarr: ToL not sure I heart the pervasive package private aliasing pattern (seems kind of complicated), but I realize that's preexisting.

mcovarr:
:+1:

backend/src/main/scala/cromwell/backend/standard/StandardFinalizationActor.scala
```diff
@@ -0,0 +1,36 @@
+package cromwell.backend.standard
+
+import cromwell.backend._
+import cromwell.core.CallOutputs
+import wdl4s.TaskCall
+
+trait StandardFinalizationActorParams {
+  val workflowDescriptor: BackendWorkflowDescriptor
```
geoffjentry: Didn't you just say that you only wanted defs in traits? :)

backend/src/main/scala/cromwell/backend/standard/StandardFinalizationActor.scala
```diff
@@ -0,0 +1,36 @@
...
+  workflowDescriptor: BackendWorkflowDescriptor,
+  calls: Set[TaskCall],
+  jobExecutionMap: JobExecutionMap,
+  workflowOutputs: CallOutputs,
+  initializationDataOption: Option[BackendInitializationData],
+  configurationDescriptor: BackendConfigurationDescriptor
+) extends StandardFinalizationActorParams
+
+trait StandardFinalizationActor extends BackendWorkflowFinalizationActor {
+  /** Standard set of parameters passed to the backend. */
```
geoffjentry: Considering `standard` and `params` exists twice each on the following line I don't think this scaladoc adds much

backend/src/main/scala/cromwell/backend/standard/StandardInitializationActor.scala
```diff
@@ -0,0 +1,67 @@
...
+  override protected def runtimeAttributeValidators: Map[String, (Option[WdlValue]) => Boolean] = {
+    runtimeAttributesBuilder.validatorMap
+  }
+
+  override protected def coerceDefaultRuntimeAttributes(options: WorkflowOptions): Try[Map[String, WdlValue]] = {
+    RuntimeAttributesDefault.workflowOptionsDefault(options, runtimeAttributesBuilder.coercionMap)
+  }
+
+  override def validate(): Future[Unit] = {
+    Future.fromTry(Try {
```
geoffjentry: (file this under TOL): In my recent travels I've seen this pattern a lot and in a couple of circumstances it actually led to lower performance. 

In this case I'm assuming it's because the backend interface is all `Future` based but this is actually a function being called as part of the Actor's internal single thread (and thus async-ified already). Is that true?

backend/src/main/scala/cromwell/backend/standard/StandardFinalizationActor.scala
kshakir: Hah! Yes. I did. Good catch.

Mainly `val`s were painful when the base trait was trying to use a `val myField = <some default>`, later overridden by a sub class / trait. Because these fields are not actually used by the base trait, there was no pain involved… yet?

I'll look at changing them all to `def`s just in case.

backend/src/main/scala/cromwell/backend/standard/StandardFinalizationActor.scala
kshakir: In addition to naming, I'm also bad at judging how little or much to document.

geoffjentry: I'm mainly just opposed to this pattern:

    // does XYZ
    // def doXyz()

which isn't what's here, but it's in the family tree. :) Feel free to ignore

backend/src/main/scala/cromwell/backend/standard/StandardInitializationActor.scala
kshakir: Yep, it's all async-ified already. I don't like our thread queues needing to run runnables split into two (one for akka, one for scala Futures) for no real reason. As long as we're using `val ec: ExecutionContext = context.dispatcher`, I prefer:

|ec runnables queue|
|---|
|Akka `receive` launching a scala `Future`|
|The scala `Future` runnable|

Instead of:

|ec runnables queue|
|---|
|Actor `receive` just runs the runnable|

Further ToL: I'm fine measuring performance and moving blocking code to a separate thread pool.  Still in theory I'd prefer a leaf-actor do the heavy lifting, and not a scala Future.

geoffjentry: Agree w/ your last statement. May or may not on your first part, not sure I understand. I prefer to not be launching (or further modifying previously launched) `Future`s inside of an actor except in the most trivial of cases (says the person who has an incoming PR where an actor does nothing *but* launching Futures).

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesRuntimeAttributes.scala
```diff
@@ -21,28 +18,17 @@ case class JesRuntimeAttributes(cpu: Int,
                                 bootDiskSize: Int,
                                 memory: MemorySize,
                                 disks: Seq[JesAttachedDisk],
+                                /*
+ TODO: There are some interesting comments within NonPreemptibleJesPipelineInfoBuilder around dockerImage, possibly
```
geoffjentry: Despite what git blame says I'm 99% sure I wrote that. It both sounds like me and I vaguely remember it. IIRC it was more of a "this could potentially happen" thing, part of my love of type safety.

backend/src/main/scala/cromwell/backend/standard/StandardInitializationActor.scala
kshakir: Whoops. I messed up above, swapping the "I prefer:" and "Instead of:".

As the actor `receive` only processes one mailbox message at a time, by design, I can understand if one wants to break up execution and not block an actor's mailbox. I'm just not a fan of using scala futures to do that, and would prefer to see child akka actors.

geoffjentry: Ah, yes. Carry on then ;)

backend/src/test/scala/cromwell/backend/standard/StandardValidatedRuntimeAttributesBuilderSpec.scala
```diff
@@ -75,7 +75,18 @@ class SharedFileSystemValidatedRuntimeAttributesBuilderSpec extends WordSpecLike
...
-    "log a warning and fail to validate an invalid Docker entry" in {
+    "log a warning and validate an invalid Docker entry" in {
+      /*
+      NOTE: The behavior used to be: when present, a "docker" runtime attribute would always be validated to ensure that
+      the value of the runtime attribute was a String-- even if the actual runtime attribute was unsupported by the
+      backend! NOW, if the backend doesn't support docker, and the runtime attributes contain say as an invalid integer,
+      say `{ docker: 1 }`, the validation will now succeed. If we want to change this, the git history contains a
+      concept of validated-yet-still-unsupported attributes. One could use this behavior to validate other standard
+      attributes such as CPU, Memory, etc.-- checking their syntax even when they are unsupported by the current
+      backend.
```
cjllanwarne: I think that's fine since Docker is ultimately a backend-specific attribute. AFAIK only memory has a specific format specified in the WDL spec (no idea why...)

So if ChrisBackend wants `docker` to custom validation, it should be able to.

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesRuntimeAttributes.scala
```diff
@@ -21,28 +18,17 @@ case class JesRuntimeAttributes(cpu: Int,
...
+ hinting that it may been None somehow / somewhere?? The ValidatedRuntimeAttributesBuilder should not be allowing
+ dockerImage to ever be None, so not sure why this is still an Option. After discussions, if no one knows, change it.
```
cjllanwarne: I don't know why. Sounds like a good thing to fix to me.

supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystemJobCachingActorHelper.scala
```diff
@@ -1,45 +1,18 @@
...
-import cromwell.backend.callcaching.JobCachingActorHelper
-import cromwell.backend.io.{JobPathsWithDocker, WorkflowPathsBackendInitializationData}
-import cromwell.backend.validation.{RuntimeAttributesValidation, ValidatedRuntimeAttributes}
+import cromwell.backend.standard.{StandardCachingActorHelper, StandardInitializationData}
 import cromwell.core.logging.JobLogging
 import cromwell.core.path.PathBuilder
 import net.ceedubs.ficus.Ficus._
 
-trait SharedFileSystemJobCachingActorHelper extends JobCachingActorHelper {
+trait SharedFileSystemJobCachingActorHelper extends StandardCachingActorHelper {
```
cjllanwarne: I get confused, is the distinction between SFS and Config backend still useful? Does anyone else extend the Config backend?

cjllanwarne:
👍

backend/src/main/scala/cromwell/backend/standard/StandardCacheHitCopyingActor.scala
```diff
@@ -0,0 +1,43 @@
...
+trait StandardCacheHitCopyingActorParams {
+  val jobDescriptor: BackendJobDescriptor
+  val backendInitializationDataOption: Option[BackendInitializationData]
+  val serviceRegistryActor: ActorRef
+  val configurationDescriptor: BackendConfigurationDescriptor
+}
+
+/** A default implementation of the cache hit copying params. */
+case class DefaultStandardCacheHitCopyingActorParams
+(
```
kshakir: Yep, for our really long class names, intellij starts indentation two spaces after the `(`. Wrapping the `(` to the beginning of the line keeps the code from scrolling off the screen.

backend/src/main/scala/cromwell/backend/standard/StandardCachingActorHelper.scala
```diff
@@ -0,0 +1,89 @@
...
+  * Like the JobCachingActorHelper, this trait should be extended by a backend, and then __that__ trait should be mixed
+  * into a async job execution actor, and cache hit copying actor.
+  */
+trait StandardCachingActorHelper extends JobCachingActorHelper {
+  this: Actor with JobLogging =>
+
+  def backendInitializationDataOption: Option[BackendInitializationData]
+
+  /** Typed backend initialization. */
+  def backendInitializationDataAs[A <: BackendInitializationData]: A =
```
kshakir: Maybe. `backend.standard` in general is up for future modifications, unlike the relatively stable base `backend` package, so I welcome folks submitting many PRs to this package.

backend/src/main/scala/cromwell/backend/standard/StandardCachingActorHelper.scala
```diff
@@ -0,0 +1,89 @@
...
+    BackendInitializationData.as[A](backendInitializationDataOption)
+
+  /**
+    * Returns the service registry actor. Both the StandardAsyncExecutorActor and StandardCacheHitCopyingActor traits
```
kshakir: It's a pointer to keep my head straight when I come back to this code in a few months.

supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystemJobCachingActorHelper.scala
kshakir: Nope, there is no other SFS implementation besides Config, and even I get confused what goes where. In general I try to keep most of the static implementation in SFS, and anything that parses config code to turn it into WDL/scala goes in Config.

There used to be Local in addition to Config, but once I figured out a way to launch jobs async by ditching Scala's blocking `SimpleProcess` I was able to make Local a Config backend too.

Now that even more is in Standard, there's not much that is SFS specific that couldn't maybe go into Config.

Not cleaning up that mess in this PR though. Happy for one to file a ticket to sanitize the SFS/Config split later.

kshakir:
One last edit, primarily to make the JRA.dockerImage required instead of an Option. Unless I hear anything else, I'll squash after tests pass, then merge once secondary tests pass.

PR #1816, opened 2017-01-05
TES Backend Prototype. Closes #1294

### LINKED ISSUE #1294: Add GA4GH Task Execution backend
The [GA4GH Task Execution API](https://github.com/ga4gh/task-execution-schemas) is an attempt to standardize on a JES-like API which can then be supported by other vendors. Cromwell should provide a backend to support this API:
- We already support JES, and due to the similarities in API it should be an easy conversion
- Having Cromwell out of the box support this API would provide greater leverage when encouraging cloud vendors to implement it
- Having a large production engine support this API provides more gravitas to the API in the community
- The Broad supporting a GA4GH API provides happiness good cheer in the community

This will require coordinating with @kellrott for his reference implementation of the API

adamstruck:
This is a prototype backend that communicates via the GA4GH task execution [schema](https://github.com/ga4gh/task-execution-schemas). The reference implementation of the task server that we have been developing against is [here](https://github.com/bmeg/task-execution-server/tree/develop).

This is a work in progress and definitely could use some feedback from you with regards to how to better fit this in with your current code base and future development plans.  

In the coming weeks we plan to implement unit tests, add documentation and refactor the code based on your recommendations.


adamstruck:
WIP - migrating this code to use the 'standard backend'.  Thanks @geoffjentry @kshakir for the walkthrough today. I will let you know when its ready for re-review.

adamstruck:
UPDATE:

Refactored to use the standard backend.  Based on my manual crude testing, basic functionality is there.  Globs definitely don't work. 

More to come soon.

adamstruck:
@geoffjentry could you review?

kshakir:
@adamstruck I still need to do a more in depth review if you're looking for scala syntax feedback (ex: `case … => { … }` could be `case … => …`).

Early feedback:

- PR 1930 contains a few more changes to the standard backend api. I tested cherry-picking 1930 onto this PR to see what would be left to patch up. Using an "Obsolete" set of bridge code for now, [these](https://github.com/kshakir/cromwell/commit/19f3bad4ca752ac47ab6f37b694dbdaec8850b36) are the minimal changes for the updated path api, plus changes for standardized command line generation. NOTE: 1930 is still under review and may change, plus the linked github commit will be deleted once these two PRs are merged.

- The standard backend will continue to change for a while as we move more common code. For example, the script generation for globs is now centralized as of PR 1930. The only CI testing I am aware of at the moment is `sbt tesBackend/test` that runs under travis. Is there a dockerized solution yet for TES that we could use with travis centaur, like we have for JES and Local? Otherwise, the minimal patches above pass the very, very basic sbt test unit tests.

- Your PR is ok as is, but I need to think about necessity of `Async.await` a bit more. The standard backend api is synchronous, requiring the `Async.await`. But the underlying "basic" backend trait is using scala futures, and I need more insight into how those are interacting with the akka actors. For example, I wouldn't want the actors receiving multiple akka poll messages in the mailbox and then queuing up dozens of overlapping poll futures in the thread pool. I also really like that your awaits have timeouts and aren't infinite futures.

More to come. Thanks!

supportedBackends/tes/src/main/scala/cromwell/backend/impl/tes/TesAsyncBackendJobExecutionActor.scala
```diff
@@ -0,0 +1,261 @@
...
+    writeScript(commandString, backendEngineFunctions.findGlobOutputs(call, jobDescriptor))
+
+    val task = TesTask(jobDescriptor, configurationDescriptor, jobLogger, tesJobPaths, runtimeAttributes)
+
+    TesTaskMessage(
+      Some(task.name),
+      Some(task.description),
+      Some(task.project),
+      Some(task.inputs),
+      Some(task.outputs),
```
kshakir: Instead of `Some()` use `Option()`

supportedBackends/tes/src/main/scala/cromwell/backend/impl/tes/TesTask.scala
```diff
@@ -0,0 +1,225 @@
...
+    workflowDescriptor.workflowOptions.getOrElse("project", workflowName)
+  }
+
+  private val commandScript = TaskParameter(
+    "commandScript",
+    Some(fullyQualifiedTaskName + ".commandScript"),
+    tesPaths.storageInput(tesPaths.script.toString),
+    tesPaths.callExecutionDockerRoot.resolve("script").toString,
+    "File",
+    Some(false)
```
kshakir: Instead of `Some()` use `Option()` (and above on line 42)

supportedBackends/tes/src/main/scala/cromwell/backend/impl/tes/TesTask.scala
```diff
@@ -0,0 +1,225 @@
...
+        case _ => false
+    }
+    .map {
+      case (inputName, f) => TaskParameter(
+        inputName,
+        Some(workflowName + "." + inputName),
+        tesPaths.storageInput(f.valueString),
+        tesPaths.toContainerPath(f).valueString,
+        "File",
+        Some(false)
```
kshakir: Instead of `Some()` use `Option()` (and above on line 60)

supportedBackends/tes/src/main/scala/cromwell/backend/impl/tes/TesTask.scala
```diff
@@ -0,0 +1,225 @@
...
+  // TODO add TES logs to standard outputs
+  private val standardOutputs = Seq("rc", "stdout", "stderr").map {
+    f =>
+      TaskParameter(
+        f,
+        Some(fullyQualifiedTaskName + "." + f),
+        tesPaths.storageOutput(f),
+        tesPaths.containerOutput(f),
+        "File",
+        Some(false)
```
kshakir: Instead of `Some()` use `Option()` (and above on line 73)

supportedBackends/tes/src/main/scala/cromwell/backend/impl/tes/TesTask.scala
```diff
@@ -0,0 +1,225 @@
...
+      case (outputName, f: WdlSingleFile) => {
+        val outputFile = f.valueString
+        Seq(
+          TaskParameter(
+            outputName,
+            Some(fullyQualifiedTaskName + "." + outputName),
+            tesPaths.storageOutput(outputFile),
+            tesPaths.containerOutput(outputFile),
+            "File",
+            Some(false)
```
kshakir: Instead of `Some()` use `Option()` (and above on line 99)

supportedBackends/tes/src/main/scala/cromwell/backend/impl/tes/TesTask.scala
```diff
@@ -0,0 +1,225 @@
...
+            "Directory",
+            Some(false)
+          ),
+          TaskParameter(
+            globListName,
+            Some(fullyQualifiedTaskName + "." + globListName),
+            tesPaths.storageOutput(globListFile),
+            tesPaths.containerOutput(globListFile),
+            "File",
+            Some(false)
```
kshakir: Instead of `Some()` use `Option()` (and above on lines 116, 120, and 124)

supportedBackends/tes/src/main/scala/cromwell/backend/impl/tes/TesTask.scala
```diff
@@ -0,0 +1,225 @@
...
+      None,
+      tesPaths.dockerWorkflowRoot.toString
+    )
+  ) ++ workingDirVolume
+
+  val resources = Resources(
+    runtimeAttributes.cpu,
+    runtimeAttributes.memory.to(MemoryUnit.GB).amount.toInt,
+    Some(false),
+    Some(volumes),
```
kshakir: Instead of `Some()` use `Option()` (and above on lines 141, 149, and 158)

supportedBackends/tes/src/test/scala/cromwell/backend/impl/tes/TesRuntimeAttributesSpec.scala
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    "fail to validate an invalid disks entry" in {
+      val runtimeAttributes = Map("docker" -> WdlString("ubuntu:latest"), "disk" -> WdlString("blah"))
+      assertFailure(runtimeAttributes, "Expecting disk runtime attribute to be an Integer or String with format '8 GB'")
+    }
+
+    "validate a valid dockerWorkingDir entry" in {
+      val runtimeAttributes = Map("docker" -> WdlString("ubuntu:latest"), "dockerWorkingDir" -> WdlString("/tmp"))
+      val expectedRuntimeAttributes = expectedDefaults.copy(dockerWorkingDir = Some("/tmp"))
```
kshakir: Instead of `Some()` use `Option()`

supportedBackends/tes/src/main/scala/cromwell/backend/impl/tes/TesTask.scala
buchanae: Just for my own personal Scala education, could you explain this one in more detail? `volumes` is definitely I non-null value I think, so isn't that Option() just converted to Some()? Is this a style thing or is there a case where one is better than the other?

I think I learned `Some` usage from articles like this, http://alvinalexander.com/scala/best-practice-option-some-none-pattern-scala-idioms

geoffjentry: This is something we picked up from my previous team where `null`s weren't verboten and we'd get bit by this now and then.`Option(null)` is `None` whereas `Some(null)` is exactly what it says. 

Generally my preference is to avoid things which the developer "knows" to be true in favor of things which *must* be true by definition as the latter doesn't require any further thinking on a future developer - also the latter might end up becoming untrue w/o people realizing.

Another example for me is `.get` on an Option, sure there are times where I *know* that it is `Some` and won't blow up but I'd much prefer to not do that for the reasons cited above. That said even I will drop my moral stance at times due to the pain in the ass factor :)

supportedBackends/tes/src/main/scala/cromwell/backend/impl/tes/TesAsyncBackendJobExecutionActor.scala
```diff
@@ -0,0 +1,261 @@
...
+import cromwell.backend.io.JobPaths
+import cromwell.backend.sfs.SharedFileSystemExpressionFunctions
+import cromwell.backend.standard.{StandardAsyncExecutionActor, StandardAsyncExecutionActorParams, StandardAsyncJob, StandardInitializationData}
+import cromwell.backend.wdl.OutputEvaluator
+import cromwell.core.path.FileImplicits._
+import cromwell.core.path.PathFactory.pathPlusSuffix
+import cromwell.core.path.{PathBuilder, PathFactory}
+import cromwell.core.retry.SimpleExponentialBackoff
+import lenthall.util.TryUtil
+import spray.client.pipelining._
```
buchanae: To throw another wrench in the mix, another idea is to use a library generated from the protobuf/grpc task execution schemas. Seems like possibly everyone is missing out on one of the primary benefits of using protobuf and grpc by implementing everything as http/json.

supportedBackends/tes/src/main/scala/cromwell/backend/impl/tes/TesTask.scala
buchanae: Excellent explanation. Thanks @geoffjentry ! Completely agree with your reasoning.

geoffjentry: I should also toss in why do I still say this for examples like `Some(false)` where it can't possibly be null. There it's more about building up the muscle memory and the uniformity of it. Admittedly I care less in those situations.

adamstruck:
@kshakir thanks for your review. I will make the changes you requested during the coming week.

kshakir:
FYI- the failing test suite "centaurJes" is due to a know limitation in our test setup, but the other three look good, including the "centaurLocal" tests. Again, without a dockerized "centaurTes" we'll certainly try to avoid issues, but there won't be any guarantees that upgrades to the standard backend API don't break TES.

Also, the 87 commits will need to be rebased and squashed correctly to a minimal set, on the order of 1 commit, but otherwise let us know when you're ready for re-review.

Let us know if you have more questions or if we can provide any other assistance.

adamstruck:
I am working on getting centuar testing up and running with a TES implementation. Rather than going through rebase hell on this branch (I have numerous merge commits), I will submit a new PR from another branch where I have squashed the commits via merge. How does that sound?

adamstruck:
Tracking on #1979

Issue #537, opened 2016-03-10
Add Scalastyle for Cromwell


gauravs90:
Once we decide on a style guide, go through the code and resolve all the issues / errors cropping up because of the check style.

geoffjentry:
@gauravs90 What's the status of this? Are we still going w/ the XML file you & I pieced together?

gauravs90:
@geoffjentry Yes, its that same XML, and a half baked version is lying in a local branch here on my laptop, after  which there were several PRs merged (possibly needing to re-fix all those changes. I think I should plan to sit with this and complete it in just one-shot-one-sitting in several increments. (I can push that branch to remote if you wanna have a look at it, or have a go at parts of it in some spare time?)

geoffjentry:
Closing this ticket. As much as I love the idea my explorations led me to believe that it'd solve none of the nitpicky stuff that annoys people while adding *more* nitpicky stuff which would annoy people

geoffjentry:
Addendum: scalafmt looks to be promising but is not yet to a place where I'd want to use it

PR #1823, opened 2017-01-08
Various patches.

kshakir:
MetadataValue.apply was throwing an NPE exception when passed null, even though it had a call to .getOrElse("").
Consolidated standard backend runtimeAttributeDefinitions implementation.
Added a GoogleAuthModeSpec.assumeHasApplicationDefaultCredentials in tests that use application default credentials.
Refactored away JesBackendLifecycleActorFactory's toJes, only used in one place where a similar standard method now exists.
Refactored away JesBackendLifecycleActorFactory's staticRuntimeAttributeDefinitions, only used in specs.
CromwellServer no longer hard codes the binding timeout.

### DIFF
--- a/engine/src/main/scala/cromwell/server/CromwellServer.scala
+++ b/engine/src/main/scala/cromwell/server/CromwellServer.scala
@@ -2,8 +2,7 @@ package cromwell.server
 
 import java.util.concurrent.TimeoutException
 
-import akka.actor.Props
-import akka.util.Timeout
+import akka.actor.{ActorContext, ActorSystem, Props}
 import com.typesafe.config.Config
 import cromwell.core.Dispatcher.EngineDispatcher
 import cromwell.webservice.WorkflowJsonSupport._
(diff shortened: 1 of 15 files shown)


engine/src/main/scala/cromwell/server/CromwellServer.scala
```diff
@@ -52,16 +52,18 @@ object CromwellServer {
...
   override val abortJobsOnTerminate = false
 
-  override def actorRefFactory = context
-  override def receive = handleTimeouts orElse runRoute(possibleRoutes)
+  override def actorRefFactory: ActorContext = context
+  override def receive: PartialFunction[Any, Unit] = handleTimeouts orElse runRoute(possibleRoutes)
 
-  val possibleRoutes = workflowRoutes.wrapped("api", config.as[Option[Boolean]]("api.routeUnwrapped").getOrElse(false)) ~ swaggerUiResourceRoute
-  val timeoutError = APIResponse.error(new TimeoutException("The server was not able to produce a timely response to your request.")).toJson.prettyPrint
+  val routeUnwrapped: Boolean = config.as[Option[Boolean]]("api.routeUnwrapped").getOrElse(false)
```
geoffjentry: Do we really need all of these type ascriptions? a lot of these seem obvious

geoffjentry:
I question if we really need to put all of those type ascriptions everywhere (many of them were obvious, IMO) but otherwise 👍

engine/src/main/scala/cromwell/server/CromwellServer.scala
kshakir: IntelliJ added this specific ascription and the others nearby. I'm not going to fight IntelliJ until our github has a competing standard to the installed default: https://blog.jetbrains.com/scala/2016/10/05/beyond-code-style/

geoffjentry: If IntelliJ encoded all of your code in ROT13 would you just let it be?

FWIW I believe you can turn that stuff off

kshakir: Let's turn it off as a team, in git, and then everyone will pick up the setting. But yeah, if the entire team was using an IDE that turned our code to ROT13, and that's the way we chose to leave the settings in git, then yep I would let it be, to follow the team settings. Or-- if I found it untenable, I'd submit a PR to turn it off. Perhaps we can re-open 537?

geoffjentry: The problem with #537 (funny, I was just thinking of that) was that it wasn't going to solve any of the things we did bicker about and add a whole bunch of new things that were going to annoy some folks.

Also note that my last comment re scalafmt, that project hasn't really progressed as much as I'd have hoped

geoffjentry: We've had this debate before but I'll say my stance again. My beef with autoformatters for complex languages like Scala is that they tend to enforce a one size fits all formatting and I don't believe that's appropriate. For simplistic languages like Python, that's a different matter. I don't know this to be true but I've always suspected that's why projects like scalacheck are really "bad" in this way.

Horneth:
👍

PR #1831, opened 2017-01-10
Io Actor Prototype

Horneth:
Before the break I was playing around trying to async-ify Cromwell's IO a bit more. It's not complete and needs clean up / refinements / tests, but considering that one of the goals is reliability/scalability, I thought I'd make a PR out of it since it might provide a base for discussion.

This branch has an IO Actor that handles *some* of the IO that has to be done both on the engine and the backend side. Specifically the script.sh upload, rc file reading, stderr file size reading, call cache copying (on JES), workflow outputs copying is done using this mechanism.
The actor is under the service registry umbrella, that was to be able to test it more rapidly (as the service registry is already wired up pretty much everywhere), but it should probably be it's own top level actor.

Due to the Future-based approach we took in the backend interface, the IO messages (copy, read, write, delete file...) are declined into 2 different flavors:
- A classic Command -> Response
- A Promise based version, that takes a promise in the command message itself to be completed when the operation finishes. This allow for the actor to integrate with parts of the code that can't (easily) handle the response as a message.

The underlying implementation of the IO Actor is a router, but could be swapped for something else. Each worker tries to perform the operation, and once it's complete (successfully or not) either sends a message back or completes the promise depending on the command flavor.
Retries are handled by keeping an exponential backoff object in the command itself. If the failure is  retryable, the worker sends the command message back to the router after waiting for the appropriate backoff time. The message will then be rerouted when a worker is available.
Note that the actual time before the command is picked up again by another worker could be longer than intended if all workers are busy and the command spends time in the mailbox. 
A command will be retried as many times as possible (considering exponentially long waiting times in between retries) until a threshold amount of time has passed since the first try (10 minutes by default).


geoffjentry:
Cool!

Have you done any load/performance testing to get before and after numbers? If not let's chat as pet of this process

Horneth:
@geoffjentry no I haven't I actually wanted to chat w/ you about it

geoffjentry:
@horneth sweet.

services/src/main/scala/cromwell/services/io/message/IoActorCommandWithAck.scala
```diff
@@ -0,0 +1,64 @@
...
+
+/**
+  * Copy source -> destination
+  */
+case class CopyCommandMessage(source: Path,
+                              destination: Path,
+                              overwrite: Boolean = true,
+                              backoff: SimpleExponentialBackoff = defaultBackoff
+                             ) extends CopyCommand with IoActorCommandWithAck[Unit] {
+  override def withNextBackoff: CopyCommand = copy(backoff = backoff.next)
```
mcovarr: any way these could be DRYed up generically in the base trait?

geoffjentry:
I could see this being a place where workpulling from a central manager vs. pushing from a central router would be worth doing in the first go. Since these by definition are going to be slower operations although falling back to our typical "let's see if it blows up before we get around to it" can also work

engine/src/main/scala/cromwell/io/ServiceRouterActor.scala
```diff
@@ -0,0 +1,48 @@
+package cromwell.io
+
+import java.io.IOException
+
+import akka.actor.SupervisorStrategy.{Escalate, Restart}
+import akka.actor.{Actor, ActorInitializationException, OneForOneStrategy, Props}
+import akka.routing.SmallestMailboxPool
```
geoffjentry: BalancingPool might also be good to consider. I say that w/o having looked to see if you need the semantics from the SMP

Horneth: Yeah I agree BalancingPool might be better.

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -103,15 +103,15 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
     *
     * @return the execution handle for the job.
     */
-  def execute(): ExecutionHandle
+  def execute(): Future[ExecutionHandle]
```
geoffjentry: I know you touched on this in the main comment but I thought one of the points behind the AsyncBackend actors was to work around the Future based contracts in the standard backend? (Which taken out of context sounds really weird). If we're exposing futures anyways, why not get rid of the Async variant?

engine/src/main/scala/cromwell/io/NioServiceWorker.scala
```diff
@@ -0,0 +1,87 @@
...
+    File(source).copyTo(destination, overwrite = overwrite)
+  }.void
+  private def readAsString(file: Path): Try[String] = Try(File(file).contentAsString)
+  private def size(file: Path): Try[Long] = Try(File(file).size)
+  private def write(file: Path, content: String)(implicit openOptions: OpenOptions): Try[Unit] = Try(File(file).write(content)).void
+  private def delete(file: Path, swallowIOExceptions: Boolean): Try[Unit] = Try(File(file).delete(swallowIOExceptions = swallowIOExceptions)).void
+}
+
+object NioServiceWorker {
+  def props() = Props(new NioServiceWorker())
```
geoffjentry: What dispatcher are these on?

engine/src/main/scala/cromwell/io/ServiceRouterActor.scala
```diff
@@ -0,0 +1,48 @@
...
+import cromwell.services.io.IoActorCommand
+import net.ceedubs.ficus.Ficus._
+
+/**
+  * Wrapper for the IO router actor.
+  * This is a sub optimal level of indirection as messages could directly be directed to the router
+  * instead of having to go through this actor.
+  * The reason is that the Io Actor is under the umbrella of the service registry which instantiates its services generically using reflection,
+  * which can't work for a router actor.
+  * Was the router IoActor to become a first class citizen, this wrapper should be removed.
```
geoffjentry: I don't picture this as being part of the service registry - those things should be stuff that one would be reasonably expected to drop in some replacement (e.g. a different DB for metadata or KV). 

Having it top level does bring up the issue of how to have both the backends and engine talking to this thing however.

services/src/main/scala/cromwell/services/io/message/IoActorCommandWithAck.scala
geoffjentry: or type class? ;)

geoffjentry:
Cool stuff.

Can you set up a group chat for interested folks at some point after you get back, doesn't have to be right away, up to you. Would be good to walk through the design.

Horneth:
Closing for now - might be worth considering depending on what comes out of the retry meeting

PR #1833, opened 2017-01-10
Add length().  Closes #1604

### LINKED ISSUE #1604: Overload `size` function to be able to get array size too
So size should be:
- if given a file, determine file size as a WdlFloat
- if given an array, determine the array size as a WdlInteger

mcovarr:
Reopening after moving tests over to wdl4s and bumping the wdl4s version.


cjllanwarne:
👍

Horneth:
👍

PR #1834, opened 2017-01-10
Custom labels sent to JES. Closes #635. Closes #1624.

### LINKED ISSUE #635: Workflow labels: ingest at submission time
As a user, I would like to be able to supply a set of labels (key strings with optional string values) to my workflow at submission time that Cromwell will store for me associated with this workflow.  It should be returned in the workflow metadata.  

The allowable values should follow the same for the Google Pipelines API (since that's a well thought out approach).  See https://cloud.google.com/genomics/reference/rest/v1alpha2/pipelines/run#RunPipelineArgs.FIELDS.labels

This will allow me to group workflows into logical units so I can manage efficiently.  

It will also be queryable #636

It will also be propagated to Google's Pipeline API #1624

### LINKED ISSUE #1624: Propagate workflow labels to JES
JES now allows for labels to be applied to pipeline runs.  Labels are important because they allow us to tag pipeline tasks with metadata which is exposed in other Google APIs, specifically in billing exports that are visible in BigQuery.  This will allow us to, for example, calculate the exact cost of a pipeline run which is immensely important for FireCloud and @abaumann 

The changes to the pipelines API are described here:

https://cloud.google.com/genomics/reference/rest/v1alpha2/pipelines/run#RunPipelineArgs.FIELDS.labels

Cromwell should propagate the workflow level labels to the JES calls for downstream use.  According to Google, we should be able to see these labels right away in the operations metadata

cjllanwarne:


README.md
```diff
@@ -1725,6 +1704,34 @@ Valid keys and their meanings:
     * **auth_bucket** - (JES backend only) defaults to the the value in **jes_gcs_root**.  This should represent a GCS URL that only Cromwell can write to.  The Cromwell account is determined by the `google.authScheme` (and the corresponding `google.userAuth` and `google.serviceAuth`)
     * **monitoring_script** - (JES backend only) Specifies a GCS URL to a script that will be invoked prior to the WDL command being run.  For example, if the value for monitoring_script is "gs://bucket/script.sh", it will be invoked as `./script.sh > monitoring.log &`.  The value `monitoring.log` file will be automatically de-localized.
 
+# Labels
```
geoffjentry: Reminder about the convo we had in person

geoffjentry:
👍

core/src/main/scala/cromwell/core/labels/Label.scala
```diff
@@ -0,0 +1,72 @@
...
+import cats.syntax.cartesian._
+import cats.syntax.validated._
+
+sealed abstract case class Label(key: String, value: String)
+
+object Label {
+
+  // Yes, 63. Not a typo for 64.
+  // See 'labels' in https://cloud.google.com/genomics/reference/rpc/google.genomics.v1alpha2#google.genomics.v1alpha2.RunPipelineArgs
+  private val MAX_LABEL_LENGTH = 63
```
mcovarr: Scala constant style should be `MaxLabelLength`

README.md
cjllanwarne: Done.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -391,6 +396,24 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
...
 
+  private def validateLabels(json: WdlJson): ErrorOr[Labels] = {
+
+    def toLabels(inputs: Map[String, JsValue]): ErrorOr[Labels] = {
+      val vectorOfValidatedLabel: Vector[ErrorOr[Label]] = inputs.toVector map {
+        case (key, JsString(s)) => Label.validateLabel(key, s)
+        case (key, other) => s"Invalid label '$key: $other': Only JS primitives can be used in labels and must match the regex [a-z]([-a-z0-9]*[a-z0-9])?".invalidNel
+      }
+
+      vectorOfValidatedLabel.sequence[ErrorOr, Label] map { validatedVectorofLabel => Labels(validatedVectorofLabel) }
```
mcovarr: `... map Labels.apply`

core/src/main/scala/cromwell/core/labels/Label.scala
```diff
@@ -0,0 +1,72 @@
...
+
+  def validateName(s: String): ErrorOr[String] = {
+    if ("[a-z]([-a-z0-9]*[a-z0-9])?".r.pattern.matcher(s).matches) {
```
mcovarr: If ever a thing should be made a constant, this string should.

mcovarr:
:+1:

Issue #1829, opened 2017-01-10
Add test(s) to CromwellServerActor


geoffjentry:
In particular the behavior for `Timeout` is indirectly tested currently by CromwellApiServiceSpec. Test this behavior directly in the (currently nonexistent) CromwellServerActorSpec instead

geoffjentry:
closed by #1854

Issue #1828, opened 2017-01-10
Add test(s) to WorkflowQueryPagination


geoffjentry:
Currently this behavior is only tested in CromwellApiServiceSpec and I intend to remove that toot sweet as that's not part of CromwellApiService.

geoffjentry:
Closing as as per @kshakir this code shouldn't even really exist in the first place, much less be tested

Issue #1811, opened 2016-12-28
Do we care about MetadataPutAcknowledgement?


geoffjentry:
Based on a quick search I believe these are ignored 100% of the time. Considering the enormous mass of metadata writes going on this generates a lot of messages which can wind up backing up things which are actually used.

geoffjentry:
It's also worth noting that `MetadataPutFailed` manages to get logged both when we send it in the metadata service and where it's used, but we don't do anything other than those log messages. It seems like one would be enough.

kcibul:
I don't think we care about the acks per-se... but we need to not lose metadata writes.  Seems like we're somewhere in the middle now?

geoffjentry:
Writes are just super slow (as per a couple of tickets now). This is just about internal bookkeeping that we set up and never used.

geoffjentry:
example of this - in one of my tests i was generating ~1000 metadata requests per second. every single one of those in turn generated a message back to my original actor which was ignored, which would slow down the original actor.

geoffjentry:
The answer is "no".

cjllanwarne:
@geoffjentry: https://www.reddit.com/r/QTWTAIN/

Issue #1835, opened 2017-01-10
metadata-summary-refresh-interval in wrong location


geoffjentry:
After careful perusal of the code it appears that this should be in `services.MetadataService.config` but since the code which uses it does an end run around the service config it wound up in `services.MetadataService` proper.

geoffjentry:
This was fixed a while back

PR #1836, opened 2017-01-10
Rework structure of workflow store and metadata, provides roughly 5-f…

geoffjentry:
…old performance increase under load prior to DB gumming up. 

Things to note:
- Effectively removes Metadata acks & failure notices (see #1811) via no longer emitting the messages but does not fully remove them. They still technically exist, I'll remove them as part of a separate PR
- Completely reworks `CromwellApiServiceSpec` to actually be testing `CromwellApiService` and not a general integration test of our REST endpoints. Two specific tests didn't make the cut (#1828 and #1829) I'll address in separate PRs. There were other tests which did not make the cut but were already effectively being tested in their appropriate units.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 * Cromwell's Config (Shared Filesystem) backend now supports invocation of commands which run in a Docker image as a non-root user.
   The non-root user could either be the default user for a given Docker image (e.g. specified in a Dockerfile via a `USER` directive),
   or the Config backend could pass an optional `"-u username"` as part of the `submit-docker` command.
+* For MySQL users, a massive scalability improvement via batched DB writing of internal metadata events. Note that one must add `rewriteBatchedStatements=true` to their JDBC URL in their config in order to take advantage of this
 
 ### Database schema changes
 * Added CUSTOM_LABELS as a field of WORKFLOW_STORE_ENTRY, to store workflow store entries.
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -397,6 +397,20 @@ services {
     config {
       # Set this value to "Inf" to turn off metadata summary refresh.  The default value is currently "2 seconds".
       # metadata-summary-refresh-interval = "Inf"
+      # For higher scale environments, e.g. many workflows and/or jobs, DB write performance for metadata events
+      # can improved by writing to the database in batches. Increasing this value can dramatically improve overall
+      # performance but will both lead to a higher memory usage as well as increase the risk that metadata events
+      # might not have been persisted in the event of a Cromwell crash.
+      #
+      # For normal usage the default value of 1 (effectively no batching) should be fine but for larger/production
+      # environments we recommend a value of at least 500. There'll be no one size fits all number here so we recommend
+      # benchmarking performance and tuning the value to match your environment
+      # db-batch-size = 1
+      #
+      # Periodically the stored metadata events will be forcibly written to the DB regardless of if the batch size
+      # has been reached. This is to prevent situations where events wind up never being written to an incomplete batch
+      # with no new events being generated. The default value is currently 5 seconds
+      # db-flush-rate = 5 seconds
     }
   }
 }
@@ -414,7 +428,7 @@ database {
   #driver = "slick.driver.MySQLDriver$"
   #db {
   #  driver = "com.mysql.jdbc.Driver"
-  #  url = "jdbc:mysql://host/cromwell"
+  #  url = "jdbc:mysql://host/cromwell?rewriteBatchedStatements=true"
   #  user = "user"
   #  password = "pass"
   #  connectionTimeout = 5000
--- a/database/sql/src/main/scala/cromwell/database/slick/MetadataSlickDatabase.scala
+++ b/database/sql/src/main/scala/cromwell/database/slick/MetadataSlickDatabase.scala
@@ -15,7 +15,7 @@ trait MetadataSlickDatabase extends MetadataSqlDatabase {
 
   override def addMetadataEntries(metadataEntries: Iterable[MetadataEntry])
                                  (implicit ec: ExecutionContext): Future[Unit] = {
-    val action = dataAccess.metadataEntryIdsAutoInc ++= metadataEntries
+    val action = DBIO.seq(metadataEntries.grouped(2000).map(dataAccess.metadataEntries ++= _).toSeq:_*)
     runTransaction(action).map(_ => ())
   }
 
--- a/engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreActor.scala
@@ -1,242 +1,33 @@
 package cromwell.engine.workflow.workflowstore
 
-import java.time.OffsetDateTime
-
-import akka.actor.{ActorLogging, ActorRef, LoggingFSM, Props}
+import akka.actor.{Actor, ActorLogging, ActorRef, Props}
 import cats.data.NonEmptyList
 import cromwell.core._
 import cromwell.core.Dispatcher.EngineDispatcher
-import cromwell.engine.workflow.WorkflowManagerActor
-import cromwell.engine.workflow.WorkflowManagerActor.WorkflowNotFoundException
-import cromwell.engine.workflow.workflowstore.WorkflowStoreActor._
-import cromwell.engine.workflow.workflowstore.WorkflowStoreState.StartableState
-import cromwell.services.metadata.MetadataService.{MetadataPutAcknowledgement, PutMetadataAction}
-import cromwell.services.metadata.{MetadataEvent, MetadataKey, MetadataValue}
-import lenthall.util.TryUtil
-import org.apache.commons.lang3.exception.ExceptionUtils
 
-import scala.concurrent.{ExecutionContext, Future}
-import scala.util.{Failure, Success, Try}
+final case class WorkflowStoreActor(store: WorkflowStore, serviceRegistryActor: ActorRef) extends Actor with ActorLogging {
+  import WorkflowStoreActor._
 
-case class WorkflowStoreActor(store: WorkflowStore, serviceRegistryActor: ActorRef)
-  extends LoggingFSM[WorkflowStoreActorState, WorkflowStoreActorData] with ActorLogging {
+  lazy val workflowStoreSubmitActor: ActorRef = context.actorOf(WorkflowStoreSubmitActor.props(store, serviceRegistryActor))
+  lazy val workflowStoreEngineActor: ActorRef = context.actorOf(WorkflowStoreEngineActor.props(store, serviceRegistryActor))
 
-  implicit val ec: ExecutionContext = context.dispatcher
-
-  startWith(Unstarted, WorkflowStoreActorData(None, List.empty))
-  self ! InitializerCommand
-
-  when(Unstarted) {
-    case Event(InitializerCommand, _) =>
-      val work = store.initialize map { _ =>
-        log.debug("Workflow store initialization successful")
-      }
-      addWorkCompletionHooks(InitializerCommand, work)
-      goto(Working) using stateData.withCurrentCommand(InitializerCommand, sender)
-    case Event(x: WorkflowStoreActorCommand, _) =>
-      stay using stateData.withPendingCommand(x, sender)
-  }
-
-  when(Idle) {
-    case Event(cmd: WorkflowStoreActorCommand, _) =>
-      if (stateData.currentOperation.nonEmpty || stateData.pendingOperations.nonEmpty) {
-        log.error("Non-empty WorkflowStoreActorData when in Idle state: {}", stateData)
-      }
-      startNewWork(cmd, sender, stateData.withCurrentCommand(cmd, sender))
-  }
-
-  when(Working) {
-    case Event(WorkDone, data) =>
-      val newData = data.pop
-      newData.currentOperation match {
-        case None => goto(Idle) using newData
-        case Some(WorkflowStoreActorCommandWithSender(cmd, sndr)) => startNewWork(cmd, sndr, newData)
-      }
-    case Event(cmd: WorkflowStoreActorCommand, data) => stay using data.withPendingCommand(cmd, sender)
-  }
-
-  whenUnhandled {
-    case Event(MetadataPutAcknowledgement(_), _) =>
-      stay // Ignored
-    case Event(msg, _) =>
-      log.warning("Unexpected message to WorkflowStoreActor in state {} with data {}: {}", stateName, stateData, msg)
-      stay
-  }
-
-  onTransition {
-    case fromState -> toState =>
-      log.debug("WorkflowStore moving from {} (using {}) to {} (using {})", fromState, stateData, toState, nextStateData)
-  }
-
-  private def startNewWork(command: WorkflowStoreActorCommand, sndr: ActorRef, nextData: WorkflowStoreActorData) = {
-    val work: Future[Any] = command match {
-      case cmd @ SubmitWorkflow(sourceFiles) =>
-        storeWorkflowSources(NonEmptyList.of(sourceFiles)) map { ids =>
-          val id = ids.head
-          registerSubmissionWithMetadataService(id, sourceFiles)
-          sndr ! WorkflowSubmittedToStore(id)
-          log.info("Workflow {} submitted.", id)
-        }
-      case cmd @ BatchSubmitWorkflows(sources) =>
-        storeWorkflowSources(sources) map { ids =>
-          val assignedSources = ids.toList.zip(sources.toList)
-          assignedSources foreach { case (id, sourceFiles) => registerSubmissionWithMetadataService(id, sourceFiles) }
-          sndr ! WorkflowsBatchSubmittedToStore(ids)
-          log.info("Workflows {} submitted.", ids.toList.mkString(", "))
-        }
-      case cmd @ FetchRunnableWorkflows(n) =>
-        newWorkflowMessage(n) map { nwm =>
-          nwm match {
-            case NewWorkflowsToStart(workflows) => log.info("{} new workflows fetched", workflows.toList.size)
-            case NoNewWorkflowsToStart => log.debug("No workflows fetched")
-            case _ => log.error("Unexpected response from newWorkflowMessage({}): {}", n, nwm)
-          }
-          sndr ! nwm
-        }
-      case cmd @ AbortWorkflow(id, manager) =>
-        store.remove(id) map { removed =>
-          if (removed) {
-            log.debug(s"Workflow $id aborted and removed from the workflow store.")
-            manager ! WorkflowManagerActor.AbortWorkflowCommand(id, sndr)
-          } else {
-            sndr ! WorkflowAbortFailed(id, new WorkflowNotFoundException(s"Couldn't abort $id because no workflow with that ID is in progress"))
-          }
-        }
-      case cmd @ RemoveWorkflow(id) =>
-        store.remove(id) map { removed =>
-          if (removed) {
-            log.debug("Workflow {} removed from store successfully.", id)
-          } else {
-            log.warning(s"Attempted to remove ID {} from the WorkflowStore but it didn't exist", id)
-          }
-        }
-      case oops =>
-        log.error("Unexpected type of start work command: {}", oops.getClass.getSimpleName)
-        Future.successful(self ! WorkDone)
-    }
-    addWorkCompletionHooks(command, work)
-    goto(Working) using nextData
-  }
-
-  private def storeWorkflowSources(sources: NonEmptyList[WorkflowSourceFilesCollection]): Future[NonEmptyList[WorkflowId]] = {
-    for {
-      processedSources <- Future.fromTry(processSources(sources, _.asPrettyJson))
-      workflowIds <- store.add(processedSources)
-    } yield workflowIds
-  }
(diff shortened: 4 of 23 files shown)


database/sql/src/main/scala/cromwell/database/slick/MetadataSlickDatabase.scala
```diff
@@ -15,7 +15,7 @@ trait MetadataSlickDatabase extends MetadataSqlDatabase {
 
   override def addMetadataEntries(metadataEntries: Iterable[MetadataEntry])
                                  (implicit ec: ExecutionContext): Future[Unit] = {
-    val action = dataAccess.metadataEntryIdsAutoInc ++= metadataEntries
+    val action = DBIO.seq(metadataEntries.grouped(2000).map(dataAccess.metadataEntries ++= _).toSeq:_*)
```
geoffjentry: I had forgotten about this. 2000 came from Rawls and the number was believed to have been Bernick wisdom. I'd like to make this configurable in a separate PR (this thing has already spidered enough as it is)

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreEngineActor.scala
```diff
@@ -0,0 +1,166 @@
...
+import WorkflowStoreEngineActor._
+import cromwell.core.WorkflowId
+import cromwell.engine.workflow.workflowstore.WorkflowStoreActor._
+import cromwell.services.metadata.MetadataService.MetadataPutAcknowledgement
+import org.apache.commons.lang3.exception.ExceptionUtils
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.{Failure, Success}
+
+final case class WorkflowStoreEngineActor(store: WorkflowStore, serviceRegistryActor: ActorRef)
```
geoffjentry: This is mostly the same as the original `WorkflowStoreActor`, I forget the exact changes but it was mostly to make the wiring work.

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
```diff
@@ -53,14 +53,19 @@ trait CromwellApiService extends HttpService with PerRequestCreator {
       perRequest(requestContext, metadataBuilderProps, message)
   }
 
-  private def failBadRequest(t: Throwable, statusCode: StatusCode = StatusCodes.BadRequest) = respondWithMediaType(`application/json`) {
+  def handleQueryMetadataRequest(parameters: Seq[(String, String)]): Route = {
```
geoffjentry: this made me cry a bit. I needed to override `handleMetadataRequest` in order to not actually have requests going into Real Cromwell. The query stuff worked a bit different and I wasn't able to figure out enough spray-fu to avoid just doing this.

mcovarr:
:+1:

CHANGELOG.md
```diff
@@ -6,12 +6,16 @@
...
+## 25 
+
+* For MySQL users, a scalability improvement via batched DB writing of internal metadata events. **IMPORTANT** one must add `rewriteBatchedStatements=true` to their JDBC URL in their config in order to take advantage of this
 
 ## 24
 
 * When emitting workflow outputs to the Cromwell log only the first 1000 characters per output will be printed
 * Added support for conditional (`if`) statements.
 * Globs for Shared File System (SFS) backends, such as local or SGE, now use bash globbing instead of Java globbing, consistent with the JES backend.
+* For MySQL users, a massive scalability improvement via batched DB writing of internal metadata events. Note that one must add `rewriteBatchedStatements=true` to their JDBC URL in their config in order to take advantage of this
```
geoffjentry: Bad rebase on my part, most of this work predated 24's release

core/src/main/resources/reference.conf
```diff
@@ -392,7 +392,25 @@ services {
   MetadataService {
     class = "cromwell.services.metadata.impl.MetadataServiceActor"
     # Set this value to "Inf" to turn off metadata summary refresh.  The default value is currently "2 seconds".
+    # FIXME: move
```
geoffjentry: no, in fact i opened #1835 instead of just doing it here

core/src/main/resources/reference.conf
```diff
@@ -392,7 +392,25 @@ services {
...
     # metadata-summary-refresh-interval = "Inf"
+    #
+    # config {
+      # For higher scale environments, e.g. many workflows and/or jobs, DB write performance for metadata events
+      # can improved by writing to the database in batches. Increasing this value can dramatically improve overall
+      # performance but will both lead to a higher memory usage as well as increase the risk that metadata events
+      # might not have been persisted in the event of a Cromwell crash.
+      #
+      # For normal usage the default value of 1 (effectively no batching) should be fine but for larger/production
+      # environments we recommend a value of at least 500. There'll be no one size fits all number here so testing
```
geoffjentry: What about benchmarking? or actually "benchmarking & tuning" or something like that, i.e. making it clear that one should have concrete data to inform the tuning

core/src/main/resources/reference.conf
```diff
@@ -392,7 +392,25 @@ services {
...
+      # is recommended.
+      # db-batch-size = 1
```
geoffjentry: I originally did, in retrospect I think that's wrong for most users and dropped it back to 1. I don't think the defaults should be catering to a larger scale deployment and turning that number up does bring along some risk of loss of data in the event of a crash. Most small scale users will never notice the difference with the setting at 1, IMO that should be the default.

services/src/main/scala/cromwell/services/metadata/impl/WriteMetadataActor.scala
```diff
@@ -1,29 +1,96 @@
...
+    case Event(ScheduledFlushToDb, curData) =>
+      log.debug("Initiating periodic metadata flush to DB")
+      goto(WritingToDb) using curData
+  }
+
+  when(WritingToDb) {
+    case Event(ScheduledFlushToDb, curData) => stay using curData
+    case Event(PutMetadataAction(events), curData) => stay using curData.addEvents(events)
+    case Event(FlushBatchToDb, curData) =>
+      curData match {
```
geoffjentry: Looking at this I don't particularly care for it either but I don't understand what you're after w/ that comment

core/src/main/resources/reference.conf
mcovarr: :1st_place_medal:

services/src/main/scala/cromwell/services/metadata/impl/WriteMetadataActor.scala
mcovarr: Assuming the ADT:

```
case Event(FlushBatchToDb, NoEvents) => ...
case Event(FlushBatchToDb, SomeEvents(events)) => ...
```

geoffjentry: Oh right. Yes, that's why I didn't particularly care for it. Now that I reread I had misread `cases` as `classes` which made no sense (for good reason)

core/src/main/resources/reference.conf
geoffjentry: Phew I'm not going crazy. I thought I had fixed this and the changelog previously, I just hadn't pushed it.

core/src/main/resources/reference.conf
```diff
@@ -409,7 +426,7 @@ database {
   #driver = "slick.driver.MySQLDriver$"
   #db {
   #  driver = "com.mysql.jdbc.Driver"
-  #  url = "jdbc:mysql://host/cromwell"
+  #  url = "jdbc:mysql://host/cromwell?rewriteBatchedStatements=true"
```
cjllanwarne: This is an odd default considering the default batch size is `1`. Not saying the default is right or wrong one way or the other, but they're currently not really consistent with each other...

database/sql/src/main/scala/cromwell/database/slick/MetadataSlickDatabase.scala
cjllanwarne: Add the ticket before merging and I'll let it pass :)

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreActor.scala
```diff
@@ -1,242 +1,33 @@
...
-
-      MetadataEvent(MetadataKey(id, None, WorkflowMetadataKeys.SubmissionSection, WorkflowMetadataKeys.SubmissionSection_Workflow), MetadataValue(sourceFiles.wdlSource)),
-      MetadataEvent(MetadataKey(id, None, WorkflowMetadataKeys.SubmissionSection, WorkflowMetadataKeys.SubmissionSection_Inputs), MetadataValue(sourceFiles.inputsJson)),
-      MetadataEvent(MetadataKey(id, None, WorkflowMetadataKeys.SubmissionSection, WorkflowMetadataKeys.SubmissionSection_Options), MetadataValue(sourceFiles.workflowOptionsJson))
-    )
-
-    serviceRegistryActor ! PutMetadataAction(submissionEvents)
+  override def receive = {
+    case cmd: WorkflowStoreActorSubmitCommand => workflowStoreSubmitActor forward cmd
+    case cmd: WorkflowStoreActorEngineCommand => workflowStoreEngineActor forward cmd
```
cjllanwarne: ToL (presumably a hangover from the single-actor days), these actors and their commands have `Submit` and `Engine` in different places.

cjllanwarne: Also ToL, `WorkflowStoreEngineActor` isn't a great name IMO (something something glass houses)... to me it should be either `EngineWorkflowStoreActor` or `WorkflowStoreManagerActor`

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreActor.scala
```diff
@@ -1,242 +1,33 @@
...
-  when(Working) {
-    case Event(WorkDone, data) =>
-      val newData = data.pop
-      newData.currentOperation match {
-        case None => goto(Idle) using newData
-        case Some(WorkflowStoreActorCommandWithSender(cmd, sndr)) => startNewWork(cmd, sndr, newData)
-      }
-    case Event(cmd: WorkflowStoreActorCommand, data) => stay using data.withPendingCommand(cmd, sender)
-  }
+final case class WorkflowStoreActor(store: WorkflowStore, serviceRegistryActor: ActorRef) extends Actor with ActorLogging {
```
cjllanwarne: Is there any benefit to keeping this redirection shell (other than not having to rewire everything)? It seems to me that we have two totally unrelated actors now that should stand alone.

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreEngineActor.scala
```diff
@@ -0,0 +1,166 @@
...
+      val newData = data.pop
+      newData.currentOperation match {
+        case None => goto(Idle) using newData
+        case Some(WorkflowStoreActorCommandWithSender(cmd, sndr)) => startNewWork(cmd, sndr, newData)
+      }
+    case Event(cmd: WorkflowStoreActorEngineCommand, data) => stay using data.withPendingCommand(cmd, sender)
+  }
+
+  whenUnhandled {
+    case Event(MetadataPutAcknowledgement(_), _) =>
```
cjllanwarne: I thought this was gone now? Rebase needed?

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreSubmitActor.scala
```diff
@@ -0,0 +1,102 @@
...
+    * Runs processing on workflow source files before they are stored.
+    *
+    * @param processOptions How to process the workflow options
+    * @param source         Original workflow source
+    * @return Attempted updated workflow source
+    */
+  private def processSource(processOptions: WorkflowOptions => WorkflowOptionsJson)
+                           (source: WorkflowSourceFilesCollection): Future[WorkflowSourceFilesCollection] = {
+    val options = Future {
+      WorkflowOptions.fromJsonString(source.workflowOptionsJson)
```
cjllanwarne: TOL: Presumably carry-over but I didn't think we even validated that it was JSON until the WFMaterializationActor, so that we get an ID and a record for every submission, even if it's invalid?

engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
```diff
@@ -460,37 +311,23 @@ class CromwellApiServiceSpec extends FlatSpec with CromwellApiService with Scala
...
+     Get(s"/workflows/$version/${MockApiService.ExistingWorkflowId}/metadata?excludeKey=testKey2b&excludeKey=testKey3") ~>
+        cromwellApiService.metadataRoute ~>
+        check {
+          status should be(StatusCodes.OK)
+          val result = responseAs[JsObject]
+          result.fields.keys should contain allOf("testKey1a", "testKey1b", "testKey2a")
+          result.fields.keys should contain noneOf("testKey2b", "testKey3")
+          result.fields("testKey1a") should be(JsString("myValue1a"))
+          result.fields("testKey1b") should be(JsString("myValue1b"))
+          result.fields("testKey2a") should be(JsString("myValue2a"))
```
cjllanwarne: ToL I'd rather have the test actor and the test expectations get these from a single known constant source, rather than having them be magical strings in both places.

services/src/main/scala/cromwell/services/metadata/impl/MetadataDatabaseAccess.scala
```diff
@@ -78,10 +78,8 @@ trait MetadataDatabaseAccess {
       val value = metadataEvent.value map { _.value }
       val valueType = metadataEvent.value map { _.valueType.typeName }
       val jobKey = key.jobKey map { jk => (jk.callFqn, jk.index, jk.attempt) }
-      MetadataEntry(
-        workflowUuid, jobKey.map(_._1), jobKey.flatMap(_._2), jobKey.map(_._3), key.key, value, valueType, timestamp)
+      MetadataEntry(workflowUuid, jobKey.map(_._1), jobKey.flatMap(_._2), jobKey.map(_._3), key.key, value, valueType, timestamp)
```
cjllanwarne: ToL: Why do we bother exploding out the jobkey aboe? It seems like it then forces us to use the horrible `_._1` syntax on this line when we could instead say:
```
MetadataEntry(workflowUuid, key.jobKey.map(_.callFqn), key.jobKey.flatMap(_.index), key.jobKey.map(_.attempt), key.key, value, valueType, timestamp)
```

services/src/main/scala/cromwell/services/metadata/impl/WriteMetadataActor.scala
```diff
@@ -1,29 +1,96 @@
...
-          val msg = MetadataPutFailed(action, t)
-          log.error(t, "Sending {} failure message {}", sndr, msg)
-          sndr ! msg
+  sealed trait WriteMetadataActorData {
+    def addEvents(newEvents: Iterable[MetadataEvent]): WriteMetadataActorData = {
+      if (newEvents.isEmpty) this
+      else {
+        val newEventsVector = newEvents.toVector
+        val combinedEvents: Vector[MetadataEvent] = this match {
+          case NoEvents => newEventsVector
```
cjllanwarne: FWIW I like the theory here, but only if `HasEvents` enforces a non-empty value. Otherwise we aren't really using the type safety.
Similarly, if `newEvents` was types as non-empty, we shouldn't need the extra `if/else` checks here

cjllanwarne:
Some genuine test failures and a lot of ToLs. Once you sort that all out, 👍 from me.

core/src/main/resources/reference.conf
geoffjentry: Disagree on inconsistency. This setting is simply saying if anything is intended to be a batch insert to allow it. The other setting is how we batch one particular setting.

More importantly however I wanted to make it as hard as possible for someone to not accidentally leave this off (it doesn't hurt anything AFAIK to leave it on), considering how many people just copy/paste this stuff that's likely.

database/sql/src/main/scala/cromwell/database/slick/MetadataSlickDatabase.scala
geoffjentry: #1867

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreActor.scala
geoffjentry: IMO it's the same concept as an interface where I changed the implementation behind the scenes. I disagree that they're wholly unrelated, together they form what we currently think of as the workflow store, whereas previously that was accomplished by a single point.

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreActor.scala
geoffjentry: I think the different places was for an annoying but real reason. I'll look and see if I can recreate my thinking.

I think the latter suggestion is far too implementation specific in terms of exactly what is reading. I could be sold on the former (but don't see the big difference). Will also see if I can think of a different name. I wanted to go with e.g. read & write but the problem is the engine side also writes, just in a different manner

database/sql/src/main/scala/cromwell/database/slick/MetadataSlickDatabase.scala
geoffjentry: Already existed as #1851

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreEngineActor.scala
geoffjentry: covered by the now poorly named #1811 - I had a comment somewhere in here (perhaps now gone) that this PR effectively removes these but doesn't remove stuff like this

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreSubmitActor.scala
geoffjentry: I'll double check that I didn't hose something up here but I think the only thing I changed around here was turning it to being Future based. If I didn't in fact hose it up then I have no idea what was intended to happen here or what should :)

services/src/main/scala/cromwell/services/metadata/impl/MetadataDatabaseAccess.scala
geoffjentry: insert shrug emoji

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreActor.scala
geoffjentry: Thinking about it more I think read & write *is* the right thing to do here now and the need to separate submit/engine is an artifact of something which no longer the case.

That does raise the question of the dispatcher but at the moment the way our dispatchers are configured i don't think it would matter anyways. I could also define a separate workflow store dispatcher

services/src/main/scala/cromwell/services/metadata/impl/WriteMetadataActor.scala
geoffjentry: re your second sentence, admittedly it's not quite the same thing, but to me that's sort of like saying that all of the examples on how to avoid stringly typed code via wrapping them in case classes isn't helpful for extra type safety when that concept is the canonical example for such things. The idea there is that you might flub up on data ingest but you at least know you're plugging the right gozoutas into the right gozintas throughout your code.

I dont recall if/why I didn't consider moving the NEL concept further out to the outer edges in our code. Early on I was trying to avoid changing things believing this would be a surgical strike. Later on I was trying to avoid changing things as I was tired of changing things ;) The most likely answer is I encountered this in one of those two time periods.

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreSubmitActor.scala
geoffjentry: Confirmed that it wasn't a me-change. You raise a good point, and I don't think this is working the way we intended. If I'm reading things right something which fails here won't return back and they'll get a timeout. 

I'll add a ticket to investigate

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreActor.scala
geoffjentry: I realized that the reason I didn't do read & write didn't go away ;) Even the read operations involve some writing.

services/src/main/scala/cromwell/services/metadata/impl/WriteMetadataActor.scala
```diff
@@ -1,29 +1,81 @@
...
+      }
+    case Event(ScheduledFlushToDb, curData) =>
+      log.debug("Initiating periodic metadata flush to DB")
+      goto(WritingToDb) using curData
+  }
+
+  when(WritingToDb) {
+    case Event(ScheduledFlushToDb, curData) => stay using curData
+    case Event(PutMetadataAction(events), curData) =>
+      val newData = curData ++ events.toVector
```
geoffjentry: Note to self: Cats provides a NonEmptyVector, investigate performance. ADT might be back in play

PR #1853, opened 2017-01-12
Move a metadata service config option to the right location. Closes #1835

### LINKED ISSUE #1835: metadata-summary-refresh-interval in wrong location

geoffjentry:


mcovarr:
:+1:

kshakir:
👍

PR #1854, opened 2017-01-12
Add a test for Timedout behavior in CromwellServerActor directly here…

geoffjentry:
… instead of in the route tests

### DIFF
--- a/engine/src/test/scala/cromwell/server/CromwellServerSpec.scala
+++ b/engine/src/test/scala/cromwell/server/CromwellServerSpec.scala
@@ -1,0 +1,33 @@
+package cromwell.server
+
+import akka.pattern.ask
+import akka.util.Timeout
+import com.typesafe.config.ConfigFactory
+import org.scalatest.concurrent.{PatienceConfiguration, ScalaFutures}
+import org.scalatest.{FlatSpec, Matchers}
+import org.specs2.mock.Mockito
+import spray.http.{ContentTypes, HttpRequest, HttpResponse, Timedout}
+
+import scala.concurrent.duration._
+
+class CromwellServerSpec extends FlatSpec with Matchers with Mockito with ScalaFutures {
+  implicit val timeout: Timeout = 5.seconds
+
+  it should "return 500 errors as Json" in {
+    val cromwellSystem = new CromwellSystem {}
+
+    val cromwellServerActor = cromwellSystem.actorSystem.actorOf(CromwellServerActor.props(ConfigFactory.empty()))
+    val response = cromwellServerActor.ask(Timedout(mock[HttpRequest]))
+
+    response.futureValue(PatienceConfiguration.Timeout(timeout.duration)) match {
+      case h: HttpResponse =>
+        h.entity.toOption match {
+          case Some(e) => e.contentType.toString() shouldBe ContentTypes.`application/json`.mediaType.value.toString
+          case None => fail()
+        }
+      case _ => fail()
+    }
+
+    cromwellSystem.shutdownActorSystem()
+  }
+}


engine/src/test/scala/cromwell/server/CromwellServerSpec.scala
```diff
@@ -0,0 +1,31 @@
...
+
+  it should "return 500 errors as Json" in {
+    val cromwellSystem = new CromwellSystem {}
+
+    val cromwellServerActor = cromwellSystem.actorSystem.actorOf(CromwellServerActor.props(ConfigFactory.empty()))
+    val response = cromwellServerActor.ask(Timedout(mock[HttpRequest]))
+
+    response.futureValue(PatienceConfiguration.Timeout(timeout.duration)) match {
+      case h: HttpResponse =>
+        h.entity.toOption shouldBe defined
```
geoffjentry: these shoulds came from the test w/ the same name in `CromwellApiServiceSpec`

mcovarr:
:+1:

engine/src/test/scala/cromwell/server/CromwellServerSpec.scala
```diff
@@ -0,0 +1,31 @@
...
+        h.entity.toOption.get.contentType.toString() shouldBe ContentTypes.`application/json`.mediaType.value.toString
```
cjllanwarne: I know it would be a case-within-a-case but this whole `Option` decomposition would work much much nicer as a `match` statement.

engine/src/test/scala/cromwell/server/CromwellServerSpec.scala
```diff
@@ -0,0 +1,31 @@
...
+import scala.concurrent.duration._
+
+class CromwellServerSpec extends FlatSpec with Matchers with Mockito with ScalaFutures {
+  implicit val timeout: Timeout = 5.seconds
+
+  it should "return 500 errors as Json" in {
+    val cromwellSystem = new CromwellSystem {}
+
+    val cromwellServerActor = cromwellSystem.actorSystem.actorOf(CromwellServerActor.props(ConfigFactory.empty()))
+    val response = cromwellServerActor.ask(Timedout(mock[HttpRequest]))
```
cjllanwarne: What is this `Timedout` request that completely ignores its argument?!

engine/src/test/scala/cromwell/server/CromwellServerSpec.scala
cjllanwarne: Who are you, and what did you do to Jeff "since you're already fixing this up..." Gentry? 😛

cjllanwarne:
No biggies. 👍

engine/src/test/scala/cromwell/server/CromwellServerSpec.scala
geoffjentry: I don't understand what you mean?

engine/src/test/scala/cromwell/server/CromwellServerSpec.scala
geoffjentry: Wow that took even me a few minutes to figure out WTF I was saying.

Ignore the `shoulds` ;) 

I was saying that these asserts came over as-is. Why I thought that needed saying I don't remember

engine/src/test/scala/cromwell/server/CromwellServerSpec.scala
geoffjentry: This might have been why ;)

geoffjentry: I don't mind case w/in case so much for tests, my sensibilities are a lot different on specs (e.g. i prefer having the multiple explicit asserts as the previous code had). But mainly I just don't care as much in general ;)

PR #1856, opened 2017-01-12
Better JES errors + RC Closes #1848

### LINKED ISSUE #1848: Always provide RC file (or as much as possible)
In JES, the use case where a task exits without producing 1 or more expected output files will end up with the JES job in a failed status (because of non-existing output files failed to be delocalized), and the error message "Failed to delocalize files".
This looks like an error from JES where it's actually most likely an error of the task failing to produce some outputs.
If the job fails Cromwell never tries to read the RC file which makes it difficult for a user to know if it's a JES error or a task error.

1) Make sure that JES will delocalize the RC file first before trying to delocalize outputs - so it's available in the bucket.
2) Have cromwell try to read the file even if the job failed and make sure it's present in the metadata.

Horneth:
- Adds a `JesError` class that maps some known JES errors to custom Exceptions to provide better error messages. Simplistic for now but avoid unnecessary stacktrace and give more explicit error messages.
- Tries to read the return code regardless of the final status of the JES job (even if it failed). If it can read it then the return code will be available in metadata.
- Sets the exec.sh content-type to `text/plain` in gcs so it opens in the browser instead of downloading a files.


supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/errors/JesKnownJobFailure.scala
```diff
@@ -0,0 +1,25 @@
...
+case class FailedToDelocalizeFailure(message: String, jobTag: String, stderrPath: Option[Path]) extends JesKnownJobFailure {
+  lazy val stderrMessage = stderrPath map { p =>
+    s"3) Look into the stderr (${p.toRealString}) file for evidence that some of the output files the command is expected to create were not."
+  } getOrElse ""
+  
+  lazy val missingFilesMessage = if (message.contains("No URLs matched")) {
+    s"""It appears that some of the expected output files for task $jobTag did not exist when the command exited.
+       |A few things to try
+       |1) Check that the output section in your WDL is correct. Remember that all output files declared in a task must exist when the command exits.
+       |2) Check that the return code is available and is valid wrt your command expected exit code
```
mcovarr: expand "wrt"

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/errors/JesKnownJobFailure.scala
```diff
@@ -0,0 +1,25 @@
...
+import java.nio.file.Path
+
+import cromwell.backend.async.KnownJobFailureException
+import cromwell.core.path.PathImplicits._
+
+sealed trait JesKnownJobFailure extends KnownJobFailureException
+
+case class FailedToDelocalizeFailure(message: String, jobTag: String, stderrPath: Option[Path]) extends JesKnownJobFailure {
+  lazy val stderrMessage = stderrPath map { p =>
+    s"3) Look into the stderr (${p.toRealString}) file for evidence that some of the output files the command is expected to create were not."
```
mcovarr: were not => were not created

mcovarr:
:+1:

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/errors/JesError.scala
```diff
@@ -0,0 +1,53 @@
...
+    ) Option(toJobFailure(errorMessage, jobTag, stderrPath))
+    else None
+  }
+  
+  protected def toJobFailure(message: String, jobTag: String, stderrPath: Option[Path]): JesKnownJobFailure
+}
+
+private case object FailedToDelocalize extends JesError(5, 10, "Failed to delocalize files") {
+  def toJobFailure(message: String, jobTag: String, stderrPath: Option[Path]) = FailedToDelocalizeFailure(message, jobTag, stderrPath)
+}
```
cjllanwarne: newline

cjllanwarne:
👍

engine/src/main/scala/cromwell/jobstore/jobstore_.scala
```diff
@@ -2,7 +2,10 @@ package cromwell.jobstore
```
geoffjentry: what's with the filename?

PR #1857, opened 2017-01-13
Write-like functions updated to handle empty optionals

cjllanwarne:


geoffjentry:
:+1:

backend/src/main/scala/cromwell/backend/wdl/WriteFunctions.scala
```diff
@@ -30,19 +31,26 @@ trait WriteFunctions { this: WdlStandardLibraryFunctions =>
...
       file <- writeContent(wdlClass.getSimpleName.toLowerCase, tsvSerialized)
     } yield file
   }
 
-  override def write_lines(params: Seq[Try[WdlValue]]): Try[WdlFile] = writeToTsv("write_lines", params, classOf[WdlArray])
-  override def write_map(params: Seq[Try[WdlValue]]): Try[WdlFile] = writeToTsv("write_map", params, classOf[WdlMap])
-  override def write_object(params: Seq[Try[WdlValue]]): Try[WdlFile] = writeToTsv("write_object", params, classOf[WdlObject])
-  override def write_objects(params: Seq[Try[WdlValue]]): Try[WdlFile] = writeToTsv("write_objects", params, classOf[WdlArray])
-  override def write_tsv(params: Seq[Try[WdlValue]]): Try[WdlFile] = writeToTsv("write_tsv", params, classOf[WdlArray])
+  override def write_lines(params: Seq[Try[WdlValue]]): Try[WdlFile] = writeToTsv("write_lines", params, WdlArray(WdlArrayType(WdlStringType), List.empty[WdlValue]))
```
Horneth: Not for here since it would be in wdl4s, but it looks like some `X.empty` methods for `WdlArray`, `WdlMap` etc... would be convenient

Horneth:
👍

backend/src/main/scala/cromwell/backend/wdl/WriteFunctions.scala
cjllanwarne: Yeah I was halfway to adding them before I realised I was in the wrong repo :'(

PR #1865, opened 2017-01-17
Users other than root. Closes #472

### LINKED ISSUE #472: Cromwell expects docker images to be run as root user
When running a docker image through Cromwell, it assumes that you are the root user for the docker container.

I was trying to run a docker image which has to be run as a non-root user, so they don't have access to the root user home folder (/root).  The problem with this is that Cromwell will place any files you pass to the container in the /root directory, so you need to be the root user or else you will get a permission error.

mcovarr:
Note these fixes only apply to the config backend; JES does not currently seem to have a provision for specifying users other than root.  This uses `/cromwell-executions` rather than `/root` as the execution root inside the container.  Permissions of files and directories are liberalized for the users inside and outside of the container.

All tests pass if I change `reference.conf` to say `String docker_user = "nobody"`, but I've currently left `docker_user` as an unfilled (old-fashioned) `String?`.

### DIFF
--- a/core/src/main/scala/cromwell/core/path/FileImplicits.scala
+++ b/core/src/main/scala/cromwell/core/path/FileImplicits.scala
@@ -1,0 +1,36 @@
+package cromwell.core.path
+
+import java.nio.file.FileAlreadyExistsException
+import java.nio.file.attribute.PosixFilePermission
+
+import better.files.File
+
+object FileImplicits {
+
+  implicit class EnhancedFile(val directory: File) extends AnyVal {
+    def createPermissionedDirectories(): directory.type = {
+      if (!directory.exists) {
+        directory.parent.createPermissionedDirectories()
+        try {
+          directory.createDirectory()
+          // When using PosixFilePermissions/FileAttributes with createDirectories, the umask Cromwell happens to be using
+          // affects the resulting directory permissions.  This is not the desired behavior, these directories should be
+          // world readable/writable/executable irrespective of the umask.
+          directory.addPermission(PosixFilePermission.OTHERS_READ)
+          directory.addPermission(PosixFilePermission.OTHERS_WRITE)
+          directory.addPermission(PosixFilePermission.OTHERS_EXECUTE)
+        }
+        catch {
+          // Race condition that's particularly likely with scatters.  Ignore.
+          case _: FileAlreadyExistsException =>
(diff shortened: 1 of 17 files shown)


geoffjentry:
:+1: 

@kshakir Can you make sure to look at this?

core/src/test/scala/cromwell/util/TestFileUtil.scala
```diff
@@ -4,15 +4,20 @@ import java.nio.file.Path
...
+import cromwell.core.path.FileImplicits._
 
 trait TestFileUtil {
   def createCannedFile(prefix: String, contents: String, dir: Option[Path] = None): Path = {
     val suffix = ".out"
-    File.newTemporaryFile(prefix, suffix, dir.map(File.apply)).write(contents).path
+    val tempFile = File.newTemporaryFile(prefix, suffix, dir.map(File.apply))
+    tempFile.createPermissionedDirectories()
+    tempFile.toJava.setReadable(true, false)
+    tempFile.toJava.setWritable(true, false)
```
kshakir: Don't use `java.io`. Use `java.nio`:

```scala
tempFile.addPermission(PosixFilePermission.OTHERS_READ)
tempFile.addPermission(PosixFilePermission.OTHERS_WRITE)
```

core/src/main/scala/cromwell/core/path/FileImplicits.scala
```diff
@@ -0,0 +1,34 @@
...
+      if (!directory.exists) {
+        directory.parent.createPermissionedDirectories()
+        try {
+          directory.createDirectory()
+          // When using PosixFilePermissions/FileAttributes with createDirectories, the umask Cromwell happens to be using
+          // affects the resulting directory permissions.  This is not the desired behavior, these directories should be
+          // world readable/writable/executable irrespective of the umask.
+          directory.toJava.setReadable(true, false)
+          directory.toJava.setWritable(true, false)
+          directory.toJava.setExecutable(true, false)
```
kshakir: Similar to other comment about using nio:

```scala
directory.addPermission(PosixFilePermission.OTHERS_READ)
directory.addPermission(PosixFilePermission.OTHERS_WRITE)
directory.addPermission(PosixFilePermission.OTHERS_EXECUTE)
```

kshakir:
Minor cleanup of legacy api usage, then 👍

PR #1873, opened 2017-01-18
Change exit codes for usage and version to 0. Closes #1870

### LINKED ISSUE #1870: "cromwell --help" should return zero exit code (success) rather than one (failure)
When run, `cromwell --help` currently returns an exit code of 1, indicating unsuccessful execution. It would be helpful for the exit code of this command to be 0. This change would make it possible to more easily check whether cromwell has been built correctly and that it runs. The exit code for bare `cromwell` can still be non-zero since such a call lacks any actionable parameters, but when `cromwell --help` is specified, the return code would ideally be zero.

geoffjentry:
It turns out that I was the one who started the non-zero exit codes anyways.

### DIFF
--- a/src/main/scala/cromwell/Main.scala
+++ b/src/main/scala/cromwell/Main.scala
@@ -131,7 +131,7 @@ object Main extends App {
         |
       """.stripMargin)
 
-    System.exit(1)
+    System.exit(0)
   }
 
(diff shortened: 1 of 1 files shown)


ruchim:
:+1:

mcovarr:
:+1:

PR #1879, opened 2017-01-19
Bring in a usable version of prefix

mcovarr:
Happy Birthday Ruchi!  Also, fix `prefix` by upgrading the `wdl4s` dependency.


ruchim:
:+1: :+1:

PR #1904, opened 2017-01-24
unhardcode insert batch size. Closes #1851

### LINKED ISSUE #1851: Remove (soon to exist) hardcode in MetadataSlickDatabase
In the not too distant future I'm merging in a PR which will hardcode the value of `2000` as our batch size to mysql. Make this tunable

geoffjentry:


database/sql/src/main/scala/cromwell/database/slick/SlickDatabase.scala
```diff
@@ -100,6 +100,8 @@ class SlickDatabase(override val originalDatabaseConfig: Config) extends SqlData
     actionThreadPool, database.executor.executionContext.reportFailure
   )
 
+  protected[this] lazy val insertBatchSize = databaseConfig.as[Option[Int]]("db.insertBatchSize").getOrElse(2000)
```
geoffjentry: the camelCase goes against our typical convention but an astute reader will notice that the other config settings in the db block are camelCase as well

kshakir: ToL: Interesting. I only use camelCase for constants. This isn't.
https://www.scala-lang.org/files/archive/spec/2.11/06-expressions.html#constant-expressions

geoffjentry: oops - I should have been more clear, I was taking about the config option "db.inserBatchSize"

geoffjentry: Regardless were you talking about CamelCase or camelCase?

kshakir: Aha- glad you mentioned the "db" part. There be dragons messing around with the "db" config sub-namespace. Keys under "db" are actually used by slick and passed into hikari connection polling and the underlying database driver:
https://github.com/slick/slick/blob/3.1.1/slick/src/main/scala/slick/backend/DatabaseConfig.scala#L72-L73

As long as the use of the "db" sub-namespace is intended, everything should be fine. Otherwise, I'd recommend keeping config keys one level up, at `database.insertBatchSize` instead of `database.db.insertBatchSize`.

geoffjentry: ah, `database` it is. Which also means I can not feel bad about reverting to our more typical `insert-batch-size`

mcovarr:
:+1:

geoffjentry:
@kshakir updated the config value

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## 25
 
+* Added a field `insert-batch-size` to the `database` stanza which defines how many values from a batch insert will be processed at a time. This value defaults to 2000.
```
cjllanwarne: Only applies to MySQL?

README.md
```diff
@@ -477,6 +477,8 @@ database {
 }
 ```
 
+By default batch inserts will be processed in blocks of 2000. To modify this value add the field `insert-batch-size` to the `database` stanza.
```
cjllanwarne: Only applicable to MySql?

core/src/main/resources/reference.conf
```diff
@@ -434,6 +434,9 @@ database {
   #  connectionTimeout = 5000
   #}
 
+  # For batch inserts the number of inserts to send to the DB at a time
+  # insert-batch-size = 2000
```
cjllanwarne: ToL: This is actually a bit of a misnomer since it only applies to metadata inserts

cjllanwarne:
👍 after you've considered my outloudened thoughts.

CHANGELOG.md
geoffjentry: No, it'd apply to hsql as well

core/src/main/resources/reference.conf
geoffjentry: That's also the only place that has a true batch insert at the moment IIRC

geoffjentry: Confirmed. At the moment there's a 1-to-1 correspondence between metadata inserts and batch inserts.

geoffjentry: Discussed in person

database/sql/src/main/scala/cromwell/database/slick/SlickDatabase.scala
```diff
@@ -100,6 +100,8 @@ class SlickDatabase(override val originalDatabaseConfig: Config) extends SqlData
     actionThreadPool, database.executor.executionContext.reportFailure
   )
 
+  protected[this] lazy val insertBatchSize = databaseConfig.as[Option[Int]]("database.insert-batch-size").getOrElse(2000)
```
kshakir: `databaseConfig` is already localized under "database". Thus your key is currently "database.database.insert-batch-size".

geoffjentry: someone got caught saying "how could this possibly break what I had previously?"

kshakir:
Can you rebase and see if centaur on travis-ci/push is happier? I suspect your branch contains stale develop code, since travis-ci/pr is green.

geoffjentry:
@kshakir If only you had been to standup ;) Yeah, that's on the todo list.

PR #1907, opened 2017-01-24
For Joel, set google project [24 Hotfix]

ruchim:
Things to make Joel's testing easier...

### DIFF
--- a/filesystems/gcs/src/main/scala/cromwell/filesystems/gcs/GcsPathBuilder.scala
+++ b/filesystems/gcs/src/main/scala/cromwell/filesystems/gcs/GcsPathBuilder.scala
@@ -53,10 +53,16 @@ class GcsPathBuilder(authMode: GoogleAuthMode,
                      options: WorkflowOptions) extends PathBuilder {
   authMode.validate(options)
 
-  protected val storageOptions = StorageOptions.builder()
-    .authCredentials(authMode.authCredentials(options))
-    .retryParams(retryParams)
-    .build()
+  protected val storageOptionsBuilder = StorageOptions.builder()
+                                  .authCredentials(authMode.authCredentials(options))
+                                  .retryParams(retryParams)
+
+  // Grab the google project from Workflow Options if specified and set
+  // that to be the project used by the StorageOptions Builder
+  options.get("google_project") map storageOptionsBuilder.projectId
+
+
+  protected val storageOptions = storageOptionsBuilder.build()
 
   // The CloudStorageFileSystemProvider constructor is not public. Currently the only way to obtain one is through a CloudStorageFileSystem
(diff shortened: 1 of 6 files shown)


geoffjentry:
General note: When this gets merged, make sure the same changes are applied to dev

filesystems/gcs/src/main/scala/cromwell/filesystems/gcs/GcsPathBuilder.scala
```diff
@@ -53,10 +53,21 @@ class GcsPathBuilder(authMode: GoogleAuthMode,
                      options: WorkflowOptions) extends PathBuilder {
   authMode.validate(options)
 
-  protected val storageOptions = StorageOptions.builder()
-    .authCredentials(authMode.authCredentials(options))
-    .retryParams(retryParams)
-    .build()
+  val googleProjectProvided = options.getOrElse("google_project", None)
+
```
geoffjentry: I believe what you should be able to do is something like this (I'm just free typing it, may not even compile as-is)

    val foo = StorageOptions.builder()
                     .authCredentials(authMode.authCredentials(options))
                     .retryParams(retryParams)
    val blah = googleProjectProvided match {
      case project: String => foo.projectId(project)
      case _ => foo
    }
    blah.build()

That looks like it could be done better but the point being that you can do most of the builder and then tack on the `projectId` part as necessary.

filesystems/gcs/src/main/scala/cromwell/filesystems/gcs/GcsPathBuilder.scala
```diff
@@ -53,10 +53,21 @@ class GcsPathBuilder(authMode: GoogleAuthMode,
...
+  protected val storageOptions = googleProjectProvided match {
+    case project: String =>
+      StorageOptions.builder()
+        .authCredentials(authMode.authCredentials(options))
+        .retryParams(retryParams)
+        .projectId(project)
+        .build()
+    case None =>
```
geoffjentry: I forget why this was here and what the right thing to do was but remember that this caused that funky type error. You don't want to have String and None as your two match types.

filesystems/gcs/src/main/scala/cromwell/filesystems/gcs/GcsPathBuilder.scala
ruchim: code has been adjusted

Horneth:
👍

cjllanwarne:
👍

cjllanwarne:
👍

Horneth:
@ruchim I think this can be closed since it's been cherry picked and merged into 24_hotfix

PR #1942, opened 2017-02-01
De-dupe EJEA events if call cache writing is enabled. Closes #1900

### LINKED ISSUE #1900: Execution event duplicated by EJEA when call caching enabled
Visible in the timing diagram, when the conf file has `call-caching.enabled = true`, the EJEA appends the execution events in two areas of the code:

screencapture-local-broadinstitute-org-8000-api-workflows-v1-1e5f7c34-6c90-4bbd-9b12-097b8de1aaa6-timing-1485208997381

The `executionEvents` appending should only happen once within the lifetime of the EJEA. The two blocks of code that append when call-caching is enabled are:
- [`saveJobCompletionToStore`](https://github.com/broadinstitute/cromwell/blob/7e2105097fb1c3f44db5a9b1969d5cdf2e0362bb/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala#L411-L413)
- a pair of sibling event handlers under [`when(RunningJob)`](https://github.com/broadinstitute/cromwell/blob/7e2105097fb1c3f44db5a9b1969d5cdf2e0362bb/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala#L214-L220)

cjllanwarne:


geoffjentry:
👍

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -218,6 +218,9 @@ class EngineJobExecutionActor(replyTo: ActorRef,
       eventList ++= response.executionEvents
       log.debug(s"Got job result for {}, awaiting hashes", jobTag)
       stay using data.withSuccessResponse(response)
+    case Event(response: JobSucceededResponse, data: ResponsePendingData) =>
+      eventList ++= response.executionEvents
+      saveJobCompletionToJobStore(data.withResponse(response))
```
kshakir: ToL: Curious, any reason `withResponse` was used here, versus the "alias" `withSuccessResponse` above?

kshakir:
👍

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
cjllanwarne: Huh. Looking at the two methods, `withSucceessData` just short-cuts a `match`/`case` check in `withResponse`. Shrug... I might as well use it I suppose!

PR #1947, opened 2017-02-02
Asynced the standard backend execute/recover.

kshakir:
Added an extra akka message check just in case we miss the abort message when our scala future eventually runs.

### DIFF
--- a/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
+++ b/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
@@ -105,7 +105,19 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
     *
     * @return the execution handle for the job.
     */
-  def execute(): ExecutionHandle
+  def execute(): ExecutionHandle = {
+    throw new NotImplementedError(s"Neither execute() nor executeAsync() implemented by $getClass")
+  }
+
+  /**
+    * Async execute the job specified in the params. Should return a `StandardAsyncPendingExecutionHandle`, or a
+    * `FailedExecutionHandle`.
+    *
+    * @return the execution handle for the job.
+    */
+  def executeAsync()(implicit ec: ExecutionContext): Future[ExecutionHandle] = {
+    Future.fromTry(Try(execute()))
+  }
 
   /**
     * Recovers the specified job id, or starts a new job. The default implementation simply calls execute().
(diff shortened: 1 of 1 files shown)


backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -238,46 +260,67 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
...
+
+  private def executeOrRecoverSuccess(executionHandle: ExecutionHandle): ExecutionHandle = {
+    executionHandle match {
+      case handle: PendingExecutionHandle[
+        StandardAsyncJob@unchecked, StandardAsyncRunInfo@unchecked, StandardAsyncRunStatus@unchecked] =>
+        tellKvJobId(handle.pendingJob)
+        jobLogger.info(s"job id: ${handle.pendingJob.jobId}")
+        tellMetadata(Map(CallMetadataKeys.JobId -> handle.pendingJob.jobId))
+        /*
+        NOTE: Because of the async nature of the scala futures, there is a point in time where we have submitted this or
```
mcovarr: the scala futures => Scala Futures

mcovarr:
:+1:

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -105,7 +105,19 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
     *
     * @return the execution handle for the job.
     */
-  def execute(): ExecutionHandle
+  def execute(): ExecutionHandle = {
+    throw new NotImplementedError(s"Neither execute() nor executeAsync() implemented by $getClass")
```
kshakir: These sync versions are only ever called by the default async implementations. If the async is overridden, ok. If the sync is overridden, the default async calls the sync, so this is ok too. If neither async nor sync is overridden, the default async calls this default sync and you get an exception.

mcovarr: What's to stop something else from calling the default sync version?

geoffjentry:
:+1:

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
kshakir: Not much. These methods should be overridden but not invoked. I could make them protected, but this is inside an actor so no one else has a reference to this object outside subclasses, specs, and akka.

The only callers to the async method, that by default just calls the sync method, should be the StandardAsyncExecutorActor itself in its executeOrRecover implementation.

PR #1949, opened 2017-02-03
More tests for conf and db

kshakir:


mcovarr:
:+1: a comment in the tests as to why we want to test those things would be awesome

geoffjentry:
👍 - for my own edification I'm curious if this is something which actually bit you (and what was the symptom) or is it just purity?

kshakir:
The conf issue was discovered while reviewing the merged reference.conf with Kate. I was showing her the difference between reference and application confs. We noticed that cromwell's reference conf was attempting to override a value and silently failing.

The db tests were just extra tests on top of the emergency fix from last friday. After the actual fix, I thought I'd add tests showing why the SqlConverters were converting empty lobs to nulls. While there, I put in a test showing a slick deadlock also.

geoffjentry:
@kshakir Cool. I forgot that there were two types of tests, was asking about the conf one.

PR #1953, opened 2017-02-03
Fix Travis Encryptions

ruchim:
So sorry that my experimentation broke your builds, hopefully this will revive everything!


.travis.yml
```diff
@@ -33,3 +33,4 @@ deploy:
   script: src/bin/travis/publishRelease.sh
   on:
     tags: true
+
```
mcovarr: please revert if this isn't necessary

mcovarr:
:+1:

PR #1969, opened 2017-02-10
Docker Hash

Horneth:
- [x] README update
- [x] Changelog update
- [x] What should happen to `call-caching.lookup-docker-hash` conf entry ?


core/src/main/scala/cromwell/core/callcaching/docker/DockerHashActor.scala
```diff
@@ -0,0 +1,209 @@
...
+   *  + expireAfterWrite sets the time after which cache entries must expire. 
+   *      We use expireAfterWrite (as opposed to expireAfterAccess because we want to the entry to expire
+   *      even if it's accessed. The goal here is to force the actor to ask again for the hash after a certain 
+   *      amount of time to guarantee its relative accuracy.
+   *      
+   *  + maximumSize sets the maximum amount of entries the cache can contain. 
+   *    If/when this size is reached, least used entries will be expired
+   */  
+  private val cache = CacheBuilder.newBuilder()
+    .concurrencyLevel(1)
```
Horneth: Should be 2 as per the comment

README.md
```diff
@@ -1609,6 +1609,35 @@ runtime {
 
 This attribute is mandatory when submitting tasks to JES. When running on other backends, they default to not running the process within Docker.
 
+### Docker Tags
+
+Docker tags are a convenient way to point to a version of an image (ubuntu:14.04), or even the latest version (ubuntu:latest).
+For that purpose, tags are mutable, in a sense that the image they point to can change, while the tag name stays the same.
+While this is very convenient in some cases, using mutable, or "floating" tags in WDL affects the reproducibility of the WDL file: the same WDL using "ubuntu:latest" run now, and a year, or even a month from now will actually run with different docker images.
```
mcovarr: Reproducibility of results, really...

README.md
```diff
@@ -1609,6 +1609,35 @@ runtime {
...
+This has an even bigger impact when Call Caching is turned on in Cromwell, and could lead to false positive cache hits.
+Docker provides another way of identifying an image version, using the specific digest of the image. The digest is guaranteed to be different if 2 images have the same byte content. For more information see https://docs.docker.com/registry/spec/api/#/content-digests
```
mcovarr: Same => different

README.md
```diff
@@ -1609,6 +1609,35 @@ runtime {
...
+
+When Cromwell finds a job ready to be run, it will first look at its docker runtime attribute, and apply the following logic:
+- The job doesn't specify a docker image:
+    The job will be dispatched and all call caching settings (read/write) will apply normally.
+- The job does specify a docker runtime attribute:
+    - The docker image uses a hash (e.g. ubuntu@sha256:71cd81252a3563a03ad8daee81047b62ab5d892ebbfbf71cf53415f29c130950):
+        All call caching settings apply normally.
+    - The docker image uses a floating tag:
+        Cromwell will look up the digest of the image, and replace the user's docker value with the digest. This way cromwell can report with certainty which specific image was used to run a job.
+        However, because this is done dynamically when a job is about to run, odd scenarios can occur where a tag can be updated in the middle of a workflow.
```
mcovarr: If we're replacing a floating tag with a hash I'm not sure where the uncertainty would be?

README.md
mcovarr: Not really "mutable"

README.md
```diff
@@ -1807,6 +1778,35 @@ Cromwell also accepts two [workflow option](#workflow-options) related to call c
...
 > **Note:** If call caching is disabled, the to workflow options `read_from_cache` and `write_to_cache` will be ignored and the options will be treated as though they were 'false'.
 
+## Docker Tags
+
+Docker tags are a convenient way to point to a version of an image (ubuntu:14.04), or even the latest version (ubuntu:latest).
+For that purpose, tags are mutable, in a sense that the image they point to can change, while the tag name stays the same.
+While this is very convenient in some cases, using mutable, or "floating" tags in WDL affects the reproducibility of the WDL file: the same WDL using "ubuntu:latest" run now, and a year, or even a month from now will actually run with different docker images.
+This has an even bigger impact when Call Caching is turned on in Cromwell, and could lead to false positive cache hits.
+Docker provides another way of identifying an image version, using the specific digest of the image. The digest is guaranteed to be different if 2 images have the same byte content. For more information see https://docs.docker.com/registry/spec/api/#/content-digests
+A docker image with digest can be referenced as follow :
```
mcovarr: Follows

README.md
```diff
@@ -1778,6 +1778,35 @@ Cromwell also accepts two [workflow option](#workflow-options) related to call c
...
+- The job does specify a docker runtime attribute:
+    - The docker image uses a hash (e.g. ubuntu@sha256:71cd81252a3563a03ad8daee81047b62ab5d892ebbfbf71cf53415f29c130950):
+        All call caching settings apply normally.
+    - The docker image uses a floating tag:
+        Cromwell will look up the digest of the image, and replace the user's docker value with the digest. This way cromwell can report with certainty which specific image was used to run a job.
+        However, because this is done dynamically when a job is about to run, odd scenarios can occur where a tag can be updated in the middle of a workflow.
+        For that reason, the job will not be eligible for call caching. Which means that even if call caching reading is turned on, Cromwell will not look for an existing job in the cache database.
+        If call caching writing is turned on, Cromwell will still write the job in the cache database, using the image with digest
+        
+Cromwell can only lookup Dockerhub and Google Container Engine image digests. Provided the right credentials, it will be able to lookup private images.
```
mcovarr: Docker Hub

README.md
```diff
@@ -1778,6 +1778,35 @@ Cromwell also accepts two [workflow option](#workflow-options) related to call c
...
+For that purpose, tags are mutable, in a sense that the image they point to can change, while the tag name stays the same.
+While this is very convenient in some cases, using mutable, or "floating" tags in WDL affects the reproducibility of the WDL file: the same WDL using "ubuntu:latest" run now, and a year, or even a month from now will actually run with different docker images.
+This has an even bigger impact when Call Caching is turned on in Cromwell, and could lead to false positive cache hits.
+Docker provides another way of identifying an image version, using the specific digest of the image. The digest is guaranteed to be different if 2 images have the same byte content. For more information see https://docs.docker.com/registry/spec/api/#/content-digests
+A docker image with digest can be referenced as follow :
+    ubuntu@sha256:71cd81252a3563a03ad8daee81047b62ab5d892ebbfbf71cf53415f29c130950
+The above image refers to a specific image of ubuntu, that does not depend on a floating tag.
+A wdl containing this docker image run now and a year from now will run on the exact same image.
```
mcovarr: WDL

mcovarr: Docker

core/src/main/scala/cromwell/core/callcaching/docker/registryv2/DockerRegistryV2AbstractFlow.scala
```diff
@@ -0,0 +1,285 @@
...
+
+/**
+  * Implements logic to build a flow that can retrieve a docker hash 
+  * from a registry server conforming to the docker registry API v2 Specification:
+  * https://docs.docker.com/registry/spec/api/
+  */
+abstract class DockerRegistryV2AbstractFlow(httpClientFlow: HttpDockerFlow)(implicit ec: ExecutionContext, materializer: ActorMaterializer) extends DockerFlow {
+  
+  // Http requests take headers as a scala.collection.immutable.Seq
+  // This implicitly converts type Seq collections to immutable
```
mcovarr: where are the mutable `Seq`s coming from?

core/src/main/scala/cromwell/core/callcaching/docker/registryv2/DockerRegistryV2AbstractFlow.scala
```diff
@@ -0,0 +1,285 @@
...
+    * @tparam B type of the output value of the flow, if successful
+    * @return flow
+    */
+  private def requestTransformFlow[A, B](
+                                          requestBuilderFunction: A => (HttpRequest, DockerHashContext),
+                                          responseHandlerFlow: Flow[(HttpResponse, DockerHashContext), (Try[B], DockerHashContext), NotUsed]
+                                        ) = GraphDSL.create() { implicit builder =>
+    import GraphDSL.Implicits._
+
+    // Build a request from a A
```
mcovarr: an

core/src/main/scala/cromwell/core/callcaching/docker/registryv2/flows/HttpFlowWithRetry.scala
```diff
@@ -0,0 +1,120 @@
...
+    source ~> mergeRequests.in(0)
+    
+    // Submit request to underlying http flow  -  Partition responses: Failure | Success
+    mergeRequests     ~>     http      ~>         partitionResponseTry.in
+                                                        // Success -> analyze response and partition into 3 (see above)
+                                                        requestSuccessful ~> partitionHttpResponse
+    // Retry failed requests (Try[HttpResponse] was a Failure)
+    mergeRequests.in(1)     <~ retryBuffer <~     requestFailed.outlet
+    // Retry retryable failed responses (the HttpResponse had a non successful return code and was deeemed retryable)
+    mergeRequests.in(2)        <~        retryBuffer         <~        responseRetryable.outlet
```
mcovarr: it's like programming in ASCII art 😮

core/src/main/scala/cromwell/core/callcaching/docker/registryv2/flows/gcr/GoogleFlow.scala
```diff
@@ -0,0 +1,59 @@
...
+
+  override def buildFlow() = GraphDSL.create() { implicit builder =>
+    import GraphDSL.Implicits._
+
+    // This flow should only receive images that have been vetted by accepts.
+    def mapContextToFlow(dockerHashContext: DockerHashContext) = {
+      if (gcrFlow.accepts(dockerHashContext.dockerImageID)) 0
+      else if (usGcrFlow.accepts(dockerHashContext.dockerImageID)) 1
+      else if (usGcrFlow.accepts(dockerHashContext.dockerImageID)) 2
+      else 3
```
mcovarr: any reason you couldn't use the "index of accepting object in the collection" magic like in the other spot?

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@
 * For MySQL users, a massive scalability improvement via batched DB writing of internal metadata events. Note that one must add `rewriteBatchedStatements=true` to their JDBC URL in their config in order to take advantage of this
 * In some cases the SFS backend, used for Local, SGE, etc., coerced `WdlFile` to `WdlString` by using `.toUri`. This
 resulted in strings prepended with `file:///path/to/file`. Now absolute file paths will not contain the uri scheme.
+* Added docker hash lookup. Docker images without an explicit digest value will disqualify the task for call caching. See https://github.com/broadinstitute/cromwell#call-caching-docker-tags for more details
```
geoffjentry: `Added docker hash lookup for call caching` or something like that, i.e. make it clear that this is CC related earlier than later.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/preparation/JobPreparationActor.scala
```diff
@@ -0,0 +1,234 @@
...
+    val requestLostTimer = newRequestLostTimer
+    val backpressureTimer = newBackPressureTimer(data.dockerHashRequest)
+    stay() using data.backpressure(requestLostTimer, backpressureTimer)
+  }
+  
+  private def callCacheIneligibleWithHash(dockerWithHash: String): CallCachingIneligible = {
+    val message = s"""You are using a floating docker tag in this task. Cromwell does not consider tasks with floating tags to be eligible for call caching.
+        |If you want this task to be eligible for call caching in the future, use a docker runtime attribute with a digest instead.
+        |This is the exact docker image that was used for this job: $dockerWithHash
+        |You can replace the docker runtime attribute in your task with the above value to make this task eligible for call caching.""".stripMargin
```
mcovarr:
```
runtime {
  contemptible: true
}
```

README.md
```diff
@@ -1778,6 +1778,35 @@ Cromwell also accepts two [workflow option](#workflow-options) related to call c
 
 > **Note:** If call caching is disabled, the to workflow options `read_from_cache` and `write_to_cache` will be ignored and the options will be treated as though they were 'false'.
 
+## Docker Tags
```
geoffjentry: @katevoss @vdauwera In terms of the tutorials and other best practice WDL stuff (vs GATK best practices) we need to make a big deal about this

mcovarr:
:+1:

Very nice!  I need to go back and re-read this after my brain cools down, but the tests give me a lot of confidence it's all working.  🙂

core/src/main/scala/cromwell/core/callcaching/CallCachingEligibility.scala
```diff
@@ -0,0 +1,5 @@
+package cromwell.core.callcaching
+
+sealed trait CallCachingEligibility
+case object CallCachingEligible extends CallCachingEligibility
+case class CallCachingIneligible(reason: String) extends CallCachingEligibility
```
geoffjentry: Generally when I say something like that, this is what I mean :) What are you looking for?

core/src/main/scala/cromwell/core/callcaching/docker/registryv2/flows/HttpFlowWithRetry.scala
```diff
@@ -0,0 +1,120 @@
...
+      case StatusCodes.InternalServerError => true
+      case _ => false
+    }
+  }
+
+  /**
+    * In order to allow for retries, the http context object needs to encapsulate the original request,
+    * so that it can be re-submitted if necessary.
+    * This type provides this encapsulation by creating a pair of (T, HttpRequest) where T is the custom type that
+    * the user wishes to use as a context for his specific use case.
```
geoffjentry: their

core/src/main/scala/cromwell/core/callcaching/CallCachingEligibility.scala
mcovarr: All the flavors of failure broken out as they are elsewhere in this PR.

engine/src/main/scala/cromwell/server/CromwellRootActor.scala
```diff
@@ -53,6 +66,20 @@ import net.ceedubs.ficus.Ficus._
   lazy val callCacheReadActor = context.actorOf(RoundRobinPool(25)
     .props(CallCacheReadActor.props(callCache)),
     "CallCacheReadActor")
+  
+  // Docker Actor
+  lazy val ioEc = context.system.dispatchers.lookup(Dispatcher.IoDispatcher)
+  lazy val gcrQueriesPer100Sec = config.getAs[Int]("docker.gcr-api-queries-per-100-seconds") getOrElse 50000
```
Horneth: It's what we have on broad-dsde-cromwell-dev :/

geoffjentry: Probably it is the real default but worth looking up in their docs.

README.md
Horneth: Let's say you have a really wide scatter and the tag gets updated in the middle of it. Half of the shards would get a hash and the other half another hash. Which means potentially half of them could have a cache hit but not the other..

core/src/main/scala/cromwell/core/callcaching/docker/registryv2/flows/HttpFlowWithRetry.scala
```diff
@@ -0,0 +1,120 @@
...
+    // Buffer for retryable requests. Will DROP requests when full
+    // Will also delay re submission
+    val retryBuffer = Flow[(HttpRequest, (T, HttpRequest))]
+      .buffer(retryBufferSize, OverflowStrategy.dropHead)
+
+    // Submit outside requests to the first port of the mergeRequests merger 
+    source ~> mergeRequests.in(0)
+    
+    // Submit request to underlying http flow  -  Partition responses: Failure | Success
+    mergeRequests     ~>     http      ~>         partitionResponseTry.in
```
geoffjentry: what's up w/ all the wackadoo spacing?

CHANGELOG.md
Horneth: It has an impact on call caching but it's actually happening wether or not call caching is turned on though

geoffjentry: Right but will anyone care that it exists if they're not using call caching?

core/src/main/scala/cromwell/core/callcaching/docker/DockerHashActor.scala
```diff
@@ -0,0 +1,209 @@
...
+      val exception = new RuntimeException(s"Failed to enqueue docker hash request ${dockerHashContext.request}. Queue was closed")
+      logAndRestart(exception)
+    case EnqueueResponse(QueueOfferResult.Failure(failure), dockerHashContext) => 
+      logAndRestart(failure)
+    case FailedToEnqueue(throwable, dockerHashContext) =>
+      logAndRestart(throwable)
+  }
+  
+  private def logAndRestart(throwable: Throwable) = {
+    log.error("Failed to process docker hash request", throwable)
```
Horneth: Would `warning` be better ? This is really not supposed to happen so if it does we should be able to see it I think

core/src/main/scala/cromwell/core/callcaching/docker/registryv2/DockerRegistryV2AbstractFlow.scala
Horneth: The `Seq` type is an alias for `scala.collection.Seq`, not `scala.collection.immutable.Seq` apparently

geoffjentry: One of the many reasons I keep asking people to stop using Seq everywhere, in particular with return types (there's no reason a return type needs to be uber-generic).

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/preparation/JobPreparationActor.scala
```diff
@@ -0,0 +1,233 @@
...
+  private final def newRequestLostTimer = context.system.scheduler.scheduleOnce(dockerNoResponseTimeout, self, DockerNoResponseTimeout)(ec, self)
+  // Schedule a DockerHashRequest to be sent to the docker hashing actor
+  private final def newBackPressureTimer(dockerHashRequest: DockerHashRequest) = context.system.scheduler.scheduleOnce(backpressureWaitTime, dockerHashingActor, dockerHashRequest)(ec, self)
+
+  /*
+   * If we get backpressured, recreate both timers
+   */
+  private def handleDockHashBackPressure(request: DockerHashRequest, data: JobPreparationActorData) = {
+    val requestLostTimer = newRequestLostTimer
+    val backpressureTimer = newBackPressureTimer(data.dockerHashRequest)
```
Horneth: The `backpressure` method on the FSM data called right after cancels them. But I think you're right it's actually better to cancel them explicitly before creating the new ones 👍

core/src/main/scala/cromwell/core/callcaching/docker/registryv2/flows/HttpFlowWithRetry.scala
Horneth: It's to try and align the blocks vertically with each other in a meaningful way:
http://doc.akka.io/docs/akka/2.4.16/scala/stream/stream-composition.html#Composing_complex_systems

engine/src/main/scala/cromwell/server/CromwellRootActor.scala
Horneth: Hmm apparently the docs say 10 / sec https://cloud.google.com/container-engine/pricing
Which is clearly not 50000 / 100 sec 
We can tune it down to 1000 / 100 sec ?

geoffjentry: Yeah, make the default whatever is actually the default.

core/src/main/scala/cromwell/core/callcaching/docker/registryv2/flows/gcr/GoogleFlow.scala
Horneth: hmm nop I think it would work here too, I'll change it

Horneth:
Note for future tech talk: The hash lookup will not work for private images when running locally, even though the job might run fine. This is because when running locally, as long as the user is logged in properly (with `docker login` or `gcloud login`), the docker command will be able to retrieve the image. However Cromwell does not go look for the user's docker config file on the machine to extract the auth information and use it when looking up the hash.
Another option would be to allow explicit declaration of authentication strategies in the config file (like we have for google). Currently only JES has a `dockerhub` entry. It could be generalize at the root level with something like
```

dockerhub {
     auths [
        {
         # this would look at ~/.docker/config.conf for example
          name = "application-default"
          scheme = "application_default"
        },
        {
          name = "custom"
          scheme = "custom"
          account = "bla"
          token = "bla"
        }
     ]
}
```

and then any backend could do

`dockerhub-lookup.auth = "application-default"`

CHANGELOG.md
Horneth: I feel like it adds extra robustness as it allows us to be able to guarantee to the user (provided that we can get the hash), that their job ran with a specific docker image. Regardless of call caching. But you're right 99.9% won't care if they don't use call caching. `Added docker hash lookup for call caching` makes it sounds like it only happens if call caching is enabled which is not true. But I can rephrase it to make it clear that unless you use call caching you probably won't care.

geoffjentry: I was just trying to view it from the lens of someone reading changelog. My first question is going to be "so? what does that mean?" - and i tend to want instant answers w/ changelog-y stuff. See your point though, so if you can think of way that hits at both sides that's also cool

cjllanwarne: I agree with everyone. I think we should say both:
 - "docker hash lookups happen! Check out the exciting new metadata entry!"
 - "docker hash lookup now used in call caching!"

README.md
katevoss: @knoblett I think you knew this already

README.md
```diff
@@ -1778,6 +1778,35 @@ Cromwell also accepts two [workflow option](#workflow-options) related to call c
...
+
+Docker tags are a convenient way to point to a version of an image (ubuntu:14.04), or even the latest version (ubuntu:latest).
+For that purpose, tags are mutable, in a sense that the image they point to can change, while the tag name stays the same.
+While this is very convenient in some cases, using mutable, or "floating" tags in WDL affects the reproducibility of the WDL file: the same WDL using "ubuntu:latest" run now, and a year, or even a month from now will actually run with different docker images.
+This has an even bigger impact when Call Caching is turned on in Cromwell, and could lead to false positive cache hits.
```
katevoss: ToL we never define what a "false positive" means in terms of Call Caching. I'd propose something like "a false positive cache hit is when Cromwell uses the incorrect file in the process of call caching".

.travis.yml
```diff
@@ -17,7 +17,7 @@ before_cache:
   - find $HOME/.sbt -name "*.lock" -delete
 env:
   global:
-    - CENTAUR_BRANCH=develop
+    - CENTAUR_BRANCH=dockerHash
```
geoffjentry: channeling my inner @cjllanwarne - don't forget to change this

README.md
mcovarr: The phrasing "when Cromwell uses the incorrect file" makes it sound like this is a thing that happens, which it should never.  It *would* happen if we allowed floating tags to be used for caching, which is why we don't.

backend/src/main/scala/cromwell/backend/BackendLifecycleActorFactory.scala
```diff
@@ -59,6 +59,12 @@ trait BackendLifecycleActorFactory {
   lazy val fileHashingActorCount: Int = 50
 
   def fileHashingActorProps: Props = FileHashingActor.props(fileHashingFunction)
+  
+  /*
+   * Returns credentials that can be used to authenticate to a docker registry server
+   * in order to obtain a docker hash.
+   */
+  def dockerHashCredentials(initializationDataOption: Option[BackendInitializationData]): Seq[Any] = Seq.empty
```
geoffjentry: Why is this an `Any`?

core/src/main/scala/cromwell/core/callcaching/docker/DockerImageIdentifier.scala
```diff
@@ -0,0 +1,83 @@
...
+
+object DockerImageIdentifier {
+  // See https://github.com/docker-library/official-images/tree/master/library
+  private val DefaultDockerRepo = "library"
+  private val DefaultDockerTag = "latest"
+  private val TagSeparator = ":"
+  private val DigestSeparator = "@"
+  
+  private val DockerStringRegex =
+    s"""
```
geoffjentry: lol I remember "loving" this the last time :)

core/src/main/scala/cromwell/core/callcaching/docker/registryv2/DockerRegistryV2AbstractFlow.scala
```diff
@@ -0,0 +1,285 @@
...
+
+  /**
+    * (e.g auth.docker.io)
+    */
+  protected def authorizationServerHostName: String
+
+  /**
+    * Builds the list of headers for the token request
+    */
+  protected def buildTokenRequestHeaders(dockerHashContext: DockerHashContext): Seq[HttpHeader]
```
geoffjentry: Is it reasonable to expect implementers to use different `Seq` subclasses in the return type here? Or is it likely to be the case that what's smart (or smart-enough) for one is going to be true for them all? If the latter, let's make this concrete.

geoffjentry: I apologize if the answer is "use your eyes, multiple impls are *already* doing different things, jackass!"

core/src/main/scala/cromwell/core/callcaching/docker/registryv2/flows/HttpFlowWithRetry.scala
geoffjentry: I'm betting it looks less odd when not in the github split screen

geoffjentry:
To the extent that I understand streams, :+1:

Two followups:

1. based on your answer to my `Any` question I might gripe, but I suspect I'll manage to deal with it
2. Can you (if you haven't already) try running this with a bunch of workflows and/or some decently size scatters to make sure it doesn't suck under realistic workloads?

backend/src/main/scala/cromwell/backend/BackendLifecycleActorFactory.scala
Horneth: It's `any` kind of credential object that the backend wants to pass along so the docker actor is able to make a proper request. It could be a generic `T` but the point is that each Docker flow can match against it and use it if they can. (Dockerhub flow recognizes `DockerCredentials`, GCR flow recognizes google `Credentials` etc...)
If there is a way to make it generic while still being able to pattern match it and avoid type erasure that would definitely be better but I couldn't find one.

geoffjentry: I put this in the category of "I think we *could* but I neither understand the problem well enough nor the solution space well enough to warrant bothering you about it"

CHANGELOG.md
```diff
@@ -21,6 +21,15 @@
...
 resulted in strings prepended with `file:///path/to/file`. Now absolute file paths will not contain the uri scheme.
+* **Call caching: Cromwell will no longer try to use the cache for WDL tasks that contain a floating docker tag.** 
+  Call caching will still behave the same for tasks having a docker image with a specific hash.
+  See https://github.com/broadinstitute/cromwell#call-caching-docker-tags for more details. 
+* Added docker hash lookup. Cromwell will try to lookup the hash for a docker image with a floating tag, and use that hash when executing the job.
+  This will be reflected in the metadata where the docker runtime attribute will contains the hash that was used.
+  If Cromwell is unable to lookup the hash hash, the job will be run with the original user defined floating tag.
+  Cromwell is currently able to lookup public and private docker hashes for images on Docker Hub and Google Container Engine for job running on the JES backend.
+  For other backends, cromwell is able to lookup public docker hashes for Docker Hub and Google Container Engine.
+  See https://github.com/broadinstitute/cromwell#call-caching-docker-tags for more details.
```
Horneth: @katevoss I updated the changelog and readme entries for this. Please let me know if it needs updates / modifications :)

README.md
```diff
@@ -1778,6 +1778,42 @@ Cromwell also accepts two [workflow option](#workflow-options) related to call c
...
+         * the floating tag otherwise.
+         
+Docker registry and access levels supported by Cromwell for docker hash lookup:
+
+|       |       DockerHub    ||       GCR       ||
+|:-----:|:---------:|:-------:|:------:|:-------:|
+|       |   Public  | Private | Public | Private |
+|  JES  |     X     |    X    |    X   |    X    |
+| Other |     X     |         |    X   |         |
+
```
Horneth: @katevoss also here

core/src/main/scala/cromwell/core/callcaching/docker/DockerImageIdentifier.scala
Horneth: Yep stole it from @mcovarr I believe 😄

cjllanwarne:
ToL: If a task has a floating docker tag, the allowResultReuse field in metadata should be `false` not the current hardcoded `true`

PR #1975, opened 2017-02-10
Swagger fix for wdlDependencies description. Closes #1971

### LINKED ISSUE #1971: wdlDependencies should state that the upload format is a zip in swagger UI
Current copy simply says, "wdlDependencies" ... proposed change to "wdlDependencies (zip)"

Since uploading a single wdl file in this box will cause an "invalid zip file error", but does not specify that the error was in the wdlDependencies field.

cjllanwarne:


mcovarr:
oopsie :+1:

Horneth:
👍

PR #1977, opened 2017-02-11
Encapsulate `ExecutionStore`, optimize a bit.

mcovarr:
The representation of the `ExecutionStore` is now encapsulated in a `private val` where nosy clients can't just fiddle with its innards.  This allows for the possibility of an optimized representation better aligned with how the `ExecutionStore` is actually used.

I made a few optimizations in the current `Map` representation because the opportunities were just staring me in the face, but I doubt they'd have much of a noticeable benefit IRL. 😛

### DIFF
--- a/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
@@ -19,24 +19,57 @@ object ExecutionStore {
       case scatter: Scatter => Option(ScatterKey(scatter))
       case conditional: If => Option(ConditionalKey(conditional, None))
       case declaration: Declaration => Option(DeclarationKey(declaration, None, workflowCoercedInputs))
-      case _ => None // Ifs will need to be added here when supported
+      case _ => None
     }
 
     new ExecutionStore(keys.flatten.map(_ -> NotStarted).toMap)
   }
 }
 
-case class ExecutionStore(store: Map[JobKey, ExecutionStatus]) {
+case class ExecutionStore(private val store: Map[JobKey, ExecutionStatus]) {
 
-  override def toString = store.map { case (j, s) => s"$j -> $s" } mkString(System.lineSeparator())
+  // This is hugely ripe for optimization, if it becomes a bottleneck
+  def isBypassedConditional(jobKey: JobKey, conditional: If): Boolean = {
+    store.exists {
+      case (key: ConditionalKey, status) =>
+        key.scope.fullyQualifiedName.equals(conditional.fullyQualifiedName) &&
+          key.index.equals(jobKey.index) &&
+          status.equals(ExecutionStatus.Bypassed)
+      case _ => false
+    }
+  }
+
+  def hasActiveJob: Boolean = {
+    def upstreamFailed(scope: Scope): Boolean = scope match {
+      case node: GraphNode => node.upstreamAncestry exists hasFailedScope
+    }
+
+    store exists {
+      case (jobKey, jobStatus) => (jobStatus == NotStarted && !upstreamFailed(jobKey.scope)) || jobStatus == QueuedInCromwell || jobStatus == Starting || jobStatus == Running
+    }
+  }
+
+  def jobStatus(jobKey: JobKey): Option[ExecutionStatus] = store.get(jobKey)
+
+  def unstartedJobs: List[BackendJobDescriptorKey] = {
+    store.collect({case (k: BackendJobDescriptorKey, status) if status != ExecutionStatus.NotStarted => k }).toList
+  }
+
+  private def hasFailedScope(s: GraphNode): Boolean = {
+    store.exists({ case (key, status) => status == Failed && key.scope == s })
+  }
+
+  def hasFailedJob: Boolean = {
+    store.values.exists(_ == ExecutionStatus.Failed)
+  }
+
+  override def toString = store.map { case (j, s) => s"$j -> $s" } mkString System.lineSeparator()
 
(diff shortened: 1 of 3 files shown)


Horneth:
👍 
Just totally ToL for more optimization later, maybe if instead of having like now a `Map[JobKey, Status]`,
what we really want is more something like a `Map[Status, List[JobKey]]` ?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
```diff
@@ -19,44 +19,76 @@ object ExecutionStore {
       case scatter: Scatter => Option(ScatterKey(scatter))
       case conditional: If => Option(ConditionalKey(conditional, None))
       case declaration: Declaration => Option(DeclarationKey(declaration, None, workflowCoercedInputs))
-      case _ => None // Ifs will need to be added here when supported
+      case _ => None
```
cjllanwarne: 😳

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
```diff
@@ -19,44 +19,76 @@ object ExecutionStore {
...
 
     new ExecutionStore(keys.flatten.map(_ -> NotStarted).toMap)
   }
 }
 
-case class ExecutionStore(store: Map[JobKey, ExecutionStatus]) {
+case class ExecutionStore(private val store: Map[JobKey, ExecutionStatus]) {
 
-  override def toString = store.map { case (j, s) => s"$j -> $s" } mkString(System.lineSeparator())
+  // This is hugely ripe for optimization, if it becomes a bottleneck
```
cjllanwarne: I'm glad that you're going to officially take ownership of this comment in the git blame!

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
```diff
@@ -19,44 +19,76 @@ object ExecutionStore {
...
+    case (k: DynamicDeclarationKey, v) if v.isDoneOrBypassed && k.scope == key.scope && k.isShard => k
   }
 
   // Just used to decide whether a collector can be run. In case the shard entries haven't been populated into the
   // execution store yet.
   case class TempJobKey(scope: Scope with GraphNode, index: Option[Int]) extends JobKey {
     // If these are ever used, we've done something wrong...
-    override def attempt: Int = ???
-    override def tag: String = ???
+    override def attempt: Int = throw new NotImplementedError("We've done something wrong.")
```
cjllanwarne: Good call. I deliberately chose not to implement these.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -175,7 +175,7 @@ case class WorkflowExecutionActor(workflowDescriptor: EngineWorkflowDescriptor,
     // Both of these Should Never Happen (tm), assuming the state data is set correctly on EJEA creation.
     // If they do, it's a big programmer error and the workflow execution fails.
     val jobKey = stateData.engineCallExecutionActors.getOrElse(actorRef, throw new RuntimeException("Programmer Error: An EJEA has terminated but was not assigned a jobKey"))
-    val jobStatus = stateData.executionStore.store.getOrElse(jobKey, throw new RuntimeException("Programmer Error: An EJEA representing a jobKey which this workflow is not running has sent up a terminated message."))
+    val jobStatus = stateData.executionStore.jobStatus(jobKey).getOrElse(throw new RuntimeException("Programmer Error: An EJEA representing a jobKey which this workflow is not running has sent up a terminated message."))
```
cjllanwarne: ToL: I wish these were `Try`s rather than unchecked exceptions

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActorData.scala
```diff
@@ -78,32 +77,16 @@ case class WorkflowExecutionActorData(workflowDescriptor: EngineWorkflowDescript
...
-      }
-    }
-    // activeJobs is the subset of the executionStore that are either running or will run in the future.
-    val activeJobs = executionStore.store.toList filter {
-      case (jobKey, jobStatus) => (jobStatus == NotStarted && upstreamFailed(jobKey.scope).isEmpty) || jobStatus == QueuedInCromwell || jobStatus == Starting || jobStatus == Running
-    }
-
-    activeJobs match {
-      case jobs if jobs.isEmpty && hasFailedJob => Option(Failed)
-      case jobs if jobs.isEmpty && !hasFailedJob => Option(Done)
```
cjllanwarne: So you removed the `isEmpty` check?

cjllanwarne:
My baseline for these kinds of refactor is "does it pass our test suites"... so 👍 but only once you've got all that sorted out

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
mcovarr: Perhaps I shall optimise the spelling to clarify the true authorship.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActorData.scala
mcovarr: Below it's been turned into a "has failed job" check.

PR #1984, opened 2017-02-14
Allow for relative symlink localization

Horneth:


core/src/main/scala/cromwell/core/path/BetterFileMethods.scala
```diff
@@ -270,7 +270,14 @@ trait BetterFileMethods {
...
 
-  final def symbolicLink: Option[Path] = betterFile.symbolicLink.map(newPath)
+  // betterFile.symbolicLink calls Files.readSymbolicLink, but then implicitly converts the java.nio.Path returned to a better.File
+  // which calls toAbsolutePath. Consequently, if the path was relative, the current directory is used to make it absolute.
+  // This is not the desired behaviour to be able to follow relative symbolic links, so bypass better files method and directly use the java one.
+  final def symbolicLink: Option[Path] = {
+    if (betterFile.isSymbolicLink) {
+      Option(newPath(Files.readSymbolicLink(betterFile.path)))
+    } else None
+  }
```
kshakir: Don't change the `BetterFilesMethods`. The names and implementations are intentionally one-to-one pass-throughs to existing behavior.

If you wish to improve on better files, add whatever you need to `EvenBetterPathMethods`. For example:

```scala
@tailrec
final def followSymbolicLinks: Path = {
  symbolicLinkRelative match {
    case Some(target) => target.followSymbolicLinks
    case None => this
  }
}

final def symbolicLinkRelative: Option[Path] = {
  if (isSymbolicLink) {
    Option(newPath(java.nio.file.Files.readSymbolicLink(nioPathPrivate)))
  } else {
    None
  }
}
```

Whenever needed within `EvenBetterPathMethods`, I'd also recommend using fully qualified names for `java.nio` as much as possible. Nearby within the package lay imports with similar names, specifically the `Obsolete.Files`, and one doesn't want to accidentally use those.

kshakir:
Good concept. As long as the implementation of the pass-throughs aren't changed 👍

mcovarr:
:+1:

Issue #1989, opened 2017-02-15
Call Caching Task Name Oversensitivity


cjllanwarne:
Tasks cannot call cache if their names are different (unless the tasks have no inputs!)

The following tasks cannot call cache:

```
task foo {
  Int i
  command { echo ${i} }
}
```

```
task bar {
  Int i
  command { echo ${i} }
}
```

The call caching simpletons shouldn't include the call name in the input hash keys:

```
+----------------------------+-----------------------------------------+----------------------------------+-----------------------+
| CALL_CACHING_HASH_ENTRY_ID | HASH_KEY                                | HASH_VALUE                       | CALL_CACHING_ENTRY_ID |
+----------------------------+-----------------------------------------+----------------------------------+-----------------------+
....
|                        137 | input: File foo.i         | 778138a97b315ec95f374425255f8b9e |                    12 |
|                        138 | input: File bar.i         | 778138a97b315ec95f374425255f8b9e |                    13 |
```

PR #1996, opened 2017-02-16
Bring sanity to the Call Caching metadata. Closes #1994

### LINKED ISSUE #1994: Split call cache metadata asunder
An additional call caching metadatum that just says whether it was a hit or miss. The existing `Call caching read result` can stay as it is, or we could standardise the name a bit?

e.g. 
```
  "cacheHit": true
  "cacheResult": "Cache Hit: b324a594-6d46-47d1-b641-7a0dbf4b21c2:call_cache_capoeira_jes.make_files:-1"
```

cjllanwarne:
Notes to self:

 - [x] Rebase to remove No call names needed in CC when #1989 merges
 - [x]  Switch travis centaur back before merging

### DIFF
--- a/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
@@ -21,6 +21,7 @@ import cromwell.engine.workflow.tokens.JobExecutionTokenDispenserActor.{JobExecu
 import cromwell.jobstore.JobStoreActor._
 import cromwell.jobstore.{Pending => _, _}
 import cromwell.services.SingletonServicesStore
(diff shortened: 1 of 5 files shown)


engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -464,6 +470,8 @@ object EngineJobExecutionActor {
   sealed trait EngineJobExecutionActorCommand
   case object Execute extends EngineJobExecutionActorCommand
 
+  val CacheMetadataKeyPrefix = "callCaching" + MetadataKey.KeySeparator
```
Horneth: can `"callCaching"` be a value in `CallMetadataKeys` ?

Horneth:
👍

mcovarr:
:+1:

PR #2003, opened 2017-02-17
Throw if cache hit copying fails on SFS

Horneth:


supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystemCacheHitCopyingActor.scala
```diff
@@ -7,7 +7,6 @@ class SharedFileSystemCacheHitCopyingActor(standardParams: StandardCacheHitCopyi
   extends StandardCacheHitCopyingActor(standardParams) with SharedFileSystemJobCachingActorHelper {
   override protected def duplicate(source: Path, destination: Path): Unit = {
     // -Ywarn-value-discard
```
mcovarr: This comment can be removed now

mcovarr:
Good catch!  :+1:

ruchim:
:+1:

PR #2004, opened 2017-02-17
Don't return invalidated cache entries

Horneth:


mcovarr:
:+1:

ruchim:
:+1:

cjllanwarne:
ToL: We should also send an override of the allowResultReuse metadata with `false`

PR #2006, opened 2017-02-17
fail file hashing if the file does not exist

Horneth:


ruchim:
:+1:

geoffjentry:
👍 

Out of curiosity, what caused this to turn up?

Horneth:
I was testing some call caching behaviors. Specifically I tried to call cache on local backend with the "hash the file path" method which would hash a file path even though the file doesn't exist. The job would then try to find a cache hit for this path even though the file doesn't exist

andy7i:
H

PR #2007, opened 2017-02-17
Disambiguate actor name by including source cache entry ID.

mcovarr:


geoffjentry:
👍

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -373,8 +373,8 @@ class EngineJobExecutionActor(replyTo: ActorRef,
     s"$workflowIdForLogging-BackendJobExecutionActor-$jobTag"
   }
 
-  private def buildCacheHitCopyingActorName(jobDescriptor: BackendJobDescriptor) = {
-    s"$workflowIdForLogging-BackendCacheHitCopyingActor-$jobTag"
+  private def buildCacheHitCopyingActorName(jobDescriptor: BackendJobDescriptor, cacheResultId: CallCachingEntryId) = {
+    s"$workflowIdForLogging-BackendCacheHitCopyingActor-$jobTag-${cacheResultId.id}"
```
cjllanwarne: ToL You could achieve the same result, without all this carrying IDs around everywhere, by just having an atomic integer which you increment every time...

cjllanwarne:
👍, if the EJEA tests all say 👍

PR #2013, opened 2017-02-21
Evaluate imports file content eagerly. Fixes "workflowName" centaur f…

Horneth:
…ailures.

### DIFF
--- a/engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
@@ -335,11 +335,11 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
   }
 
   private def validateNamespaceWithImports(w: WorkflowSourceFilesWithDependenciesZip): ErrorOr[WdlNamespaceWithWorkflow] = {
-    def getMetadatae(importsDir: Path, prefix: String = ""): Seq[(String, Path)] = {
-      importsDir.children.toSeq flatMap {
+    def getMetadatae(importsDir: Path, prefix: String = ""): List[(String, Path)] = {
+      importsDir.children.toList flatMap {
         case f: Path if f.isDirectory => getMetadatae(f, prefix + f.name + "/")
-        case f: Path if f.name.endsWith(".wdl") => Seq((prefix + f.name, f))
-        case _ => Seq.empty
+        case f: Path if f.name.endsWith(".wdl") => List((prefix + f.name, f))
+        case _ => List.empty
       }
     }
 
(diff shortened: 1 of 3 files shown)


engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -335,11 +335,11 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
   }
 
   private def validateNamespaceWithImports(w: WorkflowSourceFilesWithDependenciesZip): ErrorOr[WdlNamespaceWithWorkflow] = {
-    def getMetadatae(importsDir: Path, prefix: String = ""): Seq[(String, Path)] = {
-      importsDir.children.toSeq flatMap {
```
Horneth: @geoffjentry you should like this one: this `toSeq` is (believed to be) the reason for the `workflowName` situation

services/src/main/scala/cromwell/services/metadata/impl/MetadataServiceActor.scala
```diff
@@ -28,7 +29,14 @@ object MetadataServiceActor {
 
 case class MetadataServiceActor(serviceConfig: Config, globalConfig: Config)
   extends Actor with ActorLogging with MetadataDatabaseAccess with SingletonServicesStore {
-
+  
+  override val supervisorStrategy = new OneForOneStrategy()(SupervisorStrategy.defaultStrategy.decider) {
+    override def logFailure(context: ActorContext, child: ActorRef, cause: Throwable, decision: Directive) = {
+      val childName = if (child == readActor) "Read" else "Write"
+      log.error(s"The $childName Metadata Actor died unexpectedly. Restarting it...", cause)
```
mcovarr: "..., possibly taking your precious metadata with it."  😦

mcovarr:
:+1:

engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -335,11 +335,11 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
...
+    def getMetadatae(importsDir: Path, prefix: String = ""): List[(String, Path)] = {
```
geoffjentry: Since this ultimately winds up being used in the `WriteMetadataActor` which is converting things to a `Vector`, we can save a `toList` and then a `toVector` by just calling `toVector` here and returning `Vector`

services/src/main/scala/cromwell/services/metadata/impl/MetadataServiceActor.scala
geoffjentry: I think this is fair, the default batch size is `1` which effectively turns off batching, and there's a warning in the conf file that turning that number higher increases performance but brings about some amount of risk.

That said for the `WriteMetadataActor` we should probably identify likely problem points and catch them in the whole "error vs failure" angle.

geoffjentry:
👍

services/src/main/scala/cromwell/services/metadata/impl/MetadataServiceActor.scala
Horneth: to be clear we can still lose messages if the actor dies even if the batch size is 1. Messages are being accumulated while the actor waits for the response from the DB, regardless of the batch size.

geoffjentry: Good point.

geoffjentry: One thought is that most (all?) failures in the WriteMetadataActor could be handled w/ a resume instead of a restart. That's dangerous in the case that our accumulated state *is* the reason we're F'd up but that seems a lot less likely than something transient that causes us to lose messages

geoffjentry: Although thinking about what i just said more, it's basically the same as my previous suggestion of judiciously adding some try's here and there.

Horneth: I don't think I fully agree, all this actor does is taking messages, and calling a DB method. The DB method is encapsulated in a Future which makes error handling safe. Which means any other unexpected error will come from one of the messages in the data, so resuming will almost certainly cause the actor to perpetually fail. I don't see much room for transient errors not related to messages.

geoffjentry: What I mean is this. Whatever reason caused the WMA to crash, it certainly wasn't the accumulated queue of events which have built up after it decided to start writing to DB, so there's no reason we should be losing those. Finding a way to preserve that queue would be helpful (although it still wouldn't protect in the case that the DB was down and everything fails, but that's a separate topic).

PR #2015, opened 2017-02-22
Better NPE handling during terminal JES parsing.

kshakir:


supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/Run.scala
```diff
@@ -117,22 +117,31 @@ object Run {
...
-      Option(op.getError) match {
-        case None => Success(eventList, machineType, zone, instanceName)
-        case Some(error) => Failed(error.getCode, Option(error.getMessage), eventList, machineType, zone, instanceName)
+    try {
+      if (op.getDone) {
+        lazy val eventList = getEventList(op)
+        lazy val computeEngineOption = for {
+          runtimeMetadata <- op.getMetadata.asScala.get("runtimeMetadata")
+          computeEngine <- runtimeMetadata.asInstanceOf[GArrayMap[String, Object]].asScala.get("computeEngine")
+        } yield computeEngine.asInstanceOf[GArrayMap[String, String]].asScala
```
cjllanwarne: ToL: I'm 99% sure I'd prefer these to be a placeholder like "Undeterminable" in the metadata, rather than failing a job just because we couldn't find the machine type... PO decision?

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/Run.scala
```diff
@@ -117,22 +117,31 @@ object Run {
...
+      } else {
+        Initializing
       }
-    } else if (op.hasStarted) {
-      Running
-    } else {
-      Initializing
+    } catch {
+      case npe: NullPointerException =>
+        throw new RuntimeException(s"Caught NPE while processing operation ${op.getName}: $op", npe)
```
cjllanwarne: As far as I can tell, the only other way to get a NPE here (other than the handled section above) is if op is null... in which case using it to fill in the NPE details is going to cause another NPE!

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/Run.scala
cjllanwarne: Wait, that's exactly what you are doing! Ignore this comment!

cjllanwarne:
Apologies for missing the point! 👍

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/Run.scala
kshakir: `op.getDone` can throw a NPE, as it uses a scala implict conversion to convert from `java.lang.Boolean` to a `boolean`. `getEventList` could also throw an error when trying to call `getMetadata`. There are test specs for both.

Also you're correct that `op` itself could be also be `null`. I'll add a check and test for that too.

mcovarr:
:+1:

Issue #1126, opened 2016-07-05
Abort Local with docker kills the script but not the docker container


Horneth:
When running a task on docker locally, if the task is aborted, the script that feeds the wdl command into the docker container will be destroyed - but not the docker itself. 
Before abort:

```
wm0e9-683:cromwell tjeandet$ ps -ef | grep c72df249-ccf5-460b-8654-9606a655c669
1250915218 11002 10990   0  6:13PM ??         0:00.00 /bin/bash -c cat cromwell-executions/w/c72df249-ccf5-460b-8654-9606a655c669/call-t/script | docker run --rm -v /Users/tjeandet/Codebase/cromwell/cromwell-executions/w/c72df249-ccf5-460b-8654-9606a655c669/call-t:/root/w/c72df249-ccf5-460b-8654-9606a655c669/call-t -i ubuntu:latest /bin/bash <&0
1250915218 11004 11002   0  6:13PM ??         0:00.15 docker run --rm -v /Users/tjeandet/Codebase/cromwell/cromwell-executions/w/c72df249-ccf5-460b-8654-9606a655c669/call-t:/root/w/c72df249-ccf5-460b-8654-9606a655c669/call-t -i ubuntu:latest /bin/bash
1250915218 11015  2783   0  6:13PM ttys000    0:00.00 grep c72df249-ccf5-460b-8654-9606a655c669
```

After:

```
wm0e9-683:cromwell tjeandet$ ps -ef | grep c72df249-ccf5-460b-8654-9606a655c669
1250915218 11004     1   0  6:13PM ??         0:00.15 docker run --rm -v /Users/tjeandet/Codebase/cromwell/cromwell-executions/w/c72df249-ccf5-460b-8654-9606a655c669/call-t:/root/w/c72df249-ccf5-460b-8654-9606a655c669/call-t -i ubuntu:latest /bin/bash
1250915218 11021  2783   0  6:14PM ttys000    0:00.00 grep c72df249-ccf5-460b-8654-9606a655c669
```

kshakir:
More info on the current issue:
- The local backend, running on the Config backend, supports running commands with or without docker.
- However the Config backend only allows `run-in-background` to be set for _both_ with and without docker.
- When `run-in-background` is set to true, the process id of the backgrounded process is stored for later killing.
[...]

PR #2028, opened 2017-02-26
Config backend wasn't dockerizing script paths.

kshakir:


kshakir:
Fixes the workaround syntax in #1126, but doesn't completely repair the issue.

core/src/main/resources/reference.conf
```diff
@@ -179,7 +179,14 @@ backend {
         String? docker_user
         """
         submit = "/bin/bash ${script}"
-        submit-docker = """docker run --rm ${"--user " + docker_user} -v ${cwd}:${docker_cwd} -i ${docker} /bin/bash ${docker_cwd}/execution/script"""
+        submit-docker = """
+        docker run \
+          --rm -i \
+          ${"--user " + docker_user} \
+          -v ${cwd}:${docker_cwd} \
+          ${docker} \
```
mcovarr: ToL I kinda wish this variable name had the word "image" in it for clarity, but unless we support a new and deprecated name for this variable that would be a breaking change 🙁

supportedBackends/sfs/src/main/scala/cromwell/backend/impl/sfs/config/ConfigAsyncJobExecutionActor.scala
```diff
@@ -64,23 +64,28 @@ sealed trait ConfigAsyncJobExecutionActor extends SharedFileSystemAsyncJobExecut
...
+      CwdInput -> WdlString(jobPaths.callRoot.pathAsString)
     )
   }
 
   /**
-    * Extra arguments if this is a submit-docker command, or Map.empty.
+    * The inputs that are not specified by the config, that will be passed into a command for either submit or
+    * submit-docker.
     */
   private lazy val dockerInputs: WorkflowCoercedInputs = {
```
mcovarr: The method name doesn't seem quite right anymore.

mcovarr:
:+1:

cjllanwarne:
ToL: Is this centaurable? Or can we reconfigure centaur's Local tests so that this is verified?

cjllanwarne:
👍 not sure how docker was working at all in SFS-land before this!

kshakir:
The `script` variable was busted, but docker was working in SFS land because, instead of `${script}` the local backend was using `${cwd}/execution/script` when launching docker jobs.

supportedBackends/sfs/src/main/scala/cromwell/backend/impl/sfs/config/ConfigAsyncJobExecutionActor.scala
kshakir: Naming... my nemesis.

- `dockerizedInputs`
- `localOrDockerInputs`
- `submitOrSubmitDockerInputs`
- etc.?

kshakir:
Re "centaurable"-- I'm confident by updating the local backend to use `${script}` that centaur is actually exercising a third of the variables updated in patch. We could also change centaur to completely run on a new backend that also uses the `${out}` and `${err}` paths, as outlined in issue #1126. But because of that issue, this new centaur test would require a separate, new backend definition in `local_centaur.conf`.

FYI: I still don't have a full fix for #1126, but this is a step in that direction. When I looked at the reference.conf, I noticed the local backend wasn't using `${script}`, wanted to know why, and discovered this small issue. This patch will also make the workaround in that ticket work as expected, instead of pointing to paths outside the docker container like `/Users/kshakir/<path>` now pointing paths inside like `/cromwell_root/<path>`.

core/src/main/resources/reference.conf
kshakir: Not touching that in this PR, but open to one filing a ticket to revisit!

supportedBackends/sfs/src/main/scala/cromwell/backend/impl/sfs/config/ConfigAsyncJobExecutionActor.scala
kshakir: Ruchi hath selected `dockerizedInputs`. Updating.

Issue #2029, opened 2017-02-26
WEA transitions to failed state and leaves jobs running


kshakir:
There are two general cases where the `WorkflowExecutionActor` switches the workflow to a failed state:
- [`handleRetryableFailure()`](https://github.com/broadinstitute/cromwell/blob/f64c16e62c84b66ddba14705bede5e6fde8376b0/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala#L292-L306)
- [`handleExecutionFailure()`](https://github.com/broadinstitute/cromwell/blob/f64c16e62c84b66ddba14705bede5e6fde8376b0/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala#L240-L255)

In _both_ cases, the method:
- is triggered by the failure of an individual job, but
- fails the entire workflow, and
- doesn't send signals to other jobs to stop

Thus other jobs stay running, while the workflow gets deregistered from the system. Because the workflow manager can no longer delegate aborting the running jobs, this issue may also be related to #1414 and #1504.

katevoss:
@kshakir did this change at all with the retries work earlier this year?

Horneth:
Fixed by https://github.com/broadinstitute/cromwell/pull/2808, although other jobs are not stopped when a job fails, they're left as is and are being monitored until they reach completion.

PR #2030, opened 2017-02-26
Message, not throw, failures in WEA.startRunnableScopes.

kshakir:
Other syntax fixes.

### DIFF
--- a/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
@@ -13,7 +13,7 @@ import cromwell.core.WorkflowOptions.WorkflowFailureMode
 import cromwell.core._
 import cromwell.core.logging.WorkflowLogging
 import cromwell.engine.backend.{BackendSingletonCollection, CromwellBackends}
-import cromwell.engine.workflow.lifecycle.execution.WorkflowExecutionActor.{apply => _, _}
+import cromwell.engine.workflow.lifecycle.execution.WorkflowExecutionActor._
 import cromwell.engine.workflow.lifecycle.{EngineLifecycleActorAbortCommand, EngineLifecycleActorAbortedResponse}
 import cromwell.engine.{ContinueWhilePossible, EngineWorkflowDescriptor}
 import cromwell.util.StopAndLogSupervisor
@@ -121,7 +121,7 @@ case class WorkflowExecutionActor(workflowDescriptor: EngineWorkflowDescriptor,
       pushFailedCallMetadata(jobKey, returnCode, reason, retryableFailure = false)
       handleNonRetryableFailure(stateData, jobKey, reason, Map.empty)
       // Job Retryable
-    case Event(JobFailedRetryableResponse(jobKey, reason, returnCode), stateData) =>
+    case Event(JobFailedRetryableResponse(jobKey, reason, returnCode), _) =>
       pushFailedCallMetadata(jobKey, None, reason, retryableFailure = true)
       handleRetryableFailure(jobKey, reason, returnCode)
       // Sub Workflow - sub workflow failures are always non retryable
@@ -142,7 +142,7 @@ case class WorkflowExecutionActor(workflowDescriptor: EngineWorkflowDescriptor,
     case Event(SubWorkflowSucceededResponse(subKey, executedKeys, _), stateData) =>
       pushAbortedCallMetadata(subKey)
       handleCallAborted(stateData, subKey, executedKeys)
-    case Event(JobSucceededResponse(jobKey, returnCode, callOutputs, _, _), stateData) =>
+    case Event(JobSucceededResponse(jobKey, _, _, _, _), stateData) =>
       pushAbortedCallMetadata(jobKey)
       handleCallAborted(stateData, jobKey, Map.empty)
   }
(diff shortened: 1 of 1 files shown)


kshakir:
Centaur PR https://github.com/broadinstitute/centaur/pull/164

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -355,34 +359,36 @@ case class WorkflowExecutionActor(workflowDescriptor: EngineWorkflowDescriptor,
...
+        case k: DynamicDeclarationKey => processRunnableDynamicDeclaration(k, data)
+        case k => Failure(new UnsupportedOperationException(s"Unknown entry in execution store: ${k.tag}"))
+      }).flatten
+    } map {
+      case (_, Success(value)) => Success(value)
+      case (scope, Failure(throwable)) =>
+        self ! JobInitializationFailed(scope, throwable)
+        Failure(throwable)
+    } collect {
+      case Success(value) => value
```
Horneth: TOL: Doesn't that mean that we're potentially going to start new jobs even though the workflow is doomed to fail if one of the `JobInitializationFailed` message was sent ?

mcovarr: Great catch!  It looks like we still want the `sequence` behavior here.

.travis.yml
```diff
@@ -19,7 +19,7 @@ before_install:
   - openssl aes-256-cbc -K "$encrypted_5ebd3ff04788_key" -iv "$encrypted_5ebd3ff04788_iv" -in src/bin/travis/resources/jesConf.tar.enc -out jesConf.tar -d
 env:
   global:
-    - CENTAUR_BRANCH=develop
+    - CENTAUR_BRANCH=ks_catch_runnable_scopes
```
cjllanwarne: Remember to revert before merging!

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -259,30 +259,34 @@ case class WorkflowExecutionActor(workflowDescriptor: EngineWorkflowDescriptor,
...
-      data.expressionLanguageFunctions,
-      data.outputStore.fetchNodeOutputEntries
-    ) map { workflowOutputs =>
-       // For logging and metadata
-       val workflowScopeOutputs = workflowOutputs map {
-         case (output, value) => output.locallyQualifiedName(workflowDescriptor.workflow) -> value
-       }
-       workflowLogger.info(
-         s"""Workflow ${workflowDescriptor.workflow.unqualifiedName} complete. Final Outputs:
+    val responseAndFinalStateTry: Try[(WorkflowExecutionActorResponse, WorkflowExecutionActorState)] =
```
cjllanwarne: Channelling the honourable @geoffjentry, can we ADT this?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -259,30 +259,34 @@ case class WorkflowExecutionActor(workflowDescriptor: EngineWorkflowDescriptor,
...
+          case (output, value) => output.unqualifiedName -> JobOutput(value)
+        }
+        (WorkflowExecutionSucceededResponse(data.jobExecutionMap, unqualifiedWorkflowOutputs),
+          WorkflowExecutionSuccessfulState)
+      }
+
+    val (response, finalState) = responseAndFinalStateTry recover {
+      case throwable =>
+        (WorkflowExecutionFailedResponse(data.jobExecutionMap, throwable), WorkflowExecutionFailedState)
     } get
```
cjllanwarne: ToL: I was happy right up until I saw this!

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -493,7 +499,7 @@ case class WorkflowExecutionActor(workflowDescriptor: EngineWorkflowDescriptor,
         val conditionalStatus = if (b.value) ExecutionStatus.Done else ExecutionStatus.Bypassed
         val result = WorkflowExecutionDiff(conditionalKey.populate(workflowDescriptor.knownValues) + (conditionalKey -> conditionalStatus))
         result
-      case v: WdlValue => throw new RuntimeException("'if' condition must evaluate to a boolean")
+      case _: WdlValue => throw new RuntimeException("'if' condition must evaluate to a boolean")
```
cjllanwarne: TOL it'd be nice if we could say `... Boolean but instead got ${v.wdlType.toWdlString}`

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -510,7 +516,7 @@ case class WorkflowExecutionActor(workflowDescriptor: EngineWorkflowDescriptor,
       scatterKey.scope.collection.evaluate(lookup, data.expressionLanguageFunctions) map {
         case a: WdlArray =>
           WorkflowExecutionDiff(scatterKey.populate(a.value.size, workflowDescriptor.knownValues) + (scatterKey -> ExecutionStatus.Done))
-        case v: WdlValue => throw new RuntimeException("Scatter collection must evaluate to an array")
+        case _: WdlValue => throw new RuntimeException("Scatter collection must evaluate to an array")
```
cjllanwarne: TOL Same here

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
kshakir: A couple things:

- The other jobs have _already_ started by the point we do a `Try.sequence`. It's too late to stop them from launching. The EJEAs are already off to the races.
- Even when we do detect that something went wrong in "1 of 2000" jobs that just started, what is the duty of _this_ method?

Before, the method was throwing an exception killing the whole actor. The workflow wasn't actually stopped, and was left in an infinite "running" state. But yes even-more jobs didn't start. 😉 

Elsewhere in this WEA is other code (that isn't 100% [imho](https://github.com/broadinstitute/cromwell/issues/2029)) that decides whether to fail-fast or fail-slow when a single job fails. Again imho it's _that_ code that should decide what to do about the individual job failures, which is why I chose to send out a message to self.

So, happy to tech talk this, but I currently believe this method should be messaging, and not throwing exceptions.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
kshakir: Not green == not my git-blame

mcovarr:
discussed in person :+1:

Horneth:
👍

PR #2031, opened 2017-02-26
Updates for memory validations.

kshakir:
Fixed SGE memory parameter in reference.conf.
Split `MemoryDeclarationValidation` out of `DeclarationValidation`.
Added tests for `MemoryDeclarationValidation`.
`OptionalRuntimeAttributesValidation` were not returning the defaults from the underlying `RuntimeAttributesValidation`.

### DIFF
--- a/supportedBackends/sfs/src/main/scala/cromwell/backend/impl/sfs/config/MemoryDeclarationValidation.scala
+++ b/supportedBackends/sfs/src/main/scala/cromwell/backend/impl/sfs/config/MemoryDeclarationValidation.scala
@@ -1,0 +1,115 @@
+package cromwell.backend.impl.sfs.config
+
+import cromwell.backend.MemorySize
+import cromwell.backend.impl.sfs.config.ConfigConstants._
+import cromwell.backend.validation._
+import wdl4s.expression.NoFunctions
+import wdl4s.parser.MemoryUnit
+import wdl4s.types._
+import wdl4s.values.{WdlFloat, WdlInteger, WdlOptionalValue, WdlValue}
+import wdl4s.{Declaration, NoLookup, WdlExpression}
+
+/**
+  * Maps declarations of memory in WDL runtime attributes to the commands used to submit.
+  *
+  * The wdl runtime attributes for memory specified as strings such as:
+  *
+  * {{{
+  *   runtime {
+  *     memory: "500 MB"
+  *   }
+  * }}}
+  *
+  * However, the backend configuration only supports specifying memory in amounts in Float or Int. To specify the unit
+  * of the amount, the string "memory_" is suffixed with the unit, for example "memory_mb" or "memory_gb", or even
+  * "memory_ki".
+  *
+  * This class and companion object will do the conversion. The backend configuration should use the runtime attribute
+  * "Float? memory_gb", meaning that "memory" is now an optional runtime attribute, and will be converted to GB.
+  *
+  * Just like the runtime attribute, when no units are specified the config, the default unit is bytes.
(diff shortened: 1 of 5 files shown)


geoffjentry:
👍

supportedBackends/sfs/src/main/scala/cromwell/backend/impl/sfs/config/MemoryDeclarationValidation.scala
```diff
@@ -0,0 +1,115 @@
...
+  *
+  * The wdl runtime attributes for memory specified as strings such as:
+  *
+  * {{{
+  *   runtime {
+  *     memory: "500 MB"
+  *   }
+  * }}}
+  *
+  * However, the backend configuration only supports specifying memory in amounts in Float or Int. To specify the unit
```
Horneth: Just out of curiosity, why is that ? Couldn't the config backend also support `memory: "500 MB"` ?

Horneth:
👍

supportedBackends/sfs/src/main/scala/cromwell/backend/impl/sfs/config/MemoryDeclarationValidation.scala
kshakir: ToL:

I perhaps could have supported inferred types. In our WDL, runtime attributes are duck typed (this int/float/string looks like a memory value, make it work), while inputs and outputs are strongly typed with implicit conversions (this output is explicitly defined as a WdlFile, but if you give me a WdlString, I'll convert it).

When I implemented the config backend, I went for strongly typed declarations. It was easier, as behind the scenes the config backend is creating a WDL `Task`. The inputs to the task are the backend's `config.runtime-attributes` stanza. From the user's WDL, the `runtime` attributes are passed in to `task.instantiateCommand()`, resulting in a string that may be used to launch the job.

As the generated inputs are WDL declarations, one needs to strongly type them with a WdlType. So `memory: "500 MB"` would need to actually be `String memory: "500 MB"`.

I also needed to know what units to use when passing the memory to the submit string:

```
submit = """docker run --memory-reservation="${memory}m" ${docker} ${script} ..."""
```

In this case, memory should always be in megabytes, even if the wdl says "2GB". I also wanted to know what precision to use, fixed or floating point. So I made it so the person authoring the config could specify the precision and units in one line:

- `Float memory_mb`
- `Int memory_mb`
- `Float memory_gb`
- etc.

With more designing and engineering, perhaps one could change the runtime-attributes configuration from explicitly typed to inferred, more like the WDL `runtime` section:

Instead of defining a backends runtime attributes with:

```
runtime-attributes = """
Float memory_gb = 2.0 # obviously a float
String sge_queue      # mandatory, no default
"""
```

We could use something like:

```
runtime-attributes = """
memory_gb = 2.0  # still obviously a float
sge_queue = null # not sure how to handle mandatory runtime attributes
"""
```

It's possible that something as simple as-- instead of using `Task` inputs, generate a `WdlMap` and pass that to `instantiateCommand`-- might solve the problem of static typing. I believe effectively the `runtime` section of WDL actually is a type of dynamic map also.

Horneth: Thanks for the very detailed explanation ! 
I think strong types are better anyway and we should move to that direction more than the implicit conversion one.

Another idea would be something like

```
Memory myMemory = memory(500, "MB")
```

Anyway that's far from being required here I was just curious of the why :)

PR #2032, opened 2017-02-26
Slightly better abort behavior in standard backend.

kshakir:
Stopped closing a scala `Future` over akka's `context.become`.
Flipped the default for `requestsAbortAndDiesImmediately` from `false` to `true`.
When killing a Standard backend job with rADI false, both the rc and the standard error are required.
Writing the stderr on abort for the SFS backend.
When rADI is true, sending a backend status of `"Aborted"`.
In the engine, set and store the `ExecutionStatus.Aborted`.


geoffjentry:
👍

mcovarr:
:+1:

Issue #2037, opened 2017-02-27
Inconsistent failure messages for workflows


ansingh7115:
We are trying to expose workflow failure messages to the user in Firecloud. However, the failures field seems to have an inconsistent format:

Compare the failures sections for the following:

```
{
  "workflowName": "echo_strings",
  "submittedFiles": {
    "inputs": "{...},
  "calls": {
    "echo_strings.echo_files": [{
      "preemptible": false,
      "retryableFailure": false,
      "executionStatus": "Failed",
      "stdout": "gs://fc-2d3fd356-e3be-4953-92f1-60af623e6fa5/b6b190d6-8640-4638-94cd-15f16b194f38/echo_strings/c386672d-0248-4968-9b1a-114f5f5c4706/call-echo_files/echo_files-stdout.log",
      "backendStatus": "Failed",
      "shardIndex": -1,
      "jes": {
        "endpointUrl": "https://genomics.googleapis.com/",
        "machineType": "us-central1-c/n1-standard-1",
        "googleProject": "broad-dsde-dev",
        "executionBucket": "gs://cromwell-dev/cromwell-executions",
        "zone": "us-central1-c",
        "instanceName": "ggp-3462354720519617596"
      },
      "runtimeAttributes": {...},
      "cache": {
        "allowResultReuse": true
      },
      "Effective call caching mode": "CallCachingOff",
      "inputs": {...
      },
      "failures": [{
        "message": "Task c386672d-0248-4968-9b1a-114f5f5c4706:echo_files failed: error code 5. Message: 8: Failed to pull image ubuntu:latest: \"docker --config /tmp/.docker/ pull ubuntu:latest\" failed: exit status 1: Pulling repository docker.io/library/ubuntu\nNetwork timed out while trying to connect to https://index.docker.io/v1/repositories/library/ubuntu/images. You may want to check your internet connection or if you are behind a proxy.\n"
      }],
      "jobId": "operations/EJiq_oWfKxi8-N-X4qiwhjAgw7vetLsXKg9wcm9kdWN0aW9uUXVldWU",
      "backend": "JES",
      "end": "2017-01-30T19:14:19.708Z",
      "stderr": "gs://fc-2d3fd356-e3be-4953-92f1-60af623e6fa5/b6b190d6-8640-4638-94cd-15f16b194f38/echo_strings/c386672d-0248-4968-9b1a-114f5f5c4706/call-echo_files/echo_files-stderr.log",
[...]

PR #2038, opened 2017-02-27
Flatten Failure Metadata

cjllanwarne:


geoffjentry:
Make sure @ansingh7115 takes a look at the new output if she hasn't already

services/src/test/scala/cromwell/services/metadata/MetadataServiceSpec.scala
```diff
@@ -74,4 +77,94 @@ class MetadataServiceSpec extends FlatSpec with Matchers with TableDrivenPropert
     }
   }
 
+  // For the metadata event tests!
+  val eventKeyRegex = "path\\:to\\:failures\\[([0-9]*)\\]\\:message".r
```
mcovarr: should that `*` be a `+`?

services/src/main/scala/cromwell/services/metadata/MetadataService.scala
```diff
@@ -129,10 +131,12 @@ object MetadataService {
   def throwableToMetadataEvents(metadataKey: MetadataKey, t: Throwable): List[MetadataEvent] = {
     t match {
       case aggregation: ThrowableAggregation =>
-        aggregation.errorMessages.toList map { message => MetadataEvent(metadataKey.copy(key = s"${metadataKey.key}:message"), MetadataValue(s"${aggregation.exceptionContext}: $message")) }
+        val message = List(MetadataEvent(metadataKey.copy(key = s"${metadataKey.key}[${Random.nextInt(Int.MaxValue)}]:message"), MetadataValue(aggregation.exceptionContext)))
```
mcovarr: I realize this is somewhat a refactoring of what was here previously, but the space of nonnegative `Int`s is kind of limited for a hash...

cjllanwarne:
Approved in person. Migration options TBD

services/src/test/scala/cromwell/services/metadata/MetadataServiceSpec.scala
```diff
@@ -74,4 +77,94 @@ class MetadataServiceSpec extends FlatSpec with Matchers with TableDrivenPropert
...
+      validateEvent(events(3), workflowId, innerCauseMsg),
+      validateEvent(events(4), workflowId, causeMsg2),
+      validateEvent(events(5), workflowId, causeMsg3)
+    )
+    // Make sure all IDs are unique:
+    errorIds.toSet.size should be(6)
+  }
+
+  def validateEvent(event: MetadataEvent, workflowId: WorkflowId, message: String) = event match {
+    case MetadataEvent(k, Some(MetadataValue(v, _)), _) =>
```
mcovarr: if this doesn't match at all there's no error?

services/src/test/scala/cromwell/services/metadata/MetadataServiceSpec.scala
```diff
@@ -74,4 +77,94 @@ class MetadataServiceSpec extends FlatSpec with Matchers with TableDrivenPropert
...
+      k.workflowId should be(workflowId)
+      val messageIndex = k.key match {
+        case eventKeyRegex(x) => x
+        case _ => fail("Unexpected failure key format: " + k.key)
+      }
+      v should be(message)
```
mcovarr: the `v should...` would probably make more sense under the `k.workflowId should...` and then you can return the `k.key match {...}` expression directly

mcovarr:
:+1:

services/src/main/scala/cromwell/services/metadata/MetadataService.scala
```diff
@@ -129,10 +131,12 @@ object MetadataService {
...
+        val innards = aggregation.throwables.toList flatMap { innard => throwableToMetadataEvents(metadataKey, innard) }
+        message ++ innards
       case other =>
-        val message = List(MetadataEvent(metadataKey.copy(key = s"${metadataKey.key}:message"), MetadataValue(t.getMessage)))
-        val cause = Option(t.getCause) map { cause => throwableToMetadataEvents(metadataKey.copy(key = s"${metadataKey.key}:causedBy"), cause) } getOrElse List.empty
+        val message = List(MetadataEvent(metadataKey.copy(key = s"${metadataKey.key}[${Random.nextInt(Int.MaxValue)}]:message"), MetadataValue(t.getMessage)))
+        val cause = Option(t.getCause) map { cause => throwableToMetadataEvents(metadataKey, cause) } getOrElse List.empty
```
kshakir: Injecting the `Random.nextInt` here _inside_ `throwableToMetadataEvents` on 138 separates the exception from the recursive underlying cause on 139. IMHO that makes debugging much more confusing. These are _not_ separate events, and should be tied together.

Btw, will tech talk later, but as far as I can tell we only have **one** active json schema for exceptions, the one mentioned in the [pseudo-CFG](https://github.com/broadinstitute/cromwell/issues/2037#issuecomment-282800499).

The inactive json schema that may have triggered this whole issue appears to be an artifact of the [migration](https://github.com/broadinstitute/cromwell/blob/509ff2dc2bc2208546a76307791af05816ec83f5/database/migration/src/main/scala/cromwell/database/migration/metadata/table/FailureEventTableMigration.scala#L23) [code](https://github.com/broadinstitute/cromwell/blob/509ff2dc2bc2208546a76307791af05816ec83f5/database/migration/src/main/scala/cromwell/database/migration/metadata/table/FailureEventTableMigration.scala#L47) not matching the active code, deleted in red above.

I'm fine upgrading and migrating our current the json [schema](http://json-schema.org/) to something non-recursive, but would propose something along the lines of:

```json
{
  "$ref": "#/definitions/failures",
  "definitions": {
    "failures": {
      "type": "array",
      "items": {
        "$ref": "#/definitions/failure"
      }
    },
    "failure": {
      "required": ["message"],
      "properties": {
        "message": {
          "type": "string"
        },
        "causedBy": {
          "type": "array",
          "items": {
            "type": "string"
          }
        }
      }
    }
  }
}
```

This schema would change the recursive:

```json
{
  "failures": [
    {
      "message": "Failed to upload authentication file",
      "causedBy": {
        "message": "Error getting access token for service account: ",
        "causedBy": {
          "message": "connect timed out"
        }
      }
    }
  ]
}
```

to the more condensed:

```json
{
  "failures": [
    {
      "message": "Failed to upload authentication file",
      "causedBy": [
        "Error getting access token for service account: ",
        "connect timed out"
      ]
    }
  ]
}
```

and would still allow one to figure out what exceptions are connected, or not:

```json
{
  "failures": [
    {
      "message": "Internal unconnected exception with no cause by"
    },
    {
      "message": "A separate I/O exception that failed to upload authentication file",
      "causedBy": [
        "Error getting access token for service account: ",
        "connect timed out"
      ]
    }
  ]
}
```

Also fine with re-migrating the old data to bring it up to a current json schema implementation.

cjllanwarne: @helgridly @ansingh7115 would you like to join our tech talk, at the appropriate time?

services/src/test/scala/cromwell/services/metadata/MetadataServiceSpec.scala
cjllanwarne: Yeah... I mean it's just a case of where an exception is thrown in a test, but... yeah.

services/src/main/scala/cromwell/services/metadata/MetadataService.scala
helgridly: Yes. Short version: this seems fine. For now we're just looking at the message field of each object anyway, so you can throw whatever other <strike>junk</strike> useful information in there you'd like.

cjllanwarne:
Closing this PR. The branch may or may not be repurposed.

PR #2040, opened 2017-02-28
* make sure dir that TMPDIR is set to, exists. Fixes #2035

### LINKED ISSUE #2035: exporting non-existent TMPDIR in commandScriptContents considered harmful
https://github.com/broadinstitute/cromwell/blob/25c6573bb9342a859fee735f4190b6b6d9f2dec4/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala#L181

The environment variable $TMPDIR set to a custom value and exported in the shell script preamble, but the directory isn't actually created. This is an unpleasant surprise for some things a user might have in their command (e.g. `mktemp`) which assume that if $TMPDIR is set, then it exists.

delocalizer:


cjllanwarne:
👍 LGTM

geoffjentry:
:+1:

Issue #2033, opened 2017-02-26
update readme on spark backend


antonkulaga:
I see that in readme

```
Configuring Spark Project
When using Spark backend uncomment the following Spark configuration in the application.conf file
```

But I do not see it in application.conf Looks like this is outdated.

cjllanwarne:
I think this probably should be move the relevant section across to `application.conf` from `reference.conf`

antonkulaga:
It would be nice to have clearer docs on Spark backend (what should be put to wdl and what should be given to cromwell.jar instead of default application.conf)

geoffjentry:
Attn @francares & @gauravs90 - there's a request for some love on the Spark backend

gauravs90:
@jainh Can you update / help clarify the readme docs for spark?

jainh:
@antonkulaga I will update Readme to point right configuration, however on `what should be given when running java -jar Cromwell.jar` it doesn't need any additional runtime attributes besides uncommenting Spark backend configuration from `reference.conf`  and on `what should be put to wdl` is referred here : [WDL](https://github.com/broadinstitute/cromwell#sample-wdl)

katevoss:
The PR was merged, closing this out.

PR #2043, opened 2017-03-01
Update README.md

jainh:
Adding updated reference to reference.conf for Spark Backend. Referring to an issue [Issue](https://github.com/broadinstitute/cromwell/issues/2033)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1334,7 +1334,14 @@ It supports the following Spark deploy modes:
 *  Cluster deploy mode using Yarn resource manager
 
 ### Configuring Spark Project
-When using Spark backend, copy the Spark configuration in reference.conf (available under `core/src/main/resources`) into the main application.conf (in `src/main/resources`):
+
+Cromwell's default configuration file is located at `core/src/main/resources/reference.conf`
+
+To customize configuration it is recommended that one copies relevant stanzas from `core/src/main/resources/reference.conf` into a new file, modify it as appropriate, then pass it to Cromwell via:
+
+java -Dconfig.file=/path/to/yourOverrides.conf cromwell.jar
+
+Spark configuration stanza is as follows: 
 
 ```conf
 Spark {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1335,7 +1335,7 @@ It supports the following Spark deploy modes:
 
 ### Configuring Spark Project
 
-When using Spark backend uncomment the following Spark configuration in the application.conf file
+When using Spark backend uncomment the following Spark configuration in the reference.conf file, which is available under `core/src/resources`
```
cjllanwarne: I'm not 100% sure how these work, but I don't think `reference.conf` is ever read directly. Instead we'd want to copy the spark section across to `application.conf`?

Happy to be told otherwise...

jainh: @cjllanwarne  Thank you for pointing out, it's been a while since i've touched the Cromwell repo. Therefore, not sure about refactoring and moving pieces, if you can help me to make an appropriate configuration change would be nice otherwise i would need to compile and fix it.

cjllanwarne: oh no, I think we should leave the conf files as they are. It's just this sentence that should say "move" instead of "uncomment"

cjllanwarne:
We probably also want to say where to copy it to, e.g. 
```
When using Spark backend, copy the Spark configuration in reference.conf (available under core/src/main/resources) into the main application.conf (in src/main/resources):
```

I think we probably also don't want to have a copy/pasted version in this file, since it's unlikely to be updated if reference.conf is changed? Up to you!

geoffjentry:
Hi @jainh - I don't really like this. What we tell users to do is to take the bits of config they want to override and put them in a separate file and to invoke cromwell with the `-Dconfig.file=....` flag.

Aside: I notice that this isn't quite what it says in the [README](https://github.com/broadinstitute/cromwell#configuring-cromwell) but that's a separate topic which I'll address myself.

geoffjentry:
I updated README in my PR #2058

geoffjentry:
@antonkulaga Would this documentation change have helped you?

If so I'll merge this as-is.

cjllanwarne:
👍

cjllanwarne:
Thanks @jainh

PR #2045, opened 2017-03-01
Don't fail Travis JES for external users. Closes #1717

### LINKED ISSUE #1717: Travis scripts should gracefully exit for external users
Currently when an external user attempts to access our cloud accounts, Travis does not hand out the encryption keys. This is desired behavior, as we do not want to run up bills on unknown code from outside users.

However, currently our test scripts are not checking the variable [`TRAVIS_SECURE_ENV_VARS`](https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables), and are instead trying to access encrypted variables that are not and never will be present. This leads to false negatives where the tests are marked as failed, when instead they should best case be marked as ignored, worst case be marked as implicitly passed.

kshakir:


kshakir:
This is what a (simulated) failure looks like for the two areas edited by this PR:
https://travis-ci.org/broadinstitute/cromwell/builds/206678315

mcovarr:
I agree with the [referenced ticket](https://github.com/broadinstitute/cromwell/issues/1717), I would much prefer the test be marked ignored if we didn't even try to run it.

kshakir:
Travis doesn't have ignored-test-results afaik, nor support for a dynamic yaml test matrix based on internal-vs.-external contributions. If one knows of a way to do what the ticket wants, ignore vs. pass, I'll happily incorporate that fix. Otherwise, this PR reduces red-fatigue.

geoffjentry:
Yeah I don't know of a way to do it. At some point we're not going to be able to magic away the red-for-external-people. OTOH it's a low frequency event for now, it's not too onerous to ask one of us to run centaur manually or something else along those lines.

mcovarr:
:+1:

Horneth:
👍

PR #2049, opened 2017-03-03
Use mktemp to make temporary directory Fixes #2047

### LINKED ISSUE #2047: Standard backend's "<cwd>/tmp" dir means command cannot use own path "tmp"
The resolution of #2035 exposed this issue. A centaur issue was mitigated by https://github.com/broadinstitute/centaur/pull/165.

A proper fix in the [StandardBackend](https://github.com/broadinstitute/cromwell/blob/335e9838abce1405f0d26fe18c5dba2dfa7ad0f5/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala#L173) will enable a centaur test to run a task with command:

```
command {
  echo "hello" > tmp
}
```

This regression centaur test should run on all backends.

delocalizer:
`mktemp` exists on all mainstream Linux distros, also BSD and OSX. It creates a temporary file or directory with a name containing a random string so you won't get the name clash.

### DIFF
--- a/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
+++ b/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
@@ -176,7 +176,7 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
     val globFiles = backendEngineFunctions.findGlobOutputs(call, jobDescriptor)
 
     s"""|#!/bin/bash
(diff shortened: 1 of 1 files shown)


kshakir:
Awesome! 👍

#2047 should stay open though until we have the centaur regression test setup, mentioned in the ticket.

mcovarr:
:+1: Thank you @delocalizer this is great!  😄   Can you please squash down to one commit before we merge?

mcovarr:
I've merged this as-is to make sure it gets into the release!  🎉

PR #2053, opened 2017-03-06
temporary directory should be world writeable

mcovarr:
Not a huge deal, but otherwise the temp directory may only be accessible by the container user (likely root) without a `sudo`.  This can be an issue when trying to scan or clean up the workflow directory.

### DIFF
--- a/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
+++ b/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
@@ -177,6 +177,7 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
 
     s"""|#!/bin/bash
         |tmpDir=$$(mktemp -d $cwd/tmp.XXXXXX) 
+        |chmod 777 $$tmpDir
         |export _JAVA_OPTIONS=-Djava.io.tmpdir=$$tmpDir
         |export TMPDIR=$$tmpDir
         |$commandScriptPreamble


geoffjentry:
people might not appreciate chmoding of things to 777 in SFS backends

kshakir:
👍 

Re: @geoffjentry's comment -- see also [similar](https://github.com/broadinstitute/cromwell/blob/e7791bee565a39cb00e82dba513abd6652532bb1/supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystemAsyncJobExecutionActor.scala#L122) existing [behavior](https://github.com/broadinstitute/cromwell/blob/e7791bee565a39cb00e82dba513abd6652532bb1/core/src/main/scala/cromwell/core/path/EvenBetterPathMethods.scala#L55-L78).

geoffjentry:
@kshakir I'd say that's reason to change existing behavior. We can't be making things world readable all over the place unless they're truly inaccessible to anything other than the job.

kshakir:
ToL: How about yet another knob? May want to file an issue specifying exact behavior and maybe naming: "okToMakeDirectoriesReadWritable" / "dontFailWhenRunningNonRootDockerImage" / etc.

geoffjentry:
Now that I think about it I think I complained about this the first time around, or at least said that this would eventually cause some sort of catastrophe :)

cjllanwarne:
Can you instead create the directory outside of the script (i.e. as the `cromwell` user rather than `root`) and umask it for `777` (which is... `000`?)

cjllanwarne:
Alternatively, if we send a script to docker, it could end by transferring any files/directories it created back to `cromwell`?

kcibul:
I'm thinking perhaps this should be address instead through the use of docker User Namespaces and described in the documentation (a critical part of the solution!)

When docker runs a container as root, it isolates the user/groups from the host definitions and instead remaps them.  This could be important to a user for two reasons.  Once, files written as root in the container will instead be written as this remapped user.  Second, since IO on the underlying host VM is currently done as root, the container can read files as root.  So if you mounted in /etc/passwd you could read/write on top of that.

For a good tutorial see:

E.g. http://blog.aquasec.com/docker-1.10-user-namespace

For more details see:

https://docs.docker.com/engine/reference/commandline/dockerd/#daemon-user-namespace-options

Also pinging @davidbernick for more thoughts from a security perspective

mcovarr:
Remapping user ids would certainly be more elegant than what we're doing now.  However for a given Docker image we would need to know the user id being mapped from (i.e. we [shouldn't assume this is root](https://github.com/broadinstitute/cromwell/issues/472)).  So it sounds like we'd need to both dynamically determine the user id as which a container will run and dynamically remap that user id to the user id under which the Cromwell server is running.

All of that sounds great but is definitely more than I was trying to bite off with this little PR.  🙂  I'm happy to close this PR if it's too worrisome.

geoffjentry:
Yeah, I've seen this same issue pop up in other spots in the bioinf containers & workflows and I've not seen a wonderful solution that's been demonstrated to work real world. Disclaimer: I might have missed and or forgotten a wonderful solution

geoffjentry:
Discussed in person. we're going to merge this and then handle the docker permission thing separately.

ruchim:
:+1:

PR #2058, opened 2017-03-08
Have our config instructions match what we tell people

geoffjentry:


README.md
```diff
@@ -417,7 +417,7 @@ For many examples on how to use WDL see [the WDL site](https://github.com/broadi
 
 # Configuring Cromwell
 
-Cromwell's default configuration file is located at `src/main/resources/application.conf`.
+Cromwell's default configuration file is located at `core/src/main/resources/reference.conf`.
```
cjllanwarne: I think the default is still `application.conf`. Reference is just a bunch of handy copy/pastable samples

README.md
```diff
@@ -445,10 +445,11 @@ This allows any value to be overridden on the command line:
 java -Dwebservice.port=8080 cromwell.jar ...
 ```
 
-It is recommended that one copies `src/main/resources/application.conf`, modify it, then link to it via:
+
+To customize configuration it is recommended that one copies relevant stanzas from `core/src/main/resources/reference.conf` into a new file, modify it as appropriate, then pass it to Cromwell via:
 
 ```
-java -Dconfig.file=/path/to/application.conf cromwell.jar ...
+java -Dconfig.file=/path/to/yourOverrides.conf cromwell.jar ...
```
cjllanwarne: Suddenly, camelCase!

README.md
geoffjentry: This might be a phrasing thing. Our application.conf file is only config for non-cromwell stuff, e.g. akka. The reference.conf is our defaults (plus as you note copypasta) for cromwell.

README.md
geoffjentry: I'm not sure why the formatting of an example random file name matters here? Do we have examples of other fake file names which are otherwise?

README.md
cjllanwarne: If I read this as-is, I'd assume that anything in `reference.conf` was going to be used by Cromwell.
I'll leave it up to you (or our recently-returned travel-writer @katevoss) whether to change it

geoffjentry: Let's chat on slack or something as I don't understand what you're getting at. When I read what you wrote as-is I'd say "yeah, exactly" to it :)

README.md
cjllanwarne: It doesn't really, 100% ToL. It just surprised me.

It just happens to be wrapped up in the same "🚫 changes needed" as the `reference.conf` comment by accident

README.md
geoffjentry: Discussed in person, leaving as is unless there are further objections

cjllanwarne:
👍

kshakir:
👍

PR #15, opened 2015-05-13
Run the command in a bash subshell.

mcovarr:

PR #2070, opened 2017-03-16
stage required config changes for funnel (TES Backend)

adamstruck:
Once we merge PR #15, funnel is going to start respecting resource requests from tasks.  The defaults in the Cromwell TES backend are 1 cpu and 2 GB of RAM.  So if we want the centaur tests to finish in a reasonable timeframe we need to hard code in local resources into Funnel's config.

### DIFF
--- a/src/bin/travis/resources/tes.conf
+++ b/src/bin/travis/resources/tes.conf
@@ -7,9 +7,13 @@ Storage:
         - /cromwell-executions
         - /tmp/
 DBPath: /tmp/tes_task.db
-Schedulers:
-  Local:
-    NumWorkers: 4
+Scheduler: local
 Worker:
   LogLevel: info
-  Timeout: 1
+  Timeout: -1
+  # Funnel (TES implementation) respects resource reqs
+  # Defaults 1 cpu 2 GB ram would make centaur take too long to run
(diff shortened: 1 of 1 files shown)


kshakir:
👍

src/bin/travis/resources/tes.conf
```diff
@@ -7,9 +7,13 @@ Storage:
...
 Worker:
   LogLevel: info
-  Timeout: 1
+  Timeout: -1
+  # Funnel (TES implementation) respects resource reqs
+  # Defaults 1 cpu 2 GB ram would make centaur take too long to run
+  Resources:
+    cpus: 100
+    ram: 200
+    disk: 1000
```
cjllanwarne: OOI, where does this get provisioned from... I ask because are we going to be creating this somewhere every time we run the TES tests in Travis?

cjllanwarne:
👍 from me

src/bin/travis/resources/tes.conf
adamstruck: For these tests Funnel is running everything locally.  These declarations are to prevent Funnel from detecting available resources automatically so we can "trick" it into running more tests in parallel.

PR #2076, opened 2017-03-20
Develop - Fix custom labels

ruchim:


kshakir:
👍

database/migration/src/main/resources/changesets/replace_empty_custom_labels.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.3.xsd">
+
+    <changeSet author="rmunshi" id="replace_empty_custom_labels" dbms="mysql">
+        <sqlFile relativeToChangelogFile="true" path="migration/workflow_store/replace_empty_custom_labels.sql" />
```
cjllanwarne: Is there any reason for not doing this in a liquibase `update`?

e.g. 
```
<changeSet author="liquibase-docs" id="update-example">
    <update catalogName="cat"
            schemaName="public"
            tableName="person">
        <column name="address" type="varchar(255) value="blah"/>
        <where>A String</where>
    </update>
</changeSet>
```

cjllanwarne:
If possible, I think I'd prefer the liquibase style... but otherwise 👍

database/migration/src/main/resources/changesets/replace_empty_custom_labels.xml
```diff
@@ -0,0 +1,20 @@
...
+    </changeSet>
+
+    <changeSet id="custom_labels_not_null" author="rmunshi">
+        <addNotNullConstraint tableName="WORKFLOW_STORE_ENTRY"
+                              columnName="CUSTOM_LABELS"
+                              columnDataType="LONGTEXT"
+                              defaultNullValue="{}"/>
+    </changeSet>
+
+</databaseChangeLog>
```
kshakir: You cannot touch this file anymore. It has already has an MD5 in Liquibase as of here: https://github.com/broadinstitute/cromwell/blob/18b5b0e31f7bb5b76f27e8c3ca3a876492583786/database/migration/src/main/resources/changesets/replace_empty_custom_labels.xml

ruchim: does this mean it's too late to switch to using liquibase?

ruchim: @cjllanwarne sorry Chris, I'll have to rollback to the sql file version--but I did like the ease of using the liquibase style change, I'll keep it in mind for the future.

kshakir: Today I Learned™: One can go back and modify _some_ parts of the changelog: http://www.liquibase.org/2009/03/what-effects-changeset-checksums.html

This helps my OCD RE: newlines at the end of files, but I don't think one will get away with changing other parts of the change set contents.

PR #2080, opened 2017-03-21
Some retry improvements

Horneth:


CHANGELOG.md
```diff
@@ -2,13 +2,18 @@
...
  
 ```
 system.io {
+  # Global Throttling - This is mostly useful for GCS and can be adjusted to match
+  # the quota availble on the GCS API
   number-of-requests = 100000
   per = 100 seconds
+  
+  # Number of times an I/O operation should be attempted before giving up and failing it.
+  nb-attempts = 5
```
geoffjentry: can this be something more descriptive? e.g. `number-of-attempts`. I don't think it needs to have `io` in the name as that's implied by the structure but `nb` caused me to say "huh?"

geoffjentry:
:+1: nice

core/src/test/scala/cromwell/core/actor/RobustClientHelperSpec.scala
```diff
@@ -140,21 +140,21 @@ class RobustClientHelperSpec extends TestKitSuite with FlatSpecLike with Matcher
   
   it should "randomize backpressure timings" in {
     val delegateActor = TestProbe()
-    val backpressureTimeout = 20 seconds
+    val backpressureTimeout = 100 seconds
     val noResponseTimeout = 3 seconds
-    val randomizeFactor = 0.2D
+    val randomizeFactor = 0.5D
```
ruchim: What does this factor represent?

Horneth: It's used to create the range in which the randomized backpressure timeout will be.

So if timeout `backpressureTimeout` is 100 and `randomizeFactor` is 0.5, then the `generateBackpressureTime` should generate a number between 100 - (100 * 0.5) and 100 + (100 * 0.5)

ruchim: cool!

kshakir:
Besides concurring with the updated-names request, LGTM 👍

Issue #2067, opened 2017-03-15
cromwell failing process workflows after restart


gdlex4015:
I was running a bunch of workflows (same wdl, same method) when we had to restart and redeploy cromwell service and I got Errors with no visible message, when we dug into the cromwell metadata we discovered this message that Cris interpreted as a restart issue.
message": "Workflow input processing failed:\nWorkflow contains invalid labels JSON: Unexpected end-of-input at input index 0 (line 1, position 1), expected JSON Value:\n\n^\n"

kshakir:
Just a reminder, this needs to be eventually added to develop as well before this issue is closed.

PR #2083, opened 2017-03-22
Store custom labels.

kshakir:


ruchim:
:+1:

cjllanwarne:
Wait... was this always this way? Can we centaur this at all (hmm, assuming a JES job request its own metadata)?

cjllanwarne:
👍

kshakir:
This is a small patch on the patch of #2067. Yes, labels and the sql converters both need better unit and or centaur tests.

PR #2087, opened 2017-03-23
Improve ExecutionStore and WEA performance

Horneth:


engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
```diff
@@ -22,19 +21,22 @@ object ExecutionStore {
       case _ => None
     }
 
-    new ExecutionStore(keys.flatten.map(_ -> NotStarted).toMap)
+    new ExecutionStore(keys.flatten.map(_ -> NotStarted).toMap, keys.nonEmpty)
   }
 }
 
-case class ExecutionStore(private val store: Map[JobKey, ExecutionStatus]) {
+case class ExecutionStore(private val statusStore: Map[JobKey, ExecutionStatus], hasNewRunnables: Boolean) {
```
geoffjentry: final, and why we do we care about private here?

Horneth: Because we don't want anyone messing directly with this Map. We should only manipulate it through the methods. It's a protection against ourselves.

geoffjentry: Is that a mutable map? If so I missed that & carry on.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
```diff
@@ -44,94 +46,81 @@ case class ExecutionStore(private val store: Map[JobKey, ExecutionStatus]) {
       case node: GraphNode => node.upstreamAncestry exists hasFailedScope
     }
 
-    store exists {
-      case (jobKey, jobStatus) => (jobStatus == NotStarted && !upstreamFailed(jobKey.scope)) || jobStatus == QueuedInCromwell || jobStatus == Starting || jobStatus == Running
-    }
+    keysWithStatus(NotStarted).exists(jobKey => !upstreamFailed(jobKey.scope)) ||
+      keysWithStatus(QueuedInCromwell).nonEmpty ||
+      keysWithStatus(Starting).nonEmpty ||
+      keysWithStatus(Running).nonEmpty
```
mcovarr: would this be more efficient to order the simple `nonEmpty` tests first before the `exists`?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
Horneth: No it's not but even without modifying it, it shouldn't be randomly read without it being thought through, because of how large it can be and the performance cost associated.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
```diff
@@ -44,94 +46,81 @@ case class ExecutionStore(private val store: Map[JobKey, ExecutionStatus]) {
...
-  def hasFailedJob: Boolean = {
-    store.values.exists(_ == ExecutionStatus.Failed)
-  }
+  def hasFailedJob: Boolean = keysWithStatus(Failed).nonEmpty
 
   override def toString = store.map { case (j, s) => s"$j -> $s" } mkString System.lineSeparator()
 
-  def add(values: Map[JobKey, ExecutionStatus]) = this.copy(store = store ++ values)
+  def add(values: Map[JobKey, ExecutionStatus]) = {
+    val hasTerminalStatus = values.values.exists(_.isTerminal)
```
mcovarr: what about `lazy val`ing this so it isn't evaluated if `hasNewRunnables` is `true`?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
```diff
@@ -44,94 +46,81 @@ case class ExecutionStore(private val store: Map[JobKey, ExecutionStatus]) {
...
-  private def arePrerequisitesDone(key: JobKey): Boolean = {
-    val upstream = key.scope.upstream collect {
-      case n: Call => upstreamEntry(key, n)
-      case n: Scatter => upstreamEntry(key, n)
-      case n: Declaration => upstreamEntry(key, n)
+  private def arePrerequisitesDone(doneKeys: List[JobKey])(key: JobKey): Boolean = {
+    val upstreamAreDone = key.scope.upstream forall {
+      case n: Call => upstreamIsDone(key, n, doneKeys)
+      case n: Scatter => upstreamIsDone(key, n, doneKeys)
+      case n: Declaration => upstreamIsDone(key, n, doneKeys)
```
mcovarr: can these be `|`'d into one line?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
```diff
@@ -44,94 +46,81 @@ case class ExecutionStore(private val store: Map[JobKey, ExecutionStatus]) {
...
+      case _ => true
     }
 
-    val shardEntriesForCollector: List[JobKey] = key match {
-      case collector: CollectorKey => emulateShardEntries(collector)
-      case _ => Nil
+    val shardEntriesForCollectorAreDone: Boolean = key match {
```
mcovarr: lazy val?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -308,37 +315,38 @@ case class WorkflowExecutionActor(workflowDescriptor: EngineWorkflowDescriptor,
...
   private def handleCallBypassed(callOutputs: Map[CallKey, CallOutputs], data: WorkflowExecutionActorData) = {
     def foldFunc(d: WorkflowExecutionActorData, output: (CallKey, CallOutputs)) = d.callExecutionSuccess(output._1, output._2)
 
     val updatedData = callOutputs.foldLeft(data)(foldFunc)
-    handleExecutionSuccess(updatedData)
+    stay() using updatedData
   }
-
-  private def handleExecutionSuccess(data: WorkflowExecutionActorData) = {
+
```
mcovarr: @kshakir has made me conscious of these 😄

backend/src/main/scala/cromwell/backend/async/AsyncBackendJobExecutionActor.scala
```diff
@@ -53,10 +53,11 @@ trait AsyncBackendJobExecutionActor { this: Actor with ActorLogging =>
...
 
   private def robustExecuteOrRecover(mode: ExecutionMode) = {
-    withRetry(() => executeOrRecover(mode), executeOrRecoverBackOff) onComplete {
-      case Success(h) => self ! IssuePollRequest(h)
-      case Failure(t) => self ! FailAndStop(t)
-    }
+//    withRetry(() => executeOrRecover(mode), executeOrRecoverBackOff) onComplete {
+//      case Success(h) => self ! IssuePollRequest(h)
+//      case Failure(t) => self ! FailAndStop(t)
+//    }
```
mcovarr: ❌

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
```diff
@@ -44,94 +47,74 @@ case class ExecutionStore(private val store: Map[JobKey, ExecutionStatus]) {
...
-  def hasFailedJob: Boolean = {
-    store.values.exists(_ == ExecutionStatus.Failed)
-  }
+  def hasFailedJob: Boolean = keysWithStatus(Failed).nonEmpty
 
   override def toString = store.map { case (j, s) => s"$j -> $s" } mkString System.lineSeparator()
 
-  def add(values: Map[JobKey, ExecutionStatus]) = this.copy(store = store ++ values)
+  def add(values: Map[JobKey, ExecutionStatus]) = {
+    lazy val hasTerminalStatus = values.values.exists(_.isTerminal)
```
mcovarr: oh I guess this could be a `def` too, not sure that matters...

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
```diff
@@ -44,94 +47,74 @@ case class ExecutionStore(private val store: Map[JobKey, ExecutionStatus]) {
...
-      case _ => false
-    }
-  }
-
-  def findCompletedShardsForOutput(key: CollectorKey): List[JobKey] = store.toList collect {
-    case (k: CallKey, v) if v.isDoneOrBypassed && k.scope == key.scope && k.isShard => k
-    case (k: DynamicDeclarationKey, v) if v.isDoneOrBypassed && k.scope == key.scope && k.isShard => k
+  def findCompletedShardsForOutput(key: CollectorKey): List[JobKey] = doneKeys collect {
+    case k: CallKey if k.scope == key.scope && k.isShard => k
+    case k: DynamicDeclarationKey if k.scope == key.scope && k.isShard => k
```
mcovarr: looks like these lines could be collapsed too

mcovarr:
:+1:

backend/src/main/scala/cromwell/backend/async/AsyncBackendJobExecutionActor.scala
Horneth: Oops

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
Horneth: It's inside the add method so it will be called once at most either way no ?

geoffjentry: Not opining on that but keep in mind that there's both a mmeory & cpu overhead to lazy vals

geoffjentry:
@Horneth out of curiosity, based on however you're measuring this what sort of improvement does this lead to?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
```diff
@@ -22,19 +21,23 @@ object ExecutionStore {
       case _ => None
     }
 
-    new ExecutionStore(keys.flatten.map(_ -> NotStarted).toMap)
+    new ExecutionStore(keys.flatten.map(_ -> NotStarted).toMap, keys.nonEmpty)
   }
 }
 
-case class ExecutionStore(private val store: Map[JobKey, ExecutionStatus]) {
+final case class ExecutionStore(private val statusStore: Map[JobKey, ExecutionStatus], hasNewRunnables: Boolean) {
```
cjllanwarne: Since the `val` is private anyway, in this case might it be easier to just make this a plain old `class`? IIRC the auto-`unapply` from `case class` might ignore the privacy and let us access the `statusStore` regardless

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
```diff
@@ -22,19 +21,23 @@ object ExecutionStore {
...
+
+  // View of the statusStore more suited for lookup based on status
+  lazy val store: Map[ExecutionStatus, List[JobKey]] = statusStore.groupBy(_._2).mapValues(_.keys.toList)
+  lazy val doneKeys = store.filterKeys(_.isDoneOrBypassed).values.toList.flatten
+
+  private def keysWithStatus(status: ExecutionStatus) = store.getOrElse(status, List.empty)
 
-  // This is hugely ripe for optimization, if it becomes a bottleneck
```
cjllanwarne: Hooray!!

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
```diff
@@ -44,94 +47,72 @@ case class ExecutionStore(private val store: Map[JobKey, ExecutionStatus]) {
...
 
-  def add(values: Map[JobKey, ExecutionStatus]) = this.copy(store = store ++ values)
+  def add(values: Map[JobKey, ExecutionStatus]) = {
+    this.copy(statusStore = statusStore ++ values, hasNewRunnables = hasNewRunnables || values.values.exists(_.isTerminal))
+  }
 
   // Convert the store to a `List` before `collect`ing to sidestep expensive and pointless hashing of `Scope`s when
   // assembling the result.
-  def runnableScopes = store.toList collect { case entry if isRunnable(entry) => entry._1 }
+  def runnableScopes = keysWithStatus(NotStarted) filter arePrerequisitesDone(doneKeys)
```
cjllanwarne: That comment looks out of date now

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
```diff
@@ -44,94 +47,72 @@ case class ExecutionStore(private val store: Map[JobKey, ExecutionStatus]) {
...
-    case (k: CallKey, v) if v.isDoneOrBypassed && k.scope == key.scope && k.isShard => k
-    case (k: DynamicDeclarationKey, v) if v.isDoneOrBypassed && k.scope == key.scope && k.isShard => k
+  def findCompletedShardsForOutput(key: CollectorKey): List[JobKey] = doneKeys collect {
+    case k @ (_: CallKey | _:DynamicDeclarationKey) if k.scope == key.scope && k.isShard => k
   }
 
   // Just used to decide whether a collector can be run. In case the shard entries haven't been populated into the
   // execution store yet.
-  case class TempJobKey(scope: Scope with GraphNode, index: Option[Int]) extends JobKey {
+  private case class TempJobKey(scope: Scope with GraphNode, index: Option[Int]) extends JobKey {
```
cjllanwarne: `final` too

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -195,6 +201,7 @@ case class WorkflowExecutionActor(workflowDescriptor: EngineWorkflowDescriptor,
   }
 
   whenUnhandled {
+    case Event(CheckRunnable, data) => stay()
```
cjllanwarne: `case Event(CheckRunnable, _) => stay()`

cjllanwarne:
👍

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
Horneth: Only problem is then we lose the case class `copy` method which we use in the `add` method to update the internal map and return a new ExecutionStore.

geoffjentry: This was exactly why I balked at the `private val` FWIW. While I understand the argument of "it prevents us from possibly doing something stupid performance-wise" that seems to go against the grain of what a case class is inherently for, IMO it's fair to document "please don't do XYZ" and leave it at that.

cjllanwarne: Though, if we're using `copy` in order to change the only `val`, why not just construct a new version directly with a constructor?

Horneth:
@geoffjentry I ran this wdl on local backend, call caching off, and I mocked the backend response to immediately return success instead of running the job. So all jobs finish immediately and at the same time. It's not a realistic use case but it's an easy way to push cromwell hard in terms of execution performance for large scatter.

```
task hello {
  String addressee
  command {
    echo "Hello ${addressee}!"
  }
  runtime {
      docker: "ubuntu"
    }
}

workflow wf_hello {
  String wf_hello_input = "world"
  Array[Int] s = range(200000)
  scatter (i in s) {
    call hello {input: addressee = wf_hello_input }
  }
}
```

Here are the results:

| Branch     | JobStore Writes | ExecutionTime                                               |
|------------|-----------------|-------------------------------------------------------------|
| Develop    | On              | Still computing runnable calls after 30' - no shard started |
| ThisBranch | On              | 8'                                                          |
| ThisBranch | Off             | 1'30"                                                       |

geoffjentry:
@Horneth That's amazing. Awesome work.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
```diff
@@ -102,10 +100,12 @@ final case class ExecutionStore(private val statusStore: Map[JobKey, ExecutionSt
...
-      case collector: CollectorKey => emulateShardEntries(collector).forall(shardKey => doneKeys.exists({ doneKey =>
-        shardKey.scope.fullyQualifiedName == doneKey.scope.fullyQualifiedName &&
-          shardKey.index == doneKey.index
-      }))
+      case collector: CollectorKey =>
+        emulateShardEntries(collector)
+          .map(shard => (shard.scope.fullyQualifiedName, shard.index))
+          .diff(
+            doneKeys.map(key => (key.scope.fullyQualifiedName, key.index))
+          ).isEmpty
```
Horneth: @mcovarr @cjllanwarne I changed this in the last commit which seemed to make it even faster. It makes it go from ~17' to 8' (with job store writing on) and ~11' to ~2' with job writing off.
Just would like a double check that it makes sense and will not break everything.

cjllanwarne: LGTM. If it passes centaur then bravo!

mcovarr: Makes sense and LGTM

PR #2091, opened 2017-03-24
Report causal exceptions during Google auth failures. Closes #2090

### LINKED ISSUE #2090: Report causal exceptions during Google auth failures
While validating Google authentication, [currently](https://github.com/broadinstitute/cromwell/blob/8ededf0995d7fef1914cb744a1124e4c96328106/filesystems/gcs/src/main/scala/cromwell/filesystems/gcs/auth/GoogleAuthMode.scala#L65) only the exception message is logged. If the exception itself contains a causal exception, that information is discarded.

kshakir:


geoffjentry:
👍 

Massive ToL (as in, this just reminded me of something I saw the other day). The akka docs somewhere make the point that for performance reasons one wants to completely handle an exception as close to the call site as possible and then transform into some other object which won't ever do stack unwinding and such. We often do the opposite, passing exceptions all the way back out and then handling them there.

ruchim:
:+1:

PR #2092, opened 2017-03-25
 Move lenthall's spray code over here, with slight mods to handle our …

geoffjentry:
…extra compiler flags.

NB: I'm not going to entertain review comments on things which were taken as-is from lenthall as the goal is to get rid of spray altogether and don't want to waste time fixing up something which is on the way out anyways


geoffjentry:
Still fighting some dependency issues, there are two tests where I can only seem to get one or the other working. closing for now.

PR #2093, opened 2017-03-26
Move lenthall's spray code over here, with slight mods to handle our …

geoffjentry:
…extra compiler flags

NB: I'm not going to entertain review comments on things which were taken as-is from lenthall as the goal is to get rid of spray altogether and don't want to waste time fixing up something which is on the way out anyways

### DIFF
--- a/engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
+++ b/engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
@@ -8,7 +8,6 @@ import cromwell.engine.backend.BackendConfiguration
 import cromwell.services.metadata.MetadataService._
 import cromwell.webservice.WorkflowJsonSupport._
 import cromwell.webservice.metadata.MetadataBuilderActor
-import lenthall.spray.SwaggerUiResourceHttpService
 import spray.http.MediaTypes._
 import spray.http._
 import spray.httpx.SprayJsonSupport._
@@ -21,7 +20,7 @@ import scala.util.{Failure, Success, Try}
 trait SwaggerService extends SwaggerUiResourceHttpService {
   override def swaggerServiceName = "cromwell"
 
-  override def swaggerUiVersion = "2.1.1"
+  override def swaggerUiVersion = "2.2.2"
 }
 
 trait CromwellApiService extends HttpService with PerRequestCreator {
--- a/engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
+++ b/engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
@@ -20,7 +20,7 @@ import scala.util.{Failure, Success, Try}
 trait SwaggerService extends SwaggerUiResourceHttpService {
   override def swaggerServiceName = "cromwell"
 
-  override def swaggerUiVersion = "2.2.2"
+  override def swaggerUiVersion = "2.1.1"
 }
 
 trait CromwellApiService extends HttpService with PerRequestCreator {
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -65,7 +65,7 @@ object Dependencies {
   )
 
   private val sprayServerDependencies = List(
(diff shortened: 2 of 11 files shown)


geoffjentry:
I think some of the test code is redundant with SwaggerServiceSpec but I don't understand swagger well enough to opine. @kshakir - it looks like you did a lot of the swagger work (albeit a long time ago), any thoughts?

kshakir:
Re: redundant test code, everything is separate as far as I can tell, including looking at the coverage counts in coveralls. The various tests I see are:

Existing in cromwell pre-PR:
- Test the cromwell swagger _JSON_ is valid according to the swagger specification
- Test that the cromwell swagger route serves up the json to the expected URL

Now as of this PR:

- Test that common code for serving up a swagger endpoint can serve
  - json or yaml swagger
  - [CORS](https://en.wikipedia.org/wiki/Cross-origin_resource_sharing) support
  - Utility for redirecting browser requests to `/` to the correct swagger UI

Also included in this PR:
- Ensure that "wrapping" a route works, serving all wrapped routes under a new prefix like `/api/*`
- Ensure that common code for easier spray-can binding works

kshakir:
Maintained coveralls coverage of the copy/paste code LGTM 👍

project/Dependencies.scala
```diff
@@ -65,6 +65,7 @@ object Dependencies {
   )
 
   private val sprayServerDependencies = List(
+    "org.webjars" % "swagger-ui" % "2.2.2" % Provided,
```
Horneth: Just curious, why is this provided now ?

Horneth:
👍

project/Dependencies.scala
kshakir: Good catch. I'm guessing the provided was copy/pasta. Also should be using the value `swaggerUiVersion` (also edited in this PR) instead of `"2.2.2"`.

geoffjentry: Yeah. tl;dr Khalid was right, it was a c/p error.

Longer version: I flailed around for a long time due to a swagger version mismatch and this little line changed many times.

geoffjentry: @kshakir how the hell do i reference `swaggerUiVersion` from here? it wasn't even like that before.

kshakir: Whoops! Ignore that part. It's an old issue that doesn't need to be addressed in this PR.

Instead, the swagger version info needs to be refactored. Instead of being written separately in two places, the swagger version should be written by something like [`writeVersionConf`](https://github.com/broadinstitute/cromwell/blob/e3bb43f6ca51ca6ee6c4260f86c5760059719ec1/project/Version.scala#L22-L31), and then [`SwaggerService.swaggerUiVersion`](https://github.com/broadinstitute/cromwell/blob/e3bb43f6ca51ca6ee6c4260f86c5760059719ec1/engine/src/main/scala/cromwell/webservice/CromwellApiService.scala#L24) should be reading it from the [`cromwell-version.conf`](https://github.com/broadinstitute/cromwell/blob/e3bb43f6ca51ca6ee6c4260f86c5760059719ec1/engine/src/main/scala/cromwell/webservice/CromwellApiService.scala#L284).

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
```diff
@@ -21,7 +20,7 @@ import scala.util.{Failure, Success, Try}
 trait SwaggerService extends SwaggerUiResourceHttpService {
   override def swaggerServiceName = "cromwell"
 
-  override def swaggerUiVersion = "2.1.1"
+  override def swaggerUiVersion = "2.2.2"
```
kshakir: Revert this back to `2.1.1`. Due to a bug fixed in swagger-ui 3.x, this causes rendering issues:

screen shot 2017-03-28 at 4 25 15 pm

Unfortunately swagger-ui 3.x has its own issues with our setup that would need to be debugged.

PR #2098, opened 2017-03-27
Catch namespace load throwables Closes #2066

### LINKED ISSUE #2066: Cromwell hangs when it cannot find the WDL for a sub-workflow
- local backend
- cromwell-0.25
- using docker
- single workflow mode

The WDL is in GATK protected.  https://github.com/broadinstitute/gatk-protected/blob/master/scripts/mutect2_wdl/mutect2_multi_sample.wdl

The WDL functions fine, in general, but the sub-workflow WDL needs to be local to the cromwell execution call.  If it is not there, cromwell cannot find the sub-workflow WDL and then it hangs.

See an example of the hanging in automated gatk-protected M2 testing:

https://travis-ci.org/broadinstitute/gatk-protected/jobs/210985238

I had to manually stop the build, since it would hang forever otherwise.

Horneth:


engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -390,8 +390,10 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
           } else {
             List.empty
           }
-          // This .get is ok because we're already in a try/catch.
```
cjllanwarne: This is actually still true, but maybe we can remove the `Try` now, if nothing else in this method looks dangerous?

cjllanwarne:
I'm not sure if this can fix #2066, since we were already in a `Try`? Although this does look a lot neater...

Did you get to the bottom of why the original test wasn't catching this?

Horneth:
@cjllanwarne the reason was that `WdlNamespace.load` was throwing an `ValidationException` which unfortunately is a `Throwable` but not an `Exception`, which is why there's also a PR in lenthall to make `AggregatedException` an `Exception`...

I can't find a `missing_import` test in centaur though

cjllanwarne:
@Horneth it's in `src/main/resources/standardTestCases/missing_imports.test`

Horneth:
@cjllanwarne It's failing to catch it because it's sending the imports as a zip file, which exercises a different code path than Single workflow mode with no zip, which implicitly means "try to find the import files relatively to where I'm running cromwell"

cjllanwarne:
👍

ruchim:
:+1: LGTM

PR #2099, opened 2017-03-27
Docker hashing gets its own subproject.

mcovarr:
I am admittedly sbt challenged so it's possible this is just all wrong...


build.sbt
```diff
@@ -20,6 +20,11 @@ lazy val databaseMigration = (project in file("database/migration"))
   .dependsOn(core)
   .withTestSettings
 
+lazy val dockerHashing = (project in file("dockerHashing"))
+  .settings(dockerHashingSettings: _*)
+  .dependsOn(core)
```
Horneth: Can we make it not depend on core or is it too entangled ?

mcovarr: Not sure, I'll try.

mcovarr: I did manage to fight through most of the entanglement (see fb299b638a49418b9a4871098fd25c42ad2d2a52) with the exception of an exception: 

```
import cromwell.core.actor.StreamActorHelper.ActorRestartException
```

It looks like a common exception type would need to be used instead of this custom exception.  Also I can't guarantee there won't be more problems downstream of this.  😄

Horneth: I'm realizing that this is going to make the `updateDocker` PR much harder to rebase as it was trying to abstract most of the boilerplate code for handling enqueuing failures etc.. in a trait reused by both the docker actor and the I/O actor...
Anyway, If it turns out to be a giant pain to cut the dependency I would not bother too much..

mcovarr: OK let's at least wait until the dust settles on all the work in this area before trying to do this.  Though given the reusable trait it sounds like these subprojects are only growing more entangled.  🙂

Horneth:
👍

cjllanwarne:
LGTM, assuming the rebase goes smoothly and Centaur is happy (and actually runs...)

👍

PR #2101, opened 2017-03-28
Wait for metadata to be flushed in SWM Closes #1932

### LINKED ISSUE #1932: When writing metadata, single workflow mode should wait for complete metadata
The SingleWorkflow mode, via `java -jar cromwell.jar run …`, does not always wait for metadata to be completely processed. Thus when the actor exits, it may write out incomplete metadata json. This mainly seen during intermittent test failures that are looking for expected metadata from the output file:

```scala
[info] SingleWorkflowRunnerActorWithMetadataSpec:
[info] A SingleWorkflowRunnerActor should 
[info] - successfully run a workflow outputting metadata *** FAILED *** (6 seconds, 143 milliseconds)
[info]   java.util.NoSuchElementException: None.get
[info]   at scala.None$.get(Option.scala:347)
[info]   at scala.None$.get(Option.scala:345)
[info]   at cromwell.engine.workflow.SingleWorkflowRunnerActorSpec$OptionJsValueEnhancer$.toStringValue$extension(SingleWorkflowRunnerActorSpec.scala:44)
[info]   at cromwell.engine.workflow.SingleWorkflowRunnerActorSpec$OptionJsValueEnhancer$.toOffsetDateTime$extension(SingleWorkflowRunnerActorSpec.scala:43)
```

Horneth:


engine/src/main/scala/cromwell/engine/workflow/SingleWorkflowRunnerActor.scala
```diff
@@ -208,6 +222,7 @@ object SingleWorkflowRunnerActor {
   case object NotStarted extends RunnerState
   case object SubmittedWorkflow extends RunnerState
   case object RunningWorkflow extends RunnerState
+  case object WaitForFlushedMetadata extends RunnerState
```
cjllanwarne: `WaitingFor...` is closer to the existing state names

cjllanwarne:
Awesome! :+1:

geoffjentry:
Couple of Qs (either way it's a better situation than current):

- Is it possible for write requests to enter into the metadata service after the check events message?
- What happens if the DB is still busy writing data (as you noted last week, that can get backed up)?

mcovarr:
To @geoffjentry's first point, should there be a quiescence period in SWRA?  i.e. after finding there are no pending writes it could transition to a ReChecking state, send itself a ReCheck message on a timer, and check again.

To the second point, admittedly I didn't check this but I assume the transaction has committed after which Cromwell can't really make any additional durability guarantees?

cjllanwarne:
My gut feeling is to be extremely relaxed after the first point. The WorkflowExecutionActor has already exited at this point so nothing exists that would be sending messages to the metadata service. The only problem would be if messages get reordered somehow which IMO is unlikely

Horneth:
First point: Looking at the code again, the only thing that can possibly be sent to metadata after the status of the workflow becomes terminal are workflow failures and end time, due to the ordering of the `onTransition` blocks in the WorkflowActor. I swapped them in the last commit, so from now on nothing should be sent to the metadata after the terminal status event is sent.
I'm happy to add an extra check for robustness but I don't believe it to be necessary.

Second point: If the DB is busy writing data, the `WriteMetadataActor` will be in state `WritingToDb`, in which case it will always reply with `HasPendingWrites`

mcovarr:
:+1:

PR #2102, opened 2017-03-28
Engine and JES changes supporting custom retries. Closes #1925

### LINKED ISSUE #1925: Create the new retry mechanism wiring
Do the following:

- [x] Rename the `Preempted` `ExecutionStatus` to be `RetryableFailure`.
- [x] Change the `WorkflowExecutionActor` to blindly honor `JobFailedRetryableResponse`s. Currently it is deciding whether or not to retry when it receives these. Now it will assume the determination was already made.
- [x] Remove the `system.max-retries` config option
- [x] Change the JABJEA to track both preemptions and non-preemption retries via the KV store
- [x] For any failure from JES determine if the job is to be retried, either using the preemption count supplied by the user or a max non-preemption retry count of 2. If the job is retryable ensure a `JobFailedRetryableResponse` finds its way back to the WEA
- [x] Modify `BackendStatus` such that we can see that the job was either a) preempted or b) retried due to ReasonX

cjllanwarne:


engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/preparation/JobPreparationActor.scala
```diff
@@ -165,35 +199,42 @@ case class JobPreparationActor(executionData: WorkflowExecutionActorData,
...
 object JobPreparationActor {
-  case class JobPreparationActorData(dockerHashRequest: DockerHashRequest,
-                                     inputs: Map[Declaration, WdlValue],
-                                     attributes: Map[LocallyQualifiedName, WdlValue],
-                                     requestLostTimer: Cancellable,
-                                     maxAttempts: Int,
-                                     currentAttempts: Int = 0,
-                                     backpressureTimer: Option[Cancellable] = None
-                                    ) {
+  case class DockerHashLookupData()
```
kshakir: Delete this class. It isn't used anywhere.

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/preparation/JobPreparationActorSpec.scala
```diff
@@ -1,31 +1,39 @@
...
+import org.scalatest.{BeforeAndAfter, FlatSpecLike, Matchers}
 import wdl4s.Declaration
 import wdl4s.values.{WdlString, WdlValue}
 
 import scala.util.{Failure, Success}
 import scala.concurrent.duration._
 import scala.language.postfixOps
 
-class JobPreparationActorSpec extends TestKitSuite with FlatSpecLike with Matchers with ImplicitSender {
+class JobPreparationActorSpec extends TestKit(ActorSystem("JobPrepActorSpecSystem")) with FlatSpecLike with Matchers with ImplicitSender with BeforeAndAfter {
```
kshakir: From the scaladoc of [`TestKit`](https://github.com/akka/akka/blob/v2.4.11/akka-testkit/src/main/scala/akka/testkit/TestKit.scala#L801-L802):

> - the ActorSystem passed into the constructor needs to be shutdown,
>    otherwise thread pools and memory will be leaked

Either use `TestKitSuite("JobPrepActorSpecSystem")`, or cleanup the thread pool after this suite finishes.

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
```diff
@@ -417,7 +469,9 @@ class JesAsyncBackendJobExecutionActor(override val standardParams: StandardAsyn
       case Success(gcsPath: GcsPath) =>
         val localPath = workingDisk.mountPoint.resolve(gcsPath.pathWithoutScheme).pathAsString
         WdlFile(localPath, wdlFile.isGlob)
-      case _ => wdlFile
+      case _
+
+      => wdlFile
```
kshakir: revert

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/PreviousRetryReasons.scala
```diff
@@ -0,0 +1,51 @@
...
+import scala.util.{Failure, Success, Try}
+import JesBackendLifecycleActorFactory.preemptionCountKey
+import JesBackendLifecycleActorFactory.unexpectedRetryCountKey
+
+case class PreviousRetryReasons(preempted: Int, unexpectedRetry: Int) {
+  // E.g. preempted=3, unexpectedRetry=4, maxAttempts=7 => false
+  // E.g. preempted=3, unexpectedRetry=4, maxAttempts=8 => true
+  def retryable(maxAttempts: Int) = {
+
+  }
```
kshakir: Not sure where this was used in the past, but this dead code returns `Unit` and doesn't have any callers.

supportedBackends/jes/src/test/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActorSpec.scala
```diff
@@ -192,41 +209,48 @@ class JesAsyncBackendJobExecutionActorSpec extends TestKitSuite("JesAsyncBackend
...
+          case response: JobFailedNonRetryableResponse => if(shouldRetry) fail(s"A job ($descriptor) was sent back to the engine with: $response")
+          case response: JobFailedRetryableResponse => if(!shouldRetry) fail(s"A job ($descriptor) that shouldn't be retried was sent back to the engine with $response")
+          case huh => fail(s"Unexpected response: $huh")
         }
       }
     }
   }
 
-  it should "not restart 2 of 1 unexpected shutdowns without another preemptible VM" in {
+  it should "not restart 2 of 1 unexpected shutdowns without another preemptible VM" ignore {
```
kshakir: ToL: When are these expected to return?

backend/src/main/scala/cromwell/backend/backend.scala
```diff
@@ -69,4 +71,5 @@ final case class AttemptedLookupResult(name: String, value: Try[WdlValue]) {
   def toPair = name -> value
 }
 
+// FIXME: Remove
```
cjllanwarne: It's used only in the JABJEA tests that are currently `ignore`d... I've added a tech talk to the board for tomorrow

engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -335,11 +335,11 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
   }
 
   private def validateNamespaceWithImports(w: WorkflowSourceFilesWithDependenciesZip): ErrorOr[WdlNamespaceWithWorkflow] = {
-    def getMetadatae(importsDir: Path, prefix: String = ""): List[(String, Path)] = {
-      importsDir.children.toList flatMap {
+    def getMetadatae(importsDir: Path, prefix: String = ""): Seq[(String, Path)] = {
```
geoffjentry: Please don't change return types from concrete to generic. If anything, go the other direction

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/preparation/JobPreparationActor.scala
```diff
@@ -10,85 +10,111 @@ import cromwell.core.callcaching.docker._
...
-                                     ioActor: ActorRef,
-                                     backendSingletonActor: Option[ActorRef])
-  extends FSM[JobPreparationActorState, Option[JobPreparationActorData]] with WorkflowLogging with DockerClientHelper {
-
-  lazy val workflowIdForLogging = workflowDescriptor.id
-  
+class JobPreparationActor(executionData: WorkflowExecutionActorData,
+                          jobKey: BackendJobDescriptorKey,
+                          factory: BackendLifecycleActorFactory,
+                          val dockerHashingActor: ActorRef,
```
geoffjentry: what prompted this change?

services/src/main/scala/cromwell/services/keyvalue/KeyValueServiceActor.scala
```diff
@@ -1,45 +1,56 @@
...
+  }
+  final case class KvGet(key: ScopedKey) extends KvAction
 
-  sealed trait KvResponse extends KvMessage
-  case class KvPair(key: ScopedKey, value: Option[String]) extends KvResponse
-  case class KvFailure(action: KvAction, failure: Throwable) extends KvResponse
-  case class KvKeyLookupFailed(action: KvGet) extends KvResponse
-  case class KvPutSuccess(action: KvPut) extends KvResponse
+  sealed trait KvResponse extends KvMessage {
+    def key: ScopedKey
```
geoffjentry: Consider this request completely ignorable.

I find for these sorts trivial overrides that it winds up looking neater to handle the different values with a match in the trait. definitely not in cases where there's any sort of logic but this would IMO be a good use for it.

e.g. 

     def key = this match {
         case x: KvFailure => action.key
         case z: Foo => blah
         .....
     }

services/src/main/scala/cromwell/services/keyvalue/KeyValueServiceActor.scala
```diff
@@ -49,3 +60,22 @@ trait KeyValueServiceActor extends Actor {
     }
   }
 }
+
+/**
+  * Perhaps not great in production, but fine for test cases...!
```
geoffjentry: Perhaps it should go in the test code then?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/preparation/JobPreparationActor.scala
cjllanwarne: Two changes here:
 - Removing `case class`: Because we're extending it in tests (and it wasn't really a case class in the first place)
 - Making the `dockerHashingActor` a val: because it's no longer accessible to `DockerClientHelper` via the `case class` accessor

geoffjentry: Yeah, I was asking specifically about the val.  I didn't notice the dropping of the case, gotcha.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
cjllanwarne: Yeah this seems weird... will revert!

services/src/main/scala/cromwell/services/keyvalue/KeyValueServiceActor.scala
cjllanwarne: :man_shrugging:

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
```diff
@@ -353,62 +384,84 @@ class JesAsyncBackendJobExecutionActor(override val standardParams: StandardAsyn
...
-                 |Error code $errorCode. Message: $errorMessage)""".stripMargin)
-            FailedRetryableExecutionHandle(e, returnCode)
+    previousRetryReasons match {
+      case Valid(PreviousRetryReasons(p, ur)) =>
+        val thisPreemption = p + 1
+        val taskName = s"${workflowDescriptor.id}:${call.unqualifiedName}"
+        val baseMsg = s"Task $taskName was preempted for the ${thisPreemption.toOrdinal} time."
+
+        writeFuturePreemptedAndUnexpectedRetryCounts(thisPreemption, ur).map { case _ =>
+          if (thisPreemption < maxPreemption) {
```
Horneth: This one is a `<`, but the one on line 429 is a `<=`, although they both seem to be carrying the same logic. Is that on purpose ?

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
```diff
@@ -353,62 +384,84 @@ class JesAsyncBackendJobExecutionActor(override val standardParams: StandardAsyn
...
-      FailedNonRetryableExecutionHandle(exception, returnCode)
+    val msg = s"Retrying. $errorMessage"
+
+    previousRetryReasons match {
+      case Valid(PreviousRetryReasons(p, ur)) =>
+        val thisUnexpectedRetry = ur + 1
+        if (thisUnexpectedRetry <= maxUnexpectedRetries) {
+          // Increment unexpected retry count and preemption count stays the same
+          writeFuturePreemptedAndUnexpectedRetryCounts(p, thisUnexpectedRetry).map {
+            case _ => FailedRetryableExecutionHandle(StandardException(errorCode, msg, jobTag), jobReturnCode)
```
Horneth: the `case` is unnecessary here, could just be `_ =>`

Horneth:
A few minor comments otherwise LGTM 👍

engine/src/test/scala/cromwell/engine/workflow/WorkflowActorSpec.scala
```diff
@@ -157,9 +159,12 @@ class WorkflowActorSpec extends CromwellTestKitSpec with WorkflowDescriptorBuild
     "copy workflow logs in the event of MaterializeWorkflowDescriptorFailureResponse" in {
       val actor = createWorkflowActor(MaterializingWorkflowDescriptorState)
       deathwatch watch actor
+
+      copyWorkflowLogsProbe.expectNoMsg(AwaitAlmostNothing)
```
ruchim: What's this check for?

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
```diff
@@ -80,11 +98,20 @@ class JesAsyncBackendJobExecutionActor(override val standardParams: StandardAsyn
...
+  }
+
   override def tryAbort(job: StandardAsyncJob): Unit = {
     Run(job, initializationData.genomics).abort()
   }
 
-  override def receive: Receive = pollingActorClientReceive orElse runCreationClientReceive orElse ioReceive orElse super.receive
+  override def requestsAbortAndDiesImmediately: Boolean = true
+
+  override def receive: Receive = pollingActorClientReceive orElse runCreationClientReceive orElse ioReceive orElse kvClientReceive orElse super.receive
```
ruchim: this makes my head spin a bit....but I've also contributed to the problem

ruchim:
:+1: Minor questions, Nice testing. Travis looks like its harassing you...

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
```diff
@@ -353,62 +384,84 @@ class JesAsyncBackendJobExecutionActor(override val standardParams: StandardAsyn
...
   }
 
-  override def handleExecutionFailure(runStatus: RunStatus,
-                                      handle: StandardAsyncPendingExecutionHandle,
-                                      returnCode: Option[Int]): ExecutionHandle = {
-    import lenthall.numeric.IntegerUtil._
+  private def writeFuturePreemptedAndUnexpectedRetryCounts(p: Int, ur: Int): Future[Unit] = {
+    val updateRequests = Seq(
+      KvPut(KvPair(ScopedKey(workflowId, futureKvJobKey, JesBackendLifecycleActorFactory.unexpectedRetryCountKey), Option((ur + 1).toString))),
+      KvPut(KvPair(ScopedKey(workflowId, futureKvJobKey, JesBackendLifecycleActorFactory.preemptionCountKey), Option(p.toString)))
```
cjllanwarne: Wait, why would it?

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
cjllanwarne: It does work out, because when preemption is on the max value it retries non-preemptible, whereas when an unexpected retry is on the max value, we still want it to retry once more:

| This Preemption| Max Preemptions| Result  |
| ------------- |:-------------:| -----:|
|2|3| Retry preemptible |
|3|3| Retry non-preemptible |

| This Unexpected Retry| Max Unexpected Retries| Result  |
| ------------- |:-------------:| -----:|
|1|2| Retry |
|2|2| Retry |
|3|2| Fail |

engine/src/test/scala/cromwell/engine/workflow/WorkflowActorSpec.scala
cjllanwarne: This one makes sure that the `copyWorkflowLogsProbe` doesn't receive any messages before the materialization happens.
Line 167 makes sure the finalization actor never receives any messages throughout this test

services/src/main/scala/cromwell/services/keyvalue/KeyValueServiceActor.scala
geoffjentry: I assumed you thumbsuping a comment meant that you were going to make the change :)

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
geoffjentry: Yeah this is something we're going to need to figure out at some point as it's pretty much impossible to track stuff down now, but it is what it is at the moment

geoffjentry:
I'm still curious to hear what was the cause of pain earlier today but otherwise :+1:

cjllanwarne:
@geoffjentry the 

> Extra logging around unexpected keys

commit was the key:

screen shot 2017-04-05 at 12 13 07 pm

Before moving that message send I was seeing that programming error appear as a rare race condition (but often enough to fail a few sbt tests every time).

I think my mistake was that the `createResponseSet` wasn't necessarily called from a `receive` method so akka was quite at liberty to interleave it with calls to `fulfillOrLog`, which I had assumed would be impossible.

services/src/main/scala/cromwell/services/keyvalue/KeyValueServiceActor.scala
cjllanwarne: ... perhaps...

geoffjentry: Is it used anywhere outside of tests?

cjllanwarne: Looks like sarcasm doesn't work well in text form. Actually I just tried this and I'm getting issues with test-only artifacts not being picked up between different subprojects. IMO leaving it in `services/main` is better than copy/pasting it between multiple subprojects (sfsBackend, jesBackend) but happy to do that if you prefer?

cjllanwarne: UPDATE: I think I found how to do this

PR #2104, opened 2017-03-29
Updated to Slick 3.2.

kshakir:
Removed deadlock workaround.
Minimal updates to fix depracated usage.
Old Slick 3.1 style names are still used in scala, ex: `val driver: JdbcProfile`.
Added notes about SQL converters handling empty LOBs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,25 @@
 # Cromwell Change Log
 
+## 27
+
+
+* The update to Slick 3.2 requires a database stanza to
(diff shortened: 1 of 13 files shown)


geoffjentry:
@horneth do you still have the testing environment you were using for scale stuff recently? I'd be curious to see a before & after just to make sure there's not something else lurking here besides that one bug

geoffjentry:
Other than wanting to make sure we've seen it in action :+1:

mcovarr:
:+1:

kshakir:
Ready for merging assuming it passes a regression test.

PR #2114, opened 2017-03-30
Adds submitted and aborted status to Metadata

Horneth:


engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -192,6 +194,7 @@ class WorkflowManagerActor(params: WorkflowManagerActorParams)
         case None =>
           // All cool, if we got this far the workflow ID was found in the workflow store so this workflow must have never
           // made it to the workflow manager.
+          pushCurrentStateToMetadataService(id, WorkflowAborted)
```
cjllanwarne: Presumably the `Aborted` is already sent to the metadata for workflows that had already started?

cjllanwarne:
👍

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
Horneth: Yes, this is for workflows that have not been started yet and are sitting in the workflow store.

ruchim:
:+1:

Horneth:
@helgridly This will fix your missing status in metadata issue

helgridly:
Awesome! This makes me very happy.

PR #2116, opened 2017-03-31
Fix funnel test and pin version to git commit

buchanae:
@adamstruck


Horneth:
@buchanae Looks like it's still failing, looking at the error message seems like it gets the path to the config file wrong. Should it be `--config` instead of `-config` ?

Horneth:
👍

kshakir:
👍

PR #2118, opened 2017-03-31
Shutdown materializer and Http pool

Horneth:


geoffjentry:
👍

mcovarr:
Is this just good hygiene or does it address a specific issue?

engine/src/main/scala/cromwell/server/CromwellSystem.scala
```diff
@@ -13,9 +15,13 @@ trait CromwellSystem {
...
   val logger = LoggerFactory.getLogger(getClass.getName)
   implicit final lazy val actorSystem = newActorSystem()
+  implicit final lazy val materializer = ActorMaterializer()
+  implicit private final lazy val ec = actorSystem.dispatcher
 
   def shutdownActorSystem(): Future[Terminated] = {
-    actorSystem.terminate()
+    Http().shutdownAllConnectionPools() map { _ =>
+      materializer.shutdown()
+    } flatMap { _ => actorSystem.terminate() }
```
mcovarr: Do we really want to only do the latter things if the former succeed?

mcovarr:
:+1:

engine/src/main/scala/cromwell/server/CromwellSystem.scala
geoffjentry: IIRC (which i probably don't as I didn't go look) this is actually what the docs show to do

Horneth: @mcovarr yeah I was wondering the same. I think it wouldn't hurt putting it in an `andThen` instead of a map.

PR #2141, opened 2017-04-04
[HtCondor] Re-using SFS backend implementation

francares:
This PR is the output from an analysis made in order to see if SFS backend functionality can be re-used as part of HtCondor backend for CCC.
The end result was that most of the things can be re-used but caching since it's being centralized in someway by SFS.
The idea is to add this contribution to this repository since it reduces LOCs but in the CCC context the old version will still be supported.


geoffjentry:
@francares At this point I just took a quick skim so nothing substantive to say but in general I love that this is moving to the config backend. I was curious if you could provide more explanation on what motivates the changes to the runtime attr structure?

README.md
```diff
@@ -1267,90 +1267,132 @@ The `job-id-regex` should contain one capture group while matching against the w
 
 Allows to execute jobs using HTCondor which is a specialized workload management system for compute-intensive jobs created by the Center for High Throughput Computing in the Department of Computer Sciences at the University of Wisconsin-Madison (UW-Madison).
```
geoffjentry: I'm not asking for this to be changed here as I don't think the required structure is in place yet. However, imagine a hypothetical world I'm about to describe and let me know what your thoughts are if that world were real.

At the moment the doc for the SGE backend is a weird hodge podge of SGE and the config backend. I think it'd be nice to just have an explanation for using the config backend in general and how it is generally useful to enable common HPC systems, with pointers to default setups in the reference.conf for specific examples.

If that were to be true, would you agree that the specialized docs for HTCondor wouldn't be necessary, or no? If no, what leads you to say that?

francares:
@geoffjentry I did those modifications to runtime attributes memory validation in order to re-use same behavior/code for disk attribute. HtCondor, SGE and TES use same way (data unit) to define disk capacity which is the same way to define memory size as well.

README.md
francares: I agree. I think this can be reduced to just one section that describes config backend and then a subsection with all supported backends so the people know that HtCondor, SGE, etc. configuration is already provided.

README.md
```diff
@@ -1267,90 +1267,132 @@ The `job-id-regex` should contain one capture group while matching against the w
...
+        actor-factory = "cromwell.backend.impl.sfs.config.ConfigBackendLifecycleActorFactory"
+        # ... other configuration
+      }
+    }
+  }
+}
+```
 
-The `script` file contains:
+This backend makes the same assumption about the filesystem that the local backend does: the Cromwell process and the jobs both have read/write access to the CWD of the job.
```
ruchim: Maybe rephrase this a little bit ```This backend makes the same assumption about the filesystem that local backend does: Cromwell engine has read/write access to the CWD of the job.```

README.md
```diff
@@ -1267,90 +1267,132 @@ The `job-id-regex` should contain one capture group while matching against the w
...
+* `out` - The path to the stdout.
+* `err` - The path to the stderr.
+* `job_name` - A unique name for the job.
 
-### Docker
-This backend supports the following optional runtime attributes / workflow options for working with Docker:
-* docker: Docker image to use such as "Ubuntu".
-* dockerWorkingDir: defines the working directory in the container.
-* dockerOutputDir: defiles the output directory in the container when there is the need to define a volume for outputs within the container. By default if this attribute is not set, dockerOutputDir will be the job working directory.
+This backend also supports docker as optional feature. Configuration key `backend.providers.HtCondor.config.submit-docker` is specified for this end. When the WDL contains a docker runtime attribute, this command will be provided with two additional variables:
```
ruchim: Potential alternative: ```This backend supports docker as an optional runtime attribute, which should be configured using key `backend.providers.HtCondor.config.submit-docker`. When a WDL task uses a docker runtime attribute, the command to run the job will include two additional variables: ```

README.md
```diff
@@ -1267,90 +1267,132 @@ The `job-id-regex` should contain one capture group while matching against the w
...
+EOF
+condor_submit ${cwd}/execution/submitFile
+```
 
-### CPU, Memory and Disk
-This backend supports CPU, memory and disk size configuration through the use of the following runtime attributes / workflow options:
-* cpu: defines the amount of CPU to use. Default value: 1. Type: Integer. Ex: 4.
-* memory: defines the amount of memory to use. Default value: "512 MB". Type: String. Ex: "4 GB" or "4096 MB"
-* disk: defines the amount of disk to use. Default value: "1024 MB". Type: String. Ex: "1 GB" or "1024 MB"
+This backend support additional runtime attributes that are specified in the configuration key `backend.providers.HtCondor.config.runtime-attributes`. It uses the same syntax as specifying runtime attributes in a task in WDL.
```
ruchim: support --> supports

ruchim:
@francares @geoffjentry Taking a look here, it seems like MemoryValidation is actually a "SizeValidation", which should be reusable for things like Memory, Disk Size, etc.

README.md
geoffjentry: Ok cool. Thanks.

README.md
geoffjentry: I'd call this superfluous but I'll fall back to the "we'll clean up the docs later" thing. There's now a ticket in the dsde-docs repo to fix up all the config backend docs.

francares:
@geoffjentry @ruchim Let me know which things you want me to change in this PR.

ruchim:
@francares sorry about the delay. There are separate tickets filed to converge documentation for config based backends so I wouldn't worry about that too much. :+1:  assuming the tests pass.

geoffjentry:
Also :+1: once green.

PR #2154, opened 2017-04-07
Unstick abort, develop edition.

mcovarr:


geoffjentry:
We should probably be clear (particularly for hotfix accepters) that this is unlikely to fix-fix aborts but specifically what we believe it'll resolve.

mcovarr:
@geoffjentry Is that a request for a change here?

geoffjentry:
@mcovarr I should have made that on the hotfix one, it's more in how that's communicated (and if they even care)

Horneth:
👍

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreEngineActor.scala
```diff
@@ -77,11 +76,19 @@ final case class WorkflowStoreEngineActor(store: WorkflowStore, serviceRegistryA
...
           } else {
             sndr ! WorkflowAbortFailed(id, new WorkflowNotFoundException(s"Couldn't abort $id because no workflow with that ID is in progress"))
           }
+        } recover {
+          case t =>
+            val message = s"Error aborting workflow $id: could not remove from workflow store"
+            log.error(t, message)
+            // A generic exception type like RuntimeException will produce a 500 at the API layer, which seems appropriate
+            // given we don't know much about what went wrong here.  `t.getMessage` so the cause propagates to the client.
+            val e = new RuntimeException(s"$message: ${t.getMessage}", t)
```
cjllanwarne: ToL: Bleargh this is stretching the concept of *Runtime* exceptions... why can't we just forward on the causal `t`?

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreEngineActor.scala
```diff
@@ -77,11 +76,19 @@ final case class WorkflowStoreEngineActor(store: WorkflowStore, serviceRegistryA
...
+            sndr ! WorkflowAbortFailed(id, e)
```
cjllanwarne: ToL: Sorry, I can't remember exactly what the original problem was, but I thought it was to do with WMA not getting a response... this is a response to the API, to avoid REST timeouts. But that won't help the WMA timing out? Did I miss something?

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreEngineActor.scala
mcovarr: The causal `t` does not explain the consequence of the error: the workflow could not be removed from the workflow store, so it's still sitting there running and recoverable.

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreEngineActor.scala
mcovarr: This is different from the submission fix.  In this case there was no API response, so the client may have assumed that the abort was successful.  We now make it clear that it was not, so hopefully the client will try again with better results.

ruchim:
Spoke to @mcovarr in-person. My take on the goal of this PR is to catch database errors related to aborts and either report them back to the sender or log them. :+1:

PR #2159, opened 2017-04-07
Knob for script epilogue. (hotfix edition)

mcovarr:
Epilogue defaults to empty string, overridden to `sync` to keep Centaur Local happy.  The knob is currently undocumented, that can certainly change if desired.


Horneth:
👍  modulo comments on the develop edition

PR #2160, opened 2017-04-07
Knob for script epilogue. (develop edition)

mcovarr:


src/bin/travis/resources/local_centaur.conf
```diff
@@ -20,3 +20,4 @@ call-caching {
 backend.providers.Local.config.filesystems.local.caching.duplication-strategy = ["copy"]
 backend.providers.Local.config.filesystems.local.localization = ["soft-link", "copy"]
 backend.providers.Local.config.concurrent-job-limit = 20
+backend.providers.Local.config.script-epilogue = "sync"
```
geoffjentry: what about `sync;sync;sync`, he says only half joking

geoffjentry:
I'm thinking that putting some sync-age in as the default isn't a bad idea, this lets people turn it off if they don't like it. While it doesn't appear to be a panacea it could help and is unlikely to hurt much

mcovarr:
@cjllanwarne and I were discussing the default behavior this afternoon.  While theoretically it seems a sprinkling of `sync`s might prevent problems, I'm not aware of anything bad happening IRL without them (Centaur Local in Travis != RL).  Furthermore, we have at least two users who believe that `sync`s in the Cromwell scripts have caused them problems.

src/bin/travis/resources/local_centaur.conf
mcovarr: For Local Centaur, sure.  😄

geoffjentry:
I'm still concerned about problems happening to people using SFS backends w/o anyone being the wiser. If we can demonstrate definitively that it just doesn't happen outside of local/travis I'll shush up. To date it's always been wishy washy anecdotes on our part based on things like "I don't think I've seen it"

mcovarr:
Per Tech Talk™️, I'll keep the default `sync` behavior and document the knob so users can adjust this value if they see issues.

core/src/main/resources/reference.conf
```diff
@@ -184,6 +184,7 @@ backend {
         #concurrent-job-limit = 5
 
         run-in-background = true
+        script-epilogue = ""
```
cjllanwarne: We want the default in all of the other sample SFS backend stanzas too.

mcovarr: I would propose to delete this line and change the `StandardAsyncExecutionActor` to return `sync` instead of nothing by default.

core/src/main/resources/reference.conf
```diff
@@ -184,6 +184,8 @@ backend {
         #concurrent-job-limit = 5
 
         run-in-background = true
+        # Configures any logic which should run after the command completes
```
cjllanwarne: ToL: How about `A single line shell command to run after the execution of every command block`?

core/src/main/resources/reference.conf
```diff
@@ -184,6 +184,8 @@ backend {
...
+        # script-epilogue = ""
```
cjllanwarne: Very ToL: Would *personally* prefer if this comment showed the default rather than some-other-value (`""`).

CHANGELOG.md
```diff
@@ -43,6 +43,23 @@ system.io {
 }
 ```
 
+* Added a `script-epilogue` configuration option to adjust the logic that runs at the end of the scripts which wrap call executions.
+  This option is adjustable on a per-backend basis.  If unspecified, the default value is `sync`.  To turn off the default `sync` 
+  behavior on the `Local` backend:
```
cjllanwarne: I do this a lot and tend to agree with commenters who say, it's nicer if the details are in the real docs, and the changelog just tells you where to look.

Certainly I've regretted putting details only in the changelog before (e.g. for `concurrent-job-limit`) and then struggling to find that information when I search for it later in the README...

CHANGELOG.md
```diff
@@ -43,6 +43,23 @@ system.io {
...
+  
+```
+backend {
+  default = "Local"
+  providers {
+    Local {
+      actor-factory = "cromwell.backend.impl.sfs.config.ConfigBackendLifecycleActorFactory"
+      config {
+        script-epilogue = ""
```
cjllanwarne: Maybe a more exciting example to show the sort of things you might want to put here?

CHANGELOG.md
mcovarr: I couldn't find a good spot in the README, but I'm open to suggestions.

CHANGELOG.md
mcovarr: The users who have asked for this feature are probably going to use this particular value, but concrete suggestions for exciting alternatives are welcome.  😄

core/src/main/resources/reference.conf
mcovarr: I don't know that it has to be single line necessarily, but I like the rest of that suggestion.

core/src/main/resources/reference.conf
mcovarr: see remarks below

cjllanwarne: (still ToL) The default is `sync, so I'd put that, commented out. That way, people can see what it is they're changing it away from (assuming they copy/paste from reference.conf of course!).

CHANGELOG.md
cjllanwarne: Or just a bigger comment in `reference.conf` as a self-documenting file? And then reference that from here?

CHANGELOG.md
cjllanwarne: If it was me, I'd like to see a commented out "sync", so that I can see what is it that I'm changing it away from

cjllanwarne: (NB consider this equally as ToL as the above comment though)

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -194,7 +197,7 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
         |cd $cwd
         |${globManipulations(globFiles)}
         |)
-        |sync
+        |$scriptEpilogue
```
Horneth: Do we want to do the same `replace(...)` trick in case the user has an epilogue script with a pipe that gets eaten by the `stripMargin` ?

Horneth:
👍

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
mcovarr: ooh yeah

ruchim:
:100:  :+1:

PR #2164, opened 2017-04-10
Test_upgrade script

cjllanwarne:


mcovarr:
Some general documentation would be helpful here as to what is and isn't being tested, known limitations etc.

geoffjentry:
A hearty "seconded" to @mcovarr

geoffjentry:
Make sure to run this all through shellcheck

geoffjentry:
My main concern here is relying on it as a proof of validity until it's guaranteed that it won't go green if it's not picking up workflows in flight.

scripts/test_upgrade/test_upgrade.sh
```diff
@@ -0,0 +1,182 @@
...
+# - Sends the same 3 jobs again, in the same way
+# - Shutdown Cromwell with jobs still running
+# - Restarts Cromwell using the "new" cromwell JAR/configuration
+# - Sends the same three jobs again
+# - Waits for everything to complete
+#
+# What it doesn't do (yet... but maybe for the C27 release!):
+# - Guarantee that any jobs are in any specific states at the point that it shuts down.
+# - Check the status of jobs after completing.
+# - Check that operations IDs weren't duplicated after restarting.
```
mcovarr: Hmm these are pretty significant shortcomings.  My warmth and fuzziness may be impacted.

geoffjentry: I'm happy w/ this being a "it's a good start" but I don't want to see this used as any sort of validation until they're taken care of.

cjllanwarne:
@geoffjentry @mcovarr this script doesn't *validate* anything at all, so yeah...

The implication of that is, a person currently has to follow along, and make sure that things are happening at appropriate times. The comments are sort-of deliberately scary to encourage that :)

scripts/test_upgrade/test_upgrade.sh
cjllanwarne: Not on its own, no. Until they're automated, this is a handy list of things we should be manually checking as the script runs.

geoffjentry:
I'm :+1: insofar as this is viewed as a start in a good direction.

mcovarr:
:+1: what Jeff said

PR #2176, opened 2017-04-14
Add Job Retries on Grid Engine Workflows 

cowmoo:
Hello Cromwell People, 

Currently I believe Cromwell has retry logic for I/O issues or preemptible VMs issues for Google Cloud,

However, when Cromwell jobs that are executed via GridEngine dispatcher will fail with no re-try if the return code is deemed as a error code,

I am submitting a potential patch where Cromwell can retry failed jobs running on GridEngine with user specified retries ("backend.max-job-retries"), 

I'm not sure how the configurations should be organized but here is my starting point; let me know what you guys think

Thanks,
Paul

### DIFF
--- a/backend/src/main/scala/cromwell/backend/BackendJobExecutionActor.scala
+++ b/backend/src/main/scala/cromwell/backend/BackendJobExecutionActor.scala
@@ -2,6 +2,7 @@ package cromwell.backend
 
 import akka.actor.ActorLogging
 import akka.event.LoggingReceive
+import com.typesafe.config.ConfigFactory
 import cromwell.backend.BackendJobExecutionActor._
 import cromwell.backend.BackendLifecycleActor._
 import cromwell.backend.wdl.OutputEvaluator
@@ -46,8 +47,15 @@ trait BackendJobExecutionActor extends BackendJobLifecycleActor with ActorLoggin
   }
 
   // We need this for receive because we can't do `onFailure = ExecutionFailure` directly - because BackendJobDescriptor =/= BackendJobDescriptorKey
-  private def executionFailed = (t: Throwable) =>
-    JobFailedNonRetryableResponse(jobKey = jobDescriptor.key, throwable = t, returnCode = None)
+  private def executionFailed = (t: Throwable)  => {
+    val config = ConfigFactory.load()
+    val maxRetries = Try(config.getInt("backend.max-job-retries")).getOrElse(0)
+
+    if (jobDescriptor.key.attempt > maxRetries)
+      JobFailedNonRetryableResponse(jobKey = jobDescriptor.key, throwable = t, returnCode = None)
+    else
+      JobFailedRetryableResponse(jobKey = jobDescriptor.key, throwable = t, returnCode = None)
+  }
 
   /**
     * Execute a new job.
--- a/backend/src/main/scala/cromwell/backend/async/AsyncBackendJobExecutionActor.scala
+++ b/backend/src/main/scala/cromwell/backend/async/AsyncBackendJobExecutionActor.scala
@@ -2,13 +2,17 @@ package cromwell.backend.async
 
 
 import akka.actor.{Actor, ActorLogging, ActorRef}
+import com.typesafe.config.ConfigFactory
 import cromwell.backend.BackendJobDescriptor
+import cromwell.backend.BackendJobExecutionActor.JobFailedNonRetryableResponse
+import cromwell.backend.BackendJobExecutionActor.JobFailedRetryableResponse
 import cromwell.backend.BackendJobExecutionActor._
 import cromwell.backend.async.AsyncBackendJobExecutionActor._
 import cromwell.core.CromwellFatalExceptionMarker
 import cromwell.core.retry.{Retry, SimpleExponentialBackoff}
 import cromwell.services.metadata.MetadataService.MetadataServiceResponse
 
+import _root_.scala.util.Try
 import scala.concurrent.duration._
 import scala.concurrent.{ExecutionContext, Future, Promise}
 import scala.util.{Failure, Success}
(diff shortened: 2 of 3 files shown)


geoffjentry:
Hi @cowmoo - this appears to be retrying across all backends no matter what the error was. Is that what you intended for this?

cowmoo:
Hi Jeff, 

The short answer is I don't know about the overall Cromwell world. 

The longer answer is: it's intended for me to get around transient issues involving local filesystems and GridEngine dispatcher. 

We'd either run into paths not found when it's a network mount issue; or submitting jobs to a GridEngine queue (e.g., UGER) where the job might get killed after a certain time-period or if it takes up too much resources, 

I understand that Cromwell already have retry logic that deals with I/O issues or pre-emptible VMs in the GCP world. I'm not sure how to organize the configs and the code to harmonize these two retry world's, so I'll leave it to you except to state that we do want some kind of "Retry" in the GridEngine use case. 

Thanks,
Paul

cjllanwarne:
@cowmoo is there a way for Cromwell to tell from the result of running the "begin execution" command, or from the resulting failure, that the task should be retryable?

I like the idea of retrying certain things, but a blanket "retry everything n times, regardless of problem" is probably going to annoy more people than it helps (especially if they're paying for it directly in cloud compute!).

One thing we previously considered was a "retryOnStdoutRegex" attribute, maybe something like:
```
command {
  # ...
}
runtime {
  retryOnStdoutRegex: "path not found:"
}
```

cowmoo:
Hi Chris, 

This would work for us as long as the retryOnStdoutRegex allows for just blanket retries (since it's free for us to do it on internal GridEngine queues), 

Thanks for the follow-up, 
Paul

geoffjentry:
Hi @cowmoo and @MatthewMah (I think, I know this came up recently and I think it was you). This PR goes beyond the scope of what we're comfortable with as it paints with a pretty broad brush. If either of you were interested in implementing something which retried job submission for the config backend only (e.g. SGE, SLURM, etc) we'd be interested in that. We'll likely get to it ourselves at some point but not immediately. If that's something either of you (or anyone else coming across this) are interested in let us know and we can point you in the right direction. It should be relatively straightforward.

genomics-geek:
@geoffjentry - we are facing something similar. Our SGE was recently updated and job submissions randomly fail due `unable to contact qmaster`.   Our HPC team is looking into it.

In the meantime, I am looking for a way to configure Cromwell to retry failed job submissions using the SGE backend.  I have tried adding `maxRetries` to the runtime attributes to retry failed job submissions, but seems like this does not retry job submission errors.  Only retries task errors.  Is that correct?  Any advice would be appreciated.  Is this a feature that is currently supported? Thanks in advance.

I also have seen various different configs on the WDL/Cromwell forum, but not sure if any are still supported.  For example:

[forum post](https://gatkforums.broadinstitute.org/wdl/discussion/10475/cromewell-28-root-configuration-not-working)
```
system {
  max-retries = 10
}
backend {
  max-job-retries = 4
}
```

[forum post](https://gatkforums.broadinstitute.org/wdl/discussion/9576/is-this-error-caused-by-a-job-submission-failure)
```
system {
  max-retries = 50
  job-rate-control {
    jobs = 5
    per = 1 second
  }
}
```

natechols:
@genomics-geek Having just spent some time last week on this (also on SGE), I believe you need this:
https://cromwell.readthedocs.io/en/stable/backends/HPC/#exit-code-timeout

genomics-geek:
@natechols - thanks! I have gave it a try, seems to be what I needed.

genomics-geek:
@natechols - so it works well for failed jobs. However, there seems to be some transient errors on HPC that occur randomly and qsub/qstat go down temporarily and result in `failed (during ExecutingWorkflowState): java.lang.RuntimeException: Unable to start job.`

I was hoping this would retry failed submissions. 

This is my current config:

```

webservice {
  port = 8000
  interface = 127.0.0.1
}

#call-caching {
#  enabled = true
#  invalidate-bad-cache-results = true
#}

system {
  job-rate-control {
    jobs = 20
    per = 1 second
  }
}

backend {
  default = SGE

  providers {
    Local {

      config {
        concurrent-job-limit = 10
        root = "cromwell-executions"
        run-in-background = true

        default-runtime-attributes {
          maxRetries: 3
        }

        runtime-attributes = """
        String ? docker
        String ? docker_user
        """

        submit-docker = """
        docker run \
          --rm -i \
          ${"--user " + docker_user} \
          --entrypoint /bin/bash \
          -v ${cwd}:${docker_cwd} \
          ${docker} ${script}
        """

        filesystems {
          local {
            localization: [
              "hard-link", "soft-link", "copy"
            ]
            caching {
              duplication-strategy: [
                "hard-link", "soft-link", "copy"
              ]
              hashing-strategy: "file"
              check-sibling-md5: false
            }
          }
        }
      }
    }

    SGE {

      config {
        root = "cromwell-executions"
        exit-code-timeout-seconds = 600
        concurrent-job-limit = 100

        default-runtime-attributes {
          maxRetries: 3
        }

        runtime-attributes = """
        Int cpu = 1
        Float ? memory_gb
        String sge_queue = "dgdcloud.q"
        String ? sge_project
        """

        submit = """
        qsub \
        -terse \
        -V \
        -b n \
        -N ${job_name} \
        -wd ${cwd} \
        -o ${out}.qsub \
        -e ${err}.qsub \
        -pe smp ${cpu} \
        ${"-l h_vmem=" + memory_gb / cpu + "g"} \
        ${"-l mem_free=" + memory_gb / cpu + "g"} \
        ${"-q " + sge_queue} \
        ${"-P " + sge_project} \
        ${script}
        """

        kill = "qdel ${job_id}"
        check-alive = "qstat -j ${job_id}"
        job-id-regex = "(\\d+)"
      }
    }
  }
}
```

natechols:
@genomics-geek Just ran into the same problem here.
```
Unable to run job: failed receiving gdi request response for mid=1 (got syncron message receive timeout error)..
Exiting.
error: commlib error: got read error (closing "vm-gridmaster/qmaster/1")
```
This kind of glitch seems pretty common in SGE, unfortunately - it would be nice to have a workaround.

PR #2181, opened 2017-04-18
Fix timing diagram for preempted calls

Horneth:
26 has removed the "Preempted" status and replaced it with "RetraybleFailure", which breaks the timing diagram which only displays the last attempt.
This fixes on 26_hotfix.


mcovarr:
:+1:

geoffjentry:
👍

PR #2184, opened 2017-04-19
Support for docker hash lookup of public quay.io images

Horneth:


mcovarr:
:+1:

geoffjentry:
:+1: 

If you could add a tidbit to the changelog (and if appropriate, the readme, I don't know if there's a block in there about this stuff in general).

PR #2195, opened 2017-04-21
Remove race conditions from WriteMetadataActorSpec. Closes #2193

### LINKED ISSUE #2193: Flaky Test: WriteMetadataActor should Have one event after batch size + 1 is reached

cjllanwarne:
This was previously failing because the foreach was sending 10 events to the `WriteMetadataActor`.

Unfortunately the actor was waiting for the events to be strictly greater than 10 before triggering the write... it occasionally worked because we weren't replacing the actor between tests and the leftover state was giving us a push over the top.

It was also possible for the `WriteMetadataActor` to write the entries to its database and revert back to `WaitingToWrite` before our code was testing what state it was in.

So, I've fixed this test up by removing the inter-test dependency and precisely controlling when the actor completes its "database write".

### DIFF
--- a/services/src/main/scala/cromwell/services/metadata/impl/WriteMetadataActor.scala
+++ b/services/src/main/scala/cromwell/services/metadata/impl/WriteMetadataActor.scala
@@ -9,15 +9,16 @@ import cromwell.services.metadata.MetadataService.{MetadataServiceAction, PutMet
 import cromwell.services.metadata.impl.WriteMetadataActor.{WriteMetadataActorData, WriteMetadataActorState}
 import org.slf4j.LoggerFactory
 
+import scala.concurrent.ExecutionContext
 import scala.concurrent.duration._
 import scala.util.{Failure, Success, Try}
 
-final case class WriteMetadataActor(batchRate: Int, flushRate: FiniteDuration)
+class WriteMetadataActor(batchRate: Int, flushRate: FiniteDuration)
   extends LoggingFSM[WriteMetadataActorState, WriteMetadataActorData] with ActorLogging with
   MetadataDatabaseAccess with SingletonServicesStore {
   import WriteMetadataActor._
 
-  implicit val ec = context.dispatcher
+  implicit val ec: ExecutionContext = context.dispatcher
 
   override def preStart(): Unit = {
     context.system.scheduler.schedule(0.seconds, flushRate, self, ScheduledFlushToDb)
@@ -29,8 +30,8 @@ final case class WriteMetadataActor(batchRate: Int, flushRate: FiniteDuration)
   when(WaitingToWrite) {
     case Event(PutMetadataAction(events), curData) =>
       curData.addEvents(events) match {
-        case data@HasEvents(e) if e.length > batchRate => goto(WritingToDb) using data
-        case e => stay using e
+        case newData: HasEvents if newData.length > batchRate => goto(WritingToDb) using newData
+        case newData => stay using newData
       }
     case Event(ScheduledFlushToDb, curData) =>
       log.debug("Initiating periodic metadata flush to DB")
@@ -73,7 +74,7 @@ final case class WriteMetadataActor(batchRate: Int, flushRate: FiniteDuration)
 }
 
 object WriteMetadataActor {
-  def props(batchRate: Int, flushRate: FiniteDuration) = Props(new WriteMetadataActor(batchRate, flushRate)).withDispatcher(ServiceDispatcher)
+  def props(batchRate: Int, flushRate: FiniteDuration): Props = Props(new WriteMetadataActor(batchRate, flushRate)).withDispatcher(ServiceDispatcher)
   private lazy val logger = LoggerFactory.getLogger("WriteMetadataActor")
 
   sealed trait WriteMetadataActorMessage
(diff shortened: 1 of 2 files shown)


services/src/main/scala/cromwell/services/metadata/impl/WriteMetadataActor.scala
```diff
@@ -29,8 +30,8 @@ final case class WriteMetadataActor(batchRate: Int, flushRate: FiniteDuration)
   when(WaitingToWrite) {
     case Event(PutMetadataAction(events), curData) =>
       curData.addEvents(events) match {
-        case data@HasEvents(e) if e.length > batchRate => goto(WritingToDb) using data
-        case e => stay using e
+        case newData: HasEvents if newData.length > batchRate => goto(WritingToDb) using newData
```
geoffjentry: I'm not sure why `e` as being short for `events` is worse than `newData` as "short" for `events`?

services/src/main/scala/cromwell/services/metadata/impl/WriteMetadataActor.scala
```diff
@@ -9,15 +9,16 @@ import cromwell.services.metadata.MetadataService.{MetadataServiceAction, PutMet
 import cromwell.services.metadata.impl.WriteMetadataActor.{WriteMetadataActorData, WriteMetadataActorState}
 import org.slf4j.LoggerFactory
 
+import scala.concurrent.ExecutionContext
 import scala.concurrent.duration._
 import scala.util.{Failure, Success, Try}
 
-final case class WriteMetadataActor(batchRate: Int, flushRate: FiniteDuration)
+class WriteMetadataActor(batchRate: Int, flushRate: FiniteDuration)
```
geoffjentry: you've now made the two fields private which i think is fine as people shouldn't be accessing an object of this type directly, but figured i'd point it out :)

services/src/main/scala/cromwell/services/metadata/impl/WriteMetadataActor.scala
geoffjentry: is the switcharoo with the `@` and `length` due to the tests? If not i'm not getting why you did this.

cjllanwarne: The second `e` was never short for `events`. It was short for `WriteMetadataActorData`.
The switcheroo was part to make it easier for me to follow for my debugging. I still think it's a little neater so I left it in, but its functionally identical. (shrug)

I think you do have a point though, `WriteMetadataActorData.length` should probably become `WriteMetadataActorData.eventQueueLength`, to make it even easier to follow.

services/src/main/scala/cromwell/services/metadata/impl/WriteMetadataActor.scala
cjllanwarne: Right, we weren't using it like a `case class` and the `final` was getting in my way. Not to mention my own personal "more judicious use of case classes" crusade

services/src/main/scala/cromwell/services/metadata/impl/WriteMetadataActor.scala
geoffjentry: good point

geoffjentry:
Realized I hadn't :+1: this

services/src/test/scala/cromwell/services/metadata/impl/WriteMetadataActorSpec.scala
```diff
@@ -33,12 +40,29 @@ class WriteMetadataActorSpec extends ServicesSpec("Metadata") with Eventually {
     }
 
     "Have one event after batch size + 1 is reached" in {
-      1 to 10 foreach { _ => actor ! Action }
+      1 to WriteMetadataActorSpec.BatchRate foreach { _ => actor ! Action }
+      actor.stateName shouldBe WaitingToWrite
```
mcovarr: This doesn't need `eventually`?

services/src/test/scala/cromwell/services/metadata/impl/WriteMetadataActorSpec.scala
```diff
@@ -48,4 +72,30 @@ class WriteMetadataActorSpec extends ServicesSpec("Metadata") with Eventually {
...
+}
+
+// A WMA that won't (hopefully!) perform a time based flush during this test
+final class DelayingWriteMetadataActor extends WriteMetadataActor(WriteMetadataActorSpec.BatchRate, WriteMetadataActorSpec.FunctionallyForever) {
+
+  var writeToDbInProgress: Boolean = false
+  var writeToDbCompletionPromise: Option[Promise[Unit]] = None
+
+  override def addMetadataEvents(metadataEvents: Iterable[MetadataEvent])(implicit ec: ExecutionContext): Future[Unit] = {
+    writeToDbCompletionPromise = Some(Promise[Unit]())
```
mcovarr: `Option`

mcovarr:
:+1:

services/src/test/scala/cromwell/services/metadata/impl/WriteMetadataActorSpec.scala
cjllanwarne: No, this is checking that nothing's changed. If anything we could pause a few seconds then check, I don't like arbitrary long sleeps though. Any ideas?

PR #2198, opened 2017-04-22
Execution Store and JobPaths optimizations

Horneth:
2 other minor things popped up while call caching the 10K joint genotyping: 
- We spend a fair amount of time creating `GcsPathBuilder`s, which we shouldn't since we only need one per workflow in theory. In practice we need to create a few more because we can't quite propagate the same one around everywhere. But before this PR we would create one per job which seems inefficient. One reason for this is that `JobPaths` extends `WorkflowPaths`, so when we convert the latter to the former we effectively re-instantiate a new `WorkflowPaths` every time. This PR changes that so that `JobPaths` takes a `WorkflowPaths` instead as one of its attribute to avoid unnecessary re-allocations.
- A small optimization to the execution store which allows quicker lookup of "Done" jobs which we do a lot in `runnableCalls`. Also added a benchmark test that measures the performance of `runnableCalls`. Below are the results before and after this change. The "size" corresponds to how many jobs in "Done" and "NotStarted" states are inserted in the execution store before calling `runnableCalls`. Results are in ms.

Before:
screen shot 2017-04-19 at 3 24 15 pm

After:
screen shot 2017-04-19 at 3 25 18 pm

### DIFF
--- a/backend/src/main/scala/cromwell/backend/io/JobPathsWithDocker.scala
+++ b/backend/src/main/scala/cromwell/backend/io/JobPathsWithDocker.scala
@@ -6,17 +6,17 @@ import cromwell.core.path.{Path, PathBuilder}
 
 object JobPathsWithDocker {
   def apply(jobKey: BackendJobDescriptorKey,
-  workflowDescriptor: BackendWorkflowDescriptor,
-  config: Config,
-  pathBuilders: List[PathBuilder] = WorkflowPaths.DefaultPathBuilders) = {
+            workflowDescriptor: BackendWorkflowDescriptor,
+            config: Config,
+            pathBuilders: List[PathBuilder] = WorkflowPaths.DefaultPathBuilders) = {
     val workflowPaths = new WorkflowPathsWithDocker(workflowDescriptor, config, pathBuilders)
     new JobPathsWithDocker(workflowPaths, jobKey)
   }
 }
 
-case class JobPathsWithDocker(override val workflowPaths: WorkflowPathsWithDocker, jobKey: BackendJobDescriptorKey) extends JobPaths {
+case class JobPathsWithDocker private[io] (override val workflowPaths: WorkflowPathsWithDocker, jobKey: BackendJobDescriptorKey) extends JobPaths {
   import JobPaths._
-  
+
   override lazy val callExecutionRoot = { callRoot.resolve("execution") }
   val callDockerRoot = callPathBuilder(workflowPaths.dockerWorkflowRoot, jobKey)
   val callExecutionDockerRoot = callDockerRoot.resolve("execution")
--- a/backend/src/main/scala/cromwell/backend/io/JobPathsWithDocker.scala
+++ b/backend/src/main/scala/cromwell/backend/io/JobPathsWithDocker.scala
@@ -13,7 +13,8 @@ object JobPathsWithDocker {
     new JobPathsWithDocker(workflowPaths, jobKey)
   }
 }
-class JobPathsWithDocker(override val workflowPaths: WorkflowPathsWithDocker, val jobKey: BackendJobDescriptorKey) extends JobPaths {
+
+case class JobPathsWithDocker(override val workflowPaths: WorkflowPathsWithDocker, jobKey: BackendJobDescriptorKey) extends JobPaths {
   import JobPaths._
   
   override lazy val callExecutionRoot = { callRoot.resolve("execution") }
--- a/backend/src/main/scala/cromwell/backend/io/JobPathsWithDocker.scala
+++ b/backend/src/main/scala/cromwell/backend/io/JobPathsWithDocker.scala
@@ -4,15 +4,20 @@ import com.typesafe.config.Config
 import cromwell.backend.{BackendJobDescriptorKey, BackendWorkflowDescriptor}
 import cromwell.core.path.{Path, PathBuilder}
 
-class JobPathsWithDocker(val jobKey: BackendJobDescriptorKey,
-                         workflowDescriptor: BackendWorkflowDescriptor,
-                         config: Config,
-                         pathBuilders: List[PathBuilder] = WorkflowPaths.DefaultPathBuilders) extends WorkflowPathsWithDocker(
-  workflowDescriptor, config, pathBuilders) with JobPaths {
+object JobPathsWithDocker {
+  def apply(jobKey: BackendJobDescriptorKey,
+  workflowDescriptor: BackendWorkflowDescriptor,
+  config: Config,
+  pathBuilders: List[PathBuilder] = WorkflowPaths.DefaultPathBuilders) = {
+    val workflowPaths = new WorkflowPathsWithDocker(workflowDescriptor, config, pathBuilders)
+    new JobPathsWithDocker(workflowPaths, jobKey)
+  }
+}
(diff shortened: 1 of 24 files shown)


cjllanwarne:
Very impressive improvements!

Just OOI, how come it now takes less time when you go from 9k to 10k?!!

Horneth:
Yes that surprised me too. The numbers are averages so it's possible that it kinds of flattens out around ~20/30ms, and the 10000 happened to have slightly better runs. It does it 30 times for each of the sizes. The minimum time value would be a better indicator maybe.

backend/src/main/scala/cromwell/backend/io/WorkflowPathsWithDocker.scala
```diff
@@ -8,11 +8,12 @@ object WorkflowPathsWithDocker {
   val DockerRoot: Path = DefaultPathBuilder.get("/cromwell-executions")
 }
 
-class WorkflowPathsWithDocker(val workflowDescriptor: BackendWorkflowDescriptor, val config: Config, val pathBuilders: List[PathBuilder] = WorkflowPaths.DefaultPathBuilders) extends WorkflowPaths {
+case class WorkflowPathsWithDocker(workflowDescriptor: BackendWorkflowDescriptor, config: Config, pathBuilders: List[PathBuilder] = WorkflowPaths.DefaultPathBuilders) extends WorkflowPaths {
```
geoffjentry: `final` :)

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
```diff
@@ -3,11 +3,15 @@ package cromwell.engine.workflow.lifecycle.execution
...
 import cromwell.core.ExecutionStatus._
 import cromwell.core.{CallKey, JobKey}
+import cromwell.engine.workflow.lifecycle.execution.ExecutionStore.FqnIndex
 import cromwell.engine.workflow.lifecycle.execution.WorkflowExecutionActor.{apply => _, _}
 import wdl4s._
 
+import scala.language.postfixOps
 
 object ExecutionStore {
+  type FqnIndex = (String, Option[Int])
```
geoffjentry: I'd prefer to see case classes instead of `type` unless there's a good reason not to (e.g. "all of my fancy performance gains just went away")

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStoreBenchmark.scala
```diff
@@ -0,0 +1,52 @@
+package cromwell.engine.workflow.lifecycle.execution
```
geoffjentry: I am super +1 on having stuff like this. However it's not clear to me how I'd run this, when I'd run this, if it's automatically run, etc etc.

geoffjentry:
:+1:

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStoreBenchmark.scala
cjllanwarne: Is there CI support for running this and tracking changes over time? Maybe even giving us a red cross on the PR if we screw the benchmarks up too much (kinda like the code coverage stuff... that we ignore)

Horneth: I think all that can be done yes. To be honest I did this more as a quick way to know if what I was doing was making any (good) difference and didn't plan on spending too much time on setting something CI automated up as part of this PR, but that would definitely be nice. Happy to make a ticket for it.
Currently you can run it directly through intellij or with sbt (I'll find the command). It's not run automatically.

geoffjentry: Yeah as awesome as it'd be to have this sort of thing in our CI I think just having "here's how one could run this and here is what it is measuring" in the comments would suffice for now :)

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
Horneth: I feel like in this particular case it would only bring overhead to have a case class. It is only used in this class, in one place, as a key to a hash map, and is never accessed directly. Making a case class for it would mean instantiate an object for every `Done` entry in the execution store every time it changes. The only thing that would give us is compute the hashCode but that will be done anyway if we leave it as a pair.

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStoreBenchmark.scala
Horneth: ok 👍

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
geoffjentry: Yeah, that's why I gave you the parenthetical out :) Since you have the fancy benchmark tool I'd be curious to see what sort of impact it had. I suspect it'd be quite a lot actually.

backend/src/main/scala/cromwell/backend/io/JobPathsWithDocker.scala
```diff
@@ -4,15 +4,21 @@ import com.typesafe.config.Config
...
+  def apply(jobKey: BackendJobDescriptorKey,
+  workflowDescriptor: BackendWorkflowDescriptor,
+  config: Config,
+  pathBuilders: List[PathBuilder] = WorkflowPaths.DefaultPathBuilders) = {
+    val workflowPaths = new WorkflowPathsWithDocker(workflowDescriptor, config, pathBuilders)
+    new JobPathsWithDocker(workflowPaths, jobKey)
+  }
+}
+
+case class JobPathsWithDocker(override val workflowPaths: WorkflowPathsWithDocker, jobKey: BackendJobDescriptorKey) extends JobPaths {
```
mcovarr: If this is no longer meant to be called by anything other than `apply` I think it could be made package `private` (can't check that myself due to ongoing IntelliJ meltdown)

supportedBackends/tes/src/main/scala/cromwell/backend/impl/tes/TesJobPaths.scala
```diff
@@ -1,22 +1,29 @@
...
+  def apply(jobKey: BackendJobDescriptorKey,
+  workflowDescriptor: BackendWorkflowDescriptor,
+  config: Config,
+  pathBuilders: List[PathBuilder] = WorkflowPaths.DefaultPathBuilders) = {
+    val workflowPaths = new TesWorkflowPaths(workflowDescriptor, config, pathBuilders)
+    new TesJobPaths(workflowPaths, jobKey)
+  }
+}
+
+case class TesJobPaths(override val workflowPaths: TesWorkflowPaths,
```
mcovarr: similar comment here as above wrt package `private`

mcovarr:
:+1: awesome!

PR #2203, opened 2017-04-25
Cromwell's failure is now complete (develop edition). Closes #2201

### LINKED ISSUE #2201: Additional Failure Metadata Inconstency
Per @jmthibault79 -

```
failures: [ "single error string" ]
```
Should be:
```
failures: [{
 message: "single error string" 
 causedBy: []
}]
```

cjllanwarne:
- Modifies existing migrations since I had to support multiple statements per migrating row.
- Done in Scala so we only need to do a single pass over everything, but I'm still open to "it'd be better to use **this** SQL command!" comments.
- Tested against things that look like Joel's example inconsistency.

### DIFF
--- a/database/migration/src/main/scala/cromwell/database/migration/custom/BatchedTaskChange.scala
+++ b/database/migration/src/main/scala/cromwell/database/migration/custom/BatchedTaskChange.scala
@@ -34,7 +34,8 @@ trait BatchedTaskChange extends MigrationTaskChange {
   def readBatchQuery: String
 
   /**
-    * Used to prepare the statement that will be passed repeatedly into migrateBatchRow.
+    * Used to prepare query statements that can be passed repeatedly into migrateBatchRow. Examples of
+    * individual elements follow.
     *
     * Example:
     * {{{
@@ -49,7 +50,7 @@ trait BatchedTaskChange extends MigrationTaskChange {
     *   SET [COLUMNS];
     * }}}
     */
-  def migrateBatchQuery: String
+  def migrateBatchQueries: List[String]
 
   /**
     * Migrate a row.
@@ -58,10 +59,10 @@ trait BatchedTaskChange extends MigrationTaskChange {
     * migrateStatement.addBatch(). Return the (estimated) number of rows to be written by this batch.
     *
     * @param readRow          The row to migrate
-    * @param migrateStatement The statement to add a new migrated row
+    * @param migrateStatements The statements used to migrate a row
     * @return The number of rows updated
     */
-  def migrateBatchRow(readRow: ResultSet, migrateStatement: PreparedStatement): Int
+  def migrateBatchRow(readRow: ResultSet, migrateStatements: List[PreparedStatement]): Int
 
   /**
     * Specify the size of a "page".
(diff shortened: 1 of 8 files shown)


cjllanwarne:
Will back-port to 26-hotfix once the develop version gets the seal of approval.

database/migration/src/main/scala/cromwell/database/migration/custom/BatchedTaskChange.scala
```diff
@@ -34,7 +34,8 @@ trait BatchedTaskChange extends MigrationTaskChange {
   def readBatchQuery: String
 
   /**
-    * Used to prepare the statement that will be passed repeatedly into migrateBatchRow.
+    * Used to prepare and supplemental queries that will be passed repeatedly into migrateBatchRow. Examples of
+    * individual elements follow.
```
mcovarr: "prepare and supplemental"?

database/migration/src/main/scala/cromwell/database/migration/failuremetadata/ExpandSingleFailureStrings.scala
```diff
@@ -0,0 +1,71 @@
...
+
+  /**
+    * Find metadata keys in this range which can be fixed up.
+    */
+  override def readBatchQuery: String =
+    s"""|SELECT $primaryKeyColumn, $workflowIdColumn, $metadataKeyColumn, $callFqnColumn, $jobScatterIndexColumn, $retryAttemptColumn, $metadataTimestampColumn
+        | FROM $tableName
+        | WHERE $primaryKeyColumn >= ? AND $primaryKeyColumn < ?
+        | AND $fixableFailureMessageFilter
+        |""".stripMargin
```
mcovarr: why all the interpolations into the query strings?

database/migration/src/main/scala/cromwell/database/migration/failuremetadata/ExpandSingleFailureStrings.scala
```diff
@@ -0,0 +1,71 @@
...
+    s"UPDATE $tableName SET $metadataKeyColumn = ? WHERE $primaryKeyColumn = ?;",
+    // Insert a new element, to contain the empty causedBy[]
+    s"""INSERT INTO $tableName ($workflowIdColumn, $metadataKeyColumn, $callFqnColumn, $jobScatterIndexColumn, $retryAttemptColumn, $metadataTimestampColumn)
+       | VALUES (?, ?, ?, ?, ?, ?);
+     """.stripMargin
+  )
+
+  override def migrateBatchRow(readRow: ResultSet, migrateStatements: List[PreparedStatement]): Int = {
+    val modifyStatement = migrateStatements.head
+    val insertStatement = migrateStatements(1)
```
mcovarr: `val (modifyStatement :: insertStatement :: Nil) = migrateStatements`

mcovarr:
:+1: total failure.

database/migration/src/main/scala/cromwell/database/migration/failuremetadata/ExpandSingleFailureStrings.scala
cjllanwarne: I'm following the pattern established in the other batch migrations. I guess there's no need for it to be a `def` though, I'll switch to a `val`

Horneth:
👍

PR #2208, opened 2017-04-26
Tweak retries

Horneth:
Explicitly add `503` as a retryable error code on `StorageException`s because the `isRetryable` method seems to not be completely reliable.
Also retry SSL Exception (both top level and as a cause for `StorageException`s, I've seen both happen)

### DIFF
--- a/engine/src/main/scala/cromwell/engine/io/IoActor.scala
+++ b/engine/src/main/scala/cromwell/engine/io/IoActor.scala
@@ -137,7 +137,7 @@ object IoActor {
     case _ => false
   }
 
-  val AdditionalRetryablHttpCodes = List(
+  val AdditionalRetryableHttpCodes = List(
     // HTTP 410: Gone
     // From Google doc (https://cloud.google.com/storage/docs/json_api/v1/status-codes):
     // "You have attempted to use a resumable upload session that is no longer available.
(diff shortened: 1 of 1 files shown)


mcovarr:
:+1:

engine/src/main/scala/cromwell/engine/io/IoActor.scala
```diff
@@ -141,15 +142,21 @@ object IoActor {
...
+    503
   )
   
   /**
     * Failures that are considered retryable.
     * Retrying them should increase the "retry counter"
     */
   def isRetryable(failure: Throwable): Boolean = failure match {
-    case gcs: StorageException => gcs.isRetryable || AdditionalRetryablHttpCodes.contains(gcs.getCode)
+    case gcs: StorageException => gcs.isRetryable || AdditionalRetryablHttpCodes.contains(gcs.getCode) || isRetryable(gcs.getCause)
```
cjllanwarne: typo in `Retryabl`

mcovarr: regerttable

cjllanwarne:
The "sometimes this doesn't get retried" thing's a bit weird, but if this fixed it, 👍

engine/src/main/scala/cromwell/engine/io/IoActor.scala
geoffjentry: That one was *not* from a snickers commercial

PR #2213, opened 2017-04-27
JSON auth for Google SA

Horneth:


geoffjentry:
Is there a reason to not just forcibly switch to JSON other than backwards compat?

Horneth:
> Is there a reason to not just forcibly switch to JSON other than backwards compat?

Other than that no I don't see any

geoffjentry:
The reason I ask is that I believe the only reason we were doing the PEM file was that we had to, but it was viewed as a lesser scheme. If we can do JSON now I think we should just switch over. We're (DSDE) likely the only ones using PEM files.

mcovarr:
:+1: if we decide to support both

filesystems/gcs/src/main/scala/cromwell/filesystems/gcs/GoogleConfiguration.scala
```diff
@@ -76,7 +77,12 @@ object GoogleConfiguration {
     def buildAuth(authConfig: Config): ErrorOr[GoogleAuthMode] = {
 
       def serviceAccountAuth(authConfig: Config, name: String): ErrorOr[GoogleAuthMode] = validate {
-        ServiceAccountMode(name, authConfig.as[String]("service-account-id"), authConfig.as[String]("pem-file"), GoogleScopes)
+        (authConfig.getAs[String]("pem-file"), authConfig.getAs[String]("json-file")) match {
+          case (Some(pem), None) => ServiceAccountMode(name, PemFileFormat(authConfig.as[String]("service-account-id"), pem), GoogleScopes)
```
cjllanwarne: Can we throw out a "This will be deprecated soon" log warning here?

cjllanwarne:
In exchange for letting me mess up your centaur runs, I'll give you a thumb: 👍 

IMO it'd be nice if we can soft-deprecate the `pem` file format before disabling it

PR #2218, opened 2017-04-28
Remove unneeded SubWorkflowStoreCompleteSuccess message. Closes #2185

### LINKED ISSUE #2185: SubWorkflowStoreCompleteSuccess not needed
This is sent out by the `EmptySubWorkflowStoreActor` but but not waited for by the `SubWorkflowExecutionActor` (which indeed ignores it if it ever receives one).

This is awkward because it means we sometimes get "Message not delivered" error logs.

We should either remove this message entirely or, if it's used by tests, we should only send one out if the receiver is actually waiting for it.

cjllanwarne:
It turns out that "Complete Success" just wasn't quite right for Cromwell.

EDIT: the actual motivation for this is that this was the cause for some of our `dead letter` errors appearing at the end of `cromwell run` commands.


mcovarr:
:+1:

Horneth:
👍

PR #2224, opened 2017-04-28
Batching Job Store simpleton writes.  Closes #2085

### LINKED ISSUE #2085: JobStoreWriterActor can become a bottleneck when lots of jobs are finishing at the same time
When a lot of jobs are finishing at the same time and all try to write their result to the jobstore, it can become a bottleneck.
A possible solution (or improvement) would be to batch those requests to the DB.

mcovarr:
Makes a common `BatchingDbWriter` object to hold types for batched writers.  For the time being at least I've refrained from a deeper refactor to pull commonality up to an abstract parent FSM, though that does mean a fair amount of repetition in the implementation classes.

### DIFF
--- a/core/src/main/scala/cromwell/core/actor/BatchingDbWriter.scala
+++ b/core/src/main/scala/cromwell/core/actor/BatchingDbWriter.scala
@@ -1,0 +1,51 @@
+package cromwell.core.actor
+
+import cats.data.NonEmptyVector
+import org.slf4j.LoggerFactory
+
+import scala.util.{Failure, Success, Try}
+
+
+/** A collection of state, data, and message types to support batched database writes. */
+object BatchingDbWriter {
+  val logger = LoggerFactory.getLogger("BatchingDbWriteActor")
+
+  /** Data for batched database writes. */
+  sealed trait BatchingDbWriterData {
+    def addData[D](datum: D): BatchingDbWriterData = addData(Vector(datum))
+    def addData[D](data: Iterable[D]): BatchingDbWriterData = {
+      Try(NonEmptyVector.fromVector(data.toVector)) match {
+        case Success(Some(v)) =>
+          val newEvents = this match {
+            case NoData => v
+            case HasData(e) => e.concatNev(v)
+          }
+          HasData(newEvents)
+        case Success(None) => this
+        case Failure(f) =>
+          val dataSample = data.take(3).mkString(", ") + (if (data.size > 3) ", ..." else "")
+          logger.error(s"Failed processing batched data: $dataSample. Data will be dropped and not be sent to the database.", f)
+          this
+      }
+    }
+
+    def length: Int = this match {
+      case NoData => 0
+      case HasData(e) => e.length
+    }
+  }
+
+  case object NoData extends BatchingDbWriterData
+  case class HasData[E](events: NonEmptyVector[E]) extends BatchingDbWriterData
+
+  /** The states for batched database writes. */
+  sealed trait BatchingDbWriterState
+  case object WaitingToWrite extends BatchingDbWriterState
+  case object WritingToDb extends BatchingDbWriterState
+
+  /** The message types for batched database writes. */
+  sealed trait BatchingDbWriterMessage
+  case object DbWriteComplete extends BatchingDbWriterMessage
+  case object FlushBatchToDb extends BatchingDbWriterMessage
+  case object ScheduledFlushToDb extends BatchingDbWriterMessage
+}
--- a/core/src/main/scala/cromwell/core/path/EvenBetterPathMethods.scala
+++ b/core/src/main/scala/cromwell/core/path/EvenBetterPathMethods.scala
@@ -38,7 +38,7 @@ trait EvenBetterPathMethods {
 
   // betterFile.symbolicLink calls Files.readSymbolicLink, but then implicitly converts the java.nio.Path returned to a better.File
   // which calls toAbsolutePath. Consequently, if the path was relative, the current directory is used to make it absolute.
-  // This is not the desired behaviour to be able to follow relative symbolic links, so bypass better files method and directly use the java one.
+  // This is not the desired behavior to be able to follow relative symbolic links, so bypass better files method and directly use the java one.
   final def symbolicLinkRelative: Option[Path] = {
     if (betterFile.isSymbolicLink) {
       Option(newPath(Files.readSymbolicLink(betterFile.path)))
--- a/database/sql/src/main/scala/cromwell/database/slick/JobStoreSlickDatabase.scala
+++ b/database/sql/src/main/scala/cromwell/database/slick/JobStoreSlickDatabase.scala
@@ -1,30 +1,35 @@
 package cromwell.database.slick
 
-import cats.instances.future._
-import cats.syntax.functor._
 import cromwell.database.sql.JobStoreSqlDatabase
 import cromwell.database.sql.joins.JobStoreJoin
+import cromwell.database.sql.tables.JobStoreSimpletonEntry
 
 import scala.concurrent.{ExecutionContext, Future}
-import scala.language.postfixOps
 
 trait JobStoreSlickDatabase extends JobStoreSqlDatabase {
   this: SlickDatabase =>
 
   import dataAccess.driver.api._
 
-  private def addJobStore(jobStoreJoin: JobStoreJoin)(implicit ec: ExecutionContext): DBIO[Unit] = {
-    for {
-      jobStoreEntryId <- dataAccess.jobStoreEntryIdsAutoInc += jobStoreJoin.jobStoreEntry
-      _ <- dataAccess.jobStoreSimpletonEntryIdsAutoInc ++=
-        jobStoreJoin.jobStoreSimpletonEntries.map(_.copy(jobStoreEntryId = Option(jobStoreEntryId)))
-    } yield ()
-  }
-
-  override def addJobStores(jobStoreJoins: Seq[JobStoreJoin])
+  override def addJobStores(jobStoreJoins: Seq[JobStoreJoin], batchSize: Int)
                            (implicit ec: ExecutionContext): Future[Unit] = {
-    val action = DBIO.sequence(jobStoreJoins map addJobStore)
-    runTransaction(action) void
+
+    def assignJobStoreIdsToSimpletons(jobStoreIds: Seq[Int]): Seq[JobStoreSimpletonEntry] = {
+      val simpletonsByJobStoreEntry = jobStoreJoins map { _.jobStoreSimpletonEntries }
+      val jobStoreIdsAndSimpletons = jobStoreIds.toList zip simpletonsByJobStoreEntry
+      jobStoreIdsAndSimpletons flatMap { case (id, simpletons) => simpletons.map(_.copy(jobStoreEntryId = Option(id))) }
+    }
+
+    val jobStoreEntries = jobStoreJoins map { _.jobStoreEntry }
+
+    val action = for {
+      jobStoreIds <- dataAccess.jobStoreEntryIdsAutoInc ++= jobStoreEntries.toList
+      simpletons = assignJobStoreIdsToSimpletons(jobStoreIds)
+      simpletonInserts = simpletons.grouped(batchSize) map { dataAccess.jobStoreSimpletonEntries ++= _ }
+      _ <- DBIO.sequence(simpletonInserts)
+    } yield ()
+
+    runTransaction(action)
   }
 
   override def queryJobStores(workflowExecutionUuid: String, callFqn: String, jobScatterIndex: Int,
--- a/engine/src/main/scala/cromwell/jobstore/JobStoreWriterActor.scala
+++ b/engine/src/main/scala/cromwell/jobstore/JobStoreWriterActor.scala
@@ -1,96 +1,81 @@
 package cromwell.jobstore
 
 import akka.actor.{ActorRef, LoggingFSM, Props}
-import cromwell.jobstore.JobStoreActor._
 import cromwell.core.Dispatcher.EngineDispatcher
+import cromwell.core.actor.BatchingDbWriter
+import cromwell.core.actor.BatchingDbWriter._
+import cromwell.jobstore.JobStore.{JobCompletion, WorkflowCompletion}
+import cromwell.jobstore.JobStoreActor._
+import cromwell.jobstore.JobStoreWriterActor._
 
+import scala.concurrent.duration._
+import scala.language.postfixOps
 import scala.util.{Failure, Success}
 
-/**
-  * Singleton actor to coordinate writing job statuses to the database.
-  *
-  * State: Represents an actor either doing nothing, or currently writing to the database
-  * Data: If currently writing, the actor stores pending updates in the data. When one write completes, any further writes are written
-  */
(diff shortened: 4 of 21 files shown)


database/sql/src/main/scala/cromwell/database/slick/JobStoreSlickDatabase.scala
```diff
@@ -1,30 +1,35 @@
...
+    def assignJobStoreIdsToSimpletons(jobStoreIds: Seq[Int]): Seq[JobStoreSimpletonEntry] = {
+      val simpletonsByJobStoreEntry = jobStoreJoins map { _.jobStoreSimpletonEntries }
+      val jobStoreIdsAndSimpletons = jobStoreIds.toList zip simpletonsByJobStoreEntry
+      jobStoreIdsAndSimpletons flatMap { case (id, simpletons) => simpletons.map(_.copy(jobStoreEntryId = Option(id))) }
+    }
+
+    val jobStoreEntries = jobStoreJoins map { _.jobStoreEntry }
+
+    val action = for {
+      jobStoreIds <- dataAccess.jobStoreEntryIdsAutoInc ++= jobStoreEntries.toList
```
mcovarr: As mentioned in the comments below, the bulk `++=` for job store entries works at the Slick API level, but inspecting the SQL shows that inserts aren't actually done in bulk.  This is presumably due to using the `jobStoreEntryIdsAutoInc` instead of the `TableQuery`, but it appears we require the former in order to get the job store IDs to assign to the job store simpletons.

core/src/main/scala/cromwell/core/path/EvenBetterPathMethods.scala
```diff
@@ -38,7 +38,7 @@ trait EvenBetterPathMethods {
 
   // betterFile.symbolicLink calls Files.readSymbolicLink, but then implicitly converts the java.nio.Path returned to a better.File
   // which calls toAbsolutePath. Consequently, if the path was relative, the current directory is used to make it absolute.
-  // This is not the desired behaviour to be able to follow relative symbolic links, so bypass better files method and directly use the java one.
+  // This is not the desired behavior to be able to follow relative symbolic links, so bypass better files method and directly use the java one.
```
Horneth: Seeing this makes me think that it should be fixed in the latest better files version. We could switch back to using `betterFile.symbolicLink`

engine/src/main/scala/cromwell/jobstore/JobStoreWriterActor.scala
```diff
@@ -1,96 +1,87 @@
...
-  when(WritingToDatabase) {
-    case Event(command: JobStoreWriterCommand, stateData) =>
-      stay using stateData.withNewOperation(sender, command)
-    case Event(WriteComplete, stateData) =>
-      val newData = writeNextOperationToDatabase(stateData)
-      goto(if (newData.isEmpty) Pending else WritingToDatabase) using newData
+  when(WaitingToWrite) {
+    case Event(command: JobStoreWriterCommand, curData) =>
+      curData.addData(CommandAndReplyTo(command, sender)) match {
+        case newData: HasData[_] if newData.length > dbBatchSize => goto(WritingToDb) using newData
```
Horneth: Nitpicking but should it be `>=` ?

engine/src/main/scala/cromwell/jobstore/JobStoreWriterActor.scala
```diff
@@ -1,96 +1,87 @@
...
+      log.debug("Initiating periodic job store flush to DB")
+      goto(WritingToDb) using curData
   }
 
-  whenUnhandled {
-    case Event(someMessage, stateData) =>
-      log.error(s"JobStoreWriter: Unexpected message received in state $stateName: $someMessage")
-      stay()
+  when(WritingToDb) {
+    case Event(ScheduledFlushToDb, curData) => stay using curData
```
Horneth: no need for `using curData` I believe if we don't modify it

engine/src/main/scala/cromwell/jobstore/JobStoreWriterActor.scala
```diff
@@ -1,96 +1,87 @@
...
+      log.debug("Attempted job store flush to DB but had nothing to write")
+      goto(WaitingToWrite) using NoData
+    case Event(FlushBatchToDb, HasData(data)) =>
+      log.debug("Flushing {} job store commands to the DB", data.length)
+      val completions = data.toVector.collect({ case CommandAndReplyTo(c, _) => c.completion })
+
+      if (completions.nonEmpty) {
+        val workflowCompletions = completions collect { case w: WorkflowCompletion => w }
+        val completedWorkflowIds = workflowCompletions map { _.workflowId } toSet
+        val jobCompletions = completions.toList collect { case j: JobCompletion if !completedWorkflowIds.contains(j.key.workflowId) => j }
```
Horneth: ToL: I see why but maybe a little comment to indicate why we filter out job completions that also have a workflow completion ?

engine/src/main/scala/cromwell/jobstore/JobStoreWriterActor.scala
```diff
@@ -1,96 +1,87 @@
...
+
+        jsd.writeToDatabase(workflowCompletions, jobCompletions, dbBatchSize) onComplete {
+          case Success(_) =>
+            data map { case CommandAndReplyTo(c, r) => r ! JobStoreWriteSuccess(c) }
+            self ! DbWriteComplete
+          case Failure(regerts) =>
+            log.error("Failed to properly flush metadata to database", regerts)
```
Horneth: flush "job store entries" rather ?

database/sql/src/main/scala/cromwell/database/slick/JobStoreSlickDatabase.scala
Horneth: I imagine the answer is yes but just to double check, is it guaranteed that the Ids coming back are in the same order as their respective entry in `jobStoreEntries` ?

Horneth:
👍

core/src/main/scala/cromwell/core/path/EvenBetterPathMethods.scala
cjllanwarne: 🇬🇧 😢

database/sql/src/main/scala/cromwell/database/slick/JobStoreSlickDatabase.scala
mcovarr: Yes, that's what I've found.

engine/src/main/scala/cromwell/jobstore/JobStoreWriterActor.scala
mcovarr: yup

engine/src/main/scala/cromwell/jobstore/JobStoreWriterActor.scala
mcovarr: yup

engine/src/main/scala/cromwell/jobstore/JobStoreWriterActor.scala
mcovarr: sure

engine/src/main/scala/cromwell/jobstore/JobStoreWriterActor.scala
mcovarr: oops 😦

core/src/main/scala/cromwell/core/path/EvenBetterPathMethods.scala
mcovarr: 🇺🇸  🎆

engine/src/main/scala/cromwell/jobstore/JobStoreWriterActor.scala
```diff
@@ -1,96 +1,89 @@
...
-    case Event(someMessage, stateData) =>
-      log.error(s"JobStoreWriter: Unexpected message received in state $stateName: $someMessage")
-      stay()
+  when(WritingToDb) {
+    case Event(ScheduledFlushToDb, curData) => stay
+    case Event(command: JobStoreWriterCommand, curData) => stay using curData.addData(CommandAndReplyTo(command, sender))
+    case Event(FlushBatchToDb, NoData) =>
+      log.debug("Attempted job store flush to DB but had nothing to write")
+      goto(WaitingToWrite)
+    case Event(FlushBatchToDb, HasData(data)) =>
```
cjllanwarne: ToL: If we're going to block workflow completion behind job store writing, we want to make sure this batch rate is fast enough, otherwise we're going to have annoying wait times at the end of our workflows while we wait for the job store writing batch to happen

engine/src/main/scala/cromwell/jobstore/JobStoreWriterActor.scala
```diff
@@ -1,96 +1,89 @@
...
-            val msg = JobStoreWriteSuccess(message)
-            actor ! msg
-          }
-          self ! WriteComplete
-        case Failure(reason) =>
-          log.error(s"Failed to write to database: $reason")
-          newData.currentOperation foreach { case (actor, message) => actor ! JobStoreWriteFailure(reason) }
-          self ! WriteComplete
-      }
+  implicit class FancyJobStoreWriterCommand(val command: JobStoreWriterCommand) extends AnyVal {
```
cjllanwarne: Any reason for not just adding this directly to `JobStoreWriterCommand`?

engine/src/test/scala/cromwell/jobstore/JobStoreWriterSpec.scala
```diff
@@ -52,26 +57,26 @@ class JobStoreWriterSpec extends CromwellTestKitWordSpec with Matchers with Befo
...
 
   "JobStoreWriter" should {
     "be able to collapse writes together if they arrive while a database access is ongoing" in {
 
-      sendRegisterCompletions(attempts = 3)
-
+      // Send a job completion. The database will hang.
+      sendRegisterCompletion(1)
+      // Wait long enough to know that the writer should be trying to write.
+      Thread.sleep(sleepMillis)
```
cjllanwarne: I really don't like "just wait a long time, to make sure the concurrency will have happened" in unit tests.

Could you instead set the polling interval to effectively infinite, and instead send the internal control commands yourself?

engine/src/test/scala/cromwell/jobstore/JobStoreWriterSpec.scala
```diff
@@ -82,20 +87,22 @@ class JobStoreWriterSpec extends CromwellTestKitWordSpec with Matchers with Befo
 
     "be able to skip job-completion writes if the workflow completes, but still respond appropriately" in {
 
-      sendRegisterCompletions(attempts = 2)
-      jobStoreWriter ! RegisterWorkflowCompleted(workflowId)
-
+      // Send a job completion. The database will hang.
+      sendRegisterCompletion(1)
+      // Wait long enough to know that the writer should be trying to write.
+      Thread.sleep(sleepMillis)
```
cjllanwarne: Again.

engine/src/test/scala/cromwell/jobstore/JobStoreWriterSpec.scala
cjllanwarne: Alternatively, explicitly for the internal state to catch up before continuing:
```
eventually {
  writer.stateName shouldBe WritingToDb
}
```

core/src/main/scala/cromwell/core/actor/BatchingDbWriter.scala
```diff
@@ -0,0 +1,50 @@
...
+      Try(NonEmptyVector.fromVector(data.toVector)) match {
+        case Success(Some(v)) =>
+          val newEvents = this match {
+            case NoData => v
+            case HasData(e) => e.concatNev(v)
+          }
+          HasData(newEvents)
+        case Success(None) => this
+        case Failure(f) =>
+          logger.error("Failed processing batched data. Data will be dropped and not be sent to the database.", f)
```
mcovarr: Not easily, but I could at least dump the type of the data which should help.

engine/src/main/scala/cromwell/jobstore/JobStoreWriterActor.scala
mcovarr: Yes that's true.  Right now it's set to 1 second, and currently there isn't a config knob exposed.

engine/src/main/scala/cromwell/jobstore/JobStoreWriterActor.scala
mcovarr: I can do that if you prefer

core/src/main/scala/cromwell/core/actor/BatchingDbWriter.scala
cjllanwarne: I can imagine someone coming to office hours or the WDL forum saying "my cromwell says its dropped some data. Should I be concerned, worried, or terrified?". It'd be nice to have a better answer than :man_shrugging:

engine/src/main/scala/cromwell/jobstore/JobStoreWriterActor.scala
cjllanwarne: 1 second seems appropriately responsive. I'd be more worried if it was ~10 minutes

engine/src/main/scala/cromwell/jobstore/JobStoreWriterActor.scala
cjllanwarne: Up to you. Just struck me as odd to be doing Fancy implicit shenanigans when you already have access to the original class

mcovarr: I wrote it up and it's a little awkward as there's an abstract member in the trait and then overrides as the only functions in the concrete types.

core/src/main/scala/cromwell/core/actor/BatchingDbWriter.scala
mcovarr: Yeah it's a good point.  I think the answer is probably in the worried-terrified range, but we hopefully a sampling of the unwritten data can help give form to that fear.

core/src/main/scala/cromwell/core/actor/BatchingDbWriter.scala
```diff
@@ -0,0 +1,51 @@
...
+      Try(NonEmptyVector.fromVector(data.toVector)) match {
+        case Success(Some(v)) =>
+          val newEvents = this match {
+            case NoData => v
+            case HasData(e) => e.concatNev(v)
+          }
+          HasData(newEvents)
+        case Success(None) => this
+        case Failure(f) =>
+          val dataSample = data.mkString(", ") + (if (data.size > 3) ", ..." else "")
```
cjllanwarne: Am I missing something? This doesn't look like it's truncating `data` for the `mkString`, but your `...` makes it look like it should be?

cjllanwarne:
Assuming I did just miss something, 👍

core/src/main/scala/cromwell/core/actor/BatchingDbWriter.scala
mcovarr: You're not missing anything.  I had this right the first time then "refactored" it.  😦  Good catch!

PR #2239, opened 2017-05-03
Use the right failureMode value

Horneth:
This would work if the failure mode is passed through workflow options but not if it's set in the config.


mcovarr:
:+1: oops

cjllanwarne:
👍 brava

PR #2241, opened 2017-05-03
file paths were being stringified and causing jobs running... (25_hotfix edition)

jsotobroad:
…ration to become broken after the migration was complete


mcovarr:
:+1: oops

geoffjentry:
:+1:

mcovarr:
There should be a develop version too

PR #2242, opened 2017-05-03
JobStoreSimpletonMigration was putting quotation marks... (develop edition)

jsotobroad:
…lues which caused the jobs to fail once cromwell recovered after a migration. For Develop


mcovarr:
:+1:

cjllanwarne:
👍

PR #2243, opened 2017-05-03
Call Caching v3

Horneth:


engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
```diff
@@ -80,7 +82,14 @@ final case class ExecutionStore(private val statusStore: Map[JobKey, ExecutionSt
...
   }
 
-  def runnableScopes = keysWithStatus(NotStarted) filter arePrerequisitesDone
+  /**
+    * Returns the list of jobs ready to be run, along with a Boolean indicating whether or not the list has been truncated.
+    * The size of the list will be MaxJobsToStartPerTick at most. If more jobs where found runnable, the boolean will be true, otherwise false.
+    */
+  def runnableScopes: (List[JobKey], Boolean) = {
+    val readyToStart = keysWithStatus(NotStarted).toStream filter arePrerequisitesDone
+    (readyToStart.take(ExecutionStore.MaxJobsToStartPerTick).toList, readyToStart.size > ExecutionStore.MaxJobsToStartPerTick)
```
Horneth: Note to self: fix this: readyToStart.size evaluates the whole stream which negates the reason of making it a stream entirely.

cjllanwarne:
@Horneth could you give a quick big-picture synopsis of what this PR is?

Horneth:
@cjllanwarne Yep, it would be easier to do on a white board I was thinking of tech talking it maybe if that's ok ?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheHashingJobActor.scala
```diff
@@ -0,0 +1,267 @@
...
+    * Aggregates hash results together in a predictable and reproducible manner.
+    * Purposefully takes an Iterable, because it will be sorted appropriately in this method
+    * to ensure deterministic result, so the type of collection doesn't matter.
+    * Only aggregates hash values for which checkForHitOrMiss is true.
+    * If several hash keys are identical, the result of this method is undefined.
+    */
+  private def calculateHashAggregation(hashes: Iterable[HashResult], messageDigest: MessageDigest) = {
+    val sortedHashes = hashes.toList.filter(_.hashKey.checkForHitOrMiss).sortBy(_.hashKey.key).map(_.hashValue.value.getBytes)
+    sortedHashes foreach messageDigest.update
+    DatatypeConverter.printHexBinary(messageDigest.digest())
```
mcovarr: This is consistent with the SQL migration?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheHashingJobActor.scala
```diff
@@ -0,0 +1,267 @@
...
+/**
+  * Actor responsible for calculating individual as well as aggregated hashes for a job.
+  * First calculate the initial hashes (individual and aggregated), and send them to its parent
+  * as well as the callCacheReadingJobActor if one was provided.
+  * From there, either wait for a NextBatchOfFileHashesRequest message from the callCacheReadingJobActor to trigger the next
+  * batch of file hash requests, or send it itself if there is no callCacheReadingJobActor.
+  * When we get all the file hashes for one batch, send those hashes to the CCRead actor and wait for the next NextBatchOfFileHashesRequest.
+  * If it was the last batch and all the files have been hashed, send all the hashes along with the aggregated file hash.
+  * If at any point the callCacheReadingJobActor dies (either because it decided this was a CacheMiss or it found a successful
+  * CacheHit), either keep hashing the files if writeToCache is true, or die if it's not.
```
mcovarr: so CCRJA is communicating with a different actor (EJEA?) for hit / miss status?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheHashingJobActor.scala
```diff
@@ -0,0 +1,267 @@
...
+  }
+
+  final case class CallCacheHashingJobActorData(
+                                           fileHashRequestsRemaining: List[List[SingleFileHashRequest]],
+                                           fileHashResults: List[HashResult],
+                                           callCacheReadingJobActor: Option[ActorRef]
+                                         ) {
+    private val md5Digest = MessageDigest.getInstance("MD5")
+
+    def withFileHash(hashResult: HashResult): (CallCacheHashingJobActorData, Option[CCHJAFileHashResponse]) = {
```
mcovarr: type?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheReadingJobActor.scala
```diff
@@ -0,0 +1,101 @@
...
+      callCacheReadActor ! CacheLookupRequest(AggregatedCallHashes(data.initialHash, aggregatedFileHash), data.currentHitNumber)
+      goto(WaitingForCacheHitOrMiss) using data.withFileHash(aggregatedFileHash)
+    case Event(NoFileHashesResult, data: CCRJAWithData) =>
+      callCacheReadActor ! CacheLookupRequest(AggregatedCallHashes(data.initialHash, None), data.currentHitNumber)
+      goto(WaitingForCacheHitOrMiss)
+  }
+  
+  when(WaitingForCacheHitOrMiss) {
+    case Event(CacheLookupNextHit(hit), data: CCRJAWithData) =>
+      context.parent ! CacheHit(hit)
```
mcovarr: Ok, this answers my question about the CCRJA communicating with its parent.  😄  Wouldn't hurt to drop a comment back in the CCHJA, this is all kind of a lot to take in.

mcovarr:
:+1: as I mentioned in standup, I'm happy to rebase this if #2244 merges first.

CHANGELOG.md
```diff
@@ -2,6 +2,25 @@
...
 ## 27
 
+### Migration
+
+* Call Caching has been improved in this version of Cromwell, specifically the time needed to determine whether or not a job can be cached
+ has drastically decreased. To achieve that the database schema has been modified and a migration is required in order to preserve the pre-existing cached jobs.
+ This migration is relatively fast compared to previous migrations. To get an idea of the time needed, look at the size of your `CALL_CACHING_HASH_ENTRY` table.
+ As a landmark, it takes 1 minute for a table with 6 million rows.
+ The migration will only be executed on MySQL. Other databases will lose their previous cached jobs.
+ In order to run properly on MySQL, the following flag needs to be adjusted: https://dev.mysql.com/doc/refman/5.5/en/server-system-variables.html#sysvar_group_concat_max_len
```
Horneth: The SQL query fails so the migration fails

Horneth: Provided that you have a call for which the concatenated hash values is greater than the default value for the flag which is 1024

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheHashingJobActor.scala
```diff
@@ -0,0 +1,267 @@
...
+                               writeToCache: Boolean
+                              ) extends LoggingFSM[CallCacheHashingJobActorState, CallCacheHashingJobActorData] {
+
+  val fileHashingActor = makeFileHashingActor()
+
+  // Watch the read actor, as it will die when it's done (cache miss or successful cache hit)
+  // When that happens we want to either stop if writeToCache is false, or keep going
+  callCacheReadingJobActor map context.watch
+
+  initializeCCHJA()
```
Horneth: yeah that's what I thought too but `initialize` is an `FSM` public method so akka isn't happy if you create your own.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheHashingJobActor.scala
Horneth: Yes at least to the extent of the testing I've done they yield the same.
I ran this on the 10K mock wdl database after the migration and it call cached immediately.
`sortedHashes foreach messageDigest.update` is equivalent to md5ing the concatenation.
`DatatypeConverter.printHexBinary` just prints the bytes in hexadecimal string format.

CHANGELOG.md
mcovarr: How long does this query take to run?  Is it something we could do in a precondition with a custom `onFailMessage` in the off chance that the user hasn't carefully read the CHANGELOG.md before migrating?  😄

Horneth: Oh yeah that's a good idea ! There might also be another way to do this that doesn't involve GROUP_CONCAT and doesn't need to adjust the flag but my SQL is not powerful enough..

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheHashingJobActor.scala
mcovarr: 😢

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheHashingJobActorSpec.scala
```diff
@@ -0,0 +1,292 @@
...
+
+    fileHashingActor.expectMsg(fileHashRequest1)
+    fileHashingActor.expectMsg(fileHashRequest2)
+    cchja.stateName shouldBe HashingFiles
+  }
+
+  it should "send NoFileHashesResult and stop if there are no input files" in {
+    val parent = TestProbe()
+    val callCacheReadProbe = TestProbe()
+    // self acts as the parent
```
Horneth: Note to self: remove those comments they don't apply anymore as we have explicit parent `TestProbe`s now

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheHashingJobActor.scala
Horneth: It seems a bit superfluous for this one as this method is only ever used once inside this actor, and hence its result parsed once ? Also the name would be `NewDataAndMaybeResponse` :/
What about a scaladoc that says what the pair corresponds to ?

mcovarr:
Rebased and a minor migration fix added.  I also removed the non-default failure and error handling on the migration but did not add in a special error message.

Horneth:
@mcovarr thanks for rebasing !

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
```diff
@@ -80,7 +84,18 @@ final case class ExecutionStore(private val statusStore: Map[JobKey, ExecutionSt
...
+    * Returns the list of jobs ready to be run, along with a Boolean indicating whether or not the list has been truncated.
+    * The size of the list will be MaxJobsToStartPerTick at most. If more jobs where found runnable, the boolean will be true, otherwise false.
+    */
+  def runnableScopes: RunnableScopes = {
+    val readyToStart = keysWithStatus(NotStarted).toStream filter arePrerequisitesDone
+    // Compute the first ExecutionStore.MaxJobsToStartPerTick + 1 runnable scopes
+    val scopesToStartPlusOne = readyToStart.take(ExecutionStore.MaxJobsToStartPerTick + 1).toList
+    // Only take the first ExecutionStore.MaxJobsToStartPerTick from the above list.
+    // Use the fact that we took one more to determine whether or not we truncated the result.
+    RunnableScopes(scopesToStartPlusOne.take(ExecutionStore.MaxJobsToStartPerTick), scopesToStartPlusOne.size > ExecutionStore.MaxJobsToStartPerTick)
```
cjllanwarne: I don't get this indirection with `scopesToStartPlusOne`. Why not just do:
```
RunnableScopes(readyToStart.take(ExecutionStore.MaxJobsToStartPerTick), 
     readyToStart.size > ExecutionStore.MaxJobsToStartPerTick)
```

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCache.scala
```diff
@@ -62,20 +69,31 @@ class CallCache(database: CallCachingSqlDatabase) {
...
-  def callCachingEntryIdsMatchingHashes(callCacheHashes: CallCacheHashes)(implicit ec: ExecutionContext): Future[Set[CallCachingEntryId]] = {
-    callCachingEntryIdsMatchingHashes(NonEmptyList.fromListUnsafe(callCacheHashes.hashes.toList))
+  def hasBaseAggregatedHashMatch(baseAggregatedHash: String)(implicit ec: ExecutionContext): Future[Boolean] = {
+    database.hasMatchingCallCachingEntriesForBaseAggregation(baseAggregatedHash)
+  }
+  
+  def hasKeyValuePairHashMatch(hashes: Set[HashResult])(implicit ec: ExecutionContext): Future[Boolean] = {
+    hashes.toList match {
+        // Empty hashes should match nothing
+      case Nil => Future.successful(false)
```
cjllanwarne: Really? If a task has no input files then we want it to match anything else with no input files

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheHashingJobActor.scala
cjllanwarne: British English to the rescue? `initialise` 🇬🇧

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheHashingJobActor.scala
```diff
@@ -0,0 +1,273 @@
...
+
+  override def preStart(): Unit = {
+    if (callCacheReadingJobActor.isEmpty && !writeToCache) {
+      log.error("Programmer error ! There is no reason to have a hashing actor if both read and write to cache are off")
+      context stop self
+    }
+    super.preStart()
```
cjllanwarne: Is this safe after a `context stop self`?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheHashingJobActor.scala
```diff
@@ -0,0 +1,273 @@
...
+      stopAndStay(None)
+    case Event(Terminated(actor), data) if writeToCache =>
+      if (stateName == WaitingForHashFileRequest) self ! NextBatchOfFileHashesRequest
+      stay() using data.copy(callCacheReadingJobActor = None)
+    case Event(error: HashingFailedMessage, data) =>
+      log.error(error.reason, s"Failed to hash ${error.file}")
+      sendToCCReadActor(error, data)
+      context.parent ! error
+      stopAndStay(None)
+  }
```
cjllanwarne: Any value in a catch-all `case Event(other, data) => log.error("Panic! $other $data")`

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheHashingJobActor.scala
```diff
@@ -0,0 +1,273 @@
...
+        case (newData, None) =>
+          stay() using newData
+      }
+  }
+
+  whenUnhandled {
+    case Event(Terminated(actor), data) if !writeToCache =>
+      stopAndStay(None)
+    case Event(Terminated(actor), data) if writeToCache =>
+      if (stateName == WaitingForHashFileRequest) self ! NextBatchOfFileHashesRequest
```
cjllanwarne: Hmm, would this be better handled in the `when(stateName)` methods, rather than with a `stateName`-branch in the `whenUnhandled`?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheHashingJobActor.scala
```diff
@@ -0,0 +1,273 @@
...
+
+  /**
+    * Aggregates hash results together in a predictable and reproducible manner.
+    * Purposefully takes an Iterable, because it will be sorted appropriately in this method
+    * to ensure deterministic result, so the type of collection doesn't matter.
+    * Only aggregates hash values for which checkForHitOrMiss is true.
+    * If several hash keys are identical, the result of this method is undefined.
+    */
+  private def calculateHashAggregation(hashes: Iterable[HashResult], messageDigest: MessageDigest) = {
+    val sortedHashes = hashes.toList.filter(_.hashKey.checkForHitOrMiss).sortBy(_.hashKey.key).map(_.hashValue.value.getBytes)
```
cjllanwarne: As above, I think it's important to make sure the keys are part of the hash, as well as the values

cjllanwarne: I assume that `checkForHitOrMiss` is the same as the set of runtime attributes that you specified in the migration query?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/EngineJobHashingActor.scala
```diff
@@ -1,282 +1,126 @@
...
+
+  override def receive = {
+    case initialHashResult: InitialHashingResult => initialHash = Option(initialHashResult)
+    case finalFileHashResult: FinalFileHashingResult => sendHashes(finalFileHashResult)
+    case CacheMiss => receiver ! CacheMiss
+    case hit: CacheHit => receiver ! hit
+    case NextHit =>
+      callCacheReadingJobActor match {
+        case Some(readActor) =>
+          readActor ! NextHit
```
cjllanwarne: Is there a race condition where the `callCacheReadingJobActor` shuts down while this `NextHit` message is being processed?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/EngineJobHashingActor.scala
```diff
@@ -1,282 +1,126 @@
...
+        case None =>
+          jobLogger.error("Requested next hit but there's no Call Caching Read Actor !")
```
cjllanwarne: Erroneous space before !

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheHashingJobActorDataSpec.scala
```diff
@@ -0,0 +1,94 @@
...
+import org.scalatest.prop.TableDrivenPropertyChecks
+import org.scalatest.{FlatSpecLike, Matchers}
+
+class CallCacheHashingJobActorDataSpec extends TestKitSuite with FlatSpecLike with BackendSpec with Matchers with Eventually with TableDrivenPropertyChecks {
+  behavior of "CallCacheReadingJobActorData"
+
+  val fileHash1 = HashResult(HashKey("key"), HashValue("value"))
+  val fileHash2 = HashResult(HashKey("key2"), HashValue("value2"))
+  val fileHash3 = HashResult(HashKey("key3"), HashValue("value3"))
+  val fileHashRequest1 = mock[SingleFileHashRequest]
```
cjllanwarne: `SingleFileHashRequest` should be a `final case class` and so unmockable. 

That's ok, because you don't need to mock a case class, since this is a test we can just use `SingleFileHashRequest(null, hashKey, null null)` and allow the NPE if we set up the test wrong.

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheHashingJobActorSpec.scala
```diff
@@ -0,0 +1,292 @@
...
+  it should "send a correct InitialHashingResult upon starting" in {
+    val parent = TestProbe()
+    val inputs = Map(
+      "stringInput" -> WdlString("hello"),
+      "fileInput" -> WdlFile("world")
+    )
+    // Do not include "memory" on purpose, even though it's in the map of runtime attributes.
+    // This way we can verify that only attributes with a RuntimeAttributeDefinition are used for hashing
+    // Vice versa include a "docker" definition even though it's not in the map.
+    // This ensures that we still record the fact that there was no docker attribute with a "N/A" value
```
cjllanwarne: Nice!

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheHashingJobActorSpec.scala
```diff
@@ -0,0 +1,292 @@
...
+  it should "keep going if the read actor dies and writeToCache is on" in {
+    val callCacheReadProbe = TestProbe()
+    val parent = TestProbe()
+    val cchja = makeCCHJA(Option(callCacheReadProbe.ref), TestProbe().ref, parent.ref)
+    parent.expectMsgClass(classOf[InitialHashingResult])
+    
+    cchja.setState(WaitingForHashFileRequest)
+    system stop callCacheReadProbe.ref
+
+    parent.expectMsg(NoFileHashesResult)
```
cjllanwarne: It'd be nice if we could have some file hashes here (to make sure it did indeed keep going)

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/EJHADataSpec.scala
```diff
@@ -1,96 +0,0 @@
...
-
-class EJHADataSpec extends FlatSpec with Matchers {
-
-  behavior of "EJHA data"
-
-  val hashKey1 = HashKey("Peter Piper picked a peck of pickled peppers")
-  val hashKey2 = HashKey("I saw Susie sitting in a shoe shine shop: Where she sits she shines, and where she shines she sits")
-  val hashKey3 = HashKey("How many boards could the Mongols hoard if the Mongol hordes got bored?")
-  val hashKey4 = HashKey("The sixth sick Sikh's sixth sheep is sick.")
-  val hashKey5 = HashKey("The Doge did what a Doge does, when a Doge does his duty to a Duke, that is. When the Doge did his duty and the Duke didn't, that's when the Duchess did the dirt to the Duke with the Doge. There they were in the dark: The Duke with his dagger, the Doge with his dart and the Duchess with her dirk. The Duchess dug at the Duke just when the Duke dove at the Doge. Now the Duke ducked, the Doge dodged, and the Duchess didn't. So the Duke got the Duchess, the Duchess got the Doge, and the Doge got the Duke.")
```
cjllanwarne: it's a sad day when this text is removed from the code base :'(

cjllanwarne: 😢

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
Horneth: because `readyToStart.size` would force the stream (as in scala collection stream) to evaluate entirely and cause `arePrerequisitesDone` to be called on all subsequent scopes even though the will be ignored since we only take the first `ExecutionStore.MaxJobsToStartPerTick` ones.

database/migration/src/main/resources/changesets/call_caching_aggregation/call_caching_base_aggregation.sql
```diff
@@ -0,0 +1,18 @@
+INSERT INTO CALL_CACHING_AGGREGATION_ENTRY(
+  CALL_CACHING_ENTRY_ID,
+  BASE_AGGREGATION
+)
+  SELECT cche.CALL_CACHING_ENTRY_ID,
+  UPPER(MD5(GROUP_CONCAT(
+                cche.HASH_VALUE ORDER BY cche.HASH_KEY ASC SEPARATOR ''
```
Horneth: Why ?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheHashingJobActor.scala
Horneth: Yes the migration query leaves out the runtime attributes that have `checkForHitOrMiss` false

database/sql/src/main/scala/cromwell/database/slick/CallCachingSlickDatabase.scala
```diff
@@ -19,8 +19,8 @@ trait CallCachingSlickDatabase extends CallCachingSqlDatabase {
                              (implicit ec: ExecutionContext): Future[Unit] = {
 
     // Construct parallel lists of parent entries, hashes, simpletons, and detritus from `CallCachingJoin`s.
-    val (entries, hashes, simpletons, detritus) = joins.toList.foldMap { j =>
-      (List(j.callCachingEntry), List(j.callCachingHashEntries), List(j.callCachingSimpletonEntries), List(j.callCachingDetritusEntries)) }
+    val (entries, hashes, simpletons, detritus, aggregations) = joins.toList.foldMap { j =>
+      (List(j.callCachingEntry), List(j.callCachingHashEntries), List(j.callCachingSimpletonEntries), List(j.callCachingDetritusEntries), List(j.callCachingAggregationEntry.toList)) }
```
danbills: Just checking:  Is the intention to create a List[List[]] of Entries?

mcovarr: (This change was part of my rebase)

It was the intent that `entries` would be a `List[CallCachingEntry]` and all the others would be `List[Seq[_]]` or `List[List[_]]` of parallel dependent entities.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCache.scala
Horneth: Oh right yes ! I guess technically this would never need to be called with an empty set but still..

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheHashingJobActor.scala
Horneth: Yeah I always wondered the reason for that. Akka already logs unhandled messages by default, is there a reason we want to do this ourselves instead ?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheHashingJobActor.scala
Horneth: I believe so but I'll check

database/migration/src/main/resources/changesets/call_caching_aggregation/call_caching_base_aggregation.sql
cjllanwarne: Imagine I have two `Array[File]` inputs, `a` and `b`. Without keys I think you would cache match the following two input sets:
```
a = [ file_1, file_2 ]
b = [ file_3, file_4 ]
```
and:
```
a = [ file_1 ]
b = [ file_2, file_3, file_4 ]
```

cjllanwarne: Or for the initial hash set, you could sub out `File`s for `String`s

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheHashingJobActor.scala
geoffjentry: FWIW I've never really liked our penchant for specially handling the unhandled case unless we were going to do something other than log. As you said, akka already does this.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/EngineJobHashingActor.scala
Horneth: I don't think so ? The `callCacheReadingJobActor` only shuts down if it determines it's a `CacheMiss` at which point the EJEA will stop asking for `NextHit`

ruchim:
Perhaps I missed it, but it seems like we are still publishing the cache hit results as soon as we get a hit, not after copying succeeds? Maybe it wasn't meant to be a part of this PR.

cjllanwarne:
@ruchim writing a cache result (its hashes and its results) should be functionally atomic (i.e. either you get a cache hit, or you don't, and if you do, then you can copy all the results immediately).
If that's not part of this PR then it should 100% be part of Cromwell 27 (IMO)

engine/src/main/scala/cromwell/engine/io/nio/NioFlow.scala
```diff
@@ -77,4 +80,9 @@ class NioFlow(parallelism: Int, scheduler: Scheduler, nbAttempts: Int = IoActor.
       )
     }
   }
+  
+  private def createDirectoriesForSFSPath(path: Path) = path match {
```
ruchim: Is there risk in moving this step directly to the copyTo/write methods?

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/EJHADataSpec.scala
ruchim: I'm sure you'll bring it back one day Chris :)

cjllanwarne:
👍

PR #2244, opened 2017-05-03
Batch writes of call caching related data.

mcovarr:
This will conflict with #2243 some but not too much.


engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -142,12 +142,13 @@ object WorkflowActor {
...
             subWorkflowStoreActor: ActorRef,
             callCacheReadActor: ActorRef,
+            callCacheWriteActor: ActorRef,
             dockerHashActor: ActorRef,
             jobTokenDispenserActor: ActorRef,
             backendSingletonCollection: BackendSingletonCollection,
             serverMode: Boolean): Props = {
     Props(new WorkflowActor(workflowId, startMode, wdlSource, conf, ioActor, serviceRegistryActor, workflowLogCopyRouter,
-      jobStoreActor, subWorkflowStoreActor, callCacheReadActor, dockerHashActor, jobTokenDispenserActor, backendSingletonCollection, serverMode)).withDispatcher(EngineDispatcher)
+      jobStoreActor, subWorkflowStoreActor, callCacheReadActor, callCacheWriteActor, dockerHashActor, jobTokenDispenserActor, backendSingletonCollection, serverMode)).withDispatcher(EngineDispatcher)
```
cjllanwarne: In other cases where we have a  long list of (essentially untyped) `ActorRef` arguments, I find it pretty much necessary to add the param names to avoid mixing up the order, e.g.
```
Props(new WorkflowActor(
 ...
  jobStoreActor = jobStoreActor,
  subWorkflowStoreActor = subWorkflowStoreActor,
  ...
```

engine/src/test/scala/cromwell/CromwellTestKitSpec.scala
```diff
@@ -479,10 +481,20 @@ class EmptyCallCacheReadActor extends Actor {
   }
 }
 
+class EmptyCallCacheWriteActor extends Actor {
```
cjllanwarne: ToL: Maybe `NoOpCallCacheWriteActor` is more accurate?

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/ejea/EjeaBackendIsCopyingCachedOutputsSpec.scala
```diff
@@ -53,7 +53,6 @@ class EjeaBackendIsCopyingCachedOutputsSpec extends EngineJobExecutionActorSpec
 
           // Nothing should happen here:
           helper.jobStoreProbe.expectNoMsg(awaitAlmostNothing)
-          helper.callCacheWriteActorCreations should be(NothingYet) // Rely on the await timeout from the previous step to allow time to pass
```
cjllanwarne: Why isn't this still true?

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/ejea/PerTestHelper.scala
```diff
@@ -137,6 +135,7 @@ private[ejea] class PerTestHelper(implicit val system: ActorSystem) extends Mock
       ioActor = ioActorProbe.ref,
       jobStoreActor = jobStoreProbe.ref,
       callCacheReadActor = callCacheReadActorProbe.ref,
+      callCacheWriteActor = callCacheWriteActorProbe.ref,
```
cjllanwarne: This probe never seems to be inspected? Do we want to verify that it receives the appropriate messages (and the actors respond correctly to its messages)?

cjllanwarne:
👍 assuming CCv3 compliance

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
mcovarr: mokay

engine/src/test/scala/cromwell/CromwellTestKitSpec.scala
mcovarr: This is consistent with the existing nomenclature here

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/ejea/EjeaBackendIsCopyingCachedOutputsSpec.scala
mcovarr: These aren't created per job anymore

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/ejea/PerTestHelper.scala
mcovarr: Theoretically that sounds good, though I'm not sure of the PITA factor

Horneth:
👍

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/ejea/EjeaBackendIsCopyingCachedOutputsSpec.scala
cjllanwarne: Good point!

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/ejea/EngineJobExecutionActorSpecUtil.scala
```diff
@@ -26,12 +26,6 @@ private[ejea] trait CanExpectCacheWrites extends Eventually { self: EngineJobExe
   def expectCacheWrite(expectedResponse: JobSucceededResponse, expectedCallCacheHashes: CallCacheHashes): Unit = {
     eventually { ejea.stateName should be(UpdatingCallCache) }
     ejea.stateData should be(SucceededResponseData(expectedResponse, Some(Success(expectedCallCacheHashes))))
-    helper.callCacheWriteActorCreations match {
-      case GotOne(creation) =>
```
cjllanwarne: I'm not 100% sure but I wouldn't be surprised if `GotOne` was only ever used here

mcovarr: It's used in two other spots actually

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/ejea/PerTestHelper.scala
mcovarr: Actually I think I'm going to punt on this.  All the previous logic was doing was testing the creation of the actor, which is now done up in the CromwellRootActor.

PR #2257, opened 2017-05-11
Allow customLabels submission to Cromwell Client

ruchim:


cromwellApiClient/src/main/scala/cromwell/api/model/WorkflowSubmission.scala
```diff
@@ -6,19 +6,22 @@ sealed trait WorkflowSubmission {
...
   val options: Option[String]
+  val customLabels: Option[String]
   val zippedImports: Option[File]
   val refreshToken: Option[String]
 }
 
 final case class WorkflowSingleSubmission(wdl: String,
                                           inputsJson: Option[String],
                                           options: Option[String],
+                                          customLabels: Option[String],
```
cjllanwarne: ToL: from an API point of view it'd be cool if this was a `List[Label]` if somewhere we can have `case class Label(key: String, value: String)`. 
Thinking about it, we'd also need a custom JsonFormat for the label type that can serialize it into the appropriate JSON `"key": "value"`

cjllanwarne:
Other than my pie-in-the-sky "wouldn't it be nice" comment about the API's usability, looks good 👍

Horneth:
👍

ruchim:
@cjllanwarne can you please re-review the latest commit, I hope I'm interpreting your dreams correctly.

cromwellApiClient/src/main/scala/cromwell/api/model/Label.scala
```diff
@@ -0,0 +1,10 @@
+package cromwell.api.model
+
+import spray.json.DefaultJsonProtocol
+
+object CustomLabelsJsonSupport extends DefaultJsonProtocol {
+  implicit val CustomLabelsFormat = jsonFormat2(Label)
+  implicit val ListOfLabelsFormat = jsonFormat1(List[Label])
```
cjllanwarne: Formatting `List[X]` should be covered as long as you have a formatter for `X`

cromwellApiClient/src/main/scala/cromwell/api/model/Label.scala
```diff
@@ -0,0 +1,10 @@
+package cromwell.api.model
+
+import spray.json.DefaultJsonProtocol
+
+object CustomLabelsJsonSupport extends DefaultJsonProtocol {
+  implicit val CustomLabelsFormat = jsonFormat2(Label)
```
cjllanwarne: I think you'll want a custom formatter along the lines of:
```
implicit object LabelJsonFormat extends RootJsonFormat[Label] {
  def write(l: Label) = JsString(s"${l.key}:${l.value}"  )
  def read(value: JsValue) = value match {
    case JsString(x) => Label(x.split(':')[0], x.split(':')[1])
  }
}
```

cjllanwarne: Actually that read would be nicer with a regex...

.travis.yml
```diff
@@ -19,7 +19,7 @@ before_install:
   - openssl aes-256-cbc -K "$encrypted_5ebd3ff04788_key" -iv "$encrypted_5ebd3ff04788_iv" -in src/bin/travis/resources/jesConf.tar.enc -out jesConf.tar -d || true
 env:
   global:
-    - CENTAUR_BRANCH=develop
+    - CENTAUR_BRANCH=865b8a95846c1a145afe226eefdba1570406bf7e
```

cromwellApiClient/src/test/scala/cromwell/api/model/LabelsJsonFormatterSpec.scala
```diff
@@ -0,0 +1,52 @@
...
+    val expectedLabels =  List(Label("key-1", "value-1"),
+      Label("key-2", "value-2"),
+      Label("key-3", "value-3"))
+
+    val labelsJson: JsObject =
+      """|{
+        |  "key-1":"value-1",
+        |  "key-2":"value-2",
+        |  "key-3":"value-3"
+        |}""".stripMargin.parseJson.asJsObject
```
cjllanwarne: Since these values are copy/pasted you could probably just define them once as, e.g.
```
val inJsonFormat = """ ... """.stripMargin.parseJson.asJsObject
val inScalaFormat = List(Label(....), ...)
```
and then use them as sources/expectations as appropriate in each test

cromwellApiClient/src/test/scala/cromwell/api/model/LabelsJsonFormatterSpec.scala
```diff
@@ -0,0 +1,52 @@
...
+
+    labelsJson.convertTo[List[Label]] shouldBe expectedLabels
+  }
+}
```
cjllanwarne: newline

cjllanwarne:
👍

PR #2271, opened 2017-05-16
validation of input gcs file paths that start with \"

jsotobroad:


engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -429,6 +430,16 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
     }
   }
 
+  private def validateWdlFiles(workflowInputs: WorkflowCoercedInputs): ErrorOr[WorkflowCoercedInputs] = {
+    val failedFiles = workflowInputs.collect {
+      case (callName, WdlSingleFile(value)) if value.startsWith("\"gs://") => "Invalid value for File input '" + callName + "':"+ value + " starts with a '\"' "
```
ruchim: ToL: I find callName to be slightly misleading, it seems like it's the fullyQualifiedName for the input.

jsotobroad: thanks!  I actually wasn't sure what the proper term was in this context

engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -429,6 +430,16 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
     }
   }
 
+  private def validateWdlFiles(workflowInputs: WorkflowCoercedInputs): ErrorOr[WorkflowCoercedInputs] = {
+    val failedFiles = workflowInputs.collect {
+      case (fullyQualifiedName , WdlSingleFile(value)) if value.startsWith("\"gs://") => "Invalid value for File input '" + fullyQualifiedName  + "':"+ value + " starts with a '\"' "
```
danbills: string interpolation would make this more concise

jsotobroad: Ah thanks for the extra prod.  I was getting compilation errors and then I saw this - https://github.com/scala/bug/issues/6476 so I had all but given up but `"""` interpolation works which i just found out was a thing!

cjllanwarne:
Could we have a test in `MaterializeWorkflowDescriptorActorSpec` along the lines of `it should "correctly identify all misformed input file names in an input json"`?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -235,9 +235,10 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
 
     for {
       coercedInputs <- validateCoercedInputs(rawInputs, namespace)
-      _ = pushWfInputsToMetadataService(coercedInputs)
-      evaluatedWorkflowsDeclarations <- validateDeclarations(namespace, workflowOptions, coercedInputs, pathBuilders)
-      declarationsAndInputs <- checkTypes(evaluatedWorkflowsDeclarations ++ coercedInputs)
+      coercedValidatedFileInputs <- validateWdlFiles(coercedInputs)
+      _ = pushWfInputsToMetadataService(coercedValidatedFileInputs)
```
danbills: I'm venturing into unknown territory here as I don't really know what this does, but just checking that you want to send this off somewhere before validating the rest?

ruchim:
:+1: Seconding the idea of adding a test, otherwise LGTM

engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
cjllanwarne: I think the idea is, you want to record (to metadata) the file content submitted to us, even if it turns out that the files were bad/invalid

cjllanwarne:
👍 with a test

cjllanwarne:
Re-👍 test looks good.

PR #2274, opened 2017-05-16
Bringing map to pair coercion and basename from WDL4S

cjllanwarne:
Still needs
 - [x] Changelog

### DIFF
--- a/supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
+++ b/supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
@@ -26,7 +26,7 @@ import cromwell.filesystems.gcs.batch.GcsBatchCommandBuilder
 import cromwell.services.keyvalue.KvClient
 import org.slf4j.LoggerFactory
 import wdl4s._
-import wdl4s.expression.NoFunctions
+import wdl4s.expression.PureStandardLibraryFunctions
 import wdl4s.values._
 
 import scala.collection.JavaConverters._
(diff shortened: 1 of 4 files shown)


engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -389,7 +389,8 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
         List.empty
       }
       WdlNamespaceWithWorkflow.load(w.wdlSource, importResolvers) match {
-        case Failure(e) => s"Unable to load namespace from workflow: ${e.getMessage}".invalidNel
+        case Failure(e) =>
```
cjllanwarne: Revert this

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
```diff
@@ -191,7 +191,7 @@ class JesAsyncBackendJobExecutionActor(override val standardParams: StandardAsyn
   }
 
   private[jes] def generateJesOutputs(jobDescriptor: BackendJobDescriptor): Set[JesFileOutput] = {
-    val wdlFileOutputs = call.task.findOutputFiles(jobDescriptor.fullyQualifiedInputs, NoFunctions) map relativeLocalizationPath
+    val wdlFileOutputs = call.task.findOutputFiles(jobDescriptor.fullyQualifiedInputs, PureStandardLibraryFunctions) map relativeLocalizationPath
```
geoffjentry: Could you explain why this change? 

(personal interest/learning purposes, not a critique)

cjllanwarne: It's because otherwise we weren't able to get engine function evaluations showing up as file outputs. 
E.g. in my motivating situation
```
output {
  String x = read_string(basename(foo, ".bam") + ".bai")
}
```
With no functions, that'll fail to evaluate and the file won't be delocalized for us
With pure functions, things that don't interact externally can still be evaluated, but not the full suite of functions (e.g. we still can't do `read_string(read_string(foo))`)

geoffjentry: Gotcha, thanks

Horneth:
👍

geoffjentry:
👍

PR #2277, opened 2017-05-17
Update google nio + small optimization for copy workflow outputs

Horneth:
- Upgrade google store nio library
- While testing this I found myself in the situation where the workflow initialization would fail, in which case we still do finalization (which is fine), but as part of this finalization we stand up a `CopyWorkflowOutputs` actor, which was failing and making my workflow hang. This fixes the hanging and fails the workflow + don't even run CopyWorkflowOutputs if we failed at Initialization stage

### DIFF
--- a/backend/src/main/scala/cromwell/backend/standard/StandardLifecycleActorFactory.scala
+++ b/backend/src/main/scala/cromwell/backend/standard/StandardLifecycleActorFactory.scala
@@ -3,7 +3,6 @@ package cromwell.backend.standard
 import akka.actor.{ActorRef, Props}
 import com.typesafe.config.Config
 import cromwell.backend._
-import cromwell.backend.io.WorkflowPathsWithDocker
 import cromwell.backend.standard.callcaching._
 import cromwell.core.Dispatcher.BackendDispatcher
 import cromwell.core.path.Path
@@ -183,18 +182,16 @@ trait StandardLifecycleActorFactory extends BackendLifecycleActorFactory {
   override def getExecutionRootPath(workflowDescriptor: BackendWorkflowDescriptor, backendConfig: Config,
                                     initializationData: Option[BackendInitializationData]): Path = {
     initializationData match {
-      // Try to use the initializationData if present to avoid recreating a WorkflowPath
       case Some(data) => data.asInstanceOf[StandardInitializationData].workflowPaths.executionRoot
-      case None => new WorkflowPathsWithDocker(workflowDescriptor, backendConfig).executionRoot
+      case None => super.getWorkflowExecutionRootPath(workflowDescriptor, backendConfig, initializationData)
     }
   }
 
   override def getWorkflowExecutionRootPath(workflowDescriptor: BackendWorkflowDescriptor, backendConfig: Config,
                                             initializationData: Option[BackendInitializationData]): Path = {
     initializationData match {
-        // Try to use the initializationData if present to avoid recreating a WorkflowPath
       case Some(data) => data.asInstanceOf[StandardInitializationData].workflowPaths.workflowRoot
-      case None => new WorkflowPathsWithDocker(workflowDescriptor, backendConfig).workflowRoot
+      case None => super.getWorkflowExecutionRootPath(workflowDescriptor, backendConfig, initializationData)
     }
   }
 
--- a/backend/src/main/scala/cromwell/backend/standard/StandardLifecycleActorFactory.scala
+++ b/backend/src/main/scala/cromwell/backend/standard/StandardLifecycleActorFactory.scala
@@ -3,6 +3,7 @@ package cromwell.backend.standard
 import akka.actor.{ActorRef, Props}
 import com.typesafe.config.Config
 import cromwell.backend._
+import cromwell.backend.io.WorkflowPathsWithDocker
 import cromwell.backend.standard.callcaching._
 import cromwell.core.Dispatcher.BackendDispatcher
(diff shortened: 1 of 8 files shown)


geoffjentry:
:+1:

engine/src/main/scala/cromwell/engine/workflow/lifecycle/WorkflowFinalizationActor.scala
```diff
@@ -57,6 +63,13 @@ case class WorkflowFinalizationActor(workflowIdForLogging: WorkflowId, workflowD
   override def successResponse(data: WorkflowLifecycleActorData) = WorkflowFinalizationSucceededResponse
   override def failureResponse(reasons: Seq[Throwable]) = WorkflowFinalizationFailedResponse(reasons)
 
+  // If a finalization actor dies, send ourselves the failure and stop the child actor
+  override def supervisorStrategy = OneForOneStrategy() {
```
cjllanwarne: Either the comment is wrong or this supervision should be in the `Finalization` actor's supervisor, not the actor itself

backend/src/main/scala/cromwell/backend/standard/StandardLifecycleActorFactory.scala
```diff
@@ -181,12 +182,20 @@ trait StandardLifecycleActorFactory extends BackendLifecycleActorFactory {
 
   override def getExecutionRootPath(workflowDescriptor: BackendWorkflowDescriptor, backendConfig: Config,
                                     initializationData: Option[BackendInitializationData]): Path = {
-    initializationData.get.asInstanceOf[StandardInitializationData].workflowPaths.executionRoot
+    initializationData match {
+      // Try to use the initializationData if present to avoid recreating a WorkflowPath
+      case Some(data) => data.asInstanceOf[StandardInitializationData].workflowPaths.executionRoot
+      case None => new WorkflowPathsWithDocker(workflowDescriptor, backendConfig).executionRoot
```
cjllanwarne: Does this work even if the backend isn't using docker?

backend/src/main/scala/cromwell/backend/standard/StandardLifecycleActorFactory.scala
```diff
@@ -181,12 +182,20 @@ trait StandardLifecycleActorFactory extends BackendLifecycleActorFactory {
...
+    }
   }
 
   override def getWorkflowExecutionRootPath(workflowDescriptor: BackendWorkflowDescriptor, backendConfig: Config,
                                             initializationData: Option[BackendInitializationData]): Path = {
-    initializationData.get.asInstanceOf[StandardInitializationData].workflowPaths.workflowRoot
+    initializationData match {
+        // Try to use the initializationData if present to avoid recreating a WorkflowPath
+      case Some(data) => data.asInstanceOf[StandardInitializationData].workflowPaths.workflowRoot
+      case None => new WorkflowPathsWithDocker(workflowDescriptor, backendConfig).workflowRoot
```
cjllanwarne: Since this safety catch appears twice, is it worth making the backup `WorkflowPathsWithDocker` a `lazy val`?

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -346,7 +346,19 @@ class WorkflowActor(val workflowId: WorkflowId,
...
+    }
+    
+    context.actorOf(WorkflowFinalizationActor.props(workflowId,
+      workflowDescriptor,
+      ioActor,
+      jobExecutionMap,
+      workflowOutputs,
+      stateData.initializationData,
+      copyWorkflowOutputsActorProps
+    ), name = s"WorkflowFinalizationActor")
```
cjllanwarne: Woohoo!
Bonus points if you also explicitly name the inputs so that there's no chance of untyped `ActorRef`s getting accidentally reordered

cjllanwarne:
👍

backend/src/main/scala/cromwell/backend/standard/StandardLifecycleActorFactory.scala
Horneth: Actually in this case we could just call `super.getWorkflowExecutionRootPath` I think.
It doesn't really matter as of now anyway since it won't be called anymore if initialization fails so we should always have a Some(data)

backend/src/main/scala/cromwell/backend/standard/StandardLifecycleActorFactory.scala
```diff
@@ -181,12 +181,18 @@ trait StandardLifecycleActorFactory extends BackendLifecycleActorFactory {
 
   override def getExecutionRootPath(workflowDescriptor: BackendWorkflowDescriptor, backendConfig: Config,
                                     initializationData: Option[BackendInitializationData]): Path = {
-    initializationData.get.asInstanceOf[StandardInitializationData].workflowPaths.executionRoot
+    initializationData match {
+      case Some(data) => data.asInstanceOf[StandardInitializationData].workflowPaths.executionRoot
+      case None => super.getWorkflowExecutionRootPath(workflowDescriptor, backendConfig, initializationData)
```
pshapiro4broad: Shouldn't this be `super.getExecutionRootPath()`?

Horneth: Yep, good catch

PR #2291, opened 2017-05-23
added read limits to read_X functions 

danbills:


backend/src/test/scala/cromwell/backend/wdl/ReadLikeFunctionTest.scala
```diff
@@ -0,0 +1,101 @@
...
+        jPath
+      }
+
+      val file = createTempFileOfSize(n)
+      val params = Seq(Try(WdlString(file.toString)))
+      //apply the test "g"
+      f(rlf)(params) match {
+        case t => g(t)
+      }
+      delete(file)
```
danbills: realizing that this code will not be reached.

 test files will be left undeleted if the test `g` throws, which by design it will on 1/2 of these tests.

danbills:
Am wondering if these should be specified in the conf file.

They should be generous and static but I could envision a scenario where one would want to override read_json or something.  In that scenario a config change would be nice to have available

geoffjentry:
Yeah, it was probably lost in the back & forth on that thread but IIRC the numbers were minimums which could be increased by configuration

backend/src/test/scala/cromwell/backend/wdl/ReadLikeFunctionTest.scala
pshapiro4broad: In Java I've used JUnit `TemporaryFolder` which takes care of cleaning up all test created files on exit. This is much better than adding `delete()` to tests and also avoids cluttering up `/tmp`. Maybe there's a way to use this library in Scala, or to implement the same functionality using Scala objects.

rtitle: Or call this on the `File` to auto-delete it when the JVM terminates:
https://docs.oracle.com/javase/7/docs/api/java/io/File.html#deleteOnExit()

backend/src/main/scala/cromwell/backend/wdl/ReadLikeFunctions.scala
```diff
@@ -6,10 +6,28 @@ import wdl4s.expression.WdlStandardLibraryFunctions
...
+  val readMapLimit = 128000
+  val readObjectLimit = 128000
+}
+
+case class FileSizeTooBig(override val getMessage: String) extends Exception
+
+trait ReadLikeFunctions extends PathFactory with FileSizeLimitationConfig { this: WdlStandardLibraryFunctions =>
+
+  def fileSize: WdlValue=> Try[Long] = 
+    w => Try(buildPath(w.valueString).size)
```
Horneth: ToL: Another way to do it that could avoid an extra request to get the size (in case of GCS for example) would be to just take X bytes from an input stream. 
It could look like `new String(buildPath(path).bytes.take(100).toArray, "UTF-8")`

backend/src/test/scala/cromwell/backend/wdl/ReadLikeFunctionTest.scala
pshapiro4broad: Using `deleteOnExit()` has similar issues as using explicit `delete()`. It requires that the user remember to do it, and requires that it be done on every resource (e.g. every directory and file) that a test creates. Using `TemporaryFolder` avoids these issues.

backend/src/test/scala/cromwell/backend/wdl/ReadLikeFunctionTest.scala
```diff
@@ -0,0 +1,101 @@
...
+      override val readJsonLimit = _readJsonLimit
+      override val readTsvLimit = _readTsvLimit
+      override val readMapLimit = _readMapLimit
+      override val readObjectLimit = _readObjectLimit
+    }
+  }
+
+  def testOverUnder(command: String, n: Int, f: ReadLikeFunctions => (Seq[Try[WdlValue]] =>Try[WdlValue])) = {
+
+    def testInner(n: Int, f: ReadLikeFunctions => (Seq[Try[WdlValue]] =>Try[WdlValue]),g: Try[WdlValue] => Unit) = {
```
danbills: these `n` and `f` params need not be repeated as they are scope

backend/src/test/scala/cromwell/backend/wdl/ReadLikeFunctionTest.scala
danbills: @pshapiro4broad I'm not aware of such a magic tempFolder thing in Scalatest and some quick googling did not reveal anything.

I'm going to go w/ `deleteOnExit` (thanks @rtitle )  with a wary eye towards its caveats

backend/src/main/scala/cromwell/backend/wdl/ReadLikeFunctions.scala
danbills: per kristian on the bug: if file is larger than limit than should be an error, not read first X bytes and silently drop the rest.

I'm kind of hesitant to re-write `size` or invent something to avoid a metadata request, but perhaps @katevoss or @geoffjentry have opinions on this tradeoff?

danbills: I see this on bug too from @geoffjentry :
" Cromwell will attempt to check file size prior to reading it or pulling it across the network for cloud filesystems"

backend/src/test/scala/cromwell/backend/wdl/ReadLikeFunctionTest.scala
cjllanwarne: There's also the `beforeAndAfterAll` mixin that gives you a hook to tidy up after yourself, after all tests have completed

backend/src/test/scala/cromwell/backend/wdl/ReadLikeFunctionTest.scala
```diff
@@ -0,0 +1,101 @@
...
+        case Failure(s: FileSizeTooBig) => //success
+        case t => throw new RuntimeException(s"should not have eaten this file that is too big! msg: $t")
+      })
+    }
+
+    def testUnder(n: Int, f: ReadLikeFunctions => (Seq[Try[WdlValue]] =>Try[WdlValue])) = {
+      testInner(n - 1, f, {
+        case Success(_) => 
+        case Failure(s:FileSizeTooBig) => throw s
+        case Failure(t) => //we're not testing parsing
```
cjllanwarne: Hmm, this could still be hiding some reason why this test isn't working as intended. 

Is it possible to handle the format exceptions as OK and everything else gets thrown?

backend/src/main/scala/cromwell/backend/wdl/ReadLikeFunctions.scala
```diff
@@ -9,7 +9,24 @@ import wdl4s.values._
 
 import scala.util.{Failure, Success, Try}
 
-trait ReadLikeFunctions extends PathFactory { this: WdlStandardLibraryFunctions =>
+trait FileSizeLimitationConfig {
+  val readLinesLimit = 128000
+  val readBoolLimit = 5
```
cjllanwarne: Maybe make this one a little higher to take care of any trailing newline characters, etc?

backend/src/main/scala/cromwell/backend/wdl/ReadLikeFunctions.scala
```diff
@@ -34,13 +53,22 @@ trait ReadLikeFunctions extends PathFactory { this: WdlStandardLibraryFunctions
...
       contents <- readContentsFromSingleFileParameter("read_lines", params)
       lines = contents.split("\n")
     } yield WdlArray(WdlArrayType(WdlStringType), lines map WdlString)
   }
 
+  def validateFileSizeIsWithinLimits(functionName: String, params: Seq[Try[WdlValue]], limit: Int): Try[Unit] = 
+    for {
+      fileName <- extractSingleArgument(functionName, params)
+      fileSize <- fileSize(fileName)
+      _ = if (fileSize > limit) throw new FileSizeTooBig(s"read of $fileName failed because $fileSize > $limit")
```
cjllanwarne: You could also add in the `functionName` here, for the extra error msg info.

backend/src/test/scala/cromwell/backend/wdl/ReadLikeFunctionTest.scala
```diff
@@ -0,0 +1,101 @@
...
+import cromwell.backend.standard.{DefaultStandardExpressionFunctionsParams, StandardExpressionFunctions}
+import cromwell.core.CallContext
+import cromwell.core.path.DefaultPathBuilder
+import org.scalatest.{FlatSpec, Matchers}
+import wdl4s.values._
+import com.google.common.io.Files
+import fs2.{Task, Stream}
+import java.nio.file.Files.delete
+
+class FileSizeSpec extends FlatSpec with Matchers /*with TryValues*/ {
```
cjllanwarne: Looks like the inline comment is unneeded?

cjllanwarne: File name could be updated to match class name

project/Dependencies.scala
```diff
@@ -22,10 +22,12 @@ object Dependencies {
   lazy val googleGenomicsServicesApiV = "1.22.0"
   lazy val betterFilesV = "2.17.1"
   lazy val catsV = "0.9.0"
+  lazy val fs2V = "0.9.6"
 
   // Internal collections of dependencies
 
   private val catsDependencies = List(
+    "co.fs2" %% "fs2-io" % fs2V % "test",
```
cjllanwarne: ToL: Is this a cats dependency?

backend/src/main/scala/cromwell/backend/wdl/ReadLikeFunctions.scala
geoffjentry: Yeah, IIRC my comment came  before the one that @kcibul made. I was trying to avoid the situation where a giant file is pulled across the network and then we just read a few bytes, so in that case I agree w/ what @Horneth says. However Kristian's comment is also right, I agree that it should be an error, so size would need to be checked anyways.

project/Dependencies.scala
geoffjentry: Since I saw this coming across - for what purpose is fs2 being used?

backend/src/test/scala/cromwell/backend/wdl/ReadLikeFunctionTest.scala
danbills: The test is working as intended.  Unless I'm missing something?

But good call, I can catch those format exceptions specifically and throw anything else.

backend/src/main/scala/cromwell/backend/wdl/ReadLikeFunctions.scala
danbills: yeah I spruced this error message up a bit

project/Dependencies.scala
danbills: no it has no dependencies by design

danbills: @geoffjentry In this case, to write files of size N easily.

cjllanwarne: FWIW, I asked because it's listed under the `catsDependencies` section

danbills: @cjllanwarne Oh, that was done without much thought.  I'll put it by itself.

backend/src/test/scala/cromwell/backend/wdl/ReadLikeFunctionTest.scala
cjllanwarne: I'm sure it's working as intended! :)

But, if it wasn't (or is broken in the future), that error might be swallowed by the "all `Failure(_)`s are ok" policy.

danbills: ah, yes I agree.

backend/src/test/scala/cromwell/backend/wdl/ReadLikeFunctionTest.scala
pshapiro4broad: FWIW here's what stackoverflow says:
using the JUnit .jar: https://stackoverflow.com/a/32397751
using a scala trait: https://stackoverflow.com/a/24748621

backend/src/main/scala/cromwell/backend/wdl/ReadLikeFunctions.scala
Horneth: We get an `Iterator[Byte]` so knowing if there is more after we take what we want is just a `iterator.isEmpty`.
But on the other hand we would always take the first few bytes even if we're going to discard them after.. It's a tradeoff

danbills:
closing as I"m adding Config feature so technically "in progress".  Not 100% comfortable w/ this workflow but we'll see how it goes

cjllanwarne:
👍

cjllanwarne:
Is this worth a changelog/readme comment?

.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -1,6 +0,0 @@
-<component name="InspectionProjectProfileManager">
-  <profile version="1.0">
-    <option name="myName" value="Project Default" />
-    <inspection_tool class="TypeAnnotation" enabled="false" level="WARNING" enabled_by_default="false" />
-  </profile>
-</component>
```
mcovarr: I don't think we want to delete this file, this keeps those annoying type annotation warnings out.

backend/src/main/scala/cromwell/backend/wdl/FileSizeLimitationConfig.scala
```diff
@@ -0,0 +1,48 @@
...
+  def readStringLimit: Int
+
+  def readJsonLimit: Int
+
+  def readTsvLimit: Int
+
+  def readMapLimit: Int
+
+  def readObjectLimit: Int
+}
```
mcovarr: can't this just be a case class?

backend/src/test/scala/cromwell/backend/wdl/FileSizeSpec.scala
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  val tempDir = Files.createTempDir
+  tempDir.deleteOnExit
+
+  def testOverUnder(command: String, n: Int, f: ReadLikeFunctions => (Seq[Try[WdlValue]] =>Try[WdlValue])) = {
+
+    def testInner(n: Int, g: Try[WdlValue] => Unit) = {
+
+      def createTempFileOfSize(n: Int): Path = {
```
mcovarr: inner formal shadowing outer is confusing

backend/src/test/scala/cromwell/backend/wdl/FileSizeSpec.scala
```diff
@@ -0,0 +1,108 @@
...
+        val end = fs2.io.file.writeAll[Task](jPath, Seq(CREATE_NEW, WRITE) )
+        (start to end).run.unsafeRunSync 
+        //jPath is now a file of n bytes, we can return it
+        jPath
+      }
+
+      val file = createTempFileOfSize(n)
+      val params = Seq(Try(WdlString(file.toString)))
+
+      //apply the test "g"
```
mcovarr: descriptive parameter names would make the comment unnecessary 😛

backend/src/main/scala/cromwell/backend/wdl/FileSizeLimitationConfig.scala
danbills: it could and perhaps should, but because we're using traits and mixins the trait is more desirable.

If it were a case class it would have to be the _only_ case class being mixed in at the top, and that's not really a necessary requirement for this to work.

backend/src/test/scala/cromwell/backend/wdl/FileSizeSpec.scala
danbills: inadvertent, will change

backend/src/test/scala/cromwell/backend/wdl/FileSizeSpec.scala
danbills: true

.idea/inspectionProfiles/Project_Default.xml
danbills: ok

backend/src/main/scala/cromwell/backend/wdl/ReadLikeFunctions.scala
```diff
@@ -34,13 +42,23 @@ trait ReadLikeFunctions extends PathFactory { this: WdlStandardLibraryFunctions
...
       lines = contents.split("\n")
     } yield WdlArray(WdlArrayType(WdlStringType), lines map WdlString)
   }
 
+  def validateFileSizeIsWithinLimits(functionName: String, params: Seq[Try[WdlValue]], limit: Int): Try[Unit] = 
+    for {
+      fileName <- extractSingleArgument(functionName, params)
+      fileSize <- fileSize(fileName)
+      _ = if (fileSize > limit)
+            throw new FileSizeTooBig(s"Use of $fileName failed because the file was too big ($fileSize bytes when only files of up to  $limit bytes are permissible)")
```
mcovarr: superfluous space in the string

backend/src/test/scala/cromwell/backend/wdl/FileSizeSpec.scala
```diff
@@ -0,0 +1,107 @@
...
+  List[(String, Int, ReadLikeFunctions => (Seq[Try[WdlValue]] =>Try[WdlValue]))](
+    ("lines",  _readLinesLimit,  _.read_lines),
+    ("int",    _readIntLimit,    _.read_int),
+    ("map",    _readMapLimit,    _.read_map),
+    ("float",  _readFloatLimit,  _.read_float),
+    ("String", _readStringLimit, _.read_string),
+    ("tsv",    _readTsvLimit,    _.read_tsv),
+    ("object", _readObjectLimit, _.read_object)
+  ).flatMap{ (testOverUnder _).tupled }
+}
```
mcovarr: The whole file might benefit from a command-opt-L formatting, there are things off here and there

mcovarr:
:+1: delta nits

backend/src/test/scala/cromwell/backend/wdl/FileSizeSpec.scala
danbills: TiL formatting command.  Thanks!

PR #2293, opened 2017-05-23
TES Backend Updates

adamstruck:
* Updated to reflect the current status of the [TES schema](https://github.com/ga4gh/task-execution-schemas)
* Adds gcsPathBuilder
* Removed cpu, ram, and disk default values so that the TES implementation can handle defaults


src/bin/travis/testCentaurTes.sh
```diff
@@ -31,26 +31,48 @@ printTravisHeartbeat
...
 
 # All tests use ubuntu:latest - make sure it's there before starting the tests 
 # because pulling the image during some of the tests would cause them to fail 
 # (specifically output_redirection which expects a specific value in stderr)
 docker pull ubuntu:latest
-centaur/test_cromwell.sh -j"${CROMWELL_JAR}" -c${TES_CENTAUR_CONF} -elocaldockertest
+
+# The following tests are skipped:
+#
+# non_root_sepcified_user:   TES doesnt support switch users in the image
```
mcovarr: switch => switching

supportedBackends/tes/src/main/scala/cromwell/backend/impl/tes/TesResponseJsonFormatter.scala
```diff
@@ -2,24 +2,19 @@ package cromwell.backend.impl.tes
...
-
-final case class TesGetResponse(
-                                 jobID: String,
-                                 task: TesTaskMessage,
-                                 state: String,
-                                 logs: Option[Seq[JobLogs]]
-                               )
+final case class CreateTaskResponse(id: String)
 
+final case class CancelTaskResponse()
```
mcovarr: looks like a case object

mcovarr:
:+1:

supportedBackends/tes/src/main/scala/cromwell/backend/impl/tes/TesResponseJsonFormatter.scala
adamstruck: AFAIK  `jsonFormat0` requires this to be a case class.

src/bin/travis/testCentaurTes.sh
```diff
@@ -31,26 +31,48 @@ printTravisHeartbeat
...
+# write_lines_files:         all inputs are read-only in TES
+# lots_of_inputs:            Funnel mounts in each input seperately, this task surpasses the docker limit for volumes
+# call_cache_capoeira_local: fails on task 'read_files_without_docker' since the 'docker' runtiem key is required for this backend
```
adamstruck: clearly I wasn't paying attention to the spell checker in IntelliJ

supportedBackends/tes/src/main/scala/cromwell/backend/impl/tes/TesResponseJsonFormatter.scala
cjllanwarne: Astonishingly, this seems to be true - https://github.com/spray/spray-json/issues/186

I'm curious what this serializes to, presumable an empty json object (`{}`)?  Is that what you want?

I'm not too fussed either way but it should be possible to make a `new RootJsonFormat[CancelTaskResponse]` for the case object, where write creates a `JsString` and read would just be a string comparison vs the literal string `"CancelTaskResponse"`.

backend/src/main/scala/cromwell/backend/validation/MemoryValidation.scala
```diff
@@ -39,31 +39,31 @@ object MemoryValidation {
   }
 
   private[validation] val wrongAmountFormat =
-    s"Expecting ${RuntimeAttributesKeys.MemoryKey} runtime attribute value greater than 0 but got %s"
+    s"Expecting %s runtime attribute value greater than 0 but got %s"
```
cjllanwarne: No longer needs an `s"` prefix

src/bin/travis/testCentaurTes.sh
```diff
@@ -31,26 +31,48 @@ printTravisHeartbeat
...
 # (specifically output_redirection which expects a specific value in stderr)
 docker pull ubuntu:latest
-centaur/test_cromwell.sh -j"${CROMWELL_JAR}" -c${TES_CENTAUR_CONF} -elocaldockertest
+
+# The following tests are skipped:
+#
+# non_root_specified_user:   TES doesn't support switching users in the image
+# write_lines_files:         all inputs are read-only in TES
+# lots_of_inputs:            Funnel mounts in each input separately, this task surpasses the docker limit for volumes
+# call_cache_capoeira_local: fails on task 'read_files_without_docker' since the 'docker' runtime key is required for this backend
```
cjllanwarne: Hmm, that's a shame... perhaps we could add a `call_cache_capoeira_tes` with it removed?

supportedBackends/tes/src/main/scala/cromwell/backend/impl/tes/TesAsyncBackendJobExecutionActor.scala
```diff
@@ -146,21 +150,25 @@ class TesAsyncBackendJobExecutionActor(override val standardParams: StandardAsyn
...
+          case s if s.contains("CANCELED") =>
             jobLogger.info(s"Job ${handle.pendingJob.jobId} was canceled")
             FailedOrError
 
-          case s if s.contains("Error") =>
+          case s if s.contains("ERROR") =>
+            jobLogger.info(s"TES reported an error for Job ${handle.pendingJob.jobId}")
+            FailedOrError
+
+          case s if s.contains("SYSTEM_ERROR") =>
```
cjllanwarne: This case will never hit because `"ERROR"` (checked above) is a substring of `"SYSTEM_ERROR"`.

Since they're the same handler anyway, I'd just remove this case.

supportedBackends/tes/src/main/scala/cromwell/backend/impl/tes/TesJobPaths.scala
```diff
@@ -28,18 +28,15 @@ case class TesJobPaths private[tes] (override val workflowPaths: TesWorkflowPath
   val callInputsDockerRoot = callDockerRoot.resolve("inputs")
   val callInputsRoot = callRoot.resolve("inputs")
 
-  //TODO move to TesConfiguration
-  private def prefixScheme(path: String): String = "file://" + path
-
-  def storageInput(path: String): String = prefixScheme(path)
+  def storageInput(path: String): String = path
```
cjllanwarne: Is this still necessary? It's just an identity function now.

supportedBackends/tes/src/main/scala/cromwell/backend/impl/tes/TesResponseJsonFormatter.scala
```diff
@@ -2,24 +2,19 @@ package cromwell.backend.impl.tes
...
-  implicit val jobLogsFormat = jsonFormat8(JobLogs)
-  implicit val tesPostResponseFormat = jsonFormat1(TesPostResponse)
-  implicit val tesGetResponseFormat = jsonFormat4(TesGetResponse)
+  implicit val executorFormat = jsonFormat8(Executor)
+  implicit val executorLogFormat = jsonFormat7(ExecutorLog)
+  implicit val outputFileLogFormat = jsonFormat3(OutputFileLog)
+  implicit val taskLogFormat = jsonFormat5(TaskLog)
+  implicit val taskFormat = jsonFormat12(Task)
+  implicit val createTaskResponseFormat = jsonFormat1(CreateTaskResponse)
+  implicit val cancelTaskResponseFormat = jsonFormat0(CancelTaskResponse)
```
cjllanwarne: Just curious, what JSON object ends up these internal messages in them? It seems unlikely a user would ever want to see them

supportedBackends/tes/src/main/scala/cromwell/backend/impl/tes/TesTask.scala
```diff
@@ -125,117 +131,104 @@ final case class TesTask(jobDescriptor: BackendJobDescriptor,
...
-    ),
-    Volume(
-      Option(tesPaths.callExecutionDockerRoot.toString),
-      runtimeAttributes.disk.to(MemoryUnit.GB).amount.toInt,
-      None,
-      tesPaths.callExecutionDockerRoot.toString,
-      Option(false)
-    )
-  ) ++ workingDirVolume
+  val outputs: Seq[TaskParameter] = wdlOutputs ++ standardOutputs ++ Seq(commandScriptOut)
```
cjllanwarne: You're now including the command script as an output?

adamstruck: The TES backend is now responsible for creating this file so if we want it to end up in the final workflow output directory (or gs bucket) it needs to be an output.

supportedBackends/tes/src/main/scala/cromwell/backend/impl/tes/TesResponseJsonFormatter.scala
adamstruck: We didn't have the task view feature of the spec implemented in Funnel when I wrote this so during polling we needed this to deserialize the responses during status polling. 

We implemented this feature yesterday, so I think its safe for me to remove most of those now.

adamstruck: Actually the another reason those are there is for marshaling the Task to json before submitting the task to the TES server.

supportedBackends/tes/src/main/scala/cromwell/backend/impl/tes/TesTask.scala
```diff
@@ -124,117 +130,112 @@ final case class TesTask(jobDescriptor: BackendJobDescriptor,
...
+    case None =>
+      None
+  }
+
+  private val disk = runtimeAttributes.disk match {
+    case Some(x) =>
+      Option(x.to(MemoryUnit.GB).amount)
+    case None =>
+      None
+  }
```
adamstruck: This should be addressed in the new commits.

src/bin/travis/testCentaurTes.sh
adamstruck: Should I make a PR to centaur with your suggested change?

cjllanwarne: That'd be really helpful, thanks!

adamstruck: https://github.com/broadinstitute/centaur/pull/201

supportedBackends/tes/src/main/scala/cromwell/backend/impl/tes/TesTask.scala
cjllanwarne: Is this changed? It looks the same?

cjllanwarne:
Looking good to me 👍

supportedBackends/tes/src/main/scala/cromwell/backend/impl/tes/TesTask.scala
adamstruck: https://github.com/ohsu-comp-bio/cromwell/blob/c6ae41a8ce99a6a690a9f7954de727f29e3376a4/supportedBackends/tes/src/main/scala/cromwell/backend/impl/tes/TesTask.scala#L152-L157

adamstruck:
rebase is done - anything else I need to do on my end?

PR #2314, opened 2017-06-01
Add retries to credential validation

Horneth:
Specifically to "refresh token" mode credential validation.
Other modes should only be validated once at Cromwell startup, since they don't change for every workflow.
Only "refresh token mode" generates different credentials for every workflow (with the refresh token passed in), that need to be validated.

Retrying this asynchronously means turning into `Future`s a bunch of methods that were previously synchronous.

### DIFF
--- a/core/src/main/scala/cromwell/core/path/DefaultPathBuilderFactory.scala
+++ b/core/src/main/scala/cromwell/core/path/DefaultPathBuilderFactory.scala
@@ -3,6 +3,8 @@ package cromwell.core.path
 import akka.actor.ActorSystem
 import cromwell.core.WorkflowOptions
 
+import scala.concurrent.{ExecutionContext, Future}
+
 case object DefaultPathBuilderFactory extends PathBuilderFactory {
-  override def withOptions(options: WorkflowOptions)(implicit actorSystem: ActorSystem) = DefaultPathBuilder
+  override def withOptions(options: WorkflowOptions)(implicit as: ActorSystem, ec: ExecutionContext) = Future.successful(DefaultPathBuilder)
 }
--- a/core/src/main/scala/cromwell/core/path/DefaultPathBuilderFactory.scala
+++ b/core/src/main/scala/cromwell/core/path/DefaultPathBuilderFactory.scala
@@ -6,5 +6,5 @@ import cromwell.core.WorkflowOptions
 import scala.concurrent.{ExecutionContext, Future}
 
 case object DefaultPathBuilderFactory extends PathBuilderFactory {
-  override def withOptions(options: WorkflowOptions)(implicit as: ActorSystem, ec: ExecutionContext) = Future.successful(DefaultPathBuilder)
+  override def withOptions(options: WorkflowOptions)(implicit actorSystem: ActorSystem, ec: ExecutionContext) = Future.successful(DefaultPathBuilder)
 }
--- a/engine/src/main/scala/cromwell/engine/EngineFilesystems.scala
+++ b/engine/src/main/scala/cromwell/engine/EngineFilesystems.scala
@@ -8,6 +8,9 @@ import cromwell.core.path.{DefaultPathBuilder, PathBuilder}
 import cromwell.filesystems.gcs.{GcsPathBuilderFactory, GoogleConfiguration}
 import lenthall.exception.MessageAggregation
 import net.ceedubs.ficus.Ficus._
+import cats.instances.future._
+import cats.instances.list._
+import cats.syntax.traverse._
 
 import scala.concurrent.{ExecutionContext, Future}
 
@@ -27,13 +30,10 @@ object EngineFilesystems {
   private val gcsPathBuilderFactory = googleAuthMode map { mode =>
     GcsPathBuilderFactory(mode, googleConf.applicationName)
   }
-  
-  private val defaultFileSystem = if (config.as[Boolean]("engine.filesystems.local.enabled")) {
-    Option(Future.successful(DefaultPathBuilder))
-  } else None
 
-  def pathBuildersForWorkflow(workflowOptions: WorkflowOptions)(implicit as: ActorSystem, ec: ExecutionContext): Future[List[PathBuilder]] = {
-    val maybeEventualBuilders = List(gcsPathBuilderFactory map { _.withOptions(workflowOptions) }, defaultFileSystem).flatten
-    Future.sequence(maybeEventualBuilders)
-  }
+  private val defaultFileSystem =
+    Option(DefaultPathBuilder).filter(_ => config.as[Boolean]("engine.filesystems.local.enabled"))
+
+  def pathBuildersForWorkflow(workflowOptions: WorkflowOptions)(implicit as: ActorSystem, ec: ExecutionContext): Future[List[PathBuilder]] =
+    gcsPathBuilderFactory.toList.traverse(_.withOptions(workflowOptions)).map(_ ++ defaultFileSystem)
 }
--- a/engine/src/main/scala/cromwell/engine/EngineFilesystems.scala
+++ b/engine/src/main/scala/cromwell/engine/EngineFilesystems.scala
@@ -9,8 +9,9 @@ import cromwell.filesystems.gcs.{GcsPathBuilderFactory, GoogleConfiguration}
 import lenthall.exception.MessageAggregation
 import net.ceedubs.ficus.Ficus._
 
-case class EngineFilesystems(actorSystem: ActorSystem) {
+import scala.concurrent.{ExecutionContext, Future}
 
+object EngineFilesystems {
   private val config = ConfigFactory.load
   private val googleConf: GoogleConfiguration = GoogleConfiguration(config)
   private val googleAuthMode = config.as[Option[String]]("engine.filesystems.gcs.auth") map { confMode =>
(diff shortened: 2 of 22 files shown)


engine/src/main/scala/cromwell/engine/EngineFilesystems.scala
```diff
@@ -28,10 +29,11 @@ case class EngineFilesystems(actorSystem: ActorSystem) {
...
   private val defaultFileSystem = if (config.as[Boolean]("engine.filesystems.local.enabled")) {
-    Option(DefaultPathBuilder)
+    Option(Future.successful(DefaultPathBuilder))
   } else None
 
-  def pathBuildersForWorkflow(workflowOptions: WorkflowOptions): List[PathBuilder] = {
-    List(gcsPathBuilderFactory map { _.withOptions(workflowOptions)(actorSystem) }, defaultFileSystem).flatten
+  def pathBuildersForWorkflow(workflowOptions: WorkflowOptions)(implicit as: ActorSystem, ec: ExecutionContext): Future[List[PathBuilder]] = {
+    val maybeEventualBuilders = List(gcsPathBuilderFactory map { _.withOptions(workflowOptions) }, defaultFileSystem).flatten
+    Future.sequence(maybeEventualBuilders)
```
mcovarr:
```
import cats.instances.list._
import cats.syntax.traverse._
```

then 

```
List(gcsPathBuilderFactory map { _.withOptions(workflowOptions) }, defaultFileSystem).flatten.sequence
```

mcovarr: h/t @danbills

Horneth: Does it aggregate all the future failures instead of stopping at the first one ?

mcovarr: no but it doesn't look like the existing code is doing that either

Horneth: no right I was just wondering if the cats sequence would make it do it 😄

engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -165,18 +179,34 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
...
+                                      conf: Config,
+                                      workflowOptions: WorkflowOptions,
+                                      pathBuilders: List[PathBuilder]): ErrorOr[EngineWorkflowDescriptor] = {
     val namespaceValidation = validateNamespace(sourceFiles)
-    val workflowOptionsValidation = validateWorkflowOptions(sourceFiles.workflowOptionsJson)
     val labelsValidation = validateLabels(sourceFiles.labelsJson)
-    (namespaceValidation |@| workflowOptionsValidation |@| labelsValidation) map {
-      (_, _, _)
-    } flatMap { case (namespace, workflowOptions, labels) =>
+    (namespaceValidation |@| labelsValidation) map {
```
rtitle: I think
```
(f1 |@| f2) map { (_, _) }
```
is equivalent to
```
(f1 |@| f2).tupled
```

engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -165,18 +179,34 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
       stay
   }
 
+  private def workflowInitializationFailed(errors: NonEmptyList[String], replyTo: ActorRef) = {
+    sender() ! MaterializeWorkflowDescriptorFailureResponse(
+      new IllegalArgumentException with MessageAggregation {
+        val exceptionContext = s"Workflow input processing failed"
```
mcovarr: unneeded s

engine/src/main/scala/cromwell/engine/EngineFilesystems.scala
cjllanwarne: Is there an equivalent of `Future` in cats that conceptually wraps `Validated` instead of `Try`?

core/src/main/scala/cromwell/core/path/DefaultPathBuilderFactory.scala
```diff
@@ -3,6 +3,8 @@ package cromwell.core.path
 import akka.actor.ActorSystem
 import cromwell.core.WorkflowOptions
 
+import scala.concurrent.{ExecutionContext, Future}
+
 case object DefaultPathBuilderFactory extends PathBuilderFactory {
-  override def withOptions(options: WorkflowOptions)(implicit actorSystem: ActorSystem) = DefaultPathBuilder
+  override def withOptions(options: WorkflowOptions)(implicit as: ActorSystem, ec: ExecutionContext) = Future.successful(DefaultPathBuilder)
```
mcovarr: is there a reason for the more cryptic `as`?

engine/src/main/scala/cromwell/engine/EngineFilesystems.scala
danbills:
```
  private val defaultFileSystem =
    Option(DefaultPathBuilder).filter(_ => config.as[Boolean]("engine.filesystems.local.enabled"))

  def pathBuildersForWorkflow(workflowOptions: WorkflowOptions)(implicit as: ActorSystem, ec: ExecutionContext): Future[List[PathBuilder]] =
    gcsPathBuilderFactory.toList.traverse(_.withOptions(workflowOptions)).map(_ ++ defaultFileSystem)

```

supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystemInitializationActor.scala
```diff
@@ -36,20 +37,22 @@ class SharedFileSystemInitializationActor(standardParams: StandardInitialization
   lazy val pathBuilderFactories: List[PathBuilderFactory] =
     List(gcsPathBuilderFactory, Option(DefaultPathBuilderFactory)).flatten
 
-  override lazy val pathBuilders: List[PathBuilder] =
-    pathBuilderFactories map { _.withOptions(workflowDescriptor.workflowOptions)(context.system) }
+  override lazy val pathBuilders: Future[List[PathBuilder]] = {
+    Future.sequence(pathBuilderFactories map { _.withOptions(workflowDescriptor.workflowOptions) })
```
mcovarr: `sequence` here too

core/src/main/scala/cromwell/core/path/DefaultPathBuilderFactory.scala
Horneth: The `Retry.withRetry` needs it

engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -72,15 +74,11 @@ object MaterializeWorkflowDescriptorActor {
...
   case object ReadyToMaterializeState extends MaterializeWorkflowDescriptorActorState
+  case object MaterializingState extends MaterializeWorkflowDescriptorActorState
   case object MaterializationSuccessfulState extends MaterializeWorkflowDescriptorActorTerminalState
   case object MaterializationFailedState extends MaterializeWorkflowDescriptorActorTerminalState
   case object MaterializationAbortedState extends MaterializeWorkflowDescriptorActorTerminalState
 
-  /*
-  Data
-   */
-  case class MaterializeWorkflowDescriptorActorData()
```
mcovarr: glad this is gone, but 😕  why it was here in the first place

geoffjentry: Ask @cjllanwarne :)

core/src/main/scala/cromwell/core/path/DefaultPathBuilderFactory.scala
mcovarr: Oh I meant is there a reason the variable name was changed from `actorSystem` to `as`?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
mcovarr: Oh there's a lot of this pattern we can clean up in Cromwell if that's true.  😄

mcovarr:
:+1: delta nitpickery

core/src/main/scala/cromwell/core/path/DefaultPathBuilderFactory.scala
Horneth: oh nop that's just my cryptic side I guess 😄  I'll change it back

geoffjentry: on the flip side it's implicit and no one shall ever see it ;)

geoffjentry:
👍

filesystems/gcs/src/main/scala/cromwell/filesystems/gcs/GcsPathBuilder.scala
```diff
@@ -44,55 +46,71 @@ object GcsPathBuilder {
...
-                     options: WorkflowOptions) extends PathBuilder {
-  authMode.validate(options)
+    val storageOptionsBuilder = StorageOptions.newBuilder()
+      .setTransportOptions(transportOptions)
+      .setCredentials(credentials)
 
-  protected val transportOptions = HttpTransportOptions.newBuilder()
-    .setReadTimeout(3.minutes.toMillis.toInt)
-    .build()
+    retrySettings foreach storageOptionsBuilder.setRetrySettings
```
danbills: are you sure the builder is mutable?  I.e. this might return the modified builder

Horneth: Yeah I checked, it does return the modified builder but it is also mutable.

PR #2335, opened 2017-06-06
Call Caching Diff endpoint

Horneth:


README.md
```diff
@@ -3379,7 +3379,167 @@ Response:
 }
 ```
 
+## GET /api/workflows/:version/callcaching/diff
```
cjllanwarne: Add to ToC

README.md
```diff
@@ -3379,7 +3379,167 @@ Response:
...
 
+This endpoint returns the hash differences between 2 *completed* (successfully or not) calls.
+The following query parameters are supported:
+
+| Parameter |                                        Description                                        | Required |
+|:---------:|:-----------------------------------------------------------------------------------------:|:--------:|
+| workflowA | Workflow ID of the first call                                                             | yes      |
+| callA     | Fully qualified name of the first call. **Including workflow name**. (see example below)  | yes      |
+| indexA    | Shard index of the first call                                                             | no       |
```
cjllanwarne: ToL Hmm, presumably this is `yes`, if the call was in a scatter, otherwise `no`? I don't know how to nicely purvey that though

README.md
```diff
@@ -3379,7 +3379,167 @@ Response:
...
+    {
+      "input: String addressee": {
+        "85174842-4a44-4355-a3a9-3a711ce556f1:wf_hello.hello:-1": "D4CC65CB9B5F22D8A762532CED87FE8D",
+        "7479f8a8-efa4-46e4-af0d-802addc66e5d:wf_hello.hello:-1": "7235E005510D99CB4D5988B21AC97B6D"
+      }
+    },
+    {
+      "input: String addressee2": {
+        "85174842-4a44-4355-a3a9-3a711ce556f1:wf_hello.hello:-1": "116C7E36B4AE3EAFD07FA4C536CE092F",
+        "7479f8a8-efa4-46e4-af0d-802addc66e5d:wf_hello.hello:-1": null
```
cjllanwarne: Is this `null` as in `this job didn't have this hash`? I'd prefer the longer form if possible

README.md
```diff
@@ -3379,7 +3379,167 @@ Response:
...
+Content-Type: application/json; charset=UTF-8
+Date: Tue, 06 Jun 2017 16:44:33 GMT
+Server: spray-can/1.3.3
+
+{
+  "callA": {
+    "workflowId": "85174842-4a44-4355-a3a9-3a711ce556f1",
+    "callFqn": "wf_hello.hello",
+    "jobIndex": -1,
+    "allowResultReuse": true
```
cjllanwarne: I'd also add `"successful": true` - since an unexpected `"allowResultReuse": false` could still be hard to diagnose

README.md
Horneth: Oh right I'll try to rephrase

README.md
Horneth: ok

README.md
Horneth: Agreed that the call status would be very nice, the only problem is I don't think we can currently know that without querying metadata ?

Horneth: I think I'm gonna add a simple "callCachingFailures" in the metadata though if a job failed to be cached to with the associated exception. This way it's at least possible to work that out from the metadata

cjllanwarne: Feel free to push that back into a new ticket/PR though

README.md
Horneth: @cjllanwarne on second thoughts, if I had to implement a UI or some processing on top of this response, I'd rather have `null` than a "magic" string I have to look for to know that the job didn't have the hash

CHANGELOG.md
```diff
@@ -1,5 +1,60 @@
...
+    },
+    "callB": {
+        "workflowId": "7479f8a8-efa4-46e4-af0d-802addc66e5d",
+        "callFqn": "wf_hello.hello",
+        "jobIndex": -1,
+        "allowResultReuse": true
+    },
+    "hashDifferential": [
+        {
+            "command template": {
```
geoffjentry: Going back to the convo on whether or not to show the original values, I'm wondering if putting the hashes is useful at all here. While we're not displaying original values, what about just saying which fields didn't match?

README.md
```diff
@@ -3379,7 +3379,167 @@ Response:
...
+The following query parameters are supported:
+
+| Parameter |                                        Description                                        | Required |
+|:---------:|:-----------------------------------------------------------------------------------------:|:--------:|
+| workflowA | Workflow ID of the first call                                                             | yes      |
+| callA     | Fully qualified name of the first call. **Including workflow name**. (see example below)  | yes      |
+| indexA    | Shard index of the first call                                                             | no       |
+| workflowB | Workflow ID of the second call                                                            | yes      |
+| callB     | Fully qualified name of the second call. **Including workflow name**. (see example below) | yes      |
+| indexB    | Shard index of the second call                                                            | no       |
```
mcovarr: Same issue here

database/sql/src/main/scala/cromwell/database/slick/tables/CallCachingHashEntryComponent.scala
```diff
@@ -74,4 +74,54 @@ trait CallCachingHashEntryComponent {
...
+        })
+          // Remove duplicates
+          .distinct
+          // Project only hashKey / hashValues.
+          .map({
+            case (maybeHashA, maybeHashB) =>
+            maybeHashA.map({ hashA => hashA.hashKey -> hashA.hashValue }) -> maybeHashB.map({ hashB => hashB.hashKey -> hashB.hashValue })
+        })
+      } yield hashes
+    }
```
mcovarr: 😮

engine/src/main/resources/swagger/cromwell.yaml
```diff
@@ -384,6 +384,60 @@ paths:
       security:
         - google_oauth:
             - openid
+  '/workflows/{version}/callcaching/diff':
+    get:
+      summary: Return the hash differential between to calls
```
mcovarr: to => two

engine/src/main/resources/swagger/cromwell.yaml
```diff
@@ -384,6 +384,60 @@ paths:
...
+          type: string
+          in: query
+        - name: callA
+          description: Call FQN (including workflow name) of the first call
+          required: true
+          type: string
+          in: query
+        - name: indexA
+          description: Index of the first call
+          required: false
```
mcovarr: Per discussion above, may need to adjust optionality here and in indexB

CHANGELOG.md
Horneth: Oh ok I misunderstood then I thought the conclusion we reached was to leave the hashes. But I agree that they're mostly useless.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheDiffQueryParameter.scala
```diff
@@ -0,0 +1,54 @@
...
+
+    val callAValidation = extractAttribute(parameters, "callA")
+    val callBValidation = extractAttribute(parameters, "callB")
+
+    val indexAValidation = extractIndex(parameters, "indexA")
+    val indexBValidation = extractIndex(parameters, "indexB")
+
+    (workflowAValidation |@| callAValidation |@| indexAValidation |@|
+      workflowBValidation |@| callBValidation |@| indexBValidation).tupled match {
+      case Valid((workflowA, callA, indexA, workflowB, callB, indexB)) =>
```
danbills: ToL: why not return a `ValidatedNel`?  Could replace the `tupled match`  here w/ map, allow a `fold` instead of another pattern match where this is used.

I understand that failBadRequest takes an exception but you could worry about that when necessary.

backend/src/main/scala/cromwell/backend/standard/callcaching/StandardCacheHitCopyingActor.scala
```diff
@@ -92,26 +91,31 @@ abstract class StandardCacheHitCopyingActor(val standardParams: StandardCacheHit
...
+          processed match {
+            case Success((callOutputs, destinationDetritus, allCopyPairs)) =>
+              duplicate(allCopyPairs) match {
+                case Some(Success(_)) => succeedAndStop(returnCode, callOutputs, destinationDetritus)
+                case Some(Failure(failure)) => failAndStop(failure)
+                case None =>
+                  val allCopyCommands = allCopyPairs map { case (source, destination) => copyCommand(source, destination, overwrite = true) }
+
+                  allCopyCommands foreach {
+                    sendIoCommand(_)
```
danbills: should be able to write `allCopyCommands foreach sendIoCommand`

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/ejea/EjeaRunningJobSpec.scala
```diff
@@ -64,30 +64,64 @@ class EjeaRunningJobSpec extends EngineJobExecutionActorSpec with Eventually wit
...
-        ejea ! response
-
-        helper.jobStoreProbe.expectMsgPF(max = awaitTimeout, hint = "Job Store Write") {
-          case RegisterJobCompleted(jobKey, JobResultFailure(returnCode, reason, isRetryable)) =>
-            validateJobStoreKey(jobKey)
-            returnCode should be(failedRc)
-            reason should be(failureReason)
-            isRetryable should be(retryable)
-        }
+    s"register 'FailedRetryableResponse's with the JobStore" in {
```
cjllanwarne: I assume the decision to remove the `FailedNonRetryableResponse` test was deliberate?

CHANGELOG.md
cjllanwarne: Yeah, unless you have a rainbow table of MD5s for common integer values, they're essentially meaningless.

Except... they *do* look authoritative: 
> Oh yeah, the hashes *are* different, see, that one starts with `a5` and that one starts with `b7`...

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -218,41 +219,60 @@ class EngineJobExecutionActor(replyTo: ActorRef,
...
       log.debug(s"Got job result for {}, awaiting hashes", jobTag)
       stay using data.withSuccessResponse(response)
+        // writeToCache is OFF - complete the job
     case Event(response: JobSucceededResponse, data: ResponsePendingData) =>
       eventList ++= response.executionEvents
       saveJobCompletionToJobStore(data.withSuccessResponse(response))
+      
+      // Non-Retryable failure
+        // All hashes already retrieved - save to the cache
+    case Event(response: JobFailedNonRetryableResponse, data @ ResponsePendingData(_, _, Some(Success(hashes)), _, _)) if effectiveCallCachingMode.writeToCache =>
```
mcovarr: yikes, these cases were just not being handled before?

README.md
cjllanwarne: Yeah, kind of... I guess my complaint is that it's still a magic value, just not a humanly readable one... but I'm happy for you to leave it, especially since it **is** excellently documented :)

backend/src/main/scala/cromwell/backend/standard/callcaching/StandardCacheHitCopyingActor.scala
mcovarr: There are two args and the second is defaulted, so that doesn't work.

database/sql/src/main/scala/cromwell/database/slick/tables/CallCachingHashEntryComponent.scala
cjllanwarne: I'm curious what shocked you here...

engine/src/main/resources/swagger/cromwell.yaml
cjllanwarne: Personally, I'm happy for the API field to be optional as long as it's spelled out in the docs when it's expected

database/sql/src/main/scala/cromwell/database/slick/tables/CallCachingHashEntryComponent.scala
mcovarr: It's some pretty serious Slickery.  I'm curious / frightened to see what kind of SQL this gets compiled into.  😄

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
cjllanwarne: Before, we didn't need to separate out the cases of a failure coming back with hashes and without hashes.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -479,9 +499,16 @@ class EngineJobExecutionActor(replyTo: ActorRef,
...
 
-  private def saveCacheResults(hashes: CallCacheHashes, data: SucceededResponseData) = {
-    callCacheWriteActor ! SaveCallCacheHashes(workflowIdForLogging, hashes, data)
-    val updatedData = data.copy(hashes = Option(Success(hashes)))
+  private def saveCacheResults(hashes: CallCacheHashes, data: CacheWriteResponseData) = {
+    data match {
+      case jobSucceededData: SucceededResponseData =>
+        callCacheWriteActor ! SaveCallCacheHashes(CallCacheHashBundle(workflowIdForLogging, hashes, jobSucceededData.response))
+      case jobFailedNonRetryableData: FailedNonRetryableResponseData =>
+        callCacheWriteActor ! SaveCallCacheHashes(CallCacheHashBundle(workflowIdForLogging, hashes, jobFailedNonRetryableData.response))
```
mcovarr: it looks like these two branches are actually the same except for the metadata write on the `FailedNoRetryableResponseData`?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCache.scala
```diff
@@ -98,8 +98,54 @@ class CallCache(database: CallCachingSqlDatabase) {
...
+        jobFailedNonRetryableResponse.jobKey.index,
+        Option(jobFailedNonRetryableResponse.jobKey.attempt),
+        None,
+        false,
+        Map.empty[LocallyQualifiedName, JobOutput],
+        None
+      )
+    }
+  }
+  case class CallCacheHashBundle private (
```
cjllanwarne: eg. https://forums.manning.com/posts/list/38567.page

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -607,16 +637,26 @@ object EngineJobExecutionActor {
...
+  private [execution] sealed trait CacheWriteResponseData extends ResponseData
+  
+  private[execution] case class SucceededResponseData(response: JobSucceededResponse,
+                                                      hashes: Option[Try[CallCacheHashes]] = None) extends CacheWriteResponseData {
+    override def dockerImageUsed = response.dockerImageUsed
+    override def withHashes(hashes: Try[CallCacheHashes]) = this.copy(hashes = Option(hashes))
   }
 
+  private[execution] case class FailedNonRetryableResponseData(response: JobFailedNonRetryableResponse,
+                                                         hashes: Option[Try[CallCacheHashes]] = None) extends CacheWriteResponseData {
```
mcovarr: formatting is a bit off here

CHANGELOG.md
geoffjentry: @Horneth I think you're right that that's what we said, it was just when i saw it here that I realized it was a lot of verbosity for no real benefit.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCache.scala
```diff
@@ -98,8 +98,54 @@ class CallCache(database: CallCachingSqlDatabase) {
...
+      new CallCacheHashBundle(
+        workflowId,
+        callCacheHashes,
+        jobFailedNonRetryableResponse.jobKey.scope.fullyQualifiedName,
+        jobFailedNonRetryableResponse.jobKey.index,
+        Option(jobFailedNonRetryableResponse.jobKey.attempt),
+        None,
+        false,
+        Map.empty[LocallyQualifiedName, JobOutput],
+        None
```
mcovarr: This is making me wish for keyword args

backend/src/main/scala/cromwell/backend/standard/callcaching/StandardCacheHitCopyingActor.scala
```diff
@@ -154,11 +158,11 @@ abstract class StandardCacheHitCopyingActor(val standardParams: StandardCacheHit
     stay()
   }
 
-  private def lookupSourceCallRootPath(sourceJobDetritusFiles: Map[String, String]): Path = {
-    sourceJobDetritusFiles.get(JobPaths.CallRootPathKey).map(getPath).get recover {
-      case failure =>
-        throw new RuntimeException(s"${JobPaths.CallRootPathKey} wasn't found for call ${jobDescriptor.call.fullyQualifiedName}", failure)
-    } get
+  private def lookupSourceCallRootPath(sourceJobDetritusFiles: Map[String, String]): Try[Path] = {
```
danbills: Similar to thing above, I'd just return Option here (no chance of exception) and let the caller worry about what to do with it

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheDiffQueryParameter.scala
```diff
@@ -0,0 +1,54 @@
...
+  private def missingWorkflowError(attribute: String) = s"missing $attribute query parameter".invalidNel
+
+  private def extractAttribute(parameters: Seq[(String, String)], parameter: String) = {
+    parameters.find(_._1 == parameter) match {
+      case Some((_, value)) => value.validNel
+      case None => missingWorkflowError(parameter)
+    }
+  }
+
+  private def extractIndex(parameters: Seq[(String, String)], parameter: String) = {
```
mcovarr: return type here too plz

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheReadActor.scala
```diff
@@ -47,6 +52,41 @@ class CallCacheReadActor(cache: CallCache) extends Actor with ActorLogging {
...
+            case (None, Some((keyB, valueB))) =>
+              Map(
+                keyB -> Map(
+                  callATag -> None,
+                  callBTag -> Option(valueB)
+                )
+              )
+            // Should not be possible...
+            case (None, None) =>
+              Map.empty[String, Map[String, Option[String]]]
```
mcovarr: The value here is starting to smell like a type

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheDiffDatabaseSpec.scala
```diff
@@ -0,0 +1,219 @@
...
+    it should "fail properly if the cache entry for call A is not found" taggedAs DbmsTest in {
+      val shouldFail = for {
+        _ <- dataAccess.diffCallCacheHashes(
+          "does not", "exist", 0,
+          callCachingEntryB.workflowExecutionUuid, callCachingEntryB.callFullyQualifiedName, callCachingEntryB.jobIndex
+        )
+      } yield ()
+      
+      ScalaFutures.whenReady(shouldFail.failed) { e =>
+        e shouldBe a [Exception]
```
mcovarr: a => an

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheReadActor.scala
geoffjentry: `case class` not `type` ;)

mcovarr: yeah type not `type`

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheDiffDatabaseSpec.scala
```diff
@@ -0,0 +1,219 @@
...
+    it should "fail properly if the cache entry for call B is not found" taggedAs DbmsTest in {
+      val shouldFail = for {
+        _ <- dataAccess.diffCallCacheHashes(
+          callCachingEntryB.workflowExecutionUuid, callCachingEntryB.callFullyQualifiedName, callCachingEntryB.jobIndex,
+          "does not", "existB", 0
+        )
+      } yield ()
+
+      ScalaFutures.whenReady(shouldFail.failed) { e =>
+        e shouldBe a [Exception]
```
mcovarr: an

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheDiffDatabaseSpec.scala
```diff
@@ -0,0 +1,219 @@
...
+    it should "fail properly if none of the cache entries are found " taggedAs DbmsTest in {
+      val shouldFail = for {
+        _ <- dataAccess.diffCallCacheHashes(
+          "does not", "exist", 0,
+          "does not", "exist either", 0
+        )
+      } yield ()
+
+      ScalaFutures.whenReady(shouldFail.failed) { e =>
+        e shouldBe a [Exception]
```
mcovarr: etc

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/ejea/EjeaRunningJobSpec.scala
Horneth: Yes, this test ensure that we go into "UpadtingJobStore" mode when getting a `FailedNonRetryableResponse`, which we don't do anymore since we go into "write to cache" mode.

mcovarr:
:+1: with suggested changes and API client addition

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCache.scala
Horneth: Oh I see, the reason for the private was to try to make it hard to mess up the value of "allowedResultReuse" when creating a `CallCacheHashBundle`, as this is gonna be even more important now as we really don't want failed jobs to be cacheable. It at least adds *some* safety I guess.

database/sql/src/main/scala/cromwell/database/slick/tables/CallCachingHashEntryComponent.scala
Horneth: @mcovarr I can dig that up from the mysql log

CHANGELOG.md
Horneth: @geoffjentry If we do remove them, I think we can't leave the format as it is then or it would look awkward. I'm not sure what the new format should be then. I think it should be easy to see if both calls have the key but with different hashes, or if just one of them has it. Any idea ?

engine/src/main/resources/swagger/cromwell.yaml
Horneth: @mcovarr Just to be clear, if it is required it means we would always have to give "-1" for non sharded calls right ? I honestly don't know which one is more natural

mcovarr: @Horneth yeah you're right, please disregard; we don't want that.

database/sql/src/main/scala/cromwell/database/slick/CallCachingSlickDatabase.scala
```diff
@@ -4,12 +4,23 @@ import cats.data.NonEmptyList
...
 import scala.concurrent.{ExecutionContext, Future}
 
+object CallCachingSlickDatabase {
+  case class CacheEntriesNotFoundException(queries: NonEmptyList[String]) extends Exception(
+    if (queries.tail.isEmpty) {
+      s"Cannot find a cache entry for ${queries.head}"
+    } else {
+      s"Cannot find cache entries for ${queries.head}, ${queries.tail.mkString(",")}"
+    }
+  )
```
Horneth: Not sure what you mean ?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
Horneth: yes, except that `SucceededResponseData` and `FailedNonRetryableResponseData` both override their `response` field with a refined type of `BackendJobExecutionResponse` (respectively `JobSucceededResponse` and `JobFailedNonRetryableResponse`) which are the only ones a `CallCacheHashBundle` can be built from.  This avoids having to create yet another sealed trait that only `JobSucceededResponse` and `JobFailedNonRetryableResponse` extend from.

backend/src/main/scala/cromwell/backend/standard/callcaching/StandardCacheHitCopyingActor.scala
Horneth: `getPath` does return a `Try[Path]` though already so it would be an `Option[Try[Path]]`

README.md
Horneth: Done

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheDiffQueryParameter.scala
```diff
@@ -0,0 +1,54 @@
...
+        Success(CallCacheDiffQueryParameter(
+          CallCacheDiffQueryCall(workflowA, callA, indexA),
+          CallCacheDiffQueryCall(workflowB, callB, indexB)
+        ))
+      case Invalid(errors) => Failure(AggregatedMessageException("Wrong parameters for call cache diff query", errors.toList))
+    }
+  }
+}
+
+case class CallCacheDiffQueryParameter(callA: CallCacheDiffQueryCall, callB: CallCacheDiffQueryCall)
```
Horneth: done

README.md
Horneth: https://github.com/broadinstitute/cromwell/issues/2338

database/sql/src/main/scala/cromwell/database/slick/tables/CallCachingHashEntryComponent.scala
Horneth: I tried to format it a bit but it is pretty intense..

```sql
select x2.x3, x2.x4, x2.x5, x2.x6, x2.x7, x2.x8
from (
       select distinct x9.x10 as x11, x9.x12 as x13, x9.x14 as x5, x9.x15 as x16, x9.x17 as x8, x9.x18 as x19, x9.x20 as x4, x9.x21 as x3, x9.x22 as x6, x9.x23 as x7
       from (
              select (case when (x24.x25 is null) then null else 1 end) as x21,
                     x24.x26 as x20,
                     x24.x25 as x14,
                     x24.x27 as x15,
                     x24.x28 as x18,
                     (case when (x29.x30 is null) then null else 1 end) as x22,
                     x29.x31 as x23,
                     x29.x30 as x17,
                     x29.x32 as x12,
                     x29.x33 as x10
              from (
                     select `CALL_CACHING_ENTRY_ID` as x27,
                            `HASH_VALUE` as x25,
                            `HASH_KEY` as x26,
                            `CALL_CACHING_HASH_ENTRY_ID` as x28
                     from `CALL_CACHING_HASH_ENTRY` where `CALL_CACHING_ENTRY_ID` = 39) x24
                left outer join (
                                  select `CALL_CACHING_ENTRY_ID` as x32, `HASH_VALUE` as x30, `HASH_KEY` as x31, `CALL_CACHING_HASH_ENTRY_ID` as x33
                                  from `CALL_CACHING_HASH_ENTRY`
                                  where `CALL_CACHING_ENTRY_ID` = 42
                                ) x29
                  on x24.x26 = x29.x31
              where (((x24.x25 is null) and (x29.x30 is not null))
                     or ((x24.x25 is not null) and (x29.x30 is null)))
                    or ifnull((case when (x24.x25 is not null) then (case when (x29.x30 is not null) then (not (x24.x25 = x29.x30)) else null end) else null end),false)
              union all select (
                                 case when (null is null) then null else 1 end) as x21,
                               null as x20,
                               null as x14,
                               null as x15,
                               null as x18,
                               (case when (x34.`HASH_VALUE` is null) then null else 1 end) as x22,
                               x34.`HASH_KEY` as x23,
                               x34.`HASH_VALUE` as x17,
                               x34.`CALL_CACHING_ENTRY_ID` as x12,
                               x34.`CALL_CACHING_HASH_ENTRY_ID` as x10 from
                          `CALL_CACHING_HASH_ENTRY` x34
                        where ((x34.`CALL_CACHING_ENTRY_ID` = 42)
                               and (not exists(select `HASH_VALUE`, `HASH_KEY`, `CALL_CACHING_ENTRY_ID`, `CALL_CACHING_HASH_ENTRY_ID`
                                               from `CALL_CACHING_HASH_ENTRY`
                                               where (`CALL_CACHING_ENTRY_ID` = 39)
                                                     and (`HASH_KEY` = x34.`HASH_KEY`))))
                              and ((((null is null) and (x34.`HASH_VALUE` is not null))
                                    or ((null is not null) and (x34.`HASH_VALUE` is null)))
                                   or ifnull((case when (null is not null) then (case when (x34.`HASH_VALUE` is not null) then (not (null = x34.`HASH_VALUE`)) else null end) else null end),false))) x9) x2
```

cjllanwarne: ... I agree with 😮...

geoffjentry:
@Horneth check out the comment i left on the issue you filed. Serious seagull activity from me.

database/sql/src/main/scala/cromwell/database/slick/CallCachingSlickDatabase.scala
cjllanwarne: It's just a lot of lines between the start of `Exception(...` and the `...)

cjllanwarne:
👍, I philosophically agree with @geoffjentry's comment but I think it's find to do that in some subsequent ticket

geoffjentry:
Does "some subsequent" mean "real soon" or "it goes into the pile of issues never to be seen again"?

Horneth:
If we're going the metadata route, which I think I like too, given that w're ok with the fact that it will likely nearly double the size of the full metadata response, I think there's no point merging this because half of it will be deleted in the next PR no ?

database/sql/src/main/scala/cromwell/database/slick/tables/CallCachingHashEntryComponent.scala
Horneth: tbh pretty much any "complex" slick query will look like this

cjllanwarne:
Haha, excellent points!

So, 
- If it's going on the pile of "TBDs" then I'd merge this PR as a handy stop-gap. 
- If it's the next ticket you look at then, yeah, I'd just go straight for it.

geoffjentry:
@cjllanwarne I only want "a handy stopgap" if there's an actual plan to fix it. Things have a way of sticking around forever.

geoffjentry:
@Horneth re the size of the metadata response, my $0.02 is people should be pulling from curated views instead of the full metadata response - there just haven't been all that many differentiable use cases to date. At some point that'll change.

Horneth:
@geoffjentry yes I think you're right, I was mostly thinking about the size of the metadata table that is already approaching concerning limits in GOTC

geoffjentry:
@Horneth  Ah, in that case it helps push us towards coming up with a better scheme than how we're currently storing it all :)

Horneth:
Closing this for now, will reopen with the metadata version

mcovarr:
victorious seagull

PR #2342, opened 2017-06-08
First pass of workflow type and version.  

mcovarr:
Takes parameters on REST endpoints, roundtrips to the workflow store, and publishes to metadata.  Currently hardcodes type and version for the command line.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ before_install:
   - openssl aes-256-cbc -K "$encrypted_5ebd3ff04788_key" -iv "$encrypted_5ebd3ff04788_iv" -in src/bin/travis/resources/jesConf.tar.enc -out jesConf.tar -d || true
 env:
   global:
-    - CENTAUR_BRANCH=mlc_workflow_type
+    - CENTAUR_BRANCH=mlc_wdl_b_gone
   matrix:
     # Setting this variable twice will cause the 'script' section to run twice with the respective env var invoked
     - BUILD_TYPE=sbt
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ before_install:
   - openssl aes-256-cbc -K "$encrypted_5ebd3ff04788_key" -iv "$encrypted_5ebd3ff04788_iv" -in src/bin/travis/resources/jesConf.tar.enc -out jesConf.tar -d || true
 env:
   global:
-    - CENTAUR_BRANCH=develop
+    - CENTAUR_BRANCH=mlc_workflow_type
   matrix:
     # Setting this variable twice will cause the 'script' section to run twice with the respective env var invoked
     - BUILD_TYPE=sbt
--- a/core/src/main/scala/cromwell/core/WorkflowMetadataKeys.scala
+++ b/core/src/main/scala/cromwell/core/WorkflowMetadataKeys.scala
@@ -20,5 +20,8 @@ object WorkflowMetadataKeys {
   val SubmissionSection_Inputs = "inputs"
   val SubmissionSection_Options = "options"
   val SubmissionSection_Imports = "imports"
+  val SubmissionSection_WorkflowType = "workflowType"
+  val SubmissionSection_WorkflowTypeVersion = "workflowTypeVersion"
+
   val Labels = "labels"
 }
--- a/database/migration/src/main/resources/changesets/workflow_store_type_and_version.xml
+++ b/database/migration/src/main/resources/changesets/workflow_store_type_and_version.xml
@@ -14,10 +14,4 @@
         </addColumn>
     </changeSet>
 
-    <changeSet author="mcovarr" id="not-null-workflow-store-type">
-        <addNotNullConstraint columnDataType="VARCHAR(30)"
-                              columnName="WORKFLOW_TYPE"
-                              defaultNullValue="WDL"
-                              tableName="WORKFLOW_STORE_ENTRY"/>
-    </changeSet>
 </databaseChangeLog>
--- a/database/migration/src/main/resources/changesets/workflow_store_type_and_version.xml
+++ b/database/migration/src/main/resources/changesets/workflow_store_type_and_version.xml
@@ -1,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.3.xsd">
+
+    <changeSet author="mcovarr" id="workflow-store-type-and-version">
+        <addColumn tableName="WORKFLOW_STORE_ENTRY">
+            <column name="WORKFLOW_TYPE" type="VARCHAR(30)">
+                <constraints nullable="true"/>
+            </column>
+            <column name="WORKFLOW_TYPE_VERSION" type="VARCHAR(255)">
+                <constraints nullable="true"/>
+            </column>
+        </addColumn>
+    </changeSet>
+
+    <changeSet author="mcovarr" id="not-null-workflow-store-type">
+        <addNotNullConstraint columnDataType="VARCHAR(30)"
+                              columnName="WORKFLOW_TYPE"
+                              defaultNullValue="WDL"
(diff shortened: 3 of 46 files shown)


.travis.yml
```diff
@@ -19,7 +19,7 @@ before_install:
   - openssl aes-256-cbc -K "$encrypted_5ebd3ff04788_key" -iv "$encrypted_5ebd3ff04788_iv" -in src/bin/travis/resources/jesConf.tar.enc -out jesConf.tar -d || true
 env:
   global:
-    - CENTAUR_BRANCH=develop
+    - CENTAUR_BRANCH=mlc_workflow_type
```
mcovarr: will revert before merge (promise)

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
```diff
@@ -137,12 +137,27 @@ trait CromwellApiService extends HttpService with PerRequestCreator {
...
     workflowInputs: Vector[WdlJson],
     workflowInputsAux: Map[Int, WdlJson],
     workflowOptions: Option[WorkflowOptionsJson],
     customLabels: Option[WdlJson],
     zippedImports: Option[Array[Byte]])
+
   object PartialWorkflowSources {
+    def empty = PartialWorkflowSources(
+      wdlSource = None,
+      // TODO do not hardcode, especially not out here at the boundary layer good gravy
```
geoffjentry: maxresdefault

core/src/main/scala/cromwell/core/WorkflowMetadataKeys.scala
```diff
@@ -20,5 +20,8 @@ object WorkflowMetadataKeys {
   val SubmissionSection_Inputs = "inputs"
   val SubmissionSection_Options = "options"
   val SubmissionSection_Imports = "imports"
+  val SubmissionSection_WorkflowType = "workflowType"
```
Horneth: is `Type` the official term (rather than `Language` ?)

database/migration/src/main/resources/changesets/workflow_store_type_and_version.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.3.xsd">
+
+    <changeSet author="mcovarr" id="workflow-store-type-and-version">
+        <addColumn tableName="WORKFLOW_STORE_ENTRY">
+            <column name="WORKFLOW_TYPE" type="VARCHAR(30)">
```
ruchim: Curious--why such a large data type for workflow_type and workflow_type_version?

database/migration/src/main/resources/changesets/workflow_store_type_and_version.xml
```diff
@@ -0,0 +1,23 @@
...
+                <constraints nullable="true"/>
```
Horneth: Why is `nullable` `true` here if we add a not null constraint after ?

database/sql/src/main/scala/cromwell/database/slick/tables/WorkflowStoreEntryComponent.scala
```diff
@@ -15,6 +15,10 @@ trait WorkflowStoreEntryComponent {
 
     def workflowExecutionUuid = column[String]("WORKFLOW_EXECUTION_UUID")
 
+    def workflowType = column[String]("WORKFLOW_TYPE", O.Length(30))
+
+    def workflowTypeVersion = column[Option[String]]("WORKFLOW_TYPE_VERSION")
```
Horneth: No `O.Length` for this one ?

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreSubmitActor.scala
```diff
@@ -87,7 +87,13 @@ final case class WorkflowStoreSubmitActor(store: WorkflowStore, serviceRegistryA
         MetadataEvent(MetadataKey(id, None, WorkflowMetadataKeys.SubmissionSection, WorkflowMetadataKeys.SubmissionSection_Options), MetadataValue(sourceFiles.workflowOptionsJson))
       )
 
-      serviceRegistryActor ! PutMetadataAction(submissionEvents)
+      // Don't publish metadata for either workflow type or workflow type version if not defined.
+      val workflowTypeAndVersionEvents: List[Option[MetadataEvent]] = List(
+        sourceFiles.workflowType map { wt => MetadataEvent(MetadataKey(id, None, WorkflowMetadataKeys.SubmissionSection, WorkflowMetadataKeys.SubmissionSection_WorkflowType), MetadataValue(wt)) },
```
Horneth: I thought workflow type was not nullable so how can we not have one ?

Horneth:
👍  with some nitpickery

database/migration/src/main/resources/changesets/workflow_store_type_and_version.xml
ruchim: Do we want the workflow type to be nullable?

ruchim:
I guess that first thumb didn't count....here's a second one! :+1:

core/src/main/scala/cromwell/core/WorkflowMetadataKeys.scala
mcovarr: Yes, this is [WES](https://github.com/ga4gh/workflow-execution-schemas/blob/develop/swagger/proto/workflow_execution.swagger.json#L409) terminology.

database/migration/src/main/resources/changesets/workflow_store_type_and_version.xml
mcovarr: Hmm I think I forgot to add something to backfill with "WDL".  😦

mcovarr: Oh nm I did do it right with `defaultNullValue` 😌

mcovarr: see above/below 😄

database/sql/src/main/scala/cromwell/database/slick/tables/WorkflowStoreEntryComponent.scala
mcovarr: Apparently no `O.Length` for `VARCHAR(255)` is the Cromwell Way, cuz that's how everything else is.  :man_shrugging:

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreSubmitActor.scala
mcovarr: There may not have been a workflow type specified in the submission.  Ultimately we'll need to default to a workflow type if none is specified, but not at this layer.

database/migration/src/main/resources/changesets/workflow_store_type_and_version.xml
mcovarr: :man_shrugging: are there values you think would be more reasonable?

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreSubmitActor.scala
mcovarr: Hmm actually now that I think about it, it's not clear exactly how this should work.  I think we should record both the specified workflowType and workflowTypeVersion, but we should also record the effective values if either or both of those are unspecified.  I may make some changes here.  🤔

mcovarr:
@Horneth @ruchim   A quick review of the most recent commit would be much appreciated.  After reflecting on the purpose of the workflow store as the means of quickly capturing workflow submissions for future processing, I made workflow type optional.  This defers the problem of determining the workflow type if otherwise unspecified to materialization time, which I think is where it more properly belongs.  I also fixed some spots I forgot to de-WDL in the Swagger (oops).

ruchim:
Good catch on the swagger :+1:

geoffjentry:
Already merged but 👍 on the judgement call on pushing the WF type to materialization. It goes along w/ our notion of just dumping it in and worrying about what's there later

PR #2360, opened 2017-06-15
Make Cromwell 27 Docker hashing consistent with 26

mcovarr:
Specifically, don't force on a `library/` prefix where none was explicitly specified.

This looks ok in the debugger but I haven't figured out a good test for it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ before_install:
   - openssl aes-256-cbc -K "$encrypted_5ebd3ff04788_key" -iv "$encrypted_5ebd3ff04788_iv" -in src/bin/travis/resources/jesConf.tar.enc -out jesConf.tar -d || true
 env:
   global:
-    - CENTAUR_BRANCH=989e3d73ca72567398fd2c9fb03dad57d9cb69f5
+    - CENTAUR_BRANCH=df765d2f8b0a4e34d1dd59eb36a847942e9ed151
   matrix:
     # Setting this variable twice will cause the 'script' section to run twice with the respective env var invoked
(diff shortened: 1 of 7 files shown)


geoffjentry:
👍 

For my own edification can you explain more what the original issue was? Not sure I fully understand

mcovarr:
Docker requires all image lookups to be qualified by a repository name.  For Broad images our repository is "broadinstitute", and in our WDLs we request Broad images with the repository explicitly specified a la  "broadinstitute/genomes-in-the-cloud".  The more universal images like "ubuntu" don't require an explicit repository specification in casual parlance, but still require a default repository specification of "library" for hash lookups.

When a Docker image is specified in WDL that includes a hash, Cromwell skips the hash lookup and just hashes the image name.  Cromwell 26 would use essentially the Docker image string that had been specified in the WDL, i.e. something like `ubuntu@sha256:ea1d854d38be82f54d39efe2c67000bed1b03348bcc2f3dc094f260855dff368`.  Cromwell 27 inadvertently changed the internal representation of the image string to prepend the `library/` repository even when it hadn't been explicitly specified in the WDL.  This meant that the Docker string would hash differently on 27 than it did on 26, resulting in unwanted cache misses.  These changes look to track whether a repository has been explicitly specified or not, and only prepend the `library/` where required on hash lookups.

geoffjentry:
@mcovarr Ah thanks. Not sure which parts were the blanks but you managed to fill them in, tie things together, or whatever it was that I was missing :)

Horneth:
LGTM 👍

PR #2361, opened 2017-06-15
27 / 26 Docker hash consistency, develop edition.

mcovarr:
Don't implicitly add a library/ repo prefix to Docker images as that breaks 26-era call caches.


Horneth:
👍

PR #2364, opened 2017-06-16
Fail on bad GCS bucket names, particularly those containing underscores.

mcovarr:


ruchim:
:+1:

filesystems/gcs/src/main/scala/cromwell/filesystems/gcs/GcsPathBuilder.scala
```diff
@@ -37,8 +38,36 @@ object GcsPathBuilder {
...
+  case object ValidFullGcsPath extends GcsPathValidation
+  case object PossiblyValidRelativeGcsPath extends GcsPathValidation
+  sealed trait InvalidGcsPath extends GcsPathValidation {
+    def pathString: String
+    def errorMessage: String
+  }
+  final case class InvalidFullGcsPath(pathString: String) extends InvalidGcsPath {
+    override def errorMessage = {
+      val prefix = List(
+        s"The bucket name in GCS path '$pathString' is not compatible with",
```
Horneth: +1, one could even add that Google discourages the use of invalid DNS host names for bucket names:

> Also, for DNS compliance and future compatibility, you should not use underscores (_) or have a period adjacent to another period or dash. For example, ".." or "-." or ".-" are not valid in DNS names.

https://cloud.google.com/storage/docs/naming

Horneth:
👍

filesystems/gcs/src/main/scala/cromwell/filesystems/gcs/GcsPathBuilder.scala
mcovarr: Awesome link, thank you!

PR #2370, opened 2017-06-21
travis deprecated image

ruchim:
...because we're seeing failures with gsutil with the newest image. Need to fix that for sure, but this is more like a temporary patch.

### DIFF
--- a/src/bin/travis/testCentaurJes.sh
+++ b/src/bin/travis/testCentaurJes.sh
@@ -55,8 +55,8 @@ export PYTHONPATH="/usr/lib/python2.7/site-packages:/usr/local/lib/python2.7/sit
 export CONFIGURE_OPTS="--enable-unicode=ucs4"
 pyenv install 2.7.10
 pyenv global 2.7.10
-sudo pip install --upgrade pip
-sudo pip install pyopenssl ndg-httpsclient pyasn1 --upgrade
+sudo -H pip install --upgrade pip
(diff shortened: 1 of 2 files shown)


Horneth:
👍

src/bin/travis/testCentaurJes.sh
```diff
@@ -55,8 +55,8 @@ export PYTHONPATH="/usr/lib/python2.7/site-packages:/usr/local/lib/python2.7/sit
 export CONFIGURE_OPTS="--enable-unicode=ucs4"
 pyenv install 2.7.10
 pyenv global 2.7.10
-sudo pip install --upgrade pip
-sudo pip install pyopenssl ndg-httpsclient pyasn1 --upgrade
+sudo -H pip install --upgrade pip
+sudo -H pip install pyopenssl ndg-httpsclient pyasn1 --upgrade
```
mcovarr: why the `-H` addition?

ruchim:
In case you're wondering: https://stackoverflow.com/questions/43623025/what-does-sudo-h-do

mcovarr:
:+1:

src/bin/travis/testCentaurJes.sh
ruchim: In the Travis logs there was a warning flashing by to add a "-H" flag for the pip commands. Left a link the comments for what the flag actually does.

PR #2371, opened 2017-06-22
only check stderr if failOnStderr is true

mcovarr:
http://gatkforums.broadinstitute.org/wdl/discussion/9821/wdl-fail-with-nosuchfileexception-but-the-file-exists


Horneth:
👍

geoffjentry:
vlcsnap-2013-02-22-18h38m56s42

Kidding .... 👍

PR #2372, opened 2017-06-23
Return 404 on CC Endpoint for old workflows + cleanup from previous version

Horneth:
- There's no migration to move hashes from previous workflows to metadata. Which means the Call Caching Diff endpoint won't work for those. Instead of returning an empty diff which could be mistaken for "those calls are identical", return 404.
@katevoss @bradtaylor If that's a behavior we want to change please say so (cf email sent earlier this week).

- Reverts the code that was persisting failed jobs hashes to the hash table. We don't need it anymore because it's in the metadata and that's what the Call Caching endpoint is using.

- Update changelog / README

### DIFF
--- a/README.md
+++ b/README.md
@@ -3493,6 +3493,10 @@ Server: spray-can/1.3.3
 
 ## GET /api/workflows/:version/callcaching/diff
 
+**Disclaimer**: This endpoint depends on hash values being published ito the metadata, which only happens as of Crowmell 28.
+Workflows run with prior versions of Cromwell will not be able to benefit from this endpoint.
+A `404 NotFound` will be returned when trying to use this endpoint if either workflows have been run on a prior version.
+
 This endpoint returns the hash differences between 2 *completed* (successfully or not) calls.
 The following query parameters are supported:
 
--- a/README.md
+++ b/README.md
@@ -3493,7 +3493,7 @@ Server: spray-can/1.3.3
 
 ## GET /api/workflows/:version/callcaching/diff
 
-**Disclaimer**: This endpoint depends on hash values being published ito the metadata, which only happens as of Crowmell 28.
+**Disclaimer**: This endpoint depends on hash values being published to the metadata, which only happens as of Crowmell 28.
 Workflows run with prior versions of Cromwell will not be able to benefit from this endpoint.
 A `404 NotFound` will be returned when trying to use this endpoint if either workflows have been run on a prior version.
 
--- a/README.md
+++ b/README.md
@@ -3494,8 +3494,8 @@ Server: spray-can/1.3.3
 ## GET /api/workflows/:version/callcaching/diff
 
 **Disclaimer**: This endpoint depends on hash values being published to the metadata, which only happens as of Crowmell 28.
-Workflows run with prior versions of Cromwell will not be able to benefit from this endpoint.
-A `404 NotFound` will be returned when trying to use this endpoint if either workflows have been run on a prior version.
+Workflows run with prior versions of Cromwell cannot be used with this endpoint.
+A `404 NotFound` will be returned when trying to use this endpoint if either workflow has been run on a prior version.
 
 This endpoint returns the hash differences between 2 *completed* (successfully or not) calls.
 The following query parameters are supported:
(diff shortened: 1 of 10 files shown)


cjllanwarne:
ToL: I was wondering whether 404 is the right code here, but I think it probably is... 

My concern is that "can't diff because your job is pre-28" should be distinguishable from "can't diff because your job doesn't exist"

Horneth:
@cjllanwarne Good point I hadn't thought of that. There might be a better code to return than 404 I'll take a look.

mcovarr:
That is a good point and I also agree I didn't see a more appropriate code unfortunately...

geoffjentry:
There's alway 418 :)

More serious: 410?

Horneth:
410 as I understand it is for something that was there but is not anymore and is not coming back. Which doesn't really fit either IMO :/ But I don't see a better one either..

geoffjentry:
@Horneth admittedly it's a massive stretch but I was rationalizing it as in it's gone because we declined to store the information in the first place ;)

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/ejea/EjeaBackendIsCopyingCachedOutputsSpec.scala
```diff
@@ -112,7 +112,7 @@ class EjeaBackendIsCopyingCachedOutputsSpec extends EngineJobExecutionActorSpec
           s"not invalidate a call for caching if backend coping failed when invalidation is disabled, when it was going to receive $hashComboName, if call caching is $mode" in {
             val invalidationDisabledOptions = CallCachingOptions(invalidateBadCacheResults = false)
             val cacheInvalidationDisabledMode = mode match {
-              case CallCachingActivity(rw, _) => CallCachingActivity(rw, invalidationDisabledOptions)
+              case CallCachingActivity(rw, options) => CallCachingActivity(rw, invalidationDisabledOptions)
```
mcovarr: why is this named now if not used?

README.md
```diff
@@ -3461,6 +3461,10 @@ Server: spray-can/1.3.3
 
 ## GET /api/workflows/:version/callcaching/diff
 
+**Disclaimer**: This endpoint depends on hash values being published to the metadata, which only happens as of Crowmell 28.
+Workflows run with prior versions of Cromwell will not be able to benefit from this endpoint.
```
mcovarr: will not be able to benefit from => cannot be used with

mcovarr:
:+1:

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/ejea/EjeaBackendIsCopyingCachedOutputsSpec.scala
Horneth: Hmm because I probably changed to `_` in the first version and then git reverted it to what it was :) Will change it

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -380,13 +364,7 @@ class EngineJobExecutionActor(replyTo: ActorRef,
...
 
   private def disableCallCaching(reason: Option[Throwable] = None) = {
-    reason foreach { r =>
-      // Publish hash failure to metadata
-      import cromwell.services.metadata.MetadataService._
-      val failureAsEvents = throwableToMetadataEvents(metadataKeyForCall(jobDescriptorKey, CallMetadataKeys.CallCachingKeys.HashFailuresKey), r)
-      serviceRegistryActor ! PutMetadataAction(failureAsEvents)
-      log.error(r, "{}: Hash error, disabling call caching for this job.", jobTag)
-    }
+    reason foreach { r => log.error(r, "{}: Hash error, disabling call caching for this job.", jobTag) }
```
danbills: toL could be `{log.error(_, "...`

danbills:
👍

PR #2379, opened 2017-06-26
C(WL) gull

mcovarr:
A proposed config scheme to support multiple language frontends to Cromwell.  Basically a simplified version of the backend scheme, this is just a strawman for discussion.

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -209,6 +209,37 @@ engine {
   }
 }
 
+frontend {
+  default = "WDL"
+  providers {
+    WDL {
+      supported-versions = [
+        {
+          workflow-type = [ "WDL" ],
+          workflow-type-version = [ ] # actually require this to be blank as assent to Cromwell's disregard for WDL versions?
+        }
+      ]
+      parser-factory="cromwell.language.parser.wdl.WdlParserFactory"
+      config {
+
+      }
+    }
+
+    CWL {
+      supported-versions = [
+        {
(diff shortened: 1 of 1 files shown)


core/src/main/resources/reference.conf
```diff
@@ -209,6 +209,37 @@ engine {
...
 }
 
+frontend {
+  default = "WDL"
+  providers {
+    WDL {
+      supported-versions = [
+        {
+          workflow-type = [ "WDL" ],
+          workflow-type-version = [ ] # actually require this to be blank as assent to Cromwell's disregard for WDL versions?
```
geoffjentry: We should just assign some arbitrary version to pre-version WDL, e.g. `draft-2`, which is technically what we say it is. After that if we make clear that draft versions are moving targets I think it'll be fine.

core/src/main/resources/reference.conf
```diff
@@ -209,6 +209,37 @@ engine {
   }
 }
 
+frontend {
+  default = "WDL"
+  providers {
+    WDL {
+      supported-versions = [
+        {
+          workflow-type = [ "WDL" ],
```
geoffjentry: I don't understand why this is an array

geoffjentry:
I thought @cjllanwarne brought up a good point at standup about whether or not we wanted this to be pluggable in the first place. I didn't actually think about that when I looked at this. I suspect that usage of actual plugability would be pretty low.

mcovarr:
I'm going to table this for now as there's more interesting stuff to be working on.

PR #2380, opened 2017-06-26
Spray -> Akka Http closes #1243

### LINKED ISSUE #1243: spray -> akka http
Akka HTTP is effectively the next version of Spray. Spray is for all intents and purposes a dead technology. Considering that's been the case for over a year now, we should update at some point just because eventually we'll run into a problem that won't be resolved by developers. Also, Akka HTTP is now hitting a point in its development cycle where not only is it deemed "Better" :tm: but they've switched to ease of use and performance and it's eclipsing Spray in those categories now as well.

I've labeled this as developers choice as it certainly doesn't _need_ to happen any time soon, although if a someone wearing a PO hat thought it was otherwise important it could certainly be pushed in elsewhere (since it likely _would_ include performance enhancements and such)

geoffjentry:
- Despite what it might seem, I tried to leave as much as possible the way I found it. I've got a big list of followup issues to file - both pre-existing conditions (provided we get this in before the ACA gets passed) and some things which are now hokier than I'd like due to the changes but would cause some serious spidering going on
- I did move some stuff around, e.g. PartialWorkflowSources, so many things look by diff to be changed a lot more than they really were
- The work for this has been spread across the better part of a year, often with long pauses in between. I did my best to spot inconsistencies and such but I'm sure they're still out there


geoffjentry:
Also, regardless of where things stand, this will *not* be part of the upcoming release. I want to give downstream users a chance to more thoroughly vet this in case there are subtle changes not picked up by our testing.

NOTICE
```diff
@@ -1,4 +0,0 @@
-cromwell.webservice/PerRequest.scala (https://github.com/NET-A-PORTER/spray-actor-per-request)
```
geoffjentry: no longer using perRequest, thus this is no longer needed

engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
```diff
@@ -0,0 +1,142 @@
...
+import cats.data.Validated.{Invalid, Valid}
+import cats.syntax.validated._
+import cats.syntax.cartesian._
+import lenthall.validation.ErrorOr.ErrorOr
+import cromwell.core._
+import org.slf4j.LoggerFactory
+import spray.json.{JsObject, JsValue}
+
+import scala.util.Try
+
```
geoffjentry: this is minimally changed from the original

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreEngineActor.scala
```diff
@@ -146,8 +146,9 @@ object WorkflowStoreEngineActor {
   sealed trait WorkflowStoreEngineActorResponse
   case object NoNewWorkflowsToStart extends WorkflowStoreEngineActorResponse
   final case class NewWorkflowsToStart(workflows: NonEmptyList[WorkflowToStart]) extends WorkflowStoreEngineActorResponse
-  final case class WorkflowAborted(workflowId: WorkflowId) extends WorkflowStoreEngineActorResponse
-  final case class WorkflowAbortFailed(workflowId: WorkflowId, reason: Throwable) extends WorkflowStoreEngineActorResponse
+  sealed abstract class WorkflowStoreEngineAbortResponse extends WorkflowStoreEngineActorResponse
```
geoffjentry: I've preferred abstract classes to traits where possible for a long time now (mostly for reasons cited by [this SO answer](https://stackoverflow.com/a/35251513)) although it's never bubbled up to be a thing that I badger others about because at the end of the day it's a fairly minor point IMO.

Also it's trivial to change to a `Trait` if it ever becomes necessary.

engine/src/main/scala/cromwell/webservice/AkkaHttpService.scala
```diff
@@ -0,0 +1,246 @@
...
+      get { metadataBuilderRequest(possibleWorkflowId, (w: WorkflowId) => GetStatus(w)) }
+    } ~
+    path("workflows" / Segment / Segment / "outputs") { (version, possibleWorkflowId) =>
+      get { metadataBuilderRequest(possibleWorkflowId, (w: WorkflowId) => WorkflowOutputs(w)) }
+    } ~
+    path("workflows" / Segment / Segment / "logs") { (version, possibleWorkflowId) =>
+      get { metadataBuilderRequest(possibleWorkflowId, (w: WorkflowId) => GetLogs(w)) }
+    } ~
+    path("workflows" / Segment / "query") { version =>
+      (post | get) {
```
geoffjentry: post and get, sittin' in a (route) tree ....

engine/src/test/scala/cromwell/webservice/MetadataBuilderActorSpec.scala
```diff
@@ -292,15 +291,16 @@ class MetadataBuilderActorSpec extends TestKitSuite("Metadata") with FlatSpecLik
...
-      ("list", "res"),
+    val t = List(
       (List(kv),  """"key": "value""""),
       (List(kv, ksv2),  """"key": { "subkey": "value2" }"""),
       (List(kv, ksv2, kisv3),  """"key": [ { "subkey": "value3" } ]"""),
       (List(kv, ksv2, kisv3, kiv4),  """"key": [ "value4" ]""")
     )
 
-    forAll(t) { (l, r) => assertMetadataKeyStructure(l, r) }
+    Future.sequence(t map { case (l, r) => assertMetadataKeyStructure(l, r) }) map { assertions =>
```
geoffjentry: I actually wondered about that when I put it in and then promptly forgot. Thanks.

project/Dependencies.scala
```diff
@@ -154,27 +144,24 @@ object Dependencies {
...
     "io.swagger" % "swagger-parser" % "1.0.22" % Test,
     "org.yaml" % "snakeyaml" % "1.17" % Test
-  ) ++ sprayServerDependencies
+  ) ++ akkaHttpServerDependencies
 
   val rootDependencies = slf4jBindingDependencies
+
+  val jesBackendDependencies = refinedTypeDependenciesList
+
+  val sprayClientDependency = List("io.spray" %% "spray-client" % "1.3.3")
```
geoffjentry: Yeah, in fact this reminds me I wanted to twiddle this file a bit anyways.

engine/src/test/scala/cromwell/webservice/MetadataBuilderActorSpec.scala
geoffjentry: Hmm, not seeing how it'd help. One thing to keep in mind is that w/ the `AsyncFlatSpec` the assertions aren't `Unit`, I want to return a `Future[Assertion]` from this test block. 

So what this is doing is effectively the same as:

    val z: List[Future[Assertion]] = t map { case (l, r) => assertMetadataKeyStructure(l, r) }
    val x: Future[List[Assertion]] = Future.sequence(z)
    val a: Future[Assertion] = x should contain only Succeeded
    a

Not sure if that changes your opinion? I've always been crappy w/ `traverse` though, so perhaps not.

engine/src/main/scala/cromwell/webservice/AkkaHttpService.scala
```diff
@@ -0,0 +1,246 @@
...
+        parameterSeq { parameters =>
+          extractUri { uri =>
+            metadataQueryRequest(parameters, uri)
+          }
+        }
+      }
+    } ~
+    encodeResponseWith(Gzip, Deflate, NoCoding) {
```
geoffjentry: This (mostly) preserves the behavior in that it will default to gzipping the response even if that's not specified by the client, but allows the client to override this by specifying a different behavior.

Horneth: Should we maybe take this opportunity to change that behavior ?

geoffjentry: I actually had at first - both out of desire (I agree with the belief that unecessarily forcing a content type down someone's throat isn't cool) but also necessity (I couldn't figure out how to replicate the  old behavior for a while). 

This was a relatively late change - I forget what exactly led me to think this but at some point I came to the conclusion that the change would lead to more trouble than it was worth, in the form of people who would like for it to be gzipped and not getting it that way.

I'm more than happy to change it back if people think otherwise, the only "downside" (again if one thinks this way) is that a user would need to explicitly request it to be gzipped instead of by default.

engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
```diff
@@ -0,0 +1,142 @@
...
+         |The '$out' parameter name has been deprecated in favor of '$in'.
+         |Support for '$out' will be removed from future versions of Cromwell.
+         |Please switch to using '$in' in future submissions.
+         """.stripMargin
+    log.warn(warning)
+  }
+
+  def mergeMaps(allInputs: Seq[Option[String]]): JsObject = {
+    val convertToMap = allInputs.map(x => toMap(x))
+    JsObject(convertToMap reduce (_ ++ _))
```
Horneth: I think we might want `reduceLeft` here, otherwise the order is [nondeterministic](https://www.scala-lang.org/api/current/scala/collection/immutable/List.html#reduce[A1>:A](op:(A1,A1)=>A1):A1)

engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  private def toMap(someInput: Option[String]): Map[String, JsValue] = {
+    import spray.json._
+    someInput match {
+      case Some(inputs: String) => inputs.parseJson match {
+        case JsObject(inputMap) => inputMap
+        case _ =>
+          throw new RuntimeException(s"Submitted inputs couldn't be processed, please check for syntactical errors")
```
Horneth: Are we ok if this actually throws ?

engine/src/test/scala/cromwell/webservice/MetadataBuilderActorSpec.scala
mcovarr: It can be done, not sure it's worth it though:

```
  def assertMetadataKeyStructure(eventList: List[EventBuilder],
                                 expectedJson: String) = {
    assertMetadataKeyStructure(eventList, expectedJson, WorkflowId.randomId(), makeEvent)
  }
.
.
.
    import cats.instances.list._
    import cats.syntax.traverse._

    val assert_ = (assertMetadataKeyStructure _).tupled

    t traverse assert_ map { assertions =>
      assertions should contain only Succeeded
    }
```

mcovarr: A little better:

```
  def assertMetadataKeyStructure(tuple: (List[EventBuilder], String)) = {
    assertMetadataKeyStructure(tuple._1, tuple._2, WorkflowId.randomId(), makeEvent)
  }
.
.
.
    import cats.instances.list._
    import cats.syntax.traverse._

    t traverse assertMetadataKeyStructure map { assertions =>
      assertions should contain only Succeeded
    }
```

geoffjentry: I'm happy to do this if you want me to but that doesn't really seem worth it, particularly for test code. OTOH there's more 😼 so that's something. :) I'll leave it up to you.

mcovarr: meh

engine/src/test/scala/cromwell/webservice/AkkaHttpServiceSpec.scala
```diff
@@ -0,0 +1,497 @@
...
+                      |{
+                      |  "defaultRuntimeOptions": {
+                      |  "cpu":1
+                      |  }
+                      |}
+                      |""".stripMargin
+
+    val workflowSource = Multipart.FormData.BodyPart("workflowSource", HttpEntity(MediaTypes.`application/json`, HelloWorld.workflowSource()))
+    val workflowInputs = Multipart.FormData.BodyPart("workflowOptions", HttpEntity(MediaTypes.`application/json`, options))
+    val formData = Multipart.FormData(workflowSource, workflowInputs).toEntity()
```
mcovarr: indent's all funny here

mcovarr:
:+1:

engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
cjllanwarne: IIRC if anything in the Route processing throws, it causes a pretty generic 500 response from spray

engine/src/main/scala/cromwell/server/CromwellServer.scala
```diff
@@ -1,75 +1,64 @@
...
-        TODO:
-        If/when CromwellServer behaves like a better async citizen, we may be less paranoid about our async log messages
-        not appearing due to the actor system shutdown. For now, synchronously print to the stderr so that the user has
-        some idea of why the server failed to start up.
-         */
-        Console.err.println(s"Binding failed interface $interface port $port")
-        throwable.printStackTrace(Console.err)
-        cromwellSystem.shutdownActorSystem()
+    implicit val actorSystem = cromwellSystem.actorSystem
+    implicit val materializer = ActorMaterializer()
```
geoffjentry: I would love to claim this was a stroke of genius but instead the answer appears to involve a shruggie. Changing it.

engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
geoffjentry: so both this and the mergeMaps comment, this is pre-existing. My goal was to try not to intentionally change behavior (I'm already afraid about the unintended changes) and to flag things that don't look right for further investigation. 

Now that I think about it, the gzip issue would fall under this category as well.

Are you ok w/ that?

Horneth: Mokay 👍

geoffjentry:
Ok, now a legit PR.

@mcovarr does your thumb still hold? Changes were fairly minimal since you looked but feel free to comment

@Horneth you already did some digging and some of the followup changes involved pulling in your call cache 404 stuff, can you take a look?

@ruchim can you take a look at your label stuff?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheDiffActor.scala
```diff
@@ -10,22 +10,23 @@ import cromwell.engine.workflow.lifecycle.execution.callcaching.CallCacheDiffQue
...
 import scala.language.postfixOps
 import scala.util.{Failure, Success, Try}
 
 object CallCacheDiffActor {
   private val PlaceholderMissingHashValue = MetadataPrimitive(MetadataValue("Error: there is a hash entry for this key but the value is null !"))
-  private val CallAAndBNotFoundException = new Exception("callA and callB were run on a previous version of Cromwell on which this endpoint was not supported.")
-  private val CallANotFoundException = new Exception("callA was run on a previous version of Cromwell on which this endpoint was not supported.")
-  private val CallBNotFoundException = new Exception("callB was run on a previous version of Cromwell on which this endpoint was not supported.")
+
+  final case class CachedCallNotFoundException(message: String) extends Exception {
```
Horneth: Does this appear as a 404 in the response ? I don't see how it know that it should be that.

Horneth:
Just a quick question otherwise 👍

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheDiffActor.scala
geoffjentry: It's handled in the route itself: https://github.com/broadinstitute/cromwell/pull/2380/files#diff-77909d86deb4ace5512774b634017213R115

The pattern now is that we're no longer generating the full response anywhere but the route code unlike before where responses were generated by the various actors.

geoffjentry: If that link doesn't work awesome, it's line 115 in `CromwellApiService`

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Cromwell Change Log
 
+## 29
+
+### Breaking Changes
+
+* Request timeouts for HTTP requests on the REST API now return a 503 status code instead of 500. It also no longer contains a JSON response.
```
mcovarr: Maybe "The response for a request timeout is no longer in JSON format"?  The antecedent of the "It" seems a little ambiguous.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Cromwell Change Log
 
+## 29
+
+### Breaking Changes
+
+* Request timeouts for HTTP requests on the REST API now return a 503 status code instead of 500. It also no represents the error message as JSON.
```
mcovarr: Maybe "The response for a request timeout is no longer in JSON format"? The antecedent of the "It" seems a little ambiguous.

PR #2391, opened 2017-06-28
Metadata write batch size sanity

mcovarr:


core/src/main/resources/reference.conf
```diff
@@ -558,10 +558,8 @@ services {
       # performance but will both lead to a higher memory usage as well as increase the risk that metadata events
       # might not have been persisted in the event of a Cromwell crash.
       #
-      # For normal usage the default value of 1 (effectively no batching) should be fine but for larger/production
-      # environments we recommend a value of at least 500. There'll be no one size fits all number here so we recommend
-      # benchmarking performance and tuning the value to match your environment
-      # db-batch-size = 1
+      # This value may require benchmarking performance and tuning the value to match your environment.
```
geoffjentry: I don't recall where the "at least 500" came from for "larger/production environments" but I probably didn't pull it out of my butt. Think it's worth preserving that as well.

I was doing a lot of experimentation and benchmarking at the time, it likely was a level that seemed to manage reasonable "heavy load" without making the gap too high.

geoffjentry:
:

core/src/main/resources/reference.conf
mcovarr: As long as you're sure that's not where it came from I'll put it back.

Horneth:
👍

core/src/main/resources/reference.conf
geoffjentry: Just wear gloves to be on the safe side.

Yeah, I have no idea what significance that number held but I ran too many tests for that to have been completely made up.

PR #2395, opened 2017-06-29
Fix sbt test

jsotobroad:
Wrap test in eventually to make sure metadata of workflow is flushed before we ask for it.  This changed a bit when we set the default batch size from 1 -> 200


ruchim:
👍

Horneth:
👍

PR #2399, opened 2017-06-30
Update release WDL

Horneth:


geoffjentry:
👍

release/release_workflow.wdl
```diff
@@ -74,15 +70,11 @@ task do_release {
      # Merge develop into master
      git checkout master
      git pull --rebase
-     git merge develop
+     git merge develop --no-edit
```
mcovarr: is the `--no-edit` still required?

mcovarr:
:+1:

release/release_workflow.wdl
Horneth: well I'm not sure, last time I released I didn't seem to need it but when I tried manually this time it prompted me with the commit message edit thing. So it was more preemptively that I added it to avoid having it fail again..

PR #2406, opened 2017-06-30
[Develop] Return better 404 message when calls can't be found

Horneth:
Adjust the error message for the Call Caching diff endpoint

### DIFF
--- a/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheDiffActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheDiffActor.scala
@@ -24,9 +24,10 @@ object CallCacheDiffActor {
     override def getMessage = message
   }
 
-  private val CallAAndBNotFoundException = CachedCallNotFoundException("callA and callB were run on a previous version of Cromwell on which this endpoint was not supported.")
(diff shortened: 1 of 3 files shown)


mcovarr:
:+1:

geoffjentry:
I'll own resolving the conflict, since i'm the one who caused it

👍

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheDiffActor.scala
```diff
@@ -122,6 +131,30 @@ class CallCacheDiffActor(serviceRegistryActor: ActorRef) extends LoggingFSM[Call
...
 
   /**
+    * Returns a Failure if one or both of the calls are not found.
+    */
+  private def checkCallsExistence(queryA: MetadataQuery,
+                                  queryB: MetadataQuery,
+                                  responseA: MetadataLookupResponse,
+                                  responseB: MetadataLookupResponse): Try[Unit] = {
+    def makeTag(query: MetadataQuery) = {
+      s"${query.workflowId}:${query.jobKey.get.callFqn}:${query.jobKey.get.callFqn}"
```
Horneth: The tag's got callFqn twice, the second one should be the index

PR #2412, opened 2017-07-04
Scala 2.12 support for Cromwell. Closes #1635

### LINKED ISSUE #1635: Upgrade Cromwell to Scala 2.12
res ipsa loquitur

geoffjentry:


services/src/main/scala/cromwell/services/metadata/impl/MetadataDatabaseAccess.scala
```diff
@@ -125,6 +125,13 @@ trait MetadataDatabaseAccess {
       case MetadataQuery(_, None, None, Some(includeKeys), Some(excludeKeys), _) => Future.failed(
         new IllegalArgumentException(
           s"Include/Exclude keys may not be mixed: include = $includeKeys, exclude = $excludeKeys"))
+      case MetadataQuery(_, None, Some(_), None, Some(_), _)    |
+           MetadataQuery(_, None, Some(_), Some(_), None, _)    |
+           MetadataQuery(_, None, Some(_), Some(_), Some(_), _) |
+           MetadataQuery(_, Some(_), None, Some(_), Some(_), _) |
+           MetadataQuery(_, Some(_), Some(_), None, Some(_), _) |
+           MetadataQuery(_, Some(_), Some(_), Some(_), None, _) |
+           MetadataQuery(_, Some(_), Some(_), Some(_), Some(_), _) => Future.failed(new IllegalStateException("Invalid MetadataQuery"))
```
kshakir: 1. How about `new IllegalStateException(s"Invalid MetadataQuery $query")`
2. Is a singular `case _ =>` too loose for your liking?

kshakir:
👍  with minor questions

services/src/main/scala/cromwell/services/metadata/impl/MetadataDatabaseAccess.scala
geoffjentry: 1) sure
2) Presumably it'd satisfy the compiler, so sure. I just copy/pasted that from the error message w/o thinking about it too much (other than 'yuck') :)

engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
```diff
@@ -336,7 +336,7 @@ class CromwellApiServiceSpec extends AsyncFlatSpec with ScalatestRouteTest with
         check {
           status should be(StatusCodes.OK)
           val decoder: Decoder = Gzip
-          val result = Await.result(Unmarshal(decoder.decode(response)).to[JsObject], 1.second)
+          val result = Await.result(Unmarshal(decoder.decodeMessage(response)).to[JsObject], 1.second)
```
cjllanwarne: I *think* this fits the pattern that lets you just say `val result = responseAs[JsObject]`, avoiding the whole convoluted `Await.result` and decoding boilerplate :crossed_fingers: 

(and the examples below too, if it works here)

engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
```diff
@@ -524,10 +525,10 @@ object CromwellApiServiceSpec {
         sender ! response
       case AbortWorkflow(id, manager) =>
         val message = id match {
-          case ExistingWorkflowId =>
-            WorkflowStoreEngineActor.WorkflowAborted(id)
+          case ExistingWorkflowId => WorkflowStoreEngineActor.WorkflowAborted(id)
           case AbortedWorkflowId =>
             WorkflowAbortFailed(id, new IllegalStateException(s"Workflow ID '$id' is in terminal state 'Aborted' and cannot be aborted."))
+          case WorkflowId(_) => throw new Exception("Something untoward happened")
```
cjllanwarne: **ToL**: I like test exceptions to be obviously distinguishable between "don't worry, I wanted this to happen for the test" vs "your test has broken". It's not obvious to me which this falls into.

filesystems/gcs/src/main/scala/cromwell/filesystems/gcs/GoogleConfiguration.scala
```diff
@@ -57,7 +57,7 @@ object GoogleConfiguration {
 
   private val log = LoggerFactory.getLogger("GoogleConfiguration")
 
-  case class GoogleConfigurationException(errorMessages: List[String]) extends MessageAggregation {
+  final case class GoogleConfigurationException(errorMessages: List[String]) extends MessageAggregation {
```
cjllanwarne: 🎉

project/Dependencies.scala
```diff
@@ -4,26 +4,27 @@ object Dependencies {
...
-  lazy val akkaV = "2.4.16"
-  lazy val akkaHttpV = "10.0.5"
+  lazy val akkaV = "2.4.17"
+  lazy val akkaHttpV = "10.0.9"
 
   lazy val slickV = "3.2.0"
-  // TODO: Re-combine these when cromwell is 2.12:
-  lazy val cromwellApiClientAkkaV = "2.4.17"
-  lazy val cromwellApiClientAkkaHttpV = "10.0.6"
+
```
cjllanwarne: 🎉

services/src/main/scala/cromwell/services/metadata/impl/MetadataDatabaseAccess.scala
```diff
@@ -125,6 +125,7 @@ trait MetadataDatabaseAccess {
       case MetadataQuery(_, None, None, Some(includeKeys), Some(excludeKeys), _) => Future.failed(
         new IllegalArgumentException(
           s"Include/Exclude keys may not be mixed: include = $includeKeys, exclude = $excludeKeys"))
+      case _ => Future.failed(new IllegalStateException(s"Invalid MetadataQuery: $query"))
```
cjllanwarne: ToL: Any reason for `InvalidStateException` rather than the `InvalidArgumentException` elsewhere in the file? :man_shrugging:

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAttributes.scala
```diff
@@ -60,7 +60,7 @@ object JesAttributes {
   implicit val urlReader: ValueReader[URL] = StringReader.stringValueReader.map { URI.create(_).toURL }
   
   def apply(googleConfig: GoogleConfiguration, backendConfig: Config): JesAttributes = {
-    val configKeys = backendConfig.entrySet().toSet map { entry: java.util.Map.Entry[String, ConfigValue] => entry.getKey }
+    val configKeys = backendConfig.entrySet().asScala.toSet map { entry: java.util.Map.Entry[String, ConfigValue] => entry.getKey }
```
cjllanwarne: Same comment here about `asScala` removing the Java-y-ness

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesJobPaths.scala
```diff
@@ -11,7 +10,7 @@ object JesJobPaths {
   val GcsExecPathKey = "gcsExec"
 }
 
-case class JesJobPaths(override val workflowPaths: JesWorkflowPaths, jobKey: BackendJobDescriptorKey)(implicit actorSystem: ActorSystem) extends JobPaths {
+case class JesJobPaths(override val workflowPaths: JesWorkflowPaths, jobKey: BackendJobDescriptorKey) extends JobPaths {
```
cjllanwarne: `final`?

engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
cjllanwarne: cf the comment near the end of my lightbend support ticket:
> AKKA HTTP: responding using a Stream in high-level DSL

core/src/main/scala/cromwell/core/ConfigUtil.scala
```diff
@@ -7,15 +7,15 @@ import cats.syntax.validated._
...
+import scala.collection.JavaConverters._
 import scala.reflect.{ClassTag, classTag}
 
 object ConfigUtil {
 
   val validationLogger = LoggerFactory.getLogger("ConfigurationValidation")
 
   implicit class EnhancedConfig(val config: Config) extends AnyVal {
-    def keys = config.entrySet().toSet map { v: java.util.Map.Entry[String, ConfigValue] => v.getKey }
+    def keys = config.entrySet().asScala.toSet map { v: java.util.Map.Entry[String, ConfigValue] => v.getKey }
```
geoffjentry: I believe it just converts to a scala collection.

The key here is that the `JavaConversions` are gone now, so `JavaConverters` is the only thing available w/o doing it by hand.

engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
geoffjentry: This is "clearly the test author didn't think this could possibly happen, so this should probably blow up: :)

services/src/main/scala/cromwell/services/metadata/impl/MetadataDatabaseAccess.scala
geoffjentry: I chose `InvalidStateException` because this is covering a case which the developer(s) seemed to believe wasn't possible (and clearly has never happened before since it hasn't come up).

The 2.12 linter is more robust to detecting non-exhaustive matching

geoffjentry:
@cjllanwarne good catch on the 2.11, found a few others. I think i might have done a find on `-2.11` previously

services/src/main/scala/cromwell/services/metadata/impl/MetadataDatabaseAccess.scala
geoffjentry: although now that i look at it i realize that this is coming effectively from the user

core/src/main/scala/cromwell/core/ConfigUtil.scala
geoffjentry: w/o `asScala`: `util.Set[util.Map.Entry[String, ConfigValue]]`
w/ `asScala`: `mutable.Set[util.Map.Entry[String, ConfigValue]]`

Note that the trailing `toSet` sets it to be immutable

engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
geoffjentry: IIRC there was a reason why I couldn't do that in the first place (I *think* I used `responseAs` elsewhere in the file, I definitely played around with it). And by "couldn't", i probably mean "that was going to be a giant pain in the ass for some reason I didn't want to deal with it at the time". 

I'll take another poke.

core/src/main/scala/cromwell/core/ConfigUtil.scala
cjllanwarne: Oh I think I see... the `config.entrySet` is a (pardon the syntax) `JavaSet[JavaMap.Entry]` and `asScala` makes it a `ScalaSet[JavaMap.Entry]`. LGTM

services/src/main/scala/cromwell/services/metadata/impl/MetadataDatabaseAccess.scala
cjllanwarne: Ok

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAttributes.scala
cjllanwarne: As above. LGTM

cjllanwarne:
👍 after double checking the docs for `2.11`s

engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
geoffjentry: oh duh. it's because i couldn't see a way to use responseAs with gzip

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -289,10 +285,7 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
     * @param handle The handle of the running job.
     * @return The status of the job.
     */
-  def pollStatusAsync(handle: StandardAsyncPendingExecutionHandle)
-                     (implicit ec: ExecutionContext): Future[StandardAsyncRunStatus] = {
-    Future.fromTry(Try(pollStatus(handle)))
-  }
+  def pollStatusAsync(handle: StandardAsyncPendingExecutionHandle): Future[StandardAsyncRunStatus] = Future.fromTry(Try(pollStatus(handle)))
```
mcovarr: the Try => Future thing happens three times just in this file, perhaps some sugar is in order

project/Settings.scala
```diff
@@ -25,30 +25,49 @@ object Settings {
...
-
-    http://blog.threatstack.com/useful-scalac-options-for-better-scala-development-part-1
-
-    and
-
-    https://tpolecat.github.io/2014/04/11/scalac-flags.html
+    Other fancy flags from https://tpolecat.github.io/2017/04/25/scalac-flags.html.
 
+    The following aren't used (yet), and in general are an exercise in pain for 2.12 with Cromwell. They'd
+    certainly be nice to have, but params causes a world of hurt and patvars is just going to be a big time sink.
```
mcovarr: holy moly, it'll be a time sink just understanding all the ones that did get turned on 😄

src/bin/travis/testCentaurTes.sh
```diff
@@ -42,7 +42,7 @@ mysql -u root -e "GRANT ALL PRIVILEGES ON cromwell_test . * TO 'travis'@'localho
 WORKDIR=$(pwd)
 
 sbt assembly
-CROMWELL_JAR=$(find "$(pwd)/target/scala-2.11" -name "cromwell-*.jar")
+CROMWELL_JAR=$(find "$(pwd)/target/scala-2.12" -name "cromwell-*.jar")
```
mcovarr: which reminds me we should check what the release WDL is expecting WRT 2.11 / 2.12

project/Settings.scala
geoffjentry: You don't want to know how much time I spent over the weekend before I discovered I could just turn on everything *but* the stuff I wanted in there.

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
geoffjentry: I don't even understand why it's doing it this way in the first place instead of just a future.

that said, i'll add it to the list of preexisting conditions i claim i'm going to fix from my akka http list

mcovarr:
:+1: heroic  🎖

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
mcovarr: That is a solid point.

geoffjentry: Of course then I'd need to rollback my removal of the ECs, oh the tangled webs we weave

project/Settings.scala
geoffjentry: BTW if you want to get an idea, look at the branch jg_scala_2_12, and I didn't even manage to fix all the things those two things were complaining about before I gave up

PR #2422, opened 2017-07-05
Made 'size' accept empty file? values

cjllanwarne:


mcovarr:
:+1: aside from the whole "unfilled optionals of types that could be coerced to files have size zero" business 😛

mcovarr:
oh and some tests wouldn't hurt 😛

geoffjentry:
Including centaur :)

project/Dependencies.scala
```diff
@@ -2,7 +2,7 @@ import sbt._
 
 object Dependencies {
   lazy val lenthallV = "0.25"
-  lazy val wdl4sV = "0.13"
+  lazy val wdl4sV = "0.14-c3be413-SNAP"
```
cjllanwarne: hmm, that's automatic when we push to WDL4S develop though, right?

backend/src/main/scala/cromwell/backend/wdl/ReadLikeFunctions.scala
```diff
@@ -122,11 +122,20 @@ trait ReadLikeFunctions extends PathFactory { this: WdlStandardLibraryFunctions
   override def size(params: Seq[Try[WdlValue]]): Try[WdlFloat] = {
     def toUnit(wdlValue: Try[WdlValue]) = wdlValue flatMap { unit => Try(MemoryUnit.fromSuffix(unit.valueString)) }
 
+    def optionalSafeFileSize(value: WdlValue): Try[Double] = value match {
+      case f: WdlFile => Try(buildPath(f.valueString).size.toDouble)
+      case f if WdlFileType.isCoerceableFrom(f.wdlType) => WdlFileType.coerceRawValue(f) map { coerced => buildPath(coerced.asInstanceOf[WdlFile].valueString).size.toDouble }
```
cjllanwarne: Oh I like that!

cjllanwarne:
Added some tests. @geoffjentry I think these unit tests are comprehensive enough without (re-)testing that Cromwell wires variables into engine functions correctly in centaur. Hope you agree?

geoffjentry:
@cjllanwarne My centaur comment was more that ideally we'd be testing every function in centaur. I'm sure this wouldn't be the first one we're slacking on, don't let me hold you up.

cjllanwarne:
@geoffjentry I'll rephrase:
- I agree that the `size` function's interaction with the outside world deserves a centaur test (and I believe it has one).
- Since this PR only adds functionality that's entirely internal to Cromwell, I don't think (even in a perfect world) that we need another `empty_optionals_size` test in centaur to make sure that empty optionals give a size of 0?

geoffjentry:
I'll file that under the "don't let me hold you up" part

geoffjentry:
@cjllanwarne realized you were probably looping back due to lack of 👍

PR #2427, opened 2017-07-06
Only gzip metadata if requested. Closes #2419

### LINKED ISSUE #2419: Should metadata endpoint be forcing gzip on users?
A while back we changed things to return gzipped content in the metadata endpoint by default, despite what the client says it can handle. generally the client should be telling the server what it can expect and the server either handles it appropriately or returns a 415.

I think that we should switch this to encoding on request. The downside is that there might be users who would be much better off getting a gzipped response but not realizing they have to do something extra (depending on their client).

geoffjentry:


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Breaking Changes
 
 * Request timeouts for HTTP requests on the REST API now return a 503 status code instead of 500. The response for a request timeout is no longer in JSON format.
+* The metadata endpoint was defaulting to responding with a gzipped response. This now needs to be explicitly requested with an `Accept-Encoding: gzip` header
```
mcovarr: Maybe "no longer returns gzipped responses by default."

engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
```diff
@@ -321,22 +321,36 @@ class CromwellApiServiceSpec extends AsyncFlatSpec with ScalatestRouteTest with
...
+        check {
+          response.headers.find(x => x.name() == "Content-Encoding").get.value should be("gzip")
+        }
+    }
+
+    it should "not return with gzip encoding when not requested" in {
+      Get(s"/workflows/$version/${CromwellApiServiceSpec.ExistingWorkflowId}/metadata") ~>
+        akkaHttpService.routes ~>
+        check {
+          response.headers.find(x => x.name() == "Content-Encoding") shouldBe None
```
mcovarr: `_.name()` here and above

mcovarr:
:+1:

engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
kshakir: Edit: `_.name`

kshakir:
👍  With Miguel's comments

PR #2428, opened 2017-07-07
[Do not merge] Db restrict metadata access

danbills:
Still need info on how to apply flag to runtime metadata.

Getting this out there to ensure the correct approach.


geoffjentry:
This is *probably* not wired right, although my comment about being not wired right might itself be not right.

This will probably need to go through `JesInitializationActor.beforeAll().fileUpload()`, which is currently specific to the auth stuff, but would need to include this as well.

danbills:
handing off to @geoffjentry

PR #2433, opened 2017-07-10
patvars

mcovarr:
Simple pattern variables compiler flag cleanup.

### DIFF
--- a/project/Settings.scala
+++ b/project/Settings.scala
@@ -27,11 +27,11 @@ object Settings {
 
     Other fancy flags from https://tpolecat.github.io/2017/04/25/scalac-flags.html.
 
-    The following aren't used (yet), and in general are an exercise in pain for 2.12 with Cromwell. They'd
-    certainly be nice to have, but params causes a world of hurt and patvars is just going to be a big time sink.
-    Interested parties are encouraged to take a stab at it.
-      "-Ywarn-unused:params",              // Warn if a value parameter is unused.
-      "-Ywarn-unused:patvars",             // Warn if a variable bound in a pattern is unused.
+    The following isn't used (yet), and in general is an exercise in pain for 2.12 with Cromwell.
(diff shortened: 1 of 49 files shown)


geoffjentry:
🖖 👍 

I thought there were more than these but I guess not. Didn't have the energy to go back and do them again once I abandoned my first branch :)

mcovarr:
Nah, this was pretty straightforward if a bit tedious.  `-Ywarn-unused:params` is a genuine beast, but some of the things that points out definitely look fishy.

geoffjentry:
Oh yeah. IIRC the irritation I had here was that it'd not give me all the errors at once so i'd think i was done only to get hit by yet another block :)

I mentioned the unused vars in slack. There are some *definite* unused ones in there. Most of them were due to our (over?)reliance on overloaded function signatures where we may or may not use all of the args. Some of them were due to our pattern (particularly in the backend) of supplying a default no-op implementation of things. 

There were a handful though that were really bizarre. As in I couldn't figure out why the compiler thought it was unused.

geoffjentry:
BTW you can see a lot of my crimes against humanity here: https://github.com/broadinstitute/cromwell/compare/jg_scala_2_12?expand=1

I had a variety of terrible ways of "using" the unused args. blech.

project/Settings.scala
```diff
@@ -68,7 +68,8 @@ object Settings {
     "-Ywarn-unused:imports",
     "-Ywarn-unused:privates",
     "-Ywarn-unused:locals",
-    "-Xfatal-warnings"
+    "-Xfatal-warnings",
+    "-Ywarn-unused:patvars"
```
kshakir: Move this up so that it's grouped with the other `-Yunsused:…`.

Otherwise 👍

kshakir:
One minor comment. Otherwise 👍

PR #2454, opened 2017-07-17
Removed /api prefix from /engine.

kshakir:


engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
```diff
@@ -31,13 +32,38 @@ class CromwellApiServiceSpec extends AsyncFlatSpec with ScalatestRouteTest with
 
   implicit def default = RouteTestTimeout(5.seconds)
 
+  behavior of "REST ENGINE /stats endpoint"
+  it should "return 200 for stats" in {
```
danbills: toL: Can this be condensed to `"REST ENGINE /stats endpoint" should "return ..."

danbills:
👍

cjllanwarne:
👍

engine/src/main/scala/cromwell/server/CromwellServer.scala
```diff
@@ -43,9 +41,14 @@ class CromwellServerActor(cromwellSystem: CromwellSystem)(override implicit val
   val interface = webserviceConf.getString("interface")
   val port = webserviceConf.getInt("port")
 
-  val routeUnwrapped: Boolean = cromwellSystem.conf.as[Option[Boolean]]("api.routeUnwrapped").getOrElse(false)
-
-  val allRoutes: Route = routes.wrapped("api", routeUnwrapped) ~ swaggerUiResourceRoute
+  /**
+    * /api routes have special meaning to devops' proxy servers. NOTE: the oauth mentioned on the /api endpoints in
```
geoffjentry: Thanks for this comment

engine/src/main/scala/cromwell/server/CromwellServer.scala
```diff
@@ -43,9 +41,14 @@ class CromwellServerActor(cromwellSystem: CromwellSystem)(override implicit val
...
+    * cromwell.yaml is broken unless the swagger index.html is patched. Copy/paste the code from rawls or cromiam if
+    * actual cromwell+swagger+oauth+/api support is needed.
+    */
+  val apiRoutes: Route = pathPrefix("api")(concat(workflowRoutes))
```
geoffjentry: I'm guessing it's not really feasible to do it other than this piecemeal fashion?

One thing I really came to loathe (and thus undid w/ the akka http transition) was having to both define the route and then make sure it gets added to the aggregation. 

This isn't nearly as egregious as presumably things will fall into one of the two categories, but was curious if this was more a matter of necessity or taste.

geoffjentry:
:+1:

engine/src/main/scala/cromwell/server/CromwellServer.scala
kshakir: In this particular case-- taste. This is a collection all of the routes that are prefixed with "/api", in one place. Alternatively each route _could_ have define a full-path, but that wasn't my preference-- and would have required the specs to be updated to _also_ send requests to full-paths.

More than happy if one changes the route definitions and specs in a future PR.

PR #2456, opened 2017-07-17
Changelog message for Crom28 bug

katevoss:


CHANGELOG.md
```diff
@@ -12,6 +12,12 @@
 
 ## 28
 
+## WARNING! We have found a high priority bug on Cromwell 28. 
+
+### Please revert back to Cromwell 27 while we are fixing the issue and preparing a 28-hotfix version.
```
danbills: Just confirming: this is possible?

mcovarr: Not if using MySQL or CloudSQL.

geoffjentry:
I think what what we should do is once the hotfix is in place to replace the jar. Can have a note here that makes a mention of it, but honestly no one is going to come back here to read the changelog after they've already pulled the release, and as @mcovarr points out it's not even possible if they're using a MySQL based database.

geoffjentry:
This does raise an issue of "how do we communicate pressing issues to users outside of our immediate sphere of communication" in terms of the next time something like this happens. I don't think this is the right thing to do for that.

mcovarr:
Sounds like we should close this then?  Plus it somehow broke all the builds, yikes. 🤕

katevoss:
Closing as this is no longer necessary.

PR #2458, opened 2017-07-17
Separate out retryables from true terminals from the WEA perspective.

mcovarr:


geoffjentry:
Since this is the same as the hotfix, I think this is pre-thumbed

mcovarr:
I'd buy that logic assuming the hotfix had been thumbed.  😛

cjllanwarne:
Re-👍

PR #2462, opened 2017-07-17
Resurrected the unused `Preempted` status

cjllanwarne:
Previously, `Preempted` was not being created in `StatusPolling` because it was matching on the wrong fields. Furthermore, the `JesAsyncBackendJobExecutionActor` was not set up to handle a `Preempted` status if it had ever received one!

The practical upshot of all that is that the metadata for preempted jobs was giving the `backendStatus: "Failed"` instead of `backendStatus: "Preempted"`

### DIFF
--- a/supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
+++ b/supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
@@ -8,9 +8,8 @@ import com.google.api.client.googleapis.json.GoogleJsonResponseException
 import com.google.api.services.genomics.model.RunPipelineRequest
 import com.google.cloud.storage.contrib.nio.CloudStorageOptions
 import cromwell.backend._
-import cromwell.backend.impl.jes.RunStatus.UnsuccessfulRunStatus
-import cromwell.backend.async.{AbortedExecutionHandle, ExecutionHandle, FailedNonRetryableExecutionHandle, FailedRetryableExecutionHandle, PendingExecutionHandle}
 import cromwell.backend.impl.jes.RunStatus.TerminalRunStatus
+import cromwell.backend.async.{AbortedExecutionHandle, ExecutionHandle, FailedNonRetryableExecutionHandle, FailedRetryableExecutionHandle, PendingExecutionHandle}
 import cromwell.backend.impl.jes.errors.FailedToDelocalizeFailure
 import cromwell.backend.impl.jes.io._
 import cromwell.backend.impl.jes.statuspolling.{JesRunCreationClient, JesStatusRequestClient}
@@ -360,10 +359,8 @@ class JesAsyncBackendJobExecutionActor(override val standardParams: StandardAsyn
   override def isSuccess(runStatus: RunStatus): Boolean = {
     runStatus match {
       case _: RunStatus.Success => true
-      case _: RunStatus.Failed => false
-      case unknown =>
-        throw new RuntimeException("isSuccess not called with RunStatus.Success or RunStatus.Failed. " +
-          s"Instead got $unknown")
+      case _: RunStatus.UnsuccessfulRunStatus => false
+      case _ => throw new RuntimeException(s"Cromwell programmer blunder: isSuccess was called on an incomplete RunStatus ($runStatus).")
     }
   }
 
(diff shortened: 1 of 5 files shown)


supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
```diff
@@ -325,7 +324,15 @@ class JesAsyncBackendJobExecutionActor(override val standardParams: StandardAsyn
     }
   }
 
-  override def pollStatusAsync(handle: JesPendingExecutionHandle): Future[RunStatus] = super[JesStatusRequestClient].pollStatus(handle.runInfo.get)
+  override def pollStatusAsync(handle: JesPendingExecutionHandle): Future[RunStatus] = {
+    super[JesStatusRequestClient].pollStatus(handle.runInfo.get) map {
+      case RunStatus.Success(eventList, machineType, zone, instanceName) if jobDescriptor.key.attempt == 1 =>
```
geoffjentry: I don't understand what's going on here

cjllanwarne: oh oops, this 100% should not be left in...

that's the stuff to see what happens if every first task gets preempted 😳

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
```diff
@@ -383,29 +380,29 @@ class JesAsyncBackendJobExecutionActor(override val standardParams: StandardAsyn
...
-      Try { errorMessage.substring(0, errorMessage.indexOf(':')).toInt } toOption
-    }
-
-    val failedStatus: UnsuccessfulRunStatus = runStatus match {
-      case failedStatus: RunStatus.Failed => failedStatus
-      case preemptedStatus: RunStatus.Preempted => preemptedStatus
+    runStatus match {
+      case preemptedStatus: RunStatus.Preempted if preemptible => handlePreemption(preemptedStatus, returnCode)
+      case failedStatus: RunStatus.UnsuccessfulRunStatus => handleFailedRunStatus(failedStatus, handle, returnCode)
       case unknown => throw new RuntimeException(s"handleExecutionFailure not called with RunStatus.Failed or RunStatus.Preempted. Instead got $unknown")
```
geoffjentry: So if this gets Prempted when not preemptible this exception gets thrown? It seems like it should be an error-ful situation (I'm assuming that's expected to be unpossible) but the error message doesn't handle that situation.

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/RunStatus.scala
```diff
@@ -35,7 +43,24 @@ object RunStatus {
...
 
+  object UnsuccessfulRunStatus {
+    def apply(errorCode: Int,
+              errorMessage: Option[String],
+              eventList: Seq[ExecutionEvent],
+              machineType: Option[String],
+              zone: Option[String],
+              instanceName: Option[String]): UnsuccessfulRunStatus = {
+      val jesCode: Option[Int] = errorMessage flatMap { em => Try(em.substring(0, em.indexOf(':')).toInt).toOption }
+      if (errorCode == JesAsyncBackendJobExecutionActor.GoogleAbortedRpc && jesCode.contains(JesAsyncBackendJobExecutionActor.JesPreemption)){
```
geoffjentry: `) {`

geoffjentry:
👍

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
cjllanwarne: Nah, the `RunStatus.UnsuccessfulRunStatus` still catches it and it treats it just like any other `Failed`. IIRC, you can sometimes get `Preempted`-like symptoms from a non-preempted VM

geoffjentry: On the last part, I think what it was is that gotc had seen instances where something had been preempted but received the more general error code (`13` instead of `14` I think, or the other way around). We decided to simply treat those as errors as after talking to Google there was no perfect way to know "is this *really* preempted" in a programatic fashion.

cjllanwarne: So in this situation, your metadata says `executionStatus: Failed` and `backendStatus: Preempted` and the task will fail and won't be restarted. I think that's acceptable?

geoffjentry: Unless you've changed somethign fundamental here (and if you think you did, speak up as I don't think it should have), that wouldn't happen. The way we detect preemption is via a particular code from google. The problem that gotc saw was that sometimes a different code would be given but they could see manually that it was a preemption

cjllanwarne: Discussed externally.
Everything is good.

All is well.

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
```diff
@@ -383,29 +380,29 @@ class JesAsyncBackendJobExecutionActor(override val standardParams: StandardAsyn
...
-      case (GoogleNotFoundRpc, Some(JesFailedToDelocalize)) => Future.successful(FailedNonRetryableExecutionHandle(FailedToDelocalizeFailure(prettyPrintedError, jobTag, Option(jobPaths.stderr))))
-      case (GoogleAbortedRpc, Some(JesUnexpectedTermination)) => handleUnexpectedTermination(failedStatus.errorCode, prettyPrintedError, returnCode)
-      case (GoogleAbortedRpc, Some(JesPreemption)) if preemptible => handlePreemption(failedStatus.errorCode, prettyPrintedError, returnCode)
-      case _ => Future.successful(FailedNonRetryableExecutionHandle(StandardException(failedStatus.errorCode, prettyPrintedError, jobTag), returnCode))
+      case (GoogleNotFoundRpc, Some(JesFailedToDelocalize)) => Future.successful(FailedNonRetryableExecutionHandle(FailedToDelocalizeFailure(runStatus.prettyPrintedError, jobTag, Option(jobPaths.stderr))))
+      case (GoogleAbortedRpc, Some(JesUnexpectedTermination)) => handleUnexpectedTermination(runStatus.errorCode, runStatus.prettyPrintedError, returnCode)
+      case _ => Future.successful(FailedNonRetryableExecutionHandle(StandardException(runStatus.errorCode, runStatus.prettyPrintedError, jobTag), returnCode))
     }
```
mcovarr: ToL I prefer these inner functions above the "meat" of the code.  YMMV.

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
```diff
@@ -383,29 +380,29 @@ class JesAsyncBackendJobExecutionActor(override val standardParams: StandardAsyn
...
+      case failedStatus: RunStatus.UnsuccessfulRunStatus => handleFailedRunStatus(failedStatus, handle, returnCode)
       case unknown => throw new RuntimeException(s"handleExecutionFailure not called with RunStatus.Failed or RunStatus.Preempted. Instead got $unknown")
     }
+  }
 
-    val prettyPrintedError: String = failedStatus.errorMessage map { e => s" Message: $e" } getOrElse ""
-    val jesCode: Option[Int] = failedStatus.errorMessage flatMap getJesErrorCode
-
-    (failedStatus.errorCode, jesCode) match {
+  private def handleFailedRunStatus(runStatus: RunStatus.UnsuccessfulRunStatus,
```
mcovarr: ToL Move this inside `handleExecutionFailure` since it's only used there?

mcovarr:
:+1:

PR #2470, opened 2017-07-20
Assign a dispatcher to services and allow them to be customized

geoffjentry:


mcovarr:
:+1:

Horneth:
👍

cjllanwarne:
Can-of-wormsy ToL: if `reference.conf` doubles as our config documentation, could we include this there? Otherwise, could we write it down *somewhere*?

I like not cluttering the `reference.conf` but I also don't want to have to rummage through some (I've-already-forgotten-which) class file to find how to change this.

geoffjentry:
@cjllanwarne Totally agree. One of the things which became clear to me toot sweet was that was that we have no docs around the service registry in general, e.g. how to make a new service. All that sorta stuff is on my radar atm.

PR #2471, opened 2017-07-20
Made call cache hash endpoint more JSONic

cjllanwarne:
Previously:
```
"hashDifferential": [
  { "output expression:String hi”: 
    {
      "callA": "935C6E7EB2068B83C40B788575747EFB”, 
      "callB": “0183144CF6617D5341681C6B2F756046"
    }
  },
  …
]
```

This PR :
```
"hashDifferential": [
        "hashKey": "output expression:String hi",
        "callA": "935C6E7EB2068B83C40B788575747EFB",
        "callB": "0183144CF6617D5341681C6B2F756046"
    },
    ...
]
```

### DIFF
--- a/backend/src/test/scala/cromwell/backend/wdl/ReadLikeFunctionsSpec.scala
+++ b/backend/src/test/scala/cromwell/backend/wdl/ReadLikeFunctionsSpec.scala
@@ -68,7 +68,7 @@ class ReadLikeFunctionsSpec extends FlatSpec with Matchers {
     val oops = readLike.size(Seq(Success(WdlInteger(7))))
     oops match {
       case Success(x) => fail(s"Expected a string to not have a file length but instead got $x")
-      case Failure(e) => e.getMessage should be("The 'size' method expects a File argument but instead got Int.")
+      case Failure(e) => e.getMessage should be("The 'size' method expects a 'File' or 'File?' argument but instead got Int.")
     }
   }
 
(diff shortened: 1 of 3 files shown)


ansingh7115:
After clarification from Chris that this does not affect any current call caching stuff that firecloud uses, but is a minor formatting change to a "why didn't my call caching work???" endpoint that firecloud possibly may use in the future,  👍

helgridly:
I haven't used this endpoint and I'm not entirely sure what the use case is. However, this new version seems less usable to me. I can no longer look up the differential by a key I'm interested in - I have to iterate over all elements looking for the one I want. This seems worse -- why use a map?

Also I don't really know why it's in an array. Why isn't it just:

```
"hashDifferential": {
    "output expression:String hi”: 
    {
      "callA": "935C6E7EB2068B83C40B788575747EFB”, 
      "callB": “0183144CF6617D5341681C6B2F756046"
    },
    "output thing:blah blah"":
    { ... }
}
```

cjllanwarne:
@helgridly 
- Conceptually, this is a list of differences, not a map from known keys to values of interest. Rather, the hashKey is just another piece of information you might be interested in alongside the two calls' hash values.
- I'm keeping the schema well-defined. IMO this will make parsing easier and more precise.

helgridly:
ok. it was my $.02 without really knowing anything about any of this. feel free to ignore me, not that you needed permission

mcovarr:
:+1: beauteous

danbills:
👍

PR #2474, opened 2017-07-21
Additional clienty bits for Cromwell

cjllanwarne:
Adds endpoint support for
- logs
- outputs
- call cache diffs

### DIFF
--- a/cromwellApiClient/src/main/scala/cromwell/api/CromwellClient.scala
+++ b/cromwellApiClient/src/main/scala/cromwell/api/CromwellClient.scala
@@ -29,12 +29,21 @@ class CromwellClient(val cromwellUrl: URL, val apiVersion: String)(implicit acto
   def abortEndpoint(workflowId: WorkflowId): String = workflowSpecificEndpoint(workflowId, "abort")
   def statusEndpoint(workflowId: WorkflowId): String = workflowSpecificEndpoint(workflowId, "status")
   def metadataEndpoint(workflowId: WorkflowId): String = workflowSpecificEndpoint(workflowId, "metadata")
+  def outputsEndpoint(workflowId: WorkflowId): String = workflowSpecificEndpoint(workflowId, "outputs")
+  def logsEndpoint(workflowId: WorkflowId): String = workflowSpecificEndpoint(workflowId, "logs")
+  def diffEndpoint(workflowA: WorkflowId, callA: String, indexA: ShardIndex, workflowB: WorkflowId, callB: String, indexB: ShardIndex): String = {
+    def shardParam(aOrB: String, s: ShardIndex) = s.index.map(i => s"&index$aOrB=$i.toString").getOrElse("")
+    s"$submitEndpoint/callcaching/diff?workflowA=$workflowA&callA=$callA&workflowB=$workflowB&callB=$callB${shardParam("A", indexA)}${shardParam("B", indexB)}"
+  }
   lazy val backendsEndpoint = s"$submitEndpoint/backends"
   lazy val versionEndpoint = s"$engineEndpoint/version"
 
   import model.CromwellStatusJsonSupport._
+  import model.WorkflowOutputsJsonSupport._
+  import model.WorkflowLogsJsonSupport._
   import model.CromwellBackendsJsonSupport._
   import model.CromwellVersionJsonSupport._
+  import model.CallCacheDiffJsonSupport._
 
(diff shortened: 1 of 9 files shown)


geoffjentry:
File this under TOL but I'll channel my inner @kcibul and point out that at some point we should look and see if swagger actually *can* generate a java client for us. making a wrapper to de-suck the java would be a lot less work than building this out to completeness.

cromwellApiClient/src/main/scala/cromwell/api/model/WorkflowId.scala
```diff
@@ -2,14 +2,16 @@ package cromwell.api.model
...
+
 // ********* !!!!!!!!!! ********
 //
 // WARNING! This is the Cromwell API version of WorkflowId. If you aren't changing the API client, you probably
 // want cromwell.core.WorkflowId instead!
 //
 // ********* !!!!!!!!!! ********
 
-case class WorkflowId(id: UUID) {
+case class WorkflowId(id: UUID) extends AnyVal {
```
geoffjentry: finalize while you're here

geoffjentry:
👍

cromwellApiClient/src/main/scala/cromwell/api/model/WorkflowId.scala
mcovarr: very cool 😎

cromwellApiClient/src/main/scala/cromwell/api/CromwellClient.scala
```diff
@@ -29,12 +29,21 @@ class CromwellClient(val cromwellUrl: URL, val apiVersion: String)(implicit acto
   def abortEndpoint(workflowId: WorkflowId): String = workflowSpecificEndpoint(workflowId, "abort")
   def statusEndpoint(workflowId: WorkflowId): String = workflowSpecificEndpoint(workflowId, "status")
   def metadataEndpoint(workflowId: WorkflowId): String = workflowSpecificEndpoint(workflowId, "metadata")
+  def outputsEndpoint(workflowId: WorkflowId): String = workflowSpecificEndpoint(workflowId, "outputs")
+  def logsEndpoint(workflowId: WorkflowId): String = workflowSpecificEndpoint(workflowId, "logs")
+  def diffEndpoint(workflowA: WorkflowId, callA: String, indexA: ShardIndex, workflowB: WorkflowId, callB: String, indexB: ShardIndex): String = {
+    def shardParam(aOrB: String, s: ShardIndex) = s.index.map(i => s"&index$aOrB=$i.toString").getOrElse("")
```
mcovarr: I see the way this is being called it's "Little Bobby Tables"-proof, and I'm pretty sure nothing malicious could happen from passing junk to this "safe" endpoint anyway, but building URLs through interpolation of an unrestricted `String` still makes me somewhat uneasy.  Could this use a proper type (i.e. case class) or some fancy "refined" magic instead?

mcovarr:
:+1:

cromwellApiClient/src/main/scala/cromwell/api/model/ShardIndex.scala
```diff
@@ -0,0 +1,18 @@
...
+case class ShardIndex(index: Option[Int]) extends AnyVal {
+  override def toString: String = index.getOrElse(-1).toString
+}
+
+object ShardIndexFormatter extends DefaultJsonProtocol {
+  implicit object ShardIndexJsonFormat extends RootJsonFormat[ShardIndex] {
+    def write(si: ShardIndex) = JsNumber(si.index.getOrElse(-1))
+    def read(value: JsValue) = value match {
+      case JsNumber(i) if i.equals(0) => ShardIndex(None)
+      case JsNumber(i) if i.isValidInt => ShardIndex(Option(i.intValue()))
```
mcovarr: negative ints would not be cool

cromwellApiClient/src/main/scala/cromwell/api/CromwellClient.scala
cjllanwarne: Security wise, I think it's the Cromwell service's job to make sure "Bobby Tables" URLs don't work (otherwise anyone could craft them, regardless of this library)

So, usability wise, do you think it'd be better to:
- Tell people that their request will be rejected upfront, rather than letting Cromwell decide?
- Let Cromwell decide, and let people get their errors from Cromwell rather than the client?

PR #2482, opened 2017-07-24
updates for subprojected wdl4s

mcovarr:
Disclaimer: I didn't actually try to run this, but I did try to fix the things that were obviously affected by the recent subprojecting of wdl4s.

### DIFF
--- a/release/release_workflow.wdl
+++ b/release/release_workflow.wdl
@@ -11,9 +11,6 @@ task do_release {
    # Commands that will update previously released/published dependencies in this repo
    Array[String] dependencyCommands = []
    
-   # When true, nothing will be pushed to github, allows for some level of testing
-   Boolean dryRun = false
-   
    # Can be swapped out to try this on a fork
    String organization
     
@@ -28,8 +25,7 @@ task do_release {
      git pull --rebase
 
      # Expect the version number on develop to be the version TO BE RELEASED
-     echo "Releasing ${organization}/${repo} ${releaseV}${true=" - This a dry run, push commands won't be executed" false = "" dryRun}"
-     
+
      echo "Updating dependencies"
      ${sep='\n' dependencyCommands}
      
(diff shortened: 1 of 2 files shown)


release/release_workflow.wdl
```diff
@@ -44,7 +44,10 @@ task do_release {
        # Generate new scaladoc
        sbt 'set scalacOptions in (Compile, doc) := List("-skip-packages", "better")' doc
        git checkout gh-pages
-       mv target/scala-2.12/api ${releaseV}
+       for subproj in cwl wdl wom; do
+         mkdir -p ${releaseV}/$subproj
+         mv $subproj/target/scala-2.12/api ${releaseV}/$subproj
+       done
```
cjllanwarne: So only the scaladocs need this treatment?

release/release_workflow.wdl
```diff
@@ -259,7 +262,7 @@ workflow release_cromwell {
   call versionPrep as wdl4sPrep { input: 
     organization = organization,
     repo = "wdl4s",
-    file = "build.sbt",
+    file = "project/Version.scala",
```
cjllanwarne: Is this not a functional change

cjllanwarne:
ToL: I'm nervous to thumbs up without any way to verify this. On the other hand when build things have gone wrong before it's not been cataclysmic so...

👍 for now... and let the proof of the pudding be in the eating!

release/release_workflow.wdl
mcovarr: Not sure what you're asking.  This is where the release WDL goes to look for the version number, and when wdl4s was subprojected this is where the version number is now stored.

release/release_workflow.wdl
mcovarr: AFAICT.  The actual dependency management is handled by sbt.

mcovarr:
Yeah I agree it's not great that I didn't / wasn't able to test this, but I will say it's almost certainly more correct than what was there before given the wdl4s subprojecting.  So it's got that going for it.

Horneth:
👍

Horneth:
👍

PR #2484, opened 2017-07-25
Initialize db before serving over HTTP

kshakir:


engine/src/main/scala/cromwell/server/CromwellSystem.scala
```diff
@@ -5,11 +5,21 @@ import akka.http.scaladsl.Http
...
 import scala.concurrent.Future
 
 trait CromwellSystem {
+  /*
+  Initialize the services-store (aka a collection of DAOs), if they haven't been already.
+  This is the closest we have to:
+  - https://github.com/broadinstitute/firecloud-orchestration/blob/3c9482b/DEVNOTES.md
+  - https://github.com/broadinstitute/clio/blob/7253ec0/clio-server/src/main/scala/org/broadinstitute/clio/server/service/ServerService.scala#L58-L66
+  - https://en.wikipedia.org/wiki/Data_access_object
+   */
```
mcovarr: Maybe a comment as to the problem that is avoided by this being here, lest someone be tempted to refactor it to a spot where it no longer has the desired effect at the desired point in the initialization?

mcovarr:
:+1:

engine/src/main/scala/cromwell/server/CromwellSystem.scala
kshakir: Added above

Horneth:
👍

PR #2485, opened 2017-07-25
Don't append to lists

ruchim:
Swapping out List for Vector because appending to the list of metadata events takes a long time and the db is waiting to flush forever.


geoffjentry:
👍 

My experience has shown that in general we shouldn't be using`List` (and thus `Seq`) inside the metadata service. Feel free to change any that you see to `Vector` as well.

rtitle:
👍 
I feel that we should hotfix this in to the next Firecloud release, given the improvement we saw in Alpha. The problem was probably exacerbated recently with the increase to the metadata batch size (bigger batches == more slowdown in List.append).

ruchim:
@geoffjentry I totally agree with @rtitle that this is something that should be added to next FC release as the performance improvement is significant.

geoffjentry:
@ruchim me three

mcovarr:
:+1: for develop too plz

PR #2486, opened 2017-07-25
Don't append to lists (Develop)

ruchim:
append metadata events to a vector not list


Horneth:
Just curious, what does this fix ? 👍

geoffjentry:
@Horneth From what @ruchim and @rtitle said, it made a huge performance difference on FC alpha. Which makes sense as that list was getting appended to and we do a lot of appending there.

Horneth:
@geoffjentry yeah Ruchi explained to me, that's a nice catch !

PR #2488, opened 2017-07-25
Handle failed Cromwell requests

Horneth:


cromwellApiClient/src/main/scala/cromwell/api/CromwellClient.scala
```diff
@@ -101,9 +101,16 @@ class CromwellClient(val cromwellUrl: URL, val apiVersion: String)(implicit acto
...
     entity <- Future.fromTry(decoded.toEntity)
-    unmarshalled <- entity.to[A]
+    unmarshalled <- unmarshall(response, entity)(um, ec)
   } yield unmarshalled
-
+  
+  private def unmarshall[A](response: HttpResponse, entity: Unmarshal[ResponseEntity])(implicit um: Unmarshaller[ResponseEntity, A], ec: ExecutionContext): Future[A] = {
+    import CromwellFailedResponseExceptionJsonSupport._
+    
+    if(response.status.isSuccess()) entity.to[A]
```
mcovarr: `if (`

Horneth:
@mcovarr yes although like all the rest of the CromwellClient it's not :p 
I should be able to add something

cromwellApiClient/src/main/scala/cromwell/api/CromwellClient.scala
```diff
@@ -92,18 +92,27 @@ class CromwellClient(val cromwellUrl: URL, val apiVersion: String)(implicit acto
...
     *
     * @tparam A The type of response expected. Must be supported by an implicit unmarshaller from ResponseEntity.
     */
   private def makeRequest[A](request: HttpRequest)(implicit um: Unmarshaller[ResponseEntity, A], ec: ExecutionContext): Future[A] = for {
-    response <- Http().singleRequest(request)
+    response <- executeRequest(request)
     decoded <- Future.fromTry(decodeResponse(response))
     entity <- Future.fromTry(decoded.toEntity)
-    unmarshalled <- entity.to[A]
+    unmarshalled <- unmarshall(response, entity)(um, ec)
```
danbills: Can omit these implicit args and they'll be passed automagically

Horneth: actually there's an implicit conflict which is why I had to explicitly pass them

PR #2490, opened 2017-07-25
Graceful Server Shutdown

Horneth:
- [x] Add more docs
Reviwers please take a look at https://github.com/broadinstitute/cromwell/wiki and https://github.com/broadinstitute/cromwell/wiki/DevZone
- [x] Test at high scale (Tested at centaur scale)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ before_install:
   - sudo rm -f /etc/boto.cfg
 env:
   global:
-    - CENTAUR_BRANCH=develop
+    - CENTAUR_BRANCH=yesRestart
     - INTEGRATION_TESTS_DIR=src/main/resources/integrationTestCases
   matrix:
     # Setting this variable twice will cause the 'script' section to run twice with the respective env var invoked
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ before_install:
   - sudo rm -f /etc/boto.cfg
 env:
   global:
-    - CENTAUR_BRANCH=yesRestart
+    - CENTAUR_BRANCH=develop
     - INTEGRATION_TESTS_DIR=src/main/resources/integrationTestCases
   matrix:
     # Setting this variable twice will cause the 'script' section to run twice with the respective env var invoked
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,11 @@ retired at the same time.
 
 * The response format of the [callcaching/diff](https://github.com/broadinstitute/cromwell#get-apiworkflowsversioncallcachingdiff) endpoint has been updated.
 
+### Cromwell Server
+
+* Cromwell now attempts to gracefully shutdown when running in server mode and receiving a `SIGINT` (`Ctrl-C`) or `SIGTERM` (`kill`) signal. This includes waiting for all pending Database writes before exiting.
+A detailed explanation and information about how to configure this feature can be found in the [Cromwell Wiki](https://github.com/broadinstitute/cromwell/wiki/DevZone#graceful-server-shutdown).
+
 ## 28
 
 ### Bug Fixes
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -59,11 +59,33 @@ akka {
 
     # Note that without further configuration, all other actors run on the default dispatcher
   }
+
+  coordinated-shutdown.phases {
+    abort-all-workflows {
+      # This phase is used to give time to Cromwell to abort all workflows upon shutdown.
+      # It's only used if system.abort-jobs-on-terminate = true
+      # This timeout can be adusted to give more or less time to Cromwell to abort workflows
+      timeout = 1 hour
+      depends-on = [service-unbind]
+    }
+
+    stop-io-activity{
+      # Adjust this timeout according to the maximum amount of time Cromwell
+      # should be allowed to spend flushing its database queues
+      timeout = 30 minutes
+      depends-on = [service-stop]
+    }
+  }
 }
 
 system {
   # If 'true', a SIGINT will trigger Cromwell to attempt to abort all currently running jobs before exiting
-  #abort-jobs-on-terminate = false
+  abort-jobs-on-terminate = false
+
+  # If 'true', a SIGTERM or SIGINT will trigger Cromwell to attempt to gracefully shutdown in server mode,
+  # in particular clearing up all queued database writes before letting the JVM shut down.
+  # The shutdown is a multi-phase process, each phase having its own configurable timeout. See the Dev Wiki for more details.
+  graceful-server-shutdown = true
 
   # If 'true' then when Cromwell starts up, it tries to restart incomplete workflows
   workflow-restart = true
@@ -82,13 +104,13 @@ system {
 
   # Default number of cache read workers
   number-of-cache-read-workers = 25
-  
+
   io {
     # Global Throttling - This is mostly useful for GCS and can be adjusted to match
     # the quota availble on the GCS API
     number-of-requests = 100000
     per = 100 seconds
-    
+
     # Number of times an I/O operation should be attempted before giving up and failing it.
     number-of-attempts = 5
   }
@@ -200,11 +222,11 @@ engine {
   # You will need to provide the engine with a gcs filesystem
   # Note that the default filesystem (local) is always available.
   filesystems {
-  #  gcs {
-  #    auth = "application-default"
-  #  }
+    #  gcs {
+    #    auth = "application-default"
+    #  }
     local {
-      enabled: true      
+      enabled: true
     }
   }
 }
@@ -242,7 +264,7 @@ backend {
           ${docker} ${script}
         """
 
-        
+
 
         # Root directory where Cromwell writes job results.  This directory must be
         # visible and writeable by the Cromwell process as well as the jobs that Cromwell
@@ -342,7 +364,7 @@ backend {
     #    job-id-regex = "Job <(\\d+)>.*"
     #  }
     #}
-    
+
     #SLURM {
     #  actor-factory = "cromwell.backend.impl.sfs.config.ConfigBackendLifecycleActorFactory"
     #  config {
--- a/core/src/main/scala/cromwell/core/MonitoringCompanionActor.scala
+++ b/core/src/main/scala/cromwell/core/MonitoringCompanionActor.scala
@@ -1,0 +1,46 @@
+package cromwell.core
+
+import akka.actor.{Actor, ActorLogging, ActorRef, Props}
+import cromwell.core.MonitoringCompanionActor._
+import cromwell.util.GracefulShutdownHelper.ShutdownCommand
+
+import scala.concurrent.duration._
+import scala.language.postfixOps
+
+object MonitoringCompanionActor {
+  sealed trait MonitoringCompanionCommand
+  private [core] case object AddWork extends MonitoringCompanionCommand
+  private [core] case object RemoveWork extends MonitoringCompanionCommand
+  private [core] def props(actorToMonitor: ActorRef) = Props(new MonitoringCompanionActor(actorToMonitor))
+}
+
+private [core] class MonitoringCompanionActor(actorToMonitor: ActorRef) extends Actor with ActorLogging {
+  private var workCount: Int = 0
+  
+  override def receive = {
+    case AddWork => workCount += 1
+    case RemoveWork => workCount -= 1
+    case ShutdownCommand if workCount <= 0 =>
+      context stop actorToMonitor
+      context stop self
+    case ShutdownCommand => 
+      log.info(s"{} is still processing {} messages", actorToMonitor.path.name, workCount)
+      context.system.scheduler.scheduleOnce(1 second, self, ShutdownCommand)(context.dispatcher)
+      ()
+  }
+}
+
+trait MonitoringCompanionHelper { this: Actor =>
+  private val monitoringActor = context.actorOf(MonitoringCompanionActor.props(self))
+  private var shuttingDown: Boolean = false
+  
+  def addWork() = monitoringActor ! AddWork
+  def removeWork() = monitoringActor ! RemoveWork
+
+  val monitoringReceive: Receive = {
+    case ShutdownCommand if !shuttingDown => 
+      shuttingDown = true
+      monitoringActor ! ShutdownCommand
+    case ShutdownCommand => // Ignore if we're already shutting down
+  }
+}
--- a/core/src/main/scala/cromwell/util/GracefulShutdownHelper.scala
+++ b/core/src/main/scala/cromwell/util/GracefulShutdownHelper.scala
@@ -1,0 +1,34 @@
+package cromwell.util
+
+import akka.actor.{Actor, ActorLogging, ActorRef, Terminated}
+import akka.pattern.GracefulStopSupport
+import cats.data.NonEmptyList
+import cromwell.util.GracefulShutdownHelper.ShutdownCommand
+
+object GracefulShutdownHelper {
+  case object ShutdownCommand
+}
+
+trait GracefulShutdownHelper extends GracefulStopSupport { this: Actor with ActorLogging =>
+  private var shuttingDown: Boolean = false
+  private var shutdownList: Set[ActorRef] = Set.empty
+  
+  def isShuttingDown: Boolean = shuttingDown
+
+  def waitForActorsAndShutdown(actorsLists: NonEmptyList[ActorRef]): Unit = {
+    if (shuttingDown) {
+      log.error("Programmer error, this actor has already initiated its shutdown. Only call this once per actor !")
+    } else {
+      shuttingDown = true
+      shutdownList = actorsLists.toList.toSet
+      shutdownList foreach context.watch
+      shutdownList foreach { _ ! ShutdownCommand }
+      
+      context become {
+        case Terminated(actor) if shuttingDown && shutdownList.contains(actor) =>
+          shutdownList = shutdownList - actor
+          if (shutdownList.isEmpty) context stop self
+      }
+    }
+  }
+}
--- a/engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
@@ -20,7 +20,6 @@ import net.ceedubs.ficus.Ficus._
 import org.apache.commons.lang3.exception.ExceptionUtils
 
 import scala.concurrent.duration._
-import scala.sys.ShutdownHookThread
 
 object WorkflowManagerActor {
   val DefaultMaxWorkflowsToRun = 5000
@@ -34,6 +33,7 @@ object WorkflowManagerActor {
   /**
     * Commands
     */
+  case object PreventNewWorkflowsFromStarting extends WorkflowManagerActorMessage
   sealed trait WorkflowManagerActorCommand extends WorkflowManagerActorMessage
   case object RetrieveNewWorkflows extends WorkflowManagerActorCommand
   final case class AbortWorkflowCommand(id: WorkflowId, replyTo: ActorRef) extends WorkflowManagerActorCommand
@@ -52,11 +52,10 @@ object WorkflowManagerActor {
             dockerHashActor: ActorRef,
             jobTokenDispenserActor: ActorRef,
             backendSingletonCollection: BackendSingletonCollection,
-            abortJobsOnTerminate: Boolean,
             serverMode: Boolean): Props = {
     val params = WorkflowManagerActorParams(ConfigFactory.load, workflowStore, ioActor, serviceRegistryActor,
-      workflowLogCopyRouter, jobStoreActor, subWorkflowStoreActor, callCacheReadActor, callCacheWriteActor, dockerHashActor, jobTokenDispenserActor, backendSingletonCollection,
-      abortJobsOnTerminate, serverMode)
+      workflowLogCopyRouter, jobStoreActor, subWorkflowStoreActor, callCacheReadActor, callCacheWriteActor,
+      dockerHashActor, jobTokenDispenserActor, backendSingletonCollection, serverMode)
     Props(new WorkflowManagerActor(params)).withDispatcher(EngineDispatcher)
   }
 
@@ -65,6 +64,7 @@ object WorkflowManagerActor {
     */
   sealed trait WorkflowManagerState
   case object Running extends WorkflowManagerState
+  case object RunningAndNotStartingNewWorkflows extends WorkflowManagerState
   case object Aborting extends WorkflowManagerState
   case object Done extends WorkflowManagerState
 
@@ -100,7 +100,6 @@ case class WorkflowManagerActorParams(config: Config,
                                       dockerHashActor: ActorRef,
                                       jobTokenDispenserActor: ActorRef,
                                       backendSingletonCollection: BackendSingletonCollection,
-                                      abortJobsOnTerminate: Boolean,
                                       serverMode: Boolean)
 
 class WorkflowManagerActor(params: WorkflowManagerActorParams)
@@ -117,51 +116,16 @@ class WorkflowManagerActor(params: WorkflowManagerActorParams)
   private val tag = self.path.name
 
   private var abortingWorkflowToReplyTo = Map.empty[WorkflowId, ActorRef]
-  private var shutdownHookThreadOption: Option[ShutdownHookThread] = None
+  private var nextPollCancellable: Option[Cancellable] = None
 
   override def preStart(): Unit = {
-    addShutdownHook()
     // Starts the workflow polling cycle
     self ! RetrieveNewWorkflows
   }
 
-  override def postStop() = {
-    // If the actor is stopping, especially during error tests, then there's nothing to wait for later at JVM shutdown.
-    tryRemoveShutdownHook()
-    super.postStop()
-  }
-
-  private def addShutdownHook() = {
-    // Only abort jobs on SIGINT if the config explicitly sets system.abort-jobs-on-terminate = true.
-    val abortJobsOnTerminate =
-    config.getConfig("system").as[Option[Boolean]]("abort-jobs-on-terminate").getOrElse(params.abortJobsOnTerminate)
-
-    if (abortJobsOnTerminate) {
-      val shutdownHookThread = sys.addShutdownHook {
-        logger.info(s"$tag: Received shutdown signal.")
-        self ! AbortAllWorkflowsCommand
-        while (stateData != null && stateData.workflows.nonEmpty) {
-          log.info(s"Waiting for ${stateData.workflows.size} workflows to abort...")
-          Thread.sleep(1000)
-        }
-      }
-      shutdownHookThreadOption = Option(shutdownHookThread)
-    }
-  }
-
-  private def tryRemoveShutdownHook() = {
-    try {
-      shutdownHookThreadOption.foreach(_.remove())
-    } catch {
-      case _: IllegalStateException => /* ignore, we're probably shutting down */
-      case exception: Exception => log.error(exception, "Error while removing shutdown hook: {}", exception.getMessage)
-    }
-    shutdownHookThreadOption = None
-  }
-
   startWith(Running, WorkflowManagerData(workflows = Map.empty))
 
-  when (Running) {
+  val runningAndNotStartingNewWorkflowsStateFunction: StateFunction = {
     /*
      Commands from clients
      */
@@ -175,12 +139,10 @@ class WorkflowManagerActor(params: WorkflowManagerActorParams)
       stay()
     case Event(WorkflowStoreEngineActor.NoNewWorkflowsToStart, _) =>
       log.debug("WorkflowStore provided no new workflows to start")
-      scheduleNextNewWorkflowPoll()
       stay()
     case Event(WorkflowStoreEngineActor.NewWorkflowsToStart(newWorkflows), stateData) =>
       val newSubmissions = newWorkflows map submitWorkflow
       log.info("Retrieved {} workflows from the WorkflowStoreActor", newSubmissions.toList.size)
-      scheduleNextNewWorkflowPoll()
       stay() using stateData.withAddition(newSubmissions)
     case Event(SubscribeToWorkflowCommand(id), data) =>
       data.workflows.get(id) foreach {_ ! SubscribeTransitionCallBack(sender())}
@@ -229,6 +191,16 @@ class WorkflowManagerActor(params: WorkflowManagerActorParams)
       stay using data.without(workflowActor)
   }
 
+  val scheduleNextNewWorkflowPollStateFunction: StateFunction = {
+    case event @ Event(WorkflowStoreEngineActor.NoNewWorkflowsToStart | _: WorkflowStoreEngineActor.NewWorkflowsToStart, _) =>
+      scheduleNextNewWorkflowPoll()
+      runningAndNotStartingNewWorkflowsStateFunction(event)
+  }
+  
+  when (Running) (scheduleNextNewWorkflowPollStateFunction.orElse(runningAndNotStartingNewWorkflowsStateFunction))
+  
+  when (RunningAndNotStartingNewWorkflows) (runningAndNotStartingNewWorkflowsStateFunction)
+
   when (Aborting) {
     case Event(Transition(workflowActor, _, toState: WorkflowActorState), data) if toState.terminal =>
       // Remove this terminal actor from the workflowStore and log a progress message.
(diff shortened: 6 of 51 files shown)


geoffjentry:
@Horneth In order to assist with my squawking and other bodily functions, could you write up a bit about how this works?

core/src/main/resources/reference.conf
```diff
@@ -59,11 +59,29 @@ akka {
...
+    }
+  }
 }
 
 system {
   # If 'true', a SIGINT will trigger Cromwell to attempt to abort all currently running jobs before exiting
-  #abort-jobs-on-terminate = false
+  abort-jobs-on-terminate = false
+  
+  # If 'true', a SIGINT will trigger Cromwell to attempt to gracefully shutdown in server mode, in particular clearing up all
```
geoffjentry: Should also trap SIGTERM as that's the default signal sent from `kill`. The standard behavior is that SIGTERM will try to initiate a graceful shutdown and SIGKILL will, well, kill. You might want to also include SIGHUP as well as some people do that first.

core/src/main/resources/reference.conf
```diff
@@ -59,11 +59,29 @@ akka {
...
 }
 
 system {
   # If 'true', a SIGINT will trigger Cromwell to attempt to abort all currently running jobs before exiting
-  #abort-jobs-on-terminate = false
+  abort-jobs-on-terminate = false
+  
+  # If 'true', a SIGINT will trigger Cromwell to attempt to gracefully shutdown in server mode, in particular clearing up all 
+  # queued database writes before letting the JVM shut down.
+  # There's a 30 seconds timeout after which Cromwell will forcefully shutdown if it hasn't already.
```
geoffjentry: what about making the timeout configurable?

(egads, @geoffjentry suggests a tunable knob!)

core/src/main/scala/cromwell/core/MonitoringCompanionActor.scala
```diff
@@ -0,0 +1,45 @@
...
+
+import akka.actor.{Actor, ActorLogging, ActorRef, Props}
+import cromwell.core.MonitoringCompanionActor._
+import cromwell.util.GracefulShutdownHelper.ShutdownCommand
+
+import scala.concurrent.duration._
+import scala.language.postfixOps
+
+object MonitoringCompanionActor {
+  private [core] case object AddWork
```
geoffjentry: Tag these as extending some base trait to match the rest of our actors

core/src/main/scala/cromwell/util/GracefulShutdownHelper.scala
```diff
@@ -0,0 +1,35 @@
...
+  private [util] var shuttingDown: Boolean = false
+  private var shutdownList: Set[ActorRef] = Set.empty
+  
+  def isShuttingDown: Boolean = shuttingDown
+
+  def waitForActorsAndShutdown(actorsLists: NonEmptyList[ActorRef]): Unit = {
+    if (shuttingDown) {
+      log.error("Programmer error, this actor has already initiated its shutdown. Only call this once per actor !")
+    }
+    else {
```
mcovarr: `} else {`

engine/src/main/scala/cromwell/server/CromwellShutdown.scala
```diff
@@ -0,0 +1,219 @@
...
+    */
+  def registerUnbindTask(actorSystem: ActorSystem, serverBinding: Future[Http.ServerBinding]) = {
+    instance(actorSystem).addTask(CoordinatedShutdown.PhaseServiceUnbind, "UnbindingServerPort") { () =>
+      // At this point it's still safe to schedule work on the actor system's dispatcher
+      implicit val ec = actorSystem.dispatcher
+      serverBinding flatMap { binding =>
+        binding.unbind() map { _ =>
+          logger.info("Http server unbound.")
+          Done
+        }
```
mcovarr: looks like this can be `for`comped

core/src/main/resources/reference.conf
geoffjentry: Sure, but only in the same way that this whole PR is changing behavior. One typically shuts down a long lived server via `kill` which sends a SIGTERM. The point of this exercise is to provide graceful shutdowns for the server. But yes, shutdown procedures should be documented

core/src/main/resources/reference.conf
geoffjentry: Who *wouldn't* ask for @abaumann?

core/src/main/scala/cromwell/util/GracefulShutdownHelper.scala
```diff
@@ -0,0 +1,35 @@
...
+      shuttingDown = true
+      shutdownList = actorsLists.toList.toSet
+      shutdownList foreach context.watch
+      shutdownList foreach { _ ! ShutdownCommand }
+      
+      context become {
```
geoffjentry: @kshakir will be so proud

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -150,35 +152,21 @@ case class WorkflowExecutionActor(workflowDescriptor: EngineWorkflowDescriptor,
     FSM.NullFunction
   }
   when(WorkflowExecutionFailedState) {
-    alreadyFailedMopUp
```
geoffjentry: Once in a while I see a name that makes me chuckle and feel the need to go look up who created it. 99.99999999999999% of the time the answer is @cjllanwarne

engine/src/main/scala/cromwell/server/CromwellRootActor.scala
```diff
@@ -108,16 +117,40 @@ import scala.language.postfixOps
...
+      Try(Await.result(gracefulStop(workflowManagerActor, AbortTimeout, AbortAllWorkflowsCommand), AbortTimeout)) match {
+        case Success(_) => logger.info("All workflows aborted")
+        case Failure(f) => logger.error("Failed to abort workflows", f)
+      }
+    }
+  }
+
   override def receive = {
-    case _ => logger.error("CromwellRootActor is receiving a message. It prefers to be left alone!")
+    case _ => logger.error("Unknown message received by CromwellRootActor")
```
geoffjentry: I thought you were in the "it'll show up in dead letters" camp

core/src/main/resources/reference.conf
mcovarr: Yeah this PR changes behavior, but it's probably changing behavior to what users thought Cromwell was doing before this PR.  🙂

geoffjentry: "Wait, you all run workflows? who knew?"

Horneth: This is actually a mistake it should say `SIGTERM`. I saw this in the coordinated shutdown docs that made me wary `By default, the CoordinatedShutdown will be run when the JVM process exits, e.g. via kill SIGTERM signal (SIGINT ctrl-c doesn’t work)`.
However I'm not sure why they say that because looking at the code they just add a normal JVM shutdown hook.

src/bin/travis/testCentaurTes.sh
```diff
@@ -58,7 +58,7 @@ mkdir go-lib
 export GOPATH=$WORKDIR/go-lib
 go get github.com/ohsu-comp-bio/funnel
 cd $GOPATH/src/github.com/ohsu-comp-bio/funnel
-git checkout 04c5e03
+git checkout c4d9134
```
kshakir: Curious/ToL: What is this hash change?

core/src/main/resources/reference.conf
Horneth: Yeah I meant to make a comment about that but y'all are too fast :p 
The actual global timeout is the sum of all phases' timeouts.
Every phase has a 5s default timeout, but they can all be changed through configuration. Which means the timeout *can* be changed through the conf, it's just more than one which makes it hard to give a definitive number statically here.

core/src/main/resources/reference.conf
```diff
@@ -59,11 +59,29 @@ akka {
 
     # Note that without further configuration, all other actors run on the default dispatcher
   }
+
+  coordinated-shutdown.phases {
+    abort-all-workflows-activity {
+      depends-on = [service-unbind]
+      timeout = 1 hour
```
Horneth: The previous timeout was `Infinity` which still makes this one infinitely shorter but yeah I don't know what a sensible value would be here :/

engine/src/main/scala/cromwell/server/CromwellRootActor.scala
Horneth: I am !

src/bin/travis/testCentaurTes.sh
Horneth: This the head of develop for funnel currently. The reason is incorporating this fix https://github.com/ohsu-comp-bio/funnel/pull/164 that allows the call caching caopeira centaur test to survive a restart mid-test in TES.
However it seems to have broken another test, namely `non_root_default_user`. I excluded it for now but will follow up with the OHSU guys to figure out what's happening

Horneth:
@geoffjentry @mcovarr The `CromwellShutdown` object already documents the different phases but I'm happy to write more high-level docs about what's going on. What's a good place for that ? Seems a bit technical for the README but maybe not ?

engine/src/main/scala/cromwell/server/CromwellRootActor.scala
Horneth: Although in this case since the `receive` method has to be overridden and we really don't expect the CromwellRootActor to receive *any* message I think it makes sense to keep this as a logged error

engine/src/main/scala/cromwell/server/CromwellShutdown.scala
```diff
@@ -0,0 +1,220 @@
...
+  /**
+    * Returns true of the coordinated shutdown process is in progress. False otherwise.
+    */
+  def shutdownInProgress(): Boolean = _shutdownInProgress.get()
+
+  /**
+    * Register a task to unbind from the port during the ServiceUnbind phase.
+    */
+  def registerUnbindTask(actorSystem: ActorSystem, serverBinding: Future[Http.ServerBinding]) = {
+    instance(actorSystem).addTask(CoordinatedShutdown.PhaseServiceUnbind, "UnbindingServerPort") { () =>
```
mcovarr: should be able to `_` the `()`

engine/src/main/scala/cromwell/server/CromwellShutdown.scala
```diff
@@ -0,0 +1,220 @@
...
+import cromwell.services.SingletonServicesStore
+import cromwell.util.GracefulShutdownHelper.ShutdownCommand
+import org.slf4j.LoggerFactory
+
+import scala.concurrent.duration._
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.{Failure, Success}
+
+/**
+  * Collection of methods and object used to control Cromwell graceful shutdown process.
```
mcovarr: Collection of methods and object used => Used

engine/src/main/scala/cromwell/server/CromwellShutdown.scala
```diff
@@ -0,0 +1,220 @@
...
+                             materializer: ActorMaterializer
+                           ): Unit = {
+
+    val coordinatedShutdown = this.instance(actorSystem)
+
+    def shutdownActor(actor: ActorRef,
+                      phase: String,
+                      message: AnyRef,
+                      customTimeout: Option[FiniteDuration] = None)(implicit executionContext: ExecutionContext) = {
+      coordinatedShutdown.addTask(phase, s"stop${actor.path.name.capitalize}") { () =>
```
mcovarr: `()` => `_`

engine/src/main/scala/cromwell/server/CromwellShutdown.scala
```diff
@@ -0,0 +1,220 @@
...
+      import akka.pattern.ask
+      _shutdownInProgress.set(true)
+      implicit val timeout = Timeout(coordinatedShutdown.timeout(CoordinatedShutdown.PhaseBeforeServiceUnbind))
+      (workflowManagerActor ? InterruptNewWorkflowStart) map { _ =>
+        logger.info("Workflow polling stopped")
+        Done
+      }
+    }
+
+    /* 1') In server mode, the socket is unbound from the port in the CoordinatedShutdown.PhaseServiceUnbind
```
mcovarr: looks like that 1 is getting emotional about the Cromwell shutdown

engine/src/main/scala/cromwell/server/CromwellShutdown.scala
```diff
@@ -0,0 +1,220 @@
...
+      shutdownActor(workflowManagerActor, CoordinatedShutdown.PhaseServiceRequestsDone, PoisonPill)
+    }
+
+    /* 3) Shutdown connection pools
+      * This will close all akka http opened connection pools tied to the actor system.
+      * The pools stop accepting new work but are given a chance to execute the work submitted prior to the shutdown call.
+      * When this future returns, all outstanding connections to client will be terminated.
+      * Note that this also includes connection pools like the one used to lookup docker hashes.
+     */
+    coordinatedShutdown.addTask(CoordinatedShutdown.PhaseServiceStop, "TerminatingConnections") { () =>
```
mcovarr: `_`

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
mcovarr: 🇬🇧 👑

engine/src/main/scala/cromwell/server/CromwellShutdown.scala
```diff
@@ -0,0 +1,220 @@
...
+      Http(actorSystem).shutdownAllConnectionPools() map { _ =>
```
mcovarr: TIL Cats `Functor` has `as`:

```
  /**
   * Replaces the `A` value in `F[A]` with the supplied value.
   */
  def as[A, B](fa: F[A], b: B): F[B] = map(fa)(_ => b)
```

so here:

```
import cats.syntax.functor._
.
.
.
    Http(actorSystem).shutdownAllConnectionPools() as {
     logger.info("Connection pools shut down")
     Done
    }
```

engine/src/main/scala/cromwell/server/CromwellShutdown.scala
Horneth: Used with capital U ?

mcovarr: yes, I was suggesting to get rid of the first part of the sentence and start it with "Used".  🙂

engine/src/main/scala/cromwell/server/CromwellShutdown.scala
Horneth: That gives me a `Missing parameter type`

mcovarr: 💩

core/src/main/resources/reference.conf
```diff
@@ -59,11 +59,34 @@ akka {
 
     # Note that without further configuration, all other actors run on the default dispatcher
   }
+
+  coordinated-shutdown.phases {
+    # This phase is used to give time to Cromwell to abort all workflows upon shutdown. It's only used if system.abort-jobs-on-terminate = true
+    abort-all-workflows {
+      # /!\ This depends-on field should NOT be modified
+      depends-on = [service-unbind]
```
geoffjentry: Why is this an option if it shouldn't be modified?

core/src/main/resources/reference.conf
```diff
@@ -59,11 +59,34 @@ akka {
...
+      # Adjust this timeout according to the maximum time you're ready to wait for to allow Cromwell to abort all workflows on shutdown
```
geoffjentry: I think this would read a little more doc-y minus the "you're ready to wait for" bit, but @mcovarr is the wordsmith here

CHANGELOG.md
```diff
@@ -16,6 +16,11 @@ retired at the same time.
 
 * The response format of the [callcaching/diff](https://github.com/broadinstitute/cromwell#get-apiworkflowsversioncallcachingdiff) endpoint has been updated.
 
+### Cromwell Server
+
+* Cromwell now attempts to gracefully shutdown when running in server mode and receiving a `SIGINT` or `SIGTERM` signal. This includes waiting for all pending Database writes before exiting.
```
geoffjentry: I'd add in something making it clear that `SIGINT` == `^C` and `SIGTERM` == `kill`, most people are going to think in those terms

core/src/main/scala/cromwell/util/GracefulShutdownHelper.scala
```diff
@@ -0,0 +1,34 @@
...
+import akka.pattern.GracefulStopSupport
+import cats.data.NonEmptyList
+import cromwell.util.GracefulShutdownHelper.ShutdownCommand
+
+object GracefulShutdownHelper {
+  case object ShutdownCommand
+}
+
+trait GracefulShutdownHelper extends GracefulStopSupport { this: Actor with ActorLogging =>
+  private [util] var shuttingDown: Boolean = false
```
geoffjentry: I'm assuming this is for testing but I really don't like the idea of having mutable state in actors exposed at all. Any way around this?

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -20,13 +20,13 @@ import net.ceedubs.ficus.Ficus._
...
 
 import scala.concurrent.duration._
-import scala.sys.ShutdownHookThread
 
 object WorkflowManagerActor {
   val DefaultMaxWorkflowsToRun = 5000
   val DefaultMaxWorkflowsToLaunch = 50
   val DefaultNewWorkflowPollRate = 20
 
+  case object PreventNewWorkflowsFromStarting
```
geoffjentry: Have this extend something to make it clear it's a message

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -117,48 +115,14 @@ class WorkflowManagerActor(params: WorkflowManagerActorParams)
   private val tag = self.path.name
 
   private var abortingWorkflowToReplyTo = Map.empty[WorkflowId, ActorRef]
-  private var shutdownHookThreadOption: Option[ShutdownHookThread] = None
+  private var nextPollCancellable: Option[Cancellable] = None
+  private var preventNewWorkflowsFromStarting: Boolean = false
```
geoffjentry: Isn't this effectively creating a new state in the FSM w/o making a new state?

engine/src/main/scala/cromwell/server/CromwellRootActor.scala
```diff
@@ -63,12 +68,12 @@ import scala.language.postfixOps
...
   lazy val subWorkflowStore = new SqlSubWorkflowStore(SingletonServicesStore.databaseInterface)
   lazy val subWorkflowStoreActor = context.actorOf(SubWorkflowStoreActor.props(subWorkflowStore), "SubWorkflowStoreActor")
-  
+
   // Io Actor
   lazy val throttleElements = systemConfig.as[Option[Int]]("io.number-of-requests").getOrElse(100000)
   lazy val throttlePer = systemConfig.as[Option[FiniteDuration]]("io.per").getOrElse(100 seconds)
   lazy val ioThrottle = Throttle(throttleElements, throttlePer, throttleElements)
-  lazy val ioActor = context.actorOf(IoActor.props(1000, Option(ioThrottle)).withDispatcher(Dispatcher.IoDispatcher))
+  lazy val ioActor = context.actorOf(IoActor.props(1000, Option(ioThrottle)).withDispatcher(Dispatcher.IoDispatcher), "IoActor")
```
geoffjentry: This makes me realize that we should normalize how we add dispatchers. Some are done on the `actorOf`, some are hardcoded into the props.

engine/src/main/scala/cromwell/server/CromwellShutdown.scala
```diff
@@ -0,0 +1,222 @@
...
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.{Failure, Success}
+
+/**
+  * Collection of methods and objects used to control Cromwell graceful shutdown process.
+  */
+object CromwellShutdown extends GracefulStopSupport {
+  private val logger = LoggerFactory.getLogger("CromwellShutdown")
+  
+  // Includes DB writing actors, I/O Actor and DockerHashActor
```
geoffjentry: IMO these would read better as `FooPhase` vs `PhaseFoo`. It took me a moment to realize what this meant and I'm reading the PR.

engine/src/main/scala/cromwell/server/CromwellShutdown.scala
```diff
@@ -0,0 +1,222 @@
...
+  private val PhaseStopIoActivity = "stop-io-activity"
+  // Shutdown phase allocated when "abort-jobs-on-terminate" is true to give time to the system to abort all workflows
+  // This phase is at the same level as PhaseServiceRequestsDone in the dependency graph.
+  private val PhaseAbortAllWorkflows = "abort-all-workflows"
+
+  /**
+    * Configuration overriding pre-existing phases to wire in custom phases in the dependency graph.
+    */
+  val OverrideShutdownConfiguration = ConfigFactory.parseString(
```
geoffjentry: I don't remember why you said these were here, but I remember thinking that these were superfluous now. Is that true?

engine/src/main/scala/cromwell/server/CromwellShutdown.scala
```diff
@@ -0,0 +1,222 @@
...
+          case Failure(f) => logger.error(s"An error occurred trying to gracefully stop ${actor.path.name}.", f)
+        }
+        
+        action map { _ => Done }
+      }
+    }
+
+    implicit val ec = actorSystem.dispatcher
+
+    // 1) Stop starting new workflow. This will stop the WMA from sending pull messages to the WorkflowStore
```
geoffjentry: "new workflows"

engine/src/main/scala/cromwell/server/CromwellShutdown.scala
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    /* 5) Stop system level actors that require writing to the database or I/O
+      * - SubWorkflowStoreActor
+      * - JobStoreActor
+      * - CallCacheWriteActor
+      * - ServiceRegistryActor
+      * - DockerHashActor
+      * - IoActor
+    */
```
geoffjentry: Maybe drop a comment over where these are defined so that if someone adds a new one it gets added to the list

geoffjentry:
👍

core/src/main/resources/reference.conf
Horneth: It's an addition to the pre-defined list of akka phases so it needs to be in the conf. But if someone changes the value it will mess up the order in which the phases are executed which we don't want.
The alternative would be to hardcode it like I did for the overriding part of the config. However like you mentioned in another comment we can apparently put it in the `application.conf`, which I'm ok with as long as it's absolutely clear that this will NOT work properly if someone does not include our `application.conf` in theirs (or copy the whole thing...).

I realize this might not make a lot of sense so we can TechTalk it if you want :)

core/src/main/scala/cromwell/util/GracefulShutdownHelper.scala
Horneth: It was not for test but that makes me realize that this could actually use some testing :)

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
Horneth: Yes and no, it's very slightly modifying the behavior of the "Running" state by deciding whether or not a method should be called, which could be said to be a different state I agree. I can try to play with state functions to try and make it so, what I wanted to avoid was copy/pasting the whole "Running" state function to a new "Running but not polling" one with just 2 lines being removed.

engine/src/main/scala/cromwell/server/CromwellShutdown.scala
Horneth: Yeah agreed, the reason is that the akka coordinated shutdown pre-defined ones use this syntax so I wanted to stay consistent.

engine/src/main/scala/cromwell/server/CromwellShutdown.scala
Horneth: See my answer to your other comment but in short yes as long as we're ok with making it mandatory for a user to include `application.conf` in their conf.

engine/src/main/scala/cromwell/server/CromwellShutdown.scala
Horneth: The list is right below this comment :)

core/src/main/resources/reference.conf
geoffjentry: I have a feeling our entire `akka` block belongs in `application.conf` (@kshakir you're the expert on these things). Just keep it here since the rest of the akka stuff is here and we can address separately

core/src/main/scala/cromwell/util/GracefulShutdownHelper.scala
geoffjentry: Can you make it private-private instead of package private then?

And yeah, it probably could :)

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
geoffjentry: I was figuring the last part was the reason :)

It did seem like it'd be a waste to do that but OTOH I'm afraid of setting up a slippery slope where we start having states and pseudo-states mixed all over the place. I'm sure you can find examples of us doing this already as well as examples of lots of copypasta due to keeping it clean.

engine/src/main/scala/cromwell/server/CromwellShutdown.scala
geoffjentry: okee doke

engine/src/main/scala/cromwell/server/CromwellShutdown.scala
geoffjentry: Ok. 

Based on what I said up above I'm thinking these should go into our `akka` block in `reference.conf` an then the question of if that stuff belongs there or `application.conf` can be addressed separately.

There's all kinds of stuff that people are going to have a bad time with if they override the wrong way.

engine/src/main/scala/cromwell/server/CromwellShutdown.scala
geoffjentry: Yeah, I'm not quite *that* blind .... :)

core/src/main/resources/reference.conf
Horneth: So we touched on this at TechTalk and people seemed to agree that in order to prevent users from shooting themselves in the foot by changing those values we'd just hardcode them and see if someone complains. I'll leave the timeout values in the conf though so those CAN be overridden

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
Horneth: I think I found a way to do it all with states - it's arguably cleaner but the `var` is gone. I'll push the change once I've finished addressing the rest of the comments

core/src/main/resources/reference.conf
geoffjentry: I'm complaining ;)

Why here? Are we going to do that for *everything*? You can easily make Cromwell explode by doing dumb stuff in a config. What's special about drawing the line here on this one?

I really don't like the idea of putting these values in extremely non-standard places just because someone might be a dumbass and override stuff that isn't documented as something they should be playing with.

Horneth: I hear you. I guess in my mind anything that's in the config *should* be meant to be configurable / changed in some case that warrants it. I feel like we could make a case that this statement is true for ~~almost~~ everything currently in our conf ? Maybe it isn't in which case yes there's no reason to make an exception. My main argument is that I really don't see *any* reason why someone would want to change that - or even be aware that it exists.

geoffjentry: One quick example - if you remove the defined dispatchers you're going to have a bad time.

I see that argument but this is also a config file for us as well. If a developer wants to tweak something, the obvious place to do it is the config file. Now it is who knows where.

My stance on the user side of the config file is they should only ever be using a file that only contains the specific overrides that they want, and those overrides should have been explicitly pointed out to them in the docs as something they might want to change.

But this is also why the `akka` block should be in `application.conf` and not `reference.conf`, these aren't settings that we're providing to the outside world, they're settings that *we* are using.

geoffjentry: To put it another way, when I grab some other code using typesafe config, I don't assume that any random thing in there is something I should be messing with. I only change things if I see it documented as affecting behavior that I want to see changed.

Horneth: I think you're starting to convince me :) I'm fine with putting things in the conf and I don't like having "config strings" hardcoded anyway. If we go that route I think we could at least ping workbench and gotc to make sure they're aware that the *must* include `application.conf` and *must not* mess around with those shutdown values

geoffjentry: I think they pretty much have to already for it to work (plus, the `akka` block should be there, and w/o the dispatchers it will fail).

Hold on a bit though, in between practice runs of this talk I'm trying to see what other people do.

geoffjentry: What about something like [this?](https://groups.google.com/forum/#!topic/akka-user/A4MB5sNLm-s)

This person had a separate conf file for their akka stanza which was loaded in separately. That keeps the config-ness to it all but does put it in a slightly less accessible location.

Although now that I say this i think my dispatcher point isn't true, I don't think you'd be able to remove one via an override. It's like the issue we had w/ the local backend.

Horneth: Hmm I'm lost now -_-

geoffjentry: on my last bit I was saying I realized I was basing my argument partially on an incorrect statement although I still prefer keeping things in config format instead of hardcoded here and there in the code.

That link at the top was something I saw which seemed like an interesting middle ground between the two if one really wanted to try to obscure "dangerous" bits from a user, although it has some of the same issues as putting stuff in non-standard places.

Horneth: Mokay, what I'm hearing is put things into reference.conf / application.conf, assume that people know to include application.conf and do not care about the fact that if they change those values they'll mess up everything.

Horneth: Which means the answer to your original question `Why is this an option if it shouldn't be modified?` would be "because it's akka conf that we need to add and users should not touch it if they don't know what they're doing"

core/src/main/scala/cromwell/util/GracefulShutdownHelper.scala
Horneth: Yep !

core/src/main/resources/reference.conf
geoffjentry: @Horneth That'd be my preference, granted I wasn't at tech talk today, nor will I be there tomorrow. I do see the general conundrum but I'm afraid down the road this pattern will cause more problems than it solves.

On your second point - yeah. When I first asked that I was just scrolling through the PR and had no idea what was going on :)

.travis.yml
```diff
@@ -22,7 +22,7 @@ before_install:
   - sudo rm -f /etc/boto.cfg
 env:
   global:
-    - CENTAUR_BRANCH=develop
+    - CENTAUR_BRANCH=yesRestart
```
mcovarr: I can imagine what that branch looks like, but I haven't seen a PR for it.  🙂

Horneth: Here it is ! https://github.com/broadinstitute/centaur/pull/224 :)

Horneth:
@mcovarr @geoffjentry Like I mentioned at standup I added a second commit after seeing a failure in centaur.
What happened was JES failed the job because it couldn't localize the auth file (not found).
However I didn't see anything in Cromwell suggesting that the upload failed (which we log if it happens). So my guess is JES tried to localize the file when Cromwell was restarting the workflow and hence re-writing the file, which made sense according to the timestamps at least. This commit makes the upload of the auth file fail if the file already exists, unless it's a known restart in which case it ignores the failure and keeps going.

I think it makes sense to fail the workflow if there's already an auth file for this workflow and it's the *first* time we run it. It might indicate something is wrong and failing the workflow avoids taking chances with refresh tokens / secrets. If you disagree please voice your concerns :)

geoffjentry:
👍

PR #2493, opened 2017-07-27
Dispatcherizing undispatcherized actors. Trying to standardize on add…

geoffjentry:
…ing it to the props where appropriate

### DIFF
--- a/core/src/main/scala/cromwell/util/PromiseActor.scala
+++ b/core/src/main/scala/cromwell/util/PromiseActor.scala
@@ -1,7 +1,7 @@
 package cromwell.util
 
 import akka.actor._
-
+import cromwell.core.Dispatcher.EngineDispatcher
 import scala.concurrent.{Future, Promise}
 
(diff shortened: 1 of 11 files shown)


core/src/main/scala/cromwell/util/PromiseActor.scala
```diff
@@ -42,7 +42,7 @@ object PromiseActor {
     promise.future
   }
 
-  def props(promise: Promise[Any], sendTo: ActorRef, msg: Any): Props = Props(new PromiseActor(promise, sendTo, msg))
+  def props(promise: Promise[Any], sendTo: ActorRef, msg: Any): Props = Props(new PromiseActor(promise, sendTo, msg)).withDispatcher(EngineDispatcher)
```
geoffjentry: This one wasn't clear to me as to which dispatcher to use. It's not in `engine`, but it's fully self contained in here and this tidbit is only called in `engine`.

PR #2499, opened 2017-07-27
Dispatcherizing undispatcherized actors. Trying to standardize on add…

geoffjentry:
…ing it to the props where appropriate

### DIFF
--- a/engine/src/main/scala/cromwell/engine/workflow/WorkflowDockerLookupActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/WorkflowDockerLookupActor.scala
@@ -9,6 +9,8 @@ import cromwell.docker.{DockerClientHelper, DockerHashRequest, DockerHashResult,
 import cromwell.engine.workflow.WorkflowActor.{RestartExistingWorkflow, StartMode}
 import cromwell.engine.workflow.WorkflowDockerLookupActor._
 import cromwell.services.SingletonServicesStore
+import cromwell.core.Dispatcher.EngineDispatcher
+
 import lenthall.util.TryUtil
 
 import scala.concurrent.duration._
@@ -249,7 +251,7 @@ object WorkflowDockerLookupActor {
   final case class WorkflowDockerTerminalFailure(reason: Throwable, request: DockerHashRequest) extends WorkflowDockerLookupResponse
 
   def props(workflowId: WorkflowId, dockerHashingActor: ActorRef, startMode: StartMode, databaseInterface: SqlDatabase = SingletonServicesStore.databaseInterface) = {
-    Props(new WorkflowDockerLookupActor(workflowId, dockerHashingActor, startMode, databaseInterface))
+    Props(new WorkflowDockerLookupActor(workflowId, dockerHashingActor, startMode, databaseInterface)).withDispatcher(EngineDispatcher)
   }
 
   object WorkflowDockerLookupActorData {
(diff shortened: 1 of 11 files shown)


mcovarr:
:+1: I like dis patch.

danbills:
why not leave these dispatch settings to the caller?  i.e. CromwellRootActor

geoffjentry:
@danbills What caught my eye was noticing we were doing both and being armed with fore-knowledge that we're *really* bad at remembering to add these in general.

When I went in my thought was to harmonize it one way or the other and in every case I found all actors of a given class were assigned the same dispatcher, and in some cases they're instantiated more than once. I was a little surprised by this, I was expecting that it woudn't end up uniform in the end.

I don't really care, my main goal is to find ways to make it less frequently forgotten about. If people would rather these be on all the call sites than the definition, it's fine by me - and frankly, if we were reusing these things w/ different dispatchers I'd totally agree w/ you. 

What I *do* want is an easier time when I go scanning for rogue actors.

danbills:
yeah I hesitated to argue because I know these are very stable and basically will never be used differently.  I just kinda like the configuration all in one place personally.  The one place also makes the scanning effort easier.

geoffjentry:
Granted all of our dispatchers are defaulting to untuned forkjoinpools so it doesn't make much of a difference either way :)

engine/src/main/scala/cromwell/engine/workflow/WorkflowDockerLookupActor.scala
```diff
@@ -249,7 +251,7 @@ object WorkflowDockerLookupActor {
   final case class WorkflowDockerTerminalFailure(reason: Throwable, request: DockerHashRequest) extends WorkflowDockerLookupResponse
 
   def props(workflowId: WorkflowId, dockerHashingActor: ActorRef, startMode: StartMode, databaseInterface: SqlDatabase = SingletonServicesStore.databaseInterface) = {
-    Props(new WorkflowDockerLookupActor(workflowId, dockerHashingActor, startMode, databaseInterface))
+    Props(new WorkflowDockerLookupActor(workflowId, dockerHashingActor, startMode, databaseInterface)).withDispatcher(EngineDispatcher)
```
cjllanwarne: `WorkflowDockerLookup` sounds IO-ey?

Or is `IODispatcher` a specialer case?

geoffjentry: Good question 🤔

mcovarr: This actor is mostly doing DB stuff, so whatever's appropriate for that.  It's the `DockerHashActor` that actually chats up the Docker repos.

geoffjentry: It looks like actors of this one's ilk are EngineDispatchers

cjllanwarne: Questions because I don't really understand all this stuff:
- Does the engine dispatcher also handle all of the mission-critical EJEAs, JABJEAs, WMAs, etc?
- Are there benefits of having a `DbDispatcher` for DB stuff?

geoffjentry: - Yes
- Perhaps? Slick itself has its own threadpool

This is all a little academic right now as no one bothers to tune these dispatchers for their installation so in some ways they're not really all that useful. CaaS will give us an installation that'll be worth tuning though. I'd rather not prematurely add a ton of granularity that we might never use, let's dogfood what's there over the next few months and use that to drive how to change

PR #2500, opened 2017-07-27
Replaced use of Future.foreach with error handling

kshakir:


engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreSubmitActor.scala
```diff
@@ -21,19 +21,41 @@ final case class WorkflowStoreSubmitActor(store: WorkflowStore, serviceRegistryA
...
-        sndr ! WorkflowsBatchSubmittedToStore(ids)
-        log.info("Workflows {} submitted.", ids.toList.mkString(", "))
+
+      val futureIds = for {
+        ids <- storeWorkflowSources(cmd.sources)
+        _ <- Future sequence {
+        ids.toList zip cmd.sources.toList map {
+            case (id, source) => registerSubmissionWithMetadataService(id, source)
+          }
+        }
```
mcovarr:
```scala
import cats.instances.list._
import cats.syntax.traverse._
.
.
.
      val registerSubmission = (registerSubmissionWithMetadataService _).tupled

      val futureIds = for {
        ids <- storeWorkflowSources(cmd.sources)
        _ <- (ids.toList zip cmd.sources.toList) traverse registerSubmission
      } yield ids
```

kshakir:
Having a hard time finding the right place to unit/integration test these changes using existing specs/mocks/centaur.

If tests are required, a good bit of test refactoring will need to follow to cut through the Akka-HTTP-`Route`-to-`SubmitActor`-to-`ServicesActor`s layers.

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreSubmitActor.scala
```diff
@@ -21,19 +21,41 @@ final case class WorkflowStoreSubmitActor(store: WorkflowStore, serviceRegistryA
   override def receive = {
     case cmd: SubmitWorkflow =>
       val sndr = sender()
-      storeWorkflowSources(NonEmptyList.of(cmd.source)) foreach { ids =>
-        val id = ids.head
-        registerSubmissionWithMetadataService(id, cmd.source)
-        sndr ! WorkflowSubmittedToStore(id)
-        log.info("Workflow {} submitted.", id)
+
```
geoffjentry: didn't @Horneth just fiddle w/ this in another PR? or is it the same idea but different thing?

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
```diff
@@ -197,25 +197,39 @@ trait CromwellApiService {
...
             val e = new IllegalArgumentException("To submit more than one workflow at a time, use the batch endpoint.")
             e.failRequest(StatusCodes.BadRequest)
           case Success(workflowSourceFiles) =>
-            onComplete(workflowStoreActor.ask(WorkflowStoreActor.BatchSubmitWorkflows(NonEmptyList.fromListUnsafe(workflowSourceFiles.toList))).mapTo[WorkflowStoreSubmitActor.WorkflowsBatchSubmittedToStore]) {
-              case Success(w) =>
-                val responses = w.workflowIds map { id => WorkflowSubmitResponse(id.toString, WorkflowSubmitted.toString) }
-                complete((StatusCodes.Created, responses.toList))
-              case Failure(e) => e.failRequest(StatusCodes.InternalServerError)
-            }
+            askSubmit(WorkflowStoreActor.BatchSubmitWorkflows(NonEmptyList.fromListUnsafe(workflowSourceFiles.toList)))
```
mcovarr: is this actually safe?

supportedBackends/jes/src/test/scala/cromwell/backend/impl/jes/JesJobExecutionActorSpec.scala
```diff
@@ -83,7 +84,14 @@ class JesJobExecutionActorSpec extends TestKitSuite("JesJobExecutionActorSpec")
     // Wait for the JABJEA to be spawned. Then kill it:
     parent.expectNoMsg(max = AwaitAlmostNothing)
     deathwatch.expectNoMsg(max = AwaitAlmostNothing)
-    jabjeaConstructionPromise.future foreach { _ ! JabjeaExplode }
+    jabjeaConstructionPromise.future onComplete {
+      case Success(jabjea) =>
+        jabjea ! JabjeaExplode
+      case Failure(throwable) =>
+        val exception = new RuntimeException("Error creating jabjea for test!", throwable)
+        exception.printStackTrace()
```
mcovarr: is this intentionally left in?

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
geoffjentry: Yeah, one of the upstream cases will catch it when it isn't

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
```diff
@@ -197,25 +197,39 @@ trait CromwellApiService {
       case b: BodyPart => b.toStrict(duration).map(strict => b.name -> strict.entity.data)
     }.runFold(Map.empty[String, ByteString])((map, tuple) => map + tuple)
 
+    def toResponse(workflowId: WorkflowId): WorkflowSubmitResponse = {
+      WorkflowSubmitResponse(workflowId.toString, WorkflowSubmitted.toString)
+    }
+
+    def askSubmit(command: WorkflowStoreActor.WorkflowStoreActorSubmitCommand): Route = {
+      // NOTE: Asks were already here. Guessing this was a KISS port of per-request via:
```
mcovarr: do ask, don't tell

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreSubmitActor.scala
Horneth: Yes same thing just differently - this looks better though + it's from @kshakir so it likely is. This will probably get merged first so I'll rebase.

geoffjentry: Ok :) I didn't know if perhaps this was coordinated or an example of parallel evolution.

supportedBackends/jes/src/test/scala/cromwell/backend/impl/jes/JesJobExecutionActorSpec.scala
kshakir: Yup. Should never error, but if/when it does, it won't just silently fail. I was eliminating all uses of `Future.foreach`.

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
kshakir: > that's how one does akka http

Can you provide a doc to link to that states this and I'll include it? The link I provided in the comment has an example that _doesn't_ use `ask`, after of course mentioning `ask` a "first way".

I'm not actually convinced that the Workflow Submission needs to be an Actor anymore, and the `ask` seems to be a retrofit to talk to an akka Actor wrapper around calls to scala Futures.

FYI- Clio hasn't matured to prod status, yet, but wasn't ever using the spray-esque per-request model. It only used scala Futures and (in-one-case) an akka-stream. Everything on first glance seemed much cleaner imho, but everything looks that way when there's not much code there. ;)

geoffjentry: Why comment it? That's literally what the [docs say](http://doc.akka.io/docs/akka-http/current/scala/http/routing-dsl/index.html), what their devs say, and blogs supporting the per-request pattern are just hacks on an old and overly cargo culted pattern. The per-request thing was not due to the fact that the underlying computation might be happening via actors, it was solving a different problem that almost no one (including us) really had. Not sure yow this is possible but apparently the Spray developers themselves were unaware of per-request, so it was only "spray-esque" in that a lot of people were copypasta'ing the pattern on their own.

Also, everything else in this file that's talking to an actor is using an ask. If you want to point out something as being a KISS artifact of our per-request past, they exist in this file (`metadataBuilderActor` I'm looking at you - was too hairy to decompose). So for instance I did look at unactoring the metadataBuilderActor because in that case a new one is being constructed for every call, but with all the FSM states in there I just didn't want to bother.

mcovarr:
How about tests that submitting pictures of Gumby now produces 4xx errors (and whatever else this fixes)?

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
kshakir: Hmm, do you have another link? I feel like that's only an example, as the section is titled "Longer Example", and not meant to announce that using `ask` is the new hotness. For example, in the same block of code, the actor being communicated with by the web route _is_ the database. I don't think flattening SOC is a new best practice either, and nowhere else on the page is `akka.pattern.ask` mentioned by the akka-http team.

I don't have a history of finding good links for scala via google, but here is an example of a "best practice" type blog post with discussion: [Don't use Actors for concurrency](https://www.chrisstucchio.com/blog/2013/actors_vs_futures.html)

I think we concur that the actor may be overkill. The code-comment is a heads up that one should NOT just copy-paste this code anymore without consideration.

If you have another link that says `ask` _should_ be used, and not _can_ be, I will use it.

I _do_ plan on returning someday to unwrap the actors off of these `someService.doFuture(request): Future[Response]` methods, but I unfortunately don't have the after-hours cycles for the next few months.

And you're 100% right that unwrapping an FSM is going to be a PITA. When I get to it, perhaps I'll push that FSM into an akka-stream `Graph` with composed sub-graphs... but I'll have to also cross that bridge later.

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreSubmitActor.scala
```diff
@@ -74,8 +95,10 @@ final case class WorkflowStoreSubmitActor(store: WorkflowStore, serviceRegistryA
   /**
     * Takes the workflow id and sends it over to the metadata service w/ default empty values for inputs/outputs
     */
-  private def registerSubmissionWithMetadataService(id: WorkflowId, originalSourceFiles: WorkflowSourceFilesCollection): Unit = {
-    processSource(_.clearEncryptedValues)(originalSourceFiles) foreach { sourceFiles =>
+  private def registerSubmissionWithMetadataService(
+      id: WorkflowId,
+      originalSourceFiles: WorkflowSourceFilesCollection): Future[Unit] = {
+    processSource(_.clearEncryptedValues)(originalSourceFiles) map { sourceFiles =>
```
Horneth: I'm not sure I see the advantage of a 
```
map {
...
()
}
```

compared to a 

```
foreach {
...
}
```

if we don't handle the failed case anyway

mcovarr: It looks like the caller handles it.

Horneth: oh I see 👍

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
geoffjentry: Chatted about this in slack

kshakir:
> How about Centaur tests that submitting pictures of Gumby now produces 4xx errors (and whatever else this fixes)?

I wish! Centaur only handles `200 OK` responses. This fix returns a `400 Bad Request`, quickly, instead of a timeout.

kshakir:
Comment edited. Squashed an rebased. No centaur changes for `400 Bad Request` accompanying this PR. I'll look at adding that to centaur next week unless something blows up.

gemlam3:
@geoffjentry have you given this one a review?

kshakir:
Got a verbal re-thumb. Merging.

PR #2501, opened 2017-07-27
Model objects for Cromwell Queries

cjllanwarne:
Model objects for converting query results to and from JSON.

This is not tied into the cromwell API client by this PR, instead I'm using it in CromIAM for filtering query results.


Horneth:
How can this be used in CromIAM ?

cjllanwarne:
@Horneth in CromIAM we already have access to the entity bytes, so the plan (at least initially) is to read that in as a Query result, find out which IDs can be seen by the requester, and then make a new entity out of the filtered results

PR #2503, opened 2017-07-28
Allow "AbortedResponse" even when not aborting

cjllanwarne:
Because sometimes things other than cromwell can cancel jobs.

Also might make restarts after aborts a little more resilient in case of unexpected race conditions (not a guarantee TM)

### DIFF
--- a/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
@@ -110,20 +110,29 @@ case class WorkflowExecutionActor(workflowDescriptor: EngineWorkflowDescriptor,
     case Event(BypassedDeclaration(declKey), stateData) =>
       handleDeclarationEvaluationSuccessful(declKey, WdlOptionalValue.none(declKey.scope.wdlType), stateData)
 
-      // Failure
-        // Initialization
+    // Failure
+    // Initialization
     case Event(JobInitializationFailed(jobKey, reason), stateData) =>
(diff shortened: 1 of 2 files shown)


engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -110,20 +110,29 @@ case class WorkflowExecutionActor(workflowDescriptor: EngineWorkflowDescriptor,
...
       pushFailedCallMetadata(jobKey, returnCode, reason, retryableFailure = false)
       handleNonRetryableFailure(stateData, jobKey, reason, Map.empty)
-      // Job Retryable
+    // Aborted? But we're outside of the AbortingState!?? Could happen if
+      // - The job was aborted by something external to Cromwell
+      // - The job lasted too long (eg JES 6 day timeout)
+      // - We've reconnected to an aborting job (some sort of shutdown race condition?)
+    // Treat it like any other non-retryable failure:
+    case Event(AbortedResponse(jobKey), stateData) =>
+      val cause = new Exception("The job was aborted from outside Cromwell")
```
Horneth: If it's the third case it's actually Cromwell that aborted it but ¯\_(ツ)_/¯

cjllanwarne: That was sort of speculative... hopefully it'll never happen

geoffjentry: I know you've already merged this but this could use a more explanatory message IMO. Now we'll see if I care enough to bother making a PR. I think I know the answer.

PR #2509, opened 2017-08-01
Client only modifies request json when inserting  a refresh token.

kshakir:


Horneth:
@kshakir if you haven't you might want to rebase on develop as I re-enabled restarts in centaur

PR #2512, opened 2017-08-03
Allow dotted dirs

kshakir:
Ensure GCS file systems use custom configuration.
When an exception/timeout occurs during asyncHashing, report it as a failure.

### DIFF
--- a/backend/src/main/scala/cromwell/backend/standard/callcaching/StandardFileHashingActor.scala
+++ b/backend/src/main/scala/cromwell/backend/standard/callcaching/StandardFileHashingActor.scala
@@ -1,5 +1,7 @@
 package cromwell.backend.standard.callcaching
 
+import java.util.concurrent.TimeoutException
+
 import akka.actor.{Actor, ActorLogging, ActorRef}
 import akka.event.LoggingAdapter
 import cromwell.backend.standard.StandardCachingActorHelper
@@ -52,7 +54,7 @@ object StandardFileHashingActor {
 
 abstract class StandardFileHashingActor(standardParams: StandardFileHashingActorParams) extends Actor with ActorLogging with JobLogging with IoClientHelper with StandardCachingActorHelper {
   this: IoCommandBuilder =>
-  override val ioActor = standardParams.ioActor
+  override lazy val ioActor = standardParams.ioActor
   override lazy val jobDescriptor: BackendJobDescriptor = standardParams.jobDescriptor
   override lazy val backendInitializationDataOption: Option[BackendInitializationData] = standardParams.backendInitializationDataOption
   override lazy val serviceRegistryActor: ActorRef = standardParams.serviceRegistryActor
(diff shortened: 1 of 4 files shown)


kshakir:
Centaur PR here: https://github.com/broadinstitute/centaur/pull/227

filesystems/gcs/src/main/scala/cromwell/filesystems/gcs/GcsPathBuilder.scala
```diff
@@ -126,28 +126,24 @@ object GcsPathBuilder {
...
+                     storageOptions: StorageOptions) extends PathBuilder {
+
+  private[gcs] val projectId = storageOptions.getProjectId
+
   def build(string: String): Try[GcsPath] = {
     Try {
       val uri = getUri(string)
       GcsPathBuilder.checkValid(uri)
-      GcsPath(provider.getPath(uri), apiStorage, cloudStorage)
+      val fileSystem = CloudStorageFileSystem.forBucket(uri.getHost, cloudStorageConfiguration, storageOptions)
```
Horneth: I'd still leave a comment here with something like
```
Note that this creates a new CloudStorageFileSystemProvider for every Path created, hence making it unsuitable for file copying using the nio copy method. Cromwell currently uses a lower level API which works around this problem. If you plan on using the nio copy method make sure to take this into consideration.
Also see https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1343
```

kshakir: Comment added to the method

.travis.yml
```diff
@@ -22,7 +22,7 @@ before_install:
   - sudo rm -f /etc/boto.cfg
 env:
   global:
-    - CENTAUR_BRANCH=develop
+    - CENTAUR_BRANCH=ks_allow_dotted_dirs
```

mcovarr: revert blah blah

backend/src/main/scala/cromwell/backend/standard/callcaching/StandardFileHashingActor.scala
```diff
@@ -63,12 +65,10 @@ abstract class StandardFileHashingActor(standardParams: StandardFileHashingActor
   def fileHashingReceive: Receive = {
     // Hash Request
     case fileRequest: SingleFileHashRequest =>
-      val replyTo = sender()
-
       customHashStrategy(fileRequest) match {
         case Some(Success(result)) => context.parent ! FileHashResponse(HashResult(fileRequest.hashKey, HashValue(result)))
         case Some(Failure(failure)) => context.parent ! HashingFailedMessage(fileRequest.file.value, failure)
-        case None => asyncHashing(fileRequest, replyTo)
+        case None => asyncHashing(fileRequest, context.parent)
```
mcovarr: pourquoi?

kshakir: Conversation with @Horneth says that this actor should be consistent, and always talk to the `context.parent`, not sometimes the `sender()`.

mcovarr: See, that's why I asked in French.

PR #2516, opened 2017-08-04
Client no longer always tries to inject refresh token.

kshakir:
Still provides a tested utility function for replacing json values.

### DIFF
--- a/cromwellApiClient/src/main/scala/cromwell/api/CromwellClient.scala
+++ b/cromwellApiClient/src/main/scala/cromwell/api/CromwellClient.scala
@@ -45,25 +45,6 @@ class CromwellClient(val cromwellUrl: URL, val apiVersion: String)(implicit acto
   import model.CromwellVersionJsonSupport._
   import model.CallCacheDiffJsonSupport._
 
-  private def requestEntityForSubmit(workflowSubmission: WorkflowSubmission) = {
-    import cromwell.api.model.LabelsJsonFormatter._
-
-    val sourceBodyParts = Map(
-      "workflowSource" -> Option(workflowSubmission.wdl),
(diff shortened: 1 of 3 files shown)


cjllanwarne:
👍 
ToL: I was thinking it'd be nice if the client could use a `case class` version of `WorkflowOptions` before using a JSON formatter to write the actual bytes onto the wire.

It'd make the centaur tests more fiddly to create bad JSON but make everyone else using the client have a much nicer time!

cromwellApiClient/src/main/scala/cromwell/api/CromwellClient.scala
```diff
@@ -155,19 +135,71 @@ object CromwellClient {
...
-      secretMap = optionsMap.updated(tokenKey, JsString(refreshTokenValue))
-      secretValue = secretMap.toJson.toString
-    } yield secretValue
-
-    secretOptions orElse options
+    val newJsonOption = for {
+      _ <- Option(json)
+      if json.contains(key)
+      map = json.parseJson.asJsObject.convertTo[Map[String, JsValue]]
+      if map.contains(key)
```
mcovarr: I'm missing something: why do we check both the json and the map for containment of the key?

kshakir: If one tries to to decode a bad json value, say `{`, the code would blow up on line 172 in the `parseJson`. We only want the code to parse the json iff the string contains the key.

Later on 173, we check if the map contains the key just in case someone has done something weird like `{"these_are_a_few_of_my_favorite_things": ["refresh_token"]}`.

kshakir:
Discussed ToL in person. A future revision might have the concept of a rich case class for WorkflowOptions. Merging.

PR #2520, opened 2017-08-04
Updated README on the application.conf

kshakir:


cjllanwarne:
Content seems good 👍 

ToL: If you don't, I might reorder this so that it goes from least-scary to most-scary, eg:

1. How do I create my own configuration file for Cromwell
  - A "hello world" example
2. Description of all the options I can put in my configuration file
3. What is reference.conf for and where is it?
4. What is application.conf for and where is it?

Maybe with 2/3 swapped?

geoffjentry:
👍

cjllanwarne:
Re 👍 for content. 

ToL: My only concern given the completeness of the example.conf is keeping it up to date (it'd almost be better if it was less complete?)

Issue #2507, opened 2017-07-31
Add automatic monitoring/instrumentation of underlying job


geoffjentry:
There have been many requests over time to provide data on CPU, memory, and disk usage. 

In an ideal world this would be a rich time series of data, in a less ideal world this would be max with a couple of periodic data points, and in a much more practical world it'd just be max usage. The last one is the definition of done here but if one wants to get ambitious .......

The PAPI backend already allows for a custom script to be attached, we should be putting in something on our own for any backend which runs things via unix command lines (ie spark jobs and such likely don't make sense)

At least one group is using the following in production, this is likely a good start if not completely AOK

``````#!/bin/bash
echo ==================================
echo =========== MONITORING ===========
echo ==================================
echo --- General Information ---
echo \#CPU: $(nproc)
echo Total Memory: $(free -h | grep Mem | awk '{ print $2 }')
echo Total Disk space: $(df -h | grep cromwell_root | awk '{ print $2}')
echo 
echo --- Runtime Information ---

function runtimeInfo() {
        echo [$(date)]
        echo \* CPU usage: $(top -bn 2 -d 0.01 | grep '^%Cpu' | tail -n 1 | awk '{print $2}')%
        echo \* Memory usage: $(free -m | grep Mem | awk '{ OFMT="%.0f"; print ($3/$2)*100; }')%
        echo \* Disk usage: $(df | grep cromwell_root | awk '{ print $5 }')
}

while true; do runtimeInfo; sleep 300; done```

Horneth:
ToL: Instead of pretty printing the values like in this script, it shouldn't be too hard to output them in some kind of tsv format with timestamps that would hopefully be easily parseable to a timeseries.

pshapiro4broad:
Not sure how hard it would be to do this, but another useful metric would be network usage (as represented by bytes going in/out at a moment in time). I haven't looked at this in cromwell workflow but at my last job we had a number of workflows that were bandwidth constrained (mainly due to network drive mounts) until we moved them to to SSD

LeeTL1220:
It would be nice to also have a user name attached to this if possible.  How much are *my* workflows using?

geoffjentry:
@LeeTL1220 Cromwell doesn't understand "user"

LeeTL1220:
Is therea way to give it a unique string?  Maybe a label.

geoffjentry:
[...]

PR #2529, opened 2017-08-08
Fixup for monitoring logs

ruchim:
- monitoring logs now delocalize
- publish monitoring script used as well as the monitoring logs to call-level metadata

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ before_install:
   - sudo rm -f /etc/boto.cfg
 env:
   global:
-    - CENTAUR_BRANCH=rm_monitoring
+    - CENTAUR_BRANCH=develop
     - INTEGRATION_TESTS_DIR=src/main/resources/integrationTestCases
   matrix:
     # Setting this variable twice will cause the 'script' section to run twice with the respective env var invoked
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ before_install:
   - sudo rm -f /etc/boto.cfg
 env:
   global:
-    - CENTAUR_BRANCH=develop
+    - CENTAUR_BRANCH=rm_monitoring
     - INTEGRATION_TESTS_DIR=src/main/resources/integrationTestCases
   matrix:
     # Setting this variable twice will cause the 'script' section to run twice with the respective env var invoked
--- a/supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
+++ b/supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
@@ -230,6 +230,20 @@ class JesAsyncBackendJobExecutionActor(override val standardParams: StandardAsyn
     )
   }
 
+  lazy val jesMonitoringParamName: String = JesJobPaths.JesMonitoringKey
+  lazy val localMonitoringLogPath: Path = DefaultPathBuilder.get(jesCallPaths.jesMonitoringLogFilename)
+  lazy val localMonitoringScriptPath: Path =  DefaultPathBuilder.get(jesCallPaths.jesMonitoringScriptFilename)
+
+  lazy val monitoringScript: Option[JesInput] = {
+    jesCallPaths.workflowPaths.monitoringScriptPath map { path =>
+      JesFileInput(s"$jesMonitoringParamName-in", path.pathAsString, localMonitoringScriptPath, workingDisk)
+    }
+  }
+
+  lazy val monitoringOutput: Option[JesFileOutput] = monitoringScript map { _ => JesFileOutput(s"$jesMonitoringParamName-out",
+    jesCallPaths.jesMonitoringLogPath.pathAsString, localMonitoringLogPath, workingDisk)
+  }
+
   override lazy val commandDirectory: Path = JesWorkingDisk.MountPoint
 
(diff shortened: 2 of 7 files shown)


supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesMetadataKeys.scala
```diff
@@ -5,8 +5,9 @@ object JesMetadataKeys {
   val GoogleProject = "jes:googleProject"
   val ExecutionBucket = "jes:executionBucket"
   val EndpointUrl = "jes:endpointUrl"
-  val MonitoringLog = "jes:monitoringLog"
+  val MonitoringScript = "jes:monitoringScript"
   val MachineType = "jes:machineType"
   val Zone = "jes:zone"
   val InstanceName = "jes:instanceName"
+  val MonitoringLog = "monitoringLog"
```
mcovarr: the log moves out from under `jes` too?

ruchim: I sort of equate the monitoring lot output to be kind of like a detritus output, but I'm happy to move it under the jes key.

.travis.yml
```diff
@@ -22,7 +22,7 @@ before_install:
   - sudo rm -f /etc/boto.cfg
 env:
   global:
-    - CENTAUR_BRANCH=develop
+    - CENTAUR_BRANCH=rm_monitoring
```
ruchim: revert!

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesMetadataKeys.scala
mcovarr: It doesn't look like we're making these logs on backends other than JES, are we?

.travis.yml
kshakir: Centaur JES is failing because it's expecting the script to write a `monitor.log` file. As there isn't one, the JES/PAPI delocalizer is crashing.

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
```diff
@@ -232,11 +232,14 @@ class JesAsyncBackendJobExecutionActor(override val standardParams: StandardAsyn
 
   override lazy val commandDirectory: Path = JesWorkingDisk.MountPoint
 
+  private val DockerMonitoringLogPath: Path = JesWorkingDisk.MountPoint.resolve(jesCallPaths.jesMonitoringLogFilename)
+  private val DockerMonitoringScriptPath: Path = JesWorkingDisk.MountPoint.resolve(jesCallPaths.jesMonitoringScriptFilename)
```
kshakir: What's the difference between this… (continues)

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesJobCachingActorHelper.scala
```diff
@@ -37,20 +31,24 @@ trait JesJobCachingActorHelper extends StandardCachingActorHelper {
   lazy val jesStdoutFile: Path = jesCallPaths.stdout
   lazy val jesStderrFile: Path = jesCallPaths.stderr
   lazy val jesLogFilename: String = jesCallPaths.jesLogFilename
-  lazy val defaultMonitoringOutputPath: Path = callRootPath.resolve(JesMonitoringLogFile)
+
+  lazy val jesMonitoringParamName: String = JesJobPaths.JesMonitoringKey
+  lazy val localMonitoringLogPath: Path = DefaultPathBuilder.get(jesCallPaths.jesMonitoringLogFilename)
+  lazy val localMonitoringScriptPath: Path =  DefaultPathBuilder.get(jesCallPaths.jesMonitoringScriptFilename)
```
kshakir: (continued) and this?

I don't know what the usage differences are, but one is building an absolute path, and the other is building a relative path.

The fact that `localMonitoringScriptPath` is only a relative path and not an absolute path is is causing the test `JesAsyncBackendJobExecutionActor should create a JesFileInput for the monitoring script, when specified` to fail in the `cromwell.backend.impl.jes.JesAsyncBackendJobExecutionActorSpec`.

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesMetadataKeys.scala
geoffjentry: Mentioned this offline but #2507 asks for this to be across all backends, but for now it's just JES

.travis.yml
ruchim: Yes -- I was passing in options through the inputs key in the Centaur test setup. Option was being ignored and hence no `monitoring.log` ever created.

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesJobCachingActorHelper.scala
kshakir: Explained IRL. The local paths should be built using the `DPB`, and the docker paths built using the `JWD.MP`. Using the `DPB` versus something like `relativeLocalizationPath()` means that `/cromwell_root` will not be duplicated in the ultimate path. 👍

PR #2531, opened 2017-08-08
Does not force GCS paths to have valid hostnames as buckets

Horneth:
Allows for `_` in bucket names


filesystems/gcs/src/main/scala/cromwell/filesystems/gcs/GcsPathBuilder.scala
```diff
@@ -20,65 +19,77 @@ import cromwell.filesystems.gcs.auth.GoogleAuthMode
...
-    checkNotNull(uri.getHost, s"%s does not have a host", uri)
-  }
+  // Provides some level of validation of GCS bucket names
+  // This is meant to alert the user early if they mistyped a gcs path in their workflow / inputs and not to validate
+  // exact bucket syntax, which is done by GCS.
+  // See https://cloud.google.com/storage/docs/naming for full spec
+  val GcsBucketPattern =
+    """
+      (?x)                                      # Turn on comments and whitespace insensitivity
+      gs://
```
mcovarr: does this need a `^` anchor on the front in case this pattern occurs internally?

filesystems/gcs/src/test/scala/cromwell/filesystems/gcs/GcsPathBuilderSpec.scala
```diff
@@ -345,10 +345,10 @@ class GcsPathBuilderSpec extends TestKitSuite with FlatSpecLike with Matchers wi
 
   private def badPaths = Seq(
     BadPath("an empty path", "", " does not have a gcs scheme"),
-    BadPath("an bucketless path", "gs://", "Expected authority at index 5: gs://"),
-    BadPath("a bucket named .", "gs://./hello/world", "gs://./hello/world does not have a host"),
+    BadPath("an bucketless path", "gs://", "The specified GCS path 'gs://' does not parse as a URI.\nExpected authority at index 5: gs://"),
+    BadPath("a bucket named .", "gs://./hello/world", "The path 'gs://./hello/world' does not seem to be a valid GCS path. Please check that it starts with gs:// and that the bucket and object follow GCS naming guidelines at https://cloud.google.com/storage/docs/naming."),
     BadPath("a non ascii bucket name", "gs://nonasciibucket£€/hello/world",
-      "gs://nonasciibucket%C2%A3%E2%82%AC/hello/world does not have a host"),
+      "The path 'gs://nonasciibucket£€/hello/world' does not seem to be a valid GCS path. Please check that it starts with gs:// and that the bucket and object follow GCS naming guidelines at https://cloud.google.com/storage/docs/naming."),
```
Horneth: So actually it might not be that simple - the `pathAsString` method calls `toUri` to get back a gcs path with the `gs://`, which fails because the bucket contains an underscore...
So we can create a valid CloudStorageFilesystem / CloudStoragePath but we can't get it back to a usable String using `toUri`...

core/src/test/scala/cromwell/core/path/PathBuilderSpecUtils.scala
```diff
@@ -87,16 +83,6 @@ trait PathBuilderSpecUtils {
         Option(path.getFileName).map(_.pathAsString).orNull should be(goodPath.getFileName)
       }
 
-    it should "match expected start of toUri" in
-      withClue(s"for path ${goodPath.path}${if (goodPath.normalize) " (normalized)" else ""}:") {
-        path.toUri.toString should startWith(goodPath.toUriStartsWith)
```
kshakir: Can you also comment out `NioPathMethods.toUri`, leaving a comment with a link to the existing Google NIO issue?

I don't want anyone tripping over this in the future.

PR #2539, opened 2017-08-11
Cromwell 29 release notes

katevoss:


CHANGELOG.md
```diff
@@ -1,25 +1,31 @@
...
-* The engine endpoints are now served under `/engine`. Previousely engine endpoints were available under
-`/api/engine`. Workflow endpoints are still served under `/api/workflows`. The setting `api.routeUnwrapped` has been
-retired at the same time.
+* **Metadata endpoint**  
+The response from the metadata endpoint can be quite large depending on your workflow. To help reduce the file size Cromwell can gzip encode the metadata prior to sending it back. To enable gzipped responses, send the `Accept-Encoding: gzip` header. By default the metadata endpoint does not gzip encode responses.  
 
-* The response format of the [callcaching/diff](https://github.com/broadinstitute/cromwell#get-apiworkflowsversioncallcachingdiff) endpoint has been updated.
+* **Engine endpoints**  
+Previously engine endpoints were available under
+`/api/engine` but now the endpoints are served under `/engine` so they don't require authentication. Workflow endpoints are still served under `/api/workflows`. We also deprecated the setting `api.routeUnwrapped` as a part of this internal consistency effort.
```
cjllanwarne: I'd remove the `"so they don't require authentication"` stuff since that's just an internal deployment detail

CHANGELOG.md
```diff
@@ -1,25 +1,31 @@
...
 
-### Cromwell Server
+* **Call caching diff**  
+We updated the response format of the [callcaching/diff](https://github.com/broadinstitute/cromwell#get-apiworkflowsversioncallcachingdiff) endpoint.
+
+### Other changes
+
+* **Cromwell Server**  
+When running in server mode, Cromwell now attempts to gracefully shutdown after receiving a `SIGINT` (`Ctrl-C`) or `SIGTERM` (`kill`) signal. This means that Cromwell waits for all pending Database writes before exiting. You can find a detailed explanation and information about how to configure this feature in the [Cromwell Wiki](https://github.com/broadinstitute/cromwell/wiki/DevZone#graceful-server-shutdown).
```
cjllanwarne: @kshakir and @Horneth can confirm but I think we want to remind/inform users that this is `"as long as you include application.conf at the top of your config file"`

CHANGELOG.md
```diff
@@ -633,5 +639,4 @@ types](https://github.com/docker/distribution/blob/05b0ab0/docs/spec/manifest-v2
 multiple input files.
 
 * The `/query` endpoint now supports querying by `id`, and submitting
-parameters as a HTTP POST.
-
+parameters as a HTTP POST.
```
cjllanwarne: Was this not ever thus?

CHANGELOG.md
```diff
@@ -1,25 +1,31 @@
...
-* Command line usage has been extensively revised for Cromwell 29.  Please see the
-[README](https://github.com/broadinstitute/cromwell#command-line-usage) for details.
+* **Request timeouts**   
+HTTP requests to the REST API returns a 503 status code instead of 500 in accordance with Akka HTTP. The response for a request timeout will now be plain text, rather than a JSON format.
 
-* The engine endpoints are now served under `/engine`. Previousely engine endpoints were available under
-`/api/engine`. Workflow endpoints are still served under `/api/workflows`. The setting `api.routeUnwrapped` has been
-retired at the same time.
+* **Metadata endpoint**  
+The response from the metadata endpoint can be quite large depending on your workflow. To help reduce the file size Cromwell can gzip encode the metadata prior to sending it back. To enable gzipped responses, send the `Accept-Encoding: gzip` header. By default the metadata endpoint does not gzip encode responses.
```
geoffjentry: @cjllanwarne is correct. This has been around for a while. The change in 29 is we're being more polite by not forcing it on you

CHANGELOG.md
geoffjentry: Another agree with @cjllanwarne (don't get a big head Chris). Until CaaS arrives the "why" is pretty meaningless.

CHANGELOG.md
```diff
@@ -1,25 +1,31 @@
...
-## 29
+## 29 Release Notes
 
 ### Breaking Changes
 
-* Request timeouts for HTTP requests on the REST API now return a 503 status code instead of 500. The response for a request timeout is no longer in JSON format.
-* The metadata endpoint no longer returns gzipped responses by default. This now needs to be explicitly requested with an `Accept-Encoding: gzip` header 
+* **Command line**  
+In preparation for supporting CWL scripts (yes, you read that right!), we have extensively revised the Command Line in Cromwell 29. For more details about the usage changes, please see the
+[README](https://github.com/broadinstitute/cromwell#command-line-usage). And stay tuned to the [WDL/Cromwell blog](https://software.broadinstitute.org/wdl/blog) over the next couple of months for more news about CWL.
```
mcovarr: "over the coming months" reads almost the same without quite the same degree of potentially regrettable specificity

CHANGELOG.md
```diff
@@ -1,25 +1,31 @@
...
 
-* Command line usage has been extensively revised for Cromwell 29.  Please see the
-[README](https://github.com/broadinstitute/cromwell#command-line-usage) for details.
+* **Request timeouts**   
+HTTP requests to the REST API returns a 503 status code instead of 500 in accordance with Akka HTTP. The response for a request timeout will now be plain text, rather than a JSON format.
```
mcovarr: Maybe we can just say that 503 is more correct than 500 without name-dropping Akka HTTP, since that's just an implementation detail.

CHANGELOG.md
mcovarr: We also deprecated the setting foo ... => The foo setting was also deprecated ...

CHANGELOG.md
mcovarr: Database => database

mcovarr: a detailed explanation and information => detailed information

CHANGELOG.md
mcovarr: This looks like just a blank line kill.

CHANGELOG.md
Horneth: Correct

CHANGELOG.md
katevoss: I'd rather include an implementation detail than just say "it's more correct"; as a reader I would wonder "*why* is it more correct?". 

On the flip side, do we have any reason to withhold the implementation detail?

CHANGELOG.md
katevoss: "By not forcing it on you" you mean that users now have to opt-in to gzipping their metadata file, right? I don't want to say that Cromwell no longer encodes the files in order to behave how an API technically should.

How about:
You can now opt-in to have Cromwell gzip encode your metadata file, in order to reduce file size, by sending the header.

CHANGELOG.md
katevoss: I understand it's an internal deployment detail but that's part of release notes is to tell the user why. If it doesn't have any user-facing impact (outside of room 7003) then it shouldn't go in the release notes, IMO.

The foo setting was deprecated... is the passive voice, but release notes (and all docs) should be active.

CHANGELOG.md
katevoss: It was a blank line kill.

CHANGELOG.md
katevoss: I have faith that it will be a couple of months 😃

CHANGELOG.md
cjllanwarne: The user facing impact is that they'll need to change what URL they point their `curl`s at.

The why is, because DSP prefers it that way for our own personal reasons.

CHANGELOG.md
cjllanwarne: Oh we've strayed out of the 29 changes here. Carry on then...

CHANGELOG.md
cjllanwarne: @mcovarr I find your lack of faith disturbing

CHANGELOG.md
cjllanwarne: IMO this isn't really an interface change. This changelog entry is essentially just mentioning a slightly more accurate error message. Eg:

- Cromwell now returns more specific `503 Service Unavailable` error codes on request timeouts, rather than the more generic `500 Internal Server Error`.

geoffjentry: I like Chris' example. 

I don't see any reason to explain implementation for stuff like this. We changed, the new one is better. The HTTP people say so :)

CHANGELOG.md
geoffjentry: Yeah, that's right. 

HTTP is intended to work as a negotiation between client and server. Client says "yo, i can accept the following formats" and server says "Well you're in luck, I can handle one of those". We were ignoring the first part.

CHANGELOG.md
mcovarr: mokay, then you may need to adjust the two sentences that precede it instead. 😛

geoffjentry: @cjllanwarne stole mine. The only reason it even mattered is because it'll matter for CromIAM and we want the API to be the same. 

So this is a case of "we changed the URLs, you should update" - there's no underlying reason that'd matter to a user.

cjllanwarne: Careful @geoffjentry, if you agree with me once more in this PR that's probably an omen of the end-times 😨

CHANGELOG.md
```diff
@@ -1,35 +1,39 @@
...
-* Command line usage has been extensively revised for Cromwell 29.  Please see the
-[README](https://github.com/broadinstitute/cromwell#command-line-usage) for details.
+* **Request timeouts**   
+Cromwell now returns more specific `503 Service Unavailable` error codes on request timeouts, rather than the more generic `500 Internal Server Error`. The response for a request timeout will now be plain text, rather than a JSON format.
 
-* The engine endpoints are now served under `/engine`. Previousely engine endpoints were available under
-`/api/engine`. Workflow endpoints are still served under `/api/workflows`. The setting `api.routeUnwrapped` has been
-retired at the same time.
+* **Metadata endpoint**  
+The response from the metadata endpoint can be quite large depending on your workflow. You can now opt-in to have Cromwell gzip your metadata file, in order to reduce file size, by sending the `Accept-Encoding: gzip` header. The default behavior now does not gzip encode responses.
```
geoffjentry: I know I previously said "yeah" to this wording but now that I read it it sounds like the gzip'ing was new altogether. 

Can't think of the right place to put it but somehow indicating that the previous case was always giving you gzip'd data

geoffjentry:
👍

PR #2542, opened 2017-08-11
Do not leak out snapshots of the WEA data to other actors

Horneth:
Thanks to @mcovarr we *think* we figured out the cause of the memory leak.
The `WorkflowExecutionActor` was sending "snapshots" of its current internal data (including execution store and output store) to the `EJEA`and `SWEA` so they could pass it on to the JobPreparationActor (resp. `SubWorkflowPreparationActor`) so they could evaluate inputs using the OutputStore.
This PR rewires things so that the WEA data does not escape the WEA. Instead actors needing the OutputStore for input evaluation request it at the right time and it gets sent across but never stored in the downstream actors, therefore not holding references and allowing for proper GC.

### DIFF
--- a/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
@@ -12,7 +12,9 @@ import cromwell.core.callcaching._
 import cromwell.core.logging.WorkflowLogging
 import cromwell.core.simpleton.WdlValueSimpleton
 import cromwell.database.sql.tables.CallCachingEntry
+import cromwell.engine.EngineWorkflowDescriptor
 import cromwell.engine.workflow.lifecycle.execution.EngineJobExecutionActor._
+import cromwell.engine.workflow.lifecycle.execution.WorkflowExecutionActor.RequestOutputStore
 import cromwell.engine.workflow.lifecycle.execution.callcaching.CallCache.CallCacheHashBundle
 import cromwell.engine.workflow.lifecycle.execution.callcaching.CallCacheReadActor._
 import cromwell.engine.workflow.lifecycle.execution.callcaching.CallCacheReadingJobActor.NextHit
@@ -35,7 +37,7 @@ import scala.util.{Failure, Success, Try}
 
 class EngineJobExecutionActor(replyTo: ActorRef,
                               jobDescriptorKey: BackendJobDescriptorKey,
-                              executionData: WorkflowExecutionActorData,
+                              workflowDescriptor: EngineWorkflowDescriptor,
                               factory: BackendLifecycleActorFactory,
                               initializationData: Option[BackendInitializationData],
                               restarting: Boolean,
@@ -50,8 +52,8 @@ class EngineJobExecutionActor(replyTo: ActorRef,
                               backendName: String,
                               callCachingMode: CallCachingMode) extends LoggingFSM[EngineJobExecutionActorState, EJEAData] with WorkflowLogging with CallMetadataHelper {
 
-  override val workflowIdForLogging = executionData.workflowDescriptor.id
-  override val workflowIdForCallMetadata = executionData.workflowDescriptor.id
+  override val workflowIdForLogging = workflowDescriptor.id
+  override val workflowIdForCallMetadata = workflowDescriptor.id
 
   override val supervisorStrategy = OneForOneStrategy() {
     // If an actor fails to initialize, send the exception to self before stopping it so we can fail the job properly
@@ -111,7 +113,7 @@ class EngineJobExecutionActor(replyTo: ActorRef,
         jobStoreActor ! QueryJobCompletion(jobStoreKey, jobDescriptorKey.call.task.outputs)
         goto(CheckingJobStore)
       } else {
-        prepareJob()
+        requestOutputStore()
       }
     case Event(JobExecutionTokenDenied(positionInQueue), NoData) =>
       log.debug("Token denied so cannot start yet. Currently position {} in the queue", positionInQueue)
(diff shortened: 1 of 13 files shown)


engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/preparation/JobPreparationActor.scala
```diff
@@ -78,66 +86,72 @@ class JobPreparationActor(executionData: WorkflowExecutionActorData,
...
   private[preparation] def scopedKey(key: String) = ScopedKey(workflowDescriptor.id, KvJobKey(jobKey), key)
-  private[preparation] def lookupKeyValueEntries(lookups: Seq[String]) = {
-    val keysToLookup: Seq[ScopedKey] = lookups map scopedKey
+  private[preparation] def lookupKeyValueEntries(inputs: Map[Declaration, WdlValue],
+                                                 attributes: Map[LocallyQualifiedName, WdlValue],
+                                                 maybeCallCachingEligible: MaybeCallCachingEligible) = {
+    val keysToLookup: Seq[ScopedKey] = kvStoreKeysToPrefetch map scopedKey
     keysToLookup foreach { serviceRegistryActor ! KvGet(_) }
-    goto(FetchingKeyValueStoreEntries) using JobPreparationKeyLookupData(PartialKeyValueLookups(Map.empty, keysToLookup))
+    // Hold on to the outputStore until we can use it to evaluate inputs
```
Horneth: That is totally not true, from an early version - remove this comment

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/SubWorkflowExecutionActor.scala
```diff
@@ -50,18 +50,28 @@ class SubWorkflowExecutionActor(key: SubWorkflowKey,
...
-      prepareSubWorkflow(createSubWorkflowId())
+      requestOutputStore(createSubWorkflowId())
+  }
+  
+  when(WaitingForOutputStore) {
+    case Event(outputStore: OutputStore, SubWorkflowExecutionActorData(Some(subWorkflowId))) =>
+      prepareSubWorkflow(subWorkflowId, outputStore)
+    case Event(outputStore: OutputStore, _) =>
+      jobLogger.error("This is a programmer error, we're ready to prepare the job and should have" +
+        " a SubWorkflowId to use by now but somehow haven't. Starting with a new WorkflowId." +
```
geoffjentry: Why is it safe to start this with a new wfid like this when this happens?

geoffjentry:
@Horneth Out of curiosity, do you have any real world numbers on the impact memory-wise? Is this only really going to be a big impact for huge WFs like in the JG case or should this also help out w/ installations like FC?

Horneth:
@geoffjentry We just talked about it but just to re-iterate I think the bigger the workflow (in terms of # of jobs) the bigger the impact, but it will still help with many small workflows.

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/SubWorkflowExecutionActorSpec.scala
```diff
@@ -116,26 +116,29 @@ class SubWorkflowExecutionActorSpec extends TestKitSuite with FlatSpecLike with
...
       ewea.stateData.subWorkflowId should not be empty
     }
   }
   
-  it should "Prepare a sub workflow" in {
+  it should "Request output store" in {
     val ewea = buildEWEA()
-    ewea.setState(SubWorkflowPendingState)
+    val subWorkflowId = WorkflowId.randomId()
+    ewea.setState(WaitingForOutputStore, SubWorkflowExecutionActorData(Some(subWorkflowId)))
```
mcovarr: `Option`

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/SubWorkflowExecutionActor.scala
Horneth: Yeah I don't know I didn't want to fail the workflow for this but maybe we should actually

geoffjentry: It seems like a really sketchy state to be in. OTOH if there's no possible way that it could go awry it'd also be a shame to fail the whole WF if not necessary.

Horneth: If we fail the workflow hopefully centaur will catch this and we'll know someone broke this. Otherwise it might stay undetected.

cjllanwarne: My $0.02: 
I would prefer to fail since if we do see these things show up we'd want to know and fix it asap.

cjllanwarne: Or, another way of looking at it: if it's fine to just throw it away and start with a new subworkflow ID, why do we assign it earlier in the first place?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/preparation/JobPreparationActor.scala
cjllanwarne: ... well go on then!

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/ejea/EjeaRequestingExecutionTokenSpec.scala
```diff
@@ -41,13 +42,13 @@ class EjeaRequestingExecutionTokenSpec extends EngineJobExecutionActorSpec with
...
 
-      s"bypass the Job Store and start preparing the job for running or call caching ($mode)" in {
+      s"bypass the Job Store and request output store to start preparing the job for running or call caching ($mode)" in {
         ejea = helper.buildEJEA(restarting = false)
         ejea ! JobExecutionTokenDispensed(helper.executionToken)
 
-        helper.jobPreparationProbe.expectMsg(max = awaitTimeout, hint = "Awaiting job preparation", CallPreparation.Start)
-        helper.jobStoreProbe.expectNoMsg(awaitAlmostNothing)
-        ejea.stateName should be(PreparingJob)
+        helper.replyToProbe.expectMsg(JobStarting(helper.jobDescriptorKey))
```
cjllanwarne: TOL: FWIW I still like the hints so you get a bit of context if it fails

engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/SubWorkflowExecutionActorSpec.scala
cjllanwarne: 😢

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -150,13 +152,23 @@ class EngineJobExecutionActor(replyTo: ActorRef,
...
-      prepareJob()
+      requestOutputStore()
     case Event(CacheResultLookupFailure(reason), NoData) =>
       log.error(reason, "{}: Failure checking for cache entry existence: {}. Attempting to resume job anyway.", jobTag, reason.getMessage)
-      prepareJob()
+      requestOutputStore()
+  }
+
+  /*
+  * ! Hot Potato Warning !
```
mcovarr: I literally asked for that, didn't I?

PR #2543, opened 2017-08-12
Log configured batch size and max flush intervals.

mcovarr:
Note that there is already `debug` level logging for the actual flush events, which might be a good idea to turn on to see if the settings we've chosen are reasonable or whether we're just flushing constantly and blowing up the Slick queue.

### DIFF
--- a/core/src/main/scala/cromwell/core/actor/BatchingDbWriter.scala
+++ b/core/src/main/scala/cromwell/core/actor/BatchingDbWriter.scala
@@ -69,9 +69,13 @@ trait BatchingDbWriterActor { this: FSM[BatchingDbWriterState, BatchingDbWriterD
   
   def isShuttingDown: Boolean = shuttingDown
   def dbFlushRate: FiniteDuration
+  def dbBatchSize: Int
+  def writeActorName: String
+
   var periodicFlush: Option[Cancellable] = None
 
   override def preStart(): Unit = {
+    log.info("{} configured to write to the database with batch size {} and flush rate {}.", writeActorName, dbBatchSize, dbFlushRate)
     periodicFlush = Option(context.system.scheduler.schedule(0.seconds, dbFlushRate, self, ScheduledFlushToDb)(context.dispatcher))
   }
 
--- a/core/src/main/scala/cromwell/core/actor/BatchingDbWriter.scala
+++ b/core/src/main/scala/cromwell/core/actor/BatchingDbWriter.scala
@@ -70,12 +70,11 @@ trait BatchingDbWriterActor { this: FSM[BatchingDbWriterState, BatchingDbWriterD
   def isShuttingDown: Boolean = shuttingDown
   def dbFlushRate: FiniteDuration
   def dbBatchSize: Int
-  def writeActorName: String
 
   var periodicFlush: Option[Cancellable] = None
 
   override def preStart(): Unit = {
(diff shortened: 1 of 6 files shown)


engine/src/main/scala/cromwell/jobstore/JobStoreActor.scala
```diff
@@ -18,6 +18,7 @@ import scala.language.postfixOps
   */
 class JobStoreActor(jobStore: JobStore, dbBatchSize: Int, dbFlushRate: FiniteDuration) extends Actor with ActorLogging with GracefulShutdownHelper {
   import JobStoreActor._
+  log.info(s"JobStoreActor configured to write with batch size {} and flush rate {}.", dbBatchSize, dbFlushRate)
```
mcovarr: hmm should probably move this and the one below to the actual write actors... ☹️

Horneth:
👍 

ToL: Since they all extend the `BatchingDbWriterActor` trait we could even put this logging there instead.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheWriteActor.scala
```diff
@@ -19,6 +19,7 @@ case class CallCacheWriteActor(callCache: CallCache) extends LoggingFSM[Batching
   implicit val ec: ExecutionContext = context.dispatcher
   
   override val dbFlushRate = CallCacheWriteActor.dbFlushRate
+  log.info("CallCacheWriteActor configured to write to the database with batch size {} and flush rate {}.", dbBatchSize, dbFlushRate)
```
geoffjentry: AFAICT these are hardcoded values?

engine/src/main/scala/cromwell/jobstore/JobStoreWriterActor.scala
```diff
@@ -16,6 +16,8 @@ case class JobStoreWriterActor(jsd: JobStore, dbBatchSize: Int, override val dbF
 
   implicit val ec = context.dispatcher
 
+  log.info(s"JobStoreWriterActor configured to write to the database with batch size {} and flush rate {}.", dbBatchSize, dbFlushRate)
```
geoffjentry: I think these are also hardcoded

mcovarr: You are correct sir.  But should the observed logging of flushes inspire anyone to knobify these, the log messages are knob-ready.

core/src/main/scala/cromwell/core/actor/BatchingDbWriter.scala
```diff
@@ -69,9 +69,13 @@ trait BatchingDbWriterActor { this: FSM[BatchingDbWriterState, BatchingDbWriterD
   
   def isShuttingDown: Boolean = shuttingDown
   def dbFlushRate: FiniteDuration
+  def dbBatchSize: Int
+  def writeActorName: String
```
Horneth: ToL: I think we could just use `self.path.name` which should be the name of the actor. And if it isn't we should set it so when we create it as it shows up in the log and is good practice IMO. Again ToL :)

mcovarr: Cool, I'll try that.  I originally tried `this.getClass.getSimpleName` which did not end well due to trait name mangling, and I wasn't really happy with this solution either but at least it doesn't explode.

PR #2546, opened 2017-08-14
Allow for null input values

Horneth:
- [x] Add to the changelog

- [x] Document that empty wdl values now show as `null` json instead of `"null"`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ before_install:
   - sudo rm -f /etc/boto.cfg
 env:
   global:
-    - CENTAUR_BRANCH=nullInput
+    - CENTAUR_BRANCH=develop
     - INTEGRATION_TESTS_DIR=src/main/resources/integrationTestCases
   matrix:
     # Setting this variable twice will cause the 'script' section to run twice with the respective env var invoked
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ before_install:
   - sudo rm -f /etc/boto.cfg
 env:
   global:
-    - CENTAUR_BRANCH=develop
+    - CENTAUR_BRANCH=nullInput
     - INTEGRATION_TESTS_DIR=src/main/resources/integrationTestCases
   matrix:
     # Setting this variable twice will cause the 'script' section to run twice with the respective env var invoked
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -22,7 +22,7 @@ before_install:
   - sudo rm -f /etc/boto.cfg
 env:
   global:
-    - CENTAUR_BRANCH=develop
+    - CENTAUR_BRANCH=nullInput
```

engine/src/main/scala/cromwell/webservice/metadata/MetadataComponent.scala
```diff
@@ -120,8 +121,8 @@ object MetadataComponent {
 }
 
 sealed trait MetadataComponent
-case object MetadataEmpty extends MetadataComponent
-case object MetadataNull extends MetadataComponent
+case object MetadataEmptyComponenet extends MetadataComponent
```
mcovarr: MetadataEmptyComponenet => MetadataEmptyComponent

cjllanwarne:
Did you consider a migration at all?

Horneth:
@cjllanwarne about the migration that's a good point. I think we're going to want to add that and warn firecloud of the change.
Also looks like we're already using `"null"` in centaur in multiple tests so I'll have to change that

geoffjentry:
@Horneth @cjllanwarne we need to be less cavalier about changes which require migrations. If this requires a migration it should be verified with that in mind w/ our large production-y friends. We can't just force them to take such a change anymore.

Horneth:
@geoffjentry who decides whether or not this is migration worthy ?

cjllanwarne:
@geoffjentry I think I know but just to check, which is the thing you're worried by:
- the schema for outputs changing
- the cost of doing a migration

Because we *could* not migrate old `"null"`s but that wouldn't change the fact that the schema has changed in this PR

geoffjentry:
@cjllanwarne Initially the concern was that I've heard, well, concern from firecloud folks here and there about frivolous metadata migrations on our part as they cause downtime they don't always want to incur. It also ties in to my rapidly increasing sentiment that we should not be migrating a supposedly immutable data store for what are merely cosmetic changes - if we have to in order to fix something we know is causing issues w/ programatic clients you gotta do what you gotta do.

When we talked with @abaumann he indicated that a) they wouldn't really care if the old data stayed as-is and b) made a similar point as I did about the event store.

@Horneth with all of these taken together I don't think we should be migrating. The key constituency has said that it would be at best superfluous and IMO it's part of a habit we should be getting out of anyways.

PR #2551, opened 2017-08-15
skip wdl subproject when publishing wdl4s

danbills:


release/release_workflow.wdl
```diff
@@ -40,7 +40,7 @@ task do_release {
        # Generate new scaladoc
        sbt 'set scalacOptions in (Compile, doc) := List("-skip-packages", "better")' doc
        git checkout gh-pages
-       for subproj in cwl wdl wom; do
+       for subproj in cwl wom; do
```
mcovarr: Not sure I would fix it this way, we will have scaladocs for WDL once WDL and WOM are separated.  I would make sure the directory exists instead.

danbills: yep that's better

mcovarr: ```test -d $subproj/target/scala-2.12/api || mkdir -p $subproj/target/scala-2.12/api```

PR #2554, opened 2017-08-16
Limit the batch request size

Horneth:
This is a proposal on how to solve the "insufficient data written" error that is due to the batch request doing PAPI job creation / polling being too large. The actual limit number is unknown yet (ticket is open with google).
Depending on their answer the solution might look different but this would be one way to fix it.
I tested it with @ruchim's WDL that reproduces this problem and didn't see it.

### DIFF
--- a/supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesApiQueryManager.scala
+++ b/supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesApiQueryManager.scala
@@ -5,11 +5,11 @@ import java.io.IOException
 import akka.actor.{Actor, ActorLogging, ActorRef, Props, Terminated}
 import cats.data.NonEmptyList
 import com.google.api.client.googleapis.json.GoogleJsonError
-import com.google.api.client.http.HttpHeaders
+import com.google.api.client.http.{HttpHeaders, HttpRequest}
 import com.google.api.services.genomics.Genomics
 import com.google.api.services.genomics.model.RunPipelineRequest
 import cromwell.backend.impl.jes.Run
-import cromwell.backend.impl.jes.statuspolling.JesApiQueryManager._
+import cromwell.backend.impl.jes.statuspolling.JesApiQueryManager.{JesApiException, _}
 import cromwell.core.CromwellFatalExceptionMarker
 import cromwell.core.Dispatcher.BackendDispatcher
 import cromwell.core.retry.{Backoff, SimpleExponentialBackoff}
@@ -17,6 +17,7 @@ import cromwell.util.StopAndLogSupervisor
 import eu.timepit.refined.api.Refined
 import eu.timepit.refined.numeric._
 
+import scala.annotation.tailrec
 import scala.concurrent.duration._
 import scala.collection.immutable.Queue
 
@@ -26,7 +27,14 @@ import scala.collection.immutable.Queue
 class JesApiQueryManager(val qps: Int Refined Positive) extends Actor with ActorLogging with StopAndLogSupervisor {
 
   private implicit val ec = context.dispatcher
-  private val maxRetries = 10
+  private val maxRetries = 0
+  // FIXME: Determined empirically that we start to get errors when the batch request approaches 15MB.
+  // Find out what the real value is
+  private val maxBatchRequestSize = 14 * 1024 * 1024
+  private val requestTooLargeException = new JesApiException(new IllegalArgumentException(
+    "The JES creation request exceeds the maximum size allowed. " +
+      "If you have a task with a very large number of inputs and / or outputs in your workflow you should try to reduce it."
+  ))
 
   // workQueue is protected for the unit tests, not intended to be generally overridden
   protected[statuspolling] var workQueue: Queue[JesApiQuery] = Queue.empty
@@ -42,7 +50,11 @@ class JesApiQueryManager(val qps: Int Refined Positive) extends Actor with Actor
   override def receive = {
     case DoPoll(run) => workQueue :+= JesStatusPollQuery(sender, run)
     case DoCreateRun(genomics, rpr) =>
-      workQueue :+= JesRunCreationQuery(sender, genomics, rpr)
+      val creationQuery = JesRunCreationQuery(sender(), genomics, rpr)
+
+      if (creationQuery.contentLength > maxBatchRequestSize) {
+        creationQuery.requester ! JesApiRunCreationQueryFailed(creationQuery, requestTooLargeException)
+      } else workQueue :+= creationQuery
     case q: JesApiQuery => workQueue :+= q
     case RequestJesPollingWork(maxBatchSize) =>
       log.debug("Request for JES Polling Work received (max batch: {}, current queue size is {})", maxBatchSize, workQueue.size)
@@ -82,8 +94,27 @@ class JesApiQueryManager(val qps: Int Refined Positive) extends Actor with Actor
   private case class BeheadedWorkQueue(workToDo: Option[NonEmptyList[JesApiQuery]], newWorkQueue: Queue[JesApiQuery])
   private def beheadWorkQueue(maxBatchSize: Int): BeheadedWorkQueue = {
 
-    val head = workQueue.take(maxBatchSize).toList
-    val tail = workQueue.drop(maxBatchSize)
+    /**
+      * Take the head of the queue, making sure it stays under maxBatchSize as well as maxBatchRequestSize.
+      * Assumes that each query in the queue can fit in an empty batch (queries with a size > maxBatchSize should no be added to the queue)
+      *
+      * This is a naive approach where we keep adding queries to the head as long as they don't overflow it, and then we stop.
+      * This could leave us with half empty batches and is not very efficient.
+      *
+      * A somewhat less naive approach would be to keep looking in the queue for another query that would fit.
+      *
+      * More generally this is the knapsack problem (https://en.wikipedia.org/wiki/Knapsack_problem) which is NP-Complete.
+      * If this needs further optimization, finding an approximation algorithm that fits well this case would be a place to start.
+      */
+    @tailrec
+    def behead(queue: Queue[JesApiQuery], head: Vector[JesApiQuery]): Vector[JesApiQuery]  = queue.headOption match {
+      case Some(query) if head.size < maxBatchSize && (head :+ query).map(_.contentLength).sum < maxBatchRequestSize =>
+        behead(queue.tail, head :+ query)
+      case _ => head
+    }
+
+    val head = behead(workQueue, Vector.empty).toList
+    val tail = workQueue.drop(head.size)
 
     head match {
       case h :: t => BeheadedWorkQueue(Option(NonEmptyList(h, t)), tail)
@@ -147,6 +178,8 @@ object JesApiQueryManager {
     def genomicsInterface: Genomics
     def withFailedAttempt: JesApiQuery
     def backoff: Backoff
+    def httpRequest: HttpRequest
+    def contentLength = httpRequest.getContent.getLength
   }
   private object JesApiQuery {
     // This must be a def, we want a new one each time (they're mutable! Boo!)
@@ -154,10 +187,12 @@ object JesApiQueryManager {
   }
   private[statuspolling] final case class JesStatusPollQuery(requester: ActorRef, run: Run, failedAttempts: Int = 0, backoff: Backoff = JesApiQuery.backoff) extends JesApiQuery {
     override val genomicsInterface = run.genomicsInterface
+    override val httpRequest = run.getOperationCommand.buildHttpRequest()
     override def withFailedAttempt = this.copy(failedAttempts = failedAttempts + 1, backoff = backoff.next)
   }
   private[statuspolling] final case class JesRunCreationQuery(requester: ActorRef, genomicsInterface: Genomics, rpr: RunPipelineRequest, failedAttempts: Int = 0, backoff: Backoff = JesApiQuery.backoff) extends JesApiQuery {
     override def withFailedAttempt = this.copy(failedAttempts = failedAttempts + 1, backoff = backoff.next)
+    override val httpRequest = genomicsInterface.pipelines().run(rpr).buildHttpRequest()
   }
 
   trait JesApiQueryFailed {
--- a/supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesApiQueryManager.scala
+++ b/supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesApiQueryManager.scala
@@ -27,13 +27,15 @@ import scala.collection.immutable.Queue
 class JesApiQueryManager(val qps: Int Refined Positive) extends Actor with ActorLogging with StopAndLogSupervisor {
 
   private implicit val ec = context.dispatcher
-  private val maxRetries = 0
+  private val maxRetries = 10
   // FIXME: Determined empirically that we start to get errors when the batch request approaches 15MB.
   // Find out what the real value is
   private val maxBatchRequestSize = 14 * 1024 * 1024
   private val requestTooLargeException = new JesApiException(new IllegalArgumentException(
-    "The JES creation request exceeds the maximum size allowed. " +
-      "If you have a task with a very large number of inputs and / or outputs in your workflow you should try to reduce it."
+    "The task run request has exceeded the maximum PAPI request size." +
+      "If you have a task with a very large number of inputs and / or outputs in your workflow you should try to reduce it. " +
+      "Depending on your case you could: 1) Zip your input files together and unzip them in the command. 2) Use a file of file names " +
+      "and localize the files yourself."
   ))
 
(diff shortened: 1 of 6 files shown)


backend/src/main/scala/cromwell/backend/standard/callcaching/StandardFileHashingActor.scala
```diff
@@ -91,7 +93,7 @@ abstract class StandardFileHashingActor(standardParams: StandardFileHashingActor
     }
 
     ioHashCommandTry match {
-      case Success(ioHashCommand) => sendIoCommandWithContext(ioHashCommand, fileRequest)
+      case Success(ioHashCommand) => sendIoCommandWithContext(ioHashCommand, FileHashContext(fileRequest.hashKey, fileRequest.file.value))
```
Horneth: This is unrelated, the reason is while trying to see if this fixed the problem I noticed Cromwell was spending a ridiculous amount of time trying to compute the hashcode of the `fileRequest` as it's being inserted into an hashmap. This reduces the hashing to only what's needed.

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesApiQueryManager.scala
```diff
@@ -27,6 +28,13 @@ class JesApiQueryManager(val qps: Int Refined Positive) extends Actor with Actor
 
   private implicit val ec = context.dispatcher
   private val maxRetries = 10
+  // FIXME: Determined empirically that we start to get errors when the batch request approaches 15MB.
+  // Find out what the real value is
+  private val maxBatchRequestSize = 14 * 1024 * 1024
+  private val requestTooLargeException = new JesApiException(new IllegalArgumentException(
+    "The JES creation request exceeds the maximum size allowed. " +
```
cjllanwarne: - Should be `PAPI`.
- I would be more specific, eg "The task run request has exceeded the maximum PAPI request size."

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesApiQueryManager.scala
```diff
@@ -27,6 +28,13 @@ class JesApiQueryManager(val qps: Int Refined Positive) extends Actor with Actor
...
+      "If you have a task with a very large number of inputs and / or outputs in your workflow you should try to reduce it."
```
cjllanwarne: Can we give some concrete suggestions here? No nice answers but maybe this?
> Try bundling your inputs into a zip file rather than submitting individually

cjllanwarne: Of course when we have more backends we could also suggest switching cloud providers...

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesApiQueryManager.scala
```diff
@@ -40,9 +48,13 @@ class JesApiQueryManager(val qps: Int Refined Positive) extends Actor with Actor
   resetWorker()
 
   override def receive = {
-    case DoPoll(run) => workQueue :+= JesStatusPollQuery(sender, run)
+    case DoPoll(run) => workQueue :+= makePollQuery(sender(), run)
```
cjllanwarne: I don't think `sender` has side effects?

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesApiQueryManager.scala
```diff
@@ -147,17 +187,25 @@ object JesApiQueryManager {
...
+    def httpRequest: HttpRequest
+    def contentLength: Long = Option(httpRequest.getContent).map(_.getLength).getOrElse(0L)
   }
   private object JesApiQuery {
     // This must be a def, we want a new one each time (they're mutable! Boo!)
     def backoff: Backoff = SimpleExponentialBackoff(1.second, 1000.seconds, 1.5d)
   }
-  private[statuspolling] final case class JesStatusPollQuery(requester: ActorRef, run: Run, failedAttempts: Int = 0, backoff: Backoff = JesApiQuery.backoff) extends JesApiQuery {
+
+  // Not final so methods can be overridden in unit tests.
```
cjllanwarne: ToL: You could also achieve this with implicit extension methods

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesApiQueryManager.scala
```diff
@@ -82,8 +102,28 @@ class JesApiQueryManager(val qps: Int Refined Positive) extends Actor with Actor
   private case class BeheadedWorkQueue(workToDo: Option[NonEmptyList[JesApiQuery]], newWorkQueue: Queue[JesApiQuery])
   private def beheadWorkQueue(maxBatchSize: Int): BeheadedWorkQueue = {
 
-    val head = workQueue.take(maxBatchSize).toList
-    val tail = workQueue.drop(maxBatchSize)
+    /*
+      * Take the head of the queue, making sure it stays under maxBatchSize as well as maxBatchRequestSize.
```
cjllanwarne: Should this be:
> Keep taking from the head of the queue

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesApiQueryManager.scala
```diff
@@ -82,8 +102,28 @@ class JesApiQueryManager(val qps: Int Refined Positive) extends Actor with Actor
...
+      */
+    @tailrec
+    def behead(queue: Queue[JesApiQuery], head: Vector[JesApiQuery]): Vector[JesApiQuery]  = queue.headOption match {
+      // TODO: don't recompute the length sum every time ?
+      case Some(query) if head.size < maxBatchSize && head.map(_.contentLength).sum + query.contentLength < maxBatchRequestSize =>
+        behead(queue.tail, head :+ query)
+      case _ => head
+    }
+
+    val head = behead(workQueue, Vector.empty).toList
```
cjllanwarne: ToL: The first time `behead` is called, the `head` is empty, and if the first element is > 15MB this queue will be stuck forever.

I know we *should* be safe against that based on how the queue is built, but it'd be nice to have some "get out of jail" logic too?

cjllanwarne: (e.g. start with `behead(workQueue.tail, Vector(workQueue.head))` instead of `behead(workQueue, Vector.empty)`?

backend/src/main/scala/cromwell/backend/standard/callcaching/StandardFileHashingActor.scala
mcovarr: Should this be broken into its own PR as it sounds like something we'd definitely want?

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesApiQueryManager.scala
mcovarr: oh snap

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesApiQueryManager.scala
mcovarr: Though it can only be called in the context of a partial function from Any to Unit.  Maybe it's just a penance for that overarching sin.

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesApiQueryManager.scala
```diff
@@ -82,8 +102,28 @@ class JesApiQueryManager(val qps: Int Refined Positive) extends Actor with Actor
...
+      * Take the head of the queue, making sure it stays under maxBatchSize as well as maxBatchRequestSize.
+      * Assumes that each query in the queue can fit in an empty batch (queries with a size > maxBatchSize should no be added to the queue)
+      *
+      * This is a naive approach where we keep adding queries to the head as long as they don't overflow it, and then we stop.
+      * This could leave us with half empty batches and is not very efficient.
+      *
+      * A somewhat less naive approach would be to keep looking in the queue for another query that would fit.
+      *
+      * More generally this is the knapsack problem (https://en.wikipedia.org/wiki/Knapsack_problem) which is NP-Complete.
+      * If this needs further optimization, finding an approximation algorithm that fits well this case would be a place to start.
```
mcovarr: TIL

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/StatusPolling.scala
```diff
@@ -39,15 +39,21 @@ private[statuspolling] trait StatusPolling { this: JesPollingActor =>
...
 
-  def addStatusPollToBatch(run: Run, batch: BatchRequest, resultHandler: JsonBatchCallback[Operation]) =
-    run.getOperationCommand.queue(batch, resultHandler)
+  def addStatusPollToBatch(request: HttpRequest, batch: BatchRequest, resultHandler: JsonBatchCallback[Operation]): Unit = {
+    /*
+      * Manually enqueue the request instead of doing it through the RunPipelineRequest
+      * as it would unnecessarily rebuild the request (which we already have)
+    */
+    batch.queue(request, classOf[Operation], classOf[GoogleJsonErrorContainer], resultHandler)
+    ()
```
mcovarr: is there a reason for the `: Unit` / `()` pendaticism here and not above?

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesApiQueryManager.scala
Horneth: Yeah, TBH the reason I had the actor reject the request if it was > 15MB was to avoid having to deal with this here in the first place but I guess we can do both.

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesApiQueryManager.scala
geoffjentry: I don't think this solution is going to stand firm in the face of battle. This seems like something which would turn into a fire the first time a prominent user runs into it

geoffjentry: Is this a "there's no possible way to make this happen" issue or a "solving that would be really messy for us to do" issue?

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/StatusPolling.scala
Horneth: above does not have a type ascription that's why :)
I'll add one

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesApiQueryManager.scala
mcovarr: > Try bundling your inputs into a zip file rather than submitting individually

Oh I like this, if users have a problem Cromwell just tells them to zip it.

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesApiQueryManager.scala
Horneth: How do I do that ?

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesApiQueryManager.scala
Horneth: @geoffjentry If they truly have a task that generates a request > 15 MB

1. They have a ridiculous amount of inputs (that would probably take forever to JES to localize anyway).
As a benchmark they would need approximately 7 million input files looking like 'gs://cloud-cromwell-dev/mock/mock_firecloud_workspace_folder/mock_firecloud_submission_folder/mock_cromwell_workflow_name/mock_workflow_uuid/mock-call-alias-directory/file1.txt' to go over the limit.

I would say that there must be a better solution to what they're trying to do.

2. If we still want to try something we could try to submit it to JES directly instead of batching it but even if not the same they must have a limit too.

geoffjentry: Ok, the 7 million mark changes my mind :)

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesApiQueryManager.scala
cjllanwarne: Hmm... You'd want some class that can implement these methods instead of having them directly in the case class, and then in your test, replace the instance of that class with the test version of the methods.

But as I'm thinking about this, it's way more complicated that just allowing the methods to be overridden, so I'd probably leave it as it is, tbh...

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesApiQueryManager.scala
Horneth: Wait no -_- I think messed up this one

Horneth: It's actually 32K

geoffjentry: I feel like that's in the zone of "there's probably a better way to do what you're doing" but that's also likely to wind up in the "this is a fire! fix it!" zone as well. Even if the ultimate answer is "change what you're doing", keep in mind that we'll need to context switch to taking care of them when determining risk/reward of leaving it this way

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesApiQueryManager.scala
```diff
@@ -27,6 +28,15 @@ class JesApiQueryManager(val qps: Int Refined Positive) extends Actor with Actor
 
   private implicit val ec = context.dispatcher
   private val maxRetries = 10
+  // FIXME: Determined empirically that we start to get errors when the batch request approaches 15MB.
```
geoffjentry: Can you unfixme this? Likely just a slight rewording of the comment

geoffjentry: Hang on - just sent some stuff over to you on on slack. Starting to think that this is all correlation and not causation

geoffjentry: Ok, now that we have more clarity, can you change this from a FIXME and give a little more context on what we found out so we remember the story if/when we come back around to this? Maybe a pointer to that one github link w/ the supposed fix and an explanation that this is in the "appears to work" category + your "it's actually length of time" theory?

Horneth: @geoffjentry Sure, although if it really is length of time then it will only work for people with a connection as good or better than mine.

Horneth: By mine I mean my laptop's on the Broad network

geoffjentry: Good point. The lowest number I spotted was 10MB and the highest was up around 40, so I suspect we're in the right ballpark for a typical size here

Horneth: What would be interesting is to monitor the size (both as in number of requests and total byte size) per JES batch. It would give us an idea of the limiting factor.
Another way to do this would be to not limit the byte size of the request but simply cut it in half if it fails until it succeeds...
Also it looks like the limit for number of queries is actually 1000 even though we limit ourselves to 100 https://cloud.google.com/genomics/batch

Horneth: Anyway are we trying like this ?

PR #2558, opened 2017-08-17
point latest wdl4s scaladoc symlink to correct path

danbills:
fixes broken symlink per subprojecting on wdl4s


cjllanwarne:
If you say so... 👍

Issue #1149, opened 2016-07-12
Untag "PostMVP" tests so we can collect metrics on how many ignored, etc.


kcibul:
This will be a precursor to then having tickets that un-ignore those tests (either by fixing or removing)

katevoss:
@geoffjentry is this still a useful ticket? Are there still PostMVP tests or have they been cleaned up in the last year?

geoffjentry:
@katevoss I've said that it is not useful for quite some time now. Every time I bring it up other people disagree with removing it.

Yes, there are still ignored tests but I'm using the model of "if you haven't touched it by now it's not useful and you're never going to come back to it". 

I'd suggest we set a TTL for this ticket and anyone who feels strongly about these tests can effort to enable them between now and then.

katevoss:
It sounds like this ticket does not hold much value any longer.

@kcibul with your okay, I will close it. If you disagree, please explain more about what needs to be done and what value it provides.

mcovarr:
If we don't care about those tests then IMHO this ticket should become "delete all that PostMVP stuff you don't care about".

geoffjentry:
My long held stance is that if anyone cared about those tests they'd have unignored them by now. 

Imo the options are "someone who cares fix this immediately" or what @mcovarr said

mcovarr:
Done.

PR #2566, opened 2017-08-21
post post MVP

mcovarr:
What it would look like if #1149 were satisfied.  We do have real restart testing now, and the rest is all call caching tests; I'm looking over what we for that over in Centaur-land.


geoffjentry:
👍 make it happen

cjllanwarne:
👍 for this PR.

Are there any remaining ignored tests that we still need to be looked at?

mcovarr:
@cjllanwarne there was one `ignore`d test about default runtime attributes that pretty clearly seemed to be covered by newer tests, so I've deleted that as well.  There's one other abort test we should definitely continue to feel bad about, and another test for a "taskless workflow" for which I couldn't readily find an equivalent.

PR #2572, opened 2017-08-23
Separation of database access into engine and metadata.

kshakir:
Renamed the singleton services store to the engine services store.
Added metadata services store to the system startup.
Metadata liquibase configured via a new set of xml, with the same tables.
Added jitter to MetadataServiceActorSpec to ensure order is constistent.
Removed unused SqlDatabase from WorkflowStoreActor/WorkflowStoreEngineActor.
Fixed filename of CustomLabelEntry.scala.


CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # Cromwell Change Log
 
+## 30 Release Notes
+
+### Other changes
+
+* **Database**  
+If desired, the metadata may be optionally stored in a separate SQL database than the of the operations data. See the
```
katevoss: I'd phrase this as "You have the option of storing the metadata in a separate SQL database than the *<database?>* containing the operations data."

CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
...
+[README](https://github.com/broadinstitute/cromwell#database) or the `database` section in
+[cromwell.examples.conf](https://github.com/broadinstitute/cromwell/blob/develop/cromwell.examples.conf) for more
+information. Cromwell will not move any existing data automatically. When switching connection information for an
```
katevoss: TOL: I would move the "Cromwell will not move any existing data automatically" to be the last sentence in the paragraph, as it is a clarification on the sentence about switching the connection information, and how it will behave.

README.md
```diff
@@ -576,6 +576,36 @@ database {
 
 By default batch inserts will be processed in blocks of 2000. To modify this value add the field `insert-batch-size` to the `database` stanza.
 
+Cromwell metadata (paths to job results, start and end times, etc.) grow at a significantly faster rate than operations
+data (running workflows, previous call cache results, etc.). To use a separate database for operations and/or metadata,
```
katevoss: TOL: there is no other information about operations data in the README, so 
1) should there be more info about operations metadata? 
2) maybe you could add a little more information here about it; ex what about the running workflows is returned, or about previous call cache results; Is it the number of each? Their workflow id? Status?

geoffjentry: the real split is that the metadata event store contains information that will find its way to users. the other db stuff is for internal purposes

geoffjentry:
As I'm reading through the docs, conf, and such I'm finding `operations` to be confusing even though I know what it is. The only other thing I can think of would be `engine`, but as I noted yesterday that's not strictly a true statement - but I think it at least captures the spirit of things well enough.

Apologies in advance as I'm finding it hard to verbalize what I find confusing but I think the issue is that it doesn't map to any concept we have other than "it's a grab bag of everything that's not otherwise described", which feels odd to me.

geoffjentry:
Pinging @cjllanwarne as he was the prometheus of `operations`

README.md
```diff
@@ -576,6 +576,36 @@ database {
...
+      url = "jdbc:hsqldb:file:metadata-db-file-path;shutdown=false;hsqldb.tx=mvcc"
+      connectionTimeout = 3000
+    }
+  }
+  operations {
+    profile = "slick.jdbc.HsqldbProfile$"
+    db {
+      driver = "org.hsqldb.jdbcDriver"
+      # Continue to store operations data in memory.
+      url = "jdbc:hsqldb:mem:operations-schema;shutdown=false;hsqldb.tx=mvcc"
```
geoffjentry: This also provides the "two connections to the same DB" functionality which has come up in other circumstances, correct?

CHANGELOG.md
geoffjentry: Something which happens frequently is that a user will see a note like this in the release notes and then ask how they'd know if they'd want to do this.

Because I think our changelog is already way too verbose I don't want to see such an explanation here, but it'd be good to find a way to nip those Qs in the bud.

engine/src/main/scala/cromwell/server/CromwellSystem.scala
```diff
@@ -5,13 +5,13 @@ import akka.http.scaladsl.Http
...
 import cromwell.engine.backend.{BackendConfiguration, CromwellBackends}
-import cromwell.services.SingletonServicesStore
+import cromwell.services.{MetadataServicesStore, OperationsServicesStore}
 
 import scala.concurrent.Future
 
 trait CromwellSystem {
   /*
-  Initialize the services-store (aka a collection of DAOs), if they haven't been already, before starting any services
+  Initialize the services-stores (aka a collection of DAOs), if they haven't been already, before starting any services
```
geoffjentry: I realize you just pluralized this here but this seems funky. I suspect the issue is just with how the rest of the comment is phrased. I think what the comment was trying to get at was that the services in the service registry were being kicked off? TBH I'm not really sure, but it seems like something either about the wording or what's actually happening isn't accurate

geoffjentry:
@kshakir Beyond unit tests, how was this tested? It'd be good to try to throw this at some live mysql installations of various configurations to make sure it doesn't blow up

geoffjentry:
👍 modulo some picking on nomenclature and wording of docs

CHANGELOG.md
katevoss: I recommended to @kshakir that he add a note that this is "Experimental" so that users would be less likely to use it without knowing why they want it, and that it is likely to change. That should keep some of the questions at bay.

README.md
katevoss: That's good information for users to know!

kshakir:
I'm fine with whatever name for the "non-metadata", as those specific changes don't touch liquibase and thus _may_ be changed again in a future PR. Happy to use `engine` in place of `operations`.

Btw, I believe @cjllanwarne may have said `operational` and I mistakenly ran with `operations`. 🤔

CHANGELOG.md
kshakir: Whoops, will do.

README.md
kshakir: Yes. Even if you don't update/split the configuration, there are now _two separate_ Data Access Objects, each with their own slick queues, hikari connection pools, etc.

CHANGELOG.md
geoffjentry: I suspect that the people I have in mind will get even more excited at that but I'm willing to give it a trial balloon here :)

README.md
geoffjentry: Cool. I didn't realize you were birdstoning here. Since this comes up sometimes in our "eh, you could *try* that" performance suggestions, at least now it's already reayd to go :)

geoffjentry:
@kshakir Yeah I really apologize as it sucks when people nit without having their own suggestion but `operations` just really seems like something that's going to cause confusion. I think I'd have the same reaction to `operational` but eh.

How about "THERE_BE_DRAGONS"?

kshakir:
Don't [tempt](http://www.youtube.com/watch?v=7NEWA32jqk0) me ;)

engine/src/main/scala/cromwell/server/CromwellSystem.scala
kshakir: Will look at another edit besides just adding the `s`

Horneth:
LGTM ! 👍

CHANGELOG.md
```diff
@@ -5,12 +5,13 @@
...
-[README](https://github.com/broadinstitute/cromwell#database) or the `database` section in
+You have the option of storing the metadata in a separate SQL database than the database containing the internal engine
+data. See the [README](https://github.com/broadinstitute/cromwell#database) or the `database` section in
 [cromwell.examples.conf](https://github.com/broadinstitute/cromwell/blob/develop/cromwell.examples.conf) for more
-information. Cromwell will not move any existing data automatically. When switching connection information for an
-existing database containing historical data, the tables should be manually replicated from one database instance to
-another using the tools appropriate for your specific database types.
+information. When switching connection information for an existing database containing historical data, the tables
+should be manually replicated from one database instance to another using the tools appropriate for your specific
+database types. Cromwell will not move any existing data automatically. This feature should be considered experimental
```
katevoss: 👍

README.md
```diff
@@ -576,35 +576,27 @@ database {
 
 By default batch inserts will be processed in blocks of 2000. To modify this value add the field `insert-batch-size` to the `database` stanza.
 
-Cromwell metadata (paths to job results, start and end times, etc.) grow at a significantly faster rate than operations
-data (running workflows, previous call cache results, etc.). To use a separate database for operations and/or metadata,
-under the `database` config section, configure a sub-path for `operations` and `metadata` with custom settings.
+Cromwell stores metadata about each job and workflow intended for end users. This metadata includes paths to job
+results, start and end times, etc. The metadata grows at a significantly faster rate than the rest of the internal
```
katevoss: 👍

kshakir:
Re:

> Beyond unit tests, how was this tested?

The DBMS CI tests are testing fresh MySQL databases for both engine and metadata.

I manually started running a workflow on develop with MySql. Stopped Cromwell, switched to this branch, and restarted.

Liquibase ran successfully, marking the change_logs as "MARK_RAN" vs. "EXECUTED". The jobs/workflow resumed and succeeded.

PR #2577, opened 2017-08-25
relax deprecation warning in anticipation of cats 1.0.0-MF

danbills:


mcovarr:
uh sure I guess :+1:

danbills:
I realized that downstream libs won't automagically get this so no need to relax this

PR #2578, opened 2017-08-29
v2 submission api requires type and type version.

kshakir:
v1 submission api returns an HTTP warning when using older form data names.
All unrecognized api versions return bad statuses.
Added additional specs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## 30 Release Notes
 
+### Breaking Changes
+
+* All endpoints now require valid version numbers.
+
+* `POST /api/workflows/v2` now requires a `workflowType` and a `workflowTypeVersion`. These are optional for the `v1`
+api. See the [README](https://github.com/broadinstitute/cromwell#post-apiworkflowsversion) for more information.
+
 ### Other changes
 
 * **Database**  
--- a/README.md
+++ b/README.md
@@ -2590,7 +2590,7 @@ This endpoint accepts a POST request with a `multipart/form-data` encoded body. 
 * `workflowSource` - *Required* Contains the workflow source file to submit for execution.
 * `workflowType` - *Required* The type of the `workflowSource`, for example "WDL".
    Required starting with api version "v2".
-* `workflowTypeVersion` - *Required* The version of the `workflowType`, for example "1".
+* `workflowTypeVersion` - *Required* The version of the `workflowType`, for example "draft-2".
    Required starting with api version "v2".
 * `workflowInputs` - *Optional* JSON file containing the inputs.  For WDL workflows a skeleton file can be generated from [wdltool](https://github.com/broadinstitute/wdltool) using the "inputs" subcommand.
 * `workflowInputs_n` - *Optional* Where `n` is an integer. JSON file containing the 'n'th set of auxiliary inputs.
--- a/README.md
+++ b/README.md
@@ -2588,6 +2588,10 @@ All web server requests include an API version in the url. The current version i
 This endpoint accepts a POST request with a `multipart/form-data` encoded body.  The form fields that may be included are:
 
 * `workflowSource` - *Required* Contains the workflow source file to submit for execution.
+* `workflowType` - *Required* The type of the `workflowSource`, for example "WDL".
+   Required starting with api version "v2".
+* `workflowTypeVersion` - *Required* The version of the `workflowType`, for example "1".
+   Required starting with api version "v2".
 * `workflowInputs` - *Optional* JSON file containing the inputs.  For WDL workflows a skeleton file can be generated from [wdltool](https://github.com/broadinstitute/wdltool) using the "inputs" subcommand.
 * `workflowInputs_n` - *Optional* Where `n` is an integer. JSON file containing the 'n'th set of auxiliary inputs.
 * `workflowOptions` - *Optional* JSON file containing options for this workflow execution.  See the [run](#run) CLI sub-command for some more information about this.
--- a/engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
+++ b/engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
@@ -3,40 +3,39 @@ package cromwell.webservice
 import java.util.UUID
 
 import akka.actor.{ActorRef, ActorRefFactory}
-import akka.http.scaladsl.server.Directives._
-
-import scala.concurrent.{ExecutionContext, Future}
-import akka.http.scaladsl.model._
-import akka.http.scaladsl.model.Multipart.BodyPart
-import akka.stream.ActorMaterializer
-import cromwell.engine.backend.BackendConfiguration
 import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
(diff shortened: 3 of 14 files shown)


engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
```diff
@@ -180,59 +181,73 @@ trait CromwellApiService {
...
+      bodyPart => bodyPart.toStrict(duration).map(strict => bodyPart.name -> strict.entity.data)
     }.runFold(Map.empty[String, ByteString])((map, tuple) => map + tuple)
 
     def toResponse(workflowId: WorkflowId): WorkflowSubmitResponse = {
       WorkflowSubmitResponse(workflowId.toString, WorkflowSubmitted.toString)
     }
 
-    def askSubmit(command: WorkflowStoreActor.WorkflowStoreActorSubmitCommand): Route = {
+    def askSubmit(command: WorkflowStoreActor.WorkflowStoreActorSubmitCommand, warnings: Seq[String]): Route = {
       // NOTE: Do not blindly coppy the akka-http -to- ask-actor pattern below without knowing the pros and cons.
```
mcovarr: Also don't blindly copy this spelling of copy.

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
```diff
@@ -285,16 +300,40 @@ object CromwellApiService {
...
+      Need a quoted string.
+      https://stackoverflow.com/questions/7886782
+
+      Using a poor version of ~~#!
+      https://github.com/akka/akka-http/blob/v10.0.9/akka-http-core/src/main/scala/akka/http/impl/util/Rendering.scala#L206
+       */
+      val quotedString = "\"" + warning.replaceAll("\"","\\\\\"").replaceAll("[\\r\\n]+", " ").trim + "\""
+
+      // https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.46
+      RawHeader("Warning", s"299 cromwell/$cromwellVersion $quotedString")
```
mcovarr: kinda surprised we have to roll all this ourselves

README.md
```diff
@@ -2587,6 +2587,10 @@ All web server requests include an API version in the url. The current version i
 This endpoint accepts a POST request with a `multipart/form-data` encoded body.  The form fields that may be included are:
 
 * `workflowSource` - *Required* Contains the workflow source file to submit for execution.
+* `workflowType` - *Required* The type of the `workflowSource`, for example "WDL".
```
danbills: ToL can't we default to Wdl if not specified?

engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
```diff
@@ -56,8 +61,11 @@ object PartialWorkflowSources {
         partialSources.copy(workflowInputsAux = partialSources.workflowInputsAux + (index -> data.utf8String))
       } else if (name == "workflowOptions") {
         partialSources.copy(workflowOptions = Option(data.utf8String))
-      } else if (name == "wdlDependencies" || name == "workflowDependencies") {
-        if (name == "wdlDependencies") deprecationWarning(out = "wdlDependencies", in = "workflowDependencies")
+      } else if (name == "wdlDependencies" && version == 1) {
+        val warning = deprecationWarning(out = "wdlDependencies", in = "workflowDependencies")
+        val warnings = warning +: partialSources.warnings
+        partialSources.copy(zippedImports = Option(data.toArray), warnings = warnings)
+      } else if (name == "workflowDependencies") {
```
danbills: a pattern match on `(name, version)` would go a long way

engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
```diff
@@ -96,30 +106,54 @@ object PartialWorkflowSources {
...
-  private def deprecationWarning(out: String, in: String): Unit = {
+  private def deprecationWarning(out: String, in: String): String = {
     val warning =
       s"""
          |The '$out' parameter name has been deprecated in favor of '$in'.
          |Support for '$out' will be removed from future versions of Cromwell.
          |Please switch to using '$in' in future submissions.
          """.stripMargin
     log.warn(warning)
+    warning
```
danbills: just remove `warning` declaration

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
kshakir: Hehe, was hoping no one noticed that. I'm going to leave it as a "watermark".

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
kshakir: I couldn't find anything in the existing [APIs](https://github.com/akka/akka-http/blob/v10.0.9/akka-http-core/src/main/scala/akka/http/scaladsl/model/headers/headers.scala) or online. If anyone knows something I missed, happy to change this now or later.

README.md
kshakir: We _could_ default the type and type-version, but I believe tech-talk discussions said that the point of bumping to `v2` was the requirement that these be specified. If one doesn't want to specify the values, `v1` will still work (for now).

engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
kshakir: Can't remove this afaik. `warning` is used twice, once logged server side on the previous line, then returned here.

CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 ## 30 Release Notes
 
+### Breaking Changes
+
+* All endpoints now require valid version numbers.
+
+* `POST /api/workflows/v2` now requires a `workflowType` and a `workflowTypeVersion`. These are optional for the `v1`
+api. See the [README](https://github.com/broadinstitute/cromwell#post-apiworkflowsversion) for more information.
```
katevoss: ToL does this mean there is an official V2 of all the endpoints, or just `/api/workflows`?

README.md
geoffjentry: I missed that TT but I recall having the same initial reaction as @danbills when I was caught up. However after chatting w/ @kshakir I remember my thinking changing to the percentage of users actually constructing URLs by hand being quite low, not sure it's worth it.

CHANGELOG.md
kshakir: Metadata v2 already existed (it's the _exact_ same as v1 due to... at this point no good reason). POST /api/workflows, aka workflow submissions, will now have a v2 that behave differently.

katevoss: Huh, TIL!

geoffjentry: Perhaps this should be a TT tomorrow after standup but I don't think we should be having multiple versions of every endpoint. That's not how most sites do their APIs, at least not in my experience. Every endpoint should accept a `v2` parameter, IMO.

README.md
geoffjentry: Although if we were to do this we could get away with just keeping `v1` i believe, as then these changes would be backwards compatible

kshakir:
After discussion, replaced by #2586

PR #2581, opened 2017-08-29
Kill Docker processes for realz [PREGULL]

mcovarr:
A different approach to #1126 from that proposed in the ticket.  This records the container ID at `docker run` time and uses that to `docker kill` the process on abort.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1214,10 +1214,11 @@ backend {
 }
 ```
 
-If the backend supports docker, another optional configuration key `backend.providers.<backend>.config.submit-docker` may be specified. When the WDL contains a docker runtime attribute, this command will be provided the two additional variables:
+If the backend supports docker, another optional configuration key `backend.providers.<backend>.config.submit-docker` may be specified. When the WDL contains a docker runtime attribute, this command will be provided three additional variables:
 
 * `docker` - The docker image name.
 * `docker_cwd` - The path where `cwd` should be mounted within the docker container.
+* `docker_cid` - The host path to which the [container ID file](https://docs.docker.com/engine/reference/run/#pid-equivalent) should be written.
 
 ```
 backend {
--- a/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
+++ b/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
@@ -57,6 +57,7 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
 
   val SIGTERM = 143
   val SIGINT = 130
+  val SIGKILL = 137
 
   /** The type of the run info when a job is started. */
   type StandardAsyncRunInfo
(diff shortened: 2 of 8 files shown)


supportedBackends/sfs/src/main/scala/cromwell/backend/impl/sfs/config/ConfigAsyncJobExecutionActor.scala
```diff
@@ -119,7 +136,13 @@ sealed trait ConfigAsyncJobExecutionActor extends SharedFileSystemAsyncJobExecut
   * @param standardParams Params for running a shared file system job.
   */
 class BackgroundConfigAsyncJobExecutionActor(override val standardParams: StandardAsyncExecutionActorParams)
-  extends ConfigAsyncJobExecutionActor with BackgroundAsyncJobExecutionActor
+  extends ConfigAsyncJobExecutionActor with BackgroundAsyncJobExecutionActor {
+
+  override def killArgs(job: StandardAsyncJob): SharedFileSystemCommand = {
+    if (isDockerRun) jobScriptArgs(job, "kill", KillDockerTask, Map(DockerCidInput -> dockerCidInputValue))
```
kshakir: ToL: This will need to be revisited when we finally get UGE docker [support](https://blogs.univa.com/2016/05/new-version-of-univa-grid-engine-now-supports-docker-containers/)

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesJobPaths.scala
```diff
@@ -22,6 +22,7 @@ final case class JesJobPaths(override val workflowPaths: JesWorkflowPaths, jobKe
   override val stdoutFilename: String = s"$jesLogBasename-stdout.log"
   override val stderrFilename: String = s"$jesLogBasename-stderr.log"
   override val scriptFilename: String = s"${JesJobPaths.JesExecParamName}.sh"
+  override val dockerCidFilename: String = "not_used_on_jes.txt"
```
kshakir: Confirmed that this override is not needed for compilation.

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -213,7 +214,7 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
     * If the `command` errors for some reason, put a "-1" into the rc file.
     */
   def redirectOutputs(command: String): String = {
-    s"$command > ${jobPaths.stdout} 2> ${jobPaths.stderr} < /dev/null || echo -1 > ${jobPaths.returnCode}"
+    s"""$command > ${jobPaths.stdout} 2> ${jobPaths.stderr} < /dev/null || { rc=$$?; if [ "$$rc" -gt "127" ]; then echo $$rc; else echo -1; fi } > ${jobPaths.returnCode}"""
```
danbills: ToL what does > 127 mean?

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesJobPaths.scala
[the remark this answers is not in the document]
mcovarr: it's gone now actually.

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
mcovarr: What I *meant* for this to do was capture "death by signal", but from the reference below I learned this isn't quite right. Plus I should comment the intent here since it's not exactly obvious.  🙂 

http://www.tldp.org/LDP/abs/html/exitcodes.html

README.md
```diff
@@ -1214,10 +1214,11 @@ backend {
 }
 ```
 
-If the backend supports docker, another optional configuration key `backend.providers.<backend>.config.submit-docker` may be specified. When the WDL contains a docker runtime attribute, this command will be provided the two additional variables:
+If the backend supports docker, another optional configuration key `backend.providers.<backend>.config.submit-docker` may be specified. When the WDL contains a docker runtime attribute, this command will be provided three additional variables:
 
 * `docker` - The docker image name.
 * `docker_cwd` - The path where `cwd` should be mounted within the docker container.
+* `docker_cid` - The host path to which the container ID file should be written.
```
katevoss: TOL should this go in the Release Notes? Maybe under a section "Important fixes", as this has been broken for a while.

mcovarr: Sounds good to me.

danbills: I stared at this for a a minute too.  A link to the reference doc would be helpful too: 
https://docs.docker.com/engine/reference/run/#pid-equivalent

PR #2583, opened 2017-08-30
construct client with optional credentials

delocalizer:
This is certainly useful for us, perhaps for others.

### DIFF
--- a/cromwellApiClient/src/main/scala/cromwell/api/CromwellClient.scala
+++ b/cromwellApiClient/src/main/scala/cromwell/api/CromwellClient.scala
@@ -6,6 +6,8 @@ import akka.http.scaladsl.Http
 import akka.actor.ActorSystem
 import akka.http.scaladsl.coding.{Deflate, Gzip, NoCoding}
 import akka.http.scaladsl.model.{HttpEntity, _}
+import akka.http.scaladsl.model.headers.Authorization
+import akka.http.scaladsl.model.headers.HttpCredentials
 import akka.http.scaladsl.model.headers.HttpEncodings
 import akka.http.scaladsl.unmarshalling.{Unmarshal, Unmarshaller}
 import akka.stream.ActorMaterializer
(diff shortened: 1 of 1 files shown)


delocalizer:
As background, we run cromwell server behind an authenticating proxy.

cromwellApiClient/src/main/scala/cromwell/api/CromwellClient.scala
```diff
@@ -19,7 +21,10 @@ import cromwell.api.CromwellClient._
 
 import scala.util.{Failure, Success, Try}
 
-class CromwellClient(val cromwellUrl: URL, val apiVersion: String)(implicit actorSystem: ActorSystem, materializer: ActorMaterializer) {
+class CromwellClient(val cromwellUrl: URL, val apiVersion: String, val credentials: Option[HttpCredentials]=None)(implicit actorSystem: ActorSystem, materializer: ActorMaterializer) {
+
+  lazy val authHeader = credentials.map { Authorization(_) }
+  lazy val commonHeaders = authHeader.toList
```
geoffjentry: Why not just `toList` it up above? This is only being used in the one place down below, not sure I understand the multiple layers here.

geoffjentry:
👍 - I suspect we'll be running into this soon as well, thanks.

cjllanwarne:
Very cool! I know that we'd have needed this soon too so really helpful!

cjllanwarne:
👍

cromwellApiClient/src/main/scala/cromwell/api/CromwellClient.scala
cjllanwarne: I could be persuaded that splitting it up like this makes things more readable/understandable and the compiler will probably inline it anyway if it gets called a lot. So, I'd be happy either way.

geoffjentry: Don't completely disagree w/ that but my main question was having two basically the same variables at the top level which ultimately are only used in the one function. And really the question part of my main question was if there was a reason for that setup :)

delocalizer: Yeah, I thought for clarity and a little future-proofing that `executeRequest` should add all common headers, and creating the List of them should be separate from creating each one of them from possibly optional pieces.

geoffjentry: I was almost going to suggest inlining the whole thing in the one spot so you can see where my biases lie :) Not a big deal, carry on. Will merge this, thanks @delocalizer

PR #2586, opened 2017-08-31
Default workflow type is now configurable.

kshakir:
Submission api returns an HTTP warning when using older form data names.
Switched PWS if statements to case statements.
Switched unused version for metadata endpoint back to v1 in swagger.

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -160,6 +160,9 @@ workflow-options {
   # Workflow-failure-mode determines what happens to other calls when a call fails. Can be either ContinueWhilePossible or NoNewCalls.
   # Can also be overridden in workflow options. Defaults to NoNewCalls. Uncomment to change:
   #workflow-failure-mode: "ContinueWhilePossible"
+
+  # When a workflow type is not provided on workflow submission, this specifies the default type.
+  default.workflow-type: "WDL"
 }
 
 # Optional call-caching configuration.
--- a/core/src/main/scala/cromwell/core/WorkflowOptions.scala
+++ b/core/src/main/scala/cromwell/core/WorkflowOptions.scala
@@ -3,6 +3,7 @@ package cromwell.core
 import com.typesafe.config.ConfigFactory
 import lenthall.util.TryUtil
 import spray.json._
+import net.ceedubs.ficus.Ficus._
 
 import scala.collection.JavaConverters._
 import scala.util.{Failure, Success, Try}
@@ -131,6 +132,10 @@ object WorkflowOptions {
   }
 
   val empty = WorkflowOptions.fromMap(Map.empty).get
+
+  lazy val defaultWorkflowType: Option[String] = Option(WorkflowOptionsConf.getString("default.workflow-type"))
+  lazy val defaultWorkflowTypeVersion: Option[String] =
+    WorkflowOptionsConf.as[Option[String]]("default.workflow-type-version")
 }
 
 case class WorkflowOptions(jsObject: JsObject) {
--- a/engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
+++ b/engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
@@ -3,41 +3,40 @@ package cromwell.webservice
 import java.util.UUID
 
 import akka.actor.{ActorRef, ActorRefFactory}
-import akka.http.scaladsl.server.Directives._
-
-import scala.concurrent.{ExecutionContext, Future}
-import akka.http.scaladsl.model._
-import akka.http.scaladsl.model.Multipart.BodyPart
-import akka.stream.ActorMaterializer
-import cromwell.engine.backend.BackendConfiguration
 import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
-import cromwell.core.{WorkflowAborted, WorkflowId, WorkflowSubmitted}
-import cromwell.core.Dispatcher.ApiDispatcher
-import cromwell.engine.workflow.workflowstore.{WorkflowStoreActor, WorkflowStoreEngineActor, WorkflowStoreSubmitActor}
-import akka.pattern.{AskTimeoutException, ask}
-import akka.util.{ByteString, Timeout}
-import net.ceedubs.ficus.Ficus._
-import cromwell.engine.workflow.WorkflowManagerActor
-import cromwell.services.metadata.MetadataService._
-import cromwell.webservice.metadata.{MetadataBuilderActor, WorkflowQueryPagination}
-import cromwell.webservice.metadata.MetadataBuilderActor.{BuiltMetadataResponse, FailedMetadataResponse, MetadataBuilderActorResponse}
-import WorkflowJsonSupport._
+import akka.http.scaladsl.marshalling.{ToEntityMarshaller, ToResponseMarshallable}
+import akka.http.scaladsl.model._
+import akka.http.scaladsl.model.headers.RawHeader
+import akka.http.scaladsl.server.Directives._
 import akka.http.scaladsl.server.Route
+import akka.pattern.{AskTimeoutException, ask}
+import akka.stream.ActorMaterializer
+import akka.util.{ByteString, Timeout}
 import cats.data.NonEmptyList
 import cats.data.Validated.{Invalid, Valid}
 import com.typesafe.config.ConfigFactory
+import cromwell.core.Dispatcher.ApiDispatcher
 import cromwell.core.labels.Labels
+import cromwell.core.{WorkflowAborted, WorkflowId, WorkflowSubmitted}
+import cromwell.engine.backend.BackendConfiguration
 import cromwell.engine.instrumentation.HttpInstrumentation
+import cromwell.engine.workflow.WorkflowManagerActor
 import cromwell.engine.workflow.WorkflowManagerActor.WorkflowNotFoundException
 import cromwell.engine.workflow.lifecycle.execution.callcaching.CallCacheDiffActor.{BuiltCallCacheDiffResponse, CachedCallNotFoundException, CallCacheDiffActorResponse, FailedCallCacheDiffResponse}
 import cromwell.engine.workflow.lifecycle.execution.callcaching.{CallCacheDiffActor, CallCacheDiffQueryParameter}
 import cromwell.engine.workflow.workflowstore.WorkflowStoreEngineActor.WorkflowStoreEngineAbortResponse
+import cromwell.engine.workflow.workflowstore.{WorkflowStoreActor, WorkflowStoreEngineActor, WorkflowStoreSubmitActor}
 import cromwell.server.CromwellShutdown
+import cromwell.services.metadata.MetadataService._
 import cromwell.webservice.LabelsManagerActor._
+import cromwell.webservice.WorkflowJsonSupport._
+import cromwell.webservice.metadata.MetadataBuilderActor.{BuiltMetadataResponse, FailedMetadataResponse, MetadataBuilderActorResponse}
+import cromwell.webservice.metadata.{MetadataBuilderActor, WorkflowQueryPagination}
 import lenthall.exception.AggregatedMessageException
-import spray.json.JsObject
+import net.ceedubs.ficus.Ficus._
 
 import scala.concurrent.duration._
+import scala.concurrent.{ExecutionContext, Future}
 import scala.util.{Failure, Success, Try}
 
 trait CromwellApiService extends HttpInstrumentation {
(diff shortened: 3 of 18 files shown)


engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
```diff
@@ -20,51 +20,54 @@ final case class PartialWorkflowSources(workflowSource: Option[WorkflowSource],
...
-        throw new IllegalArgumentException(s"Unexpected body part name: $name")
+  def fromSubmitRoute(formData: Map[String, ByteString],
+                      allowNoInputs: Boolean): Try[Seq[WorkflowSourceFilesCollection]] = {
+    val partialSources: Try[PartialWorkflowSources] = Try {
+      formData.foldLeft(PartialWorkflowSources.empty) {
+        (partialSources: PartialWorkflowSources, kv: (String, ByteString)) =>
+          val (name, data) = kv
+
+          name match {
+            case "wdlSource" =>
```
mcovarr: yay match

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
```diff
@@ -285,16 +290,40 @@ object CromwellApiService {
...
+                         (implicit mt: ToEntityMarshaller[A]): Route = {
+    val warningHeaders = warnings.toIndexedSeq map { warning =>
+      /*
+      Need a quoted string.
+      https://stackoverflow.com/questions/7886782
+
+      Using a poor version of ~~#!
+      https://github.com/akka/akka-http/blob/v10.0.9/akka-http-core/src/main/scala/akka/http/impl/util/Rendering.scala#L206
+       */
+      val quotedString = "\"" + warning.replaceAll("\"","\\\\\"").replaceAll("[\\r\\n]+", " ").trim + "\""
```
cjllanwarne: Whoa! here be { 🐲 🐉 🐲 }!

"I'm going to escape my own HTML" sounds like a bad idea if this is a security mechanism. If it's just "so it's not a horrible UI" then I'm less worried...

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
```diff
@@ -285,16 +290,40 @@ object CromwellApiService {
...
+
+      // https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.46
+      RawHeader("Warning", s"299 cromwell/$cromwellVersion $quotedString")
```
cjllanwarne: ToL: Is this actually looked at? 

This feels a little bit like `... > /dev/null`

engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
```diff
@@ -20,51 +20,54 @@ final case class PartialWorkflowSources(workflowSource: Option[WorkflowSource],
...
-                                        zippedImports: Option[Array[Byte]])
+                                        zippedImports: Option[Array[Byte]],
+                                        warnings: Seq[String])
 
 object PartialWorkflowSources {
   val log = LoggerFactory.getLogger(classOf[PartialWorkflowSources])
 
   def empty = PartialWorkflowSources(
     workflowSource = None,
-    // TODO do not hardcode, especially not out here at the boundary layer good gravy
```
cjllanwarne: - Removing a `TODO`: 😄 
- Removing a comment about gravy: 😭

engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
```diff
@@ -20,51 +20,54 @@ final case class PartialWorkflowSources(workflowSource: Option[WorkflowSource],
...
+              val index = name.stripPrefix("workflowInputs_").toInt
+              partialSources.copy(workflowInputsAux = partialSources.workflowInputsAux + (index -> data.utf8String))
+            case "workflowOptions" => partialSources.copy(workflowOptions = Option(data.utf8String))
+            case "wdlDependencies" =>
+              val warning = deprecationWarning(out = "wdlDependencies", in = "workflowDependencies")
+              val warnings = warning +: partialSources.warnings
+              partialSources.copy(zippedImports = Option(data.toArray), warnings = warnings)
+            case "workflowDependencies" => partialSources.copy(zippedImports = Option(data.toArray))
+            case "customLabels" => partialSources.copy(customLabels = Option(data.utf8String))
+            case _ => throw new IllegalArgumentException(s"Unexpected body part name: $name")
```
cjllanwarne: ToL, NYF: I think this might be a better UX as a warning?

engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
```diff
@@ -96,30 +100,51 @@ object PartialWorkflowSources {
       case _ => s"Incomplete workflow submission: $partialSource".invalidNel
     }
 
+    def validateWorkflowType(partialSource: PartialWorkflowSources): ErrorOr[Option[WorkflowType]] = {
+      partialSource.workflowType match {
+        case Some(src) => Option(src).validNel
```
cjllanwarne: - What's the value of the `ErrorOr[]` wrapper if they're both valid?
- ToL: `Some(src) => Option(src)` makes me cry a little inside. Could it be:
 `Some(_) => partialSource.workflowType.validNel`
- ToL: it feels like there's probably some helper function somewhere along the lines of:
`partialSource.workflowType.noneMap(WorkflowOptions.defaultWorkflowType)`... 🤷‍♂️

engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
```diff
@@ -96,30 +100,51 @@ object PartialWorkflowSources {
...
+    }
+
     partialSources match {
       case Valid(partialSource) =>
-        (validateWorkflowSource(partialSource) |@| validateInputs(partialSource) |@| validateOptions(partialSource.workflowOptions)) map {
-          case (wfSource, wfInputs, wfOptions) =>
+        (validateWorkflowSource(partialSource) |@|
+          validateInputs(partialSource) |@|
+          validateOptions(partialSource.workflowOptions) |@|
+          validateWorkflowType(partialSource) |@|
```
cjllanwarne: ToL: All these `|@|`s will be gone in a month (for the `cats` update)

src/main/scala/cromwell/CommandLineParser.scala
```diff
@@ -14,8 +15,8 @@ object CommandLineParser extends App {
                                   workflowSource: Option[Path] = None,
                                   workflowInputs: Option[Path] = None,
                                   workflowOptions: Option[Path] = None,
-                                  workflowType: Option[String] = Option("WDL"),
-                                  workflowTypeVersion: Option[String] = Option("v2.0-draft"),
+                                  workflowType: Option[String] = WorkflowOptions.defaultWorkflowType,
```
cjllanwarne: Does whatever adds the failsafe "or else just use WDL" still get applied in command line mode?

src/main/scala/cromwell/CromwellEntryPoint.scala
```diff
@@ -153,7 +154,8 @@ object CromwellEntryPoint extends GracefulStopSupport {
           workflowTypeVersion = None,
```
cjllanwarne: ToL: is this really meant to be hard-coded to `None`

core/src/main/resources/reference.conf
```diff
@@ -157,6 +157,9 @@ workflow-options {
   # Workflow-failure-mode determines what happens to other calls when a call fails. Can be either ContinueWhilePossible or NoNewCalls.
   # Can also be overridden in workflow options. Defaults to NoNewCalls. Uncomment to change:
   #workflow-failure-mode: "ContinueWhilePossible"
+
+  # When a workflow type is not provided on workflow submission, this specifies the default type.
+  default.workflow-type: WDL
```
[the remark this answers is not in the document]
kshakir: HOCON doesn't require quoting strings like JSON does, but I'll change it anyway.

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
kshakir: It's not security, and not HTML. It's HTTP. HTTP has some weird type of `quoted-string` type that goes in the headers. It's is poorly defined IMHO, and I really wish someone else would have implemented this so I could look at what they're doing. The closest I could find was the cryptically named `~~#!` operator buried privately in akka-http-core.

See the links in the comments for more info, or if more comments would be better, let me know.

engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
kshakir: Someday. This line was only touched after a request to switch the existing `if` statements to `case` matches.

engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
kshakir: > What's the value of the ErrorOr[] wrapper if they're both valid?

So our wonderful cartesian ~~combine~~ tie-fighter operator may be used.

> ToL: Some(src) => Option(src) makes me cry a little inside.

Will do.

> ToL: it feels like there's probably some helper function somewhere along the lines of:
partialSource.workflowType.noneMap(WorkflowOptions.defaultWorkflowType)...

Probably is? I'm still absorbing cats knowledge from others. If the correct type-class appears before I merge, I'll use it! Otherwise, we can update these later.

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
cjllanwarne: Is there any reason you can't use the existing `~~#!` method?

src/main/scala/cromwell/CommandLineParser.scala
kshakir: I believe so, but will double check. Doesn't look like `MainSpec` exists anymore...

src/main/scala/cromwell/CromwellEntryPoint.scala
kshakir: Hm... thanks for the ping. Upon further inspection, I'm also finding even more "WDL" hardcodings than I thought. Will address in the next PR patch.

Horneth:
LGTM 👍

core/src/main/scala/cromwell/core/WorkflowOptions.scala
```diff
@@ -131,6 +132,10 @@ object WorkflowOptions {
   }
 
   val empty = WorkflowOptions.fromMap(Map.empty).get
+
+  val defaultWorkflowType: Option[String] = Option(WorkflowOptionsConf.getString("default.workflow-type"))
+  val defaultWorkflowTypeVersion: Option[String] =
```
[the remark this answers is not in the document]
kshakir: The `WorkflowOptionsConf` is `lazy`. If these aren't also lazy, that defeats the original lazy-purpose.

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
kshakir: All of `akka.http.impl.util.Rendering` is package-private. In the past I've used runtime-reflection to "fix" private access methods. But that method didn't strip the newlines we're adding. So I went with just adding a reference pointer showing that this wasn't the first method to handle this weird (again imho) HTTP quoted-string.

core/src/main/scala/cromwell/core/WorkflowOptions.scala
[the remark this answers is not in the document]
geoffjentry: I for one welcome our lazy overlords

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
geoffjentry: This isn't a seagull request or anything, but since my eyebrow was literally just raised I figured I should throw it out there. How is it that we're running into something which apparently no one else does if it requires hijinks of this caliber? 

Is this a case where technically it's supposed to work a certain way but most people don't bother? Is it that typically people aren't putting quoted strings in that header? Something else?

src/main/scala/cromwell/CommandLineParser.scala
kshakir: "Failsafe" works.

src/main/scala/cromwell/CromwellEntryPoint.scala
kshakir: This and all related "WDL" hard codings in main code are gone!

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
kshakir: > How is it that we're running into something which apparently no one else does if it requires hijinks of this caliber?

Here is _one_ example of someone else jumping through loops to return a proper Warning 299: https://github.com/elastic/elasticsearch/blob/v5.5.2/core/src/main/java/org/elasticsearch/common/logging/DeprecationLogger.java#L307-L315

There are more examples even in this file of returning proper deprecation warnings over HTTP. Oracle Java and Akka Http don't have a library to help with this header, afaik. Maybe Apache does?

Overall, I get the impression that this was a well intentioned spec that needed a [second round](https://stackoverflow.com/questions/7886782/what-is-the-exact-syntax-and-semantics-of-a-quoted-string-in-the-http1-1-rfc2616#7888497) of cleanup.

> Is this a case where technically it's supposed to work a certain way but most people don't bother?

Most people (ex: see rest of code in ES class) are returning a very basic "go fix this" warning messages to their clients. Our warnings originate as a paragraph including line feeds. Not horrible, but requires some extra stripping of line feeds before it's sent over HTTP, where a CRLF is also defined as the official separator for header fields.

geoffjentry: mentioned this at standup but this wasn't a request for action. i just wanted to understand what was happening

kshakir:
Coveralls be damned. No idea what's wrong with the github hooks. Merging.

PR #2588, opened 2017-09-01
Are you a fan of upcat?

cjllanwarne:
Also bumped up some minor revisions of `akka` and `akka-http`

### DIFF
--- a/engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
@@ -6,7 +6,7 @@ import cats.data.NonEmptyList
 import cats.data.Validated._
 import cats.instances.list._
 import cats.instances.vector._
-import cats.syntax.cartesian._
+import cats.syntax.apply._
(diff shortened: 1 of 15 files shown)


mcovarr:
yes

engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -205,7 +205,7 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
     val namespaceValidation = validateNamespace(sourceFiles)
     val labelsValidation = validateLabels(sourceFiles.labelsJson)
 
-    (namespaceValidation |@| labelsValidation).tupled flatMap { case (namespace, labels) =>
+    (namespaceValidation, labelsValidation).tupled flatMap { case (namespace, labels) =>
```
mcovarr: I kind of liked your `flatMapN` idea if you still think that makes sense

cjllanwarne: I'll give it a go via lenthall. I think I'll do that via a separate PR though

PR #2601, opened 2017-09-08
Create placeholder file when appropriate

ruchim:
Issue: Currently, regardless of the duplication strategy in JES, the call cache placeholder file is created in case of cache hits. 

Fix: Added a statement that confirms the caching strategy is set to reference in order to upload the call_caching_placeholder.txt.


Horneth:
👍  Thanks for fixing this ! :)

mcovarr:
:+1: 😬  oops

PR #2609, opened 2017-09-11
flatMapN

cjllanwarne:


mcovarr:
huge fan

:+1: PS There were only three usages?

Horneth:
👍

PR #2619, opened 2017-09-12
Adding minor fixes to make spark backend functioning 

jainh:
There are a few minor changes to bring back Sparkbackend to live as it was the last time.


supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkJobExecutionActor.scala
```diff
@@ -210,4 +203,4 @@ class SparkJobExecutionActor(override val jobDescriptor: BackendJobDescriptor,
     }
   }
 
-}
+}
```
mcovarr: \n

geoffjentry:
@jainh I think you need to rebase, my suspicion is that the test failures are due to not being quite up to date on develop.

Also, not for this PR but I'd suggest looking into porting this backend to the standard backend trait, it might help wiht these divergences in the future

cjllanwarne:
Looks good. I agree with Jeff that the red should go away when you rebase

cjllanwarne:
👍

PR #2630, opened 2017-09-14
Fix cromwell.examples.conf

cjllanwarne:
Per comment in https://gatkforums.broadinstitute.org/wdl/discussion/10191/color-in-the-output-to-screen-from-cromwell-run#latest


mcovarr:
:+1:

PR #2633, opened 2017-09-18
Don't create a partial database.metadata stanza in examples.

kshakir:


geoffjentry:
👍

kshakir:
Tested manually via

```bash
docker run -it --rm -v $PWD:$PWD -w $PWD \
  -e 'JAVA_OPTS=-Dconfig.file=cromwell.examples.conf' \
  broadinstitute/cromwell:develop server
```

cromwell.examples.conf
```diff
@@ -622,12 +622,12 @@ database {
...
-    #profile = "slick.jdbc.HsqldbProfile$"
-    #db {
-    #  driver = "org.hsqldb.jdbcDriver"
-    #  url = "jdbc:hsqldb:file:metadata-${uniqueSchema};shutdown=false;hsqldb.tx=mvcc"
-    #  connectionTimeout = 3000
-    #}
-  }
+  #metadata {
+  #  profile = "slick.jdbc.HsqldbProfile$"
+  #  db {
```
cjllanwarne: I thought that recreating the key `metadata` wouldn't erase an existing key `metadata.db`? I thought that was the reason why we couldn't easily remove certain conf entries which exist in the default?

cjllanwarne:
Hooray! 👍

cromwell.examples.conf
kshakir: In this case, the code is detecting that the stanza `metadata` exists, then failing to find the `metadata.db` stanza.

Issue #2624, opened 2017-09-13
Cromwell quickstart


cjllanwarne:
eg something like:
```
$ cromwell quickstart
## BACKEND ##
Would you like to run tasks [local], [sge] or [papi] (Google cloud)?
> [local] | papi
Installing gcloud......
Done! Would you like me to run [gcloud auth] for you to get default credentials set up?
> [yes] |
Running gcloud auth.......
Done!
Backend setup complete!

## DATABASE ##
Would you like to keep a database of past runs (eg so that you can call-cache?)
> [yes] | 
Would you like to use [mysql] or [hsqldb] file?
> [mysql] | 
MySQL detected. Will not reinstall.
Please enter a MySQL username:
> [root] | chris
Please enter the MySQL password:
> [] |
Database setup complete!

## SETTING UP YOUR SYSTEM ##
Writing start/stop script to /usr/local/bin/cromwell.server...
Writing configuration file to /etc/cromwell/cromwell.conf...

To run Cromwell you can now run:
# cromwell.server start
# cromwell submit <wdl> -i <inputs.json>

Good luck!
```

katevoss:
![supercomputer_cyoa_large](https://user-images.githubusercontent.com/2395211/30393179-454aeb34-988d-11e7-955b-490778436f03.jpg)

PR #2635, opened 2017-09-18
Seething Seafowl Suitable Server Start Stop and Status Script

cjllanwarne:
- Starter-for-10 PR.
- Lets us start up Cromwell with a simple
```
# cromwell.server start
```
- Lets us stop Cromwell with a simple
```
# cromwell.server stop
```
- Lets us see whether Cromwell is already running with a simple
```
# cromwell.server status
Cromwell is stopped

# echo $?
1
```
- My ultimate dream is to make this part of the [cromwell quickstart process](https://github.com/broadinstitute/cromwell/issues/2624) to get people up and ready for server mode in just a few simple commands.

### DIFF
--- a/scripts/server/cromwell.server
+++ b/scripts/server/cromwell.server
@@ -1,0 +1,122 @@
+#!/bin/sh
+
+# Startup script for a Cromwell server.
+# Place me somewhere accessible (eg /usr/local/bin/cromwell.server) and use 'cromwell.server start' to begin your Cromwell server.
+
+# Cromwell PID:
+CROMPID=/usr/local/bin/cromwell/cromwell.pid
+
+# Cromwell jar:
+CROMJAR=/usr/local/bin/cromwell/cromwell.jar
+
+# Cromwell conf:
+CROMCONF=/etc/cromwell.conf
+
+# Log redirect:
+CROMLOG=/usr/local/bin/cromwell/cromwell.log
+
+usage()
+{
+  echo "Usage: /bin/sh $0 <start|stop|status|logs|config>"
+}
+
+status()
+{
+  VERBOSE=$1
+  if [ -f "$CROMPID" ]
+    then
+      PID=$(cat $CROMPID)
+      # Check alive:
+      if kill -0 "$PID" >/dev/null 2>&1
+      then
+        [ "$VERBOSE" = "verbose" ] && echo "Cromwell process running as $PID"
+        return 0
+      else
+        echo "Cromwell was terminated unexpectedly. Removing $CROMPID"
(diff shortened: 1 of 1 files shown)


scripts/server/cromwell.server
```diff
@@ -0,0 +1,122 @@
+#!/bin/sh
+
+# Startup script for a Cromwell server.
+# Place me somewhere accessible (eg /usr/local/bin/cromwell.server) and use 'cromwell.server start' to begin your Cromwell server.
+
+# Cromwell PID:
```
danbills: Maybe be explicit w/ `Process ID (PID)`.

scripts/server/cromwell.server
```diff
@@ -0,0 +1,122 @@
...
+CROMPID=/usr/local/bin/cromwell/cromwell.pid
```
danbills: is this directory generally writable by non-root?  perhaps `/tmp`?

scripts/server/cromwell.server
```diff
@@ -0,0 +1,122 @@
...
+
+# Cromwell jar:
+CROMJAR=/usr/local/bin/cromwell/cromwell.jar
```
Horneth: ToL: It would be nice if I could somehow choose where my cromwell jar is. I understand that this is meant to be a "just run cromwell.server start" script but I likely will have to put the jar there before anyway.

Also it seems that this location might not play nicely with someone that has installed cromwell through brew as brew will create a symlink at `/usr/local/bin/cromwell`pointing to the cromwell jar. It could actually be nice to take advantage of that so that you could do
```
brew install cromwell
cromwell.server start
```

scripts/server/cromwell.server
```diff
@@ -0,0 +1,122 @@
+#!/bin/sh
```
geoffjentry: If you haven't already, run these through the shell linter

scripts/server/cromwell.server
geoffjentry: I could see a similar issue with the conf file

scripts/server/cromwell.server
cjllanwarne: Done.

scripts/server/cromwell.server
cjllanwarne: That's actually exactly what I was hoping to do ultimately! The reason I dumped everything in `/usr/local/bin/cromwell/` was because I didn't know any better.
Eg if we can get brew to put down a cromwell jar, and a quickstart script to put down `cromwell.conf` and `cromwell.server` in appropriate places, we should be golden!

PR #2646, opened 2017-09-21
glob mechanism updated from wdl4s

danbills:
This incorporates the changes made from Wdl4s' globbing implementation seen here:
https://github.com/broadinstitute/wdl4s/pull/225/files

Basically a one-line change in `StandardAsyncExecutionActor` and a few details.

### DIFF
--- a/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
+++ b/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
@@ -182,7 +182,7 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
     val rcPath = cwd./(jobPaths.returnCodeFilename)
     val rcTmpPath = rcPath.plusExt("tmp")
 
-    val globFiles = backendEngineFunctions.findGlobOutputs(call, jobDescriptor)
+    val globFiles: Set[WdlGlobFile] = call.globFiles(jobDescriptor.fullyQualifiedInputs)
 
     s"""|#!/bin/bash
         |tmpDir=$$(mktemp -d $cwd/tmp.XXXXXX)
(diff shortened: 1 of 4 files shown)


danbills:
All the dependencies I removed are pulled in transitively from wdl4s.  T

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -182,7 +182,7 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
     val rcPath = cwd./(jobPaths.returnCodeFilename)
     val rcTmpPath = rcPath.plusExt("tmp")
 
-    val globFiles = backendEngineFunctions.findGlobOutputs(call, jobDescriptor)
+    val globFiles: Set[WdlGlobFile] = call.globFiles(jobDescriptor.fullyQualifiedInputs)
```
Horneth: ToL: I think I lied and that we *can* (doesn't mean we have to) do this with what we have already:

```scala
def findGlobOutputs(call: TaskCallNode, jobDescriptor: BackendJobDescriptor): ErrorOr[List[WdlGlobFile]] = {
    call.callable.outputs.flatTraverse[ErrorOr, WdlGlobFile] {
      _.expression.evaluateFiles(jobDescriptor.fullyQualifiedInputs, this, WdlAnyType) map {
        _.toList collect { case glob: WdlGlobFile => glob }
      }
    }
  }
```

danbills: By all means, let's use what we already have!  Does this fall in `TaskDefinition`?  I.e. what is `this`?

Horneth: It falls in the `GlobFunctions` trait. `this` is the `IoFunctionSet`. WDL needs it (or at least claims it does...) to find out what the globs are.

danbills:
using approach suggested by @Horneth in #2656

PR #2648, opened 2017-09-22
Fix compilation

mcovarr:
Should my [wdl4s offering](https://github.com/broadinstitute/wdl4s/pull/222) be fortunate enough to be graced by another thumb, I would want to bump the wdl4s and lenthall refs in Cromwell to the heads of develop of those repos. It looks like some changes crept in to wdl4s before my most recent changes that require a little adjusting here for that to happen.


geoffjentry:
how can i say no to compiling

:+1:

mcovarr:
Compilation included at no extra charge in #2649

PR #2649, opened 2017-09-22
Update MaterializeWorkflowActor to use Parse monad

danbills:


project/Dependencies.scala
```diff
@@ -23,8 +22,7 @@ object Dependencies {
   private val fs2Test = "co.fs2" %% "fs2-io" % fs2V % "test"
 
   private val catsDependencies = List(
-    "org.typelevel" %% "cats-core" % catsV,
-    "com.github.benhutchison" %% "mouse" % "0.9"
+    "org.typelevel" %% "cats-core" % catsV
   ) map (_
     /*
     Exclude test framework cats-laws and its transitive dependency scalacheck.
```
kshakir: While you're here: Now that we're only using `cats-core`, remove these exclusions for `cats`.

danbills:
closing, will re-open from another branch because reasons

danbills:
@mcovarr featuring latest wdl4s and fixes

engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -429,8 +429,9 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
       for {
         wf <- CwlDecoder.decodeAllCwl(cwlFile).map {
           _.select[Workflow].get
+          //TODO move this unsafeRunSync to the fringe so as to leave more power to the caller
         }.value.unsafeRunSync.toValidated
-        executable <- wf.womExecutable
+        executable <- wf.womExecutable.toValidated
```
mcovarr: oh I would have thought this would have gone to `Checked` to take advantage of its real `flatMap`

danbills: it should be, it should really be a `Parse` like decodeAllCwl.  

I was lazy and wrote a TODO.  am doing the hard work now.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -418,20 +425,18 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
...
     try {
       for {
-        wf <- CwlDecoder.decodeAllCwl(cwlFile).map {
-          _.select[Workflow].get
-        }.value.unsafeRunSync.toValidated
-        executable <- wf.womExecutable
-        graph <- executable.graph
+        cwl <- CwlDecoder.decodeAllCwl(cwlFile)
+        wf <- EitherT{ IO{cwl.select[Workflow].toRight(NonEmptyList.one(s"expected a workflow but got a $cwl"))}}
+        executable <- EitherT{IO {wf.womExecutable}}
```
geoffjentry: I apologize as I know you hate these sorts of comments, and yes this is an argument for scalafmt but I hate when sometimes it is "{ I} and then "{I" in other places. I have an opinion on which I prefer but it's really the back and forth which jumps out at me

geoffjentry:
:+1:

engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
danbills: I totally agree, my point w/ scalafmt is that I want the computer to do it for me.  I'll fix it.

geoffjentry: yeah that's why i try to not make this comment every time i see it, i feel a little bad when i do :)

engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -418,20 +425,18 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
...
       for {
-        wf <- CwlDecoder.decodeAllCwl(cwlFile).map {
-          _.select[Workflow].get
-        }.value.unsafeRunSync.toValidated
-        executable <- wf.womExecutable
-        graph <- executable.graph
+        cwl <- CwlDecoder.decodeAllCwl(cwlFile)
+        wf <- EitherT{ IO{ cwl.select[Workflow].toRight(NonEmptyList.one(s"expected a workflow but got a $cwl")) }}
+        executable <- EitherT{ IO{ wf.womExecutable }}
+        graph <- EitherT{ IO{ executable.graph.toEither }}
```
mcovarr: Is it possible to sugar the `EitherT { IO { ... }}`? There's a lot of that in this PR.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -213,34 +218,36 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
...
-      publishLabelsToMetadata(id, validatedNamespace.executable.entryPoint.name, labels)
-      buildWorkflowDescriptor(id, sourceFiles, validatedNamespace, workflowOptions, labels, conf, pathBuilders)
-    }
+    for {
+      validatedNamespace <- namespaceValidation
+      labels <- labelsValidation
+       _ <- pushWfNameMetadataService(validatedNamespace.executable.entryPoint.name)
+      _ <- publishLabelsToMetadata(id, validatedNamespace.executable.entryPoint.name, labels)
+      ewd <- EitherT{ IO{ buildWorkflowDescriptor(id, sourceFiles, validatedNamespace, workflowOptions, labels, conf, pathBuilders).toEither }}
+    } yield ewd
```
mcovarr: ewd?

geoffjentry: as in `eitherWorkflowDescriptor` I assume. And since I dislike having the `modifierRealVarName` thing anyways I'd extra prefer to keep this over the long form :)

danbills: `engineWorkflowDescriptor`, so not following `modifierRealVarName`, just being lazy and potentially more readable.

I go back and forth on long names, if there is consensus I will do whichever.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
danbills: fixed it up using [`fromEither[IO]`](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/data/EitherT.scala#L351)

engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
geoffjentry: I'm :+1: on acronyms for stuff like this but not everyone agrees :)

PR #2654, opened 2017-09-24
Initial health monitor support

geoffjentry:
o Provide health monitor infrastructure to know status of underlying systems
o Provide status endpoint which will query current contents of health monitor
o Provide health monitor implementations for typical and workbench usages
o Moved some google code to a new cloudSupport project

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,7 @@ A [Workflow Management System](https://en.wikipedia.org/wiki/Workflow_management
   * [GET /api/workflows/:version/callcaching/diff](#get-apiworkflowsversioncallcachingdiff)
   * [GET /engine/:version/stats](#get-engineversionstats)
   * [GET /engine/:version/version](#get-engineversionversion)
+  * [GET /engine/:version/status](#get-engineversionstatus)
   * [Error handling](#error-handling)
 * [Developer](#developer)
   * [Generating table of contents on Markdown files](#generating-table-of-contents-on-markdown-files)
@@ -4034,6 +4035,31 @@ Response:
   "cromwell": 23-8be799a-SNAP
 }
 ```
+
+## GET /engine/:version/status
+
+Cromwell will track the status of underlying systems on which it depends, typically connectivity to the database and to Docker Hub. The `status` endpoint will return the current status of these systems. 
+
+Response:
+
+This endpoint will return an `Internal Server Error` if any systems are marked as failing or `OK` otherwise. The exact response will vary based on what is being monitored but adheres to the following pattern. Each system has a boolean `ok` field and an optional array field named `messages` which will be populated with any known errors if the `ok` status is `false`:
+
+```
+{
+  "System Name 1": {
+    "ok": false,
+    "messages": [
+      "Unknown status"
+    ]
+  },
+  "System Name 2": {
+    "ok": true
+  }
+}
+
+```
+ 
+
 
 ## Error handling
 Requests that Cromwell can't process return a failure in the form of a JSON response respecting the following JSON schema:
--- a/cloudSupport/src/main/scala/cromwell/cloudSupport/gcp/gcs/GcsStorage.scala
+++ b/cloudSupport/src/main/scala/cromwell/cloudSupport/gcp/gcs/GcsStorage.scala
@@ -1,0 +1,56 @@
+package cromwell.cloudsupport.gcp.gcs
+
+import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport
+import com.google.api.client.googleapis.media.MediaHttpUploader
+import com.google.api.client.json.jackson2.JacksonFactory
+import com.google.api.gax.retrying.RetrySettings
+import com.google.cloud.storage.StorageOptions
+import com.google.api.services.storage.Storage
+import com.google.auth.Credentials
+import com.google.cloud.storage.contrib.nio.CloudStorageConfiguration
+import com.typesafe.config.ConfigFactory
+import cromwell.cloudsupport.gcp.GoogleConfiguration
+import cromwell.cloudsupport.gcp.http.GoogleHttpTransportOptions.TransportOptions
+import net.ceedubs.ficus.Ficus._
+
+object GcsStorage {
+  val JsonFactory = JacksonFactory.getDefaultInstance
+  val HttpTransport = GoogleNetHttpTransport.newTrustedTransport
+
+  val DefaultCloudStorageConfiguration = {
+    val UploadBufferBytes = ConfigFactory.load().as[Option[Int]]("google.upload-buffer-bytes").getOrElse(MediaHttpUploader.MINIMUM_CHUNK_SIZE)
+
+    CloudStorageConfiguration.builder()
+      .blockSize(UploadBufferBytes)
+      .permitEmptyPathComponents(true)
+      .stripPrefixSlash(true)
+      .usePseudoDirectories(true)
+      .build()
+  }
+
+  def gcsStorage(applicationName: String,
+                 storageOptions: StorageOptions): Storage = {
+    new Storage.Builder(HttpTransport,
+      JsonFactory,
+      GoogleConfiguration.withCustomTimeouts(TransportOptions.getHttpRequestInitializer(storageOptions)))
+      .setApplicationName(applicationName)
+      .build()
+  }
+
+  def gcsStorage(applicationName: String, credentials: Credentials): Storage = {
+    gcsStorage(applicationName, gcsStorageOptions(credentials))
+  }
+
+  def gcsStorageOptions(credentials: Credentials,
+                        project: Option[String] = None,
+                        retrySettings: Option[RetrySettings] = None): StorageOptions = {
+    val storageOptionsBuilder = StorageOptions.newBuilder()
+      .setTransportOptions(TransportOptions)
+      .setCredentials(credentials)
+
+    retrySettings foreach storageOptionsBuilder.setRetrySettings
+    project foreach storageOptionsBuilder.setProjectId
+
+    storageOptionsBuilder.build()
+  }
+}
--- a/cromwell.examples.conf
+++ b/cromwell.examples.conf
@@ -596,6 +596,31 @@ services {
     #   flush-rate = 1 second # rate at which aggregated metrics will be sent to statsd
     # }
   }
+  HealthMonitor {
+    config {
+      # How long to wait between status check sweeps
+      # check-refresh-time = 5 minutes
+      # For any given status check, how long to wait before assuming failure
+      # check-timeout = 1 minute
+      # For any given status datum, the maximum time a value will be kept before reverting back to "Unknown"
+      # status-ttl = 15 minutes
+
+      ## When using the WorkbenchHealthMonitorServiceActor, the following are possibilities
+
+      # This *MUST* be set to the name of the PAPI (aka JES) backend one defined in the Backends stanza. Most likely
+      # it is "Jes" or "JES"
+      # papi-backend-name = JES
+
+      # The name of an authentication scheme to use for e.g. pinging PAPI and GCS. This should be either an application
+      # default or service account auth, otherwise things won't work as there'll not be a refresh token where you need
+      # them.
+      # google-auth-name = application-default
+
+      # A bucket in GCS to periodically stat to check for connectivity. This must be accessible by the auth mode
+      # specified by google-auth-name
+      # gcs-bucket-to-check = some-bucket-name
+    }
+  }
 }
 
 database {
--- a/dockerHashing/src/main/scala/cromwell/docker/local/DockerCliClient.scala
+++ b/dockerHashing/src/main/scala/cromwell/docker/local/DockerCliClient.scala
@@ -1,5 +1,6 @@
 package cromwell.docker.local
 
+import scala.Function.const
 import scala.util.Try
 
 /**
@@ -33,7 +34,7 @@ trait DockerCliClient {
     * @return Failure if an error occurs.
     */
   def pull(dockerCliKey: DockerCliKey): Try[Unit] = {
-    forRun("docker", "pull", dockerCliKey.fullName) { _ => () }
+    forRun("docker", "pull", dockerCliKey.fullName) { const(()) }
   }
 
   /**
@@ -79,7 +80,7 @@ trait DockerCliClient {
     * @param hashLine The line output by the stdout of the `lookupHash` command.
     * @return An optional `DockerCliHash`, if the all the columns are found.
     */
-  private[local] def parseHashLine(hashLine: String): Option[DockerCliHash] = {
+  private def parseHashLine(hashLine: String): Option[DockerCliHash] = {
     val none = "<none>"
     val tokens = hashLine.split("\t").lift
     for {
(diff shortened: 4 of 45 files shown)


cloudSupport/src/main/scala/cromwell/cloudSupport/gcp/gcs/GcsStorage.scala
```diff
@@ -0,0 +1,56 @@
...
+import com.google.api.gax.retrying.RetrySettings
+import com.google.cloud.storage.StorageOptions
+import com.google.api.services.storage.Storage
+import com.google.auth.Credentials
+import com.google.cloud.storage.contrib.nio.CloudStorageConfiguration
+import com.typesafe.config.ConfigFactory
+import cromwell.cloudSupport.gcp.GoogleConfiguration
+import cromwell.cloudSupport.gcp.http.GoogleHttpTransportOptions.TransportOptions
+import net.ceedubs.ficus.Ficus._
+
```
geoffjentry: This was really just moved around and has already faced an angry flock of seabirds but have at it

dockerHashing/src/main/scala/cromwell/docker/local/DockerCliClient.scala
```diff
@@ -33,7 +34,7 @@ trait DockerCliClient {
     * @return Failure if an error occurs.
     */
   def pull(dockerCliKey: DockerCliKey): Try[Unit] = {
-    forRun("docker", "pull", dockerCliKey.fullName) { _ => () }
+    forRun("docker", "pull", dockerCliKey.fullName) { const(()) }
   }
```
geoffjentry: I'd previously changed code around here and @mcovarr saw fit to comment on nearby code. I wound up removing my changes but left this in to prevent heartache for the next soul that modified something within a few lines of this

services/src/main/scala/cromwell/services/healthmonitor/impl/workbench/WorkbenchHealthMonitorServiceActor.scala
```diff
@@ -0,0 +1,83 @@
...
+
+    val genomicsFactory = GenomicsFactory(googleConfig.applicationName, papiAuthMode, endpointUrl)
+    val genomicsInterface = papiAuthMode.credential(WorkflowOptions.empty) map genomicsFactory.fromCredentials
+
+    genomicsInterface map { _.pipelines().list().setProjectId(papiProjectId).setPageSize(1).execute() } as OkStatus
+  }
+}
+
+object WorkbenchHealthMonitorServiceActor {
+  /*
```
geoffjentry: Change this to come from the config, and just leave a note that for multitenant type stuff that it should be a public read bucket

cloudSupport/src/main/scala/cromwell/cloudSupport/gcp/gcs/GcsStorage.scala
```diff
@@ -0,0 +1,56 @@
...
+object GcsStorage {
+  val JsonFactory = JacksonFactory.getDefaultInstance
+  val HttpTransport = GoogleNetHttpTransport.newTrustedTransport
+
+  val DefaultCloudStorageConfiguration = {
+    val UploadBufferBytes = ConfigFactory.load().as[Option[Int]]("google.upload-buffer-bytes").getOrElse(MediaHttpUploader.MINIMUM_CHUNK_SIZE)
```
danbills: ToL consider making this a class and taking `JsonFactory`, `HttpTransport`, and `Int`  as constructor params?

services/src/main/scala/cromwell/services/healthmonitor/HealthMonitorServiceActor.scala
```diff
@@ -0,0 +1,131 @@
...
+  val serviceConfig: Config
+  def subsystems: Set[MonitoredSubsystem]
+
+  implicit val ec: ExecutionContext = context.system.dispatcher
+
+  lazy val futureTimeout: FiniteDuration = serviceConfig.as[Option[FiniteDuration]]("services.HealthMonitor.check-timeout").getOrElse(DefaultFutureTimeout)
+  lazy val staleThreshold: FiniteDuration = serviceConfig.as[Option[FiniteDuration]]("services.HealthMonitor.status-ttl").getOrElse(DefaultStaleThreshold)
+
+  override def preStart(): Unit = {
+    val sweepTime = serviceConfig.as[Option[FiniteDuration]]("services.HealthMonitor.check-refresh-time").getOrElse(DefaultSweepTime)
```
danbills: as above. It'd be nice to take out `serviceConfig` and either declare some kind of config class or just declare each field as abstract.

services/src/main/scala/cromwell/services/healthmonitor/impl/workbench/WorkbenchHealthMonitorServiceActor.scala
```diff
@@ -0,0 +1,87 @@
...
+  private def checkGcs(): Future[SubsystemStatus] = {
+    import WorkbenchHealthMonitorServiceActor.GcsBucketToCheck
+
+    // For any expected production usage of this check, the GCS bucket should be public read */
+    val gcsBucketToCheck = serviceConfig.as[String]("services.HealthMonitor.config.GcsBucketToCheck")
+
+    val gcsAuthName = papiConfig.as[String]("filesystems.gcs.auth")
+
+    val cred = googleConfig.auth(gcsAuthName) match {
+      case Valid(a) => a.credential(WorkflowOptions.empty)
```
Horneth: Sorry to be so annoying with this -_- But in the current FC config (granted it's going to change soon when we get rid of refresh tokens), the `filesystems.gcs.auth` is going to be a "refresh token" auth mode, which will need a refresh token that it expects to get from workflow options, so I don't *think* this will work (I'm happy to give it a shot with a refresh token config if you want).
Basically since there is only one of those actors it cannot depend on a per workflow or per user type of auth.

I would suggest to just use application-default here since the bucket is expected to be public, or have a separate config stanza that defines what the auth for this check should be.
It will also have the added benefit to keep this GCS check working even without a JES backend configured.

services/src/main/scala/cromwell/services/healthmonitor/HealthMonitorServiceActor.scala
```diff
@@ -0,0 +1,131 @@
...
+
+  private def checkSubsystem(subsystem: MonitoredSubsystem): Unit = {
+    val result = subsystem.check()
+    val errMsg = s"Timed out after ${futureTimeout.toString} waiting for a response from ${subsystem.toString}"
+    result.withTimeout(futureTimeout, errMsg, context.system.scheduler)
+      .recover { case NonFatal(ex) =>
+        failedStatus(ex.getMessage)
+      } map {
+      Store(subsystem, _)
+    } pipeTo self
```
danbills: ToL curious why not call `store` here as opposed to this `pipeTo`?

services/src/main/scala/cromwell/services/healthmonitor/impl/common/DockerHubMonitor.scala
```diff
@@ -0,0 +1,49 @@
...
+  implicit val timeout = Timeout(5.seconds)
+
+  lazy val dockerHttpPool = Http().superPool[ContextWithRequest[DockerHashContext]]()
+  lazy val dockerHubFlow = List(new DockerHubFlow(dockerHttpPool)(ec, materializer, system.scheduler))
+  lazy val dockerHashActor = system.actorOf(DockerHashActor.props(dockerHubFlow, 500, 0.minutes, 0)(materializer), "DockerHashActor")
+
+  lazy val DockerHub = MonitoredSubsystem("DockerHub", checkDockerhub _)
+
+  /**
+    * Demonstrates connectivity to DockerHub by periodically pulling a small image. Remove the image afterwards in
```
Horneth: This comment seems to not reflect anymore what this is doing

README.md
```diff
@@ -4035,6 +4036,31 @@ Response:
 }
 ```
 
+## GET /engine/:version/status
+
+Cromwell will track the status of underlying systems on which it depends, typically connectivity to the database and to Dockerhub. The `status` endpoint will return the current status of these systems.
```
mcovarr: Pedantry alert: Docker Hub

cloudSupport/src/main/scala/cromwell/cloudSupport/gcp/gcs/GcsStorage.scala
mcovarr: Why a singleton class rather than an object?

cromwell.examples.conf
```diff
@@ -585,6 +585,26 @@ services {
...
+      # How long to wait between status check sweeps
+      # check-refresh-time = 5 minutes
+      # For any given status check, how long to wait before assuming failure
+      # check-timeout = 1 minute
+      # For any given status datum, the maximum time a value will be kept before reverting back to "Unknown"
+      # status-ttl = 15 minutes
+
+      ## When using the WorkbenchHealthMonitorServiceActor, the following are possibilities
+
+      # This *MUST* be set to the name of the PAPI aka JES) backend one defined in the Backends stanza. Most likely
```
mcovarr: missing a `(` for your `)`

services/src/main/scala/cromwell/services/healthmonitor/HealthMonitorServiceActor.scala
mcovarr: what happens if the failure is `Fatal`?

services/src/main/scala/cromwell/services/healthmonitor/HealthMonitorServiceActor.scala
geoffjentry: Talked about in person, but `serviceConfig` is part of the contract of the service registry (whether that's a good idea or not is debatable but I'd like to think of it as out of scope here)

services/src/main/scala/cromwell/services/healthmonitor/HealthMonitorServiceActor.scala
geoffjentry: The real answer is this block was taken verbatim from Rawls and any answer I'd give would be guessing on the intent.

services/src/main/scala/cromwell/services/healthmonitor/impl/workbench/WorkbenchHealthMonitorServiceActor.scala
geoffjentry: You mean have this implementation require the existence of an auth other than refresh token (e.g. app default)? If I understood you that seems like a good idea to me

Horneth: yes exactly

cromwell.examples.conf
geoffjentry: (oops)

cloudSupport/src/main/scala/cromwell/cloudSupport/gcp/gcs/GcsStorage.scala
geoffjentry: @mcovarr was that to me or to @danbills ?

mcovarr: @danbills

services/src/main/scala/cromwell/services/healthmonitor/HealthMonitorServiceActor.scala
geoffjentry: @mcovarr That's a good question. `pipeTo` will send a `Status.Failure` message which is not handled as far as I can tell. I don't *think* there's any underlying akka magic for that. Let me double check w/ @rtitle that there's not some slick way this is really being handled that I missed.

mcovarr: There doesn't seem to be anything in `receive` that would deal with an Akka `Failure`, and unless I've missed something that seems like a situation that could happen.

geoffjentry: @mcovarr That was exactly what I was just saying, but using different words. The other bit I was getting at is that it's possible there was another part to this which I missed and didn't bring across to Cromwell.

cloudSupport/src/main/scala/cromwell/cloudSupport/gcp/gcs/GcsStorage.scala
danbills: object doesn't take parameters.  You could stick to object if every function returned a `SomeConfigType => X`, i.e. taking the dependencies as function inputs.  I.e. dependency injection via `Reader` monad.

geoffjentry: Watching the convo unfold, since this was just a move from a different package I'm going to raise "activation energy" as a potential blocker for suggestions here :)

services/src/main/scala/cromwell/services/healthmonitor/HealthMonitorServiceActor.scala
geoffjentry: Have chatted with @rtitle - will be setting up an experiment to see what happens. I'd like to think that this is a non-issue as this feels like a real easy trap to fall into with pipeTo but early investigation was unclear

geoffjentry: Confirmed that nothing good happens here

geoffjentry: @danbills - @rtitle pointed out that the reason for the `pipeTo` was so that actor state wasn't being modified from inside a `Future` which IMO is a very good reason

mcovarr: That's an excellent reason which ought to be immortalized in a comment. 🙂

geoffjentry: @mcovarr A year ago I'd have said that that should be so obvious that no one would ever question it. And then this thread happened.

geoffjentry: @mcovarr So it turns out that my test was wrong. I was constructing a hard error with `Future.failed` which doesn't get caught by `NonFatal()`. With a true fatal error this does the rights thing and shuts things down.

That said ..... if a check was doing this intentionally (e.g. `Future.failed(new OutOfMemoryError())` it wouldn't get caught. I'd argue that this isn't worth the extra hijinks to protect against since even if this were to happen it'd be programmer hijinks and not a true fatal error.

cloudSupport/src/main/scala/cromwell/cloudSupport/gcp/gcs/GcsStorage.scala
geoffjentry: BTW that's not to say that I won't change anything, but come to a consensus easier than wiring `Reader` around :)

Horneth:
👍

Issue #2655, opened 2017-09-25
Make tmpDir configurable in reference.conf


psfoley:
In the latest releases of Cromwell (28+), the tmpDir variable is set to be created under the [current working directory of the workflow task](https://github.com/broadinstitute/cromwell/blob/master/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala#L186) by default. In a cluster environment, we have observed that this limits how quickly a task can execute when using a shared file system (NFS/Lustre/etc.) because these filesystems perform poorly with small reads and writes. In earlier versions of Cromwell, we worked around this by setting tmpDir within each task of the WDL to a path mounted on a local disk, but now the tmpDir variable seems to be set globally resulting in the local tmpDir variable getting overwritten. 

We can modify the tmpDir path in StandardAsyncExecutionActor, but this requires recompiling Cromwell  for each environment where Cromwell is deployed. Would it be possible to make the tmpDir variable configurable within reference.conf? That way a site admin can deploy Cromwell without needing to recompile, and quickly choose the tmpDir path that is optimal for their environment.

PR #2657, opened 2017-09-26
Enable configuration of the temporary directory.

kshakir:


cromwell.examples.conf
```diff
@@ -212,6 +212,17 @@ backend {
...
+        # `temporary-directory` creates the temporary directory for commands.
+        #
+        # If this value is not set explicitly, the default value creates a unique temporary directory, equivalent to:
+        # temporary-directory = "$(mktemp -d \"$PWD\"/tmp.XXXXXX)"
+        #
+        # The expression is run from the execution directory for the script. The expression must create the directory
+        # if it does not exist, and then return the full path to the directory.
+        #
+        # To create and return a constant temporary directory, use something like:
+        # temporary-directory = "$(mkdir -p /tmp/mydir && echo /tmp/mydir)"
```
mcovarr: Hopefully users wouldn't expect this to be visible across Docker invocations (at least without poking in new volume mounts)?

mcovarr: And perhaps a counterindication for Mac whose Docker I believe doesn't like volumes mounted in that are not under the home directory?

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -187,7 +190,12 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
 
     globFiles.map(globFiles =>
     s"""|#!/bin/bash
-        |tmpDir=$$(mktemp -d $cwd/tmp.XXXXXX)
+        |tmpDir=$$(
+        |  set -e
+        |  cd $cwd
+        |  tmpDir="$temporaryDirectory"
+        |  echo "$$tmpDir"
```
Horneth: does this print to stdout ?

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -98,6 +98,9 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
 
   lazy val scriptEpilogue = configurationDescriptor.backendConfig.as[Option[String]]("script-epilogue").getOrElse("sync")
 
+  lazy val temporaryDirectory = configurationDescriptor.backendConfig
+    .as[Option[String]]("temporary-directory").getOrElse("""$(mktemp -d "$PWD"/tmp.XXXXXX)""")
```
danbills: ToL I realizing I'm getting to the point of annoying belaboring this on separate PR's, but it would really be nice if this were a field on `BackendConfigurationDescriptor` rather than looking it up w/ the knowlege of a type, special key name, and the `orElse` value.

cromwell.examples.conf
kshakir: Oops, I didn't mean constant as in reusable, but constant as in a non-random path. The example above was meant to ensure that config editors realize when not using the default `mktemp` that they must still create the directory iff it doesn't exist, and return the path as a string. I'll edit the sentence.

Most folks should not need to edit this config value. See the original ticket #2655 for the use cases.

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
kshakir: In `$()`, and the deprecated ` `` ` syntax, stdout is captured while stderr is printed.

http://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_03

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
kshakir: Good points. Is there a ticket that's tracking the improvement? This PR should be linked as an example.

PR #2664, opened 2017-09-28
comment for posterity

geoffjentry:
I'm not actually going to wait for :+1: but will wait for it to go green. My claim will be out of safety but really it's because I'm lazy and will do it later.


mcovarr:
thumbing this anyway! :+1:

geoffjentry:
Thank you @mcovarr - your support means the world.

PR #2665, opened 2017-09-28
Running pre/post script in cwd.

kshakir:


backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -209,9 +212,15 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
...
+        |(
+        |cd $cwd
         |SCRIPT_EPILOGUE
+        |)
         |mv $rcTmpPath $rcPath
-        |""".stripMargin.replace("INSTANTIATED_COMMAND", instantiatedCommand).replace("SCRIPT_EPILOGUE", scriptEpilogue))
+        |""".stripMargin
+      .replace("COMMAND_SCRIPT_PREAMBLE", commandScriptPreamble)
+      .replace("INSTANTIATED_COMMAND", instantiatedCommand)
+      .replace("SCRIPT_EPILOGUE", scriptEpilogue))
```
Horneth: nitpick, feel free to ignore: can the preamble and epilogue be named the same way ?
either `COMMAND_SCRIPT_PREAMBLE` and `COMMAND_ SCRIPT_EPILOGUE` or `SCRIPT_PREAMBLE` and `SCRIPT_EPILOGUE`

kshakir: Will do, but just a heads up that they are very different while the names will make them look similar.

Currently:
- The preamble is only wired for the occasional jes monitoring script, per workflow-options.
- The epilogue is used at some sites(?) for disabling `sync`, per backend config.

Redmar does have further [plans](https://gatkforums.broadinstitute.org/wdl/discussion/comment/42753/#Comment_42753) for the epilogue. We can always iterate as we flush out use cases.

PR #2667, opened 2017-10-02
A develop Cromwell to match develop wdl4s

mcovarr:
Mostly package movement with a splash of real API changes.

### DIFF
--- a/engine/src/test/scala/cromwell/SimpleWorkflowActorSpec.scala
+++ b/engine/src/test/scala/cromwell/SimpleWorkflowActorSpec.scala
@@ -90,7 +90,7 @@ class SimpleWorkflowActorSpec extends CromwellTestKitWordSpec with BeforeAndAfte
     }
 
     "fail to construct with missing inputs" in {
-      val expectedError = "Required workflow input 'wf_hello.hello.addressee' not specified."
+      val expectedError = "Required workflow input 'wf_hello.hello.addressee' not specified"
       val failureMatcher = FailureMatcher(expectedError)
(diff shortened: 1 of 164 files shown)


engine/src/test/scala/cromwell/MetadataWatchActor.scala
```diff
@@ -18,6 +18,7 @@ final case class MetadataWatchActor(promise: Promise[Unit], matchers: Matcher*)
 
   override def receive = {
     case PutMetadataAction(events) if unsatisfiedMatchers.nonEmpty =>
+      println(events.map(_.value.map(_.value)))
```
danbills: I'm curious if this was intentionally added?

mcovarr: I think this was @Horneth ?

Horneth: it was intentionally added but unintentionally not removed 😄 
It was to help me debug some failing tests, it can be reverted

engine/src/test/scala/cromwell/SimpleWorkflowActorSpec.scala
```diff
@@ -107,7 +107,7 @@ class SimpleWorkflowActorSpec extends CromwellTestKitWordSpec with BeforeAndAfte
     }
 
     "fail to construct with inputs of the wrong type" in {
-      val expectedError = "Could not coerce JsNumber value for 'wf_hello.hello.addressee' (3) into: WdlStringType"
+      val expectedError = "No coercion defined from '3' of type 'spray.json.JsNumber' to 'String'."
```
Horneth: it would have been nice to keep the name of the offending input here :/ I probably unintentionally changed that in the last wdl4s PR 😞

PR #2668, opened 2017-10-02
Build correct CWL Namespace

Horneth:


engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -444,7 +444,8 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
         wf <- fromEither[IO](cwl.select[Workflow].toRight(NonEmptyList.one(s"expected a workflow but got a $cwl")))
         executable <-  fromEither[IO](wf.womExecutable(Option(source.inputsJson)))
         graph <- fromEither[IO](executable.graph.toEither)
-      } yield ValidatedWomNamespace(executable, graph, Map.empty)
+        validatedWomNamespace <- fromEither[IO](validateWomNamespace(executable))
```
mcovarr: would this be better as a method on `Executable`?

Horneth: By this do you mean this whole flatMap or `validateWomNamespace` ?

mcovarr: `validateWomNamespace`

Horneth: I think it's on the line between WOM and Cromwell. I created this `ValidatedWomNamespace` thing because of how the WMDA builds the workflow descriptor later so in some sense it's somewhat Cromwell specific but I could be convinced either way.
There is something checking that gcs files look right in it which should stay in Cromwell though IMO

PR #2676, opened 2017-10-04
User service account auth mode. Closes #2669

### LINKED ISSUE #2669: Allow for user provided service accounts
For now treat this like refresh tokens, e.g. encrypted, deleted afterwards, etc.

user provides json key via workflow options.

geoffjentry:
Opened a [Centaur ticket](https://github.com/broadinstitute/centaur/issues/234) to provide centaur testing for this, but since this won't be used in anger for a while punting for now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@ Abort of Dockerized tasks on the Local backend should now work as expected as `d
 * **StatsD**
 Added initial support for StatsD instrumentation. See the [README](https://github.com/broadinstitute/cromwell#statsd) for details on how to use it.
 
+* **User Service Account auth mode for Google**
+Added a new authentication mode for Google Cloud Platform which will allow a user to supply the JSON key file in their workflow options to allow for per-workflow authentication via service account. This is analogous to the previously existing refresh token authentication scheme. As with the refresh token scheme it is encouraged that the **user_service_account_json** workflow option field is added to the **encrypted-fields** list in the configuration.
+
 ## 29 Release Notes
 
 ### Breaking Changes
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ A [Workflow Management System](https://en.wikipedia.org/wiki/Workflow_management
       * [Application Default Credentials](#application-default-credentials)
       * [Service Account](#service-account)
       * [Refresh Token](#refresh-token)
+      * [User Service Account](#user-service-account)
     * [Docker](#docker)
     * [Monitoring](#monitoring)
   * [GA4GH TES Backend](#ga4gh-tes-backend)
@@ -948,7 +949,8 @@ authentication schemes that might be used:
 * `application_default` - (default, recommended) Use [application default](https://developers.google.com/identity/protocols/application-default-credentials) credentials.
 * `service_account` - Use a specific service account and key file (in PEM format) to authenticate.
 * `user_account` - Authenticate as a user.
-* `refresh_token` - Authenticate using a refresh token supplied in the workflow options.
+* `refresh_token` - Authenticate each individual workflow using a refresh token supplied in the workflow options.
+* `user_service_account` - Authenticate each individual workflow using service account credentials supplied in the workflow options.
 
 The `auths` block in the `google` stanza defines the authorization schemes within a Cromwell deployment:
 
@@ -971,6 +973,10 @@ google {
       scheme = "service_account"
       service-account-id = "my-service-account"
       pem-file = "/path/to/file.pem"
+    },
+    {
+      name = "user-service-account"
+      scheme = "user_service_account"
     }
   ]
 }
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,9 @@ Abort of Dockerized tasks on the Local backend should now work as expected as `d
 * **StatsD**
 Added initial support for StatsD instrumentation. See the [README](https://github.com/broadinstitute/cromwell#statsd) for details on how to use it.
 
+* **User Service Account auth mode for Google**
+Added a new authentication mode for Google Cloud Platform which will allow a user to supply the JSON key file in their workflow options to allow for per-workflow authentication via service account. This is analogous to the refresh token authentication scheme except using a service account credentials instead of user credentials. As with the refresh token scheme it is highly encouraged that the **user_service_account_json** workflow option field is added to the **encrypted-fields** list in the configuration.
```
katevoss: Could you add a link to the User Service Account section of the Readme docs, so if users want more details they don't have to search?

README.md
```diff
@@ -1017,14 +1023,20 @@ Creating the account will cause the JSON file to be downloaded.  The structure o
...
 
 A **refresh_token** field must be specified in the [workflow options](#workflow-options) when submitting the job.  Omitting this field will cause the workflow to fail.
 
 The refresh token is passed to JES along with the `client-id` and `client-secret` pair specified in the corresponding entry in `auths`.
 
+#### User Service Account
+
+A [JSON key file for the service account](####service-acocunt) must be passed in via the **user_service_account_json** field in the [workflow options](#workflow-options) when submitting the job. Omitting this field will cause the workflow to fail. The JSON should be passed as a string and must have no newlines and all instances of *"* and *\n* escaped. 
+
+In the likely event that this service account does not have access to Cromwell's default google project the **google_project** workflow option must be set. In the similarly likely case that this service account can not access Cromwell's defualt google bucket, the **jes_gcs_root** workflow option should be set appropriately.
```
Horneth: defualt -> default

geoffjentry: @Horneth I almost removed the `google_compute_service_account` option but left it in for two reasons - basically it's not *quite* the same thing (and if the old thing works for you it's technically a superior option, it just is very limited) and also in case anyone is using it.

cloudSupport/src/main/scala/cromwell/cloudSupport/gcp/auth/GoogleAuthMode.scala
```diff
@@ -118,6 +120,23 @@ final case class ServiceAccountMode(override val name: String,
...
+
+  private def extractServiceAccount(options: WorkflowOptions): String = {
+    options.get(UserServiceAccountKey) getOrElse {
+      throw new IllegalArgumentException(s"Missing parameters in workflow options: $UserServiceAccountKey")
+    }
+  }
+
+  override def credential(options: WorkflowOptions)(implicit as: ActorSystem, ec: ExecutionContext): Future[Credentials] = {
+    Future {
+      ServiceAccountCredentials.fromStream(new ByteArrayInputStream(extractServiceAccount(options).getBytes)).createScoped(scopes)
```
kshakir: Very very ToL:

A safer version of this specific line is `extractServiceAccount(options).getBytes("UTF-8")`, as `ServiceAccountCredentials` is not using the platform default encoding, but is specifying that the bytes [must be UTF-8](https://github.com/google/google-auth-library-java/blob/v0.7.0/oauth2_http/java/com/google/auth/oauth2/ServiceAccountCredentials.java#L308-L309).

This line will work on 99.999% of systems. But every once in a while, you get someone who's set their default String encoding to something ~~weird~~ different. Generally I try to keep String<->Bytes conversions using the platform default encoding, or _always_ using UTF-8. This line mixes them.

That said, anyone who switches the default encoding on their system just might end up with a ton of other problems elsewhere in cromwell…

geoffjentry:
@kshakir agreed on the json as string but without having a new mechanism to pass stuff it was the only thing that would work. I figured that in the not too distant future people wouldn't be directly interacting with his auth mode so it would be ok

README.md
```diff
@@ -1017,14 +1023,20 @@ Creating the account will cause the JSON file to be downloaded.  The structure o
...
 
-While technically not part of Service Account authorization mode, one can also override the default service account that the compute VM is started with via the configuration option `JES.config.genomics.compute-service-account` or through the workflow options parameter `google_compute_service_account`.  It's important that this service account, and the service account specified in `JES.config.genomics.auth` can both read/write the location specified by `JES.config.root`
+While technically not part of Service Account authorization mode, one can also override the default service account that the compute VM is started with via the configuration option `JES.config.genomics.compute-service-account` or through the workflow options parameter `google_compute_service_account`.  The service account you provide must have granted Service Account Actor role to Cromwell's primary service account. As this only affects Google Pipelines API and not GCS, it's important that this service account, and the service account specified in `JES.config.genomics.auth` can both read/write the location specified by `JES.config.root`
```
mcovarr: must have granted => must have been granted
?

geoffjentry:
@Horneth I confirmed that the behavior w/ the compute SA is the same as the standard SA auth. Since my intention was to be "the same" but with per-workflow SA I think that's right here. One can always set the compute-service-account option(s)

PR #2680, opened 2017-10-06
Cromwell WOM update

Horneth:
Updates Cromwell with the latest changes from wdl4s
I was able to unignore half of the unit tests that were previously ignored, and re-enable 60 centaur tests (on local and TES).
I haven't looked at the JES failures yet, so this is not be merged until they're fixed, but I though I would make a PR a bit early as this re-wires a decent amount of stuff that was broken due to missing FQNs.


backend/src/test/scala/cromwell/backend/BackendSpec.scala
```diff
@@ -83,15 +82,23 @@ trait BackendSpec extends ScalaFutures with Matchers with Mockito {
...
-        val value: WdlValue = inputs.getOrElse(
-          optionalWithDefault.name, optionalWithDefault.default.evaluateValue(inputs, NoIoFunctionSet)
-          .getOrElse(fail(s"Can't evaluate input ${optionalWithDefault.name}")))
-        optionalWithDefault -> value
-      case optional: OptionalInputDefinition => optional -> inputs.getOrElse(optional.name, WdlOptionalValue.none(optional.womType))
-    } toMap
+    
+    val inputDeclarations: Map[InputDefinition, WdlValue] = call.inputDefinitionMappings.map {
+      case (inputDef, resolved) => inputDef -> 
+        resolved.select[WdlValue].orElse(
```
danbills: ToL worried about the `.get`s in here.

backend/src/main/scala/cromwell/backend/wdl/OutputEvaluator.scala
```diff
@@ -1,31 +1,56 @@
...
+    }
+    
+    def validateOutput(output: OutputDefinition): OutputResult[JobOutput] = for {
+      evaluated <- evaluateOutputExpression(output.expression)
+      coerced <- coerceOutputValue(evaluated, output.womType)
+      postProcessed <- fromEither[Try](postMapper(coerced).toErrorOr.toEither)
+    } yield JobOutput(postProcessed)
+    
+    jobDescriptor.call.callable.outputs.traverse[OutputResult, (String, JobOutput)]({ output =>
+      validateOutput(output) map { output.name -> _ }
```
danbills: consider traversing w/ `Validated` so you get all validation errors?  You can get Applicative for  `Try[Validated[_]]` for free if you explicitly pass in `Applicative[Try] compose Applicative[Validated]` to  `traverse`

danbills:
I'm curious why the "do not merge?"  The `OutputEvaluator` happens to intersect w/ the spike I'm working on.

Horneth:
@danbills because I want to fix the JES tests too and I know they're still broken for now

backend/src/main/scala/cromwell/backend/wdl/OutputEvaluator.scala
```diff
@@ -1,31 +1,56 @@
...
-      // TODO WOM: Should evaluateValue return a Future ?
-      // TODO WOM: Aggregate failures
-      output.name -> 
-        JobOutput(
-          output.expression.evaluateValue(knownValues, ioFunctions).flatMap(value => postMapper(value).toErrorOr)  
-          .getOrElse(throw new Exception("Output evaluation failed"))
-        )
-    } toMap
+    def evaluateOutputExpression(expression: WomExpression): OutputResult[WdlValue] = {
+      Try(expression.evaluateValue(knownValues, ioFunctions)) match {
```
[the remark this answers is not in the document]
Horneth: Yep, I think we might even want them to return an `EitherT`.
I've been thinking about it and I think it does make sense to differentiate between an error as a message (like "method foo is unknown, cannot evaluate this expression"), and an exception thrown while trying to read a file or do some actual IO for example. The same way you did with `Parse` for the CWL parsing.
I think we could aim for an `EitherT[IO, NonEmptyList[String], WdlValue]` or something like that

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -79,6 +78,13 @@ case class WorkflowExecutionActor(workflowDescriptor: EngineWorkflowDescriptor,
 
   when(WorkflowExecutionPendingState) {
     case Event(ExecuteWorkflowCommand, _) =>
+      // TODO WOM: Remove this when conditional and scatters are supported. It prevents the workflow from hanging
+      if(workflowDescriptor.namespace.innerGraph.nodes.collect({
```
mcovarr: ToL `collectFirst` to stop at the first one, but hopefully this is not long for this world. 😄

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -268,54 +274,47 @@ case class WorkflowExecutionActor(workflowDescriptor: EngineWorkflowDescriptor,
...
-    val responseAndState = ResponseAndFinalState(
-      WorkflowExecutionSucceededResponse(data.jobExecutionMap, unqualifiedWorkflowOutputs),
-      WorkflowExecutionSuccessfulState)
 
-
-    context.parent ! responseAndState.response
-    goto(responseAndState.finalState) using data
+    workflowDescriptor.namespace.innerGraph.outputNodes
+      .flatMap(_.outputPorts)
+      .map(op => op.identifier -> data.outputStore.get(op, None)).toList
```
mcovarr: looking a little bit forcompy

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -780,30 +779,47 @@ object WorkflowExecutionActor {
...
     }
 
-    def evaluate(lookup: Map[String, WdlValue], wdlFunctions: IoFunctionSet) = {
-      node.instantiatedExpression.expression.evaluateValue(lookup, wdlFunctions) flatMap { node.womType.coerceRawValue(_).toErrorOr }
+    def evaluate(lookup: Map[String, WdlValue], ioFunctions: IoFunctionSet) = {
+      instantiatedExpression.expression.evaluateValue(lookup, ioFunctions) flatMap { womType.coerceRawValue(_).toErrorOr }
     }
   }
 
+  case class IntermediateValueKey(node: ExpressionNode, index: ExecutionIndex) extends ExpressionKey {
```
mcovarr: I don't know what this is.

src/bin/travis/resources/funnel.conf
```diff
@@ -1,10 +1,10 @@
...
-  - Local:
-      AllowedDirs:
-        - /home/
-        - /cromwell-executions
-        - /tmp/
+  Local:
+    AllowedDirs:
+      - /home/
+      - /cromwell-executions
+      - /tmp/
```
mcovarr: hrm?

backend/src/test/scala/cromwell/backend/BackendSpec.scala
danbills: Ah!  Carry on 😄

src/bin/travis/resources/funnel.conf
Horneth: I don't know how it was working before but this conf was outdated, they changed the format and the `Storage` thing is an object not a list

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
Horneth: It's an expression that is not a workflow output. I took the name from what is in wdl4s but I can try to find something more meaningful

backend/src/main/scala/cromwell/backend/backend.scala
```diff
@@ -36,7 +35,10 @@ case class BackendJobDescriptor(workflowDescriptor: BackendWorkflowDescriptor,
                                 maybeCallCachingEligible: MaybeCallCachingEligible,
                                 prefetchedKvStoreEntries: Map[String, KvResponse]) {
   val fullyQualifiedInputs = inputDeclarations map { case (declaration, value) =>
-    key.call.fullyQualifiedName + "." + declaration -> value 
+    key.call.identifier.combine(declaration.name).fullyQualifiedName.value -> value 
+  }
+  val unqualifiedInputs = inputDeclarations map { case (declaration, value) =>
```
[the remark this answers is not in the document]
Horneth: yep :) there's likely more`unqualified`s everywhere that should be renamed

backend/src/main/scala/cromwell/backend/wdl/OutputEvaluator.scala
```diff
@@ -1,31 +1,56 @@
...
+import wom.callable.Callable.OutputDefinition
+import wom.expression.{IoFunctionSet, WomExpression}
 
-import scala.language.postfixOps
-import scala.util.{Success, Try}
+import scala.util.{Failure, Success, Try}
 object OutputEvaluator {
+  type OutputResult[A] = EitherT[Try, NonEmptyList[String], A]
+
+  implicit val composedApplicative = Applicative[Try] compose Applicative[ErrorOr]
```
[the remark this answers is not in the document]
danbills: This is needed to traverse w/ `Try[ErrorOr[_]]`.

Perhaps you could explicitly pass this into `traverse` and skip the implicit declaration to make it clearer?  

Could even take it further and skip the formal declaration as a `val`.

backend/src/main/scala/cromwell/backend/wdl/OutputEvaluator.scala
```diff
@@ -1,31 +1,56 @@
...
 import wdl.values.WdlValue
-import wom.expression.IoFunctionSet
+import wom.callable.Callable.OutputDefinition
+import wom.expression.{IoFunctionSet, WomExpression}
 
-import scala.language.postfixOps
-import scala.util.{Success, Try}
+import scala.util.{Failure, Success, Try}
 object OutputEvaluator {
+  type OutputResult[A] = EitherT[Try, NonEmptyList[String], A]
```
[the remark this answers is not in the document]
Horneth: Because `evaluateValue` returns an `ErrorOr`, but can also throw an exception, and we want to handle both. I think it makes sense to have both because if whatever we use in `readFile` throws an exception for example, we want to keep it including the stacktrace (could be a GCS exception for example)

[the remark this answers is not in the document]
Horneth: It's the same concept as what @danbills did with CWL Parsing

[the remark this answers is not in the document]
mcovarr: I'm thinking an ADT might make it more clear what's going on here rather than `Success(Left(...))` is one kind of failure and `Failure(..)` is a different kind of failure, but neither communicates to the reader what it means.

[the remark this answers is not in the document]
Horneth: The advantage of this is it can be `flatMap`ped easily. But I guess once it's been `flatMap`ped we could convert the result to an ADT

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -79,6 +78,13 @@ case class WorkflowExecutionActor(workflowDescriptor: EngineWorkflowDescriptor,
 
   when(WorkflowExecutionPendingState) {
     case Event(ExecuteWorkflowCommand, _) =>
+      // TODO WOM: Remove this when conditional and scatters are supported. It prevents the workflow from hanging
+      if(workflowDescriptor.namespace.innerGraph.nodes.collect({
+        case _: ScatterNode | _: ConditionalNode | _: WorkflowCallNode => true
+      }).nonEmpty) {
+        context.parent ! WorkflowExecutionFailedResponse(Map.empty, new Exception("Scatter and Conditional not supported yet"))
```
mcovarr: and workflow call

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
mcovarr: OK, or maybe just docs if that's easier. 🙂

mcovarr:
The only JES failure is `sizeenginefunction` which seems to be suffering from some file path chimerism. In my selfish desire to move forward with Cromwell unification, I'd be more than OK with disabling that temporarily and merging what's here. 😄

Horneth:
So I just realized this is not totally correct because it only traverses the outputs but it needs to foldLeft them because we can use output A in output B if A is declared before B.
I'll merge it anyway since @mcovarr is waiting on it and will make a ticket to fix that

backend/src/main/scala/cromwell/backend/wdl/OutputEvaluator.scala
```diff
@@ -18,22 +19,22 @@ import wom.expression.{IoFunctionSet, WomExpression}
 
 import scala.util.{Failure, Success, Try}
 object OutputEvaluator {
+  sealed trait EvaluatedJobOutputs
```
danbills: Let's tech talk this on Tuesday please.   My concern is that we're re-implementing the `Either` and `Try` effects.

PR #2674, opened 2017-10-04
[PREGULL] Unified Cromwell

mcovarr:
Definitely not done yet but hopefully getting close.

PR #2683, opened 2017-10-10
Fix output evaluation

Horneth:
This can wait for https://github.com/broadinstitute/cromwell/pull/2674 to be merged

### DIFF
--- a/backend/src/main/scala/cromwell/backend/backend.scala
+++ b/backend/src/main/scala/cromwell/backend/backend.scala
@@ -41,7 +41,7 @@ case class BackendJobDescriptor(workflowDescriptor: BackendWorkflowDescriptor,
     declaration.name -> value
   }
   val call = key.call
-  override val toString = s"${key.mkTag(workflowDescriptor.id)}"
+  override lazy val toString = s"${key.mkTag(workflowDescriptor.id)}"
 }
 
 object BackendWorkflowDescriptor {
--- a/backend/src/main/scala/cromwell/backend/backend.scala
+++ b/backend/src/main/scala/cromwell/backend/backend.scala
@@ -41,7 +41,7 @@ case class BackendJobDescriptor(workflowDescriptor: BackendWorkflowDescriptor,
     declaration.name -> value
   }
   val call = key.call
-  override lazy val toString = s"${key.mkTag(workflowDescriptor.id)}"
+  override lazy val toString = key.mkTag(workflowDescriptor.id)
 }
 
 object BackendWorkflowDescriptor {
--- a/backend/src/main/scala/cromwell/backend/wdl/OutputEvaluator.scala
+++ b/backend/src/main/scala/cromwell/backend/wdl/OutputEvaluator.scala
@@ -1,13 +1,12 @@
 package cromwell.backend.wdl
 
-import cats.Applicative
 import cats.data.EitherT._
 import cats.data.Validated.{Invalid, Valid}
 import cats.data.{EitherT, NonEmptyList}
-import cats.instances.list._
 import cats.instances.try_._
+import cats.syntax.apply._
 import cats.syntax.either._
-import cats.syntax.traverse._
+import cats.syntax.validated._
 import cromwell.backend.BackendJobDescriptor
 import cromwell.core.{CallOutputs, JobOutput}
 import lenthall.validation.ErrorOr.ErrorOr
@@ -15,7 +14,7 @@ import lenthall.validation.Validation._
 import wdl.types.WdlType
 import wdl.values.WdlValue
 import wom.callable.Callable.OutputDefinition
-import wom.expression.{IoFunctionSet, WomExpression}
+import wom.expression.IoFunctionSet
 
 import scala.util.{Failure, Success, Try}
 object OutputEvaluator {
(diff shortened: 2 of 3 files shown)


backend/src/main/scala/cromwell/backend/wdl/OutputEvaluator.scala
```diff
@@ -1,71 +1,82 @@
...
+       */
+      (for {
+        evaluated <- evaluateOutputExpression
+        coerced <- coerceOutputValue(evaluated, output.womType)
+        postProcessed <- fromEither[Try](postMapper(coerced).toErrorOr.toEither)
+        pair = output.name -> postProcessed
+      } yield pair).toValidated map { evaluatedOutput: ErrorOr[(String, WdlValue)] =>
+        (accumulated, evaluatedOutput) mapN {
+          case (validValues, validOutput) => validValues :+ validOutput
+        }
```
Horneth: suggestions to make this look better welcome

cjllanwarne: I think with `mapN` you shouldn't need the `case`?

cjllanwarne: does it work like `(accumulated, evaluatedOutput) mapN { _ :+ _ }`?

Horneth: yep that seems to work !

backend/src/main/scala/cromwell/backend/wdl/OutputEvaluator.scala
```diff
@@ -1,71 +1,80 @@
...
-        .leftMap(t => NonEmptyList.one(t.getMessage))
+    val emptyValue = Success(List.empty[(String, WdlValue)].validNel): Try[ErrorOr[List[(String, WdlValue)]]]
+
+    // Fold over the outputs to evaluate them in order, map the result to an EvaluatedJobOutputs
+    jobDescriptor.call.callable.outputs.foldLeft(emptyValue)(foldFunction) match {
+      case Success(Valid(outputs)) => ValidJobOutputs(
+        // Wrap the wdlValues in JobOutput objects
+        outputs.toMap map {
+          case (name, wdlValue) => name -> JobOutput(wdlValue)
+        }
```
danbills: is this a non-exhaustive match?

backend/src/main/scala/cromwell/backend/wdl/OutputEvaluator.scala
```diff
@@ -1,71 +1,80 @@
...
-    }
+    val knownValues = jobDescriptor.localInputs
     
-    def evaluateOutputExpression(expression: WomExpression): OutputResult[WdlValue] = {
-      EitherT { Try(expression.evaluateValue(knownValues, ioFunctions)).map(_.toEither) }
+    def foldFunction(accumulatedOutputs: Try[ErrorOr[List[(String, WdlValue)]]], output: OutputDefinition) = accumulatedOutputs flatMap { accumulated =>
+      // Extract the valid pairs from the job outputs accumulated so far, and add to it the inputs (outputs can also reference inputs)
+      val allKnownValues = accumulated match {
+        case Valid(outputs) => outputs.toMap[String, WdlValue] ++ knownValues
+        case Invalid(_) => knownValues
```
danbills: This logic seems fishy.  Seems like knownValues should be passed into initial state and then invalid is really invalid no?

backend/src/main/scala/cromwell/backend/wdl/OutputEvaluator.scala
```diff
@@ -1,71 +1,80 @@
...
+
+      /*
+        * Go through evaluation, coercion and post processing.
+        * Transform the result to a validated Try[ErrorOr[(String, WdlValue)]] with toValidated
+        * If we have a valid pair, add it to the previously accumulated outputs, otherwise combine the Nels of errors
+       */
+      (for {
+        evaluated <- evaluateOutputExpression
+        coerced <- coerceOutputValue(evaluated, output.womType)
+        postProcessed <- fromEither[Try](postMapper(coerced).toErrorOr.toEither)
```
danbills: Seems like `postMapper` should be mapped with a `Right` (then wrap the whole thing w/ `EitherT`).  This approach takes Try failures and puts them in a Left.  Is this intentional?

backend/src/main/scala/cromwell/backend/wdl/OutputEvaluator.scala
```diff
@@ -1,71 +1,80 @@
...
+      } yield pair).toValidated map { evaluatedOutput: ErrorOr[(String, WdlValue)] =>
+        (accumulated, evaluatedOutput) mapN { _ :+ _ }
+      }
     }
     
-    def coerceOutputValue(wdlValue: WdlValue, coerceTo: WdlType): OutputResult[WdlValue] = {
-      fromEither[Try](
-        coerceTo.coerceRawValue(wdlValue).toEither
-        .leftMap(t => NonEmptyList.one(t.getMessage))
+    val emptyValue = Success(List.empty[(String, WdlValue)].validNel): Try[ErrorOr[List[(String, WdlValue)]]]
```
danbills: Is `ErrorOr` still appropriate given that we're no longer `traverse`-ing?

backend/src/main/scala/cromwell/backend/wdl/OutputEvaluator.scala
```diff
@@ -1,71 +1,80 @@
...
+
+    // Fold over the outputs to evaluate them in order, map the result to an EvaluatedJobOutputs
+    jobDescriptor.call.callable.outputs.foldLeft(emptyValue)(foldFunction) match {
```
mcovarr: ToL so much IntelliJ red here that seems bogus to me 😢

backend/src/main/scala/cromwell/backend/wdl/OutputEvaluator.scala
Horneth: If we pass `knownValues` (which contains input values) in the initial state then it will also end up in the final map, and we only want the outputs.

backend/src/main/scala/cromwell/backend/wdl/OutputEvaluator.scala
Horneth: `evaluateValues` still returns an `ErrorOr` so if we want to accumulate multiple potential errors from all outputs we still need it no ?

backend/src/main/scala/cromwell/backend/wdl/OutputEvaluator.scala
Horneth: Hmm I don't think so ? What `case` is missing ?

backend/src/main/scala/cromwell/backend/backend.scala
```diff
@@ -41,7 +41,7 @@ case class BackendJobDescriptor(workflowDescriptor: BackendWorkflowDescriptor,
     declaration.name -> value
   }
   val call = key.call
-  override val toString = s"${key.mkTag(workflowDescriptor.id)}"
+  override lazy val toString = s"${key.mkTag(workflowDescriptor.id)}"
```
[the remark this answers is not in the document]
Horneth: no idea

backend/src/main/scala/cromwell/backend/wdl/OutputEvaluator.scala
Horneth: oh I don't have any red here : /

backend/src/main/scala/cromwell/backend/wdl/OutputEvaluator.scala
danbills: I put the comment on the wrong line, but it looks like you are missing `Success(Invalid(` and `Failure`.

danbills: oh my, I see them.  sorry my bad

backend/src/main/scala/cromwell/backend/wdl/OutputEvaluator.scala
danbills: yes I see `mapN`, carry on!  I'm 0/2 on this PR :)

backend/src/main/scala/cromwell/backend/wdl/OutputEvaluator.scala
Horneth: Do you mean `EitherT { postMapper(coerced).map(_.asRight[NonEmptyList[String]]) }` ?

Horneth:
@danbills agree with the `IO` standardization. I'd rather not do it in this PR but I'll make a ticket

backend/src/main/scala/cromwell/backend/wdl/OutputEvaluator.scala
danbills: yes 

(to keep the exceptions from postMapper in the Try )

PR #2686, opened 2017-10-11
Add the return code to JES error messages.

kshakir:


supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
```diff
@@ -56,8 +56,14 @@ object JesAsyncBackendJobExecutionActor {
   val JesUnexpectedTermination = 13
   val JesPreemption = 14
 
-  def StandardException(errorCode: Status, message: String, jobTag: String) = {
-    new Exception(s"Task $jobTag failed. JES error code ${errorCode.getCode.value}. $message")
+  def StandardException(errorCode: Status, message: String, jobTag: String, returnCodeOption: Option[Int]) = {
+    val returnCodeMessage = returnCodeOption match {
+      case Some(returnCode) if returnCode == 0 => " Job exited without an error, exit code 0."
+      case Some(returnCode) => s" Job exit code $returnCode. Check the stderr for more information."
```
Horneth: ToL : If you have access to jobPaths we could also put the path of stderr in the message

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
```diff
@@ -56,8 +56,18 @@ object JesAsyncBackendJobExecutionActor {
...
-    new Exception(s"Task $jobTag failed. JES error code ${errorCode.getCode.value}. $message")
+  def StandardException(errorCode: Status,
+                        message: String,
+                        jobTag: String,
+                        returnCodeOption: Option[Int],
+                        stderrPath: Path) = {
+    val returnCodeMessage = returnCodeOption match {
+      case Some(returnCode) if returnCode == 0 => " Job exited without an error, exit code 0."
+      case Some(returnCode) => s" Job exit code $returnCode. Check $stderrPath for more information."
+      case None => ""
```
cjllanwarne: Actually I'd say that this could be useful information (`"the job was stopped before the command finished"`). In fact I'd guess this scenario could be the majority of exceptional cases.

cjllanwarne: Side benefit of adding text here would be that we can move the space (` `) to the containing string. Woohoo!

PR #2689, opened 2017-10-11
Correctly interpolate optionals into strings

cjllanwarne:


cjllanwarne:
@mcovarr see new test. Before this change they fail because they were inserting extra `"`s:
```
Expected :WdlString(someStr)
Actual   :WdlString("someStr")
```

PR #2734, opened 2017-10-11
Hotfix support for public http-based imports

kcibul:
* disabled build failures due to deprecation warnings... it's a hotfix, deadend branch!
* support public http-based imports in cromwell
* fixed a few necessary classes due to newer cats being pulled in


Horneth:
LGTM ! 👍

PR #2745, opened 2017-10-13
CWL Expressions & Spike wrapup

danbills:
* Introduced `WorkflowStepInputExpression`.
* Standardized FullyQualifiedName to handle the separate cases in a more coherent way.
* Introduced `CwlExpressionCommandPart`.

I've not written tests for these new expression as I was focused on getting 3step running.  I am open to testing suggestions but request that we merge first and I pick that up when I'm back from vaca.


danbills:
Hm it appears using the web editor does a merge instead of rebase.  I'm hoping squash paves over this mistake.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -476,7 +477,7 @@ case class WorkflowExecutionActor(workflowDescriptor: EngineWorkflowDescriptor,
   private def resolve(jobKey: JobKey, data: WorkflowExecutionActorData)(outputPort: OutputPort): ErrorOr[WdlValue] = {
     data.outputStore.get(outputPort, jobKey.index).map(_.validNel) orElse {
       workflowDescriptor.defaultExpressions.get(outputPort) map {
-        case expr if expr.inputs.isEmpty => expr.evaluateValue(Map.empty, data.expressionLanguageFunctions)
+        case expr if expr.inputs.isEmpty => println("\n\nexpression inputs are empty!!!\n\n"); expr.evaluateValue(Map.empty, data.expressionLanguageFunctions)
```
kshakir: Change to a log message or delete the `println`.

geoffjentry:
images

danbills:
For the record I successfully ran 3step w/ pattern input specified.

cwl/src/main/scala/cwl/WorkflowStep.scala
```diff
@@ -48,7 +46,7 @@ case class WorkflowStep(
 
   def fileName: Option[String] = run.select[String]
 
-  val unqualifiedStepId = WomIdentifier(Try(WorkflowStepId(id)).map(_.stepId).getOrElse(id))
+  val unqualifiedStepId = WomIdentifier(Try(FullyQualifiedName(id)).map(_.id).getOrElse(id))
```
cjllanwarne: This seems a little smelly.... can we have an alternative `FullyQualifiedName.tryConstruct(id)` that returns the `Try` directly?

I just really don't like exceptions from constructors

cwl/src/test/resources/three_step.cwl
```diff
@@ -1,47 +1,46 @@
 cwlVersion: v1.0
 class: Workflow
 inputs:
-- id: file:///Users/danb/wdl4s/r.cwl#pattern
+- id: pattern
```
cjllanwarne: FMI, I'm assuming this only works because we re-SALAD this file before processing?

cwl/src/test/scala/cwl/CommandOutputExpressionSpec.scala
```diff
@@ -11,30 +11,49 @@ import eu.timepit.refined.string.MatchesRegex
...
 
 class CommandOutputExpressionSpec extends FlatSpec with Matchers {
 
   behavior of "CommandOutputExpression"
 
+  def ioFunctionSet(data: String) =
+    new IoFunctionSet {
+      override def readFile(path: String) = Future.successful(data)
+
+      override def writeFile(path: String, content: String) = ???
```
cjllanwarne: I'd prefer real throws here (ala "this method should never have been called by this test"), to show that this is the final state of the test case

cwl/src/test/scala/cwl/CwlWorkflowWomSpec.scala
```diff
@@ -170,19 +170,21 @@ class CwlWorkflowWomSpec extends FlatSpec with Matchers with TableDrivenProperty
...
-      "file:///Users/danb/wdl4s/r.cwl#cgrep-count",
-      "file:///Users/danb/wdl4s/r.cwl#wc-count"
+    nodes collect { case gon: GraphOutputNode => FullyQualifiedName(gon.localName).id } should be(Set(
+      "cgrep-count",
+      "wc-count"
     ))
 
     nodes collect { case cn: CallNode => cn.localName } should be(Set("ps", "cgrep", "wc"))
 
+    //def stepAndId(in: String) = s"${FileStepAndId(in).stepId}/${FileStepAndId(in).id}"
```
cjllanwarne: Remove

wom/src/main/scala/wdl/util/JsUtil.scala
```diff
@@ -80,6 +80,7 @@ object JsUtil {
       case WdlFloat(double) => double.asInstanceOf[java.lang.Double]
       case WdlBoolean(boolean) => boolean.asInstanceOf[java.lang.Boolean]
       case WdlArray(_, array) => array.map(toJavascript).toArray
+      case WdlSingleFile(path) => path
```
cjllanwarne: Huh, I'm curious where exactly this is being used (and why we never stuck problems with no "File to javascript" conversion in the past)

cjllanwarne: And if this is being used somewhere important, can we add `WdlPair` and non-empty `WdlOptional`s in too?

wom/src/main/scala/wom/expression/WomExpression.scala
```diff
@@ -14,13 +20,33 @@ trait WomExpression {
...
+      linkedInputs = linkedInputList.toMap
+      inputTypes = linkedInputs map { case (k, v) => k -> v.womType }
+      evaluatedType <- evaluateType(inputTypes)
+    } yield new InstantiatedExpression(this, evaluatedType, linkedInputs)
+  }
 }
 
 final case class PlaceholderWomExpression(inputs: Set[String], fixedWomType: WdlType) extends WomExpression {
   override def sourceString: String = "placeholder"
-  override def evaluateValue(inputValues: Map[String, WdlValue], ioFunctionSet: IoFunctionSet): ErrorOr[WdlValue] = Valid(WdlString("42"))
```
cjllanwarne: Did this not break tests?

and if this didn't... are we actually using `PlaceholderWomExpression` anywhere any more?

wom/src/main/scala/wom/graph/InstantiatedExpression.scala
```diff
@@ -1,15 +1,11 @@
...
-import cats.syntax.validated._
-import cats.data.Validated.Valid
 import lenthall.validation.ErrorOr.ErrorOr
 import wdl.types.WdlType
 import wom.expression.WomExpression
-import wom.graph.GraphNodePort.{ConnectedInputPort, InputPort, OutputPort}
+import wom.graph.GraphNodePort.{InputPort, OutputPort}
 
-class InstantiatedExpression private(val expression: WomExpression, val womReturnType: WdlType, val inputMapping: Map[String, InputPort]) {
+class InstantiatedExpression (val expression: WomExpression, val womReturnType: WdlType, val inputMapping: Map[String, InputPort]) {
```
cjllanwarne: Hmm... I had this private to avoid people (by which I mean mostly myself) accidentally constructing invalid ones of these (e.g. by calling the constructor without validating the mapping). I *think* I'd prefer to have the verification here locally, rather than making this public

wom/src/main/scala/wom/graph/InstantiatedExpression.scala
```diff
@@ -19,26 +15,10 @@ object InstantiatedExpression {
...
-
-    import lenthall.validation.ErrorOr.ShortCircuitingFlatMap
-    for {
-      linkedInputList <- expression.inputs.toList traverse linkInput
-      linkedInputs = linkedInputList.toMap
-      inputTypes = linkedInputs map { case (k, v) => k -> v.womType }
-      evaluatedType <- expression.evaluateType(inputTypes)
-    } yield new InstantiatedExpression(expression, evaluatedType, linkedInputs)
-  }
 }
```
cjllanwarne: While you're here, might as well add a newline

cwl/src/main/scala/cwl/CommandOutputBinding.scala
```diff
@@ -63,9 +66,10 @@ case class CommandOutputBinding(
 
   private def load64KiB(path: String, ioFunctionSet: IoFunctionSet): String = {
     // This suggests the IoFunctionSet should have a length-limited read API as both CWL and WDL support this concept.
-    // val content = ioFunctionSet.readFile(path)
-    val content = better.files.File(path).bytes.take(64 * 1024).toArray
-    new String(content)
+    val content = ioFunctionSet.readFile(path)
+
+    //TODO: propagate IO, Try, or Future or something all the way out via "commandOutputBindingtoWdlValue" signature
+    Await.result(content, 5 seconds)
```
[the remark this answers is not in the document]
danbills: Yeah this needs to be revisited.  I will truncate at 64k for now to be compliant w/ spec.

cwl/src/main/scala/cwl/CwlExpressionCommandPart.scala
```diff
@@ -19,3 +20,25 @@ case class CwlExpressionCommandPart(expr: Expression) extends CommandPart {
...
+      case (LocalName(localName), WdlSingleFile(path)) => localName -> WdlString(path)
+      case (LocalName(localName), value) => localName -> value
+    }), functions)
+
+    val wdlValue: WdlValue= argument match {
+      case CommandLineBinding(_, _, _, _, _, Some(Inl(expression: Expression)), Some(false)) =>
+        expression.fold(EvaluateExpression).apply(pc)
+      case CommandLineBinding(_, _, _, _, _, Some(Inr(Inl(string: String))), Some(false)) =>
+        WdlString(string)
+      case _ => ???
```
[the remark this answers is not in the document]
danbills: Given the signature, I think that `NotImplementedYet` is pretty accurate, no?  

I think perhaps these functions should return `Either`, this is an issue that is been accumulating technical debt, I will revisit when I get back

cwl/src/main/scala/cwl/CwlWomExpression.scala
```diff
@@ -15,37 +20,67 @@ sealed trait CwlWomExpression extends WomExpression {
...
   override def sourceString = outputBinding.toString
   override def evaluateValue(inputValues: Map[String, WdlValue], ioFunctionSet: IoFunctionSet): ErrorOr[WdlValue] = {
     val parameterContext = ParameterContext.Empty.withInputs(inputValues, ioFunctionSet)
 
-    val wdlValue = outputBinding.commandOutputBindingToWdlValue(parameterContext, ioFunctionSet)
-    cwlExpressionType.coerceRawValue(wdlValue).toErrorOr
+    val wdlValue: WdlValue = outputBinding.commandOutputBindingToWdlValue(parameterContext, ioFunctionSet)
+    val extractFile: WdlValue =
+      wdlValue match {
+        case WdlArray(WdlMaybeEmptyArrayType(WdlMapType(WdlStringType, WdlStringType)), Seq(WdlMap(WdlMapType(WdlStringType, WdlStringType), map))) =>
```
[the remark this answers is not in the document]
danbills: Right now all inputs to expressions are converted to parameter context by creating a map of a single type.

I have ticket #2748 out to accept multiple input types, which will then affect this pattern match.

cwl/src/main/scala/cwl/CwlWomExpression.scala
```diff
@@ -15,37 +20,67 @@ sealed trait CwlWomExpression extends WomExpression {
...
 
-  override def inputs: Set[String] = ???
-
   /*
   TODO:
    DB: It doesn't make sense to me that this function returns type WdlFile but accepts a type to which it coerces.
    Wouldn't coerceTo always == WdlFileType, and if not then what?
    */
-  override def evaluateFiles(inputTypes: Map[String, WdlValue], ioFunctionSet: IoFunctionSet, coerceTo: WdlType): ErrorOr[Set[WdlFile]] ={
+  override def evaluateFiles(inputs: Map[String, WdlValue], ioFunctionSet: IoFunctionSet, coerceTo: WdlType): ErrorOr[Set[WdlFile]] ={
```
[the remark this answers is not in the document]
danbills: > whether it's about to be coerced into an Array[File] - in which case we evaluate 2 files to delocalize

I'm confused.  Is this function actually doing the delocalizing?  And why wouldn't an array of strings be treated the same way as Files?  

Could this function's name be improved?

cwl/src/main/scala/cwl/WorkflowStep.scala
danbills: It is smelly indeed!  I'm not really sure why it's within a `Try` or why it could throw exceptions.

I would prefer to call `WorkflowStep`'s `id` field `fullyQualifiedId` or something similar but it would deviate from the spec name.  I guess we could change `FullyQualifiedName`'s `id` field to something better but I don't know what that better name is.

cwl/src/test/resources/three_step.cwl
danbills: yep

cwl/src/test/scala/cwl/CommandOutputExpressionSpec.scala
danbills: ok. 

for the record I think a better solution here is to pass in a `readFile: String => IO[String]` function to `evaluateValue` which would make this mock unnecessary.

wom/src/main/scala/wdl/util/JsUtil.scala
danbills: This code should probably be in the CWL package.

It is being used to convert inputs to javascript data.  I don't think we've ever done this in the past, but I think you know this so I must be misunderstanding your point...

CWL don't support pairs nor optionals.

wom/src/main/scala/wom/graph/InstantiatedExpression.scala
danbills: Ok I had moved to bring `linkWithInputs` to `WomExpression`, but I'll put it back

wom/src/main/scala/wom/graph/InstantiatedExpression.scala
danbills: I'm not trying to start a bike shed discussion, but why does it matter?

geoffjentry: It's a @kshakir hobby horse because if you cat a file it doesn't put the prompt on its own line.

something something autoformatting something something

cwl/src/main/scala/cwl/CwlWomExpression.scala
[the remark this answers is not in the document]
cjllanwarne: This function works out all of the files which will be needed in order to evaluate the expression.

Cromwell uses this to work out what to tell JES to delocalize (which it has to do before the task starts running)

wom/src/main/scala/wdl/util/JsUtil.scala
cjllanwarne: Oh right, I didn't realize this was a CWL specific thing (which explains why we never hit it before)! 

I asked about non-empty optionals because there actually is a handler for empty optionals in this chain

cjllanwarne:
Closing in favour of #2761

PR #2749, opened 2017-10-14
* lost a few hours debugging this test, made it more robust

kcibul:
* if you happen to have a 3step.wdl[options or inputs] in your local working directory, this test fails because it expects NOT to find them there.  This isn't so crazy since (a) 3step is real workflow name and (b) people might (as I did) run it to test out cromwell from their local working directory


geoffjentry:
@mcovarr for something like this i'd imagine most of us have a standard workflow (har har har) that we tend to follow which doesn't turn up stuff like this

:+1:

kcibul:
That's why you need some chump like me bumbling around ;)

I'm getting a 'X' on codecov because there is no previous version to compare to.  I'm just gonna ignore that...

PR #2758, opened 2017-10-18
adding public http import support

kcibul:


engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -479,9 +479,9 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
       case w: WorkflowSourceFilesWithDependenciesZip => validateNamespaceWithImports(w)
       case w: WorkflowSourceFilesWithoutImports =>
         val importResolvers: Seq[ImportResolver] = if (importLocalFilesystem) {
-          List(WdlNamespace.fileResolver)
+          List(WdlNamespace.fileResolver, WdlNamespace.httpResolver)
         } else {
-          List.empty
+          List(WdlNamespace.httpResolver)
```
cjllanwarne: So just to be clear, there's no way to turn this off?

kcibul: Yeah -- there's actually no way to turn any of these off really.  This whole section could do with a "configurability" ticket IMHO and streamline these code paths.  @katevoss future ticket?

geoffjentry: What is the argument for turning these off? 

security concerns?

cjllanwarne: Unforeseen security concerns would be my big one. Based on our history of other things that we thought should be on for everyone ending up with red faces and an off switch a few weeks later

PR #2759, opened 2017-10-18
Removes gap between workflow submission and abort request

Horneth:
For all endpoints taking a workflow ID, Cromwell first checks that it can find something about this workflow in the metadata. Although there's some time after submission when the workflow hasn't made it to the metadata yet even though we have a workflow ID. 
What this does is fallback to checking in the WorkflowStore if we can't find it in the metadata.

### DIFF
--- a/engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreActor.scala
@@ -1,24 +1,48 @@
 package cromwell.engine.workflow.workflowstore
 
 import akka.actor.{Actor, ActorLogging, ActorRef, Props}
+import akka.pattern.pipe
 import cats.data.NonEmptyList
 import cromwell.core.Dispatcher.EngineDispatcher
 import cromwell.core._
+import cromwell.services.metadata.MetadataService._
 import cromwell.util.GracefulShutdownHelper
 import cromwell.util.GracefulShutdownHelper.ShutdownCommand
+
+import scala.concurrent.Future
 
 final case class WorkflowStoreActor private(store: WorkflowStore, serviceRegistryActor: ActorRef)
   extends Actor with ActorLogging with GracefulShutdownHelper {
   import WorkflowStoreActor._
-
+  implicit val ec = context.dispatcher
+  
   lazy val workflowStoreSubmitActor: ActorRef = context.actorOf(WorkflowStoreSubmitActor.props(store, serviceRegistryActor), "WorkflowStoreSubmitActor")
   lazy val workflowStoreEngineActor: ActorRef = context.actorOf(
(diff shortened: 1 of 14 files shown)


geoffjentry:
Consider that it's likely that all metadata endpoints will wind up split offable fro the engine in the near future, would this hinder that?

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreActor.scala
```diff
@@ -1,37 +1,77 @@
...
+  implicit val ec = context.dispatcher
+  
   lazy val workflowStoreSubmitActor: ActorRef = context.actorOf(WorkflowStoreSubmitActor.props(store, serviceRegistryActor), "WorkflowStoreSubmitActor")
   lazy val workflowStoreEngineActor: ActorRef = context.actorOf(
     WorkflowStoreEngineActor.props(store, serviceRegistryActor), "WorkflowStoreEngineActor")
 
   override def receive = {
     case ShutdownCommand => waitForActorsAndShutdown(NonEmptyList.of(workflowStoreSubmitActor))
+    case WorkflowStateRequest(id) => 
+      pipe(statusOf(id)).to(sender())
```
mcovarr: `statusOf(id) pipeTo sender`?

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreActor.scala
```diff
@@ -1,37 +1,77 @@
...
+      ()
+    case ValidateWorkflowId(id) => 
+      pipe(checkWorkflowIdExistence(id)).to(sender())
```
mcovarr: here too

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreActor.scala
Horneth: I can never get this `pipeTo` to compile but I'll try again

mcovarr: I usually have that problem too, but I actually tried it and it worked. 🙂

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
```diff
@@ -351,8 +354,9 @@ object CromwellApiService {
 
   final case class BackendResponse(supportedBackends: List[String], defaultBackend: String)
 
-  final case class UnrecognizedWorkflowException(message: String) extends Exception(message)
-  final case class InvalidWorkflowException(message: String) extends Exception(message)
+  final case class UnrecognizedWorkflowException(id: WorkflowId) extends Exception(s"Unrecognized workflow ID: $id")
```
geoffjentry: nice catch

geoffjentry:
:+1: now with 100% less seagulling

PR #2760, opened 2017-10-18
CWL workflow id is the name

cjllanwarne:


cwl/src/main/scala/cwl/Workflow.scala
```diff
@@ -107,7 +108,6 @@ case class Workflow private(
   }
 
   def womDefinition: Checked[WorkflowDefinition] = {
-    // TODO: need to find a way to get a meaningful name here
     val name: String = "MyCwlWorkflow"
```
Horneth: Looks like the name is still `"MyCwlWorkflow"` here though ?

PR #2762, opened 2017-10-19
wdltool -> womtool -> cromwell

mcovarr:
The only really interesting commit is 80dd47f1a4e136e7c57971ffa4467ac92f3390f9. WOM graphing is broken because the method to create WOM Executables now requires compatible inputs, which womtool won't have. Pretty sure this was a change I asked for in a PR, whoops. 😦  I'll ticket the fix. Hopefully this mistake won't happen again with a monorepo.


Horneth:
> WOM graphing is broken because the method to create WOM Executables now requires compatible inputs, which womtool won't have. Pretty sure this was a change I asked for in a PR, whoops.

@mcovarr sounds like something I forgot to do 😨  I'll fix it after this PR is merged, I'm on bug rotation anyway :)

geoffjentry:
@mcovarr to save me from wading through a few hundred files, what's the project structure here, ie where would womtool now live?

mcovarr:
@Horneth pretty sure I made a comment asking you to break it like that, oops. 😄

Horneth:
@mcovarr Oh right yeah I remember, I thought you meant you made a comment about making sure it doesn't break wdltool, which I probably should have done anyway 😄

mcovarr:
@geoffjentry womtool is now a subproject under cromwell like all the others. The artifact that gets built for it can be run with `java -jar /path/to/cromwell-womtool-blah.jar` just like before (except for the WOM graphing which is temporarily broken).

cjllanwarne:
is it indended that "WomTypes" parts I and II are being hoovered in by this PR?

mcovarr:
@cjllanwarne This is meant to merge into the branch containing parts I and II.

geoffjentry:
@mcovarr I thought we said that the move to unicromwell wouldn't require `cromwell-` in front of jar names? I noticed that this morning in the convo about where the `wdl4s` artifacts went.

If this is a program designed to generically test wdl & cwl (which IMO it **is**) it shouldn't have `cromwell-` in front of the name.

mcovarr:
@geoffjentry I don't remember discussing that but I can certainly see the objection to it in this case. Let me look into what would be required to get rid of the `cromwell-`.

geoffjentry:
@mcovarr when people were asking if we could move stuff into cromwell i said as long as we have the ability to have the jars be independently branded and folks said "yeah, no problem"

Horneth:
LGTM 👍

cjllanwarne:
Things aren't looking good for the "let's see how well we use CodeCov" analysis...!

Notwithstanding that... 👍

PR #2765, opened 2017-10-20
Upgrade to 2.12.4

geoffjentry:


.travis.yml
```diff
@@ -5,7 +5,7 @@ services:
   - mysql
 language: scala
 scala:
-  - 2.12.2
+  - 2.12.4
```
mcovarr: oh lovely we weren't being consistent before 😦

PR #2770, opened 2017-10-20
womtool and common

mcovarr:
`lenthall` has been posthumously executed and reinterred under `common`. The former `cromwell-womtool` artifact is now published as just `womtool`.


Horneth:
I guess `lenthall` is now part of the `common` wealth

mcovarr:
Centaur will need to be adjusted to reference the new name.

cjllanwarne:
👍 I don't particularly like "common" since it's not at all clear what now goes in "common" vs "core" but 🤷‍♂️ I can't think of anything better

mcovarr:
@cjllanwarne I agree with that and I think we should spell out what belongs where after everything is in one big happy repo.

PR #2772, opened 2017-10-23
Pattern matching StringOrExpression

cjllanwarne:
One possible solution to the "`Poly`s are too much boilerplate" vs "`inl` and `inr` are too brittle" debate.

Also IMO much much easier to read the `match` `case`s


Horneth:
That does look pretty good to me, enough to make me reverse my concerns about `unapply` at least in this case. 👍  as far as I"m concerned

cwl/src/main/scala/cwl/package.scala
```diff
@@ -49,9 +49,16 @@ package object cwl extends TypeAliases {
...
 
   type StringOrExpression = Expression :+: String :+: CNil
-
   type Expression = ECMAScriptExpression :+: ECMAScriptFunction :+: CNil
 
+  object StringOrExpression {
+    object String {
+      def unapply(soe: StringOrExpression): Option[String] = soe.select[String]
+    }
+    object Expression {
```
danbills: can we take this one further and offer `ScriptExpression` and `ScriptFunction` that match to the nested `ECMAScriptExpression` and `ECMAScriptFunction`, respectively?

After all my talk of pattern matching being useful in nested coproducts I didn't even do it!

mcovarr:
:+1: 🐦-proof

cwl/src/main/scala/cwl/package.scala
cjllanwarne: I hadn't done initially since it wasn't being used (😄), but I don't see why not. Added.

PR #2775, opened 2017-10-23
Fixup file naming for CommandLineTool, remove some cruft

cjllanwarne:
The diff looks worse than it is. This PR:
- Moves `CommandLineTool` from `CwlFile.scala` to `CommandLineTool.scala`
- Moves the `case class`es in `CommandLineTool.scala` into the companion object.
- Removes some cruft, including an unused `Poly1`


Horneth:
LGTM when travis is green 👍

cwl/src/main/scala/cwl/CommandLineTool.scala
```diff
@@ -1,66 +1,191 @@
...
+
+  def womExecutable(inputFile: Option[String] = None): Checked[Executable] =
+    CwlExecutableValidation.builWomExecutable(taskDefinition.validNelCheck, inputFile)
+
+  def taskDefinition: TaskDefinition = {
+
+    val id = this.id
+
+    val commandTemplate: Seq[CommandPart] = baseCommand.toSeq.flatMap(_.fold(BaseCommandToCommandParts)) ++
+      arguments.toSeq.flatMap(_.map(_.fold(ArgumentToCommandPart)))
```
mcovarr: is this forcompable?

cwl/src/main/scala/cwl/CommandLineTool.scala
```diff
@@ -1,66 +1,191 @@
...
-                                   `type`: Option[MyriadOutputType] = None)
+  case class CommandOutputParameter(
+                                     id: String,
+                                     label: Option[String] = None,
+                                     secondaryFiles: Option[Expression :+: String :+: Array[Expression :+: String :+: CNil] :+: CNil] = None,
+                                     format: Option[Expression :+: Array[String] :+: String :+: CNil] = None, //only valid when type: File
+                                     streamable: Option[Boolean] = None, //only valid when type: File
+                                     doc: Option[String :+: Array[String] :+: CNil] = None,
+                                     outputBinding: Option[CommandOutputBinding] = None,
+                                     `type`: Option[MyriadOutputType] = None)
```
mcovarr: ToL and this is obviously preexisting, but it seems like there could be separate types here where these "only valid when type: File" attributes are only valid when the type is File

cwl/src/main/scala/cwl/CommandLineTool.scala
cjllanwarne: Perhaps. But this is just a small rename so I'll leave that for another ticket

cwl/src/main/scala/cwl/CommandLineTool.scala
cjllanwarne: As above.

PR #2777, opened 2017-10-24
Unzip CWL Dependencies and make available to pre-processing

danbills:
I'm throwing in some nice-to-haves here:

* deleted unused "mungeId" artifact before FullyQualifiedName was born
* added "sbt-pack" SBT plugin which makes nice run scripts and doesn't assemble fat jar (slow)

### DIFF
--- a/engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
@@ -3,6 +3,7 @@ package cromwell.engine.workflow.lifecycle
 import akka.actor.{ActorRef, FSM, LoggingFSM, Props, Status}
 import akka.pattern.pipe
 import cats.Monad
+import cats.data.{EitherT, NonEmptyList}
 import cats.data.NonEmptyList
 import cats.data.EitherT._
 import cats.data.Validated.{Invalid, Valid}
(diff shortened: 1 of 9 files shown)


engine/src/main/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -434,22 +435,26 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
...
-      for {
-        cwl <- CwlDecoder.decodeAllCwl(cwlFile)
-        wf <- fromEither[IO](cwl.select[Workflow].toRight(NonEmptyList.one(s"expected a workflow but got a $cwl")))
-        executable <-  fromEither[IO](wf.womExecutable(Option(source.inputsJson)))
-        graph <- fromEither[IO](executable.graph.toEither)
-        ioFunctions = new WdlFunctions(pathBuilders)
-        validatedWomNamespace <- fromEither[IO](validateWomNamespace(executable, ioFunctions))
-      } yield validatedWomNamespace
-    } finally {
-      cwlFile.delete(swallowIOExceptions = true)
```
danbills: This `finally` block was for some reason executing before the parsing step, throwing `FileNotFound` error from `cwltool`.  

The file and the dependencies are being created in a temporary directory though, and they're small, so I don't think this is that bad.

I am thinking of another way to ensure they get deleted but short of using fs2 I don't know how to ensure resource safety without the (problematic) `finally` block.

mcovarr:
Can you please open this in the main Cromwell PR?

mcovarr:
Er, what I meant to write was, can you please create this branch in the main Cromwell repo and PR that instead? Having PRs in private repos has been causing review and maintenance pain.

danbills:
Opening from main code repo

PR #2783, opened 2017-10-24
String only stdout/stderr redirection

cjllanwarne:
Expression-based `stdout` and `stderr` is a bit of a can of worms (TL;DR, which came first, the expression chicken or the stdout path egg?) and probably less valuable to our alpha than string-based stdout/stderr redirection (which is crucial).

For future reference: [a partially opened can of worms](https://github.com/broadinstitute/cromwell/tree/cjl_stdout_as_a_runtime_attr).

### DIFF
--- a/backend/src/main/scala/cromwell/backend/io/JobPaths.scala
+++ b/backend/src/main/scala/cromwell/backend/io/JobPaths.scala
@@ -1,5 +1,6 @@
 package cromwell.backend.io
 
+import cromwell.backend.BackendJobDescriptorKey
 import cromwell.core.path.Path
 import cromwell.core.{CallContext, JobKey}
 import cromwell.services.metadata.CallMetadataKeys
@@ -30,12 +31,15 @@ trait JobPaths {
 
   def workflowPaths: WorkflowPaths
   def returnCodeFilename: String = "rc"
-  def stdoutFilename: String = "stdout"
-  def stderrFilename: String = "stderr"
+  def defaultStdoutFilename = "stdout"
+  def defaultStderrFilename = "stderr"
+
+  final def stdoutFilename: String = jobKey.node.callable.stdoutRedirection.getOrElse(defaultStdoutFilename)
+  final def stderrFilename: String = jobKey.node.callable.stderrRedirection.getOrElse(defaultStderrFilename)
   def scriptFilename: String = "script"
   def dockerCidFilename: String = "docker_cid"
-  
-  def jobKey: JobKey
+
+  def jobKey: BackendJobDescriptorKey
   lazy val callRoot = callPathBuilder(workflowPaths.workflowRoot, jobKey)
   lazy val callExecutionRoot = callRoot
   lazy val stdout = callExecutionRoot.resolve(stdoutFilename)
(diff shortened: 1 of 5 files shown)


backend/src/main/scala/cromwell/backend/io/JobPaths.scala
```diff
@@ -30,12 +31,21 @@ trait JobPaths {
...
+  final def stderrFilename: String = jobKey.node match {
+    case t: TaskCallNode => t.callable.stdoutRedirection.getOrElse(defaultStderrFilename)
+    case _ => ??? // TODO: JobKey should be typesafe
+  }
   def scriptFilename: String = "script"
   def dockerCidFilename: String = "docker_cid"
-  
-  def jobKey: JobKey
+
+  def jobKey: CallKey
```
Horneth: Can you make it a `BackendJobDescriptorKey` ?

cwl/src/main/scala/cwl/CommandLineTool.scala
```diff
@@ -86,6 +86,11 @@ case class CommandLineTool private(
         RequiredInputDefinition(FullyQualifiedName(cip.id).id, tpe)
       }.toList
 
+    def stringOrExpressionToString(soe: Option[StringOrExpression]): Option[String] = soe match {
+      case Some(StringOrExpression.String(str)) => Some(str)
```
danbills: very small nit but if `match` were `flatMap` you could get rid of the `Some` in the case statement

backend/src/main/scala/cromwell/backend/io/JobPaths.scala
```diff
@@ -30,12 +31,21 @@ trait JobPaths {
...
   def workflowPaths: WorkflowPaths
   def returnCodeFilename: String = "rc"
-  def stdoutFilename: String = "stdout"
-  def stderrFilename: String = "stderr"
+  def defaultStdoutFilename = "stdout"
+  def defaultStderrFilename = "stderr"
+
+  final def stdoutFilename: String = jobKey.node match {
+    case t: TaskCallNode => t.callable.stdoutRedirection.getOrElse(defaultStdoutFilename)
+    case _ => ??? // TODO: JobKey should be typesafe
```
danbills: could `node` be `node: TaskCallNode` on CallKey?

backend/src/main/scala/cromwell/backend/io/JobPaths.scala
danbills: not to pile on but a variable rename would be nice too

backend/src/main/scala/cromwell/backend/io/JobPaths.scala
Horneth: No it could also be a `WorkflowCallNode` (for sub workflows).
But it should be a `TaskCallNode` on `BackendJobDescriptorKey`

backend/src/main/scala/cromwell/backend/io/JobPaths.scala
cjllanwarne: @Horneth will change, thanks!
@danbills I think I still like `jobKey` here, IMO the `JobKey` trait is more in need of a rename (eg to `GraphNodeKey` or similar)

cwl/src/test/resources/three_step.cwl
```diff
@@ -3,6 +3,7 @@ class: Workflow
 id: three_step
 inputs:
 - id: pattern
+  default: "v"
```
mcovarr: ?

cjllanwarne: Oops. Shouldn't have been committed! 😳

PR #2792, opened 2017-10-26
CWL Input Bindings into bash command

danbills:
This PR takes inputs marked w/ "inputBinding" positions and appends them to the bash script in the specified position.

### DIFF
--- a/cwl/src/main/scala/cwl/CommandLineTool.scala
+++ b/cwl/src/main/scala/cwl/CommandLineTool.scala
@@ -120,6 +120,10 @@ case class CommandLineTool private(
 
   def asCwl = Coproduct[Cwl](this)
 
+  /**
+    * Sort according to position. If position does not exist, use 0 per spec:
+    * http://www.commonwl.org/v1.0/CommandLineTool.html#CommandLineBinding
+    */
   def orderedInputs: Seq[CommandPart] =
     inputs.
       sortBy(_.inputBinding.flatMap(_.position).getOrElse(0)).
--- a/cwl/src/main/scala/cwl/CommandLineTool.scala
+++ b/cwl/src/main/scala/cwl/CommandLineTool.scala
@@ -52,7 +52,8 @@ case class CommandLineTool private(
     val id = this.id
 
     val commandTemplate: Seq[CommandPart] = baseCommand.toSeq.flatMap(_.fold(BaseCommandToCommandParts)) ++
-      arguments.toSeq.flatMap(_.map(_.fold(ArgumentToCommandPart)))
+      arguments.toSeq.flatMap(_.map(_.fold(ArgumentToCommandPart))) ++
+      sortedInputs
 
     val runtimeAttributes: RuntimeAttributes = RuntimeAttributes(Map.empty[String, WomExpression])
 
@@ -118,6 +119,12 @@ case class CommandLineTool private(
   }
 
   def asCwl = Coproduct[Cwl](this)
+
+  def sortedInputs: Seq[CommandPart] =
+    inputs.
+      sortBy(_.inputBinding.flatMap(_.position).getOrElse(0)).
+      toSeq.
+      map(InputParameterCommandPart.apply)
 }
 
 object CommandLineTool {
--- a/cwl/src/main/scala/cwl/CommandLineTool.scala
+++ b/cwl/src/main/scala/cwl/CommandLineTool.scala
@@ -53,7 +53,7 @@ case class CommandLineTool private(
 
     val commandTemplate: Seq[CommandPart] = baseCommand.toSeq.flatMap(_.fold(BaseCommandToCommandParts)) ++
       arguments.toSeq.flatMap(_.map(_.fold(ArgumentToCommandPart))) ++
-      sortedInputs
+      orderedInputs
 
     val runtimeAttributes: RuntimeAttributes = RuntimeAttributes(Map.empty[String, WomExpression])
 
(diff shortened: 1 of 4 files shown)


cwl/src/main/scala/cwl/CwlExpressionCommandPart.scala
```diff
@@ -43,3 +44,29 @@ case class CwlArgumentCommandPart(argument: CommandLineBinding) extends CommandP
...
+                           functions: IoFunctionSet,
+                           valueMapper: (WomValue) => WomValue) = {
+
+    val localInputsMap = inputsMap.map{
+      case (LocalName(localName), WomSingleFile(path)) => localName -> WomString(path)
+      case (LocalName(localName), value) => localName -> value
+    }
+
+
+    val womValue: WomValue = argument match {
```
cjllanwarne: At first I thought this was a straight copy/paste from above, but it does seem to be fundamentally different. 

Since it's so hard to parse the chains of `Some`s and `None`s and `_`s, could you add comments on what situation each case is trying to hit?

cwl/src/main/scala/cwl/CwlExpressionCommandPart.scala
```diff
@@ -43,3 +44,29 @@ case class CwlArgumentCommandPart(argument: CommandLineBinding) extends CommandP
...
 }
 
+case class InputParameterCommandPart(argument: CommandInputParameter) extends CommandPart {
+
+  override def instantiate(inputsMap: Map[LocalName, WomValue],
+                           functions: IoFunctionSet,
+                           valueMapper: (WomValue) => WomValue) = {
+
+    val localInputsMap = inputsMap.map{
+      case (LocalName(localName), WomSingleFile(path)) => localName -> WomString(path)
```
cjllanwarne: Just FMI, could you explain this?

cwl/src/main/scala/cwl/CwlExpressionCommandPart.scala
```diff
@@ -43,3 +44,29 @@ case class CwlArgumentCommandPart(argument: CommandLineBinding) extends CommandP
   }
 }
 
+case class InputParameterCommandPart(argument: CommandInputParameter) extends CommandPart {
+
+  override def instantiate(inputsMap: Map[LocalName, WomValue],
+                           functions: IoFunctionSet,
+                           valueMapper: (WomValue) => WomValue) = {
+
+    val localInputsMap = inputsMap.map{
```
cjllanwarne: ToL: Rather than preprocessing the entire list ahead of time, why not make a method (`eg "formatForCommandInsertion(womValue: WomValue)"`) and call it once we know which value it is we want to insert (on line 62)?

cwl/src/main/scala/cwl/CwlExpressionCommandPart.scala
```diff
@@ -43,3 +44,29 @@ case class CwlArgumentCommandPart(argument: CommandLineBinding) extends CommandP
...
+      case (LocalName(localName), WomSingleFile(path)) => localName -> WomString(path)
+      case (LocalName(localName), value) => localName -> value
+    }
+
+
+    val womValue: WomValue = argument match {
+      case CommandInputParameter(id, _,_,_,_,_, Some(CommandLineBinding(None,Some(_),None,None,None,None,None)),_,_) =>
```
Horneth: I wonder if at some point it just becomes better to just do `if blah.isDefined`, `blah.get` :smile:

danbills: This is done because I have nested options, i.e. command line binding and the position of the command line binding.

I can't match the case where there is a command line binding *and* the command line binding has a position binding with an if statement but I can with a pattern match.

danbills: The Nones are probably too strict.  I'm looking to see what I can make `_`

Horneth: What about adding a method on `CommandInputParameter` like 
```scala
def hasPosition = inputBinding.exists(_.position.isDefined)
```

danbills: So it turns out all those `None`'s being enforced is correct.  This case only handles when the input is bound by the id and not by any of the (many) options in `CommandLineBinding`.

Now I could write a function enforcing all this too, but it would only ever be used here and I would use a pattern match to implement that, too.  

I promise I would make this cleaner if I could :).

cwl/src/main/scala/cwl/CommandLineTool.scala
```diff
@@ -101,6 +102,12 @@ case class CommandLineTool private(
   }
 
   def asCwl = Coproduct[Cwl](this)
+
+  def sortedInputs: Seq[CommandPart] =
```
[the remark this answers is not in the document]
danbills: Roger, I think `orderedInputs` is best.

cwl/src/main/scala/cwl/CommandLineTool.scala
```diff
@@ -101,6 +102,12 @@ case class CommandLineTool private(
...
+    inputs.
+      sortBy(_.inputBinding.flatMap(_.position).getOrElse(0)).
```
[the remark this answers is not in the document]
danbills: Yep!  See `position` in http://www.commonwl.org/v1.0/CommandLineTool.html#CommandLineBinding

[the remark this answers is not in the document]
danbills: It now occurs to me that this is best immortalized in a comment.

Horneth:
👍

PR #2793, opened 2017-10-26
Allow CommandLineTools to run directly

cjllanwarne:
This PR is a lot less scary than it looks. Most of the changed lines spring from two changes (and the changes are actually make everything a lot simpler!):

- `TaskDefinition` is split into two: `CallableTaskDefinition` which can be called, and `ExecutableTaskDefinition` which can be executed.
- `Callable` now doesn't have a `graph: ErrorOr[Graph]`. Instead the new, more specific `ExecutableCallable` has a `graph: Graph` method.

### DIFF
--- a/cwl/src/main/scala/cwl/CommandLineTool.scala
+++ b/cwl/src/main/scala/cwl/CommandLineTool.scala
@@ -1,19 +1,22 @@
 package cwl
 
+import java.nio.file.Paths
+
 import common.Checked
-import common.validation.Checked._
 import shapeless.syntax.singleton._
-import shapeless.{:+:, CNil, Witness, Coproduct}
+import shapeless.{:+:, CNil, Coproduct, Witness}
 import CommandLineTool._
 import CwlType.CwlType
 import CwlVersion._
 import eu.timepit.refined.W
-import wom.callable.Callable.{OutputDefinition, RequiredInputDefinition}
-import wom.callable.{Callable, TaskDefinition}
+import wom.callable.Callable.{InputDefinitionWithDefault, OutputDefinition, RequiredInputDefinition}
+import wom.callable.{Callable, CallableTaskDefinition, ExecutableTaskDefinition}
 import wom.executable.Executable
-import wom.expression.WomExpression
+import wom.expression.{ValueAsAnExpression, WomExpression}
 import wom.types.WomFileType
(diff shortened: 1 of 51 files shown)


cwl/src/main/scala/cwl/CommandLineTool.scala
```diff
@@ -79,26 +82,33 @@ case class CommandLineTool private(
         OutputDefinition(FullyQualifiedName(cop_id).id, womType, CommandOutputExpression(outputBinding, womType, inputNames))
     }.toList
 
-    val inputs: List[_ <: Callable.InputDefinition] =
+    val inputDefinitions: List[_ <: Callable.InputDefinition] =
       this.inputs.map { cip =>
-        val tpe = cip.`type`.flatMap(_.select[CwlType]).map(cwlTypeToWdlType).get
-
-        RequiredInputDefinition(FullyQualifiedName(cip.id).id, tpe)
+        val inputType = cip.`type`.flatMap(_.select[CwlType]).map(cwlTypeToWdlType).get
```
mcovarr: forcompable?

cjllanwarne: Actually, I think I can tidy it somewhat by just bringing forwards the `.get`.

Of course it'll still be a `.get` but that's a known `TODO`

wom/src/main/scala/wom/callable/TaskDefinition.scala
```diff
@@ -5,84 +5,89 @@ import common.validation.ErrorOr.ErrorOr
...
-//      val jobOutput = output -> (coerced flatMap postMapper).recoverWith {
-//        case t: Throwable => Failure(new RuntimeException(s"Could not evaluate ${output.name} = ${output.expression.toWdlString}", t))
-//      }
-//      outputMap + jobOutput
-//    }) map { case (k, v) => k -> v }
-//
-//    TryUtil.sequenceMap(evaluatedOutputs, "Failed to evaluate outputs.")
-//  }
+object ExecutableTaskDefinition {
+  def tryApply(callableTaskDefinition: CallableTaskDefinition): ErrorOr[ExecutableTaskDefinition] =
```
danbills: ToL I'd prefer the name `apply` here but I predict you are reserving that for a constructor that returns the value w/ no effect.

PR #2803, opened 2017-10-31
Reinstate simple conditionals

cjllanwarne:
Reinstates the ability to run at least the `simple_if` WDL from centaur:
```wdl
task runMe {
  command {
    echo "done"
  }
  output {
    String s = read_string(stdout())
  }
  runtime {
    docker: "ubuntu:latest"
  }
}

workflow simple_if {
  if (true) {
    call runMe as runMeTrue
  }

  if (false) {
    call runMe as runMeFalse
  }

  output {
    runMeTrue.s
    runMeFalse.s
  }
}
```

### DIFF
--- a/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
@@ -133,8 +133,10 @@ final case class ExecutionStore(private val statusStore: Map[JobKey, ExecutionSt
     def calculateUpstreamDone(upstream: GraphNode, index: ExecutionIndex): Boolean = upstream match {
         // OuterGraphInputNode signals that an input comes from outside the graph.
         // Depending on whether or not this input is outside of a scatter graph will change the index which we need to look at
-      case outerNode: OuterGraphInputNode if key.node.isInstanceOf[ScatterNode] => calculateUpstreamDone(outerNode, None)
-      case outerNode: OuterGraphInputNode => calculateUpstreamDone(outerNode, index)
+      case outerNode: ScatterVariableNode =>
+        calculateUpstreamDone(outerNode.linkToOuterGraph.graphNode, None)
+      case outerNode: OuterGraphInputNode =>
+        calculateUpstreamDone(outerNode.linkToOuterGraph.graphNode, index)
       case _: CallNode | _: ScatterNode | _: ExpressionNode | _: ConditionalNode => doneKeys.contains(upstream, index)
       case _ => true
     }
--- a/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
@@ -75,12 +75,6 @@ final case class ExecutionStore(private val statusStore: Map[JobKey, ExecutionSt
 
   private def keysWithStatus(status: ExecutionStatus) = store.getOrElse(status, List.empty)
 
-  def isInBypassedConditional(jobKey: JobKey): Boolean = keysWithStatus(Bypassed).exists {
-    case conditional: ConditionalKey if conditional.node.innerGraph.nodes.contains(jobKey.node)
-      && conditional.index.equals(jobKey.index) => true
-    case _ => false
-  }
-
   def hasActiveJob: Boolean = {
     def upstreamFailed(node: GraphNode): Boolean = node.upstreamAncestry exists hasFailedNode
 
--- a/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
@@ -31,8 +31,6 @@ object ExecutionStore {
   
   case class RunnableScopes(scopes: List[JobKey], truncated: Boolean)
 
-  def empty = ExecutionStore(Map.empty[JobKey, ExecutionStatus], hasNewRunnables = false)
-
   def apply(callable: ExecutableCallable): ExecutionStore = {
     // Keys that are added in a NotStarted Status
     val notStartedKeys = callable.graph.nodes collect {
@@ -71,7 +69,7 @@ final case class ExecutionStore(private val statusStore: Map[JobKey, ExecutionSt
         val newDone = if (status.isDoneOrBypassed) done.addAll(newMapEntries) else done
         val newTerminal = if (status.isTerminal) terminal.addAll(newMapEntries) else terminal
 
-        newDone -> newTerminal
+        (newDone, newTerminal)
     })
   }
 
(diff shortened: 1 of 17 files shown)


cjllanwarne:
Things to investigate as we move to more complex cases: 
- Can I remove the `processBypassedScope` method? I think so since everything should be covered by filling in the `ConditionalGatherPort`s with `None`s
- The `ExecutionStore` seemed to start with some of the `ConditionalNode`'s inner graph already present. Double check

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -445,18 +445,17 @@ case class WorkflowExecutionActor(workflowDescriptor: EngineWorkflowDescriptor,
   private def processBypassedScope(jobKey: JobKey, data: WorkflowExecutionActorData): Try[WorkflowExecutionDiff] = {
     Success(
       WorkflowExecutionDiff(
-        executionStoreChanges = Map(jobKey -> ExecutionStatus.Bypassed),
-        valueStoreAdditions = bypassedScopeResults(jobKey)
+        executionStoreChanges = Map.empty, // Map(jobKey -> ExecutionStatus.Bypassed),
+        valueStoreAdditions = Map.empty //bypassedScopeResults(jobKey)
```
Horneth: Is this method still useful then ? The purpose of the `WorkflowExecutionDiff` is to represent changes in the execution store and value store after a key has been processed. So if they're both empty here it's not doing anything ?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
```diff
@@ -124,18 +130,22 @@ final case class ExecutionStore(private val statusStore: Map[JobKey, ExecutionSt
...
+        calculateUpstreamDone(outerNode.linkToOuterGraph.graphNode, index)
       case _: CallNode | _: ScatterNode | _: ExpressionNode | _: ConditionalNode => doneKeys.contains(upstream, index)
       case _ => true
     }
 
     def upstreamAreDone = key.node.upstreamPorts forall {
       // The collector is at index None, so if this is a scatter gather port ignore the key index
       case upstreamPort: ScatterGathererPort => calculateUpstreamDone(upstreamPort.executionNode, None)
+      case upstreamPort: ConditionalOutputPort =>
+        statusStore.get(ConditionalCollectorKey(upstreamPort, key.index, upstreamPort.g(()))).exists(_.isDoneOrBypassed)
```
Horneth: Re-creating a fake key just to check that one with the same hash is in the store feels kind of weird.
I think you could use the `doneKeys` table that contains all the keys in `DoneOrBypassed` status, grouped by node and index and check that the node is in there instead

wom/src/main/scala/wom/types/WomOptionalType.scala
```diff
@@ -15,14 +15,16 @@ case class WomOptionalType(memberType: WomType) extends WomType {
     */
   override protected def coercion: PartialFunction[Any, WomValue] = {
 
-    // It's safe to box up values implicitly:
-    case womValue: WomValue if memberType.equals(womValue.womType) => WomOptionalValue(womValue)
-    case coerceable: Any if memberType.coercionDefined(coerceable) => WomOptionalValue(memberType.coerceRawValue(coerceable).get)
+    case alreadyGood @ WomOptionalValue(WomOptionalType(`memberType`), _) => alreadyGood
```
mcovarr: Does `alreadyGood` mean `alreadyAnExactMatchForTheDesiredType`?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
```diff
@@ -71,7 +69,7 @@ final case class ExecutionStore(private val statusStore: Map[JobKey, ExecutionSt
         val newDone = if (status.isDoneOrBypassed) done.addAll(newMapEntries) else done
         val newTerminal = if (status.isTerminal) terminal.addAll(newMapEntries) else terminal
 
-        newDone -> newTerminal
+        (newDone, newTerminal)
```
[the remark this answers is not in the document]
cjllanwarne: Because it's folding over a single updated tuple, not creating a list of map elements

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
cjllanwarne: This is because `doneKeys` is based on GraphNode, whereas I need to check the `ConditionalCollector` (which doesn't have it's own Node)

`Collector` != `ConditionalNode`

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
cjllanwarne: Yeah I mentioned this in my "look at again for complex cases". FWIW I'm pretty sure the answer is "it can be removed"

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -570,12 +569,20 @@ case class WorkflowExecutionActor(workflowDescriptor: EngineWorkflowDescriptor,
...
+    // This is the output port from the conditional's 'condition' input:
+    val conditionOutputPort = conditionalKey.node.conditionExpression.singleExpressionOutputPort
     data.valueStore.get(conditionOutputPort, conditionalKey.index) match {
       case Some(b: WomBoolean) =>
         val conditionalStatus = if (b.value) ExecutionStatus.Done else ExecutionStatus.Bypassed
-        Success(WorkflowExecutionDiff(conditionalKey.populate + (conditionalKey -> conditionalStatus)))
+
+        val valueStoreAdditions: Map[ValueKey, WomValue] = if (!b.value) {
+          conditionalKey.node.outputPorts.map(op => ValueKey(op, conditionalKey.index) -> WomOptionalValue(op.womType, None)).toMap
+        } else Map.empty
```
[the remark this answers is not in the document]
cjllanwarne: Right, we can only collect the outputs immediately if we aren't going to run it

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -642,10 +650,20 @@ case class WorkflowExecutionActor(workflowDescriptor: EngineWorkflowDescriptor,
         override def exceptionContext: String = s"Failed to collect conditional output value for node ${collector.tag}"
         override def errorMessages: Traversable[String] = e.toList
       })
-      case Valid(outputs) => Success(WorkflowExecutionDiff(
-        executionStoreChanges = Map(collector -> ExecutionStatus.Done),
-        valueStoreAdditions = outputs
-      ))
+      case Valid(outputs) => {
+        if (outputs.size != 1) {
+          Failure(new Exception(s"Expected exactly 1 result back for $collector"))
```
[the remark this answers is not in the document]
cjllanwarne: Would you be happier if I put "Developer error! This should be impossible!" at the start of this error message?

wom/src/main/scala/wom/types/WomOptionalType.scala
cjllanwarne: Yes

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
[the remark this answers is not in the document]
mcovarr: mokay, if you think this is more clear. They should be equivalent.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/ExecutionStore.scala
mcovarr: please comment this

wom/src/main/scala/wom/types/WomOptionalType.scala
mcovarr: comment plz

cjllanwarne:
This PR will return when less simple `if`s are added too

PR #2807, opened 2017-10-31
Improving type evaluation and lookup

orodeh:
This is a port of the original pull [request](https://github.com/broadinstitute/wdl4s/pull/256) to the Cromwell tree. I hope this was done correctly, while addressing the review comments. Thanks for pointing out how to use `Try` and `should equal` in the testing framework.

There are two outstanding comments, which I am answering here, since I am assuming you want to dismantle the old wdl4s git repository. 

1) TypeEvaluator.scala
*Q: Isn't the caller perhaps trying to ascertain if the expression actually can be coerced to the declared type?* 
A: That could be added as an assert. 

2) WdlSubworkflowWomSpec.scala
*Q: The magic conversion of a single Thing to an Array[Thing] was a feature someone explicitly asked for way back when. I never liked this feature but I'm sure there's WDL out there that expects this to work.*
A: I can see why a user might want that. A side effect is that Array[T] can automatically be coerced to Array[Array[T]]. I ran into a case where Array[String] was coerced into Array[Array[File]]. For example, the workflow below executes under Cromwell v0.29, involving the coercion of `Array[Int]` to `Array[Array[Int?]]`.
  
```
# Trying out file copy operations
task Num {
  Array[Array[Int?]] numbers
  command {
  }
  output {
    Array[Array[Int?]] result = numbers
  }
}


workflow w {
  Array[Int] primes = [2, 3, 5, 7, 11]
  call Num { input: numbers = primes }
  output {
    Num.result
  }
}
```

This raises two issues in my mind: 
1)  There are many ways to convert a single dimensional array into a two dimensional array. Which one does the WDL language specify? 
2) A user can mistakenly pass an incorrectly typed argument to a task without getting a compiler error.

### DIFF
--- a/wdl/src/main/scala/wdl/WdlNamespace.scala
+++ b/wdl/src/main/scala/wdl/WdlNamespace.scala
@@ -457,13 +457,16 @@ object WdlNamespace {
     invalidVariableReferences ++ typeMismatches
   }
 
-  private def lookupType(from: Scope)(n: String): WomType = {
+  def lookupType(from: Scope)(n: String): WomType = {
     val resolved = from.resolveVariable(n)
     resolved match {
       case Some(d: DeclarationInterface) => d.relativeWdlType(from)
       case Some(c: WdlCall) => WdlCallOutputsObjectType(c)
       case Some(s: Scatter) => s.collection.evaluateType(lookupType(s), new WdlStandardLibraryFunctionsType, Option(from)) match {
-        case Success(a: WomArrayType) => a.memberType
+        case Success(WomArrayType(aType)) => aType
+        case Success(WomMapType(kType,vType)) =>
+          // Collection is a map
+          WomPairType(kType, vType)
         case _ => throw new VariableLookupException(s"Variable $n references a scatter block ${s.fullyQualifiedName}, but the collection does not evaluate to an array")
       }
       case Some(_: WdlNamespace) => WdlNamespaceType
--- a/wdl/src/main/scala/wdl/WdlNamespace.scala
+++ b/wdl/src/main/scala/wdl/WdlNamespace.scala
@@ -464,9 +464,8 @@ object WdlNamespace {
       case Some(c: WdlCall) => WdlCallOutputsObjectType(c)
       case Some(s: Scatter) => s.collection.evaluateType(lookupType(s), new WdlStandardLibraryFunctionsType, Option(from)) match {
         case Success(WomArrayType(aType)) => aType
-        case Success(WomMapType(kType,vType)) =>
-          // Collection is a map
-          WomPairType(kType, vType)
+        // We don't need to check for a WOM map type, because
+        // of the custom unapply in object WomArrayType
         case _ => throw new VariableLookupException(s"Variable $n references a scatter block ${s.fullyQualifiedName}, but the collection does not evaluate to an array")
       }
       case Some(_: WdlNamespace) => WdlNamespaceType
--- a/wdl/src/main/scala/wdl/WdlNamespace.scala
+++ b/wdl/src/main/scala/wdl/WdlNamespace.scala
@@ -457,7 +457,7 @@ object WdlNamespace {
     invalidVariableReferences ++ typeMismatches
   }
 
-  def lookupType(from: Scope)(n: String): WomType = {
+  private [wdl] def lookupType(from: Scope)(n: String): WomType = {
     val resolved = from.resolveVariable(n)
     resolved match {
       case Some(d: DeclarationInterface) => d.relativeWdlType(from)
--- a/wdl/src/main/scala/wdl/expression/TypeEvaluator.scala
+++ b/wdl/src/main/scala/wdl/expression/TypeEvaluator.scala
@@ -85,10 +85,13 @@ case class TypeEvaluator(override val lookup: String => WomType, override val fu
           evaluate(a.getAttribute("lhs")).flatMap {
             case o: WdlCallOutputsObjectType =>
               o.call.outputs.find(_.unqualifiedName == rhs.getSourceString) match {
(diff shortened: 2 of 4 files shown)


mcovarr:
The build failures are in unrelated Swagger tests, rebasing on develop should get this green.

wdl/src/main/scala/wdl/WdlNamespace.scala
```diff
@@ -457,13 +457,16 @@ object WdlNamespace {
     invalidVariableReferences ++ typeMismatches
   }
 
-  private def lookupType(from: Scope)(n: String): WomType = {
+  def lookupType(from: Scope)(n: String): WomType = {
```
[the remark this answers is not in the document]
orodeh: Thanks, fixed!

orodeh:
Thank @mcovarr and @cjllanwarne, I believe I am done. Please let me know if there is anything else. I am not authorized to merge this PR, so I am leaving this up to your team.

wdl/src/main/scala/wdl/WdlNamespace.scala
```diff
@@ -457,13 +457,16 @@ object WdlNamespace {
...
-  private def lookupType(from: Scope)(n: String): WomType = {
+  private [wdl] def lookupType(from: Scope)(n: String): WomType = {
     val resolved = from.resolveVariable(n)
     resolved match {
       case Some(d: DeclarationInterface) => d.relativeWdlType(from)
       case Some(c: WdlCall) => WdlCallOutputsObjectType(c)
       case Some(s: Scatter) => s.collection.evaluateType(lookupType(s), new WdlStandardLibraryFunctionsType, Option(from)) match {
-        case Success(a: WomArrayType) => a.memberType
+        case Success(WomArrayType(aType)) => aType
+        case Success(WomMapType(kType,vType)) =>
```
cjllanwarne: This second line will never hit because of the custom unapply in `object WomArrayType`. I'd suggest removing it and adding a comment to that effect

orodeh: Fixed!

cjllanwarne:
@orodeh in case you're curious, the build failure is currently being caused by:
```
[0m[[0minfo[0m] [0m[31m*** 1 TEST FAILED ***[0m[0m
[0m[[0minfo[0m] [0m[31mWdlSubworkflowWomSpec:[0m[0m
[0m[[0minfo[0m] [0m[31mWdlNamespaces with subworkflows [0m[0m
[0m[[0minfo[0m] [0m[31m- should support WDL to WOM conversion of subworkflow calls *** FAILED *** (51 milliseconds)[0m[0m
[0m[[0minfo[0m] [0m[31m  wdl4s.parser.WdlParser$SyntaxError: ERROR: out is declared as a Array[String] but the expression evaluates to a String:[0m[0m
[0m[[0minfo[0m] [0m[31m[0m[0m
[0m[[0minfo[0m] [0m[31m    Array[String] out = inner.out[0m[0m
[0m[[0minfo[0m] [0m[31m                  ^[0m[0m
[0m[[0minfo[0m] [0m[31m  at wdl.WdlNamespace$.$anonfun$typeCheckDeclaration$1(WdlNamespace.scala:493)[0m[0m
[0m[[0minfo[0m] [0m[31m  at scala.Option.flatMap(Option.scala:171)[0m[0m
[0m[[0minfo[0m] [0m[31m  at wdl.WdlNamespace$.typeCheckDeclaration(WdlNamespace.scala:488)[0m[0m
[0m[[0minfo[0m] [0m[31m  at wdl.WdlNamespace$.validateDeclaration(WdlNamespace.scala:466)[0m[0m
[0m[[0minfo[0m] [0m[31m  at wdl.WdlNamespace$.$anonfun$apply$35(WdlNamespace.scala:381)[0m[0m
[0m[[0minfo[0m] [0m[31m  at scala.collection.TraversableLike.$anonfun$flatMap$1(TraversableLike.scala:241)[0m[0m
[0m[[0minfo[0m] [0m[31m  at scala.collection.Iterator.foreach(Iterator.scala:929)[0m[0m
[0m[[0minfo[0m] [0m[31m  at scala.collection.Iterator.foreach$(Iterator.scala:929)[0m[0m
[0m[[0minfo[0m] [0m[31m  at scala.collection.AbstractIterator.foreach(Iterator.scala:1417)[0m[0m
[0m[[0minfo[0m] [0m[31m  at scala.collection.IterableLike.foreach(IterableLike.scala:71)[0m[0m
```

I'm not sure whether you intended to roll back that change at the same time as rolling back the test case? I think we can argue to make the set of coercions explicit in draft 3 (and not include `X => Array[X]`), but IMO we shouldn't "unsupport" that Cromwell feature with this PR.

orodeh:
@cjllanwarne, I am fine with supporting this only when moving to draft 3. I don't know how to do that in this codebase though.

By the way, this was never a declared WDL or Cromwell feature. I only stumbled across it by accident, when implementing a WDL-runner, and comparing against Cromwell.

geoffjentry:
@orodeh re `draft-3` and Cromwell what @cjllanwarne and I were discussing this morning is that we can just not worry about it as long as `draft-3` is fully backwards compatible with `draft-2` and/or "what cromwell already does", so that'll buy us some time.

cjllanwarne:
@orodeh @geoffjentry this is probably a case where the WDL spec has failed to keep up with developments in Cromwell. What's awkward in this case is I think Cromwell could be argued to be overreaching and adding things that **should not** be in the WDL spec!

I'd be fine to remove this as a "bug fix", but you'll still need to "fix" the test case before I can merge it.

orodeh:
@cjllanwarne I believe I fixed the problem. 

I will know where to look for the travis problems next time.

PR #2808, opened 2017-11-01
Aborts v2

Horneth:


core/src/main/scala/cromwell/core/ExecutionStatus.scala
```diff
@@ -2,21 +2,24 @@ package cromwell.core
...
       case QueuedInCromwell => 1
       case Starting => 2
       case Running => 3
-      case Aborted => 4
-      case Bypassed => 5
-      case RetryableFailure => 6
-      case Failed => 7
-      case Done => 8
+      case Aborting => 4
+      case UnReachable => 5
```
geoffjentry: Is unreachable 2 words? :)

Horneth: Argh I knew I forgot to fix this somewhere :p
Also congratulations you just officially started the seagulling ! Let it rain ! : smile:

geoffjentry: It's what i do

mcovarr: That's not rain.

cjllanwarne: 🐦 🐦 🐦 
; ; ; ; ;
  ☔️
  😨

cjllanwarne: (could it be `Unstartable` or `InterruptedBeforeStarting` or `AintGonnaHappen`?)

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/stores/ExecutionStore.scala
```diff
@@ -0,0 +1,253 @@
...
+    * If needsUpdate, returns an empty list of runnable keys and this instance of the store.
+    * 
+    * This method can expansive to run for very large workflows if needsUpdate is true.
+    */
+  def update: ExecutionStoreUpdate = if (needsUpdate) {
+    // When looking for runnable keys, keep track of the ones that are unreachable so we can mark them as such
+    var unreachables = Map.empty[JobKey, ExecutionStatus]
+
+    // filter the keys that are runnable. In the process remember the ones that are unreachable
+    val readyToStart = keysWithStatus(NotStarted).toStream.filter(key => {
```
danbills: ToL curious why `toStream`?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/stores/ExecutionStore.scala
```diff
@@ -0,0 +1,253 @@
...
+    * In the process, identifies and updates the status of keys that are unreachable.
+    * Returns an ExecutionStoreUpdate which is the list of runnable keys and an updated execution store.
+    * 
+    * If needsUpdate, returns an empty list of runnable keys and this instance of the store.
+    * 
+    * This method can expansive to run for very large workflows if needsUpdate is true.
+    */
+  def update: ExecutionStoreUpdate = if (needsUpdate) {
+    // When looking for runnable keys, keep track of the ones that are unreachable so we can mark them as such
+    var unreachables = Map.empty[JobKey, ExecutionStatus]
```
danbills: ToL could do a `foldMap` below and skip the var here.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -311,34 +299,102 @@ case class WorkflowExecutionActor(workflowDescriptor: EngineWorkflowDescriptor,
...
+     * 
+     * 8) We get a response saying that B was already successful
+     * 9) We try to reconnect to B1 (and only reconnect, not recover.
+     * We know to do that because we're in Failing state and restarting - see processRunnableJob method)
+     * 10) We reconnect to B1 and wait for a terminal status
+     * 11) If B1 had a dependency B2, then we'd try to reconnect to it as well, until we find a job that was never started
+     * and for which reconnection will fail. When that happens we'll fail the job (see failing state).
+     * 
+     * This guarantees that we'll try to reconnect to all potentially running jobs on restart.
+    */
```
danbills: 💯

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/keys/ScatterCollectorKey.scala
```diff
@@ -15,4 +17,19 @@ private [execution] case class ScatterCollectorKey(node: GraphNode,
   override val index = None
   override val attempt = 1
   override val tag = s"Collector-${node.localName}"
+
+  def processRunnable(data: WorkflowExecutionActorData): ErrorOr[WorkflowExecutionDiff] = {
+    data.valueStore.collectShards(this) map { outputs =>
+      // TODO WOM: fix
```
danbills: create issue perhaps?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/stores/ExecutionStore.scala
```diff
@@ -0,0 +1,253 @@
...
+        // For a scatter collector key, the dependencies are all the shards
+        case scatterCollector: ScatterCollectorKey => statusTable.row(scatterCollector.node).size == scatterCollector.scatterWidth
+        case conditionalCollector: ConditionalCollectorKey => statusTable.contains(conditionalCollector.node.executionNode, key.index)
+        // In the general case, the dependencies are held by the upstreamPorts
+        case _ => key.node.upstreamPorts forall portIsInStatus
+      }
+    }
+  }
+  
+  implicit class EnhancedOutputPort(val outputPort: OutputPort) extends AnyVal {
```
danbills: could this be tacked on to `OutputPort`?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/stores/ExecutionStore.scala
Horneth: Two things:
- To avoid making Cromwell cry too much, a little while ago we limited the number of "runnable keys" that this function will return per call so that if you have a 1M wide scatter, Cromwell doesn't starts 1M job literally at the same time. Instead, we return up to the first 1000 ones. This is called every second so it's still relatively fast but give some breathing room to Cromwell.

- It's potentially expensive to determine if all the dependencies of a key are in a given status (it's actually less now than it used to be since we've optimized the data structures but...).

Because of those 2 things, I wanted to avoid calculating "are you runnable?" for all of the keys if we're only going to take the first 1000 anyway. A stream / filter / take lazily computes only the first 1000 runnable keys and then stops, so we don't lose any cycle on the rest.

If you wonder why all this fuss, we had performance issues at a few different occasions as people started throwing bigger and bigger scatters to Cromwell and tracked it down to lots of CPU spent computing runnable keys.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/stores/ExecutionStore.scala
Horneth: I think we'd need some sort of short-circuiting fold if we want to keep the "stream" behavior (see my answer to your stream question) ?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/stores/ExecutionStore.scala
Horneth: Definitely could, I would say it depends on if we consider it being an artifact of how Cromwell execute WOM workflows in which case I'd argue it should be in Cromwell, or a WOM concept in which case putting it in the OutputPort class makes more sense.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/keys/ScatterCollectorKey.scala
Horneth: Yeah this is related to conditionals so I'd say the "Reinstate conditional" ticket covers it but I'll link to it specifically in the TODO.

engine/src/test/scala/cromwell/engine/workflow/WorkflowActorSpec.scala
```diff
@@ -10,15 +10,16 @@ import cromwell.core.path.DefaultPathBuilder
 import cromwell.engine.EngineWorkflowDescriptor
 import cromwell.engine.backend.BackendSingletonCollection
 import cromwell.engine.workflow.WorkflowActor._
-import cromwell.engine.workflow.lifecycle.MaterializeWorkflowDescriptorActor.MaterializeWorkflowDescriptorFailureResponse
-import cromwell.engine.workflow.lifecycle.WorkflowFinalizationActor.{StartFinalizationCommand, WorkflowFinalizationSucceededResponse}
-import cromwell.engine.workflow.lifecycle.WorkflowInitializationActor.{WorkflowInitializationAbortedResponse, WorkflowInitializationFailedResponse}
+import cromwell.engine.workflow.lifecycle.materialization.MaterializeWorkflowDescriptorActor.MaterializeWorkflowDescriptorFailureResponse
+import cromwell.engine.workflow.lifecycle.finalization.WorkflowFinalizationActor.{StartFinalizationCommand, WorkflowFinalizationSucceededResponse}
+import cromwell.engine.workflow.lifecycle.initialization.WorkflowInitializationActor.{WorkflowInitializationAbortedResponse, WorkflowInitializationFailedResponse}
```
mcovarr: oh all they all the big workflow actor states get their own packages! 😄

out
```diff
@@ -0,0 +1,3 @@
+gs://bucket/file1.bam
+gs://bucket/file2.bam
```
mcovarr: this file looks like a stowaway?

engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
```diff
@@ -603,12 +602,12 @@ object CromwellApiServiceSpec {
...
+      case AbortWorkflowCommand(id) =>
         val message = id match {
-          case ExistingWorkflowId => WorkflowStoreEngineActor.WorkflowAborted(id)
-          case UnrecognizedWorkflowId => WorkflowStoreEngineActor.WorkflowAbortFailed(id, new WorkflowNotFoundException(s"Couldn't abort $id because no workflow with that ID is in progress"))
+          case ExistingWorkflowId => WorkflowAbortingResponse(id)
+          case UnrecognizedWorkflowId => WorkflowAbortFailureResponse(id, new WorkflowNotFoundException(s"Couldn't abort $id because no workflow with that ID is in progress"))
           case AbortedWorkflowId =>
-            WorkflowAbortFailed(id, new IllegalStateException(s"Workflow ID '$id' is in terminal state 'Aborted' and cannot be aborted."))
+            WorkflowAbortFailureResponse(id, new IllegalStateException(s"Workflow ID '$id' is in terminal state 'Aborted' and cannot be aborted."))
           case WorkflowId(_) => throw new Exception("Something untoward happened")
```
mcovarr: So much better than Aw crap.

supportedBackends/sfs/src/test/scala/cromwell/backend/sfs/SharedFileSystemJobExecutionActorSpec.scala
```diff
@@ -37,9 +36,9 @@ class SharedFileSystemJobExecutionActorSpec extends TestKitSuite("SharedFileSyst
     StandardValidatedRuntimeAttributesBuilder.default(Some(TestConfig.optionalRuntimeConfig)).definitions.toSet
 
   def executeSpec(docker: Boolean): Any = {
-    val expectedOutputs: CallOutputs = Map(
-      "salutation" -> JobOutput(WomString("Hello you !"))
-    )
+    WomMocks
```
mcovarr: ?

backend/src/main/scala/cromwell/backend/async/AsyncBackendJobExecutionActor.scala
```diff
@@ -46,6 +52,7 @@ trait AsyncBackendJobExecutionActor { this: Actor with ActorLogging =>
     case _ => false
   }
 
+  // This seems dangerous as any exception that is not one of the above (fatal) will trigger infinite retries !
```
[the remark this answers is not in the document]
geoffjentry: Agree. We should have *some* sort of limit on everything

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -293,6 +293,32 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
   def recoverAsync(jobId: StandardAsyncJob): Future[ExecutionHandle] = Future.fromTry(Try(recover(jobId)))
 
   /**
+    * A correct implementation should reconnect to the specified job id, and upon reconnection try to abort the job.
+    * With the understanding that the job might already be complete or aborted.
```
geoffjentry: I'm not the grammarian that @mcovarr is but this sentence seems like it's not a sentence.

wom/src/main/scala/wom/graph/GraphNodePort.scala
```diff
@@ -50,6 +51,12 @@ object GraphNodePort {
     }
   }
   case class GraphNodeOutputPort(override val identifier: WomIdentifier, womType: WomType, graphNode: GraphNode) extends OutputPort
+  object ExpressionBasedOutputPort {
+    def apply(localName: LocalName, womType: WomType, graphNode: GraphNode, expression: WomExpression): ExpressionBasedOutputPort = {
+      ExpressionBasedOutputPort(WomIdentifier(localName, graphNode.identifier.fullyQualifiedName.combine(localName.value)), womType, graphNode, expression)
+    }
+  }
+  case class ExpressionBasedOutputPort(override val identifier: WomIdentifier, womType: WomType, graphNode: GraphNode, expression: WomExpression) extends OutputPort
```
cjllanwarne: OOI, what's happening here?

Doesn't seem super-related to the aborts stuff?

core/src/main/scala/cromwell/core/ExecutionStatus.scala
```diff
@@ -2,21 +2,24 @@ package cromwell.core
...
       case QueuedInCromwell => 1
       case Starting => 2
       case Running => 3
-      case Aborted => 4
-      case Bypassed => 5
-      case RetryableFailure => 6
-      case Failed => 7
-      case Done => 8
+      case Aborting => 4
+      case Unstartable => 5
```
geoffjentry: This keeps making me think of https://youtu.be/Gg--VIL2I6Y?t=75 :)

docs/Abort.md
```diff
@@ -0,0 +1,30 @@
...
+Note that by the time the backend is asked to abort a job, it might already be successful, or failed, in which case Cromwell will report the job as such.
+
+As an example, the Google backend when asked to abort a job will send an abort request to the Pipelines API. When Pipelines API indicates that the status of the job is aborted, Cromwell will mark it as such.
+However the abort action is entirely dependent on the backend, in this particular case, PAPI does not guarantee the success of an abort request (see [Pipelines API documentation on abort](https://cloud.google.com/genomics/reference/rest/v1alpha2/operations/cancel)).
+
+You'll also notice that the workflow is finalized even though being aborted.
+Finalization is the last step in the execution of a workflow and a chance for each backend to do some work before the workflow is terminated.
+For example, the Google backend in some cases uploads a file containing authentication information at the beginning of the workflow. This file is only necessary during the execution of the workflow and is deleted in the finalization step. This will be done even if the workflow is aborted.
```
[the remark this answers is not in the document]
geoffjentry: I usually just refer to "control files" as a generic term for the notion that sometimes cromwell puts files on behalf of the user for general plumbing purposes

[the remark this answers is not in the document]
Horneth: yeah good point, I'll rephrase

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/keys/ConditionalKey.scala
```diff
@@ -45,4 +49,18 @@ private [execution] case class ConditionalKey(node: ConditionalNode, index: Exec
...
+    data.valueStore.get(conditionOutputPort, index) match {
+      case Some(b: WomBoolean) =>
+        val conditionalStatus = if (b.value) ExecutionStatus.Done else ExecutionStatus.Bypassed
+        WorkflowExecutionDiff(populate + (this -> conditionalStatus)).validNel
+      case Some(v: WomValue) =>
+        s"'if' condition ${node.conditionExpression.womExpression.sourceString} must evaluate to a boolean but instead got ${v.womType.toDisplayString}".invalidNel
+      case None =>
+        s"Could not find the boolean value for conditional $tag. Missing boolean should have come from expression ${node.conditionExpression.womExpression.sourceString}".invalidNel
+    }
+  }
```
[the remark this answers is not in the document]
Horneth: I moved the "processRunnable" method for each key in the key class to make some room in the WEA. So this is just a copy paste.

backend/src/main/scala/cromwell/backend/async/AsyncBackendJobExecutionActor.scala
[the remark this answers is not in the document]
Horneth: https://github.com/broadinstitute/cromwell/issues/2813

wom/src/main/scala/wom/graph/GraphNodePort.scala
Horneth: Yeah this is what we discussed about having a port that contains the expression for task outputs so that we can return a `Map[OutputPort, WomValue]` when evaluating outputs and avoid having Cromwell do some string matching to find the right one. It has nothing to do with abort but I sneaked it in here.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/stores/ExecutionStore.scala
danbills: Very interesting, thanks for the in-depth explanation!  

In the interest of future me forgetting all this, could you please paste this explanation into a comment?

centaur/src/main/resources/standardTestCases/failures.restart_while_failing_tes.test
```diff
@@ -0,0 +1,27 @@
+name: failures.restart_while_failing_tes
+testFormat: WorkflowFailureRestartWithoutRecover
+callMark: restart_while_failing.B1
+backends: [TES]
+tags: [localdockertest]
```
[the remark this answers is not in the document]
Horneth: yeah I agree.. ultimately I think instead of targeting backends specifically we should categorize by feature. For example "only run this test on backends that support recover", or "support gcs" etc..

centaur/src/main/scala/centaur/test/Test.scala
```diff
@@ -133,7 +133,7 @@ object Operations {
     new Test[Unit] {
       override def run = {
         // Give some margin to the timeout
-        Try(Await.result(Future { Thread.sleep(duration.toMillis) }, duration.plus(2.seconds)))
+        Try(Await.result(Future { Thread.sleep(duration.toMillis) }, duration.plus(1.minute)))
```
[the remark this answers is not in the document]
Horneth: Yeah I'm not sure what's going on with this but I had timeouts here, which I don't see how could happen. It's just a sleep in a future, so I would expect that just a few seconds more would be enough but apparently not. I increased it a lot to see if it fixed and so far it did but I'll investigate more

centaur/src/main/scala/centaur/api/CentaurCromwellClient.scala
```diff
@@ -69,7 +70,8 @@ object CentaurCromwellClient {
     Try(Await.result(x(), timeout)) recoverWith {
       case _: TimeoutException |
            _: StreamTcpException |
-           _: IOException if !CromwellManager.isReady && attempt < 5 =>
+           _: IOException |
+           _: UnsupportedContentTypeException if !CromwellManager.isReady && attempt < 5 =>
```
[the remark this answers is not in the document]
Horneth: I imagine something changed in the cromwell client between when we ignored centaur tests (specifically restart ones) and now because now that I re-enabled restart tests, this is a new failure that appeared when trying to make a request to Cromwell while it's down. Apparently the response (whatever it is) is not json so the client fails to unmarshall with this exception.

PR #2810, opened 2017-11-01
Support `runtime` in ECMAScript Expressions

danbills:
Introduces a `RuntimeEnvironment` type in order to provide CWL expressions the proper values for the `runtime` ECMAscript variable.  

See spec for explicit detail: http://www.commonwl.org/v1.0/CommandLineTool.html#Runtime_environment

Also introduce `MinimumRuntimeSettings` as the spec says 
> "if an implementation can't provide the actual number of reserved cores during the expression evaluation time, it should report back the minimal requested amount."

I've made a few tradeoffs which I intend to document as tickets unless there are objections.  To be clear the tradeoff is these compromises for speed, as I'm trying to "spike" on 1st-workflow and get it working :

* MinimumRuntimeSettings should come from the config.  I see a major dependency tree coming all the way down from `RootCromwellActor` and I'm trying to think of a better way.  In this PR I've taken the shortcut of instantiating MinimumRuntimeSettings with default values hardcoded.
* The values of `outdirSize` and `tmpdirSize` are specified in CWL but I haven't yet figured out how to provide those values accurately.  I will likely create an issue to do this effectively as I doubt this is regularly used.
* I think we could constrain the types of `RuntimeEnvironment` better than the `String` and `Int` we are using currently, e.g. using `Path` and `MemorySize`.  This requires moving these types up to the `wom` package.

### DIFF
--- a/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
+++ b/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
@@ -21,7 +21,9 @@ import cromwell.services.metadata.CallMetadataKeys
 import common.exception.MessageAggregation
 import common.util.TryUtil
 import common.validation.ErrorOr.ErrorOr
+import cromwell.backend.io.JobPaths
 import net.ceedubs.ficus.Ficus._
+import wom.callable.RuntimeEnvironment
 import wom.values._
 
 import scala.concurrent.{ExecutionContext, ExecutionContextExecutor, Future, Promise}
@@ -224,8 +226,13 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
   }
 
   /** The instantiated command. */
-  lazy val instantiatedCommand: String = Command.instantiate(
-    jobDescriptor, backendEngineFunctions, commandLinePreProcessor, commandLineValueMapper).get
+  lazy val instantiatedCommand: String =  {
+
+    jobDescriptor.runtimeAttributes
+
+    Command.instantiate(
+      jobDescriptor, backendEngineFunctions, commandLinePreProcessor, commandLineValueMapper, jobPaths.callRoot).get
+  }
 
   /**
     * Redirect the stdout and stderr to the appropriate files. While not necessary, mark the job as not receiving any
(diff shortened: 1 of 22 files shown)


cwl/src/main/scala/cwl/Decoder.scala
```diff
@@ -37,7 +37,12 @@ object CwlDecoder {
   }
 
   def parseJson(json: String): Parse[Cwl] =
-    EitherT{IO{CwlCodecs.decodeCwl(json).leftMap(_.getMessage).leftMap(NonEmptyList.one)} }
+    EitherT{IO{
+      CwlCodecs.decodeCwl(json).
+        leftMap{t =>
+          NonEmptyList.of(t.getMessage, t.getStackTrace.mkString("\n"))
```
Horneth: ToL: shouldn't we keep the the actual exception at this point instead ?

cwl/src/main/scala/cwl/CwlExpressionCommandPart.scala
```diff
@@ -10,9 +11,15 @@ import wom.values._
...
+                           valueMapper: (WomValue) => WomValue,
+                           runtimeEnvironment: RuntimeEnvironment ): String = {
 
-    val pc = ParameterContext.Empty.withInputs(inputsMap.map({ case (LocalName(localName), value) => localName -> value }), functions)
+    val stringKeyMap = inputsMap.map{ case (LocalName(localName), value) => localName -> value }
+
+    val pc =
+      ParameterContext(
+        runtime = runtimeEnvironment.cwlMap).
+        withInputs(stringKeyMap, functions)
```
Horneth: ToL: this formatting is a bit strange it makes it look like `withInputs` is an attribute of `ParameterContext`.
Can we do
```scala
val pc =
  ParameterContext(
     runtime = runtimeEnvironment.cwlMap
  ).withInputs(stringKeyMap, functions)
```

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -224,8 +226,19 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
...
-    jobDescriptor, backendEngineFunctions, commandLinePreProcessor, commandLineValueMapper).get
+  lazy val instantiatedCommand: String =  {
+
+    val todoMoveMeMinimums =
+      MinimumRuntimeSettings(
+        cores = 1,
+        ram = MemorySize.apply(4, MemoryUnit.GiB),
+        outputPathSize = Long.MaxValue,
+        tempPathSize = Long.MaxValue
+      )
```
Horneth: Isn't this `MinimumRuntimeSettings()` ?

cwl/src/main/scala/cwl/CwlExpressionCommandPart.scala
danbills: Ew.  yep I'll fix this up

cwl/src/main/scala/cwl/Decoder.scala
danbills: Well we are working within the `Parse` type as defined as:
`EitherT[IO, NonEmptyList[String], A]`

I suppose we could go w/ exceptions in the `Nel` type.  Need to think about it a little more.

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
danbills: whoops, I meant to remove this and take in the argument :)

danbills:
@Horneth right, this does not add support for Resource Requirement.

cwl/src/main/scala/cwl/Decoder.scala
Horneth: Yeah this could also be punted as part of the "harmonize monads" tech debt.
I just think that whatever we end up choosing it should leave the possibility for keeping an exception as is when it makes sense, without having to "convert" it to an ErrorOr.

Horneth:
@danbills cool thanks !

PR #2819, opened 2017-11-02
swap running jobs to aborting in the workflow store if abort-on-termi…

Horneth:
…nate is true

It doesn't matter in run mode since it apparently uses an in-memory workflow store, but in the unlikely case in which someone sets `abort-on-terminate` to `true` in server mode, we want to set the status of all running jobs in the workflow store to `Aborting` so that when we restart the server we don't keep running those jobs.

It also fixes a bug which was preventing run mode to exit properly.

### DIFF
--- a/engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreEngineActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreEngineActor.scala
@@ -1,6 +1,6 @@
 package cromwell.engine.workflow.workflowstore
 
-import akka.actor.{ActorLogging, ActorRef, LoggingFSM, Props}
+import akka.actor.{ActorLogging, ActorRef, LoggingFSM, PoisonPill, Props}
 import cats.data.NonEmptyList
 import cromwell.core.Dispatcher._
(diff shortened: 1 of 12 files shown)


geoffjentry:
👍

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreEngineActor.scala
```diff
@@ -100,6 +107,11 @@ final case class WorkflowStoreEngineActor private(store: WorkflowStore, serviceR
             // given we don't know much about what went wrong here.  `t.getMessage` so the cause propagates to the client.
             sndr ! WorkflowAbortFailureResponse(id, new RuntimeException(s"$message: ${t.getMessage}", t))
         }
+      case AbortAllRunningWorkflowsCommandAndStop =>
+        store.abortAllRunning() map { _ =>
+          log.info(s"Aborting all running workflows.")
+          self ! PoisonPill
```
mcovarr: why a `PoisonPill` here?

Horneth: because this is happening in the `map` of a future and I'm worried about calling `context stop self` in here. But I do want the actor to stop once that's done.

PR #2821, opened 2017-11-03
Added CentaurCwlRunner.

kshakir:
If a java.util.concurrent.ExecutionException, check if it has an inner cause that is actually fatal.
Pre-update to SBT 1.x.
The cromwell multi-project with SBT 1.x consumes a lot of memory. To avoid OOME:
 - Bump IntelliJ from 768 to something like 1786.
 - Bump .sbtopts to something like 3072.

Build updates while prepping fo sbt version bump, including using sbt-doge temporarily for simplified SBT 0.13 cross version build syntax.
DRYed out a bit of the module dependency graph.
Publish hotfix libraries, not just executables.
Replaced deprecated commons-lang with commons-text replacements.
Fixed test class errantly nested within its companion object.
Fixed test that was only liquibasing metadata, and not the original database.
Removed unused TestActorSystem.
Fixed missing test in "common".
Removed old "common" sbt files.
Cleaner version of assembly.
SwaggerUI injected into the cromwell-version.conf, so it only needs to be editted in one place.
Replaced usage of 'sbt "project foo" sometask' with 'sbt foo/sometask'.


services/src/test/scala/cromwell/services/metadata/impl/MetadataDatabaseAccessSpec.scala
```diff
@@ -38,9 +38,16 @@ class MetadataDatabaseAccessSpec extends FlatSpec with Matchers with ScalaFuture
...
+      override val metadataDatabaseInterface = {
+        val databaseConfig = ConfigFactory.load.getConfig(configPath)
+
+        // NOTE: EngineLiquibaseSettings **MUST** always run before the MetadataLiquibaseSettings
+        new EngineSlickDatabase(databaseConfig)
+          .initialized(EngineServicesStore.EngineLiquibaseSettings)
+
+        new MetadataSlickDatabase(databaseConfig)
           .initialized(MetadataServicesStore.MetadataLiquibaseSettings)
+      }
```
mcovarr: gotta ask what triggered this realization 🙂

kshakir: Running `sbt alltests:test` with sbt 1.0.3 on my laptop triggered the same error travis was seeing. Looking in the resulting mysql changelog rows, I could see that the metadata changelog had been run _before_ the default/engine changelog.

kshakir:
> Just OOI why did WdlParser change ?

`org.apache.commons.lang3.StringEscapeUtils` is [deprecated](http://commons.apache.org/proper/commons-lang/javadocs/api-3.6/org/apache/commons/lang3/StringEscapeUtils.html) and was producing a compiler warning. Running the "Hermes command" at the top of the generated file, I replaced the import with the new `org.apache.commons.text.StringEscapeUtils`.

services/src/test/scala/cromwell/services/metadata/impl/MetadataDatabaseAccessSpec.scala
danbills: so does this mean we can upgrade sbt?

kshakir: Yep. I've tested with sbt 1.0.3 and LGTM.

During previous tests last weekend, I had to up IntelliJ's sbt memory from 768 (I used 1768M), and give my shell .sbtopts even more memory to run `sbt alltests:test` (I used 3072M), but I haven't retried recently since I also slimmed some of the redundant library dependencies.

centaur/src/main/scala/centaur/api/DaemonizedDefaultThreadFactory.scala
```diff
@@ -0,0 +1,25 @@
...
+
+import java.util.concurrent.ThreadFactory
+import java.util.concurrent.atomic.AtomicInteger
+
+/**
+  * A static version of java.util.concurrent.Executors.DefaultThreadFactory that creates daemon threads that exit when
+  * the application exits.
+  *
+  * java.util.concurrent.Executors.DefaultThreadFactory
+  * http://dev.bizo.com/2014/06/cached-thread-pool-considered-harmlful.html
```
geoffjentry: Is this URL intended to be the explanation of why this thing exists? If so can you give a brief summary. If not, can you add an explanation of why this is here?

geoffjentry:
👍

centaur/src/main/scala/centaur/api/DaemonizedDefaultThreadFactory.scala
kshakir: Will do. TL;DR: non-daemon threads exit when they want to, not when a user presses control-C or System.exit() is called.

project/Settings.scala
```diff
@@ -109,42 +109,57 @@ object Settings {
...
 
         // If you use the 'exec' form for an entry point, shell processing is not performed and
         // environment variable substitution does not occur.  Thus we have to /bin/bash here
         // and pass along any subsequent command line arguments
         // See https://docs.docker.com/engine/reference/builder/#/entrypoint
-        entryPoint("/bin/bash", "-c", "java ${JAVA_OPTS} -jar /app/cromwell.jar ${CROMWELL_ARGS} ${*}", "--")
+        entryPoint(
+          "/bin/bash",
+          "-c",
+          s"java $${JAVA_OPTS} -jar /app/$artifactName.jar $${${artifactName.toUpperCase}_ARGS} $${*}",
```
mcovarr: The more stringent compiler flags weren't being applied to Centaur before this PR? 😦

centaurCwlRunner/src/main/scala/centaur/cwl/CentaurCwlRunner.scala
```diff
@@ -0,0 +1,161 @@
...
+      def status: Int = exitCodeValue.asInstanceOf[Val].status
+    }
+
+    val Success = Val(0)
+    val Failure = Val(1)
+    val NotImplemented = Val(33)
+  }
+
+  private val parser = buildParser()
+  private lazy val centaurVersion = ConfigFactory.load("cromwell-version.conf").getString("version.cromwell")
```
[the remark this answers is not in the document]
kshakir: The version is "required" by the CWL conformance test suite.

This does read strangely, but is easy to clean up in the future. I didn't want to change it yet. Specifically, each artifact could write _its_ version instead of the "cromwell", aka "root" version.

Right now, the two artifacts that are writing versions are "cromwell-engine" and "centaur-cwl-runner". I didn't want to change too much at the moment to start reading from `cromwell-engine-version.conf`, but perhaps in a future PR.

project/Settings.scala
kshakir: The highlighted change specifically allows us to build dockers for any executable within the monorepo, replacing `cromwell` with `$artifactName`. AFAIK, the above isn't related to the compiler flags, but maybe I missed your question.

mcovarr: Hmm I thought you had to add the "dollar doubling" here to silence the "fatal warnings" about missing an 's' when the compiler thought you were trying to do interpolations. But perhaps the double dollars were added because you explicitly turned on interpolation to be able to get the artifact name in. 😄

PR #2828, opened 2017-11-06
Handle globs in cwl output expressions

danbills:
CWL was treating output glob strings as if they were filenames, and thus was not returning the filename that Cromwell expects, namely `glob-${md5(fileName)}.list`.

The implementation boils down to `OutputEvaluator` trying to detect whether the output of the expression is a glob.  If it is _is_ a glob, it changes the output to be the filename as listed above.

### DIFF
--- a/backend/src/main/scala/cromwell/backend/io/GlobFunctions.scala
+++ b/backend/src/main/scala/cromwell/backend/io/GlobFunctions.scala
@@ -40,11 +40,10 @@ trait GlobFunctions extends IoFunctionSet {
     *
     * The paths are currently read from a list file based on the pattern, and the path parameter is not used.
     *
-    * @param path    The path string returned by globPath. This isn't currently used.
     * @param pattern The pattern of the glob. This is the same "glob" passed to globPath().
     * @return The paths that match the pattern.
     */
-  override def glob(path: String, pattern: String): Seq[String] = {
+  override def glob(pattern: String): Seq[String] = {
     val globPatternName = globName(pattern)
     val listFilePath = callContext.root.resolve(s"${globName(pattern)}.list")
     // This "lines" is technically a read file and hence should use the readFile IO method
--- a/backend/src/main/scala/cromwell/backend/io/GlobFunctions.scala
+++ b/backend/src/main/scala/cromwell/backend/io/GlobFunctions.scala
@@ -22,9 +22,7 @@ trait GlobFunctions extends IoFunctionSet {
       }
     }
 
-  def globDirectory(glob: String): String = globName(glob) + "/"
-  def globName(glob: String) = s"glob-${glob.md5Sum}"
-
+  def globDirectory(glob: String): String = GlobFunctions.globName(glob) + "/"
   /**
     * Returns a path to the glob.
     *
@@ -44,6 +42,7 @@ trait GlobFunctions extends IoFunctionSet {
     * @return The paths that match the pattern.
     */
   override def glob(pattern: String): Seq[String] = {
+    import GlobFunctions._
     val globPatternName = globName(pattern)
     val listFilePath = callContext.root.resolve(s"${globName(pattern)}.list")
     // This "lines" is technically a read file and hence should use the readFile IO method
@@ -51,4 +50,8 @@ trait GlobFunctions extends IoFunctionSet {
       (callContext.root /  globPatternName  / fileName).pathAsString
     }
   }
+}
+
+object GlobFunctions {
+  def globName(glob: String) = s"glob-${glob.md5Sum}"
 }
--- a/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
+++ b/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
@@ -13,7 +13,7 @@ import cromwell.backend.async.AsyncBackendJobExecutionActor._
 import cromwell.backend.async.{AbortedExecutionHandle, AsyncBackendJobExecutionActor, ExecutionHandle, FailedNonRetryableExecutionHandle, FailedRetryableExecutionHandle, PendingExecutionHandle, ReturnCodeIsNotAnInt, StderrNonEmpty, SuccessfulExecutionHandle, WrongReturnCode}
 import cromwell.backend.validation._
 import cromwell.backend.wdl.OutputEvaluator._
-import cromwell.backend.wdl.{Command, OutputEvaluator, WdlFileMapper}
+import cromwell.backend.wdl.{Command, OutputEvaluator, WomFileMapper}
 import cromwell.backend._
 import cromwell.core.io.{AsyncIo, DefaultIoCommandBuilder}
 import cromwell.core.path.Path
@@ -116,9 +116,11 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
 
   /** @see [[Command.instantiate]] */
   final lazy val commandLinePreProcessor: WomEvaluatedCallInputs => Try[WomEvaluatedCallInputs] = {
-    inputs => TryUtil.sequenceMap(inputs mapValues WdlFileMapper.mapWdlFiles(preProcessWdlFile)) recoverWith {
-      case e => Failure(new IOException(e.getMessage) with CromwellFatalExceptionMarker)
-    }
+    inputs =>
+      TryUtil.sequenceMap(inputs mapValues WomFileMapper.mapWomFiles(preProcessWdlFile)).
+        recoverWith {
+          case e => Failure(new IOException(e.getMessage) with CromwellFatalExceptionMarker)
+        }
   }
 
   /**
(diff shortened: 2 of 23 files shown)


backend/src/main/scala/cromwell/backend/wdl/OutputEvaluator.scala
```diff
@@ -40,9 +41,17 @@ object OutputEvaluator {
...
       // Attempt to evaluate the expression using all known values
-      def evaluateOutputExpression: OutputResult[WomValue] = {
-        EitherT { Try(output.expression.evaluateValue(allKnownValues, ioFunctions)).map(_.toEither) }
-      }
+      def evaluateOutputExpression: OutputResult[WomValue] = EitherT{ Try{
+        output.expression.evaluateValue(allKnownValues, ioFunctions).toEither.map{
+
+          //If the string contained in a filename is actually a glob,
+          // The output of this expression is expected to be glob-md5(fileName).list
+          case WomSingleFile(fileName) if fileName.contains("*") =>
```
Horneth: I'm curious as why we need this, I would have thought that the fact that the expression contains a glob had already been handled in the `evaluateValue`, so by the time we get here we don't have to do any "is this a glob" checks. Also treating any filename with * as a glob might have unintended consequences as I'm sure someone somewhere named their file `my_awesome_file_with_a_*_in_the_name.boom`

backend/src/main/scala/cromwell/backend/io/GlobFunctions.scala
```diff
@@ -40,11 +38,11 @@ trait GlobFunctions extends IoFunctionSet {
     *
     * The paths are currently read from a list file based on the pattern, and the path parameter is not used.
     *
-    * @param path    The path string returned by globPath. This isn't currently used.
     * @param pattern The pattern of the glob. This is the same "glob" passed to globPath().
     * @return The paths that match the pattern.
     */
-  override def glob(path: String, pattern: String): Seq[String] = {
+  override def glob(pattern: String): Seq[String] = {
```
Horneth: thanks for bringing some sanity to this

backend/src/main/scala/cromwell/backend/wdl/OutputEvaluator.scala
danbills: I had originally written this within `evaluateValue`, but it is really custom to cromwell impl so it seemed like it belonged here.

Part of the reason for the talk tomorrow is that I am not particularly happy with this solution and I need some fresh eyes on this problem.

cjllanwarne: This seems a little off to me:
- The output evaluator should be returning the final value, not some intermediate thing in need of further interpretation.
- As part of evaluation, it should notice that it's using a glob and do the stuff with the `.list` file behind the scenes
- Maybe the standard `IoFunctions` that Cromwell passes to the languages should contain a `resolveGlobOutput` method from `String` to `Array[File]`?

wdl/src/main/scala/wdl/expression/FileEvaluator.scala
```diff
@@ -34,9 +34,8 @@ class FileEvaluatorWdlFunctions
  * evaluate("b.txt") would return Seq().
  */
 case class FileEvaluator(valueEvaluator: ValueEvaluator, coerceTo: WomType = WomAnyType) {
-  type T = Seq[WomFile]
-  def lookup = (s:String) => Seq.empty[WomFile]
 
+  def lookup = (s:String) => Seq.empty[WomFile]
```
cjllanwarne: Would this work like this? Feels nicer to me...:
```scala
def lookup: String => Seq[WomFile] = _ => Seq.empty
```

backend/src/main/scala/cromwell/backend/wdl/OutputEvaluator.scala
danbills: `IoFunctions` has 
```scala
glob(pattern: String): Seq[String]
```
However the Cromwell implementation is expecting to see a `.list` file from this output evaluation, and it does the dereferencing downstream of this output

wdl/src/main/scala/wdl/expression/FileEvaluator.scala
danbills: Hm.  Not sure why this is highlighted, I didn't change this line.  I deleted the one above it.

(FWIW I agree)

backend/src/main/scala/cromwell/backend/wdl/OutputEvaluator.scala
Horneth: Totally not your fault but if that's the case then I would tend to say it shouldn't be. The whole point of `IoFunctions` is that they are provided by the engine which gives it full flexibility as to how it wants to perform them, and if that includes dereferencing things it should be done as part of this method IMO.

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -156,22 +162,28 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
...
     * Returns the shell scripting for hard linking a glob results using ln.
     *
     * @param globFile The glob.
     * @return The shell scripting.
     */
-  def globManipulation(globFile: WomGlobFile): String = {
+  def globScript(globFile: WomGlobFile): String = {
     val parentDirectory = globParentDirectory(globFile)
-    val globDir = backendEngineFunctions.globName(globFile.value)
+    val globDir = GlobFunctions.globName(globFile.value)
```
cjllanwarne: Just to make sure, there was no backend-specific code in any `globName` function override?

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -156,22 +162,28 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
...
     val globList = parentDirectory./(s"$globDir.list")
 
-    s"""|mkdir $globDirectory
+    s"""|# make the directory which will keep the matching files
+        |mkdir $globDirectory
+        |
+        |# symlink all the files into the glob directory
         |( ln -L ${globFile.value} $globDirectory 2> /dev/null ) || ( ln ${globFile.value} $globDirectory )
+        |
+        |# list all the files that match teh glob into a file called glob-[md5 of glob].list
```
cjllanwarne: `teh glob`

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -211,7 +224,7 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
         |echo $$? > $rcTmpPath
         |(
         |cd $cwd
-        |${globManipulations(globFiles)}
+        |${globScripts(globFiles)}
```
cjllanwarne: ToL: I'm not sure `scripts` is correct, they're more like `snippets`... 🤷‍♂️

backend/src/main/scala/cromwell/backend/wdl/WomFileMapper.scala
```diff
@@ -5,29 +5,29 @@ import wom.values._
 
 import scala.util.{Success, Try}
 
-object WdlFileMapper {
-  def mapWdlFiles(mapper: (WomFile => WomFile))(womValue: WomValue): Try[WomValue] = {
+object WomFileMapper {
```
cjllanwarne: This is still in a package called `cromwell.backend.wdl`

cwl/src/main/scala/cwl/CwlWomExpression.scala
```diff
@@ -22,16 +22,29 @@ case class CommandOutputExpression(outputBinding: CommandOutputBinding,
 
   // TODO WOM: outputBinding.toString is probably not be the best representation of the outputBinding
   override def sourceString = outputBinding.toString
+
   override def evaluateValue(inputValues: Map[String, WomValue], ioFunctionSet: IoFunctionSet): ErrorOr[WomValue] = {
     val parameterContext = ParameterContext.Empty.withInputs(inputValues, ioFunctionSet)
 
-    val wdlValue: WomValue = outputBinding.commandOutputBindingToWdlValue(parameterContext, ioFunctionSet)
+    val wdlValue: WomValue = outputBinding.
```
cjllanwarne: Not a `wdl` value

cwl/src/main/scala/cwl/CwlWomExpression.scala
```diff
@@ -22,16 +22,29 @@ case class CommandOutputExpression(outputBinding: CommandOutputBinding,
...
+      (extractFile, cwlExpressionType) match {
+        case (WomString(glob), WomFileType) =>
+          ioFunctionSet.glob(glob) match {
+            case head :: Nil => WomString(head)
+            case list => throw new RuntimeException(s"expecting a single File glob but instead got $list")
+          }
+        case _  => extractFile
+      }
+
+    cwlExpressionType.coerceRawValue(globIfFile).toErrorOr
```
cjllanwarne: Some comments would be nice here. I'm not sure I followed what's happening

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
danbills: Confirming there was no backend-specific code.

PR #2829, opened 2017-11-06
Abort harder

Horneth:
In situations where a workflow is being restarted, receives an abort command but is only at materialization stage or initialization stage we currently assume it's ok to stop right here and declare it `Aborted`. This assumption is wrong as jobs are probably running and we need to re-connect to them so we can 1) abort them 2) update their status.
In the same vein, if a workflow is in the workflow store waiting to be restarted and gets aborted, we need to still pick it up and reconnect to jobs to abort them.


database/sql/src/main/scala/cromwell/database/slick/WorkflowStoreSlickDatabase.scala
```diff
@@ -13,19 +15,31 @@ trait WorkflowStoreSlickDatabase extends WorkflowStoreSqlDatabase {
...
-  override def updateWorkflowState(workflowId: String, newWorkflowState: String)
-                                     (implicit ec: ExecutionContext): Future[Int] = {
-    val action = dataAccess.workflowStateForId(workflowId).update(newWorkflowState)
+  override def initializeRestartedFlags()
+                                       (implicit ec: ExecutionContext): Future[Unit] = {
+    val action = for {
+      _ <- dataAccess.restartedFlagForRunningAndAborting.update(true)
+    } yield ()
+
+    runTransaction(action) void
```
mcovarr:
```
    val action = dataAccess.restartedFlagForRunningAndAborting.update(true)
    runTransaction(action) void
```

database/sql/src/main/scala/cromwell/database/slick/WorkflowStoreSlickDatabase.scala
```diff
@@ -13,19 +15,31 @@ trait WorkflowStoreSlickDatabase extends WorkflowStoreSqlDatabase {
 
   import dataAccess.driver.api._
 
-  override def updateWorkflowsInState(updates: List[(String, String)])
-                                     (implicit ec: ExecutionContext): Future[Unit] = {
-    val updateQueries = updates.map({
-      case (oldState, newState) => dataAccess.workflowStateForWorkflowState(oldState).update(newState)
-    })
+  override def abortAllRunning()
```
mcovarr: The name of this method makes it sound like the workflows will have been aborted after it returns; perhaps it should have a name more consistent with only marking the workflows as aborting.

database/sql/src/main/scala/cromwell/database/slick/WorkflowStoreSlickDatabase.scala
```diff
@@ -13,19 +15,31 @@ trait WorkflowStoreSlickDatabase extends WorkflowStoreSqlDatabase {
...
+      .update(WorkflowStoreState.Aborting)
 
-    val action = DBIO.sequence(updateQueries)
     runTransaction(action) void
   }
 
-  override def updateWorkflowState(workflowId: String, newWorkflowState: String)
-                                     (implicit ec: ExecutionContext): Future[Int] = {
-    val action = dataAccess.workflowStateForId(workflowId).update(newWorkflowState)
+  override def initializeRestartedFlags()
```
mcovarr: It's not really clear what 'initialize' means, perhaps 'markRunningAndAbortingWorkflowsAsRestarting' or a less awkward equivalent of that

database/sql/src/main/scala/cromwell/database/slick/WorkflowStoreSlickDatabase.scala
```diff
@@ -13,19 +15,31 @@ trait WorkflowStoreSlickDatabase extends WorkflowStoreSqlDatabase {
...
+                                       (implicit ec: ExecutionContext): Future[Unit] = {
+    val action = for {
+      _ <- dataAccess.restartedFlagForRunningAndAborting.update(true)
+    } yield ()
+
+    runTransaction(action) void
+  }
+
+  override def abort(workflowId: String)
```
mcovarr: Naming again, something less certain-sounding.

database/sql/src/main/scala/cromwell/database/slick/WorkflowStoreSlickDatabase.scala
```diff
@@ -35,22 +49,29 @@ trait WorkflowStoreSlickDatabase extends WorkflowStoreSqlDatabase {
...
   }
 
-  private def updateWorkflowStateForWorkflowExecutionUuid(updateWorkflowState: String)
-                                                         (workflowStoreEntry: WorkflowStoreEntry)
+  private def updateWorkflowStateAndRestartedForWorkflowExecutionUuid(workflowStoreEntry: WorkflowStoreEntry)
                                                          (implicit ec: ExecutionContext): DBIO[Unit] = {
     val workflowExecutionUuid = workflowStoreEntry.workflowExecutionUuid
+    val updateState = workflowStoreEntry.workflowState match {
+        // Submitted workflows become running when fetched
+      case WorkflowStoreState.Submitted => WorkflowStoreState.Running
```
mcovarr: This isn't saying that all submitted workflows go to running on restart, does it?

database/sql/src/main/scala/cromwell/database/slick/tables/WorkflowStoreEntryComponent.scala
```diff
@@ -56,13 +69,15 @@ trait WorkflowStoreEntryComponent {
...
-  val workflowStoreEntriesForWorkflowState = Compiled(
-    (workflowState: Rep[String], limit: ConstColumn[Long]) => {
+  val fetchStartableWorkflows = Compiled(
+    (limit: ConstColumn[Long]) => {
       val query = for {
         workflowStoreEntryRow <- workflowStoreEntries
-        if workflowStoreEntryRow.workflowState === workflowState
+        if (workflowStoreEntryRow.workflowState === WorkflowStoreState.Submitted && workflowStoreEntryRow.restarted === false) ||
+           (workflowStoreEntryRow.workflowState === WorkflowStoreState.Running && workflowStoreEntryRow.restarted === true) ||
+           (workflowStoreEntryRow.workflowState === WorkflowStoreState.Aborting && workflowStoreEntryRow.restarted === true)
```
mcovarr: it might be nice to give a precedence: aborting, running, submitted

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -234,7 +236,8 @@ class WorkflowActor(val workflowId: WorkflowId,
       pushWorkflowStart(workflowId)
       actor ! MaterializeWorkflowDescriptorCommand(workflowSourceFilesCollection, conf)
       goto(MaterializingWorkflowDescriptorState) using stateData.copy(currentLifecycleStateActor = Option(actor))
-    case Event(AbortWorkflowCommand, _) => goto(WorkflowAbortedState)
+    // If the workflow is not being restarted - then we can abort it immediately as nothing happened yet
```
mcovarr: not sure what's up with the dash here 🙂

engine/src/main/scala/cromwell/engine/workflow/WorkflowActor.scala
```diff
@@ -249,17 +252,16 @@ class WorkflowActor(val workflowId: WorkflowId,
       goto(InitializingWorkflowState) using data.copy(currentLifecycleStateActor = Option(initializerActor), workflowDescriptor = Option(workflowDescriptor))
     case Event(MaterializeWorkflowDescriptorFailureResponse(reason: Throwable), data) =>
       goto(WorkflowFailedState) using data.copy(lastStateReached = StateCheckpoint(MaterializingWorkflowDescriptorState, Option(List(reason))))
-    case Event(AbortWorkflowCommand, _) =>
-      // No lifecycle sub-actors exist yet, so no indirection via WorkflowAbortingState is necessary:
-      goto(WorkflowAbortedState)
+    // If the workflow is not being restarted - then we can abort it immediately as nothing happened yet
```
mcovarr: more dashing commentary

database/sql/src/main/scala/cromwell/database/slick/WorkflowStoreSlickDatabase.scala
Horneth: No it's the state the workflow is updated to when it's fetched by the WMA. Submitted becomes Running, the rest (Running or Aborting with restarted = true) stay the same (but the restarted flag is set to false)

Horneth:
@mcovarr On top of addressing your comments I added a small migration as I realized it's technically possible for some workflows to be in `RestartableRunning` or `RestartableAborting` when Cromwell starts, which are now replaced by `Running` and `Aborting` with the restarted flag to `true`. If you don't mind re-taking a look at the liquibase :)

mcovarr:
@Horneth LGTM 👍

database/migration/src/main/resources/changesets/workflow_store_restarted_column.xml
```diff
@@ -0,0 +1,31 @@
...
+        </update>
+
+        <!--RestartableAborting to Aborting-->
+        <update
+                tableName="WORKFLOW_STORE_ENTRY">
+            <column name="WORKFLOW_STATE" value="Aborting" />
+            <where>WORKFLOW_STATE = 'RestartableAborting'</where>
+        </update>
+
+        <!--Note: there's not need to set the flags to true because that's the first thing Cromwell will do when it starts-->
```
cjllanwarne: `not` ➡️  `no`

database/sql/src/main/scala/cromwell/database/slick/WorkflowStoreSlickDatabase.scala
```diff
@@ -13,19 +15,29 @@ trait WorkflowStoreSlickDatabase extends WorkflowStoreSqlDatabase {
 
   import dataAccess.driver.api._
 
-  override def updateWorkflowsInState(updates: List[(String, String)])
-                                     (implicit ec: ExecutionContext): Future[Unit] = {
-    val updateQueries = updates.map({
-      case (oldState, newState) => dataAccess.workflowStateForWorkflowState(oldState).update(newState)
-    })
+  override def setAllRunningToAborting()
+                                      (implicit ec: ExecutionContext): Future[Unit] = {
```
cjllanwarne: ToL: Do we need to preserve the `()` and/or second line?

database/sql/src/main/scala/cromwell/database/slick/WorkflowStoreSlickDatabase.scala
```diff
@@ -13,19 +15,29 @@ trait WorkflowStoreSlickDatabase extends WorkflowStoreSqlDatabase {
...
+    val action = dataAccess
+      .workflowStateForWorkflowState(WorkflowStoreState.Running)
+      .update(WorkflowStoreState.Aborting)
+
+    runTransaction(action) void
```
cjllanwarne: ToL: Can we use `.`s and remove the unfashionable `import scala.language.postfixOps`

database/sql/src/main/scala/cromwell/database/slick/WorkflowStoreSlickDatabase.scala
```diff
@@ -13,19 +15,29 @@ trait WorkflowStoreSlickDatabase extends WorkflowStoreSqlDatabase {
...
+  }
+
+  override def markRunningAndAbortingAsRestarted()
+                                                (implicit ec: ExecutionContext): Future[Unit] = {
```
cjllanwarne: ToL: Same comment re `()` and newlines

database/sql/src/main/scala/cromwell/database/slick/WorkflowStoreSlickDatabase.scala
```diff
@@ -13,19 +15,29 @@ trait WorkflowStoreSlickDatabase extends WorkflowStoreSqlDatabase {
...
 
-  override def updateWorkflowState(workflowId: String, newWorkflowState: String)
-                                     (implicit ec: ExecutionContext): Future[Int] = {
-    val action = dataAccess.workflowStateForId(workflowId).update(newWorkflowState)
+  override def setToAborting(workflowId: String)
+                            (implicit ec: ExecutionContext): Future[Option[Boolean]] = {
+    val action =  for {
+      restarted <- dataAccess.workflowRestartedForId(workflowId).result.headOption
+      _ <- dataAccess.workflowStateForId(workflowId).update(WorkflowStoreState.Aborting)
+    } yield restarted
```
cjllanwarne: Could you comment what `setToAborting` has to do with `"restarting"`?

database/sql/src/main/scala/cromwell/database/slick/tables/WorkflowStoreEntryComponent.scala
```diff
@@ -106,4 +131,14 @@ trait WorkflowStoreEntryComponent {
       if workflowStoreEntry.workflowExecutionUuid === workflowId
     } yield workflowStoreEntry.workflowState
   )
+
+  /**
+    * Useful for updating a given workflow to a new state
```
cjllanwarne: Really?

engine/src/main/scala/cromwell/engine/workflow/workflowstore/SqlWorkflowStore.scala
```diff
@@ -53,9 +45,13 @@ case class SqlWorkflowStore(sqlDatabase: WorkflowStoreSqlDatabase) extends Workf
...
-  override def fetchRunnableWorkflows(n: Int, state: StartableState)(implicit ec: ExecutionContext): Future[List[WorkflowToStart]] = {
-    sqlDatabase.queryWorkflowStoreEntries(n, state.toString, state.afterFetchedState.toString) map {
-      _.toList map fromWorkflowStoreEntry
+  override def fetchRunnableWorkflows(n: Int)(implicit ec: ExecutionContext): Future[List[WorkflowToStart]] = {
+    import cats.instances.list._
+    import cats.syntax.traverse._
+    import common.validation.Validation._
+    sqlDatabase.fetchRunnableWorkflows(n) map {
+      // .get on purpose here to fail the future if something went wrong
+      _.toList.traverse(fromWorkflowStoreEntry).toTry.get
```
cjllanwarne: TIL that's a thing. Is it really not possible to do it nicelier though?

Horneth:
@cjllanwarne example scenario:

- Workflow starts
- a few jobs start
- cromwell is restarted
- immediately after, someone aborts this workflow

the workflow can be anywhere from still in the workflow store (waiting to be picked up and restarted) to executing. If it's anywhere but executing (could be materialization, workflow initialization...), the behavior right now is "we're not executing this workflow yet, which means we haven't started any jobs, so it's fine to just mark it aborted and stop". This doesn't work because this workflow does have running jobs that will 1) never be aborted 2) stay "running" forever in the metadata

database/sql/src/main/scala/cromwell/database/slick/WorkflowStoreSlickDatabase.scala
```diff
@@ -35,22 +47,29 @@ trait WorkflowStoreSlickDatabase extends WorkflowStoreSqlDatabase {
...
   }
 
-  override def queryWorkflowStoreEntries(limit: Int, queryWorkflowState: String, updateWorkflowState: String)
-                                        (implicit ec: ExecutionContext): Future[Seq[WorkflowStoreEntry]] = {
+  override def fetchRunnableWorkflows(limit: Int)
+                                     (implicit ec: ExecutionContext): Future[Seq[WorkflowStoreEntry]] = {
     val action = for {
-      workflowStoreEntries <- dataAccess.workflowStoreEntriesForWorkflowState((queryWorkflowState, limit.toLong)).result
-      _ <- DBIO.sequence(workflowStoreEntries map updateWorkflowStateForWorkflowExecutionUuid(updateWorkflowState))
+      workflowStoreEntries <- dataAccess.fetchStartableWorkflows(limit.toLong).result
```
geoffjentry: what's the difference between `runnable` and `startable` workflows?

Horneth: there isn't, I'll pick one to homogenize

engine/src/main/scala/cromwell/engine/workflow/workflowstore/SqlWorkflowStore.scala
Horneth: I could `flatMap` it and use `Future.fromTry` which would remove the `get` but I don't know how much nicer that would be ?

cjllanwarne: I meant I didn't realise you could throw in a `Future`'s `map` and everything would be fine... I guess it does make sense when I think about it

database/sql/src/main/scala/cromwell/database/slick/WorkflowStoreSlickDatabase.scala
Horneth: There's a comment in the method definition in the trait but I'll copy it here : 
```
/**
    * Set the workflow to aborting state.
    * @return Some(restarted) if the workflow exists in the store, where restarted is the value of its restarted flag
    *         None if the workflow does not exist in the store
    */
```
We need to know if this workflow is being restarted because if it's not and we haven't started it yet then we can just consider it aborted, but if is, we need to leave it there and pick it up to reconnect to running jobs (while making sure we know we want to abort it)

database/sql/src/main/scala/cromwell/database/slick/tables/WorkflowStoreEntryComponent.scala
Horneth: Blah I was just copy pasting it from other queries :p

docs/execution/WorkflowExecution.md
```diff
@@ -0,0 +1,96 @@
+This section explains at a high level how Cromwell runs a workflow, and how workflow completeness is determined.
+
+In order to run a workflow, Cromwell uses the backends available to it to create jobs and monitor them until they are complete. When no new jobs can or should be started, and all jobs have reached a terminal status (Success, Failure, Aborted), the workflow itself is terminated.
+
+In a ideal situation, all jobs succeed and the workflow succeeds.
+Unfortunately things don't always go as planned. Here is what to expect from Cromwell when things go off rail.
```
mcovarr: off rail => off the rails

docs/execution/WorkflowExecution.md
```diff
@@ -0,0 +1,96 @@
+This section explains at a high level how Cromwell runs a workflow, and how workflow completeness is determined.
+
+In order to run a workflow, Cromwell uses the backends available to it to create jobs and monitor them until they are complete. When no new jobs can or should be started, and all jobs have reached a terminal status (Success, Failure, Aborted), the workflow itself is terminated.
+
+In a ideal situation, all jobs succeed and the workflow succeeds.
```
geoffjentry: The doc is called "workflow execution" but 99% of it appears to be "something has gone horribly awry". Not a good message to be sending :)

docs/execution/WorkflowExecution.md
```diff
@@ -0,0 +1,96 @@
...
+## Failure Modes
+
+Cromwell supports two failure modes:
+
+* `NoNewCalls` (default)
+* `ContinueWhilePossible`
+
+They specify how Cromwell behaves when a job fails during the execution of a workflow.
+
+`NoNewCalls` will (fairly obviously) not start any new call as soon as a job fails. Cromwell will still monitor the rest of the jobs until they complete (successfully or not).
```
mcovarr: I'd probably remove the `(fairly obviously)` 🙂

docs/execution/WorkflowExecution.md
Horneth: hmm that's true, any suggestion ? :)

docs/execution/WorkflowExecution.md
```diff
@@ -0,0 +1,96 @@
...
+
+If the failure mode is `ContinueWhilePossible`, then A **will be** retried:
+
+![](CWP_B_fail_A_retryable.png)
+
+## Abort
+
+In both Run and Server mode, you have the ability to ask Cromwell to abort a running workflow. This section explains what that entails.
+
+When aborting workflow, either through the [abort endpoint](api/RESTAPI#abort-a-running-workflow) or by terminating the [Cromwell run process](Modes) (if [configured](Configuring#abort) to do so), Cromwell will take the following steps:
```
mcovarr: workflow => a workflow

docs/execution/WorkflowExecution.md
geoffjentry: If I had a good one I'd have made it .... "when good workflows go bad"? :)

docs/execution/WorkflowExecution.md
```diff
@@ -0,0 +1,96 @@
...
+
+- Change the status of the workflow to `Aborting`
+- Do not start any new job going forward
+- Ask every running job to abort
+- Wait for all running jobs to reach completion
+- Finalize the workflow
+- Change the status of the workflow to `Aborted`
+
+The action of aborting a job is backend specific. Cromwell can only ask a backend to abort a job and wait for the backend to notify when it is in fact aborted.
```
mcovarr: to notify => to notify it

docs/execution/WorkflowExecution.md
mcovarr: just keepin' it real

docs/execution/WorkflowExecution.md
```diff
@@ -0,0 +1,96 @@
...
+Note that by the time the backend is asked to abort a job, it might already be successful, or failed, in which case Cromwell will report the job as such.
+
+As an example, the Google backend when asked to abort a job will send an abort request to the Pipelines API. When Pipelines API indicates that the status of the job is aborted, Cromwell will mark it as such.
+However the abort action is entirely dependent on the backend, in this particular case, Pipelines API does not guarantee the success of an abort request (see [Pipelines API documentation on abort](https://cloud.google.com/genomics/reference/rest/v1alpha2/operations/cancel)).
+
+You'll also notice that the workflow is finalized even though being aborted.
+Finalization is the last step in the execution of a workflow and a chance for each backend to do some work before the workflow is terminated.
+Abort won't deny the chance for backends to finalize the workflow.
```
mcovarr: ugh active voice pedantry makes for some awkward sentences at times 🙂

docs/execution/WorkflowExecution.md
```diff
@@ -0,0 +1,96 @@
...
+
+If a job fails with a retryable failure (e.g is preempted), it will **not** be attempted again when the workflow is aborting.
+
+## Restart
+
+When Cromwell is run in server mode, it is likely that the server will have to be stopped and started again at some point (to upgrade Cromwell to a new version for example).
+In such a case, if the `workflow-restart` configuration option is set to `true` (it is by default), Cromwell will restart all jobs that were in progress.
+
+Specifically it will attempt to reconnect to all running jobs of a workflow.
+The ability to reconnect to an existing job is again backend dependent. The Google Backend and the HPC backends currently support it.
```
mcovarr: hmm I wonder if it works in Docker local? 🤔

docs/execution/WorkflowExecution.md
geoffjentry: Since the good captain didn't note it, "a ideal" => "an ideal"

docs/execution/WorkflowExecution.md
Horneth: Good question, maybe @kshakir knows ? Can we "recover" a docker job ?

mcovarr: I made some changes recently that record the container ID in the execution directory. Since the actually processes should run as children of the Docker daemon, I think this is at least theoretically possible.

mcovarr:
🎉

PR #2832, opened 2017-11-07
Yaml inputs

Horneth:
Accepts yaml input files during submission.
Whether the input is in yaml or json, it will be written in **json** in the database, as well as metadata.
Also `cats`ify things a little to get better error reporting on incorrect submit request.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,17 +5,6 @@
 ### Other changes
 
 * **API**  
-    + The submit endpoint will now present all errors in an array:
-
-    ```
-    $ curl -X POST -F foo=wrong -F bar=wrong_again http://localhost:8000/api/workflows/v1
-    {
-      "status": "fail",
-      "message": "Invalid submit request",
-      "errors": ["Unexpected body part name: foo", "Unexpected body part name: bar"]
-    }
-    ```
-
     + Cromwell now supports input files in the yaml format (json format is still supported).
 
 * **Database**  
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,20 @@
 
 ### Other changes
 
+* **API**  
+    + The submit endpoint will now present all errors in an array:
+
+    ```
+    $ curl -X POST -F foo=wrong -F bar=wrong_again http://localhost:8000/api/workflows/v1
+    {
+      "status": "fail",
+      "message": "Invalid submit request",
+      "errors": ["Unexpected body part name: foo", "Unexpected body part name: bar"]
+    }
+    ```
+
+    + Cromwell now supports input files in the yaml format (json format is still supported).
+
 * **Database**  
 You have the option of storing the metadata in a separate SQL database than the database containing the internal engine
 data. See the [README](https://github.com/broadinstitute/cromwell#database) or the `database` section in
--- a/backend/src/main/scala/cromwell/backend/async/AsyncBackendJobExecutionActor.scala
+++ b/backend/src/main/scala/cromwell/backend/async/AsyncBackendJobExecutionActor.scala
@@ -55,8 +55,7 @@ trait AsyncBackendJobExecutionActor { this: Actor with ActorLogging =>
     case _ => false
   }
 
-  // This seems dangerous as any exception that is not one of the above (fatal) will trigger infinite retries !
-  def isTransient(throwable: Throwable): Boolean = !isFatal(throwable)
+  def isTransient(throwable: Throwable): Boolean = false
 
   private def withRetry[A](work: () => Future[A], backOff: SimpleExponentialBackoff): Future[A] = {
     Retry.withRetry(work, isTransient = isTransient, isFatal = isFatal, backoff = backOff)(context.system)
--- a/common/src/main/scala/common/validation/Validation.scala
+++ b/common/src/main/scala/common/validation/Validation.scala
@@ -7,6 +7,7 @@ import cats.data.{NonEmptyList, Validated}
 import cats.syntax.validated._
 import cats.syntax.either._
 import common.Checked
+import common.exception.AggregatedMessageException
 import common.validation.ErrorOr.ErrorOr
 import net.ceedubs.ficus.readers.{StringReader, ValueReader}
 import org.slf4j.Logger
@@ -41,7 +42,11 @@ object Validation {
   implicit class ValidationTry[A](val e: ErrorOr[A]) extends AnyVal {
     def toTry: Try[A] = e match {
       case Valid(options) => Success(options)
-      case Invalid(err) => Failure(new RuntimeException(s"Error(s): ${err.toList.mkString(",")}"))
+      case Invalid(err) => Failure(AggregatedMessageException("Error(s)", err.toList))
+    }
+    def toTry(context: String): Try[A] = e match {
+      case Valid(options) => Success(options)
+      case Invalid(err) => Failure(AggregatedMessageException(context, err.toList))
     }
   }
 }
--- a/engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
+++ b/engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
@@ -87,11 +87,12 @@ object PartialWorkflowSources {
 
   private def workflowInputs(data: String): ErrorOr[Vector[WorkflowJson]] = {
     import cats.syntax.validated._
+    import _root_.io.circe.Printer
     
     yaml.parser.parse(data) match {
       // If it's an array, treat each element as an individual input object, otherwise simply toString the whole thing
-      case Right(json) => json.asArray.map(_.map(_.toString())).getOrElse(Vector(json.toString)).validNel
-      case Left(error) => s"Input file is not valid json: ${error.getMessage}".invalidNel
+      case Right(json) => json.asArray.map(_.map(_.toString())).getOrElse(Vector(json.pretty(Printer.noSpaces))).validNel
+      case Left(error) => s"Input file is not valid yaml nor json: ${error.getMessage}".invalidNel
     }
   }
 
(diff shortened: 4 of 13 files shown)


geoffjentry:
Heads up that updating the cromwell.yaml doesn't update the docs. We can mostly autogenerate docs but @katevoss did some hand editing.

CHANGELOG.md
```diff
@@ -4,6 +4,20 @@
 
 ### Other changes
 
+* **API**  
+    + The submit endpoint will now present all errors in an array:
```
cjllanwarne: Can we run this change past a fireclouder? It seems like it might be more obnoxious for them than it would appear to us...

backend/src/main/scala/cromwell/backend/async/AsyncBackendJobExecutionActor.scala
```diff
@@ -55,8 +55,7 @@ trait AsyncBackendJobExecutionActor { this: Actor with ActorLogging =>
     case _ => false
   }
 
-  // This seems dangerous as any exception that is not one of the above (fatal) will trigger infinite retries !
-  def isTransient(throwable: Throwable): Boolean = !isFatal(throwable)
+  def isTransient(throwable: Throwable): Boolean = false
```
cjllanwarne: Can we get rid of this method (and `isFatal`) and only keep the `false` paths elsewhere?

engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
```diff
@@ -1,84 +1,98 @@
...
-    zippedImports = None,
-    warnings = Vector.empty
-  )
-
+  implicit val partialWorkflowSourcesMonoid = new Monoid[PartialWorkflowSources] {
+    override def empty: PartialWorkflowSources = PartialWorkflowSources()
+    
+    override def combine(x: PartialWorkflowSources, y: PartialWorkflowSources): PartialWorkflowSources = {
+      x.copy(
+        workflowSource = x.workflowSource.orElse(y.workflowSource),
```
cjllanwarne: ToL: I wonder whether combining two `Some`s should be an error, here and in some of the other fields below

project/Dependencies.scala
```diff
@@ -262,7 +262,7 @@ object Dependencies {
     "com.fasterxml.jackson.module" %% "jackson-module-scala" % jacksonV,
     "io.swagger" % "swagger-parser" % swaggerParserV % Test,
     "org.yaml" % "snakeyaml" % snakeyamlV % Test
-  ) ++ swaggerUiDependencies ++ akkaHttpDependencies
+  ) ++ swaggerUiDependencies ++ akkaHttpDependencies ++ List(circeYamlDependency)
```
cjllanwarne: `++ List(x)` <==> `:+ x`?

CHANGELOG.md
geoffjentry: What was it before?

engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
geoffjentry: why?

backend/src/main/scala/cromwell/backend/async/AsyncBackendJobExecutionActor.scala
Horneth: not sure what you mean ? `isFatal` and `isTransient` are needed by the `withRetry` method

CHANGELOG.md
Horneth: Well our generic "transform an exception in a json response" was already doing that if the exception was an `AggregatedException`, but the logic that looks at the form parameters in the request was not accumulated errors and throwing at the first one so it the `errors` array was never there.

geoffjentry: Got it. This is demonstrably better. As long as we make it clear that FC will need to update it's fine.

engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
```diff
@@ -222,14 +222,19 @@ class CromwellApiServiceSpec extends AsyncFlatSpec with ScalatestRouteTest with
   }
 
     it should "return 400 for an unrecognized form data request parameter " in {
-      val formData = Multipart.FormData(Multipart.FormData.BodyPart("incorrectParameter", HttpEntity(MediaTypes.`application/json`, HelloWorld.workflowSource()))).toEntity()
+      val formData = Multipart.FormData(Map(
+        "incorrectParameter" -> HttpEntity(MediaTypes.`application/json`, HelloWorld.workflowSource()),
+        "incorrectParameter2" -> HttpEntity(MediaTypes.`application/json`, HelloWorld.workflowSource())
+      )).toEntity()
```
kshakir: ToL: I think of this like `.toString`, in that we don't add the parens in scala

backend/src/main/scala/cromwell/backend/async/AsyncBackendJobExecutionActor.scala
cjllanwarne: Ignore me. Since it's now hard-coded to false, I thought we could remove the method call... but I guess it must be overridden so we *do* have to keep it!

engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
cjllanwarne: because having two workflow sources is an error

kshakir: > ToL: I wonder whether combining two Somes should be an error, here and in some of the other fields below

I'm guessing you're saying one might want `PartialWorkflowSources(workflowSource = "wf 1")` combined with `PartialWorkflowSources(workflowSource = "wf 2")` to be an error. Instead as is the result would be `PartialWorkflowSources(workflowSource = "wf 1")` and "wf 2" is ignored.

Is there an existing Monoid[Option]-ish function of ExclusiveOr-or-Error?

CHANGELOG.md
Horneth: To clarify further:

Up until now, the logic processing the submit request is wrapped in a `Try`, and the first thing that throws becomes the single exception we record. We then fail the request using this exception which results in something like 
```
{
  "status": "fail",
  "message": "Error(s): blablabla"
}
```

Note that because errors are not accumulated, there can't possibly be more than one in the message.

Now the same logic that transforms this exception into the above json (which we use for all endpoints), is smart enough to detect if the exception is an `AggregatedException` and will then create an `errors` array which will contain all the "sub" exceptions.

This change makes use of that, but @cjllanwarne is right that in doing so the "message" field doesn't contain the error anymore, but rather a context message that pretty much says "your submit request failed, see the errors array to know why".

A couple options would be 

1) if there's actually only one sub exception, put it in the "message" field and only use the "errors" array when there is more.
2) put a big string with all the exceptions concatenated in message and forget about the "errors" array (although it can definitely already appear in other endpoints that throws aggregated exceptions)

Once again, since it is true that FC might have to do some changes to adapt to this, I'm totally fine with saying "let's keep this for later". I saw it as a "nice to have" while doing this but it's totally unrelated to the yaml stuff.

engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
Horneth: I agree, but I don't think this is technically possible since the form data is a `Map[String, ByteString]`, so there can't possibly be 2 "workflowSource"s, or anything else for that matter.

CHANGELOG.md
geoffjentry: Ok, got it now. FWIW the `cromwell.yaml` (ie the now source of truth for our API, right? right?!?) doesn't have it as an array so this would indeed be a change.

If option #2 is indeed viable let's either go forward as-is now or fall back to that. I'll double check on the FC side to see what they want to do.

geoffjentry:
👍 modulo the ultimate resolution of the error thing

engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
mcovarr: if combining two `Some`s is an error I have to wonder if this is more a big `orElse` than a monoid

Horneth: You're right it's probably not a mathematically correct monoid but the result is still correct and can be `foldMap`ed with `ErrorOr` which makes things more concise IMO.

Horneth:
@kshakir I updated the swagger description of the input files in the submit endpoint to say "JSON or YAML". It doesn't get automatically updated in the docs but @geoffjentry said we'll figure that out when @katevoss is back.

engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
cjllanwarne: It's because "it can't happen" that I was advocating for an exception (which should never be thrown)

cjllanwarne: (because if we're wrong, or our assumptions about how it gets called change, we definitely want it to explode, rather than silently hiding the error)

Horneth:
@cjllanwarne @mcovarr @kshakir I tried to address the concerns about the monoid stuff, if you want take a last look, it should do what people asked.

common/src/main/scala/common/validation/Validation.scala
```diff
@@ -41,7 +42,11 @@ object Validation {
   implicit class ValidationTry[A](val e: ErrorOr[A]) extends AnyVal {
     def toTry: Try[A] = e match {
       case Valid(options) => Success(options)
-      case Invalid(err) => Failure(new RuntimeException(s"Error(s): ${err.toList.mkString(",")}"))
+      case Invalid(err) => Failure(AggregatedMessageException("Error(s)", err.toList))
+    }
+    def toTry(context: String): Try[A] = e match {
+      case Valid(options) => Success(options)
+      case Invalid(err) => Failure(AggregatedMessageException(context, err.toList))
```
kshakir: The codecov browser plugin says these `cromwell-common` lines aren't covered, but I'll put it back to 100% in another commit. ;)

PR #2834, opened 2017-11-08
This wasn't an example of 2724

cjllanwarne:


cjllanwarne:
... but it's still broken. Closing this down

PR #2857, opened 2017-11-13
Allow WomGraphs to be made for workflows that require inputs

cjllanwarne:


womtool/src/main/scala/womtool/graph/WomGraph.scala
```diff
@@ -175,8 +164,9 @@ object WomGraph {
 
   private def womExecutableFromWdl(filePath: String): Executable = {
     val namespace = WdlNamespaceWithWorkflow.load(readFile(filePath), Seq(WdlNamespace.fileResolver _)).get
-    // TODO WOM don't think anything good is going to happen with a None inputs file.
-    namespace.womExecutable(None) match {
+    val sampleInputs = JsObject(namespace.workflow.inputs map { i => i._1 -> sampleInput(i._2.womType) })
```
geoffjentry: what is going on w/ this sampleInputs thing?

cjllanwarne: @geoffjentry `namespace.womExecutable` now fails if the inputs provided don't satisfy the required graph inputs.

So, this generates fake (but valid) inputs to the workflow so that we can succeed in making the WOM graph and thus draw it.

mcovarr: `fake` or `dummy` might make that more clear. And/or a comment. 🙂

PR #2858, opened 2017-11-13
Reinstate a whole bunch of centaur tests. Let the shoes drop...!

cjllanwarne:


geoffjentry:
@cjllanwarne Outside of someone saying "I like my shoes not fallen" is there any reason this needs a review?

backend/src/main/scala/cromwell/backend/io/GlobFunctions.scala
```diff
@@ -8,16 +8,15 @@ import common.validation.ErrorOr.ErrorOr
...
 
 trait GlobFunctions extends IoFunctionSet {
 
   def callContext: CallContext
 
   def findGlobOutputs(call: TaskCallNode, jobDescriptor: BackendJobDescriptor): ErrorOr[List[WomGlobFile]] =
-    call.callable.outputs.flatTraverse[ErrorOr, WomGlobFile] {
-      _.expression.evaluateFiles(jobDescriptor.localInputs, this, WomAnyType) map {
+    call.callable.outputs.flatTraverse[ErrorOr, WomGlobFile] { outputDefinition =>
+      outputDefinition.expression.evaluateFiles(jobDescriptor.localInputs, this, outputDefinition.womType) map {
```
mcovarr: what's up with the type change?

centaur/src/main/resources/standardTestCases/array_literal_locations.test
```diff
@@ -14,6 +10,6 @@ metadata {
   workflowName: array_literal_locations
   status: Succeeded
   "outputs.array_literal_locations.array_literal_locations_i.out": "0:1:2"
-  "outputs.array_literal_locations.array_literal_locations_ii.0.out": "0"
-  "outputs.array_literal_locations.array_literal_locations_ii.1.out": "1"
+  "outputs.array_literal_locations.array_literal_locations_ii.out.0": "0"
+  "outputs.array_literal_locations.array_literal_locations_ii.out.1": "1"
```
mcovarr: 😄

backend/src/main/scala/cromwell/backend/io/GlobFunctions.scala
cjllanwarne: The FileEvaluator needs to know what it's coercing inputs into. Previously it was failing to read pair inputs from JSON (even though they didn't have any globs, they still need to go through the "does it have globs" method)

PR #2859, opened 2017-11-14
Added REST API docs generator.

kshakir:
Minor updates to YAML to fix capitalization.
Cleaned up sbt eviction warnings for both project and project/project.


katevoss:
Blargh apparently pull approve doesn't work if I edit my comment? 👍

geoffjentry:
Yeah pullapprove is wise but not all seeing

PR #2909, opened 2017-11-15
Fixup CWL graph outputs

cjllanwarne:


womtool/src/main/scala/womtool/graph/WomGraph.scala
```diff
@@ -178,7 +178,9 @@ object WomGraph {
       clt <- CwlDecoder.decodeAllCwl(File(filePath)).
         value.
         unsafeRunSync
-      wom <- clt.womExecutable()
+      inputs = clt.requiredInputs
+      fakedInputs = JsObject(inputs map { i => i._1 -> fakeInput(i._2) })
+      wom <- clt.womExecutable(Some(fakedInputs.prettyPrint))
```
mcovarr: why the fresh fakery in this PR?

cwl/src/main/scala/cwl/package.scala
```diff
@@ -86,5 +86,20 @@ package object cwl extends TypeAliases {
...
     }
+
+    def requiredInputs: Map[String, WomType] = cwl match {
+      case Cwl.Workflow(w) => selectWomTypeInputs(w.inputs collect {
+        case i if i.`type`.isDefined => FullyQualifiedName(i.id).id -> i.`type`.get
+      })
+      case Cwl.CommandLineTool(clt) => selectWomTypeInputs(clt.inputs collect {
+        case i if i.`type`.isDefined => FullyQualifiedName(i.id).id -> i.`type`.get
+      })
+    }
```
mcovarr: is the code duplication suggesting a common supertype? 🙂

womtool/src/main/scala/womtool/graph/WomGraph.scala
mcovarr: `Option`

cwl/src/main/scala/cwl/TypeAliases.scala
```diff
@@ -58,6 +59,21 @@ trait TypeAliases {
...
+  object MyriadInputType {
+    object CwlType {
+      def unapply(m: MyriadInputType): Option[CwlType] = {
+        m.select[CwlType]
+      }
+    }
+
+    object WomType {
+      def unapply(m: MyriadInputType): Option[WomType] = m match {
+        case CwlType(c) => Some(cwl.cwlTypeToWdlType(c))
```
mcovarr: `Option`

womtool/src/main/scala/womtool/graph/WomGraph.scala
geoffjentry: The faked stuff still seems like a smell to me but I don't have a better suggestion so ....

geoffjentry:
I agree w/ the points @mcovarr made, otherwise 👍

cwl/src/main/scala/cwl/package.scala
cjllanwarne: Perhaps, I'll let @danbills look into that. As far as I can tell, `InputParameter` and `CommandInputParameter` are the same thing

danbills: they are indeed!  as well as all the `*Schema` classes too it seems.  I think we could de-duplicate broadly if we feel confident that CWL would not diverge in the future.  It would be nice to know why they have distinct names in the spec.

danbills:
I think a rebase will make Travis CI and then we can merge

cwl/src/main/scala/cwl/package.scala
geoffjentry: IIRC originally there was just a single entity and then when it became clear that workflows and individual tools have slightly different needs they diverged. The other thing was that at the time, at least from my observation, that community was dominated by people who think/thought in terms of single tools and not workflows, despite the W in the acronym.

danbills:
Your new pattern-matching `MyriadInputType` object should be moved out of the trait into it's own space.

So instead of 
```scala
trait TypeAliases {
  object MyriadInputType { 
  
```
do 
```scala
trait TypeAliases { ...
}
object MyriadInputType { ...
}  
```

PR #2911, opened 2017-11-16
Reinstate refresh_token tests.

kshakir:


cloudSupport/src/test/scala/cromwell/cloudsupport/gcp/auth/GoogleAuthModeSpec.scala
```diff
@@ -0,0 +1,126 @@
...
+    val privateKey = keyPair.getPrivate
+    val byteEncoded = privateKey.getEncoded
+    val base64Encoded = Base64.getEncoder.encodeToString(byteEncoded)
+    s"""|-----BEGIN PRIVATE KEY-----
+        |$base64Encoded
+        |-----END PRIVATE KEY-----
+        |""".stripMargin
+  }
+
+  // Hide me from git secrets false positives
```
mcovarr: :rofl:

common/src/main/scala/common/validation/Validation.scala
```diff
@@ -40,10 +40,8 @@ object Validation {
   }
 
   implicit class ValidationTry[A](val e: ErrorOr[A]) extends AnyVal {
-    def toTry: Try[A] = e match {
-      case Valid(options) => Success(options)
-      case Invalid(err) => Failure(AggregatedMessageException("Error(s)", err.toList))
-    }
+    def toTry: Try[A] = toTry("Error(s)")
```
mcovarr: so happy there isn't a separate PR in a separate repo for this 😂

common/src/test/scala/common/validation/ValidationSpec.scala
```diff
@@ -78,4 +79,14 @@ class ValidationSpec extends FlatSpec with Matchers {
     failure.toChecked shouldBe Left(NonEmptyList.of(":("))
   }
 
+  it should "convert an ErrorOr to a Try" in {
+    val valid = "yeah".valid
+    val invalid = ":(".invalidNel
```
mcovarr: a unicode frowny might be more clear 😄

cloudSupport/src/main/scala/cromwell/cloudsupport/gcp/auth/GoogleAuthMode.scala
```diff
@@ -81,14 +71,26 @@ sealed trait GoogleAuthMode {
 
   def requiresAuthFile: Boolean = false
 
+  // changing this is for testing only. all the inheritance is otherwise locked down.
```
geoffjentry: I think I understand what this sentence is saying but I'm not sure, which makes me think it should be rephrased somehow

cloudSupport/src/main/scala/cromwell/cloudsupport/gcp/auth/GoogleAuthMode.scala
```diff
@@ -81,14 +71,26 @@ sealed trait GoogleAuthMode {
...
   protected def validateCredential(credential: Credentials) = {
-    Try(credential.refresh()) match {
+    Try(credentialValidation(credential)) match {
       case Failure(ex) => throw new RuntimeException(s"Google credentials are invalid: ${ex.getMessage}", ex)
       case Success(_) => credential
     }
   }
 }
 
+case object MockAuthMode extends GoogleAuthMode {
```
geoffjentry: Is this used only in test code? if so would it make more sense to live over in test-land?

kshakir: That would be nice, but I don't think that works with `sealed trait`s. In this PR I just moved the class from an inner-class to a top-level-class.

gemlam3:
@geoffjentry can you re-review this and approve if it looks good?

PR #2913, opened 2017-11-17
Hello sub workflow world

cjllanwarne:


engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/job/preparation/SubWorkflowPreparationActor.scala
```diff
@@ -20,24 +24,48 @@ class SubWorkflowPreparationActor(workflowDescriptor: EngineWorkflowDescriptor,
...
+      case (Some(k), v) => k -> v
+    }
+
+    // We don't need to supply these:
+    val optionalsAndDefaults: Set[GraphNode] = workflowInputs collect {
+      case optionalInputNode: OptionalGraphInputNode => optionalInputNode
+      case optionalWithDefault: OptionalGraphInputNodeWithDefault => optionalWithDefault
+    }
+
+    // Make sure we got everything:
```
cjllanwarne: Question to reviewers: should we also make sure we didn't get anything we didn't want?

wdl/src/main/scala/wdl/WdlNamespace.scala
```diff
@@ -102,40 +101,6 @@ case class WdlNamespaceWithWorkflow(importedAs: Option[String],
 
   override def toString: String = s"[WdlNamespace importedAs=$importedAs]"
 
-  /**
-    * Confirm all required inputs are present and attempt to coerce raw inputs to `WomValue`s.
-    * This can fail if required raw inputs are missing or if the values for a specified raw input
-    * cannot be coerced to the target type of the input as specified in the namespace.
-    */
-  def coerceRawInputs(rawInputs: ExecutableInputMap): Try[WorkflowCoercedInputs] = {
```
cjllanwarne: Note to reviewers: In case you're curious why this is going, it was unused apart from in tests.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/job/preparation/SubWorkflowPreparationActor.scala
kshakir: Good question. In general, I'm satisfied if a unit test makes sure we didn't get anything we didn't want, instead of production code testing/asserting during real workflow runs.

However, at the moment, the codecov chrome plugin says this file has _zero_ lines covered.

cjllanwarne: This is more asking “should we (the engine) check that the ‘{ input: x=... }‘ block didn’t give us inputs that we (the sub workflow) weren’t expecting (albeit, perhaps we should have caught that before subworkflow preparation...)

kshakir: I'm still digesting all of this code, so apologies if I'm missing something obvious.

What are reasons why the engine would/should check for extra inputs? If the check would just be for ensuring that the code is correct, then I'd say put sanity checks in test code, not in main code. If a bug happens, I'd put the check into a regression test.

The one case I have seen main-code-asserts is something like (pseudo-code):

```
if (logger.isDebugModeEnabled) {
   val extraCheckResults = doExtraCheckHere()
   extraCheckResults foreach logger.debug
}
```

But those blocks I've usually reserved for cases where some sort of transient/heisen bug was happening in production environments. When discovered I usually replaced the blocks with a proper regression test.

mcovarr: I *think* the question is if a *user* has written WDL where the main workflow calls a subworkflow with inputs that aren't valid for that subworkflow, should Cromwell call that out (where "call that out" is left ambiguous as to warning or error)? Right now I believe this condition is silently ignored.

cjllanwarne: Right, @mcovarr!

Hopefully by this point something else would have checked statically so this could be a sanity check, but OTOH I don’t think that “something else” exists yet...

wdl/src/main/scala/wdl/Declaration.scala
```diff
@@ -66,9 +67,17 @@ trait DeclarationInterface extends WdlGraphNodeWithUpstreamReferences {
     case None => Option(TaskInput(unqualifiedName, womType))
   }
 
-  def asWorkflowInput: Option[WorkflowInput] = expression match {
-    case Some(_) => None
-    case None => Some(WorkflowInput(fullyQualifiedName, womType))
+  def asWorkflowInput: Option[InputDefinition] = (expression, womType) match {
```
mcovarr: docs!

wdl/src/main/scala/wdl/Declaration.scala
```diff
@@ -127,11 +136,17 @@ object Declaration {
...
 
-    decl match {
-      case Declaration(opt: WomOptionalType, _, None, _, _) => Valid(InputDeclarationNode(OptionalGraphInputNode(decl.womIdentifier, opt)))
-      case Declaration(_, _, None, _, _) => Valid(InputDeclarationNode(RequiredGraphInputNode(decl.womIdentifier, decl.womType)))
-      case Declaration(_, _, Some(expr), _, _) => declarationAsExpressionNode(expr)
-      case WorkflowOutput(_, _, expr, _, _) => workflowOutputAsGraphOutputNode(expr)
+    def asWorkflowInput(inputDefinition: InputDefinition): GraphInputNode = inputDefinition match {
+      case RequiredInputDefinition(_, womType) => RequiredGraphInputNode(decl.womIdentifier, womType)
+      case OptionalInputDefinition(_, optionalType) => OptionalGraphInputNode(decl.womIdentifier, optionalType)
+      case InputDefinitionWithDefault(_, womType, default) => OptionalGraphInputNodeWithDefault(decl.womIdentifier, womType, default)
```
mcovarr: I'm not sure exactly what an `InputDefinitionWithDefault` is. Is this only optionals with a default, which are overrideable? Or does that also include non-optionals and those are overrideable too?

wdl/src/main/scala/wdl/Declaration.scala
```diff
@@ -66,9 +67,17 @@ trait DeclarationInterface extends WdlGraphNodeWithUpstreamReferences {
...
 
-  def asWorkflowInput: Option[WorkflowInput] = expression match {
-    case Some(_) => None
-    case None => Some(WorkflowInput(fullyQualifiedName, womType))
+  def asWorkflowInput: Option[InputDefinition] = (expression, womType) match {
+    case (None, optionalType: WomOptionalType) =>
+      Some(OptionalInputDefinition(fullyQualifiedName, optionalType.flatOptionalType))
+    case (None, nonOptionalType) =>
+      Some(RequiredInputDefinition(fullyQualifiedName, nonOptionalType))
+    case (Some(expr), optionalType: WomOptionalType) if upstreamAncestry.isEmpty =>
```
mcovarr: no clue what the upstream ancestry is about

wdl/src/main/scala/wdl/WdlNamespace.scala
mcovarr: and who needs tests?

wdl/src/test/scala/wom/WdlSubworkflowWomSpec.scala
```diff
@@ -214,7 +214,13 @@ class WdlSubworkflowWomSpec extends FlatSpec with Matchers {
 
     def validateOuter(workflowGraph: Graph) = {
       // One input, x
-      workflowGraph.inputNodes shouldBe empty
+      workflowGraph.inputNodes.size should be(1)
+      workflowGraph.inputNodes foreach {
+        case opt: OptionalGraphInputNodeWithDefault =>
+          opt.localName should be("i")
+          opt.womType should be(WomIntegerType)
+        case other => fail("Unexoected input node to outer graph: " + other.localName)
```
mcovarr: expected the unexpected spellings i guess

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/job/preparation/SubWorkflowPreparationActor.scala
kshakir: That's a WDL spec decision for the committee to be debated elsewhere outside of this particular engine.

That said, and invoking "ToL" ;)

ToL: I'm fine with (and assumed that) an engine would ignore extra inputs to a sub workflow. Only missing inputs are an error. This is based on the behavior of the inputs.json, that I believe _does not_ cause an error if there are extra fields not found in the workflow. I'm not sure if that's in the "WDL Spec" either? Honestly I'm fine with whatever, but I'd like consistency.

geoffjentry: The spec currently isn't too particular on this topic: https://github.com/openwdl/wdl/blob/master/versions/draft-3/SPEC.md#specifying-workflow-inputs-in-json

This means that until someone wants to try to make it part of the spec, it isn't.

wdl/src/main/scala/wdl/WdlNamespace.scala
cjllanwarne: To be clear: this function was only ever used by tests that only tested this function :)

wdl/src/main/scala/wdl/Declaration.scala
cjllanwarne: Maybe a rename would help? It's a workflow input that has a default. Could be`Int x = 5`, could be `String? s = "hello"`

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -167,8 +160,17 @@ case class WorkflowExecutionActor(params: WorkflowExecutionActorParams)
...
+    case Event(SubWorkflowSucceededResponse(jobKey, descendantJobKeys, callOutputs), currentStateData) =>
+      // Update call outputs to come from sub-workflow output ports:
+      val subworkflowOutputs: Map[OutputPort, WomValue] = callOutputs.outputs map { case (port, value) =>
+        jobKey.node.subworkflowCallOutputPorts.find(_.outputToExpose.graphOutputPort eq port) -> value
+      } collect {
+        // Only keep the outputs that the subworkflow graph Node intended to expose!
+        case (Some(port), value: WomValue) => port -> value
+      }
+      // TODO: Fail if the subworkflow didn't give up an output we need
+      assert(subworkflowOutputs.size == callOutputs.outputs.size)
```
mcovarr: why is this an `assert`?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -167,8 +160,17 @@ case class WorkflowExecutionActor(params: WorkflowExecutionActorParams)
...
-    case Event(SubWorkflowSucceededResponse(jobKey, descendantJobKeys, callOutputs), stateData) =>
-      handleCallSuccessful(jobKey, callOutputs, None, stateData, descendantJobKeys)
+    case Event(SubWorkflowSucceededResponse(jobKey, descendantJobKeys, callOutputs), currentStateData) =>
+      // Update call outputs to come from sub-workflow output ports:
+      val subworkflowOutputs: Map[OutputPort, WomValue] = callOutputs.outputs map { case (port, value) =>
+        jobKey.node.subworkflowCallOutputPorts.find(_.outputToExpose.graphOutputPort eq port) -> value
+      } collect {
+        // Only keep the outputs that the subworkflow graph Node intended to expose!
+        case (Some(port), value: WomValue) => port -> value
+      }
```
mcovarr: This looks like the other code that was turned into a single `flatMap`

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/job/preparation/SubWorkflowPreparationActor.scala
```diff
@@ -11,33 +12,65 @@ import cromwell.engine.workflow.lifecycle.execution.job.preparation.CallPreparat
...
+  }
+
+  /**
+    * Work out a set of  "workflow inputs" to pass to this subworkflow as though it were a top-level workflow receiving inputs
+    */
+  private def evaluateStartingKnownValues(inputEvaluation: WomEvaluatedCallInputs, workflowInputs: Set[GraphInputNode]): ErrorOr[Map[OutputPort, WomValue]] = {
+
+    /**
+      * Find the values in the provided inputs that match up with subworkflow inputs. Silently drop the rest on the floor.
+      */
```
mcovarr: should not be scaladoc, doc generation tools will error with "unmoored comment" messages

wdl/src/main/scala/wdl/Declaration.scala
mcovarr: I don't have a suggestion for a better name, as long as the doc on this thing says what you just said that should be fine. 🙂

mcovarr:
:+1:

wdl/src/main/scala/wdl/Declaration.scala
cjllanwarne: Added scaladocs to the `case class`

Issue #2835, opened 2017-11-08
Run CWL tests in Centaur


danbills:
We currently do not have automated CWL testing in Centaur.  this puts us at risk for regressions.

Start w/ 
* 1st-workflow.cwl
* three_step.cwl

PR #2920, opened 2017-11-20
Enabled three_step_cwl centaur test.

kshakir:


centaur/src/main/resources/standardTestCases/three_step.test
```diff
@@ -1,14 +1,20 @@
...
 workflowType: CWL
 workflowTypeVersion: v1.0
 backends: [Local]
-backendsMode: "only"
 tags: [localdockertest]
 
 files {
   wdl: three_step/three_step.cwl
   inputs: three_step/inputs.json
+  options: three_step/three_step.options
```
cjllanwarne: Does this really fix the problem or just force the JES centaur to run the test on its (should be removed) “Local” backend?

centaur/src/main/resources/standardTestCases/three_step.test
```diff
@@ -1,14 +1,20 @@
...
 }
 
-metadata { }
+metadata {
+  "submittedFiles.workflowType": CWL
+  "submittedFiles.workflowTypeVersion": v1.0
+  "calls.wc.executionStatus": Done
+  "calls.ps.executionStatus": Done
+  "calls.cgrep.executionStatus": Done
```
cjllanwarne: Can we test outputs too? To make sure CWL support isn’t regressing

centaur/src/main/resources/standardTestCases/three_step.test
kshakir: Specifically, this forces TES-centaur to run this test on the (should be removed, maybe, see below) "Local" backend.

JES-centaur passes [`-elocaldockertest`](https://github.com/broadinstitute/cromwell/blob/c6ed64617c51c572863b87d324fa8e68fa085b1a/src/bin/travis/resources/centaur.wdl#L34), meaning that _all_ tests tagged as such are skipped. Meanwhile, TES uses a very specific [blacklist](https://github.com/broadinstitute/cromwell/blob/c6ed64617c51c572863b87d324fa8e68fa085b1a/src/bin/travis/testCentaurTes.sh#L71-L78) of incompatible tests, allowing those that can run to specify via a workflow-options file that the test should run on the (should be removed..?) Local.

centaur/src/main/resources/standardTestCases/three_step.test
kshakir: Added a new issue https://github.com/broadinstitute/cromwell/issues/2924 that also describes the current issue with 3step.

kshakir:
FYI was continuation of #2835

PR #2921, opened 2017-11-21
Fix writeFile in JES and TES.

mcovarr:


centaur/src/main/resources/standardTestCases/fofn_caching.test
```diff
@@ -1,4 +1,4 @@
-ignore: true # Depends on 2864
+ignore: true # Depends on 2864 and also some TODO WOM fun in cromwell.engine.WdlFunctions
```
mcovarr: This test is still broken due to that TODO WOM.

mcovarr:
Actually on further reflection I'm going to close this and split it in two. One of the new PRs will just be noisy mechanical stuff, the other will be mostly substantive and should have better docs than what's here.

PR #2923, opened 2017-11-21
rename cleanup

mcovarr:


backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -108,38 +105,38 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
...
   }
 
   /**
-    * Maps WdlFile to a local path, for use in the commandLineValueMapper.
+    * Maps WomFile to a local path, for use in the commandLineValueMapper.
     *
-    * @param wdlFile The wdlFile.
-    * @return The updated wdlFile.
+    * @param womFile The womFile.
+    * @return The updated womFile.
```
cjllanwarne: ToL I always hate these "that's the name of the parameter!" parrot-docs

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
```diff
@@ -133,9 +133,9 @@ class JesAsyncBackendJobExecutionActor(override val standardParams: StandardAsyn
   }
 
   /**
-    * Takes two arrays of remote and local WDL File paths and generates the necessary JesInputs.
+    * Takes two arrays of remote and local WOM File paths and generates the necessary JesInputs.
     */
-  private def jesInputsFromWdlFiles(jesNamePrefix: String,
+  private def jesInputsFromWomFiles(jesNamePrefix: String,
```
cjllanwarne: Might as well rename fully to `papiInputsFromWomFiles`?

mcovarr: At some point we should PAPI All The Things but I think that's a separate PR. This is just the stuff I bumped into fixing `writeFile`.

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
geoffjentry: Agreed. Unless the scaladoc is telling me something non-obvious from the name/type I see no point in having it

mcovarr: I'll whack this in the sequel PR.

PR #2925, opened 2017-11-21
Reinstated no_new_calls centaur test.

kshakir:


centaur/src/main/resources/standardTestCases/no_new_calls.test
```diff
@@ -16,7 +10,23 @@ metadata {
...
+absent-metadata-keys: [
+  "calls.no_new_calls.shouldNotStart.executionStatus",
+
+  # Currently, if cromwell restarts, it may attempt execute "Restarting calls" on boundToFail, shouldSucceed, AND then
+  # when shouldSucceed finally finishes, it "Restarting calls" on delayedTask.
+  #
+  # Cromwell then catches the mistaken restart by failing the call and setting the error message to:
+  # > Cromwell server was restarted while this workflow was running. As part of the restart process, Cromwell attempted
+  # > to reconnect to this job, however it was never started in the first place. This is a benign failure and not the
+  # > cause of failure for this workflow, it can be safely ignored.
```
cjllanwarne: It seems like the problem is:
- The engine can't know whether a job was started or not when it comes back from a restart.
- So it has to try to reconnect.
- The job execution actor immediately finds out that it's in a failing workflow and this job hadn't started.
- We don't have a "JobNeverStarted" response in the WEA, so the job execution actor has no choice but to respond with a failure
- And that failure gets recorded even though we know it's not really this job's problem.

Comments:
- Do we have a ticket to tidy this up...?
- IMO a clearer message in the absence of a bigger fixup might be "This job has not been started because the workflow has already failed elsewhere"

kshakir: Yep, ticket is linked above as a reference-- #2926

mcovarr: I'd like to get Thibault's input when he gets back, I thought this condition had been modeled clearly.

mcovarr:
:+1:

centaur/src/main/resources/standardTestCases/no_new_calls.test
Horneth: @cjllanwarne This sounds like what I would expect from a "what was implemented" point of view. It might not be what we want, although we did tech talk this and agreed that "restarting unstarted calls was ok as long as the error message indicated clearly what happened". 
Other than clarifying the message, the options I see are
1) add a "JobNeverStarted" status, which could end up being awkward because only jobs that we attempted to restart but had never started would have it, although technically all jobs that were never started could be in this state.
2) Actually record in the DB when we start a job so that we don't have to reconnect to unstarted jobs. This would need some re-factoring of the JobStore and we'd also have to be careful of race conditions when Cromwell restarts (seems to me like it would be hard to guarantee that jobs are marked "started" in the DB when and only when the backend effectively starts it given that Cromwell can restart in the middle of it, this is the reason why I didn't go this route in the first place as it already seemed complicated enough).

Let me know if there's anything I can do to help or clarify on this. I'll be at reInvent next week but I'll be online

Issue #2901, opened 2017-11-14
Wom: Reinstate task outputs dependant on previous outputs


cjllanwarne:
This seems to be the reason that this centaur test is failing even after #2900:
- [x] `globbingindex`
- [ ] `sub_workflow_var_refs` (in this case, the outputs depend on inputs)

cjllanwarne:
Fixed. Addressing the unrelated problem in `sub_workflow_var_refs` as part of #2937

PR #2933, opened 2017-11-22
For subworkflows, no outputs means no outputs.

cjllanwarne:
Fixes the second half of #2901 (which turned out to not be caused by task outputs at all, but incorrectly added subworkflow outputs)


cjllanwarne:
Side note: we probably do want to test outputs from this test case, but since it found a bug, an empty `output {}` block in a subworkflow is definitely worth having a test for too

cjllanwarne:
Needs more work.... but it will return!

PR #2935, opened 2017-11-26
Unify terminology for submitting labels to cromwell

geoffjentry:
Workflow submission has a form field `customLabels` whereas the label update endpoint has the form field `labels`. When one is trying to abstract logic involving labels to the akka http DSL this makes it yucky. There's no reason why these need to be different. Unify them to `labels` terminology

There are still internal references to customLabels, but not in a user visible way.

edit: Looks like the command line API also has `labels` as the terminology

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 * **API**  
     + Cromwell now supports input files in the yaml format (json format is still supported).
+    + The `customLabels` form field for workflow submission has been renamed to `labels`.
 
 * **Database**  
 You have the option of storing the metadata in a separate SQL database than the database containing the internal engine
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,14 @@
 
 ## 30 Release Notes
 
+### Breaking Changes
+
+* The `customLabels` form field for workflow submission has been renamed to `labels`.
+
 ### Other changes
 
 * **API**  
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 * **API**  
     + Cromwell now supports input files in the yaml format (json format is still supported).
+    + The `customLabels` form field for workflow submission has been renamed to `labels`.
```
cjllanwarne: Put under a "breaking changes" section?

cjllanwarne:
Esprit d'escalier: Does FC know this is coming? Is it going to slow down their adoption of Cromwell?

geoffjentry:
@cjllanwarne in this case it'll be a simple find/replace text change, the underlying API in terms of data structures remains the same.

kshakir:
> Does FC know this is coming?

And GOTC. These are the types of changes that drive them insane when they don't know about them ahead of time. ("wtf happened to _insert-renamed-field-here_???")

geoffjentry:
That's why it's in the changelog :P

geoffjentry:
@cjllanwarne re `customLabels` and `custom labels` - I did a search previously and all remaining mentions were either deeper in the code (which were semi-defensible in terms of naming) and in references to the concept in the docs which also seemed defensible in that they added more clarity. As an example which I just made up - "One may set custom labels by setting the form field `labels`"

PR #2938, opened 2017-11-27
Remove cromwell-core dependency from cloud-support

kshakir:
PR 1 of 3:
1. Remove cromwell-core dependency from cloud-support
2. Run jes centaur on travis
3. Generate coverage for integration tests

---

Instead of credentials requiring WorkflowOptions, any String => String will do, including Map[String, String].
Retrieving credentials only requires actorSystem/executionContext when retrying.
Moved logback dependencies from common library over to testing.
Added mockito to all artifact tests.
Fixed akka-stream-testkit dependency appearing in core's main instead of test.
Split confusingly named baseDependencies into configDependencies ++ catsDependencies.
Other dependency cleanup to reduce duplicates and extra transitive dependencies.
Log stderr from centaur'ed cromwell failures.
The total attempt time to connect to cromwell for a test is now longer than the timeout of a cromwell restart.


services/src/main/scala/cromwell/services/healthmonitor/impl/workbench/WorkbenchHealthMonitorServiceActor.scala
```diff
@@ -30,8 +29,8 @@ class WorkbenchHealthMonitorServiceActor(val serviceConfig: Config, globalConfig
 
   override lazy val subsystems: Set[MonitoredSubsystem] = Set(DockerHub, EngineDb, Papi, Gcs)
 
-  private lazy val Gcs = MonitoredSubsystem("GCS", checkGcs _)
```
geoffjentry: iirc this fuckery was due to one of the linters we have (eta expansion?) which makes me wonder what changed to make it not necessary?

services/src/main/scala/cromwell/services/healthmonitor/impl/workbench/WorkbenchHealthMonitorServiceActor.scala
```diff
@@ -51,7 +50,7 @@ class WorkbenchHealthMonitorServiceActor(val serviceConfig: Config, globalConfig
   private def checkGcs(): Future[SubsystemStatus] = {
     // For any expected production usage of this check, the GCS bucket should be public read */
     val gcsBucketToCheck = serviceConfig.as[String]("gcs-bucket-to-check")
-    val storage = googleAuth.credential(WorkflowOptions.empty) map { c => GcsStorage.gcsStorage(googleConfig.applicationName, c) }
+    val storage = Future(googleAuth.credential(Map.empty)) map { c => GcsStorage.gcsStorage(googleConfig.applicationName, c) }
```
geoffjentry: Will these wind up retrying? I don't think so but maybe it's lurking somewhere underneath. Since these are firing off fairly regularly anyways I don't think retrying here would be right.

services/src/main/scala/cromwell/services/healthmonitor/impl/workbench/WorkbenchHealthMonitorServiceActor.scala
kshakir: - The original `MonitoredSubsystem("GCS", checkGcs _)` works in the linter, but gives a yellow box in intellij instead of the green checkmark.
- The intellij "fix", `MonitoredSubsystem("GCS", checkGcs)`, trips up the linter with the "eta expansion" message you referred to.
- The replacement `MonitoredSubsystem("GCS", () => checkGcs())` gives a green intellij checkmark and passes the scalac linter!

services/src/main/scala/cromwell/services/healthmonitor/impl/workbench/WorkbenchHealthMonitorServiceActor.scala
kshakir: I thought also that retrying would _not_ be right in this particular case for the same reasons.

These no longer retry. They were previously, via an implicit actor system being sent in to the old methods. That logic has been separated into a different wrapper.

services/src/main/scala/cromwell/services/healthmonitor/impl/workbench/WorkbenchHealthMonitorServiceActor.scala
geoffjentry: Sweet. As an aside, the new signature (which I swear I tried, but perhaps something else was different when I did) is exactly why I wanted to make them `IO` and not `Future` but that opened a can of worms I didn't have time to dig into

PR #2940, opened 2017-11-27
Reinstante invalid_runtime_attributes test.

kshakir:
A val instead of lazy-val was throwing an exception "too early".
The additional ActorInitializationException then affected the depth of the expected causeBy messages.


geoffjentry:
👍

geoffjentry:
Oops, hit the wrong button :)

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
```diff
@@ -113,7 +113,7 @@ class JesAsyncBackendJobExecutionActor(override val standardParams: StandardAsyn
 
   override lazy val dockerImageUsed: Option[String] = Option(jobDockerImage)
 
-  override val preemptible: Boolean = previousRetryReasons match {
+  override lazy val preemptible: Boolean = previousRetryReasons match {
```
cjllanwarne: 🤦‍♂️

Issue #2864, opened 2017-11-14
Wom: Reinstate write_lines in JES


cjllanwarne:
Reinstate centaur tests:
- [x] `write_lines`
- [ ] `fofn_caching`
- [x] `array_io`
- [x] `write_lines_files`

mcovarr:
[Oh my](https://github.com/broadinstitute/cromwell/blob/05c8dfdc914c40ef6b5443cad0361073e6b837f0/supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala#L162)

cjllanwarne:
Closing because `fofn_caching` is being fixed in #2922

PR #2942, opened 2017-11-28
Fix PAPI and TES writeFile.

mcovarr:
Gull away. There are things here that need discussion, in particular `WdlTask#instantiateCommand` that is called only by tests and "example" code. Also I'm not sure how to properly honor initialized optionals for the config backend; I'm currently forcing in `none` defaults since if I don't expression evaluation with uninitialized optionals blows up.


wdl/src/main/scala/wdl/command/ParameterCommandPart.scala
```diff
@@ -49,29 +45,53 @@ case class ParameterCommandPart(attributes: Map[String, String], expression: Wdl
...
+          // running Cromwell, or a cloud path. Capture these newly created files and their engine paths here.
+          val enginePathFile = value.asInstanceOf[WomFile]
+          InstantiatedCommand(f.valueString, List(enginePathFile)).validNel
+        case f: WomFile =>
+          InstantiatedCommand(f.valueString).validNel
+        case p: WomPrimitive => InstantiatedCommand(p.valueString).validNel
+        case a: WomArray if attributes.contains("sep") =>
+          InstantiatedCommand(a.value.map(_.valueString).mkString(attributes.get("sep").head)).validNel
+        case _: WomArray =>
+          s"Expression '${expression.toString}' evaluated to an Array but no 'sep' was specified".invalidNel
```
kshakir: I've seen `${expression.toWomString}` used elsewhere. If that's not the case here, then this should just be `$expression`.

wdl/src/main/scala/wdl/command/ParameterCommandPart.scala
```diff
@@ -49,29 +45,53 @@ case class ParameterCommandPart(attributes: Map[String, String], expression: Wdl
...
   }
 
   private def defaultString = {
-    if (attributes.contains("default")) {
-      WomString(attributes.get("default").head)
-    } else {
-      WomString("")
-    }
+    val value = if (attributes.contains("default")) attributes.get("default").head else ""
+    WomString(value)
```
kshakir: `WomString(attributes.getOrElse("default", ""))`

cwl/src/main/scala/cwl/CwlExpressionCommandPart.scala
```diff
@@ -1,76 +1,79 @@
...
+      val womValue: WomValue = commandInputParameter match {
+        case cip: CommandInputParameter =>
+          val localizedId = LocalName(FullyQualifiedName(cip.id).id)
+          inputsMap.get(localizedId) match {
+            case Some(x) =>x
+            case _ => throw new RuntimeException(s"could not find $localizedId in map $inputsMap")
+          }
+
+        // There's a fair few other cases to add, but until then...
+        case other => throw new NotImplementedError(s"As-yet-unsupported commandPart from  command input parameters: $other")
```
kshakir: While you're touching this line, please fix the two spaces between "from  command".

supportedBackends/sfs/src/main/scala/cromwell/backend/impl/sfs/config/ConfigAsyncJobExecutionActor.scala
```diff
@@ -47,8 +51,32 @@ sealed trait ConfigAsyncJobExecutionActor extends SharedFileSystemAsyncJobExecut
...
+    } yield womInputDefinition -> womValue).toMap
+
+    // TODO WOM FIXME
+    // This may force `none`s onto potentially initialized optionals. This is currently required because otherwise
+    // expressions using uninitialized optionals explode on evaluation.
+    val optionalsForciblyDefaultedToNone = for {
+      optional <- taskDefinition.inputs.collect { case oid: OptionalInputDefinition => oid }
+      // Only default to `none` if there was no input value explicitly provided.
+      if !inputs.contains(optional.localName.value)
+    } yield optional -> WomOptionalValue.none(optional.womType.memberType)
```
kshakir: Would love to chat IRL about what's missing here. These config-specified-tasks are usually relatively simple, but  true they do often [contain optionals](https://github.com/broadinstitute/cromwell/blob/5e325e98785670d2bf12bd98590a082c89e25a94/cromwell.examples.conf#L250-L257).

mcovarr: Discussed IRL. I'm going to leave the code effectively as it is now with a ticket and more comments.

wdl/src/main/scala/wdl/command/ParameterCommandPart.scala
mcovarr: This was just massaging the code on the old line 65 to fit the new return type. I'll change this to `${expression.toWomString}` as I suspect that will produce something far more readable. 😄

wdl/src/main/scala/wdl/command/ParameterCommandPart.scala
```diff
@@ -49,29 +45,50 @@ case class ParameterCommandPart(attributes: Map[String, String], expression: Wdl
...
+    // Create the stringified version of the command and record any file created by `writeFile`.
+    def instantiateCommand(value: WomValue): ErrorOr[InstantiatedCommand] = {
+      valueMapper(value) match {
+        case b: WomBoolean if attributes.contains("true") && attributes.contains("false") =>
+          InstantiatedCommand(if (b.value) attributes.get("true").head else attributes.get("false").head).validNel
+        case f: WomFile if !representTheSamePath(f, value) =>
+          // Files generated by writeFiles have "engine-relative" paths which will be different from the container paths
+          // calculated by `valueMapper`. "engine-relative" may mean either the non-Docker container path on the host
+          // running Cromwell, or a cloud path. Capture these newly created files and their engine paths here.
+          val enginePathFile = value.asInstanceOf[WomFile]
```
Horneth: ToL: this assumes that the `valueMapper` maps only `WomFile`s to `WomFile`s right ?
Since it's provided by the backend and could potentially be anything we might want to safe guard against weird value mappers here that would map `WomSomethingElse` to `WomFile`. Or at least leave a comment.

mcovarr: Yeah this isn't quite right. I think it was right at one point but I refactored it to be not right. Good catch!

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -231,14 +228,16 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
...
       .replace("SCRIPT_EPILOGUE", scriptEpilogue))
   }
 
   /** The instantiated command. */
-  lazy val instantiatedCommand: String =
+  lazy val instantiatedCommand: InstantiatedCommand = {
+    val runtimeEnvironment = RuntimeEnvironmentBuilder(jobDescriptor.runtimeAttributes, jobPaths)(standardParams.minimumRuntimeSettings)
     Command.instantiate(
-      jobDescriptor, backendEngineFunctions, commandLinePreProcessor, commandLineValueMapper, RuntimeEnvironmentBuilder(jobDescriptor.runtimeAttributes, jobPaths)(standardParams.minimumRuntimeSettings)).get
+      jobDescriptor, backendEngineFunctions, commandLinePreProcessor, commandLineValueMapper, runtimeEnvironment).toTry.get
```
cjllanwarne: `toTry.get`?

centaur/src/main/resources/standardTestCases/fofn_caching.test
```diff
@@ -1,4 +1,4 @@
-ignore: true # Depends on 2864
+ignore: true # Depends on 2864 and also some TODO WOM fun in cromwell.engine.WdlFunctions in 2922
```
cjllanwarne: No longer on #2864 once this merges

cwl/src/main/scala/cwl/CwlExpressionCommandPart.scala
```diff
@@ -1,76 +1,79 @@
...
 
 case class CwlExpressionCommandPart(expr: Expression) extends CommandPart {
   override def instantiate(inputsMap: Map[LocalName, WomValue],
                            functions: IoFunctionSet,
                            valueMapper: (WomValue) => WomValue,
-                           runtimeEnvironment: RuntimeEnvironment ): String = {
-
-    val stringKeyMap = inputsMap.map{ case (LocalName(localName), value) => localName -> value }
+                           runtimeEnvironment: RuntimeEnvironment ): ErrorOr[InstantiatedCommand] =
+    Try {
```
cjllanwarne: What might `throw` here, and can we wrap that small piece instead of wrapping the whole top level function here?
(also applies on the `Try`s a few times below)

supportedBackends/sfs/src/main/scala/cromwell/backend/impl/sfs/config/ConfigAsyncJobExecutionActor.scala
```diff
@@ -47,8 +51,56 @@ sealed trait ConfigAsyncJobExecutionActor extends SharedFileSystemAsyncJobExecut
...
+     *   --entrypoint /bin/bash \
+     *   -v ${cwd}:${docker_cwd} \
+     *   ${docker} ${script}
+     *  """
+     *
+     * The `docker_user` variable is an uninitialized optional. But the expression `${"--user " + docker_user}` will
+     * fail to evaluate if `docker_user` is not provided as an input by the caller of `TaskDefinition#instantiateCommand`.
+     * This only appears to be the case when calling `instantiateCommand` on a `TaskDefinition` that came from
+     * `WdlTask#womDefinition`, which makes it seem that there is some special initialization of task optionals taking
+     * place when create a `WomExecutable` from a `WdlWorkflow`.
```
cjllanwarne: I believe the string interpolation does indeed do some magic for optionals along the lines of "if it doesn't exist, use the `default=` tag, and if that doesn't exist, use `""`"

cjllanwarne: So in this case, I think they might actually not need further "fixing" (as long as set values aren't overridden to `None` too?)

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkJobExecutionActor.scala
```diff
@@ -172,7 +174,7 @@ class SparkJobExecutionActor(override val jobDescriptor: BackendJobDescriptor,
         SparkCommands.ExecutorCores -> runtimeAttributes.executorCores,
         SparkCommands.ExecutorMemory -> runtimeAttributes.executorMemory.toString.replaceAll("\\s","").toLowerCase,
         SparkCommands.AdditionalArgs -> runtimeAttributes.additionalArgs.getOrElse(""),
-        SparkCommands.SparkAppWithArgs -> command.get,
+        SparkCommands.SparkAppWithArgs -> command.toTry.get,
```
cjllanwarne: I'm gonna flag this one too since it isn't in a `Spec`

wdl/src/main/scala/wdl/WdlTask.scala
```diff
@@ -109,93 +111,27 @@ case class WdlTask(name: String,
     * @param valueMapper Optional WomValue => WomValue function that is called on the result of each expression
     *                    evaluation (i.e. evaluation of $${...} blocks).
     * @return String instantiation of the command
+    *
+    * TODO WOM this method should go away in the #2944 purge
```
cjllanwarne: If the test is only hitting this function, and this function is only used in that test, can we just chuck them both out immediateyly?

wdl/src/main/scala/wdl/command/ParameterCommandPart.scala
```diff
@@ -49,29 +45,45 @@ case class ParameterCommandPart(attributes: Map[String, String], expression: Wdl
...
 
-    valueMapper(value) match {
-      case b: WomBoolean if attributes.contains("true") && attributes.contains("false") => if (b.value) attributes.get("true").head else attributes.get("false").head
-      case p: WomPrimitive => p.valueString
-      case a: WomArray if attributes.contains("sep") => a.value.map(_.valueString).mkString(attributes.get("sep").head)
-      case _: WomArray => throw new UnsupportedOperationException(s"Expression '${expression.toString}' evaluated to an Array but no 'sep' was specified")
-      case _ => throw new UnsupportedOperationException(s"Could not string-ify value: $value")
-    }
-  }
+    // Create the stringified version of the command and record any file created by `writeFile`.
```
cjllanwarne: `"by 'writeFile"` => `"in the process"`?

Is there any chance CWL might like this to be in the WOM layer?

wdl/src/main/scala/wdl/examples/ex7.scala
```diff
@@ -37,7 +38,7 @@ object ex7 {
     ns.taskCalls.find( _.unqualifiedName == "a") foreach { call =>
       val wdlFunctions: CustomFunctions = new CustomFunctions
       val evaluatedInputs = call.evaluateTaskInputs(inputs, wdlFunctions).get
-      println(call.task.instantiateCommand(evaluatedInputs, wdlFunctions).get)
+      println(call.task.instantiateCommand(evaluatedInputs, wdlFunctions).toTry.get)
```
cjllanwarne: I see that this is in `src/main`, but it feels like more of a test to me? Can we move it?

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
mcovarr: This went from `.get` to `.toTry.get` since the callee returns a `Try` rather than an `ErrorOr`. Yes there's a lot of tech debt here, I'm just hoping I didn't make it worse.

centaur/src/main/resources/standardTestCases/fofn_caching.test
mcovarr: true dat

cwl/src/main/scala/cwl/CwlExpressionCommandPart.scala
mcovarr: Honestly I have no idea and didn't have time to investigate.

supportedBackends/sfs/src/main/scala/cromwell/backend/impl/sfs/config/ConfigAsyncJobExecutionActor.scala
mcovarr: Let's discuss this IRL tomorrow.

supportedBackends/spark/src/main/scala/cromwell/backend/impl/spark/SparkJobExecutionActor.scala
mcovarr: This is existing tech debt.

wdl/src/main/scala/wdl/command/ParameterCommandPart.scala
mcovarr: I don't know but we can refactor if that turns out to be the case.

wdl/src/main/scala/wdl/examples/ex7.scala
mcovarr: That seems out of scope for what this PR is trying to do.

mcovarr: There's also the larger concern that all those examples are demonstrating usage of an API we might not want to people to use anymore.

PR #2947, opened 2017-11-28
Reinstating member access

cjllanwarne:
1-minute description of this: VariableReference had to become aware of where it was, because `p.left` might need a `Pair` called `p`'s `"left"` field, or it might need a `task` called `p`'s `"left"` output, depending on its scope, and the variable reference is different in each case (`p` and `p.left` respectively).

Determining whether a reference is a member access or a task output reference is a bit inefficient right now, but I think it should be ok since it's only called during WDL instantiation (thereafter it's all just following the links in WOM objects)


wdl/src/main/scala/wdl/AstTools.scala
```diff
@@ -304,21 +304,40 @@ object AstTools {
...
+      lazy val popLastNamePiece: Option[String] = {
+        name.lastIndexOf(".") match {
+          case -1 => None
+          case i => Some(name.substring(0, i))
+        }
+      }
+
+      from.resolveVariable(name) match {
+        case Some(_: WdlTaskCall) => Some(previous)
+        case Some(_) => Some(name)
```
cjllanwarne: Need to treat workflow calls like task calls too

mcovarr: Yes and `Option(name)`.

wdl/src/main/scala/wdl/WdlExpression.scala
```diff
@@ -215,7 +215,7 @@ final case class WdlWomExpression(wdlExpression: WdlExpression, from: Option[Sco
   override def evaluateType(inputTypes: Map[String, WomType]): ErrorOr[WomType] =
     // All current usages of WdlExpression#evaluateType trace back to WdlNamespace, but this is not the
     // case in the brave new WOM-world.
-    wdlExpression.evaluateType(inputTypes.apply, new WdlStandardLibraryFunctionsType, from).toErrorOr
+    wdlExpression.evaluateType(inputTypes.apply, new WdlStandardLibraryFunctionsType, Some(from)).toErrorOr
```
mcovarr: `Option`

centaur/src/main/resources/standardTestCases/array_and_map_indexing/array_and_map_indexing.wdl
```diff
@@ -1,32 +1,20 @@
...
   Map[String, Int] m = {"a": 0, "b": 1, "c": 2}
   Array[String] a = ["foo", "bar", "baz"]
+  Pair[String, String] p = ("left", "right")
+
   call echo_str {
-    input: s = a[1]
-  }
-  call echo_int {
-    input: i = m["c"]*100
+    input: s = [a[1], m["c"], p.left]
```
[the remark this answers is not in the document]
mcovarr: Your description states the need to disambiguate `.left` accesses on `Pair`s from accesses for an output named `pair`; it might be good to actually test that.

geoffjentry:
Some(codeReviewComment)

wdl/src/main/scala/wdl/AstTools.scala
```diff
@@ -304,21 +304,40 @@ object AstTools {
...
     }
 
-    def fullVariableReferenceString: String = terminal.getSourceString + (terminalSubIdentifier map { "." + _.getSourceString } getOrElse "")
+    private lazy val fullVariableReferenceString: String = terminal.getSourceString + (terminalSubIdentifier map { "." + _.getSourceString } getOrElse "")
+
+    private def findResolvableSubstring(name: String, previous: String): Option[String] = {
+      lazy val popLastNamePiece: Option[String] = {
+        name.lastIndexOf(".") match {
+          case -1 => None
+          case i => Some(name.substring(0, i))
```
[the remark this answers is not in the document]
cjllanwarne: Bah, humbug

PR #2948, opened 2017-11-29
Run jes centaur on travis

kshakir:
PR 2 of 3:
1. Remove cromwell-core dependency from cloud-support
2. Run jes centaur on travis
3. Generate coverage for integration tests

---

Centaur now uses configurable cloudSupport for auth, instead of always using application-default.
Refresh token is passed as a file path, then read by centaur.


centaur/src/main/scala/centaur/test/Test.scala
```diff
@@ -76,17 +81,46 @@ object Test {
...
+  lazy val configuration: GoogleConfiguration = GoogleConfiguration(CentaurConfig.conf)
+  lazy val googleConf: Config = CentaurConfig.conf.getConfig("google")
+  lazy val authName: String = googleConf.getString("auth")
+  lazy val genomicsEndpointUrl: String = googleConf.getString("genomics.endpoint-url")
+  lazy val credentials: Credentials = configuration.auth(authName).toTry.get.credential(Map.empty)
+  lazy val credentialsProjectOption: Option[String] = {
+    credentials match {
+      case serviceAccountCredentials: ServiceAccountCredentials => Option(serviceAccountCredentials.getProjectId)
+      case _ => None
+    }
```
mcovarr: I think this is equivalent
```
lazy val credentialsProjectOption: Option[String] = Option(credentials) collect { case sac: ServiceAccountCredentials => sac.getProjectId } 
```

geoffjentry:
Ooooh none of us could remember the exact reason but I think you’re right. And iirc there was some performance problem which was eventually tracked down to be something silly we hadn’t accounted for. This makes sense

kshakir:
Travis has been running cromwell+centaur+funnel for a while now. If it can't run cromwell+centaur+(sending-files-off-to-gcloud)... we have other issues. ;)

mcovarr:
Yeah I recall Jeff tried a simpler setup like this first and then had to resort to nesting Cromwells when that didn't work. But clearly Cromwell has become that much more efficient so now it does work hooray!

Issue #2905, opened 2017-11-15
Wom: Reinstate invalidate bad caches local


cjllanwarne:
This might be related to centaur configuration, which IIRC requires specific, non-default, options for this test. NB fails on both `centaurLocal` and `centaurTest`

Once fixed, reinstate centaur test:
- [x] `invalidate_bad_caches_local`

kshakir:
Close this ticket once the test has been turned back on and verified. PR #2952 should fix this test as well.

cjllanwarne:
Fixed in #2956

PR #2952, opened 2017-11-30
Reinstates workflow outputs (et al)

cjllanwarne:
As well as allowing workflow outputs to refer to other outputs (commit 1), to make the four tests in this ticket work involved:
- Making the result from not-run conditionals be of type `X?`, not the incorrect `X??` (commit 2)
- Making task input evaluation honor optionals and defaults, and be evaluated in the right order (commit 3)

### DIFF
--- a/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/job/preparation/InputPointerToWdlValue.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/job/preparation/InputPointerToWdlValue.scala
@@ -1,31 +1,41 @@
 package cromwell.engine.workflow.lifecycle.execution.job.preparation
 
+import cats.data.Validated.Valid
 import cats.syntax.validated._
 import cromwell.core.ExecutionIndex.ExecutionIndex
 import common.validation.ErrorOr.ErrorOr
 import cromwell.engine.workflow.lifecycle.execution.stores.ValueStore
 import shapeless.Poly1
+import wom.callable.Callable.{InputDefinition, InputDefinitionWithDefault, OptionalInputDefinition}
 import wom.expression.{IoFunctionSet, WomExpression}
-import wom.graph.GraphNode
 import wom.graph.GraphNodePort.OutputPort
 import wom.values.WomValue
 
 object InputPointerToWdlValue extends Poly1 {
   // Function that can transform any of the coproduct types to an ErrorOr[WomValue]
-  type ToWdlValueFn = (GraphNode, Map[String, WomValue], IoFunctionSet, ValueStore, ExecutionIndex) => ErrorOr[WomValue]
+  type ToWdlValueFn = (Map[String, WomValue], IoFunctionSet, ValueStore, ExecutionIndex, InputDefinition) => ErrorOr[WomValue]
 
-  implicit def fromWdlValue: Case.Aux[WomValue, ToWdlValueFn] = at[WomValue] {
-    WomValue => (_: GraphNode, _: Map[String, WomValue], _: IoFunctionSet, _: ValueStore, _: ExecutionIndex) =>
-      WomValue.validNel: ErrorOr[WomValue]
+  implicit def fromWomValue: Case.Aux[WomValue, ToWdlValueFn] = at[WomValue] {
+    womValue => (_: Map[String, WomValue], _: IoFunctionSet, _: ValueStore, _: ExecutionIndex, _: InputDefinition) =>
+      womValue.validNel: ErrorOr[WomValue]
   }
 
   implicit def fromOutputPort: Case.Aux[OutputPort, ToWdlValueFn] = at[OutputPort] {
-    port => (_: GraphNode, _: Map[String, WomValue], _: IoFunctionSet, valueStore: ValueStore, index : ExecutionIndex) =>
-      valueStore.resolve(index)(port)
+    port => (knownValues: Map[String, WomValue], ioFunctions: IoFunctionSet, valueStore: ValueStore, index: ExecutionIndex, inputDefinition: InputDefinition) =>
+      (valueStore.resolve(index)(port), inputDefinition) match {
+        case (v: Valid[WomValue], _) => v
+        case (_, InputDefinitionWithDefault(_, _, defaultExpression)) =>
+          evaluate(defaultExpression, knownValues, ioFunctions)
+        case (_, OptionalInputDefinition(_, optionalType)) => optionalType.none.validNel
+        case _ => s"Failed to lookup input value for required input ${port.name} at index $index in value store $valueStore".invalidNel
+      }
   }
 
   implicit def fromWomExpression: Case.Aux[WomExpression, ToWdlValueFn] = at[WomExpression] { 
-    womExpression => (_: GraphNode, knownValues: Map[String, WomValue], ioFunctions: IoFunctionSet, _: ValueStore, _ : ExecutionIndex) =>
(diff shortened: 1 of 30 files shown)


wdl/src/test/scala/wdl/WdlWorkflowSpec.scala
```diff
@@ -130,14 +131,35 @@ class WdlWorkflowSpec extends WordSpec with Matchers {
       }
     }
 
+    def evaluateOutputs(workflow: WdlWorkflow,
```
cjllanwarne: NB since this is now only used in this test, I've moved it into the test class which used it

kshakir:
@cjllanwarne Please also reinstate `invalidate_bad_caches_local` in your branch. I made a similar fix for the conditionals ([in a different area](https://github.com/broadinstitute/cromwell/commit/b0195c6956ac761349efbb3e75aa5311c37583ce)) while investigating and fixing ticket #2905. The conditionals issue fixes `invalidate_bad_caches_local`.

As your branch has even more changes in it, I will delete mine once this PR is merged.

cc @gemlam3

wom/src/main/scala/wom/graph/CallNode.scala
```diff
@@ -106,14 +106,14 @@ object CallNode {
...
-  final case class InputDefinitionFold(mappings: InputDefinitionMappings = Map.empty,
+  final case class InputDefinitionFold(mappings: InputDefinitionMappings = List.empty,
                                        callInputPorts: Set[InputPort] = Set.empty,
                                        newGraphInputNodes: Set[ExternalGraphInputNode] = Set.empty,
                                        newExpressionNodes: Set[ExpressionNode] = Set.empty,
                                        usedOuterGraphInputNodes: Set[OuterGraphInputNode] = Set.empty)
 
   type InputDefinitionPointer = OutputPort :+: WomExpression :+: WomValue :+: CNil
-  type InputDefinitionMappings = Map[InputDefinition, InputDefinitionPointer]
+  type InputDefinitionMappings = List[(InputDefinition, InputDefinitionPointer)]
```
kshakir: Please leave a comment as to why one shouldn't come back and refactor this into a `Map`.

ToL: After our chat about `Set`s (including key sets) yesterday, I'm more than happy to have this as a list, but I'm curious as to if this was a bug fix, or a preference fix. If this is a bug fix, do the centaur tests fix this if it's a regression? Someday I'd love to also see more unit tests in this area too.

cjllanwarne: Yes, there's a centaur test called `optional_parameter` that was failing before this change was made.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/job/preparation/InputPointerToWdlValue.scala
```diff
@@ -1,31 +1,45 @@
...
   // Function that can transform any of the coproduct types to an ErrorOr[WomValue]
-  type ToWdlValueFn = (GraphNode, Map[String, WomValue], IoFunctionSet, ValueStore, ExecutionIndex) => ErrorOr[WomValue]
+  type ToWdlValueFn = (Map[String, WomValue], IoFunctionSet, ValueStore, ExecutionIndex, InputDefinition) => ErrorOr[WomValue]
 
-  implicit def fromWdlValue: Case.Aux[WomValue, ToWdlValueFn] = at[WomValue] {
-    WomValue => (_: GraphNode, _: Map[String, WomValue], _: IoFunctionSet, _: ValueStore, _: ExecutionIndex) =>
-      WomValue.validNel: ErrorOr[WomValue]
+  implicit def fromWomValue: Case.Aux[WomValue, ToWdlValueFn] = at[WomValue] {
+    womValue => (_: Map[String, WomValue], _: IoFunctionSet, _: ValueStore, _: ExecutionIndex, _: InputDefinition) =>
+      println(s"Using WomValue $womValue")
```
mcovarr: 👁

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/job/preparation/InputPointerToWdlValue.scala
```diff
@@ -1,31 +1,45 @@
...
+      womValue.validNel: ErrorOr[WomValue]
   }
 
   implicit def fromOutputPort: Case.Aux[OutputPort, ToWdlValueFn] = at[OutputPort] {
-    port => (_: GraphNode, _: Map[String, WomValue], _: IoFunctionSet, valueStore: ValueStore, index : ExecutionIndex) =>
-      valueStore.resolve(index)(port)
+    port => (knownValues: Map[String, WomValue], ioFunctions: IoFunctionSet, valueStore: ValueStore, index: ExecutionIndex, inputDefinition: InputDefinition) =>
+      println(s"Looking up output port ${port.name}")
```
mcovarr: 🙄

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/job/preparation/InputPointerToWdlValue.scala
```diff
@@ -1,31 +1,45 @@
...
+      (valueStore.resolve(index)(port), inputDefinition) match {
+        case (v: Valid[WomValue], _) => v
+        case (_, InputDefinitionWithDefault(_, _, defaultExpression)) =>
+          println(s"Falling back to evaluating expression ${port.name} = ${defaultExpression.sourceString}")
```
mcovarr: 😮

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/job/preparation/InputPointerToWdlValue.scala
```diff
@@ -1,31 +1,45 @@
...
+          evaluate(defaultExpression, knownValues, ioFunctions)
+        case (_, OptionalInputDefinition(_, optionalType)) => optionalType.none.validNel
+        case _ => s"Failed to lookup input value for required input ${port.name} at index $index in value store $valueStore".invalidNel
+      }
   }
 
   implicit def fromWomExpression: Case.Aux[WomExpression, ToWdlValueFn] = at[WomExpression] { 
-    womExpression => (_: GraphNode, knownValues: Map[String, WomValue], ioFunctions: IoFunctionSet, _: ValueStore, _ : ExecutionIndex) =>
+    womExpression => (knownValues: Map[String, WomValue], ioFunctions: IoFunctionSet, _: ValueStore, _ : ExecutionIndex, _: InputDefinition) =>
+      println(s"Using WOM expression ${womExpression.sourceString}")
```
mcovarr: :no_good_man:

wom/src/main/scala/wom/graph/CallNode.scala
mcovarr: I worry about having centaur tests that exist to test other things and only incidentally test the important functionality here. i.e. those tests might get refactored and we wouldn't know if this was still working. tl;dr can we have an explicit test

cjllanwarne: I dunno whether I'd agree about it being incidental... the only reason this is a `List` is so that we can evaluate inputs in the correct order. 

If someone somehow refactors this and that test still works... I'm quite happy to let that happen

cjllanwarne: `select_first_with_optionals` is an explicit test of using optional inputs in subsequent expressions, so I don't think it's likely to be refactored. 

To make it even less likely I've added a comment to preserve that feature of the test.

PR #2959, opened 2017-11-30
Add flatten type detection

cjllanwarne:
@orodeh it looks like the `WdlStandardLibraryFunctionsType#flatten` method was missing, which meant Cromwell wasn't able to validate workflows with `flatten`s in them.

I also added the coercion so that `flatten(_: Array[Map[_]])` works as expected.


orodeh:
I see, so the function to calculate the type was missing. Sorry about that, I didn't have a test to check it. 

👍

PR #2960, opened 2017-12-01
fix fofn caching

mcovarr:


engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -430,6 +433,29 @@ case class WorkflowExecutionActor(params: WorkflowExecutionActorParams)
...
+      callNode <- (Option(taskCallNode) collect { case t: TaskCallNode => t }).toValidNel(s"Not a TaskCallNode: $taskCallNode").toEither
+      jobKey <- data.executionStore.keyForNode(callNode).toValidNel(s"No job key found for call node $callNode").toEither
+      backendJobDescriptorKey <- Option(jobKey).collectFirst({ case c: BackendJobDescriptorKey => c }).toValidNel(s"Job key not a BackendJobDescriptorKey: $jobKey").toEither
+      backendName <- workflowDescriptor
+        .backendAssignments.get(callNode).toValidNel(s"Cannot find an assigned backend for call ${callNode.fullyQualifiedName}").toEither
+      backendInitializationData = params.initializationData.get(backendName)
+      backendFactory <- backendFactories.get(backendName).toValidNel(s"Cannot find a backend factory for backend $backendName").toEither
+      functions = backendFactory.expressionLanguageFunctions(workflowDescriptor.backendDescriptor, backendJobDescriptorKey, backendInitializationData)
+      diff <- key.processRunnable(functions, data.valueStore, self).toEither
+    } yield diff).toValidated
```
kshakir: Huge ToL: I could be way off here, but I'm curious if someday cromwell could use _a lot less_ conversion back-and-forth between `Try`/`Either`/`Coproduct`/`Validated`/etc.

mcovarr: Yeah I'm not super happy with the way it is now, and it's not a great excuse but this follows the pattern elsewhere in this file. In the short term I'll attempt to DRY and sugar a bit but I very much agree with what you're saying.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -467,18 +499,16 @@ case class WorkflowExecutionActor(params: WorkflowExecutionActorParams)
...
-    } yield startEJEA(key, backendName, backendFactory, command)).toValidated
+    } yield startEJEA(key, factory, command)).toValidated
   }
 
   private def startEJEA(jobKey: BackendJobDescriptorKey,
-                        backendName: String,
                         backendFactory: BackendLifecycleActorFactory,
                         command: BackendJobExecutionActorCommand): WorkflowExecutionDiff = {
     val ejeaName = s"${workflowDescriptor.id}-EngineJobExecutionActor-${jobKey.tag}"
+    val backendName = backendFactory.name
```
cjllanwarne: Just want to check: this is valid if there are two instances of the same backend active at once, right?

wom/src/main/scala/wom/graph/GraphNode.scala
```diff
@@ -80,6 +82,9 @@ object GraphNode {
     var _graphNode: A = _
     private def getGraphNode = _graphNode
     def get: Unit => A = _ => getGraphNode
+    def isAssignedAndOfType[T: ClassTag]: Boolean = {
+      Option(_graphNode) exists classTag[T].runtimeClass.isInstance
```
cjllanwarne: FMI: Is this better than .isInstanceOf, or a simple match?

wom/src/main/scala/wom/graph/expression/AnonymousExpressionNode.scala
```diff
@@ -23,4 +26,8 @@ object AnonymousExpressionNode {
 /**
   * An expression node that is purely an internal expression and shouldn't be visible outside the graph
   */
-trait AnonymousExpressionNode extends ExpressionNode
+trait AnonymousExpressionNode extends ExpressionNode {
+  val callNodeReceivingInput: GraphNodeSetter[CallNode]
```
cjllanwarne: Maybe Scala docs are a good idea on this, given our uneasiness over the name?

wom/src/main/scala/wom/graph/expression/AnonymousExpressionNode.scala
```diff
@@ -23,4 +26,8 @@ object AnonymousExpressionNode {
...
+
+  def feedsTaskCallNode: Boolean = callNodeReceivingInput.isAssignedAndOfType[TaskCallNode]
```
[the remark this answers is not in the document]
cjllanwarne: Oh, I see, the field isn’t set for Ifs and Scatters?

I’m a little uneasy about “how do we know it wasn’t a task but just wasn’t set yet”... but I think it’s pretty safe. Hopefully!

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -430,17 +434,45 @@ case class WorkflowExecutionActor(params: WorkflowExecutionActorParams)
...
+   /*
+    * If this ExpressionKey has an AnonymousExpressionNode that feeds a task call input, use the backend IoFunctionSet
+    * instead of the engine's IoFunctionSet to properly handle `writeFile` or `readFile` invocations.
+    */
+  private def processRunnableTaskCallInputExpression(key: ExpressionKey,
+                                                     data: WorkflowExecutionActorData,
+                                                     anonymousExpressionNode: AnonymousExpressionNode): ErrorOr[WorkflowExecutionDiff] = {
+    import cats.syntax.either._
+
+    val callNode = anonymousExpressionNode.callNodeReceivingInput.get(())
```
cjllanwarne: TOLIf this get is unsafe, can we move it into the for comp?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -430,17 +434,45 @@ case class WorkflowExecutionActor(params: WorkflowExecutionActorParams)
...
+      diff <- key.processRunnable(functions, data.valueStore, self).toEither
+    } yield diff).toValidated
+  }
+
+  private def backendFactoryForTaskCallNode(taskCallNode: TaskCallNode): Checked[BackendLifecycleActorFactory] = {
+    for {
+      name <- workflowDescriptor
+        .backendAssignments.get(taskCallNode).toChecked(s"Cannot find an assigned backend for call ${taskCallNode.fullyQualifiedName}")
+      factory <- backendFactories.get(name).toChecked(s"Cannot find a backend factory for backend $name")
+    } yield factory
```
cjllanwarne: TOL curious if there’s a reason for splitting this out from the main for comp?

mcovarr: This chunk of logic is called in two spots, the evaluate-an-anonymous-expression-that-feeds-a-task-call code and the actual okay-now-actually-go-run-the-task-call code.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
mcovarr: The code claims the name corresponds to the name in the config.

wom/src/main/scala/wom/graph/expression/AnonymousExpressionNode.scala
mcovarr: yes

wom/src/main/scala/wom/graph/expression/AnonymousExpressionNode.scala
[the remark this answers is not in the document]
mcovarr: I think that should be safe since those relationships are established when the task call is built, but the WEA doesn't look at this until after the task call is built and in the execution store.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/stores/ExecutionStore.scala
```diff
@@ -130,6 +130,10 @@ sealed abstract class ExecutionStore private[stores](statusStore: Map[JobKey, Ex
   // View of the statusStore more suited for lookup based on status
   lazy val store: Map[ExecutionStatus, List[JobKey]] = statusStore.groupBy(_._2).mapValues(_.keys.toList)
 
+  def keyForNode(node: GraphNode): Option[JobKey] = {
```
cjllanwarne: ToL if it doesn't break an abstraction layer, maybe this fits in `EnhancedGraphNode` above?

wom/src/main/scala/wom/graph/GraphNode.scala
mcovarr: It turns out to be required for an API that looks like this since otherwise the type is erased. I'll leave a comment taunting the reader to improve it if they can.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/stores/ExecutionStore.scala
mcovarr: Yes exactly. I wanted to move that other big chunk of logic over here as well and had a similar problem.

PR #2971, opened 2017-12-02
Include additional keys in workflow query response

samanehsan:
Update the `query` endpoint to take in an `additionalKeys` parameter, which will include the specified keys in the returned workflow metadata. This currently just supports "labels" and "parentWorkflowId" but can be expanded later if necessary.

E.g. A GET request to `http://localhost:8000/api/workflows/v1/query?additionalKeys=parentWorkflowId` would return: 

```
{
  "results": [
    {
      "name": "test",
      "id": "4e1bf4b5-caea-4e8a-9cd3-f48ac8d11393",
      "status": "Failed",
      "parentWorkflowId": "21b58d24-7f14-4e45-a45a-5e8678fd83cb",
      "end": "2017-12-01T19:00:23.778-05:00",
      "start": "2017-12-01T19:00:16.738-05:00"
    }
  ]
}
```

A GET request to `http://localhost:8000/api/workflows/v1/query?additionalKeys=labels` would return: 

```
{
  "results": [
    {
      "name": "use_sub_wfs",
      "id": "21b58d24-7f14-4e45-a45a-5e8678fd83cb",
      "labels": {
        "project": "test",
        "version": "2",
        "cromwell-workflow-id": "cromwell-21b58d24-7f14-4e45-a45a-5e8678fd83cb"
      },
      "status": "Failed",
      "end": "2017-12-01T19:00:24.627-05:00",
      "start": "2017-12-01T19:00:14.210-05:00"
    }
  ]
}
```

### DIFF
--- a/engine/src/main/resources/swagger/cromwell.yaml
+++ b/engine/src/main/resources/swagger/cromwell.yaml
@@ -356,6 +356,16 @@ paths:
             returns workflows with all of the specified label keys. Specify the label key
             and label value pair as separated with
             "label-key:label-value"
+        - name: additionalKeys
+          required: false
+          in: query
+          type: array
+          items:
+            type: string
+          collectionFormat: multi
+          pattern: ^[a-zA-Z][a-zA-Z0-9_]*$
+          description: >
+            Includes the specified keys in the metadata for the returned workflows.
       tags:
         - Workflows
       responses:
--- a/engine/src/main/resources/swagger/cromwell.yaml
+++ b/engine/src/main/resources/swagger/cromwell.yaml
@@ -356,7 +356,7 @@ paths:
             returns workflows with all of the specified label keys. Specify the label key
             and label value pair as separated with
             "label-key:label-value"
-        - name: additionalKeys
+        - name: additionalQueryResultFields
           required: false
           in: query
           type: array
--- a/engine/src/main/resources/swagger/cromwell.yaml
+++ b/engine/src/main/resources/swagger/cromwell.yaml
@@ -363,7 +363,6 @@ paths:
           items:
             type: string
           collectionFormat: multi
-          pattern: ^[a-zA-Z][a-zA-Z0-9_]*$
           description: >
             Includes the specified keys in the metadata for the returned workflows.
       tags:
--- a/engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
+++ b/engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
@@ -20,6 +20,7 @@ import cromwell.services.healthmonitor.HealthMonitorServiceActor.{GetCurrentStat
 import cromwell.services.metadata.MetadataService._
 import cromwell.services.metadata._
 import cromwell.util.SampleWdl.HelloWorld
+import mouse.boolean._
 import org.scalatest.{AsyncFlatSpec, Matchers}
 import spray.json.DefaultJsonProtocol._
 import spray.json._
@@ -488,6 +489,33 @@ class CromwellApiServiceSpec extends AsyncFlatSpec with ScalatestRouteTest with 
         }
     }
 
+    it should "return labels if specified in additionalQueryResultFields param" in {
+      Get(s"/workflows/$version/query?additionalQueryResultFields=labels&id=${CromwellApiServiceSpec.ExistingWorkflowId}") ~>
+        akkaHttpService.workflowRoutes ~>
+        check {
+          status should be(StatusCodes.OK)
+          contentType should be(ContentTypes.`application/json`)
+          val results = responseAs[JsObject].fields("results").convertTo[Seq[JsObject]]
+          val fields = results.head.fields
+          fields("id") should be(JsString(CromwellApiServiceSpec.ExistingWorkflowId.toString))
+          fields(WorkflowMetadataKeys.Labels).asJsObject.fields("key1") should be(JsString("label1"))
+          fields(WorkflowMetadataKeys.Labels).asJsObject.fields("key2") should be(JsString("label2"))
+        }
+    }
+
+    it should "return parentWorkflowId if specified in additionalQueryResultFields param" in {
+      Get(s"/workflows/$version/query?additionalQueryResultFields=parentWorkflowId&id=${CromwellApiServiceSpec.ExistingWorkflowId}") ~>
+        akkaHttpService.workflowRoutes ~>
+        check {
+          status should be(StatusCodes.OK)
+          contentType should be(ContentTypes.`application/json`)
+          val results = responseAs[JsObject].fields("results").convertTo[Seq[JsObject]]
+          val fields = results.head.fields
+          fields("id") should be(JsString(CromwellApiServiceSpec.ExistingWorkflowId.toString))
+          fields(WorkflowMetadataKeys.ParentWorkflowId) should be(JsString("pid"))
+        }
+    }
+
     behavior of "REST API /query POST endpoint"
     it should "return good results for a good query map body" in {
       Post(s"/workflows/$version/query", HttpEntity(ContentTypes.`application/json`, """[{"status":"Succeeded"}]""")) ~>
@@ -502,34 +530,6 @@ class CromwellApiServiceSpec extends AsyncFlatSpec with ScalatestRouteTest with 
         }
     }
 
-    behavior of "REST API /query GET endpoint"
-    it should "return labels if specified in additionalQueryResultFields param" in {
-      Get(s"/workflows/$version/query?additionalQueryResultFields=labels&id=${CromwellApiServiceSpec.ExistingWorkflowId}") ~>
-        akkaHttpService.workflowRoutes ~>
-        check {
-          status should be(StatusCodes.OK)
-          contentType should be(ContentTypes.`application/json`)
-          val results = responseAs[JsObject].fields("results").convertTo[Seq[JsObject]]
-          results.head.fields("id") should be(JsString(CromwellApiServiceSpec.ExistingWorkflowId.toString))
-          results.head.fields(WorkflowMetadataKeys.Labels).asJsObject.fields("key1") should be(JsString("label1"))
-          results.head.fields(WorkflowMetadataKeys.Labels).asJsObject.fields("key2") should be(JsString("label2"))
-        }
-    }
-
-    behavior of "REST API /query GET endpoint"
-    it should "return parentWorkflowId if specified in additionalQueryResultFields param" in {
-      Get(s"/workflows/$version/query?additionalQueryResultFields=parentWorkflowId&id=${CromwellApiServiceSpec.ExistingWorkflowId}") ~>
-        akkaHttpService.workflowRoutes ~>
-        check {
-          status should be(StatusCodes.OK)
-          contentType should be(ContentTypes.`application/json`)
-          val results = responseAs[JsObject].fields("results").convertTo[Seq[JsObject]]
-          results.head.fields("id") should be(JsString(CromwellApiServiceSpec.ExistingWorkflowId.toString))
-          results.head.fields(WorkflowMetadataKeys.ParentWorkflowId) should be(JsString("pid"))
-        }
-    }
-
-    behavior of "REST API /query POST endpoint"
     it should "return labels if specified in additionalQueryResultFields param" in {
       Post(s"/workflows/$version/query", HttpEntity(ContentTypes.`application/json`, """[{"additionalQueryResultFields":"labels"}]""")) ~>
         akkaHttpService.workflowRoutes ~>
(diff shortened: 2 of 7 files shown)


cjllanwarne:
Thanks Saman!

Quick Question: What is “parent workflow ID” here?

I thought that QUERY should only be returning top level workflow result? Otherwise you’re going to get lots of workflow entries in QUERY for the same SUBMIT request which seems odd to me. Maybe it’d be useful to make getting subworkflows opt in/out? But that’s me thinking out loud and not for this PR!

Since we *do* expose subworkflows, we probably should give the “top level” ID (the one returned by SUBMIT) rather than direct “parent” ID.

engine/src/main/resources/swagger/cromwell.yaml
```diff
@@ -356,6 +356,15 @@ paths:
             returns workflows with all of the specified label keys. Specify the label key
             and label value pair as separated with
             "label-key:label-value"
+        - name: additionalKeys
```
mcovarr: Not 100% sure about the name. It isn't clear this is related to the query endpoint, and if it is understood to relate to the query endpoint it sounds like it allows for additional keys (parameters) to the query rather than additional key/value pairs in the results.

services/src/main/scala/cromwell/services/metadata/impl/MetadataDatabaseAccess.scala
```diff
@@ -178,14 +178,73 @@ trait MetadataDatabaseAccess {
...
+    def summaryToQueryResult(workflow: WorkflowMetadataSummaryEntry): Future[MetadataService.WorkflowQueryResult] = {
+
+      def queryResult(labels: Map[String, String], parentId: Option[String]): MetadataService.WorkflowQueryResult = {
+        MetadataService.WorkflowQueryResult(
+          id = workflow.workflowExecutionUuid,
+          name = workflow.workflowName,
+          status = workflow.workflowStatus,
+          start = workflow.startTimestamp map { _.toSystemOffsetDateTime },
+          end = workflow.endTimestamp map { _.toSystemOffsetDateTime },
+          labels = if (labels.nonEmpty) { Some(labels) } else None,
```
mcovarr: `labels = labels.nonEmpty.option(labels)` with `mouse`

services/src/main/scala/cromwell/services/metadata/impl/MetadataDatabaseAccess.scala
```diff
@@ -178,14 +178,73 @@ trait MetadataDatabaseAccess {
...
+
+      def metadataEntriesToValue(entries: Seq[MetadataEntry]): Option[String] = {
+        entries.headOption.flatMap(_.metadataValue.toRawStringOption)
+      }
+
+      def keyToMetadataValue(key: String): Future[Option[String]] = {
+        val metadataEntries: Future[Seq[MetadataEntry]] = metadataDatabaseInterface.queryMetadataEntries(workflow.workflowExecutionUuid, key)
+        metadataEntries map { entries =>
+          metadataEntriesToValue(entries)
+        }
```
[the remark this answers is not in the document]
mcovarr: at which point this might become a one liner, or be inlined 🙂

engine/src/main/resources/swagger/cromwell.yaml
mcovarr: Maybe `additionalQueryResultFields`? 🙂

samanehsan:
@cjllanwarne, the query endpoint currently returns all workflows, not just top-level workflows and an option to include/exclude subworkflows would be ideal. 

The `parentWorkflowId` is the id of the workflow that called this workflow (which could be different from the `rootWorkflowId` if this workflow gets called by a subworkflow). It lets us differentiate between top-level workflows and subworkflows, and could be useful if you have deeply nested workflows. I thought including the `rootWorkflowId` would also tell us whether a workflow is a subworkflow, but I don't see it in the subworkflow metadata.

engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
```diff
@@ -502,6 +502,61 @@ class CromwellApiServiceSpec extends AsyncFlatSpec with ScalatestRouteTest with
...
+          status should be(StatusCodes.OK)
+          contentType should be(ContentTypes.`application/json`)
+          val results = responseAs[JsObject].fields("results").convertTo[Seq[JsObject]]
+          results.head.fields("id") should be(JsString(CromwellApiServiceSpec.ExistingWorkflowId.toString))
+          results.head.fields(WorkflowMetadataKeys.Labels).asJsObject.fields("key1") should be(JsString("label1"))
+          results.head.fields(WorkflowMetadataKeys.Labels).asJsObject.fields("key2") should be(JsString("label2"))
+        }
+    }
+
+    behavior of "REST API /query GET endpoint"
```
cjllanwarne: You should only need this line once, per `behavior of` name, then it's carried forwards by the `it`s

engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
```diff
@@ -502,6 +502,61 @@ class CromwellApiServiceSpec extends AsyncFlatSpec with ScalatestRouteTest with
...
     }
 
+    behavior of "REST API /query GET endpoint"
+    it should "return labels if specified in query additionalKeys param" in {
+      Get(s"/workflows/$version/query?additionalKeys=labels&id=${CromwellApiServiceSpec.ExistingWorkflowId}") ~>
+        akkaHttpService.workflowRoutes ~>
+        check {
+          status should be(StatusCodes.OK)
+          contentType should be(ContentTypes.`application/json`)
+          val results = responseAs[JsObject].fields("results").convertTo[Seq[JsObject]]
```
cjllanwarne: Your `WorkflowJsonSupport` wil allow you to convert from JSON to scala too, if that makes this test easier

engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
```diff
@@ -588,9 +643,23 @@ object CromwellApiServiceSpec {
   class MockServiceRegistryActor extends Actor {
     import MockServiceRegistryActor._
     override def receive = {
-      case WorkflowQuery(_) =>
+      case WorkflowQuery(parameters) =>
+        val labels: Option[Map[String, String]] = {
+          if (parameters.contains(("additionalKeys", "labels"))) {
```
cjllanwarne: Hmm, I'm quite possibly misreading this but it looks like this is saying, 
> "if my test `GET` wants labels then my tests adds labels to the result it returns, and then my test checks that it added the labels"

As far as I can tell, this doesn't actually test that the underlying logic is really adding labels when it needs to?

services/src/main/scala/cromwell/services/metadata/WorkflowQueryKey.scala
```diff
@@ -89,6 +89,20 @@ object WorkflowQueryKey {
...
     }
   }
+
+  case object AdditionalKeys extends SeqWorkflowQueryKey[String] {
+    override val name = "Additionalkeys"
+
+    override def validate(grouped: Map[String, Seq[(String, String)]]): ErrorOr[List[String]] = {
+      val values = valuesFromMap(grouped).toList
+      val allowedValues = Seq(WorkflowMetadataKeys.Labels, WorkflowMetadataKeys.ParentWorkflowId)
+      val nels:List[data.ValidatedNel[String,String]] = values map { v => {
```
cjllanwarne: Add space to `val nels: List[...`

services/src/main/scala/cromwell/services/metadata/WorkflowQueryParameters.scala
```diff
@@ -84,7 +86,8 @@ object WorkflowQueryParameters {
       startDateValidation,
       endDateValidation,
       pageValidation,
-      pageSizeValidation) mapN { (_, _, statuses, names, ids, labels, startDate, endDate, page, pageSize) => WorkflowQueryParameters(statuses, names, ids, labels, startDate, endDate, page, pageSize) }
+      pageSizeValidation,
+      additionalKeysValidation) mapN { (_, _, statuses, names, ids, labels, startDate, endDate, page, pageSize, additionalKeys) => WorkflowQueryParameters(statuses, names, ids, labels, startDate, endDate, page, pageSize, additionalKeys) }
```
cjllanwarne: ~~Is this equivalent to `...) mapN WorkflowQueryParameters`?~~
EDIT: never mind, I missed the two `_`s at the start!

services/src/main/scala/cromwell/services/metadata/impl/MetadataDatabaseAccess.scala
```diff
@@ -178,14 +178,73 @@ trait MetadataDatabaseAccess {
...
-          //only return metadata if page is defined
-          queryParameters.page map { _ => QueryMetadata(queryParameters.page, queryParameters.pageSize, Option(count)) })
+    def queryMetadata(count: Int): Option[QueryMetadata] = {
+      queryParameters.page map { _ => QueryMetadata(queryParameters.page, queryParameters.pageSize, Option(count)) }
+    }
+
+    def summariesToQueryResults(workflows: Traversable[WorkflowMetadataSummaryEntry]): Future[Traversable[MetadataService.WorkflowQueryResult]] = {
+      val mapped: Traversable[Future[MetadataService.WorkflowQueryResult]] = workflows map { summaryToQueryResult }
+      val result: Future[Traversable[MetadataService.WorkflowQueryResult]] = Future.sequence(mapped)
+      result
```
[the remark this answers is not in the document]
cjllanwarne: Also another superfluous `result` variable

services/src/main/scala/cromwell/services/metadata/impl/MetadataDatabaseAccess.scala
```diff
@@ -178,14 +178,73 @@ trait MetadataDatabaseAccess {
...
+
+      def queryResult(labels: Map[String, String], parentId: Option[String]): MetadataService.WorkflowQueryResult = {
+        MetadataService.WorkflowQueryResult(
+          id = workflow.workflowExecutionUuid,
+          name = workflow.workflowName,
+          status = workflow.workflowStatus,
+          start = workflow.startTimestamp map { _.toSystemOffsetDateTime },
+          end = workflow.endTimestamp map { _.toSystemOffsetDateTime },
+          labels = if (labels.nonEmpty) { Some(labels) } else None,
+          parentWorkflowId = parentId
```
cjllanwarne: I think it'd be nicer to wrap the "do we want labels" and "do we want parentWorkflowId" into the original metadata query, so that we only need to go to the database once (in this scheme I think we might end up querying the entire metadata table *three* times per QUERY request).

An added benefit would be that these two lines would be easier to write if the fields were already part of `WorkflowQueryResult`

services/src/main/scala/cromwell/services/metadata/impl/MetadataDatabaseAccess.scala
```diff
@@ -178,14 +178,73 @@ trait MetadataDatabaseAccess {
...
+
+      def getWorkflowLabels: Future[Map[String, String]] = {
+        if (queryParameters.additionalKeys.contains(WorkflowMetadataKeys.Labels)) {
+          metadataDatabaseInterface.getWorkflowLabels(workflow.workflowExecutionUuid)
+        } else {
+          Future.successful(Map.empty)
+        }
+      }
+
+      val getParentWorkflowId: Future[Option[String]] = {
```
cjllanwarne: As mentioned above, I think it's gonna be much nicer and more scalable as we add more possible "additional fields" to make these part of the original query rather than added on as subsequent queries to the DB (since metadata is a *big* table)

engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
samanehsan: How do you suggest modifying the test? Previously, this was always returning the same `WorkflowQueryResponse` no matter what the query was.

services/src/main/scala/cromwell/services/metadata/impl/MetadataDatabaseAccess.scala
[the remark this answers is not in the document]
samanehsan: I'm getting an error when I switch to using the `traverse` method. Do you know where to specify the implicit value for the evidence parameter? 

```
Error: could not find implicit value for evidence parameter of type cats.Applicative[scala.concurrent.Future] workflows.toList.traverse[Future, MetadataService.WorkflowQueryResult](summaryToQueryResult)
```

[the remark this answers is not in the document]
danbills: try `import cats.instances.future._`

services/src/main/scala/cromwell/services/metadata/impl/MetadataDatabaseAccess.scala
samanehsan: Do you mean add these fields to the `WorkflowMetadataSummaryEntry` model? I agree it would be more efficient to get all of the information from the database at the same time, but if we want to support more "additional fields" that would require adding each field to the database model.

mcovarr: IIRC the first query hits the summary table, but you're right that these follow on queries go against the potentially enormous metadata table.

services/src/main/scala/cromwell/services/metadata/impl/MetadataDatabaseAccess.scala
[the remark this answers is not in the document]
samanehsan: thanks @danbills! that fixed it!

samanehsan:
Thanks for the helpful feedback! I've addressed all of the comments if you want to take another look! @cjllanwarne @mcovarr

PR #2974, opened 2017-12-02
Purging old files.

kshakir:


common/.gitignore
```diff
@@ -1,12 +0,0 @@
-# common scala config
-*~
-.DS_Store
-.artifactory
-.idea/*
-!/.idea/inspectionProfiles/
-.idea/inspectionProfiles/*
-!/.idea/inspectionProfiles/Project_Default.xml
-target
```
cjllanwarne: Should we also add `common/target` to the main `.gitignore`?

kshakir: `common/target` is already [covered](https://stackoverflow.com/questions/8783093/gitignore-syntax-bin-vs-bin-vs-bin-vs-bin#8783675) by `target` in the root [`.gitignore`](https://github.com/broadinstitute/cromwell/blob/d45f00106f51f3e639cfebb149d11ff60a4a373a/.gitignore#L9)

PR #2978, opened 2017-12-04
Rerun on "failed to start due to preemption". Closes #2970

### LINKED ISSUE #2970: Add new retryable case: JES error code 2.  Message: Instance failed to start due to preemption.
There’s a new response from JES that should be retryable: 
JES error code 2.  Message: Instance failed to start due to preemption.” 

THE CATCH: Henry chatted with Google, and it sounds like JES error code 2 isn’t *always* about preemption. (But in the past few weeks we’ve had a handful of cloud workflows failing each day from this response, always with the message about preemption). So to be safe, we want to make this one retryable based on the combination of the code + the message. 

Example: 
UUID: 7da0394b-371f-4fdb-ae70-737833c4fbfa
OPERATION_ID: operations/EKmIx96ALBjh373VhLH0ui8gkYad9-AKKg9wcm9kdWN0aW9uUXVldWU

Relevant workflow metadata: 
"failures"
: [
{
"causedBy": [],
"message": "Task PairedEndSingleSampleWorkflow.CollectUnsortedReadgroupBamQualityMetrics:2:1 failed. JES error code 2.  Message: Instance failed to start due to preemption."
}
],

cjllanwarne:
See #2970 for the current error case which we now think should be treated as preemptible.

### DIFF
--- a/supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/RunStatus.scala
+++ b/supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/RunStatus.scala
@@ -51,13 +51,14 @@ object RunStatus {
               zone: Option[String],
               instanceName: Option[String]): UnsuccessfulRunStatus = {
       val jesCode: Option[Int] = errorMessage flatMap { em => Try(em.substring(0, em.indexOf(':')).toInt).toOption }
-      if (errorCode == Status.ABORTED && jesCode.contains(JesAsyncBackendJobExecutionActor.JesPreemption)) {
-        Preempted(errorCode, jesCode, errorMessage, eventList, machineType, zone, instanceName)
-      } else if (errorCode == Status.CANCELLED) {
-        Cancelled(errorCode, jesCode, errorMessage, eventList, machineType, zone, instanceName)
-      } else {
-        Failed(errorCode, jesCode, errorMessage, eventList, machineType, zone, instanceName)
+
+      val unsuccessfulStatusBuilder = errorCode match {
+        case Status.ABORTED if jesCode.contains(JesAsyncBackendJobExecutionActor.JesPreemption) => Preempted.apply _
+        case Status.UNKNOWN if errorMessage.exists(_.contains(JesAsyncBackendJobExecutionActor.FailedToStartDueToPreemptionSubstring)) => Preempted.apply _
+        case Status.CANCELLED => Cancelled.apply _
+        case _ => Failed.apply _
       }
+      unsuccessfulStatusBuilder.apply(errorCode, jesCode, errorMessage, eventList, machineType, zone, instanceName)
     }
   }
 
--- a/supportedBackends/jes/src/test/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActorSpec.scala
+++ b/supportedBackends/jes/src/test/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActorSpec.scala
@@ -196,9 +196,9 @@ class JesAsyncBackendJobExecutionActorSpec extends TestKitSuite("JesAsyncBackend
     system.actorOf(Props(new ExecuteOrRecoverActor), "ExecuteOrRecoverActor-" + UUID.randomUUID)
   }
 
-  private def runAndFail(previousPreemptions: Int, previousUnexpectedRetries: Int, preemptible: Int, errorCode: Status, innerErrorCode: Int, expectPreemptible: Boolean): BackendJobExecutionResponse = {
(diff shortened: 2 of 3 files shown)


supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/RunStatus.scala
```diff
@@ -51,12 +51,16 @@ object RunStatus {
...
-        Preempted(errorCode, jesCode, errorMessage, eventList, machineType, zone, instanceName)
-      } else if (errorCode == Status.CANCELLED) {
-        Cancelled(errorCode, jesCode, errorMessage, eventList, machineType, zone, instanceName)
-      } else {
-        Failed(errorCode, jesCode, errorMessage, eventList, machineType, zone, instanceName)
+
+      errorCode match {
+        case Status.ABORTED if jesCode.contains(JesAsyncBackendJobExecutionActor.JesPreemption) =>
+          Preempted(errorCode, jesCode, errorMessage, eventList, machineType, zone, instanceName)
+        case Status.UNKNOWN if errorMessage.exists(_.contains("failed to start due to preemption")) =>
```
mcovarr: why does the one above get a constant and not this one?

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/RunStatus.scala
```diff
@@ -51,12 +51,16 @@ object RunStatus {
...
+          Preempted(errorCode, jesCode, errorMessage, eventList, machineType, zone, instanceName)
+        case Status.CANCELLED =>
+          Cancelled(errorCode, jesCode, errorMessage, eventList, machineType, zone, instanceName)
+        case _ =>
+          Failed(errorCode, jesCode, errorMessage, eventList, machineType, zone, instanceName)
       }
```
mcovarr: Even DRYer, arguably more readable:
```scala
val unsuccessfulStatusBuilder = errorCode match {
  case Status.ABORTED if jesCode.contains(JesAsyncBackendJobExecutionActor.JesPreemption) => Preempted
  case Status.UNKNOWN if errorMessage.exists(_.contains("failed to start due to preemption")) => Preempted
  case Status.CANCELLED => Cancelled
  case _ => Failed
}
unsuccessfulStatusBuilder.apply(errorCode, jesCode, errorMessage, eventList, machineType, zone, instanceName)
```

supportedBackends/jes/src/test/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActorSpec.scala
```diff
@@ -196,9 +196,9 @@ class JesAsyncBackendJobExecutionActorSpec extends TestKitSuite("JesAsyncBackend
     system.actorOf(Props(new ExecuteOrRecoverActor), "ExecuteOrRecoverActor-" + UUID.randomUUID)
   }
 
-  private def runAndFail(previousPreemptions: Int, previousUnexpectedRetries: Int, preemptible: Int, errorCode: Status, innerErrorCode: Int, expectPreemptible: Boolean): BackendJobExecutionResponse = {
+  private def runAndFail(previousPreemptions: Int, previousUnexpectedRetries: Int, preemptible: Int, errorCode: Status, innerErrorMessage: String, expectPreemptible: Boolean): BackendJobExecutionResponse = {
 
-    val runStatus = UnsuccessfulRunStatus(errorCode, Option(s"$innerErrorCode: I seen some things man"), Seq.empty, Option("fakeMachine"), Option("fakeZone"), Option("fakeInstance"))
```
mcovarr: 😢

cjllanwarne: 🙈

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/RunStatus.scala
cjllanwarne: It's because the one above is a JES error code (ie an `Integer`) whereas this one is just a portion of an error message `String`. I can make it a constant if you prefer?

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/RunStatus.scala
cjllanwarne: Is this an acceptable compromise? I find it more obvious what's happening:
```scala
case Status.CANCELLED => Cancelled.apply _
```

Issue #2854, opened 2017-11-13
Incorrect REST deprecation warning (or incorrect swagger)


cjllanwarne:
To recreate:
- Make a submission via swagger with a WDL and an `inputs.json`
- Look to the logs. Lo:
```
The 'wdlDependencies' parameter name has been deprecated in favor of 'workflowDependencies'.
Support for 'wdlDependencies' will be removed from future versions of Cromwell.
Please switch to using 'workflowDependencies' in future submissions.
```

I didn't specify any dependencies, and swagger shouldn't be submitting them as `wdlDependencies` anyway. Maybe our REST deprecation is triggering too eagerly?

PR #2983, opened 2017-12-04
fix utterly broken deprecation warnings

mcovarr:


cjllanwarne:
Does this fix https://github.com/broadinstitute/cromwell/issues/2854 ?

engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
```diff
@@ -68,12 +70,27 @@ object PartialWorkflowSources {
...
+
+      val wdlSourceDeprecationWarning: String => String = deprecationWarning(out = WdlSourceKey, in = WorkflowSourceKey)
+      val wdlSourceWarning = wdlSource.as(WdlSourceKey) map wdlSourceDeprecationWarning
+
       val workflowSourceFinal: ErrorOr[String] = (wdlSource, workflowSource) match {
         case (Some(source), None) => source.validNel
         case (None, Some(source)) => source.validNel
-        case (Some(_), Some(_)) => "wdlSource and workflowSource can't both be supplied".invalidNel
-        case (None, None) => "workflowSource needs to be supplied".invalidNel
+        case (Some(_), Some(_)) => s"$WdlSourceKey and $WorkflowSourceKey can't both be supplied".invalidNel
```
cjllanwarne: `workflowSourceKey`, surely?

engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
```diff
@@ -68,12 +70,27 @@ object PartialWorkflowSources {
       // workflow source
       val wdlSource = getStringValue(WdlSourceKey)
       val workflowSource = getStringValue(WorkflowSourceKey)
-      val wdlSourceWarning = wdlSource.map(_ => wdlSourceDeprecationWarning)
+
+      def deprecationWarning(out: String, in: String)(actual: String): String = {
```
cjllanwarne: ToL: It'd be nice if this also included a `(..., expiration: String)` so we can be more exact, eg `"... will be removed from future versions of Cromwell as of July 2018"`

Horneth: If we do that we should also create a ticket to actually remove support for it or it'll very likely be forgotten

mcovarr: I'd consider that a PO call since we could be breaking existing installations. Paging @katevoss.

mcovarr:
@cjllanwarne yes I added the link

engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
cjllanwarne: Good point, I was thinking we'd at least know "ah yes, now is a fine time to remove it" next time we happened to look at it.

katevoss: Thanks for the page, @mcovarr. If I understand the issue, we are deprecating something and want to add a date of deprecation. After that date, we want to remove it, which would break existing installations. 
As long as we're not breaking anything for 30, could someone make the ticket and we can discuss implications and a plan of action there.

Horneth: Makes me remember that the PEM files are "deprecated" in Cromwell since April... might be time to cut the cord 😄

cjllanwarne: @katevoss see here: https://github.com/broadinstitute/cromwell/issues/2985

PR #2987, opened 2017-12-04
Parse Cwl's "Any" type 

danbills:
This does assert at least that the value is valid Json.

We may consider going to a coproduct in the future but this will get us past parsing failure in the conformance tests.


geoffjentry:
:+1:

PR #2988, opened 2017-12-04
Customizable input identifier

Horneth:
A bunch of CWL tests fail at materialization time because CWL expects the keys in the input file to be the "local name" of the input whereas WDL expects the "fully qualified name" 
e.g

cwl:
```
{
  "my_input": "hello"
}
``` 

wdl:
```
{
  "my_workflow.my_input": "hello"
}
```

This allows each language to define how an `ExternalGraphInputNode` should be mapped to a `String` to look for in the input file.

### DIFF
--- a/cwl/src/main/scala/cwl/Workflow.scala
+++ b/cwl/src/main/scala/cwl/Workflow.scala
@@ -65,11 +65,11 @@ case class Workflow private(
 
         // TODO: Eurgh! But until we have something better ...
         val womValue = womType.coerceRawValue(inputParameter.default.get).get
-        OptionalGraphInputNodeWithDefault(WomIdentifier(parsedInputId), womType, ValueAsAnExpression(womValue))
+        OptionalGraphInputNodeWithDefault(WomIdentifier(parsedInputId, inputParameter.id), womType, ValueAsAnExpression(womValue), parsedInputId)
       case input =>
         val parsedInputId = FileAndId(input.id).id
 
-        RequiredGraphInputNode(WomIdentifier(parsedInputId), wdlTypeForInputParameter(input).get)
+        RequiredGraphInputNode(WomIdentifier(parsedInputId, input.id), wdlTypeForInputParameter(input).get, parsedInputId)
     }.toSet
 
     val workflowInputs: Map[String, GraphNodeOutputPort] =
--- a/wom/src/main/scala/wom/graph/CallNode.scala
+++ b/wom/src/main/scala/wom/graph/CallNode.scala
@@ -51,6 +51,7 @@ object TaskCall {
     
     // Creates an identifier for an input or an output
     // The localName is the name of the input or output
+    
     // The FQN combines the name of the task to the name of the input or output
     def identifier(name: LocalName) = WomIdentifier(name, taskDefinitionLocalName.combineToFullyQualifiedName(name))
 
@@ -64,9 +65,9 @@ object TaskCall {
     val inputDefinitionFold = taskDefinition.inputs.foldMap({ inputDef =>
     {
       val newNode = inputDef match {
-        case RequiredInputDefinition(name, womType) => RequiredGraphInputNode(identifier(name), womType)
-        case InputDefinitionWithDefault(name, womType, default) => OptionalGraphInputNodeWithDefault(identifier(name), womType, default)
-        case OptionalInputDefinition(name, womType) => OptionalGraphInputNode(identifier(name), womType)
+        case RequiredInputDefinition(name, womType) => RequiredGraphInputNode(identifier(name), womType, name.value)
+        case InputDefinitionWithDefault(name, womType, default) => OptionalGraphInputNodeWithDefault(identifier(name), womType, default, name.value)
+        case OptionalInputDefinition(name, womType) => OptionalGraphInputNode(identifier(name), womType, name.value)
       }
 
       InputDefinitionFold(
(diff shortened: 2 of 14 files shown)


wom/src/main/scala-2.11/wom/executable/ExecutableValidation.scala
```diff
@@ -5,7 +5,7 @@ import common.Checked
 import common.validation.Checked._
 import common.validation.ErrorOr.ErrorOr
 import wom.callable.ExecutableCallable
-import wom.executable.Executable.{DelayedCoercionFunction, InputParsingFunction, ResolvedExecutableInputs}
+import wom.executable.Executable.{DelayedCoercionFunction, EGINInputIdentifier, InputParsingFunction, ResolvedExecutableInputs}
```
kshakir: ToL: should this (and others) be `egin` vs. `eGIN`?
- Example 1: [`ogin`](https://github.com/broadinstitute/cromwell/blob/1b1a56372659b9cb7a168bb1fa2a2296103e1256/wdl/src/main/scala/wdl/WdlCall.scala#L155)
- Example 2: [`wom`](https://github.com/broadinstitute/cromwell/blob/1b1a56372659b9cb7a168bb1fa2a2296103e1256/wom/src/main/scala/wom/values/WomValue.scala#L10)
- Example 3: [`ejea`](https://github.com/broadinstitute/cromwell/blob/1b1a56372659b9cb7a168bb1fa2a2296103e1256/engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/ejea/EjeaRunningJobSpec.scala#L14)
- etc.

cjllanwarne:
TOL: To me this feels like it’d be way neater if the EGIN had a field or def called “inputFileName”, to encapsulate all this into the egin itself rather than having to add it later externally?

cjllanwarne:
FWIW in my ideal world we’d have pluggable languages which should only need to define one function like “readWorkflowIntoWom(content: String, l: Set[ImportResolver]): WomExecutable” and everything else would be included/encapsulated in that result

Horneth:
> TOL: To me this feels like it’d be way neater if the EGIN had a field or def called inputFileName nameInInputSet, to encapsulate all this into the egin itself rather than having to add it later externally?

Good point, I was just reticent to the idea of jamming yet another attribute injected by the language that will only ever be used once during the lifetime of the workflow but I agree it's still neater.

> FWIW in my ideal world we’d have pluggable languages which should only need to define one function like “readWorkflowIntoWom(content: String, l: Set[ImportResolver]): WomExecutable” and everything else would be included/encapsulated in that result

Yes but there would be some non-DRYness by having each language implement entirely how they ingest inputs (most of the logic is the same), plus having it in WOM guarantees that all EGIN are handled the same way w.r.t coercion, validation etc.. It could use some refactoring though I agree

mcovarr:
We've had the pluggability discussion before and I believe we decided it would happen when N > 2, which will likely be the first time we need to support distinctly different versions of either CWL or WDL.

cjllanwarne:
@mcovarr I agree (although that day might be sooner than you think, re WDLd3)... read my pluggable comment only as "this is background info on why I don't want another function if we can avoid it"

Horneth:
@mcovarr @kshakir a quick second pass would be appreciated :) I followed (one of :p) @cjllanwarne suggestions and added an attribute to EGIN instead of passing a function pointer.
I also fixed the FQNs for CWL EGINs which makes the pre-existing `CwlInputValidationSpec` actually test that this works now.

wom/src/main/scala/wom/graph/CallNode.scala
```diff
@@ -50,7 +50,8 @@ object TaskCall {
     val taskDefinitionLocalName = LocalName(taskDefinition.name)
     
     // Creates an identifier for an input or an output
-    // The localName is the name of the input or output
+    // The localName is the name of the input or outputB8z2@kft/16!
```
mcovarr: um what

wom/src/main/scala/wom/graph/CallNode.scala
```diff
@@ -51,6 +51,7 @@ object TaskCall {
     
     // Creates an identifier for an input or an output
     // The localName is the name of the input or output
+
```
kshakir: While you're here, can you covert this comment block from `//` to `/* */`? I would say `/** */`, but I don't think local-def may use `/** */`.

cwl/src/main/scala/cwl/Workflow.scala
```diff
@@ -65,11 +65,11 @@ case class Workflow private(
 
         // TODO: Eurgh! But until we have something better ...
         val womValue = womType.coerceRawValue(inputParameter.default.get).get
-        OptionalGraphInputNodeWithDefault(WomIdentifier(parsedInputId), womType, ValueAsAnExpression(womValue))
+        OptionalGraphInputNodeWithDefault(WomIdentifier(parsedInputId, inputParameter.id), womType, ValueAsAnExpression(womValue), parsedInputId)
```
cjllanwarne: ToL: I'm not sure what it referred to before, but does this satisfy the `TODO: Eurgh` comment above?

wom/src/main/scala/wom/graph/GraphInputNode.scala
```diff
@@ -38,18 +38,26 @@ sealed trait ExternalGraphInputNode extends GraphInputNode {
     */
   
   override lazy val singleOutputPort: GraphNodeOutputPort = GraphNodeOutputPort(identifier, womType, this)
+
+  /**
+    * Key that should be looked for in the input set to satisfy this GIN
```
cjllanwarne: `GIN` => `EGIN`

cjllanwarne:
Rebase to unredden TES tests

cwl/src/main/scala/cwl/Workflow.scala
Horneth: I think the `Eurgh` is for the `.get.get` just below but I could be wrong

PR #2954, opened 2017-11-30
Reinstate runtime attributes

danbills:
As `WomValue` can now be represented in `WomExpression` using `ValueAsExpression`, this validation  was reinstated.

PR #2989, opened 2017-12-05
Re-added kind-projector compiler plugin to backend.

kshakir:
Custom scaladoc settings must be appended (++=) instead of overwriting (:=).
Fixed unmoored doc warnings/errors.


backend/src/main/scala/cromwell/backend/BackendWorkflowInitializationActor.scala
```diff
@@ -79,7 +79,7 @@ object BackendWorkflowInitializationActor {
       //This map append will overwrite default key/values with runtime settings upon key collisions
       val lookups = defaultRuntimeAttributes.mapValues(_.asWomExpression) ++ runtimeAttributes
 
-      runtimeAttributeValidators.toList.traverse[WishKindProjectorWorked, Unit]{
```
mcovarr: 😢

project/Dependencies.scala
```diff
@@ -47,6 +48,7 @@ object Dependencies {
   val mouseV = "0.10-MF"
   val mysqlV = "5.1.42"
   val nettyHandlerV = "4.0.51.Final"
+  val paradiseV = "2.1.0"
```
cjllanwarne: ToL `paradise 2.1` sounds like it's straight out of a horror/dystopia movie

kshakir:
> Bonus points if you can ELI5?

Perhaps after I finish the FP book club??

At the moment, this is just fixing an sbt bug that was blocking [this commit](https://github.com/broadinstitute/cromwell/pull/2954/commits/8ce921581169f8466b67a6f03ceee8b498e4b483).

Issue #2612, opened 2017-09-11
Run I/O methods from the IoFunctionSet through the I/O Actor


Horneth:
This will give us retries, batching and throttling also for expression evaluation

cjllanwarne:
@katevoss heads up that this was reported and requested in the forums here: https://gatkforums.broadinstitute.org/wdl/discussion/10853/consistent-503-error-for-engine-functions#latest

katevoss:
Is it related to (or the same as) #2576 ?

cjllanwarne:
@katevoss seems to be, yeah. This is a specific fix to the problem in #2576 which also comes with extra benefits like throttling and batching to make it generally much more reliable and scalable.

Issue #2576, opened 2017-08-25
503 StorageException failed a workflow, possibly without retry


kshakir:
A 503 StorageException seems to have failed one of the centaur JES jobs, and hence the workflow. Via the [cromwell.log](https://console.cloud.google.com/storage/browser/cloud-cromwell-dev/cromwell_execution/travis/centaur_workflow/0310fa51-e985-4c54-8cdb-5058155f452e/call-centaur/cromwell_root/logs/).

```java
2017-08-25 05:43:25,399 cromwell-system-akka.dispatchers.engine-dispatcher-51 ERROR - WorkflowManagerActor Workflow dabddbe7-a385-4df4-be97-c1ef7b884823 failed (during ExecutingWorkflowState): Could not evaluate composeEngineFunctions.y = read_int(stderr()) + x + read_string(blah)
java.lang.RuntimeException: Could not evaluate composeEngineFunctions.y = read_int(stderr()) + x + read_string(blah)
	at wdl4s.wdl.WdlTask$$anonfun$4.applyOrElse(WdlTask.scala:190)
	at wdl4s.wdl.WdlTask$$anonfun$4.applyOrElse(WdlTask.scala:189)
	at scala.runtime.AbstractPartialFunction.apply(AbstractPartialFunction.scala:34)
	at scala.util.Failure.recoverWith(Try.scala:232)
	at wdl4s.wdl.WdlTask.$anonfun$evaluateOutputs$2(WdlTask.scala:189)
	at scala.collection.TraversableOnce.$anonfun$foldLeft$1(TraversableOnce.scala:157)
	at scala.collection.TraversableOnce.$anonfun$foldLeft$1$adapted(TraversableOnce.scala:157)
	at scala.collection.Iterator.foreach(Iterator.scala:929)
	at scala.collection.Iterator.foreach$(Iterator.scala:929)
	at scala.collection.AbstractIterator.foreach(Iterator.scala:1417)
	at scala.collection.IterableLike.foreach(IterableLike.scala:71)
	at scala.collection.IterableLike.foreach$(IterableLike.scala:70)
	at scala.collection.AbstractIterable.foreach(Iterable.scala:54)
	at scala.collection.TraversableOnce.foldLeft(TraversableOnce.scala:157)
	at scala.collection.TraversableOnce.foldLeft$(TraversableOnce.scala:155)
	at scala.collection.AbstractTraversable.foldLeft(Traversable.scala:104)
[...]

PR #2993, opened 2017-12-05
Imperfect retries for GCS requests

cjllanwarne:
- Hopefully addresses #2576 
- But not in the more complete way suggested in #2612

nb only applies defaults if we create a `GcsPathBuilder` via a `GcsPathBuilderFactory`

### DIFF
--- a/filesystems/gcs/src/main/scala/cromwell/filesystems/gcs/GcsPathBuilderFactory.scala
+++ b/filesystems/gcs/src/main/scala/cromwell/filesystems/gcs/GcsPathBuilderFactory.scala
@@ -2,23 +2,24 @@ package cromwell.filesystems.gcs
 
 import akka.actor.ActorSystem
 import com.google.api.gax.retrying.RetrySettings
-import com.google.auth.Credentials
 import com.google.cloud.storage.contrib.nio.CloudStorageConfiguration
+import com.google.auth.Credentials
 import cromwell.cloudsupport.gcp.auth.GoogleAuthMode
 import cromwell.cloudsupport.gcp.gcs.GcsStorage
 import cromwell.core.WorkflowOptions
 import cromwell.core.path.PathBuilderFactory
+import org.threeten.bp.Duration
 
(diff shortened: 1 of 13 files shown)


filesystems/gcs/src/main/scala/cromwell/filesystems/gcs/GcsPathBuilderFactory.scala
```diff
@@ -2,30 +2,56 @@ package cromwell.filesystems.gcs
...
+            applicationName: String,
+            retrySettings: Option[RetrySettings]): GcsPathBuilderFactory = {
+
+    val actualRetrySettings: RetrySettings = retrySettings.getOrElse(defaultRetrySettings)
+    val actualCloudStorage = defaultCloudStorageConfiguration
+
+    new GcsPathBuilderFactory(authMode, applicationName, actualRetrySettings, actualCloudStorage)
+  }
+
+  lazy val defaultRetrySettings: RetrySettings = RetrySettings.newBuilder()
```
mcovarr: `DefaultRetrySettings` to make clear it's a constant

filesystems/gcs/src/main/scala/cromwell/filesystems/gcs/GcsPathBuilderFactory.scala
```diff
@@ -2,30 +2,56 @@ package cromwell.filesystems.gcs
...
+    .setInitialRetryDelay(Duration.ofMillis(100))
+    .setRetryDelayMultiplier(1.1)
+    .setMaxRetryDelay(Duration.ofSeconds(1))
+    .setJittered(true)
+    .setInitialRpcTimeout(Duration.ofMillis(100))
+    .setRpcTimeoutMultiplier(1.1)
+    .setMaxRpcTimeout(Duration.ofSeconds(5))
+    .build()
+
+  lazy val defaultCloudStorageConfiguration = GcsStorage.DefaultCloudStorageConfiguration
```
mcovarr: same here

PR #2994, opened 2017-12-05
Run Cromwell in CWL test dir

Horneth:
With https://github.com/broadinstitute/cromwell/pull/2988 it should enable a bunch of new CWL conformance tests to run on travis. They will still need the outputs to look right to fully "pass"

### DIFF
--- a/src/bin/travis/testCentaurCwlConformance.sh
+++ b/src/bin/travis/testCentaurCwlConformance.sh
@@ -10,6 +10,7 @@ CROMWELL_JAR=$(find "$(pwd)/target/scala-2.12" -name "cromwell-*.jar")
 CENTAUR_CWL_RUNNER="$(pwd)/centaurCwlRunner/src/bin/centaur-cwl-runner.bash"
 
 git clone --depth 1 https://github.com/common-workflow-language/common-workflow-language.git
+CWL_TEST_DIR=$(pwd)/common-workflow-language/v1.0/v1.0
 
 shutdownCromwell() {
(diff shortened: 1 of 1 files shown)


src/bin/travis/testCentaurCwlConformance.sh
```diff
@@ -19,8 +20,12 @@ shutdownCromwell() {
 
 trap "shutdownCromwell" EXIT
 
+# Start the Cromwell server in the directory containing input files so it can access them via their relative path
+CURRENT_DIR=$(pwd)
+cd CWL_TEST_DIR
 java -jar ${CROMWELL_JAR} server &
 CROMWELL_PID=$$
+cd CURRENT_DIR
```
mcovarr: you might subshell this instead

Horneth: If I subshell it will `CROMWELL_PID` still be defined outside the subshell ?

mcovarr: ugh nm. you can do disgusting things with globals but that's totally not worth it.

cjllanwarne:
I was going to say "rebase on `develop` to unredden TES". Unfortunately the CWL conformance tests are also red (and since we ignore the results right now, that means they are failing to run at all, I believe)

kshakir:
Aha, [a week ago](https://github.com/common-workflow-language/common-workflow-language/commit/915d49f8fb912774cbf5cbe6ba910a430ef69941) 100% of tests were moved under the `v1.0` directory. So this "fix" should now work!

PR #3013, opened 2017-12-06
Manipulate outputs to make File

danbills:
CWL expects `File` output to include lots of additional metadata.

CWL also expects output keys to be the "local name" whereas Cromwell outputs fully qualified names.

This PR addresses these issues.


build.sbt
```diff
@@ -50,6 +50,7 @@ lazy val dockerHashing = project
 
 lazy val cromwellApiClient = project
   .withLibrarySettings("cromwell-api-client", cromwellApiClientDependencies)
+  .dependsOn(engine)
```
mcovarr: that doesn't seem right

centaurCwlRunner/src/main/scala/centaur/cwl/Outputs.scala
```diff
@@ -0,0 +1,104 @@
...
+import cromwell.api.model.SubmittedWorkflow
+import cwl.{CwlDecoder, CwlType, MyriadOutputType, File => CwlFile}
+import io.circe.Json
+import io.circe.generic.auto._
+import io.circe.literal._
+import io.circe.syntax._
+import shapeless.{Inl, Poly1}
+import spray.json.{JsNumber, JsObject, JsString, JsValue}
+
+import scalaz.syntax.std.map._
```
mcovarr: why scalaz and not 🐱

centaurCwlRunner/src/main/scala/centaur/cwl/CentaurCwlRunner.scala
```diff
@@ -130,11 +130,11 @@ object CentaurCwlRunner {
               println(s"Unexpected failure.")
               ExitCode.Failure
             case Succeeded =>
-              val outputs = CentaurCromwellClient.outputs(submittedWorkflow).get.outputs
+
               if (!args.quiet) {
                 println("Result:")
               }
-              println(outputs)
+              println(centaur.cwl.Outputs.handleOutput(submittedWorkflow))
```
mcovarr: why not import this?

centaurCwlRunner/src/main/scala/centaur/cwl/Outputs.scala
```diff
@@ -0,0 +1,106 @@
...
+}
+
+object CwlOutputsFold extends Poly1 {
+  import cwl._
+
+  implicit def wf: Case.Aux[cwl.Workflow, Map[String, MyriadOutputType]] = at[cwl.Workflow] {
+    _.outputs.map(output => output.id -> output.`type`.get).toMap
+  }
+
+  implicit def clt: Case.Aux[cwl.CommandLineTool, Map[String, MyriadOutputType]] = at[cwl.CommandLineTool] {_.outputs.map(output => output.id -> output.`type`.get).toMap}
```
mcovarr: this might be better across multiple lines

cwl/src/main/scala/cwl/Decoder.scala
```diff
@@ -54,6 +56,18 @@ object CwlDecoder {
...
 
+  def decodeTopLevelCwl(fileName: BFile): Parse[Cwl] =
+    for {
+      jsonString <- preprocess(fileName)
+      unmodifiedCwl <- parseJson(jsonString)
+    } yield unmodifiedCwl
+
+  def decodeTopLevelCwl(cwl: String): Parse[Cwl] =
+    for {
+     file <-  EitherT{ IO{ better.files.File.newTemporaryFile().write(cwl).asRight }}
```
mcovarr: import

project/Dependencies.scala
```diff
@@ -63,6 +63,7 @@ object Dependencies {
   val scalacticV = "3.0.1"
   val scalameterV = "0.8.2"
   val scalatestV = "3.0.2"
+  val scalazV = "7.2.17"
```
mcovarr: 😲

centaurCwlRunner/src/main/scala/centaur/cwl/Outputs.scala
```diff
@@ -0,0 +1,106 @@
...
+
+        CwlFile(
+          location = Option(fileName.name),
+          checksum = Option("sha1$" + hex),
+          size = Option(fileName.size)
+        ).asJson
+      case (JsNumber(metadata), Inl(CwlType.Long)) => metadata.longValue.asJson
+      case (JsNumber(metadata), Inl(CwlType.Float)) => metadata.floatValue().asJson
+      case (JsNumber(metadata), Inl(CwlType.Double)) => metadata.doubleValue().asJson
+      case (JsNumber(metadata), Inl(CwlType.Int)) => metadata.intValue().asJson
```
mcovarr: why do these have parens

centaurCwlRunner/src/main/scala/centaur/cwl/Outputs.scala
```diff
@@ -0,0 +1,106 @@
...
+object Outputs {
+
+  //When the string returned is not valid JSON, it is effectively an exception as CWL runner expects JSON to be returned
+  def handleOutput(submittedWorkflow: SubmittedWorkflow): String = {
+    val metadata: Map[String, JsValue] = CentaurCromwellClient.metadata(submittedWorkflow).get.value
+
+    //Sorry for all the nesting, but spray json JsValue doesn't have optional access methods like Argonaut/circe,
+    //thus no for comprehensions for us :(
+    metadata.get("submittedFiles.workflow") match {
+      case Some(JsString(some)) =>
```
cjllanwarne: ToL: `some` to `stringValue`? Or even better what the value represents. IMO the `Some`ness is implied

centaurCwlRunner/src/main/scala/centaur/cwl/Outputs.scala
```diff
@@ -0,0 +1,106 @@
...
+                val mungeTypeMap = typeMap.mapKeys(stripTypeMapKey)
+
+                val mungeOutputMap = map.mapKeys(stripOutputKey)
+
+                mungeOutputMap.
+                  //This lets us operate on the values of the output values and types for a particular output key
+                  intersectWith(mungeTypeMap)(resolveOutput).
+                  //converting the whole response to Json using Circe's auto-encoder derivation
+                  asJson.
+                  //drop null values so that we don't print when Option == None
```
cjllanwarne: I hope these are not `WomOptional`s that are being dropped?

centaurCwlRunner/src/main/scala/centaur/cwl/Outputs.scala
cjllanwarne: because they're 2-ples?

project/Dependencies.scala
cjllanwarne: 😿

centaurCwlRunner/src/main/scala/centaur/cwl/Outputs.scala
```diff
@@ -0,0 +1,106 @@
...
+                  pretty(io.circe.Printer.spaces2.copy(dropNullValues = true))
+              case other => s"it seems cromwell is not returning outputs as a Jsobject but is instead a $other"
+            }
+          case Right(Left(error)) => s"couldn't parse $some: $error"
+          case Left(error) => s"couldn't parse $some: $error"
```
Horneth: What's the difference between a `Right(Left(error))` and a `Left(error)` ? Do they really both mean parsing failed or do they have a different cause ?

centaurCwlRunner/src/main/scala/centaur/cwl/Outputs.scala
kshakir: On the right side, ex: `metadata.intValue().asJson` should be `metadata.intValue.asJson`

centaurCwlRunner/src/main/scala/centaur/cwl/Outputs.scala
danbills: Map.intersectWith is :fire: : https://github.com/scalaz/scalaz/blob/series/7.3.x/core/src/main/scala/scalaz/syntax/std/MapOps.scala#L12

Not in cats :frowning:

centaurCwlRunner/src/main/scala/centaur/cwl/Outputs.scala
danbills: tuples

centaurCwlRunner/src/main/scala/centaur/cwl/Outputs.scala
danbills: the outer is a result of running `IO.attempt` which catches any exceptions.

The inner is a result of a parsing error, i.e. non-exception. 

I see I left some copypasta here and should fix the error message.

build.sbt
danbills: This is an artifact of me trying to use Cromwell's Metadata types to parse the return types.  

I do need a dependency on `cwl` for it's `File` class.

centaurCwlRunner/src/main/scala/centaur/cwl/CentaurCwlRunner.scala
danbills: I'll import.

centaurCwlRunner/src/main/scala/centaur/cwl/Outputs.scala
```diff
@@ -0,0 +1,106 @@
...
+
+    //Sorry for all the nesting, but spray json JsValue doesn't have optional access methods like Argonaut/circe,
+    //thus no for comprehensions for us :(
+    metadata.get("submittedFiles.workflow") match {
+      case Some(JsString(some)) =>
+
+        val cwl = CwlDecoder.decodeTopLevelCwl(some)
+
+        cwl.value.attempt.unsafeRunSync() match {
+          case Right(Right(cwl)) =>
```
cjllanwarne: Suspicious shadowing by a variable

centaurCwlRunner/src/main/scala/centaur/cwl/Outputs.scala
```diff
@@ -0,0 +1,106 @@
...
+        val fileName = better.files.File(metadata)
+
+        val file = fs2.io.file.readAll[IO](fileName.path, 65536)
+        val hash = fs2.hash.sha1[IO]
+
+        val bytes = (file through hash).runLog.unsafeRunSync
+
+        //hexify the bytes into a string
+        val hex = bytes.map {
+          b => String.format("%02X", new Integer(b & 0xff))
```
cjllanwarne: IntelliJ has issues with this line:
screen shot 2017-12-06 at 2 50 03 pm

centaurCwlRunner/src/main/scala/centaur/cwl/Outputs.scala
danbills: I'm using shadowing on purpose to get around the naming convention of `xContext` and then `x` for the value we actually care about.

In this extreme case `cwl` is an `IO[Either[..]]` so in order to avoid `cwlIOEither` grossness I used shadowing.

centaurCwlRunner/src/main/scala/centaur/cwl/Outputs.scala
danbills: I'm afraid IntelliJ is mistaken, it works as advertised.

centaurCwlRunner/src/main/scala/centaur/cwl/Outputs.scala
danbills: ah I will clean up the right side, thanks @kshakir

centaurCwlRunner/src/main/scala/centaur/cwl/Outputs.scala
danbills: Hm, now my test is failing on the hex value.  Perhaps *I'm* mistaken.

centaurCwlRunner/src/main/scala/centaur/cwl/Outputs.scala
danbills: These are scala `Option` where they == `None`.  Otherwise they print out as `format=null`.  Not sure if it fails the CWL conformance tester but it is unsightly to be sure.

build.sbt
```diff
@@ -50,6 +50,7 @@ lazy val dockerHashing = project
 
 lazy val cromwellApiClient = project
   .withLibrarySettings("cromwell-api-client", cromwellApiClientDependencies)
+  .dependsOn(cwl)
```
mcovarr: while certainly an improvement over depending on `engine` I'm still not sure conceptually why the API client should need this? can this be just the `centaurCwlRunner` that needs `cwl`?

danbills: :man_facepalming: .  I didn't realize this was in the wrong place!

PR #3016, opened 2017-12-06
Cromwell should be able to take a hint.

mcovarr:
Allows parsing hints and requirements. ~~The requirement parsing is more liberal than it should be, I'm still investigating that.~~ No I'm just a dummy and was feeding my Cromwell the wrong copy of the `three_step.cwl` file.

### DIFF
--- a/cwl/src/main/scala/cwl/Workflow.scala
+++ b/cwl/src/main/scala/cwl/Workflow.scala
@@ -25,7 +25,9 @@ case class Workflow private(
                      id: String,
                      inputs: Array[InputParameter],
                      outputs: Array[WorkflowOutputParameter],
-                     steps: Array[WorkflowStep]) {
+                     steps: Array[WorkflowStep],
+                     requirements: Option[Array[Requirement]],
+                     hints: Option[Array[CwlAny]]) {
   def womExecutable(inputFile: Option[String] = None): Checked[Executable] = {
     CwlExecutableValidation.buildWomExecutable(womDefinition, inputFile)
   }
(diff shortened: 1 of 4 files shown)


cwl/src/test/resources/three_step.cwl
```diff
@@ -1,6 +1,15 @@
 cwlVersion: v1.0
 class: Workflow
 id: three_step
+hints:
+  # hints can be anything (CwlAny)
+  - fee: fi
+    fo: fum
+  - eenie: meenie
+    minie: mo
+  - 42
```
cjllanwarne: ToL: I don't know whether the `three_step.cwl` is the best place to be testing the limits of `CwlAny`'s parsing

mcovarr: It's not, this is a quick hack. 🙂

mcovarr:
Lovely, this parses and runs when actually submitted to Cromwell but the unit tests won't even parse it 😡

mcovarr:
Ok talked with Dan and there are a couple of problems here: 
We have two copies of three_step.cwl, one of which is used by unit tests and the other is used by Centaur. The unit tests currently don't SALAD input CWL so their input needs to be formatted just so. Real Cromwell does SALAD input CWL so it's a lot more flexible. Dan is working on a separate PR so the unit tests run SALAD.

cwl/src/main/scala/cwl/Workflow.scala
```diff
@@ -25,7 +25,9 @@ case class Workflow private(
                      id: String,
                      inputs: Array[InputParameter],
                      outputs: Array[WorkflowOutputParameter],
-                     steps: Array[WorkflowStep]) {
+                     steps: Array[WorkflowStep],
+                     requirements: Option[Array[Requirement]],
+                     hints: Option[Array[CwlAny]]) {
```
cjllanwarne: These must also be added to `ExpressionTool`

mcovarr: Strangely enough, they are already there (I did not add them). 😮

cjllanwarne: Huh, I found my CNil was due to an errant `ExpressionTool`, I guess it wasn't the requirement that it wasn't liking after all 😨

mcovarr: Unit test or Centaur? Unit tests currently are not SALADing so you have to format requirements just so.

PR #3022, opened 2017-12-06
Improve Error messaging surrounding Cwl parsing

danbills:
this will spit out more informative errors when a user encounters a parsing error:

image

Unfortunately this has to be done by hand as the `Coproduct` parser in circe only returns the `CNil` by virtue of its automaticness.

As such we will continue to hit `CNil` in deeper parts of the code(as seen in this example!) unless we write `Decoder`s by hand for all of our coproducts or figure out a way to implement error accumulation in Coproduct decoder derivation and submit a patch to Circe. For reference the offending code is [here](https://github.com/circe/circe/blob/master/modules/shapes/src/main/scala/io/circe/shapes/CoproductInstances.scala#L18)

### DIFF
--- a/cwl/src/main/scala/cwl/CwlCodecs.scala
+++ b/cwl/src/main/scala/cwl/CwlCodecs.scala
@@ -7,16 +7,23 @@ import io.circe.generic.auto._
 import eu.timepit.refined.string._
 import io.circe.refined._
 import io.circe.literal._
+import shapeless.Coproduct
+import cats.syntax.either._
 
 object CwlCodecs {
   import Implicits._
 
-  //According to automatic derivation, these instances should not be required.  But
-  //removing these breaks decodeCwl, so...
-  implicit val wfD = implicitly[Decoder[Workflow]]
-  implicit val cltD = implicitly[Decoder[CommandLineTool]]
-
-  def decodeCwl(in: String): Either[Error, Cwl] = decode[Cwl](in)
+  def decodeCwl(in: String): Either[Error, Cwl] = {
+    //try to parse both and combine errors if they fail
+    (decode[Workflow](in), decode[CommandLineTool](in)) match {
+      case (Right(wf), _) => Coproduct[Cwl](wf).asRight
+      case (_, Right(clt)) => Coproduct[Cwl](clt).asRight
+      case (Left(wfError), Left(cltError)) =>
+        //This is not really suppressed but there is no other way to compose errors at the Exception level API AFAIK
+        wfError.addSuppressed(cltError)
+        wfError.asLeft
+    }
+  }
 
   def encodeCwlCommandLineTool(commandLineTool: CommandLineTool): Json = {
     import io.circe.syntax._
--- a/cwl/src/main/scala/cwl/CwlCodecs.scala
+++ b/cwl/src/main/scala/cwl/CwlCodecs.scala
@@ -16,13 +16,12 @@ import cats.syntax.show._
 object CwlCodecs {
   import Implicits._
 
-  def decodeCwl(in: String): Either[NonEmptyList[String], Cwl] = {
+  def decodeCwl(cwlWorkflow: String): Either[NonEmptyList[String], Cwl] = {
     //try to parse both and combine errors if they fail
-    (decode[Workflow](in), decode[CommandLineTool](in)) match {
+    (decode[Workflow](cwlWorkflow), decode[CommandLineTool](cwlWorkflow)) match {
       case (Right(wf), _) => Coproduct[Cwl](wf).asRight
       case (_, Right(clt)) => Coproduct[Cwl](clt).asRight
       case (Left(wfError), Left(cltError)) =>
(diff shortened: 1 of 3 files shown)


cwl/src/main/scala/cwl/CwlCodecs.scala
```diff
@@ -1,22 +1,34 @@
...
-  implicit val cltD = implicitly[Decoder[CommandLineTool]]
-
-  def decodeCwl(in: String): Either[Error, Cwl] = decode[Cwl](in)
+  def decodeCwl(in: String): Either[NonEmptyList[String], Cwl] = {
+    //try to parse both and combine errors if they fail
+    (decode[Workflow](in), decode[CommandLineTool](in)) match {
+      case (Right(wf), _) => Coproduct[Cwl](wf).asRight
+      case (_, Right(clt)) => Coproduct[Cwl](clt).asRight
+      case (Left(wfError), Left(cltError)) =>
+        //This is not really suppressed but there is no other way to compose errors at the Exception level API AFAIK
```
danbills: this is an old comment and will be removed

cwl/src/main/scala/cwl/CwlCodecs.scala
```diff
@@ -1,22 +1,34 @@
...
 
-  //According to automatic derivation, these instances should not be required.  But
-  //removing these breaks decodeCwl, so...
-  implicit val wfD = implicitly[Decoder[Workflow]]
-  implicit val cltD = implicitly[Decoder[CommandLineTool]]
-
-  def decodeCwl(in: String): Either[Error, Cwl] = decode[Cwl](in)
+  def decodeCwl(in: String): Either[NonEmptyList[String], Cwl] = {
+    //try to parse both and combine errors if they fail
+    (decode[Workflow](in), decode[CommandLineTool](in)) match {
```
mcovarr: can you `lazy` the `CommandLineTool` version to save a parse in the likely event you get a `Workflow`?

cwl/src/main/scala/cwl/CwlCodecs.scala
```diff
@@ -1,22 +1,34 @@
...
 object CwlCodecs {
   import Implicits._
 
-  //According to automatic derivation, these instances should not be required.  But
-  //removing these breaks decodeCwl, so...
-  implicit val wfD = implicitly[Decoder[Workflow]]
-  implicit val cltD = implicitly[Decoder[CommandLineTool]]
-
-  def decodeCwl(in: String): Either[Error, Cwl] = decode[Cwl](in)
+  def decodeCwl(in: String): Either[NonEmptyList[String], Cwl] = {
```
danbills: use better variable name than `in`

cwl/src/main/scala/cwl/CwlCodecs.scala
danbills: If there is a way I'm not aware of how to do that.  I could do one-then-the-other but I lose pattern matching niceness.

Plus I don't think it's terribly wasteful to just try both. :man_shrugging:

cwl/src/main/scala/cwl/CwlCodecs.scala
```diff
@@ -1,22 +1,33 @@
...
-  //According to automatic derivation, these instances should not be required.  But
-  //removing these breaks decodeCwl, so...
-  implicit val wfD = implicitly[Decoder[Workflow]]
-  implicit val cltD = implicitly[Decoder[CommandLineTool]]
-
-  def decodeCwl(in: String): Either[Error, Cwl] = decode[Cwl](in)
+  def decodeCwl(cwlWorkflow: String): Either[NonEmptyList[String], Cwl] = {
+    //try to parse both and combine errors if they fail
+    (decode[Workflow](cwlWorkflow), decode[CommandLineTool](cwlWorkflow)) match {
+      case (Right(wf), _) => Coproduct[Cwl](wf).asRight
```
mcovarr: I was essentially suggesting this: 
```
    lazy val commandLineToolDecode = decode[CommandLineTool](cwlWorkflow) 
    (decode[Workflow](cwlWorkflow), commandLineToolDecode) match {
      case (Right(wf), _) => Coproduct[Cwl](wf).asRight
```
I *think* this works out to not evaluating the `CommandLineTool` decode if you get a `Workflow`.

mcovarr: that could be instrumented with a 

```
lazy val commandLineToolDecode = {
  println("BAHH the CLT decode is being evaluated!!!")
  decode[CommandLineTool](cwlWorkflow)
}
```

danbills: It seems the tupling evaluates each argument.

```bash
val i = 2
lazy val j = {
  println("ahh")
  3
}

(i,j) match {
  case (2,_) => println("good")
  case _ => println("bad")
}
danb@GPBD5-C11:~/cromwell$ scala r.scala 
ahh
good
```

mcovarr: 😢

PR #3029, opened 2017-12-07
Fixup coercion into the WomNonEmptyArrayType

### LINKED ISSUE #3028: Non-empty array literals
eg:
```wdl
workflow foo {
  Array[Int]+ array_literal = [1, 2, 3]
]
```

Also make sure we can box these into optionals:
```wdl
workflow foo {
  Array[Int]+? array_literal = [1, 2, 3]
}
```

cjllanwarne:
Fixed #3028


engine/src/test/scala/cromwell/engine/workflow/lifecycle/MaterializeWorkflowDescriptorActorSpec.scala
```diff
@@ -241,39 +241,6 @@ class MaterializeWorkflowDescriptorActorSpec extends CromwellTestKitWordSpec wit
       system.stop(materializeWfActor)
     }
 
-
-    "reject a taskless WDL source" ignore {
```
cjllanwarne: Not entirely related to the rest of the PR, but this test case is just invalid now

PR #3030, opened 2017-12-07
Run individual conformance tests.

kshakir:
Check that results are as expected.


kshakir:
Tests are now green! Had to double the amount of changes from of the original commit to fix some logging issues.

PR #3040, opened 2017-12-08
DockerRequirement [PREGULL]

mcovarr:
This does appear to work for 3 step but I'd like to add metadata assertions to the existing Centaur test assuming the PR doesn't get gulled out of existence.

TODO

- [x] Set parent workflow into subworkflows
- [x] Fix input path localization
- [x] Make "please turn on Docker" an explicit thing in `core`
- [x] Unbreak all the conformance tests that are broken on this branch
- [x] Centaur assertions that the expected Docker images are being used
- [ ] Turn on conformance tests in backends that require Docker
- [ ] Fix globs on JES
- [ ] Fix inputs on JES
- [ ] Return validation failures to the caller

### DIFF
--- a/centaur/src/main/resources/standardTestCases/three_step/three_step.cwl
+++ b/centaur/src/main/resources/standardTestCases/three_step/three_step.cwl
@@ -2,13 +2,7 @@ cwlVersion: v1.0
 class: Workflow
 id: three_step
 hints:
-# hints can be anything (CwlAny)
-- fee: fi
-  fo: fum
-- eenie: meenie
-  minie: mo
-- 42
-- DockerRequirement:
+  DockerRequirement:
     dockerPull: "ubuntu:latest"
 inputs:
 - id: pattern
@@ -37,7 +31,7 @@ steps:
       # Check it: this DockerRequirement does not use the `class` formatting but should still parse
       # thanks to the magic of SALAD.
       DockerRequirement:
-        dockerPull: "debian:wheezy"
+        dockerPull: "ubuntu:bionic"
     baseCommand: ps
     stdout: ps-stdOut.txt
 - id: cgrep
--- a/cwl/src/main/scala-2.11/cwl/CwlExecutableValidation.scala
+++ b/cwl/src/main/scala-2.11/cwl/CwlExecutableValidation.scala
@@ -11,7 +11,8 @@ import io.circe.yaml
 import io.circe.literal._
 import common.Checked
 import common.validation.Checked._
-import wom.callable.ExecutableCallable
+import common.validation.ErrorOr.ErrorOr
+import wom.callable.{CallableTaskDefinition, ExecutableCallable, ExecutableTaskDefinition}
 import wom.executable.Executable
 import wom.executable.Executable.{InputParsingFunction, ParsedInputMap}
 
@@ -36,5 +37,15 @@ object CwlExecutableValidation {
       womDefinition <- callable
       executable <- Executable.withInputs(womDefinition, inputCoercionFunction, inputFile)
     } yield executable
+  }
+
+  def buildWomExecutable(callableTaskDefinition: ErrorOr[CallableTaskDefinition], inputFile: Option[String]): Checked[Executable] = {
+    import cats.data.Validated._
+    import common.validation.ErrorOr._
+    (for {
+      taskDefinition <- callableTaskDefinition
+      executableTaskDefinition = ExecutableTaskDefinition.tryApply(taskDefinition)
+      executable <- fromEither(CwlExecutableValidation.buildWomExecutable(executableTaskDefinition.toEither, inputFile))
+    } yield executable).toEither
   }
 }
--- a/cwl/src/main/scala/cwl/CommandLineTool.scala
+++ b/cwl/src/main/scala/cwl/CommandLineTool.scala
@@ -3,6 +3,7 @@ package cwl
 import java.nio.file.Paths
 
 import common.Checked
+import common.validation.ErrorOr.ErrorOr
 import cwl.CommandLineTool._
 import cwl.CwlType.CwlType
 import cwl.CwlVersion._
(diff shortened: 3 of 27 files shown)


mcovarr:
It might also be a plus if the unit tests compiled.

cwl/src/main/scala/cwl/CommandLineTool.scala
```diff
@@ -40,23 +43,58 @@ case class CommandLineTool private(
...
+      val validHints = (hintRequirements ++ parentHintRequirements).collect { case req if validator(req).isValid => req }
+      validRequirements ++ validHints
+    }
+  }
 
-  def taskDefinition: CallableTaskDefinition = {
+  def taskDefinition(parentWorkflow: Option[Workflow], validator: Requirement => ErrorOr[Requirement]): ErrorOr[CallableTaskDefinition] = {
 
-    val id = this.id
+    validateRequirementsAndHints(parentWorkflow, validator) map { requirementsAndHints =>
```
mcovarr: everything below here in this file is the same as before just indented due to being wrapped by a `map`

cwl/src/main/scala/cwl/TypeAliases.scala
```diff
@@ -30,6 +30,11 @@ trait TypeAliases {
       StepInputExpressionRequirement :+:
       CNil
 
+  type Hint =
+    Requirement :+:
```
mcovarr: Having an explicit `Hint` type with a `Requirement` as an element of its `CoProduct` makes it easy to `select` out only `Requirement`s.

project/Dependencies.scala
```diff
@@ -57,7 +57,7 @@ object Dependencies {
   val reactiveStreamsV = "1.0.1"
   val refinedV = "0.8.3"
   val scalaGraphV = "1.12.0"
-  val scalaLoggingV = "3.6.0"
+  val scalaLoggingV = "3.7.1"
```
mcovarr: so the 2.11 compiler doesn't explode

centaur/src/main/resources/standardTestCases/three_step/three_step.cwl
```diff
@@ -37,7 +31,7 @@ steps:
       # Check it: this DockerRequirement does not use the `class` formatting but should still parse
       # thanks to the magic of SALAD.
       DockerRequirement:
-        dockerPull: "debian:wheezy"
+        dockerPull: "ubuntu:bionic"
```
mcovarr: The `ps` command specifies `ubuntu:bionic` as one of its `requirements` and gets that. `cgrep` and `wc` don't express `hints` or `requirements`, but there is a `hints` at the workflow level that requests `ubuntu:latest`.

cwl/src/main/scala/cwl/CommandLineTool.scala
```diff
@@ -40,23 +43,58 @@ case class CommandLineTool private(
...
+        arguments.toSeq.flatMap(_.map(_.fold(ArgumentToCommandPart))) ++
+        CommandLineTool.orderedForCommandLine(inputs).map(InputParameterCommandPart.apply)
 
-    val runtimeAttributes: RuntimeAttributes = RuntimeAttributes(Map.empty[String, WomExpression])
+      val dockerRequirement = requirementsAndHints.toStream flatMap { _.select[DockerRequirement] } headOption
+      val attributeMap = for {
+        requirement <- dockerRequirement
+        // TODO CWL this only supports `dockerPull`. Not clear from the spec what to do if multiple attributes are specified.
+        pull <- requirement.dockerPull
+      } yield "docker" -> ValueAsAnExpression(WomString(pull))
```
mcovarr: The other occurrences of the `"docker"` literal used as a runtime attribute key are in packages that aren't visible to `cwl`. Not sure where the right place for these would be, maybe `common`?

cwl/src/main/scala-2.11/cwl/CwlExecutableValidation.scala
```diff
@@ -37,4 +38,14 @@ object CwlExecutableValidation {
       executable <- Executable.withInputs(womDefinition, inputCoercionFunction, inputFile)
     } yield executable
   }
+
+  def buildWomExecutable(callableTaskDefinition: ErrorOr[CallableTaskDefinition], inputFile: Option[String]): Checked[Executable] = {
```
mcovarr: ongoing 2.11 support is the gift that keeps on giving

cwl/src/main/scala/cwl/CommandLineTool.scala
```diff
@@ -40,23 +43,55 @@ case class CommandLineTool private(
...
+    CwlExecutableValidation.buildWomExecutable(taskDefinition, inputFile)
+  }
+
+  private def validateRequirementsAndHints(parentWorkflow: Option[Workflow], validator: Requirement => ErrorOr[Requirement]): ErrorOr[List[Requirement]] = {
+    import cats.instances.list._
+    import cats.syntax.traverse._
+
+    val allRequirements = requirements.toList.flatten ++ parentWorkflow.toList.flatMap(_.allRequirements)
+    // All requirements must validate or this fails.
+    val errorOrValidatedRequirements: ErrorOr[List[Requirement]] = allRequirements traverse validator
```
mcovarr: So this is perhaps a bit aggressive as it will validate requirements at the workflow level even if they are always overridden by requirements specified in subworkflows or tools.

cwl/src/main/scala/cwl/CommandLineTool.scala
```diff
@@ -40,23 +43,55 @@ case class CommandLineTool private(
                                    temporaryFailCodes: Option[Array[Int]],
                                    permanentFailCodes: Option[Array[Int]]) {
 
-  def womExecutable(inputFile: Option[String] = None): Checked[Executable] =
-    CwlExecutableValidation.buildWomExecutable(ExecutableTaskDefinition.tryApply(taskDefinition).toEither, inputFile)
+  /** Builds an `Executable` directly from a `CommandLineTool` CWL with no parent workflow. */
+  def womExecutable(validator: Requirement => ErrorOr[Requirement], inputFile: Option[String] = None): Checked[Executable] = {
```
mcovarr: This is the current shape for a `Requirement` validator function. It takes a `Requirement` and opines on its acceptability, returning the same `Requirement` in a `Valid` or an explanation for why it was unacceptable in an `Invalid[NonEmptyList[String]]`. Unacceptable `Requirement`s need to be handled differently depending on whether they are true `requirements` or merely `hints`. Any unsuccessful validations should be returned to the caller to address the error handling that's beyond the scope of this code.

cwl/src/main/scala/cwl/Workflow.scala
```diff
@@ -27,11 +27,22 @@ case class Workflow private(
...
   def womExecutable(inputFile: Option[String] = None): Checked[Executable] = {
     CwlExecutableValidation.buildWomExecutable(womDefinition, inputFile)
   }
 
+  var parentWorkflow: Option[Workflow] = None
+
+  val allRequirements: List[Requirement] = requirements.toList.flatten ++ parentWorkflow.toList.flatMap { _.allRequirements }
+
+  val allHints: List[Requirement] = {
+    // Just ignore any hint that isn't a Requirement.
```
mcovarr: TODO don't do that

cwl/src/main/scala/cwl/WorkflowStep.scala
```diff
@@ -67,7 +66,7 @@ case class WorkflowStep(
       // Create a task definition for the underlying run.
       // For sub workflows, we'll need to handle the case where this could be a workflow definition
       //TODO: turn this select into a fold that supports other types of runnables
-      val taskDefinition = run.select[CommandLineTool].map { _.taskDefinition } get
+      val taskDefinition = run.select[CommandLineTool].map { _.buildTaskDefinition(Option(workflow), _.validNel) } get
```
mcovarr: The `_.validNel` is a particularly easygoing implementation of a `Requirement` validator. This is just a placeholder to keep what everyone else is working on from breaking, unless you're working on the `ex:BlibberBlubberFakeRequirement` conformance test.

cwl/src/main/scala/cwl/package.scala
```diff
@@ -81,10 +82,14 @@ package object cwl extends TypeAliases {
 
   type WdlTypeMap = Map[String, WomType]
 
+  type RequirementsValidator = Requirement => ErrorOr[Requirement]
```
mcovarr: This is the current shape for a `Requirement` validator function. It takes a `Requirement` and opines on its acceptability, returning the same `Requirement` in a `Valid` or an explanation for why it was unacceptable in an `Invalid[NonEmptyList[String]]`. Unacceptable `Requirement`s need to be handled differently depending on whether they are true `requirements` or merely `hints`. Any unsuccessful validations should be returned to the caller to address the error handling that's beyond the scope of this code.

cwl/src/main/scala/cwl/CommandLineTool.scala
```diff
@@ -40,23 +43,55 @@ case class CommandLineTool private(
...
+
+    errorOrValidatedRequirements map { validRequirements =>
+      // Only Requirement hints, everything else is thrown out.
+      // TODO CWL don't throw them out but pass them back to the caller to do with as the caller pleases.
+      val hintRequirements = hints.toList.flatten.flatMap { _.select[Requirement] }
+      val parentHintRequirements = parentWorkflow.toList.flatMap(_.allHints)
+
+      // Throw out invalid Requirement hints.
+      // TODO CWL pass invalid hints back to the caller to do with as the caller pleases.
+      val validHints = (hintRequirements ++ parentHintRequirements).collect { case req if validator(req).isValid => req }
```
Horneth: aren't you revalidating them here ? I thought `val errorOrValidatedRequirements: ErrorOr[List[Requirement]] = allRequirements traverse validator` already validated them

cwl/src/main/scala/cwl/CommandLineTool.scala
```diff
@@ -40,23 +43,55 @@ case class CommandLineTool private(
...
 
-    val commandTemplate: Seq[CommandPart] = baseCommand.toSeq.flatMap(_.fold(BaseCommandToCommandParts)) ++
-      arguments.toSeq.flatMap(_.map(_.fold(ArgumentToCommandPart))) ++
-      CommandLineTool.orderedForCommandLine(inputs).map(InputParameterCommandPart.apply)
+      val commandTemplate: Seq[CommandPart] = baseCommand.toSeq.flatMap(_.fold(BaseCommandToCommandParts)) ++
+        arguments.toSeq.flatMap(_.map(_.fold(ArgumentToCommandPart))) ++
+        CommandLineTool.orderedForCommandLine(inputs).map(InputParameterCommandPart.apply)
 
-    val runtimeAttributes: RuntimeAttributes = RuntimeAttributes(Map.empty[String, WomExpression])
+      val dockerRequirement = requirementsAndHints.toStream flatMap { _.select[DockerRequirement] } headOption
```
Horneth: isn't this a `find` ?

cwl/src/main/scala/cwl/CommandLineTool.scala
Horneth: IMO I feel like it could even be in `core`. I see `common` more as a bunch of mostly non business logic, scala related code.

cwl/src/main/scala/cwl/Workflow.scala
```diff
@@ -27,9 +27,20 @@ case class Workflow private(
...
-                     hints: Option[Array[CwlAny]]) {
-  def womExecutable(inputFile: Option[String] = None): Checked[Executable] = {
-    CwlExecutableValidation.buildWomExecutable(womDefinition, inputFile)
+                     hints: Option[Array[Hint]]) {
+  /** Builds an `Executable` from a `Workflow` CWL with no parent `Workflow` */
+  def womExecutable(validator: RequirementsValidator, inputFile: Option[String] = None): Checked[Executable] = {
+    CwlExecutableValidation.buildWomExecutable(womDefinition(validator), inputFile)
+  }
+
+  var parentWorkflow: Option[Workflow] = None
```
Horneth: This should probably be `private` or `private[cwl]`

Horneth: Also I don't see it being set anywhere ?

cwl/src/main/scala/cwl/CommandLineTool.scala
mcovarr: Oh right, yes. I meant `core` not `common` née `lenthall`.

cwl/src/main/scala/cwl/CommandLineTool.scala
mcovarr: these are the `hints`

cwl/src/main/scala/cwl/Workflow.scala
mcovarr: It's not being set currently, the subworkflow stuff hadn't landed when I started this.

cwl/src/main/scala/cwl/CommandLineTool.scala
mcovarr: no this is picking out the first `Requirement` it finds in priority order, if any

cwl/src/main/scala/cwl/CommandLineTool.scala
```diff
@@ -3,60 +3,96 @@ package cwl
...
+                                   arguments: Option[Array[CommandLineTool.Argument]],
+                                   stdin: Option[StringOrExpression],
+                                   stderr: Option[StringOrExpression],
+                                   stdout: Option[StringOrExpression],
+                                   successCodes: Option[Array[Int]],
+                                   temporaryFailCodes: Option[Array[Int]],
+                                   permanentFailCodes: Option[Array[Int]]) {
+
+  /** Builds an `Executable` directly from a `CommandLineTool` CWL with no parent workflow. */
+  def womExecutable(validator: RequirementsValidator, inputFile: Option[String] = None): Checked[Executable] = {
```
danbills: ToL Maybe creating a new type using the Reader Monad could clean up the passing around of this validator.  I'll give it a quick shot and see if it just makes things better or worse.

cwl/src/main/scala/cwl/CwlExpressionCommandPart.scala
```diff
@@ -67,7 +67,7 @@ case class InputParameterCommandPart(commandInputParameter: CommandInputParamete
         case cip: CommandInputParameter =>
           val localizedId = LocalName(FullyQualifiedName(cip.id).id)
           inputsMap.get(localizedId) match {
-            case Some(x) =>x
+            case Some(x) => valueMapper(x)
```
danbills: ToL (a little outside scope) a `fold` on the option could let you specify `valueMapper` w/o arguments and generally reduce code here.

cwl/src/main/scala/cwl/ParameterContext.scala
```diff
@@ -12,11 +12,11 @@ case class ParameterContext(inputs: WomValue = WomOptionalValue(WomNothingType,
                             self: WomValue = WomOptionalValue(WomNothingType, None),
                             runtime: WomValue = WomOptionalValue(WomNothingType, None)) {
   def withInputs(inputValues: Map[String, WomValue], ioFunctionSet: IoFunctionSet): ParameterContext = {
-    val wdlValueType = WomStringType
+    val womValueType = WomStringType
     copy(
       inputs = WomMap(
-        WomMapType(WomStringType, wdlValueType),
-        // TODO: WOM: convert inputValues (including WdlFile?) to inputs using the ioFunctionSet
+        WomMapType(WomStringType, womValueType),
```
danbills: I do appreciate the wdl -> wom effort but there's so much it's a little distracting

cwl/src/main/scala/cwl/CwlExpressionCommandPart.scala
mcovarr: I'll fix that in my next PR

cwl/src/main/scala/cwl/ParameterContext.scala
mcovarr: this is everything

PR #3049, opened 2017-12-11
Allow 0 shard scatters

### LINKED ISSUE #3037: Scattering over an empty array fails the workflow

cjllanwarne:
Closes #3037

Allows a scatter to be collected even if it's 0-width

### DIFF
--- a/centaur/src/main/resources/standardTestCases/empty_scatter/empty_scatter.wdl
+++ b/centaur/src/main/resources/standardTestCases/empty_scatter/empty_scatter.wdl
@@ -1,0 +1,28 @@
+workflow empty_scatter {
+  Array[Int] xs = []
+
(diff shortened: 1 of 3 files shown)


centaur/src/main/resources/standardTestCases/empty_scatter/empty_scatter.wdl
```diff
@@ -0,0 +1,25 @@
...
+  output {
+    Int task_outs_length = length(do_nothing.str)
+    Int decls_length = length(decl)
+  }
+}
+
+# We should never invoke this since we're scattering over an empty array
+task do_nothing {
+  Int whoa
+  command { # Do nothing }
```
mcovarr: `exit 1` might be better in this case

PR #3050, opened 2017-12-11
Show optional inputs in womtool

cjllanwarne:
And help @abaumann who also wants to use the `workflow.inputs` method


danbills:
@cjllanwarne I'm going to see this through, feel better soon!

abaumann:
:success_kid:

PR #3053, opened 2017-12-12
Fix 2.11 compile & a small fixup for InputLookupExpression

danbills:


danbills:
Going to merge this as it fixes the 2.11 build and is miniscule.

Issue #2611, opened 2017-09-11
Language specific configuration (not enforcing read_X limits)


Horneth:
Design a way for Cromwell to pass language specific configuration to the various bindings.
The use case for this is read_*** bytes limitation that are different for different WDL functions which Cromwell has no knowledge of anymore.
CWL also requires the same type of limitation.

danbills:
not fixed yet

PR #3057, opened 2017-12-13
Split WomFileType into WomSingleFileType and WomGlobFileType.

kshakir:
Replaced use of isGlob with using the womType.
Removed unused WomFile.appendPathsWithSlashSeparators.
Added a WomFile.mapFile to use instead of WomFile.apply(value, isGlob).


backend/src/test/scala/cromwell/backend/wdl/ReadLikeFunctionsSpec.scala
```diff
@@ -35,32 +35,40 @@ class ReadLikeFunctionsSpec extends FlatSpec with Matchers {
 
   it should "correctly report the size of a supplied, optional, 2048 byte file" in {
     val readLike = new TestReadLikeFunctions(Success(2048d))
-    readLike.size(Seq(Success(WomOptionalValue(WomFileType, Some(WomSingleFile("blah")))))) should be(Success(WomFloat(2048d)))
+    readLike.size(Seq(Success(WomOptionalValue(WomSingleFileType, Option(WomSingleFile("blah")))))) should
+      be(Success(WomFloat(2048d)))
   }
 
   it should "correctly report the size of a supplied, optional optional, 2048 byte file" taggedAs PostWomTest ignore {
```
mcovarr: TOL so many `ignore`d `PostWomTest`s

wom/src/main/scala/wom/values/WomFile.scala
```diff
@@ -1,48 +1,69 @@
...
+  }
 
-  override def valueString = value.toString
+  def mapSingleFile(womFile: WomSingleFile)(f: String => String): WomSingleFile = {
+    WomSingleFile(f(womFile.value))
+  }
+
+  def mapGlobFile(womFile: WomGlobFile)(f: String => String): WomGlobFile = {
+    womFile match {
+      case WomGlobFile(value) => WomGlobFile(f(value))
```
mcovarr: why the pattern match on the known type here but just the member access above in `mapSingleFile`?

wom/src/main/scala/wom/values/WomFile.scala
```diff
@@ -1,48 +1,69 @@
...
+
+  def mapFile(womFile: WomFile)(f: String => String): WomFile = {
+    womFile match {
+      case womSingleFile: WomSingleFile => mapSingleFile(womSingleFile)(f)
+      case womGlobFile: WomGlobFile => mapGlobFile(womGlobFile)(f)
+    }
+  }
 
-  override def valueString = value.toString
+  def mapSingleFile(womFile: WomSingleFile)(f: String => String): WomSingleFile = {
```
mcovarr: why are these curried?

backend/src/test/scala/cromwell/backend/wdl/ReadLikeFunctionsSpec.scala
Horneth: This one got a ticket https://github.com/broadinstitute/cromwell/issues/2611

wom/src/main/scala/wom/values/WomFile.scala
kshakir: Muscle memory from the FP book club.

wom/src/main/scala/wom/values/WomFile.scala
kshakir: Uhm...... this was a test for reviewers. You passed!

wom/src/main/scala/wom/values/WomFile.scala
Horneth: maybe they could be define directly in `WomSingleFile` ?

kshakir: In a future commit for adding directories and additional-files, there is a sub-trait of WomFile coming that specifies _FileOrDirButNotGlob_. There I need a map over the sub-trait to return the sub-trait. I have a feeling that typeclasses/FP-ness might help, but I've been hesitating to start introducing a lot of that into Wom, yet.

wom/src/main/scala/wom/types/WomFileType.scala
```diff
@@ -1,28 +1,54 @@
...
+  override def add(rhs: WomType): Try[WomType] = rhs match {
+    case WomStringType => Success(WomSingleFileType)
+    case WomOptionalType(memberType) => add(memberType)
+    case _ => invalid(s"$this + $rhs")
+  }
+
+  override def equals(rhs: WomType): Try[WomType] = rhs match {
+    case WomSingleFileType => Success(WomBooleanType)
+    case WomStringType => Success(WomBooleanType)
+    case WomOptionalType(memberType) => equals(memberType)
```
Horneth: Is this true ? `WomSingleFileType == WomOptionalType(WomSingleFileType))` ?

kshakir: Interesting question. I don't know, and mainly refactored what was there. I haven't checked the spec, and the unit tests don't explicitly check for this case. This code is "covered by codecov", but only for the negative case via:

https://github.com/broadinstitute/cromwell/blob/bfa8378f0fa7b645df9c145e21f3b1a6f0c569c7/wdl/src/test/scala/wdl/expression/ValueEvaluatorSpec.scala#L31-L44

https://github.com/broadinstitute/cromwell/blob/bfa8378f0fa7b645df9c145e21f3b1a6f0c569c7/wdl/src/test/scala/wdl/expression/ValueEvaluatorSpec.scala#L472-L473

Horneth: okay, must be here for a reason. Maybe @cjllanwarne knows.
Anyway don't block the merging for this :)

cjllanwarne: @kshakir @Horneth hmmmm, this looks a lot like a "this helps in some awkward situation and doesn't seem to break anything" kind of a line. 

I guess I'd like to see what breaks if we remove it and see if we can resolve that more appropriately (or maybe I've just forgotten what `equals` is used for in types). On the other hand if a tree falls down in the woods and nothing else seems to be broken...

PR #3064, opened 2017-12-14
Allow member accesses to parse

### LINKED ISSUE #3060: Cromwell v30 object type regression
I'm getting an error when trying to run the following WDL, which is using an `Object` type for the output of one of the tasks: https://github.com/HumanCellAtlas/pipeline-tools/blob/master/adapter_pipelines/smart_seq2/adapter.wdl#L46. This WDL previously worked in Cromwell 29

The WDL fails immediately on validation with this error:
```
 "failures": [
    {
      "causedBy": [
        {
          "causedBy": [],
          "message": "Some([Declaration type=Object name=prep.inputs expr=Some(prep.inputs)]) (of class scala.Some)"
        }
      ],
      "message": "Workflow input processing failed"
    }
  ]
```

We're relying on objects in our HCA pipelines so it would be great if this could get fixed soon!

cjllanwarne:
Fixes #3060 (or at least the parsing of it)

### DIFF
--- a/core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala
+++ b/core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala
@@ -92,7 +92,8 @@ object WomValueBuilder {
     }
 
     outputType match {
-      case _: WomPrimitiveType => components collectFirst { case SimpletonComponent(_, v) => v } get
+      case _: WomPrimitiveType =>
+        components collectFirst { case SimpletonComponent(_, v) => v } get
       case opt: WomOptionalType =>
         if (components.isEmpty) {
           WomOptionalValue(opt.memberType, None)
@@ -109,6 +110,26 @@ object WomValueBuilder {
       case pairType: WomPairType =>
         val groupedByLeftOrRight: Map[PairLeftOrRight, Traversable[SimpletonComponent]] = group(components map descendIntoPair)
         WomPair(toWomValue(pairType.leftType, groupedByLeftOrRight(PairLeft)), toWomValue(pairType.rightType, groupedByLeftOrRight(PairRight)))
+      case WomObjectType =>
+        val groupedByMapKey: Map[String, Traversable[SimpletonComponent]] = group(components map descendIntoMap)
+        // map keys are guaranteed by the WDL spec to be primitives, so the "coerceRawValue(..).get" is safe.
+        val map: Map[String, WomValue] = groupedByMapKey map { case (k, ss) => k -> toWomValue(WomAnyType, ss) }
+        WomObject(map)
+      case WomAnyType =>
+        // Ok, we're going to have to guess, but the keys should give us some clues:
+        if (components forall { component => MapElementPattern.findFirstMatchIn(component.path).isDefined }) {
+          // Looks like an object
+          toWomValue(WomObjectType, components)
+        } else if (components forall { component => ArrayElementPattern.findFirstMatchIn(component.path).isDefined }) {
+          // Looks like an array:
+          toWomValue(WomArrayType(WomAnyType), components) match {
+            case WomArray(_, values) => WomArray(values)
+          }
+
+        } else {
+          // Treat it as a primitive:
+          components collectFirst { case SimpletonComponent(_, v) => v } get
+        }
     }
   }
 
@@ -143,8 +164,9 @@ object WomValueBuilder {
     // "dehydrated" to WomValueSimpletons correctly. This code is not robust to corrupt input whatsoever.
     val types = taskOutputs map { o => o -> o.womType } toMap
     val simpletonsByOutputName = simpletons groupBy { _.simpletonKey match { case IdentifierAndPathPattern(i, _) => i } }
-    val simpletonComponentsByOutputName = simpletonsByOutputName map { case (name, ss) => name -> (ss map simpletonToComponent(name)) }
-    types map { case (outputPort, outputType) => outputPort -> toWomValue(outputType, simpletonComponentsByOutputName(outputPort.name))}
+    val simpletonComponentsByOutputName: Map[String, Traversable[SimpletonComponent]] =
+      simpletonsByOutputName map { case (name, ss) => name -> (ss map simpletonToComponent(name)) }
+    types map { case (outputPort, outputType) => outputPort -> toWomValue(outputType, simpletonComponentsByOutputName.getOrElse(outputPort.name, Seq.empty))}
   }
 }
 
--- a/wdl/src/main/scala/wdl/MemberAccess.scala
+++ b/wdl/src/main/scala/wdl/MemberAccess.scala
@@ -25,7 +25,9 @@ import wdl.AstTools.EnhancedAstNode
  *  }
  *}
  */
-case class MemberAccess(lhs: String, rhs: String)
+case class MemberAccess(lhs: String, rhs: String, ast: Ast) {
+  def memberAccessString = s"$lhs.$rhs"
+}
 
 object MemberAccess {
   def apply(ast: Ast): MemberAccess = {
@@ -36,6 +38,6 @@ object MemberAccess {
       case terminal: Terminal => terminal.sourceString
     }
 
-    MemberAccess(lhs, rhs)
+    MemberAccess(lhs, rhs, ast)
   }
 }
--- a/wdl/src/main/scala/wdl/WdlNamespace.scala
+++ b/wdl/src/main/scala/wdl/WdlNamespace.scala
@@ -13,6 +13,7 @@ import wdl.command.ParameterCommandPart
 import wdl.exception._
 import wdl.expression.{NoFunctions, WdlStandardLibraryFunctions, WdlStandardLibraryFunctionsType}
 import wdl.types.{WdlCallOutputsObjectType, WdlNamespaceType}
+import wdl4s.parser.WdlParser
 import wdl4s.parser.WdlParser._
 import wom.core._
 import wom.executable.Executable
@@ -486,10 +487,8 @@ object WdlNamespace {
      * if we have "input: var=ns.ns1.my_task" it does not make sense to validate "ns1.my_task" by itself as it only
      * makes sense to validate that "ns.ns1.my_task" as a whole is coherent
      *
-     * Run for its side effect (i.e. Exception) but we were previously using a Try and immediately calling .get on it
-     * so it's the same thing
      */
-    val invalidMemberAccesses = callInputSections flatMap { ast =>
+    val invalidMemberAccesses: Seq[WdlParser.SyntaxError] = callInputSections flatMap { ast =>
       ast.getAttribute("value").findTopLevelMemberAccesses flatMap { memberAccessAst =>
         val memberAccess = MemberAccess(memberAccessAst)
         val requestedValue = memberAccess.rhs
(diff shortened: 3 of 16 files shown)


wdl/src/main/scala/wdl/WdlNamespace.scala
```diff
@@ -13,10 +13,11 @@ import wdl.command.ParameterCommandPart
 import wdl.exception._
 import wdl.expression.{NoFunctions, WdlStandardLibraryFunctions, WdlStandardLibraryFunctionsType}
 import wdl.types.{WdlCallOutputsObjectType, WdlNamespaceType}
+import wdl4s.parser.WdlParser
 import wdl4s.parser.WdlParser._
 import wom.core._
 import wom.executable.Executable
-import wom.types._
+import wom.types.{WomObjectType, _}
```
danbills: Revert

cjllanwarne: IntelliJ! :shakes_fist_at_sky:

core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala
```diff
@@ -109,6 +111,27 @@ object WomValueBuilder {
...
         WomPair(toWomValue(pairType.leftType, groupedByLeftOrRight(PairLeft)), toWomValue(pairType.rightType, groupedByLeftOrRight(PairRight)))
+      case WomObjectType =>
+        val groupedByMapKey: Map[String, Traversable[SimpletonComponent]] = group(components map descendIntoMap)
+        // map keys are guaranteed by the WDL spec to be primitives, so the "coerceRawValue(..).get" is safe.
+        val map: Map[String, WomValue] = groupedByMapKey map { case (k, ss) => k -> toWomValue(WomAnyType, ss) }
+        WomObject(map)
+      case WomAnyType =>
+        // Ok, we're going to have to guess, but the keys should give us some clues:
+        if (components forall { component => MapElementPattern.findFirstMatchIn(component.path).isDefined }) {
+          // Looks like an object
```
cjllanwarne: I guess I should test this case too

core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala
```diff
@@ -92,7 +92,9 @@ object WomValueBuilder {
     }
 
     outputType match {
-      case _: WomPrimitiveType => components collectFirst { case SimpletonComponent(_, v) => v } get
+      case _: WomPrimitiveType =>
+        assert(components.size == 1)
```
Horneth: I think assert can take a second argument that describes the failure. Could be nice to add some explanation.

core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala
```diff
@@ -109,6 +111,27 @@ object WomValueBuilder {
...
+      case WomObjectType =>
+        val groupedByMapKey: Map[String, Traversable[SimpletonComponent]] = group(components map descendIntoMap)
+        // map keys are guaranteed by the WDL spec to be primitives, so the "coerceRawValue(..).get" is safe.
+        val map: Map[String, WomValue] = groupedByMapKey map { case (k, ss) => k -> toWomValue(WomAnyType, ss) }
+        WomObject(map)
+      case WomAnyType =>
+        // Ok, we're going to have to guess, but the keys should give us some clues:
+        if (components forall { component => MapElementPattern.findFirstMatchIn(component.path).isDefined }) {
+          // Looks like an object
+          toWomValue(WomObjectType, components)
```
Horneth: Could it not be a `Map` ?

core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala
mcovarr: why are we doing `assert`s?

core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala
```diff
@@ -109,6 +111,27 @@ object WomValueBuilder {
       case pairType: WomPairType =>
         val groupedByLeftOrRight: Map[PairLeftOrRight, Traversable[SimpletonComponent]] = group(components map descendIntoPair)
         WomPair(toWomValue(pairType.leftType, groupedByLeftOrRight(PairLeft)), toWomValue(pairType.rightType, groupedByLeftOrRight(PairRight)))
+      case WomObjectType =>
+        val groupedByMapKey: Map[String, Traversable[SimpletonComponent]] = group(components map descendIntoMap)
+        // map keys are guaranteed by the WDL spec to be primitives, so the "coerceRawValue(..).get" is safe.
```
mcovarr: doesn't the spec also guarantee the case you're `assert`ing on above?

core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala
cjllanwarne: It could, but to check that I'd need to analyze every entry to make sure the value types are always the same. This just seemed easier (and I believe they're coerceable with a run-time check)

Structs should make this oh so much easier since we won't have to guess anymore

core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala
mcovarr: 👁

core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala
```diff
@@ -109,6 +111,27 @@ object WomValueBuilder {
...
+          toWomValue(WomObjectType, components)
+        } else if (components forall { component => ArrayElementPattern.findFirstMatchIn(component.path).isDefined }) {
+          // Looks like an array:
+          toWomValue(WomArrayType(WomAnyType), components) match {
+            case WomArray(_, values) => WomArray(values)
+          }
+
+        } else {
+          // Treat it as a primitive:
+          assert(components.size == 1)
```
mcovarr: 👁

wdl/src/main/scala/wdl/MemberAccess.scala
```diff
@@ -36,6 +38,6 @@ object MemberAccess {
       case terminal: Terminal => terminal.sourceString
     }
 
-    MemberAccess(lhs, rhs)
+    MemberAccess(lhs, rhs, ast)
   }
 }
```
mcovarr: \n while you're in here

wdl/src/main/scala/wdl/WdlNamespace.scala
```diff
@@ -480,16 +481,14 @@ object WdlNamespace {
...
-     * so it's the same thing
-     */
-    val invalidMemberAccesses = callInputSections flatMap { ast =>
+//    /*
+//     * Ensures that the lhs corresponds to a call and the rhs corresponds to one of its outputs. We're only checking
+//     * top level MemberAccess ASTs because the sub-ASTs don't make sense w/o the context of the parent. For example
+//     * if we have "input: var=ns.ns1.my_task" it does not make sense to validate "ns1.my_task" by itself as it only
+//     * makes sense to validate that "ns.ns1.my_task" as a whole is coherent
+//     *
+//     */
```
mcovarr: um what

core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala
kshakir: ToL: IMHO this and all others usages in Cromwell should be `Predef.require` instead of `Predef.assert`.

I don't like `Predef.assert` for the same reasons as java's `assert` keyword:
1. They have different meanings depending on how the program is [run](https://stackoverflow.com/questions/2758224/what-does-the-java-assert-keyword-do-and-when-should-it-be-used) or [compiled](http://daily-scala.blogspot.com/2010/04/elidable-remove-method-calls-at-compile.html). `require` _always_ throws an exception on bad inputs. `assert` only _sometimes_ throws an error on bad inputs.
2. When an `Predef.assert` or a java `assert` does trip an [`Error`](https://docs.oracle.com/javase/8/docs/api/java/lang/AssertionError.html) it indicates a "[serious problems that a reasonable application should not try to catch](https://docs.oracle.com/javase/8/docs/api/java/lang/Error.html)", often shutting down those reasonable applications. Again, IMHO, instead of Cromwell not being able to run at all, an `Exception` via `Predef.require` should be thrown instead where the one workflow will fail and the others will keep running until the bug can be reported and fixed.

wom/src/main/scala/wom/types/WomPrimitiveType.scala
```diff
@@ -3,6 +3,7 @@ package wom.types
 abstract class WomPrimitiveType extends WomType {
   lazy val coercionMap: Map[WomType, Seq[WomType]] = Map(
     // From type -> To type
+    WomAnyType -> Seq(WomStringType, WomIntegerType, WomFloatType, WomSingleFileType, WomBooleanType),
```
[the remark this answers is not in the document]
cjllanwarne: `Object` is not a primitive type

core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala
cjllanwarne: This is sort of why I like asserts. They get run with your tests so you know they're getting validated, but aren't part of the final code.

They're like comments saying "this should never happen" that actually get validated during tests.

wom/src/main/scala/wom/types/WomPrimitiveType.scala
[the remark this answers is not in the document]
mcovarr: oh duh nm... 😳

core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala
geoffjentry: I hate asserts in real code

wdl/src/main/scala/wdl/WdlNamespace.scala
cjllanwarne: 😳 this probably got swept up in a big <kbd>ALT</kbd> <kbd>/</kbd> and then didn't get unswept by the subsequent "reinstate everything"

core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala
cjllanwarne: IMO they're strictly better than comments like `"I know Condition X will always be true here"`

geoffjentry: Also true, but that doesn’t make assert non-sucky

cjllanwarne: I think they can be a useful tool, but will agree that they do have with a *very* narrow use case. 

I'm happy to remove them if I'm the only one in Team "asserts are ok sometimes" (I added them for my own sanity checking when running the tests, I thought others might enjoy similar sanity verification at a later date)

cjllanwarne: OK, vote here via "reaction emoji" on this comment:
- thumbs up: keep the `assert`s
- thumbs down: remove
- grin: "I'm just observing this"
- party: "I'm not even observing this"
- confused: "what's happening"
- heart: "please don't take my vote against you as a personal sleight"

core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala
cjllanwarne: Done

core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala
cjllanwarne: I added a test case to make sure we can coerce the result back into the original `Map` type

PR #3066, opened 2017-12-15
somatic test cases for gatk-launch

ruchim:


ruchim:
@cjllanwarne new lines adjusted, so are my intellij settings thanks to Jose. I'm thinking the integrationTestCases should run weekly instead of nightly even. @jsotobroad as the original creator of the integrationTestCases -- does that sound okay?

geoffjentry:
@ruchim Can you explain the reasoning behind daily vs weekly?

jsotobroad:
The team should decide at what cadence they want to see that current develop does have any regression as far as whatever workflows it runs is concerned.  At the time when we created those tests it was decided the team wanted to see them daily.  If its a larger test bed then maybe weekly makes sense in terms of $$ and how little/much does the team get out of seeing it daily.

geoffjentry:
In that case I’d side with more and dialing down instead of less and dialing up

ruchim:
@geoffjentry The main advantage of running these test cases daily vs weekly seems to be that it’s easier to narrow down which change couldve caused this test suite to fail. However, it seems unlikely to me that these tests could find breaking changes everyday that the centaur standard test cases wouldn’t already uncover. I see these tests as a release requirement for Cromwell more than anything else. However, it’s totally upto the team on whatever makes them most comfortable, I don’t have a strong opinion on it.

centaur/src/main/resources/integrationTestCases/Somatic/cnv_common_tasks.wdl
```diff
@@ -0,0 +1,172 @@
...
+        memory: machine_mem + " MB"
+        disks: "local-disk " + disk_size + " HDD"
+        preemptible: select_first([preemptible_attempts, 5])
+    }
+
+    output {
+        String entity_id = base_filename
+        File allelic_counts = allelic_counts_filename
+    }
+}
```
mcovarr: \n

PR #3067, opened 2017-12-15
Use correct IO Functions for JES outputs generator

ruchim:


supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
```diff
@@ -212,7 +212,7 @@ class JesAsyncBackendJobExecutionActor(override val standardParams: StandardAsyn
     import cats.syntax.validated._
     def evaluateFiles(output: OutputDefinition): List[WomFile] = {
       Try (
-        output.expression.evaluateFiles(jobDescriptor.localInputs, NoIoFunctionSet, output.womType).map(_.toList)
+        output.expression.evaluateFiles(jobDescriptor.localInputs, backendEngineFunctions, output.womType).map(_.toList)
```
mcovarr: :man_facepalming:

centaur/src/main/resources/standardTestCases/output_filename_interpolation/output_filename_interpolation.wdl
```diff
@@ -8,9 +8,11 @@ task output_filename_interpolation {
         echo a > ${outputMain}
         echo b > ${outputMain + "." + outputSuffix}
         echo c > ${outputPrefix}/${outputMain}.${outputSuffix}
+        echo ${outputMain} > someFile.txt
     }
 
     output {
+        File f = read_string("someFile.txt")
```
cjllanwarne: NB this only works because `Upside` is already being delocalized already in another task.

ruchim: :man_facepalming: definitely the wrong test...its been adjusted!

centaur/src/main/resources/standardTestCases/output_filename_interpolation/output_filename_interpolation.wdl
```diff
@@ -8,11 +8,12 @@ task output_filename_interpolation {
...
         echo b > ${outputMain + "." + outputSuffix}
         echo c > ${outputPrefix}/${outputMain}.${outputSuffix}
-        echo ${outputMain} > someFile.txt
+        echo foo > bar
+        echo bar > baz
     }
 
     output {
-        File f = read_string("someFile.txt")
+        File f = read_string("baz")
```
cjllanwarne: I don't believe this will delocalize `bar`. So if you try to use `f` in a subsequent task, it won't be available

centaur/src/main/resources/standardTestCases/output_filename_interpolation.test
```diff
@@ -15,5 +15,5 @@ metadata {
 
 fileSystemCheck: "gcs"
 outputExpectations {
-  "gs://cloud-cromwell-dev/output_filename_interpolation_wf/<<UUID>>/call-output_filename_interpolation/Upside": 1
+  "gs://cloud-cromwell-dev/output_filename_interpolation_wf/<<UUID>>/call-output_filename_interpolation/baz": 1
```
cjllanwarne: I believe `baz` would actually contain `gs://[...]/bar`. And I also believe that the `bar` file won't exist.

PR #3069, opened 2017-12-15
Abort more

Horneth:
Stumbled upon this while trying to abort a CWL workflow that was failing to submit to JES.
Aborting it was not short circuiting the "try 10 times" mechanism of creating JES runs or polling.
The workflow do end up in `Aborted` state after the 10 retries so we don't necessarily have to hotfix it.


supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesApiQueryManager.scala
```diff
@@ -72,6 +71,7 @@ class JesApiQueryManager(val qps: Int Refined Positive, override val serviceRegi
   // workQueue is protected for the unit tests, not intended to be generally overridden
   protected[statuspolling] var workQueue: Queue[JesApiQuery] = Queue.empty
   private var workInProgress: Map[ActorRef, JesPollingWorkBatch] = Map.empty
+  private var queriesWaitingForRetry: Set[JesApiQuery] = Set.empty[JesApiQuery]
```
cjllanwarne: Could you add a comment along the lines of "this is the set of queries that are going to be retried on schedule, unless they get cancelled in the meantime"?

PR #3072, opened 2017-12-15
Check for existence of auth file

Horneth:
It seems that sometimes the request to write the auth file fails but it still ends up being written on GCS. However because we write it in "fail if this file already exists mode", every subsequent retry to write the file fails.
Instead check for existence before and act accordingly. This is less clean as the "check if exist then write" is not atomic anymore but I don't see another simple way to fix this.


supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesInitializationActor.scala
```diff
@@ -81,18 +84,18 @@ class JesInitializationActor(jesParams: JesInitializationActorParams)
...
-          Future.successful(())
-        case CromwellFatalException(e: StorageException) if e.getCode == StatusCodes.PreconditionFailed.intValue =>
-            Future.failed(new IOException(s"Failed to upload authentication file:" +
-              s" there was already a file at the same location and this workflow was not being restarted: ${e.getMessage}"))
-        case failure => Future.failed(new IOException(s"Failed to upload authentication file", failure))
-      }
+    def fileUpload(paths: JesWorkflowPaths) = existsAsync(paths.gcsAuthFilePath) flatMap {
+      // if it is then something is definitely wrong
+      case true if !jesParams.restarting =>
+        workflowLogger.debug(s"Authentication file already exists but this is a restart, proceeding.")
```
mcovarr: looks like this is *not* a restart? and this is failing?

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesInitializationActor.scala
```diff
@@ -81,18 +84,18 @@ class JesInitializationActor(jesParams: JesInitializationActorParams)
...
+        Future.failed(AuthFileAlreadyExistsException(paths.gcsAuthFilePath))
+      case true if jesParams.restarting => Future.successful(())
```
mcovarr: don't really need the `if` here

engine/src/main/scala/cromwell/engine/io/gcs/GcsBatchFlow.scala
```diff
@@ -104,7 +104,9 @@ class GcsBatchFlow(batchSize: Int, scheduler: Scheduler, onRetry: IoCommandConte
     }
 
     // Add all requests to the batch
-    contexts foreach { _.queue(batch) }
+    contexts foreach { context =>
+      context.queue(batch) 
+    }
```
mcovarr: ?

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesInitializationActor.scala
Horneth: Yes if it is **not** a restart it means it should be the first time we ever try to upload the auth file so we definitely don't expect it to be there already.

mcovarr: so then this message is not right?

Horneth: Oh yep :) good catch :p

Horneth:
There was indeed a case where a workflow could get stuck. I added a commit to this PR that fixed it.
I think we should hotfix this to 30 (at least the last commit for sure)

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -669,7 +669,7 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
           customPollStatusFailure orElse {
             case (_: ExecutionHandle, exception: Exception) if isFatal(exception) =>
               // Log exceptions and return the original handle to try again.
-              jobLogger.warn(s"Fatal exception polling for status. Job will fail.")
+              jobLogger.warn(s"Fatal exception polling for status. Job will fail.", exception)
```
cjllanwarne: ToL: I assume we only see this if the job is really actually about to fail? I don't want to see 100s of these popping up for retried jobs.

centaur/src/main/resources/standardTestCases/chainfail.test
```diff
@@ -0,0 +1,9 @@
+name: chainfail
```
cjllanwarne: Please add some commentary here on what the test is for, why we need it, etc

core/src/main/scala/cromwell/core/io/IoCommand.scala
```diff
@@ -101,3 +101,11 @@ abstract class IoTouchCommand(val file: Path) extends SingleFileIoCommand[Unit]
   override def toString = s"touch ${file.pathAsString}"
   override lazy val name = "touch"
 }
+
+/**
+  * Check if a file exists
+  */
+abstract class IoExistsCommand(val file: Path) extends SingleFileIoCommand[Boolean] {
+  override def toString = s"check if ${file.pathAsString} exists"
```
cjllanwarne: `"... if ..."` => `"... whether ..."`

engine/src/main/scala/cromwell/engine/io/IoActor.scala
```diff
@@ -170,12 +170,19 @@ object IoActor {
     503
   )
   
+  val AdditionalRetryableErrorMessages = List(
```
cjllanwarne: Comment pls

cjllanwarne: also, why "Additional"?

engine/src/main/scala/cromwell/engine/io/IoActor.scala
```diff
@@ -170,12 +170,19 @@ object IoActor {
...
   /**
     * Failures that are considered retryable.
     * Retrying them should increase the "retry counter"
     */
   def isRetryable(failure: Throwable): Boolean = failure match {
-    case gcs: StorageException => gcs.isRetryable || AdditionalRetryableHttpCodes.contains(gcs.getCode) || isRetryable(gcs.getCause)
+    case gcs: StorageException => gcs.isRetryable || 
+      AdditionalRetryableHttpCodes.contains(gcs.getCode) || 
+      isRetryable(gcs.getCause) ||
+      AdditionalRetryableErrorMessages.exists(_.contains(gcs.getMessage))
```
cjllanwarne: Shouldn't this be `[...].exists(gcs.getMessage.contains(_))`? It seems like the list we store is substrings of messages that they might return?

cjllanwarne: Also do we want to throw in an `ignoreCase`?

engine/src/main/scala/cromwell/engine/io/gcs/GcsBatchCommandContext.scala
```diff
@@ -84,12 +84,27 @@ final case class GcsBatchCommandContext[T, U](request: GcsBatchIoCommand[T, U],
     promise.trySuccess(promiseResponse)
     ()
   }
+  
+  private def handleSuccess(successResult: Either[T, GcsBatchIoCommand[T, U]]) = {
+    val promiseResponse: BatchResponse = successResult match {
+      // Left means the command is complete, so just create the corresponding IoSuccess with the value
+      case Left(responseValue) => Left(success(responseValue))
```
cjllanwarne: ToL: I'll let @danbills or @mcovarr speak with more authority, but I think I've been warned off of using `Either` as a simple type of Coproduct, I think because `Either` is so often used as "`Left`=bad, `Right`=good" that it's confusing to see it used any other way.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/keys/ScatterKey.scala
```diff
@@ -22,6 +22,42 @@ private [execution] case class ScatterKey(node: ScatterNode) extends JobKey {
...
+  * is a TaskCallNode. This is because we expect all the output ports of a TaskCallNode to be available at the same time,
+  * indeed we get all the outputs for a task call at once (when the call is complete). If this assumption is no longer
+  * true and we get partial output values while a task is running this optimization would need to be removed.
+  * We make that same assumption for WorkflowCallNodes (sub workflows) but this could be changed as well as it's not
+  * unlikely that we'd want workflow outputs to be available as soon as possible.
+  * 
+  * 
+  * In order to do this optimization, we group the scatter gather ports by node they point to
+  * (not the PortBasedGraphOutputNode that is just a proxy, but the source node). If that node is a CallNode
+  * we create only one collector for all those scatter gather ports.
```
cjllanwarne: ToL: I don't think this is actually what happens anymore

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/stores/ExecutionStore.scala
```diff
@@ -41,6 +41,12 @@ object ExecutionStore {
         case _ => key.node.upstreamPorts forall { p => p.executionNode.isInStatus(chooseIndex(p), statusTable) }
       }
     }
+    
+    def nonStartableOutputKeys: Set[JobKey] = key match {
+      case scatterKey: ScatterKey => scatterKey.makeCollectors(0, scatterKey.node.scatterCollectionFunctionBuilder(List.empty)).toSet[JobKey]
```
cjllanwarne: ToL: Do we have to recreate the entire set of collectors every time we run this `def nonStartableOutputKeys`?

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesInitializationActor.scala
```diff
@@ -81,18 +84,21 @@ class JesInitializationActor(jesParams: JesInitializationActorParams)
...
-        case CromwellFatalException(e: StorageException) if e.getCode == StatusCodes.PreconditionFailed.intValue && jesParams.restarting =>
-          workflowLogger.debug(s"Authentication file already exists but this is a restart, proceeding.")
-          Future.successful(())
-        case CromwellFatalException(e: StorageException) if e.getCode == StatusCodes.PreconditionFailed.intValue =>
-            Future.failed(new IOException(s"Failed to upload authentication file:" +
-              s" there was already a file at the same location and this workflow was not being restarted: ${e.getMessage}"))
-        case failure => Future.failed(new IOException(s"Failed to upload authentication file", failure))
-      }
+    def fileUpload(paths: JesWorkflowPaths) = existsAsync(paths.gcsAuthFilePath) flatMap {
+      // if it is then something is definitely wrong
```
cjllanwarne: `// if we aren't restarting then something is definitely wrong`

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesInitializationActor.scala
```diff
@@ -81,18 +84,21 @@ class JesInitializationActor(jesParams: JesInitializationActorParams)
...
+      case true if !jesParams.restarting =>
+        Future.failed(AuthFileAlreadyExistsException(paths.gcsAuthFilePath))
+      case true => 
+        workflowLogger.debug(s"Authentication file already exists but this is a restart, proceeding.")
+        Future.successful(())
```
cjllanwarne: ToL: What if it was partially uploaded? Can we do a sanity check on the contents?

engine/src/main/scala/cromwell/engine/io/gcs/GcsBatchCommandContext.scala
Horneth: This predates the use of coproducts in the Codebase and I'm not going to change all that in this PR at least :) But definitely a TechDebt™️ for later if people care about this.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/stores/ExecutionStore.scala
Horneth: This is really going to *maybe* be called only once per key per workflow anyway. Since it's in an implicit class it has to be a `def`.

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
Horneth: Yep

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesInitializationActor.scala
Horneth: I guess we could, I kind of want to leave that to another ticket though as I'm not trying to fix everything wrong with initialization but to fix what causes travis to fail now.

engine/src/main/scala/cromwell/engine/io/IoActor.scala
Horneth: Because the gcs library already has some error messages that it considers retryable but this one is not one of them

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesInitializationActor.scala
mcovarr: Also I know the cloud is magical and I shouldn't make assumptions about how anything is implemented but I'd be pretty surprised if we had an issue with a partially uploaded auth file.

PR #3077, opened 2017-12-19
CWL Resource Requirements

Horneth:
Validates `cpuMin`, `cpuMax`, `ramMin` and `ramMax` as valid cpu / memory runtime attributes
Maps `cpuMin` and `ramMin` to `cpu` and `memory` (respectively)
All the resource requirements are then available in the runtime attributes map for the backend to play with if it wants to / can.

PR #3078, opened 2017-12-19
InitialWorkDirRequirement for inline content

cjllanwarne:
See [the new centaur cwl](https://github.com/broadinstitute/cromwell/blob/cjl_initial_work_dir_requirement/centaur/src/main/resources/standardTestCases/InitialWorkDirRequirement/inline_file.cwl) for the specific case that this PR enables.

There's plenty more to come for the other scenarios of `InitialWorkDirRequirement`, but one PR at a time...


centaur/src/main/resources/standardTestCases/InitialWorkDirRequirement.inline_file.test
```diff
@@ -0,0 +1,15 @@
...
+
+files {
+  wdl: InitialWorkDirRequirement/inline_file.cwl
+}
+
+metadata {
+  "submittedFiles.workflowType": CWL
+  "submittedFiles.workflowTypeVersion": v1.0
+  # TODO CWL: Re-add this check when outputs are stable:
+  #"outputs.prime_list": "[2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47, 53, 59, 61, 67, 71, 73, 79, 83, 89, 97]"
```
cjllanwarne: If somebody knows how I can add this back please tell me! I hate that this is commented out

Horneth: You can do this to have predictable workflow names https://github.com/broadinstitute/cromwell/pull/3077/files#diff-aba1b9ae7445e5816fa7b4d4b1d2c171

cjllanwarne:
Looks like my JES localization has accidentally thrown off the `write_lines` in WDL. This PR will return...

cjllanwarne:
I think I've fixed the `write_lines` problem now

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
```diff
@@ -149,15 +149,19 @@ class JesAsyncBackendJobExecutionActor(override val standardParams: StandardAsyn
...
   /**
-    * Turns WomFiles into relative paths.  These paths are relative to the working disk
+    * Turns WomFiles into relative paths.  These paths are relative to the working disk.
     *
     * relativeLocalizationPath("foo/bar.txt") -> "foo/bar.txt"
     * relativeLocalizationPath("gs://some/bucket/foo.txt") -> "some/bucket/foo.txt"
+    * relativeLocalizationPath("gs://some/bucket/foo.txt", keepPrefix = false) -> "foo.txt"
     */
-  private def relativeLocalizationPath(file: WomFile): WomFile = {
+  private def relativeLocalizationPath(file: WomFile, keepPrefix: Boolean): WomFile = {
```
Horneth: I don't see this called anywhere with `keepPrefix = false` ?

cwl/src/main/scala/cwl/CwlWomExpression.scala
```diff
@@ -85,12 +90,31 @@ final case class WorkflowStepInputExpression(input: WorkflowStepInput, override
...
+
+  override def evaluateValue(inputValues: Map[String, WomValue], ioFunctionSet: IoFunctionSet): ErrorOr[WomValue] = entry match {
+    case IwdrListingArrayEntry.StringDirent(content, StringOrExpression.String(entryname), _) =>
+      Try(Await.result(ioFunctionSet.writeFile(entryname, content), Duration.Inf)).toErrorOr
+    case _ => ??? // TODO WOM and the rest....
+  }
+
 
+  override def evaluateFiles(inputTypes: Map[String, WomValue], ioFunctionSet: IoFunctionSet, coerceTo: WomType): ErrorOr[Set[WomFile]] =
+    "Programmer error: Shouldn't use InitialWorkDirRequirement listing to find output files. You silly goose.".invalidNel
```
Horneth: ToL: should this not conceptually return the file that this expression aims to create ?

cjllanwarne: `evaluateFiles` is "examine this output expression and tell me what files I'm going to need to have delocalized from JES in order to evaluate it". 

So IMO it's not applicable to the InitialWorkDirRequirement expressions that create `File`s

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
cjllanwarne: Yeah, the parameter is a remnant from before I fixed `write_lines`. I'll remove

cwl/src/main/scala/cwl/CwlWomExpression.scala
Horneth: Ok I've always thought we were also using it to figure out the files that were going to be created prior to running the task (like by `write_lines` for example). Anyway, works for me 👍

cjllanwarne: It's a side effect of PAPI needing to know delocalizations before we start running the job. If/when PAPI2 happens, this entire function will hopefully go away

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -236,8 +239,24 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
...
+      jobDescriptor,
+      backendEngineFunctions,
+      commandLinePreProcessor,
+      commandLineValueMapper,
+      runtimeEnvironment
+    )
+
+    ((adHocFileCreations, instantiatedCommandValidation) mapN { (adHocFiles, command) =>
+        command.copy(createdFiles = command.createdFiles ++ adHocFiles)
+    }).toTry.get
```
danbills: ToL not sure the ripple effect but perhaps acknowledge the `get` here?

cwl/src/main/scala/cwl/CwlWomExpression.scala
danbills: ToL I think we should encode `Input` vs `Output` expressions as types, perhaps in an output extends input manner.  Then we'd only evaluate the ones that have outputs.  I'll add an issue w/ these thoughts.

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala
```diff
@@ -167,17 +167,27 @@ class JesAsyncBackendJobExecutionActor(override val standardParams: StandardAsyn
     // We need to tell PAPI about files that were created as part of command instantiation (these need to be defined
     // as inputs that will be localized down to the VM). Make up 'names' for these files that are just the short
     // md5's of their paths.
-    val writeFunctionFiles = instantiatedCommand.createdFiles map { f => f.value.md5SumShort -> List(f) } toMap
+    val writeFunctionFiles = instantiatedCommand.createdFiles map { f => f.file.value.md5SumShort -> List(f) } toMap
+
+    def localizationPath(f: CommandSetupSideEffectFile) = f.relativeLocalPath match {
+      case Some(l) => WomSingleFile(l)
```
danbills: ToL I've been finding `fold` a nice shortcut for this idiom lately.

cwl/src/main/scala/cwl/InitialWorkDirRequirement.scala
```diff
@@ -0,0 +1,70 @@
...
+}
+
+/**
+  *  Short for "Directory Entry"
+  *  @see <a href="http://www.commonwl.org/v1.0/CommandLineTool.html#Dirent">Dirent Specification</a>
+  *
+  *  Split into two cases because entryName is only optional if entry is an Expression
+  */
+trait Dirent {
+  def writable: Option[Boolean]
```
danbills: consider making this `protected`?  Seems like one would always want `writableWithDefault`.  

Although off the top of my head I don't know how the override w/ case class would work, it might anger the compiler if it made the field public.

cwl/src/main/scala/cwl/InitialWorkDirRequirement.scala
```diff
@@ -0,0 +1,70 @@
...
+                             ) extends Dirent
+
+object InitialWorkDirRequirement {
+
+  final type IwdrListingArrayEntry = File :+: Directory :+: StringDirent :+: ExpressionDirent :+: ExpressionOrString :+: CNil
+
+  object IwdrListingArrayEntry {
+    object StringDirent {
+      def unapply(e: IwdrListingArrayEntry): Option[(String, ExpressionOrString, Boolean)] =
+        e.select[StringDirent].map(sd => (sd.entry, sd.entryname, sd.writableWithDefault))
```
danbills: I see here the effort to use writableWithDefault 👍 ( but I still wonder if it could be enforced via the type system).

PR #3086, opened 2017-12-21
PAPI CWL conformance.

mcovarr:
All the tests that pass on the Local backend currently pass on PAPI with the hackery documented below. This does not attempt to run tests that fail on Local because there's no realistic reason to believe a conformance test that fails Local would succeed on PAPI and the PAPI conformance test run would take hours. On the topic of slowness, the PAPI conformance run will probably have to be converted to a cron job once more conformance tests start passing.

The hacks:
- Cromwell now allows for a default Docker image to be specified in config. This is required for those conformance tests that don't specify a `DockerRequirement`.
- Cromwell allows for a "GCS default input prefix" of any input files when using the JES backend. This allows for the conformance input JSONs to be used as-is. It also works to hack the input JSONs to specify the GCS paths where input files are staged and not bother with the "GCS default input prefix".
- I had to turn off call caching since the hashing actor was looking at the non-GCS paths of files. It's possible this could be worked around if anyone felt it was worth the effort, but I didn't.


centaurCwlRunner/src/main/scala/centaur/cwl/OutputManipulator.scala
```diff
@@ -15,41 +13,28 @@ import io.circe.shapes._
...
-        val file = readAll[IO](fileName.path, 65536)
-        val hash = sha1[IO]
-
-        val bytes = (file through hash).runLog.unsafeRunSync
-
-        //hexify the bytes into a string
-        //credit: https://stackoverflow.com/questions/2756166/what-is-are-the-scala-ways-to-implement-this-java-byte-to-hex-class
-        val hex = bytes.map {
-          b => String.format("%02X", new Integer(b & 0xff))
-        }.mkString.toLowerCase
```
danbills: I'm not sorry to see this go

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesParameters.scala
```diff
@@ -1,15 +1,33 @@
...
   def name: String
   def toGooglePipelineParameter: PipelineParameter
   def toGoogleRunParameter: String
 }
 
+object JesInput {
+  // This is used for CWL conformance testing on PAPI. The current assumption is that the input files have already
+  // been staged up to this path.
+  lazy val defaultGcsPrefix = ConfigFactory.load().as[Option[String]]("papi.default-input-gcs-prefix") map {
+    p => if (p.endsWith("/")) p else p + "/"}
```
Horneth: Can the last `}` be sent to a new line ?

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesParameters.scala
```diff
@@ -1,15 +1,33 @@
...
+
+  private[jes] def prefixIfNecessary(url: String): String = {
+    val prefix = (for {
+      p <- defaultGcsPrefix
+      if !url.startsWith("gs://")
+    } yield p) getOrElse ""
```
Horneth: ToL: I personally find more readable something like

```scala
defaultGcsPrefix match {
  case Some(prefix) if !url.startsWith("gs://") => prefix + url
  case _ => url
}
```

centaurCwlRunner/src/main/scala/centaur/cwl/OutputManipulator.scala
```diff
@@ -61,20 +46,19 @@ object OutputManipulator extends Poly1 {
           schema <- tpe.select[OutputArraySchema]
           items = schema.items
           innerType <- items.select[MyriadOutputInnerType]
-          outputJson = metadata.map(resolveOutputViaInnerType(innerType)).asJson
+          outputJson = metadata.map(m => resolveOutputViaInnerType(innerType)(m, pathBuilder)).asJson
```
Horneth: ToL: If the pathBuilder parameter of `resolveOutputViaInnerType` is moved next to `innerType` instead you could keep it as  `metadata.map(resolveOutputViaInnerType(innerType, pathBuilder)).asJson`

gemlam3:
@danbills @Horneth @kshakir 2 people to re-review please 👍

mcovarr:
I don't think this one needs re-review?

gemlam3:
oops! sorry! I think I confused a different pr

PR #3087, opened 2017-12-21
InitialWorkDirRequirement part 2: StringDirent entryname expressions

cjllanwarne:
Motivating case: [link](https://github.com/broadinstitute/cromwell/blob/cjl_initial_work_dir_requirement_2/centaur/src/main/resources/standardTestCases/InitialWorkDirRequirement/inline_file_custom_entryname.cwl)

Allows the `entryname` to be an expression in `InitialWorkDirRequirement` `StringDirent` entries.

~~EDIT: Will rebase and target `develop` after merging https://github.com/broadinstitute/cromwell/pull/3078~~

PR #3089, opened 2017-12-21
InitialWorkDirRequirement Part 3: Allow String entry expressions for Dirents

cjllanwarne:
Motivating case: [link](https://github.com/broadinstitute/cromwell/blob/cjl_initial_work_dir_requirement_3/centaur/src/main/resources/standardTestCases/InitialWorkDirRequirement/input_string.cwl)

The entry field of a Dirent can now be used to access inputs and evaluate ECMAScript expressions.

- [x] Rebase and target develop after merging https://github.com/broadinstitute/cromwell/pull/3087

### DIFF
--- a/cwl/src/main/scala/cwl/CwlWomExpression.scala
+++ b/cwl/src/main/scala/cwl/CwlWomExpression.scala
@@ -4,6 +4,11 @@ import cats.syntax.option._
 import cats.syntax.validated._
 import common.validation.ErrorOr.ErrorOr
 import common.validation.Validation._
+import common.validation.ErrorOr.ShortCircuitingFlatMap
+import wom.types._
+import wom.values._
+import wom.expression.{IoFunctionSet, WomExpression}
+import cats.syntax.validated._
 import cwl.InitialWorkDirRequirement.IwdrListingArrayEntry
 import cwl.WorkflowStepInput.InputSource
 import cwl.command.ParentName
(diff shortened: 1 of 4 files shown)


cwl/src/main/scala/cwl/CwlWomExpression.scala
```diff
@@ -103,30 +104,39 @@ final case class InitialWorkDirFileGeneratorExpression(entry: IwdrListingArrayEn
...
       } yield writtenFile
-
-
+    case IwdrListingArrayEntry.ExpressionDirent(Expression.ECMAScriptExpression(contentExpression), Some(StringOrExpression.String(entryname)), _) =>
+      val entryEvaluation: WomValue = ExpressionEvaluator.evalExpression(contentExpression, ParameterContext().withInputs(inputValues, ioFunctionSet))
+
+      entryEvaluation match {
+        case f: WomFile => f.validNel
+        case other => for {
+          coerced <- WomStringType.coerceRawValue(other).toErrorOr
```
mcovarr: maybe stick a `collect` in here too instead of `instanceOf`

cwl/src/main/scala/cwl/InitialWorkDirRequirement.scala
```diff
@@ -44,7 +44,7 @@ final case class StringDirent(
 
 object InitialWorkDirRequirement {
 
-  final type IwdrListingArrayEntry = File :+: Directory :+: StringDirent :+: ExpressionDirent :+: StringOrExpression :+: CNil
+  final type IwdrListingArrayEntry = File :+: Directory :+: ExpressionDirent :+: StringDirent :+: StringOrExpression :+: CNil
```
mcovarr: maybe throw a comment in here as to why you did this

cjllanwarne: Sure. FYI, it's because expressions were also getting matched by the string case, so I never saw `Expression` being hit

Issue #2729, opened 2017-10-11
Make WomExpression evaluateValue method asynchronous


katevoss:
@Horneth commented on [Mon Sep 11 2017](https://github.com/broadinstitute/wdl4s/issues/200)

This is a nice to have. `IoFunctionSet` already returns `Future`s for I/O methods. It would be nice to have the evaluate method itself be asynchronous to allow for a fully asynchronous evaluation.
Whether or not WDL evaluation should be made asynchronous too is left to implementor.

katevoss:
@Horneth if this is a nice-to-have then I'll take it out of the Milestone, which is our goal for Q4

Horneth:
If we don't do it now we'll very very unlikely do it later and you might as well close it.

PR #3091, opened 2017-12-22
Use the I/O Actor to execute IoFunctionSet methods

Horneth:
The gist is:

The `ioActor` that is already wired through from the CromwellRoot into the system has been swapped for an `IoActorEndpoint`. This endpoint just forwards existing requests to the IoActor so nothing changes on that side. The only difference is it forwards `IoCommandWithPromise` requests to a proxy actor that will communicate with the IoActor and complete the promise when the answer comes back, thus providing a globally accessible (even outside of an actor) `Future`-like interface that performs operations through the I/O actor.
This is then used in the IoFunctionSet to add retries, throttling and GCS optimizations to it.

### DIFF
--- a/core/src/main/scala/cromwell/core/io/AsyncIo.scala
+++ b/core/src/main/scala/cromwell/core/io/AsyncIo.scala
@@ -1,27 +1,25 @@
 package cromwell.core.io
 
-import akka.actor.{Actor, ActorLogging, ActorRef}
+import akka.actor.ActorRef
+import cromwell.core.io.IoPromiseProxyActor.IoCommandWithPromise
 import cromwell.core.path.BetterFileMethods.OpenOptions
 import cromwell.core.path.Path
 
+import scala.concurrent.Future
 import scala.concurrent.duration._
-import scala.concurrent.{Future, Promise}
-import scala.language.postfixOps
 
+object AsyncIo {
+  val ioTimeout = 3.minutes
+}
 
-trait AsyncIo extends IoClientHelper { this: Actor with ActorLogging with IoCommandBuilder =>
-  
-  protected val ioTimeout = 3 minutes
-  
-  override private [core] def ioResponseReceive: Receive = {
-    case (promise: Promise[_], ack: IoAck[Any] @unchecked) =>
-      cancelTimeout(promise -> ack.command)
-      // This is not typesafe. 
-      // However the sendIoCommand method ensures that the command and the promise have the same generic type
-      // Which means as long as only the sendIoCommand method is used to send requests, and the ioActor honors his contract
-      // and send back the right context with the right response, the types are virtually guaranteed to match.
-      promise.asInstanceOf[Promise[Any]].complete(ack.toTry)
-      ()
+/**
+  * Provides Futurized methods for I/O actions processed through the IoActor
+  */
+class AsyncIo(ioEndpoint: ActorRef, ioCommandBuilder: IoCommandBuilder) {
+  private def asyncCommand[A](command: IoCommand[A], timeout: FiniteDuration = AsyncIo.ioTimeout) = {
+    val commandWithPromise = IoCommandWithPromise(command, timeout)
+    ioEndpoint ! commandWithPromise
+    commandWithPromise.promise.future
   }
   
   /**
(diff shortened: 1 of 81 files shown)


geoffjentry:
What's the genesis of the `Endpoint` terminology?

Horneth:
It's technically not what we call the IoActor anymore but instead _some_ actor that will route I/O related commands to either the IoActor or its proxy, so I came up with "endpoint" but I'm totally open to other naming suggestions !

core/src/main/scala/cromwell/core/io/IoCommandBuilder.scala
```diff
@@ -4,24 +4,89 @@ import cromwell.core.io.DefaultIoCommand._
...
+abstract class PartialIoCommandBuilder {
+  def contentAsStringCommand: PartialFunction[Path, IoContentAsStringCommand] = PartialFunction.empty
+  def writeCommand: PartialFunction[(Path, String, OpenOptions), IoWriteCommand] = PartialFunction.empty
+  def sizeCommand: PartialFunction[Path, IoSizeCommand] = PartialFunction.empty
+  def deleteCommand: PartialFunction[(Path, Boolean), IoDeleteCommand] = PartialFunction.empty
+  def copyCommand: PartialFunction[(Path, Path, Boolean), IoCopyCommand] = PartialFunction.empty
+  def hashCommand: PartialFunction[Path, IoHashCommand] = PartialFunction.empty
+  def touchCommand: PartialFunction[Path, IoTouchCommand] = PartialFunction.empty
+  def existsCommand: PartialFunction[Path, IoExistsCommand] = PartialFunction.empty
+  def readLinesCommand: PartialFunction[Path, IoReadLinesCommand] = PartialFunction.empty
```
mcovarr: TIL `PartialFunction.empty` 😄

cwl/src/main/scala/cwl/CwlWomExpression.scala
```diff
@@ -63,7 +63,7 @@ case class CommandOutputExpression(outputBinding: CommandOutputBinding,
 
         //In the case of a single file being expected, we must enforce that the glob only represents a single file
         case (WomString(glob), WomSingleFileType) =>
-          ioFunctionSet.glob(glob) match {
+          Await.result(ioFunctionSet.glob(glob), Duration.Inf) match {
```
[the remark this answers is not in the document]
Horneth: Yeah if we had [that](https://github.com/broadinstitute/cromwell/issues/2729) we wouldn't need the `Await.result` but...

core/src/main/scala/cromwell/core/io/AsyncIo.scala
```diff
@@ -1,84 +1,61 @@
...
 import cromwell.core.path.BetterFileMethods.OpenOptions
 import cromwell.core.path.Path
 
+import scala.concurrent.Future
 import scala.concurrent.duration._
-import scala.concurrent.{Future, Promise}
-import scala.language.postfixOps
 
+object AsyncIo {
+  val ioTimeout = 3.minutes
```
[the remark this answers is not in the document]
Horneth: I vaguely remember someone saying "if someone complains we'll make a knob" but I'm not sure. I'm sure @geoffjentry will have an opinion on this. I'm happy to add a knob in case someone wants it though.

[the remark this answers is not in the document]
geoffjentry: I caught myself in a contradiction lately where I was talking up all of the configuration options.

I guess the guiding principle should be "does it seem reasonable that someone is going to want to fiddle with this" instead of simply "OMG, it's a hardcoded value! knob it!"

geoffjentry:
@Horneth my concern on `endpoint` is its overloaded with the REST API endpoints. Realizing that it's also not an *akka* router, it seems to be serving a similar role. Proxy also seems viable (granted you're using that term in your description, but not in the code)

filesystems/gcs/src/main/scala/cromwell/filesystems/gcs/batch/GcsBatchCommandBuilder.scala
```diff
@@ -1,37 +1,32 @@
...
   
-  override def copyCommand(src: Path, dest: Path, overwrite: Boolean = true) =  (src, dest) match {
-    case (gcsSrc: GcsPath, gcsDest: GcsPath) => GcsBatchCopyCommand(gcsSrc, gcsDest, overwrite)
-    case _ => super.copyCommand(src, dest, overwrite)
+  override def copyCommand = {
+    case (gcsSrc: GcsPath, gcsDest: GcsPath, overwrite) => GcsBatchCopyCommand(gcsSrc, gcsDest, overwrite)
   }
   
-  override def hashCommand(path: Path) =  path match {
+  override def hashCommand = {
```
danbills: ToL so these will throw exceptions whenever a GcsPath is not being passed in?

core/src/main/scala/cromwell/core/io/AsyncIoActorClient.scala
```diff
@@ -0,0 +1,43 @@
...
+import akka.actor.ActorRef
+import cromwell.core.path.BetterFileMethods.OpenOptions
+import cromwell.core.path.Path
+
+import scala.concurrent.Future
+
+/**
+  * Helper trait to used async Io method easily
+  */
+trait AsyncIoActorClient {
```
danbills: ToL Consider making `asyncIO` public, thus avoiding the need to write the wrapper methods.

filesystems/gcs/src/main/scala/cromwell/filesystems/gcs/batch/GcsBatchCommandBuilder.scala
Horneth: No they're partial functions and they're used if defined for the given path, otherwise the `IoCommandBuilder` defaults to the default command.

PR #3098, opened 2018-01-03
proof of concept metadata implementation which publishes to google pu…

geoffjentry:
…bsub

### DIFF
--- a/cromwell.examples.conf
+++ b/cromwell.examples.conf
@@ -622,22 +622,36 @@ backend {
 services {
   MetadataService {
     config {
-      # Set this value to "Inf" to turn off metadata summary refresh.  The default value is currently "2 seconds".
-      # metadata-summary-refresh-interval = "Inf"
-      # For higher scale environments, e.g. many workflows and/or jobs, DB write performance for metadata events
-      # can improved by writing to the database in batches. Increasing this value can dramatically improve overall
-      # performance but will both lead to a higher memory usage as well as increase the risk that metadata events
-      # might not have been persisted in the event of a Cromwell crash.
-      #
-      # For normal usage the default value of 200 should be fine but for larger/production environments we recommend a
-      # value of at least 500. There'll be no one size fits all number here so we recommend benchmarking performance and
-      # tuning the value to match your environment.
-      # db-batch-size = 200
-      #
-      # Periodically the stored metadata events will be forcibly written to the DB regardless of if the batch size
-      # has been reached. This is to prevent situations where events wind up never being written to an incomplete batch
-      # with no new events being generated. The default value is currently 5 seconds
-      # db-flush-rate = 5 seconds
+      # For the standard MetadataService implementation, cromwell.services.metadata.impl.MetadataServiceActor:
+        # Set this value to "Inf" to turn off metadata summary refresh.  The default value is currently "2 seconds".
+        # metadata-summary-refresh-interval = "Inf"
+        # For higher scale environments, e.g. many workflows and/or jobs, DB write performance for metadata events
+        # can improved by writing to the database in batches. Increasing this value can dramatically improve overall
+        # performance but will both lead to a higher memory usage as well as increase the risk that metadata events
+        # might not have been persisted in the event of a Cromwell crash.
+        #
+        # For normal usage the default value of 200 should be fine but for larger/production environments we recommend a
+        # value of at least 500. There'll be no one size fits all number here so we recommend benchmarking performance and
+        # tuning the value to match your environment.
+        # db-batch-size = 200
+        #
+        # Periodically the stored metadata events will be forcibly written to the DB regardless of if the batch size
+        # has been reached. This is to prevent situations where events wind up never being written to an incomplete batch
+        # with no new events being generated. The default value is currently 5 seconds
+        # db-flush-rate = 5 seconds
+
+      # For the Google PubSub MetadataService implementation: cromwell.services.metadata.impl.pubsub.PubSubMetadataServiceActor:
+        # Google project
+        # project = "my-project"
+        # The auth *must* be a service-account auth with PEM file configured (not JSON).
+        # auth = "service-account"
+        # The PubSub topic to write to. Will be created if it doesn't already exist. Defaults to "cromwell-metadata"
+        # topic = "cromwell-metadata"
+        # An optional PubSub subscription name. If supplied and if it doesn't already exist, it will be created and
+        # subscribed to the topic
+        # subscription = "optional-subscription"
+        # An application name to set on your PubSub interaction.
+        # appName = "cromwell"
     }
   }
 
--- a/cromwell.examples.conf
+++ b/cromwell.examples.conf
@@ -623,35 +623,35 @@ services {
   MetadataService {
     config {
       # For the standard MetadataService implementation, cromwell.services.metadata.impl.MetadataServiceActor:
-        # Set this value to "Inf" to turn off metadata summary refresh.  The default value is currently "2 seconds".
-        # metadata-summary-refresh-interval = "Inf"
-        # For higher scale environments, e.g. many workflows and/or jobs, DB write performance for metadata events
-        # can improved by writing to the database in batches. Increasing this value can dramatically improve overall
-        # performance but will both lead to a higher memory usage as well as increase the risk that metadata events
-        # might not have been persisted in the event of a Cromwell crash.
-        #
-        # For normal usage the default value of 200 should be fine but for larger/production environments we recommend a
-        # value of at least 500. There'll be no one size fits all number here so we recommend benchmarking performance and
-        # tuning the value to match your environment.
-        # db-batch-size = 200
-        #
-        # Periodically the stored metadata events will be forcibly written to the DB regardless of if the batch size
-        # has been reached. This is to prevent situations where events wind up never being written to an incomplete batch
-        # with no new events being generated. The default value is currently 5 seconds
-        # db-flush-rate = 5 seconds
+      #   Set this value to "Inf" to turn off metadata summary refresh.  The default value is currently "2 seconds".
+      #   metadata-summary-refresh-interval = "Inf"
+      #   For higher scale environments, e.g. many workflows and/or jobs, DB write performance for metadata events
+      #   can improved by writing to the database in batches. Increasing this value can dramatically improve overall
+      #   performance but will both lead to a higher memory usage as well as increase the risk that metadata events
+      #   might not have been persisted in the event of a Cromwell crash.
+      #
+      #   For normal usage the default value of 200 should be fine but for larger/production environments we recommend a
+      #   value of at least 500. There'll be no one size fits all number here so we recommend benchmarking performance and
+      #   tuning the value to match your environment.
+      #   db-batch-size = 200
+      #
+      #   Periodically the stored metadata events will be forcibly written to the DB regardless of if the batch size
+      #   has been reached. This is to prevent situations where events wind up never being written to an incomplete batch
+      #   with no new events being generated. The default value is currently 5 seconds
+      #   db-flush-rate = 5 seconds
 
       # For the Google PubSub MetadataService implementation: cromwell.services.metadata.impl.pubsub.PubSubMetadataServiceActor:
-        # Google project
-        # project = "my-project"
-        # The auth *must* be a service-account auth with PEM file configured (not JSON).
-        # auth = "service-account"
-        # The PubSub topic to write to. Will be created if it doesn't already exist. Defaults to "cromwell-metadata"
-        # topic = "cromwell-metadata"
-        # An optional PubSub subscription name. If supplied and if it doesn't already exist, it will be created and
-        # subscribed to the topic
-        # subscription = "optional-subscription"
-        # An application name to set on your PubSub interaction.
-        # appName = "cromwell"
+      #   Google project
+      #   project = "my-project"
+      #   The auth *must* be a service-account auth with PEM file configured (not JSON).
+      #   auth = "service-account"
+      #   The PubSub topic to write to. Will be created if it doesn't already exist. Defaults to "cromwell-metadata"
+      #   topic = "cromwell-metadata"
+      #   An optional PubSub subscription name. If supplied and if it doesn't already exist, it will be created and
+      #   subscribed to the topic
+      #   subscription = "optional-subscription"
+      #   An application name to set on your PubSub interaction.
+      #   appName = "cromwell"
     }
   }
 
--- a/cromwell.examples.conf
+++ b/cromwell.examples.conf
@@ -643,7 +643,7 @@ services {
       # For the Google PubSub MetadataService implementation: cromwell.services.metadata.impl.pubsub.PubSubMetadataServiceActor:
       #   Google project
       #   project = "my-project"
-      #   The auth *must* be a service-account auth with PEM file configured (not JSON).
+      #   The auth *must* be a service-account auth with JSON auth.
       #   auth = "service-account"
       #   The PubSub topic to write to. Will be created if it doesn't already exist. Defaults to "cromwell-metadata"
       #   topic = "cromwell-metadata"
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -3,6 +3,9 @@ import sbt._
 object Dependencies {
   val akkaHttpV = "10.0.10"
   val akkaV = "2.5.4"
+  val alibabaCloudCoreV = "3.2.3"
+  val alibabaCloudOssV = "2.8.3"
+  val alibabaCloudBcsV = "5.1.0"
   val ammoniteOpsV = "1.0.1"
   val apacheHttpClientV = "4.5.3"
   val apacheHttpCoreV = "4.4.6"
@@ -75,10 +78,7 @@ object Dependencies {
   val swaggerParserV = "1.0.22"
   val swaggerUiV = "3.2.2"
   val typesafeConfigV = "1.3.1"
-  val alibabaCloudCoreV = "3.2.3"
-  val alibabaCloudOssV = "2.8.3"
-  val alibabaCloudBcsV = "5.1.0"
-
+  val workbenchGoogleV = "0.15-2fc79a3"
 
   /*
   If you see warnings from SBT about evictions, insert a specific dependency version into this list.
@@ -207,7 +207,7 @@ object Dependencies {
       exclude("com.google.cloud.datastore", "datastore-v1-protos")
       exclude("org.apache.httpcomponents", "httpclient"),
     "com.google.cloud" % "google-cloud-compute" % googleCloudComputeV,
-    "org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.10-0967a99"
+    "org.broadinstitute.dsde.workbench" %% "workbench-google" % workbenchGoogleV
       exclude("com.google.apis", "google-api-services-genomics"),
     "org.apache.httpcomponents" % "httpclient" % apacheHttpClientV
   )
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -207,6 +207,8 @@ object Dependencies {
       exclude("com.google.cloud.datastore", "datastore-v1-protos")
       exclude("org.apache.httpcomponents", "httpclient"),
     "com.google.cloud" % "google-cloud-compute" % googleCloudComputeV,
+    "org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.10-0967a99"
+      exclude("com.google.apis", "google-api-services-genomics"),
     "org.apache.httpcomponents" % "httpclient" % apacheHttpClientV
   )
 
--- a/services/src/main/scala/cromwell/services/metadata/MetadataJsonSupport.scala
+++ b/services/src/main/scala/cromwell/services/metadata/MetadataJsonSupport.scala
@@ -1,0 +1,28 @@
+package cromwell.services.metadata
+
+import java.time.OffsetDateTime
+
+import cromwell.core.WorkflowId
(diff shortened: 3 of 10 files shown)


kcibul:
Very cool to see!

cromwell.examples.conf
```diff
@@ -573,22 +573,36 @@ backend {
...
-      #
-      # Periodically the stored metadata events will be forcibly written to the DB regardless of if the batch size
-      # has been reached. This is to prevent situations where events wind up never being written to an incomplete batch
-      # with no new events being generated. The default value is currently 5 seconds
-      # db-flush-rate = 5 seconds
+      # For the standard MetadataService implementation, cromwell.services.metadata.impl.MetadataServiceActor:
+        # Set this value to "Inf" to turn off metadata summary refresh.  The default value is currently "2 seconds".
+        # metadata-summary-refresh-interval = "Inf"
+        # For higher scale environments, e.g. many workflows and/or jobs, DB write performance for metadata events
+        # can improved by writing to the database in batches. Increasing this value can dramatically improve overall
```
mcovarr: this sentence can improved too

services/src/main/scala/cromwell/services/metadata/impl/pubsub/PubSubMetadataServiceActor.scala
```diff
@@ -0,0 +1,132 @@
...
+  /**
+    * Creates a subscription if pubsubTopicName is defined. Note that we're ignoring the bool in the Future[bool]
+    * return type of HttpGooglePubSubDAO.createSubscription and only caring about the success of the Future. It is not
+    * considered an error if the subscription already exists (in fact, this is a likely circumstance).
+    */
+  private def createSubscription(): Future[Unit] = {
+    pubSubSubscriptionName match {
+      case Some(name) =>
+        log.info("Creating subscription " + name)
+        pubSubConnection.createSubscription(pubSubTopicName, name) map { _ => () }
```
mcovarr: `void`

services/src/main/scala/cromwell/services/metadata/impl/pubsub/PubSubMetadataServiceActor.scala
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  /**
+    * Creates the specified topic, and optionally the subscription if it was defined. We are ignoring the bool return
+    * type from the HttpGooglePubSubDAO return types as it is not considered an error if the topic/subscription already
+    * exist (in fact, this is a likely circumstance).
+    */
+  private def createTopicAndSubscription(): Future[Unit] = {
+    log.info("Ensuring topic " + pubSubTopicName + " exists")
+    pubSubConnection.createTopic(pubSubTopicName) map { _ => createSubscription() } map { _ => () }
```
mcovarr: `as`/`void`

services/src/main/scala/cromwell/services/metadata/impl/pubsub/PubSubMetadataServiceActor.scala
geoffjentry: @mcovarr lol I knew you had some little thing on these but I couldn't remember, so I figured your eagle eyes would spot it here

cromwell.examples.conf
geoffjentry: So help me, I better not do a blame mode and find out you were the original author of this sentence :P

geoffjentry: Damn, *I* was the original author

geoffjentry:
@kcibul now with a hybrid metadata impl as well

cromwell.examples.conf
geoffjentry: @mcovarr however, since you **did** directly comment on this exact sentence the first time it was created, I'll let you describe your evolving thoughts on sentence structure :)

services/src/main/scala/cromwell/services/metadata/impl/pubsub/PubSubMetadataServiceActor.scala
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  /**
+    * Creates the specified topic, and optionally the subscription if it was defined. We are ignoring the bool return
+    * type from the HttpGooglePubSubDAO return types as it is not considered an error if the topic/subscription already
+    * exist (in fact, this is a likely circumstance).
+    */
+  private def createTopicAndSubscription(): Future[Unit] = {
+    log.info("Ensuring topic " + pubSubTopicName + " exists")
+    pubSubConnection.createTopic(pubSubTopicName).flatMap(_ => createSubscription()).void
```
geoffjentry: @mcovarr you had asked for an `as` here on top of the `void`, what did you mean?

mcovarr: ignore me, I missed the fact that's a `flatMap` and not a `map`.

geoffjentry: oh duh. it *was* a `map` but when i added the void I realized i had a `Future[Future[Unit]]` and it was ok because of the `Future[Unit]` return type

services/src/main/scala/cromwell/services/metadata/impl/hybridpubsub/HybridPubSubMetadataServiceActor.scala
```diff
@@ -0,0 +1,35 @@
...
+  *
+  * Under the hood it maintains its own MetadataServiceActor and PubSubMetadataServiceActor. All messages are routed
+  * to the MetadataServiceActor. PutMetadataActions are also sent to the PubSubMetadataServiceActor. PutMetadataActionAndRespond
+  * messages will be sent to the PubSubMetadataServiceActor as a standard PutMetadataAction, i.e. only the standard
+  * metadata service will be ACKing the request.
+  *
+  * It is expected that the user will supply any desired config fields for both MetadataServiceActor and
+  * PubSubMetadataServiceActor as the serviceConfig block will be passed along to both of them.
+  */
+class HybridPubSubMetadataServiceActor(serviceConfig: Config, globalConfig: Config) extends Actor with ActorLogging {
```
danbills: ToL are there dispatchers for this kind of thing?  i.e. a broadcast one?

services/src/main/scala/cromwell/services/metadata/impl/pubsub/PubSubMetadataServiceActor.scala
```diff
@@ -0,0 +1,133 @@
...
+  *
+  * For now it requires a google service account auth with a PEM file. If the underlying PubSub library is updated to
+  * allow for other forms of auth this becomes viable for us. Note that currently Cromwell complains that PEM is
+  * deprecated, so we might need to contribute a patch to workbench-lib's pubsub tooling.
+  *
+  * This is intentionally not documented, considering there exists no way to consume/read metadata if this is active
+  * anyone using this is by definition more advanced than the Cromwell developers
+  *
+  * Potential TODOs to consider if moving this forward from proof-of-concept:
+  *  - This will probably need some measure of batching, need to see it in action before proceeding down that path
```
danbills: Doesn't Google's client have batching?

services/src/main/scala/cromwell/services/metadata/impl/pubsub/PubSubMetadataServiceActor.scala
```diff
@@ -0,0 +1,133 @@
...
+  *  - Graceful shutdown could also be useful. The pubsub calls will autoretry in the face of transient errors which
```
danbills: as above, I would hope a retry is configurable in the client.

services/src/main/scala/cromwell/services/metadata/impl/pubsub/PubSubMetadataServiceActor.scala
```diff
@@ -0,0 +1,133 @@
...
+  *     means there might be stuff hanging around on a shutdown request.
+  *  - Currently service actors which fail to instantiate properly don't bring down Cromwell. It's easier to screw
+  *     it up here, we might want to revisit that strategy
+  *  - Revisit the Alpakka connector instead of the workbench-lib version. There was a bug they fixed but hadn't released
+  *     it yet. Also I'm pretty sure that same bug was repeated in a few other places.
+  */
+class PubSubMetadataServiceActor(serviceConfig: Config, globalConfig: Config) extends Actor with ActorLogging {
```
danbills: ToL In general it'd be nice to have specific types injected here rather than the Config sledgehammer.  Puts the onus of validation, `getOrElses` and failure decisions outside of the scope of this code.

services/src/main/scala/cromwell/services/metadata/impl/pubsub/PubSubMetadataServiceActor.scala
```diff
@@ -0,0 +1,133 @@
...
+
+  val googleProject = serviceConfig.as[String]("project")
+  val pubSubTopicName = serviceConfig.as[Option[String]]("topic").getOrElse("cromwell-metadata")
+  val pubSubSubscriptionName = serviceConfig.as[Option[String]]("subscription")
+  val pubSubAppName = serviceConfig.as[Option[String]]("appName").getOrElse("cromwell")
+
+  val pubSubConnection = createPubSubConnection()
+  createTopicAndSubscription().failed foreach { e => throw e }
+
+  override def receive = {
```
danbills: ToL IMO this actor should be limited to this `receive` and `publishMessages`

services/src/main/scala/cromwell/services/metadata/impl/pubsub/PubSubMetadataServiceActor.scala
```diff
@@ -0,0 +1,133 @@
...
+  // The auth *must* be a service account auth but it might not be named service-account.
+  val googleAuthName = serviceConfig.as[Option[String]]("auth").getOrElse("service-account")
+
+  val googleProject = serviceConfig.as[String]("project")
+  val pubSubTopicName = serviceConfig.as[Option[String]]("topic").getOrElse("cromwell-metadata")
+  val pubSubSubscriptionName = serviceConfig.as[Option[String]]("subscription")
+  val pubSubAppName = serviceConfig.as[Option[String]]("appName").getOrElse("cromwell")
+
+  val pubSubConnection = createPubSubConnection()
+  createTopicAndSubscription().failed foreach { e => throw e }
```
danbills: Sorry to beat this horse but this side effecty stuff should be done elsewhere

services/src/main/scala/cromwell/services/metadata/impl/hybridpubsub/HybridPubSubMetadataServiceActor.scala
geoffjentry: This isn't blindly throwing **everything** to both actors, not sure what you're looking for here.

services/src/main/scala/cromwell/services/metadata/impl/pubsub/PubSubMetadataServiceActor.scala
```diff
@@ -0,0 +1,133 @@
...
+      case Invalid(e) => throw new IllegalArgumentException("Unable to configure PubSubMetadataServiceActor: " + e.toList.mkString(", "))
+    }
+
+    val pemFile = googleAuth.fileFormat match {
+      case p: PemFileFormat => p
+      case _ => throw new IllegalArgumentException("Unable to configure PubSubMetadataServiceActor: the service account must supply a PEM file")
+    }
+
+    // The metric name is pretty useless. It's an artifact of the workbench libraries which we're not using
+    new HttpGooglePubSubDAO(pemFile.accountId, pemFile.file, pubSubAppName, googleProject, "cromwell.pubsub")
```
kshakir: _PEM files begone!_ Can this be switched to a JSON instead?

services/src/main/scala/cromwell/services/metadata/impl/pubsub/PubSubMetadataServiceActor.scala
geoffjentry: This is using the client built in `workbench-lib` which has batching. Batching isn't cut & dry in pubsub and has some nuance. This comment is saying that for now that nuance is being punted on.

services/src/main/scala/cromwell/services/metadata/impl/pubsub/PubSubMetadataServiceActor.scala
geoffjentry: The library being used autoretries on specific situations. However those situations are the same across the entirety of workbench pubsub usage and we also can open a PR at any point :)

services/src/test/scala/cromwell/services/metadata/impl/pubsub/PubSubMetadataServiceActorSpec.scala
```diff
@@ -0,0 +1,190 @@
...
+
+  class FailToPublishMockGooglePubSubDao extends SuccessfulMockGooglePubSubDao {
+    override def publishMessages(topicName: String, messages: Seq[String]): Future[Unit] = Future.failed(new RuntimeException("sorry charlie"))
+  }
+
+  // This doesn't include a project so should be a failure
+  val emptyConfig = ConfigFactory.parseString(
+    """
+      |
+    """.stripMargin
```
kshakir: A little verbose for what could be `""` (or in this case `"\n\n  \n"`)

services/src/main/scala/cromwell/services/metadata/impl/pubsub/PubSubMetadataServiceActor.scala
geoffjentry: I believe we've had this convo before. This is part of an implicit interface for the service registry via reflective loading and needs to be this way. I'm not opposed to that changing, but that needs to be changed across the whole service registry - which also theoretically has an impact on 3rd parties

services/src/main/scala/cromwell/services/metadata/impl/pubsub/PubSubMetadataServiceActor.scala
geoffjentry: Where? 

This kills the actor which bubbles up the supervision chain and ultimately stops Cromwell from loading.

services/src/main/scala/cromwell/services/metadata/impl/pubsub/PubSubMetadataServiceActor.scala
geoffjentry: To make sure I understand, this ultimately is tied to the `Config` commentary, right?

services/src/main/scala/cromwell/services/metadata/impl/pubsub/PubSubMetadataServiceActor.scala
```diff
@@ -0,0 +1,133 @@
...
+  *  - Currently service actors which fail to instantiate properly don't bring down Cromwell. It's easier to screw
+  *     it up here, we might want to revisit that strategy
+  *  - Revisit the Alpakka connector instead of the workbench-lib version. There was a bug they fixed but hadn't released
+  *     it yet. Also I'm pretty sure that same bug was repeated in a few other places.
+  */
+class PubSubMetadataServiceActor(serviceConfig: Config, globalConfig: Config) extends Actor with ActorLogging {
+  implicit val ec = context.dispatcher
+
+  // The auth *must* be a service account auth but it might not be named service-account.
+  val googleAuthName = serviceConfig.as[Option[String]]("auth").getOrElse("service-account")
```
kshakir: Consider the less verbose `getOrElse` instead of `as[Option[A]]`. Ex:

```scala
val googleAuthName = serviceConfig.getOrElse("auth", "service-account")
```

Same for others below.

cromwell.examples.conf
```diff
@@ -573,22 +573,36 @@ backend {
...
-      # value of at least 500. There'll be no one size fits all number here so we recommend benchmarking performance and
-      # tuning the value to match your environment.
-      # db-batch-size = 200
-      #
-      # Periodically the stored metadata events will be forcibly written to the DB regardless of if the batch size
-      # has been reached. This is to prevent situations where events wind up never being written to an incomplete batch
-      # with no new events being generated. The default value is currently 5 seconds
-      # db-flush-rate = 5 seconds
+      # For the standard MetadataService implementation, cromwell.services.metadata.impl.MetadataServiceActor:
+        # Set this value to "Inf" to turn off metadata summary refresh.  The default value is currently "2 seconds".
```
kshakir:
```hocon
# Don't indent  
    # like this.
```

Line up the hash marks, or IntelliJ will do it for the next editor who uses the pretty-formatter.

```hocon
# Can still indent
#     like this.
```

project/Dependencies.scala
```diff
@@ -203,6 +203,8 @@ object Dependencies {
       exclude("com.google.cloud.datastore", "datastore-v1-protos")
       exclude("org.apache.httpcomponents", "httpclient"),
     "com.google.cloud" % "google-cloud-compute" % googleCloudComputeV,
+    "org.broadinstitute.dsde.workbench" %% "workbench-google" % "0.10-0967a99"
```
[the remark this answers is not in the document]
geoffjentry: Yeah, I forgot about that in the interim. the upside is this has hung around so long that I no longer need to make a PR to workbench-libs like I Said I would.

re the constant: it's only used in one place, IMO string constants should only be used when n > 1

services/src/main/scala/cromwell/services/metadata/MetadataJsonSupport.scala
```diff
@@ -0,0 +1,28 @@
...
+
+import java.time.OffsetDateTime
+
+import cromwell.core.WorkflowId
+import spray.json.{DefaultJsonProtocol, JsString, JsValue, RootJsonFormat}
+
+object MetadataJsonSupport extends DefaultJsonProtocol {
+  implicit object WorkflowIdJsonFormatter extends RootJsonFormat[WorkflowId] {
+    def write(id: WorkflowId) = JsString(id.id.toString)
+    def read(value: JsValue) = throw new UnsupportedOperationException("Reading WorkflowId from JSON is currently unsupported")
```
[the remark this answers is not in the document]
geoffjentry: This was a YAGNI situation, but yeah

services/src/main/scala/cromwell/services/metadata/impl/pubsub/PubSubMetadataServiceActor.scala
geoffjentry: TIL

services/src/main/scala/cromwell/services/metadata/impl/pubsub/PubSubMetadataServiceActor.scala
geoffjentry: Yeah. If it helps, me doing this project and then whining to them is what led to them getting into the 21st century

services/src/test/scala/cromwell/services/metadata/impl/pubsub/PubSubMetadataServiceActorSpec.scala
geoffjentry: Or apparently `ConfigFactory.empty()`

project/Dependencies.scala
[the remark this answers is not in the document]
kshakir: Yeah, I get you on constants when n == 1. Still a trip through git history shows our habit of copy-pasting and then _not_ creating the constant when going to n == 2, n == 3, etc. The worst is when different versions of the same library to two different sub-projects.

Also having the constants in one block, alphabetically, makes it easier to update when running `sbt dependencyUpdates`. I suspect with our large number of libraries we'll have at least some patching to do soon enough after security audits, etc. (ask GOTC about their recent cleanups)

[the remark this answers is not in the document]
geoffjentry: My answer to your first point about poor history is to fix that poor behavior ;)

However you won me over with your second paragraph. Particularly considering I've had [this](https://github.com/rtimush/sbt-updates) open as a tab in my browser for a while because I get so excited about dependency updates :)

services/src/main/scala/cromwell/services/metadata/impl/pubsub/PubSubMetadataServiceActor.scala
geoffjentry: Although today that doesn't seem to be the case, need to figure that out

services/src/main/scala/cromwell/services/metadata/impl/hybridpubsub/HybridPubSubMetadataServiceActor.scala
danbills: Oh I didn't notice that anythingElse is omitting the pubSubMetadataActor 🤦‍♂️

services/src/main/scala/cromwell/services/metadata/impl/pubsub/PubSubMetadataServiceActor.scala
danbills: This comment probably belonged in a larger comment on the PR.

I do still think there is a valid concern about a couple things. They are subjective, I'm fine with the current pattern, and thus my ToL tag, but here they are :

* the configuration items e.g. the String "project", "topic", etc. could be passed in even though you are receiving the Config which contains them.  This moves the logic for key names, types, getOrElse situations, and failures out.  
* Same with the side-effecting `createTopicAndSubscription`

geoffjentry: That’s my point, it can’t. Not without sticking another actor in front of this which has just he configs in the constructor and passes everything on to this one, which doesn’t seem great.

The service registry requires you have an actor with this exact signature

danbills: I got it, I didn't understand the constraints.

PR #3099, opened 2018-01-03
bump sbt version as per project/build.properties

mr-c:


docs/Building.md
```diff
@@ -4,7 +4,7 @@ If for some reason you require a non-release version of Cromwell or are developi
 features or fixes, the following are required to build Cromwell from source:
 
 * [Scala 2.12.2](http://www.scala-lang.org/news/2.12.1#scala-212-notes)
```
mcovarr: In a similar vein there are two different patch level versions of Scala called out here but neither is the 2.12.4 version we're actually using. 😢

docs/Building.md
```diff
@@ -4,7 +4,7 @@ If for some reason you require a non-release version of Cromwell or are developi
...
-* [SBT 0.13.12](https://github.com/sbt/sbt/releases/tag/v0.13.12)
+* [SBT 0.13.16](https://github.com/sbt/sbt/releases/tag/v0.13.16)
 * [Java 8](http://www.oracle.com/technetwork/java/javase/overview/java8-2100321.html)
```
mcovarr: We may also want to add a note that old versions of Java 8 are known to have bugs that affect Cromwell usage and encourage users to get the most recent Java 8 version possible. One of those old Java 8 issues was a bug parsing Javascript lambdas that is likely to affect CWL users.

PR #3100, opened 2018-01-03
SGE cluster configuration correction in documentation

rhpvorderman:
`actor-factory` should be defined within `SGE` not within SGE.config


geoffjentry:
:+1:

PR #3106, opened 2018-01-05
IWDR: Allow Array[File] expressions as listings

cjllanwarne:
See [here](https://github.com/broadinstitute/cromwell/blob/cjl_initial_work_dir_requirement_6/centaur/src/main/resources/standardTestCases/InitialWorkDirRequirement/file_array_listing.cwl#L10) for the motivating use case. 

This needs improving as a centaur tests because we need to generate the file so that it can be passed in (which needs a WF rather than a simple CLT).

### DIFF
--- a/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
+++ b/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
@@ -8,7 +8,6 @@ import cats.syntax.traverse._
 import cats.syntax.validated._
 import akka.actor.{Actor, ActorLogging, ActorRef}
 import akka.event.LoggingReceive
-import cats.data.Validated.Valid
 import common.exception.MessageAggregation
 import common.util.TryUtil
 import common.validation.ErrorOr.ErrorOr
@@ -249,14 +248,15 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
 
     val adHocFileCreationInputs = jobDescriptor.evaluatedTaskInputs.map { case (k,v) => k.localName.value -> v }
 
-    def validateAdHocFile(value: WomValue): ErrorOr[WomFile] = value match {
-        case f: WomFile => Valid(f)
+    def validateAdHocFile(value: WomValue): ErrorOr[List[WomFile]] = value match {
+        case f: WomFile => List(f).valid
+        case a: WomArray => a.value.toList.traverse(validateAdHocFile).map(_.flatten)
         case other => s"Ad-hoc file creation expression invalidly created a ${other.womType.toDisplayString} result.".invalidNel
     }
 
     val adHocFileCreations: ErrorOr[List[WomFile]] = jobDescriptor.taskCall.callable.adHocFileCreation.toList.traverse {
       _.evaluateValue(adHocFileCreationInputs, backendEngineFunctions).flatMap(validateAdHocFile)
-    }
+    }.map(_.flatten)
 
     val adHocFileCreationSideEffectFiles: ErrorOr[List[CommandSetupSideEffectFile]] = adHocFileCreations map { _ map {
       f => CommandSetupSideEffectFile(f,  Option(adHocFileLocalization(f)))
@@ -274,7 +274,10 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
     // TODO CWL: toTry.get here. Is throwing an exception the best way to indicate command generation failure?
     ((adHocFileCreationSideEffectFiles, instantiatedCommandValidation) mapN { (adHocFiles, command) =>
         command.copy(createdFiles = command.createdFiles ++ adHocFiles)
-    }).toTry.get
+    }).toTry match {
+      case Success(ic) => ic
+      case Failure(e) => throw new Exception("Failed to evaluate ad hoc files", e)
+    }
   }
 
   /**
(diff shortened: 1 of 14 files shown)


cwl/src/main/scala/cwl/CwlWomExpression.scala
```diff
@@ -121,12 +123,17 @@ final case class InitialWorkDirFileGeneratorExpression(entry: IwdrListingArrayEn
...
+  override def evaluateValue(inputValues: Map[String, WomValue], ioFunctionSet: IoFunctionSet): ErrorOr[WomValue] = {
     def mustBeString(womValue: WomValue): ErrorOr[String] = womValue match {
       case WomString(s) => s.validNel
       case other => WomStringType.coerceRawValue(other).map(_.asInstanceOf[WomString].value).toErrorOr
     }
 
+    def mustBeFileArray(womValue: WomValue): ErrorOr[Seq[WomFile]] = womValue match {
+      case WomArray(WomArrayType(WomSingleFileType), value) => value.collect({case f: WomFile => f}).validNel
+      case other => WomArrayType(WomSingleFileType).coerceRawValue(other).map(_.asInstanceOf[WomArray]).toErrorOr.flatMap(mustBeFileArray)
+    }
```
mcovarr: This isn't used anywhere.

cwl/src/main/scala/cwl/CwlWomExpression.scala
```diff
@@ -155,8 +166,22 @@ final case class InitialWorkDirFileGeneratorExpression(entry: IwdrListingArrayEn
...
         }
+      case IwdrListingArrayEntry.ECMAScriptExpression(expression) =>
+        // A single expression which must evaluate to an array of Files
+        val expressionEvaluation: ErrorOr[WomValue] = ExpressionEvaluator.evalExpression(expression, ParameterContext().withInputs(inputValues, ioFunctionSet)).toErrorOr
+
+        expressionEvaluation flatMap {
+          case array: WomArray if WomArrayType(WomSingleFileType).coercionDefined(array) =>
+            val newFileArray: ErrorOr[List[WomFile]] = array.value.map(_.valueString).toList.traverse { source: String =>
+              val dest = source.split('/').last
+              Try(Await.result(ioFunctionSet.copyFile(source, dest), Duration.Inf)).toErrorOr
```
mcovarr: ToL The 5 `Await.result(..., Duration.Inf)`s in this file suggest to me these functions maybe shouldn't be synchronous... 🙂

mcovarr: This does look to be very similar to the yuck above.

cwl/src/main/scala/cwl/InitialWorkDirRequirement.scala
```diff
@@ -58,6 +59,21 @@ object InitialWorkDirRequirement {
...
+      def unapply(e: IwdrListingArrayEntry): Option[String] = e.select[File].flatMap(file => file.location)
+    }
+    object String {
+      def unapply(e: IwdrListingArrayEntry): Option[String] = e.select[StringOrExpression].flatMap(_.select[String])
+    }
+    object ECMAScriptExpression {
+      def unapply(e: IwdrListingArrayEntry): Option[ECMAScriptExpression] = for {
+        soe <- e.select[StringOrExpression]
+        expr <- soe.select[Expression]
+        ecmaScriptExpression <- expr.select[ECMAScriptExpression]
```
kshakir: Why does this only apply to `ECMAScriptExpression` and not `Expression`? According to the way I read [the spec](http://www.commonwl.org/v1.0/CommandLineTool.html#InitialWorkDirRequirement), `ECMAScriptFunction` should be valid wherever `ECMAScriptExpression` are valid.

Also, ToL, I haven't made up my mind yet whether I find `Poly1`s easier to read than these `unapply`s used in match statements.

cwl/src/main/scala/cwl/TypeAliases.scala
```diff
@@ -100,8 +102,14 @@ trait TypeAliases {
 }
 
 object CwlAny {
-  implicit class EnhancedCwlAny(val cwlAny: CwlAny) extends AnyVal {
-    def stringRepresentation: String = cwlAny.select[Json] map { json => json.asString.getOrElse(json.toString) } getOrElse "CwlAny (not Json)"
+
+  implicit class EnhancedJson(val json: Json) extends AnyVal {
+    // Yes, urgh! This is disgusting but until we rewrite the entire JS coercion set for circe...
+    def sprayJsonRepresentation: JsValue = json.toString().parseJson
```
kshakir: `json.toString().parseJson` -> `json.noSpaces.parseJson`

cwl/src/main/scala/cwl/CwlWomExpression.scala
cjllanwarne: I agree, but this is probably a bigger problem than just `InitialWorkDirRequirement` (this is itself copy/paste from elsewhere)

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -249,14 +249,15 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
 
     val adHocFileCreationInputs = jobDescriptor.evaluatedTaskInputs.map { case (k,v) => k.localName.value -> v }
 
-    def validateAdHocFile(value: WomValue): ErrorOr[WomFile] = value match {
-        case f: WomFile => Valid(f)
+    def validateAdHocFile(value: WomValue): ErrorOr[List[WomFile]] = value match {
+        case f: WomFile => Valid(List(f))
```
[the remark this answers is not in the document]
Horneth: `import cats.syntax.validated._` pls :)

cwl/src/main/scala/cwl/InitialWorkDirRequirement.scala
cjllanwarne: I'll address this in an iwdr_7 branch

PR #3108, opened 2018-01-05
Fix requirement nesting

mcovarr:
Whilst working on env vars I noticed the Run / WorkflowStep / Workflow relationship wasn't modeled correctly wrt Requirements. That's something the env var conformance tests in particular will check, but it's generally useful to have this be correct so here it is in its own wee PR.

### DIFF
--- a/centaur/src/main/resources/standardTestCases/three_step.test
+++ b/centaur/src/main/resources/standardTestCases/three_step.test
@@ -16,5 +16,5 @@ metadata {
   "calls.cgrep.executionStatus": Done
   "calls.ps.runtimeAttributes.docker": "ubuntu:bionic"
   "calls.wc.runtimeAttributes.docker": "ubuntu:latest"
-  "calls.cgrep.runtimeAttributes.docker": "ubuntu:latest"
+  "calls.cgrep.runtimeAttributes.docker": "debian:jessie"
 }
--- a/cwl/src/main/scala/cwl/CommandLineTool.scala
+++ b/cwl/src/main/scala/cwl/CommandLineTool.scala
@@ -49,17 +49,22 @@ case class CommandLineTool private(
   private [cwl] implicit val explicitWorkflowName = ParentName(id)
   private val inputNames = this.inputs.map(i => FullyQualifiedName(i.id).id).toSet
 
+  // Circe can't create bidirectional links between workflow steps and runs (including `CommandLineTool`s) so this
+  // ugly var is here to link back to a possible parent workflow step. This is needed to navigate upward for finding
+  // requirements in the containment hierarchy. There isn't always a containing workflow step so this is an `Option`.
+  private[cwl] var parentWorkflowStep: Option[WorkflowStep] = None
+
   /** Builds an `Executable` directly from a `CommandLineTool` CWL with no parent workflow. */
   def womExecutable(validator: RequirementsValidator, inputFile: Option[String] = None): Checked[Executable] = {
-    val taskDefinition = buildTaskDefinition(parentWorkflow = None, validator)
+    val taskDefinition = buildTaskDefinition(validator)
     CwlExecutableValidation.buildWomExecutable(taskDefinition, inputFile)
   }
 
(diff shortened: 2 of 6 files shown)


centaur/src/main/resources/standardTestCases/three_step.test
```diff
@@ -16,5 +16,5 @@ metadata {
   "calls.cgrep.executionStatus": Done
   "calls.ps.runtimeAttributes.docker": "ubuntu:bionic"
   "calls.wc.runtimeAttributes.docker": "ubuntu:latest"
-  "calls.cgrep.runtimeAttributes.docker": "ubuntu:latest"
+  "calls.cgrep.runtimeAttributes.docker": "debian:jessie"
```
geoffjentry: pure curiosity here, why the change?

mcovarr: So I could assert that the `DockerRequirement` on the workflow step was correctly being given precedence over the `DockerRequirement` on the workflow.

cwl/src/main/scala/cwl/CommandLineTool.scala
```diff
@@ -49,25 +49,27 @@ case class CommandLineTool private(
   private [cwl] implicit val explicitWorkflowName = ParentName(id)
   private val inputNames = this.inputs.map(i => FullyQualifiedName(i.id).id).toSet
 
+  private[cwl] var parentWorkflowStep: Option[WorkflowStep] = None
```
kshakir: _Shame!_ 🔔 🔔 

No really-- can you leave a comment beside this var with the briefest of descriptions for future developers as to why the `var` is required?

mcovarr: Yeah not my proudest moment but I did what I had to. 🙂 I'll add some docs.

cwl/src/main/scala/cwl/CommandLineTool.scala
```diff
@@ -49,25 +49,30 @@ case class CommandLineTool private(
   private [cwl] implicit val explicitWorkflowName = ParentName(id)
   private val inputNames = this.inputs.map(i => FullyQualifiedName(i.id).id).toSet
 
+  // Circe can't create bidirectional links between workflow steps and runs (including `CommandLineTool`s) so this
+  // ugly var is here to link back to a possible parent workflow step. This is needed to navigate upward for finding
+  // requirements in the containment hierarchy. There isn't always a containing workflow step so this is an `Option`.
+  private[cwl] var parentWorkflowStep: Option[WorkflowStep] = None
```
cjllanwarne: Does this work even in a multiple file scenario? (I notice your test case is an all-in-one cwl file)

cwl/src/main/scala/cwl/Workflow.scala
```diff
@@ -30,21 +30,24 @@ case class Workflow private(
...
+
+  steps.foreach { _.parentWorkflow = this }
+
   /** Builds an `Executable` from a `Workflow` CWL with no parent `Workflow` */
   def womExecutable(validator: RequirementsValidator, inputFile: Option[String] = None): Checked[Executable] = {
     CwlExecutableValidation.buildWomExecutable(womDefinition(validator), inputFile)
   }
 
-  private[cwl] var parentWorkflow: Option[Workflow] = None
+  private[cwl] var parentWorkflowStep: Option[WorkflowStep] = None
```
cjllanwarne: ToL: This ugly var doesn't get the comment that the other two do

cwl/src/main/scala/cwl/CommandLineTool.scala
mcovarr: Doesn't SALAD all-in-one everything before we begin to parse?

cwl/src/main/scala/cwl/Workflow.scala
mcovarr: oops

mcovarr:
@cjllanwarne while the ExpressionStep has requirement and hint fields it currently isn't doing anything with them at all. 🤔

PR #3121, opened 2018-01-09
reference slick docs in our db config

danbills:
We see questions on these often and it would be nice for users to know where to look to find answers.


cjllanwarne:
Any place for this link in RTFD too?

cjllanwarne:
👍

PR #3126, opened 2018-01-10
query to return total results count

ruchim:
Motivation: This is a feature requested by the Mint team. They want the total number of results matching a query to be included in the query response. Mint wants to expose the total results count through the Job-Manager UI.


engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
```diff
@@ -557,6 +557,16 @@ class CromwellApiServiceSpec extends AsyncFlatSpec with ScalatestRouteTest with
         }
     }
 
+    it should "include totalResultCount in workflow query response" in {
```
cjllanwarne: Is there any way to be sure that this runs at exactly the right time for the answer to be `1`?

(ie to avoid the "unit tests might run in any order" thing)

engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
```diff
@@ -652,7 +662,7 @@ object CromwellApiServiceSpec {
           parameters.contains(("additionalQueryResultFields", "parentWorkflowId")).option("pid")
         }
         val response = WorkflowQuerySuccess(WorkflowQueryResponse(List(WorkflowQueryResult(ExistingWorkflowId.toString,
-          None, Some(WorkflowSucceeded.toString), None, None, labels, parentWorkflowId))), None)
+          None, Some(WorkflowSucceeded.toString), None, None, labels, parentWorkflowId)), 1), None)
```
cjllanwarne: This looks like it already covers the "should be 1" aspect of the previous test

CHANGELOG.md
```diff
@@ -4,6 +4,13 @@
 
 * Added a configuration option under `docker.hash-lookup.enabled` to disable docker hash lookup.
  Disabling it will also disable call caching for jobs with floating docker tags.
+ 
+* **API**    
+    + Updated the `/query` response to include the total number of query results returned. See [here](http://cromwell.readthedocs.io/en/develop/api/RESTAPI/#workflowqueryresponse) for more information.
+
+## 30.1 Release Notes
+
+* A set of bug fixes following the migration of Cromwell to WOM (the Workflow Object Model) in version 30.
```
kshakir: ToL: weird. not sure how the 30.1 change notes are just now being added.

geoffjentry: I remember noticing that they weren't on there when it was released and since I could make an argument either way I didn't say anything.

OTOH this would be for a `30.2` release as `30.1` is already out ....

ruchim: I just didn't like that they were missing.....but I can take them out of this PR and add them separately?

cjllanwarne: I like them (although it's a little strange to see them in this PR, they're better here than nowhere!)

cjllanwarne:
Great 👍 again!

PR #3129, opened 2018-01-10
Set "Trying to use another call cache hit" messages to debug [29 Hotfix]

cjllanwarne:


geoffjentry:
👍

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/EngineJobExecutionActor.scala
```diff
@@ -515,11 +515,11 @@ class EngineJobExecutionActor(replyTo: ActorRef,
 
     data.ejha match {
       case Some(ejha) =>
-        log.info("Trying to use another cache hit for job: {}", jobDescriptorKey)
+        workflowLogger.debug("Trying to use another cache hit for job: {}", jobDescriptorKey)
```
mcovarr: is the stringification of the `jobDescriptorKey` okay?

cjllanwarne: it's not *great* but it's surprisingly readable

cjllanwarne: Example:
```
Trying to use another cache hit for job: BackendJobDescriptorKey_TaskCallNode_foo.bar:20:1
```

mcovarr: I can has workflow id?

cjllanwarne: Yeah, that's why I also switched it to workflowLogger

mcovarr: sweet

PR #3134, opened 2018-01-10
[PREGULL] (De)Simpletonize WomFiles

Horneth:
This might be controversial hence the PREGULL.
The problem is `WomFile` can now take different forms.
`WomSingleFile` is just a file path (for now)
`WomMaybePopulatedFile` can have a whole bunch of attributes, including a `List[WomFile]`
Same for `WomMaybeListedDirectory`.
`WomFile` => Simpletons is not the problem since we can just simpletonize a "complex file" as an object.
Simpletons => `WomFile` is harder because we don't necessarily know which kind of `WomFile` it is that we're trying to build.
The solution in this PR is to leave a "marker" in the simpleton key to give us a hint that what's following is a `WomMaybeListedDirectory` or a `WomMaybePopulatedFile` and re-build it appropriately.

### DIFF
--- a/core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala
+++ b/core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala
@@ -106,8 +106,8 @@ object WomValueBuilder {
       else {
         // Otherwise make a map of the components and detect the type of file from the marker
         val groupedListing = components.asMap
-        if (groupedListing.get("<<directory>>").isDefined) toWomValue(WomMaybeListedDirectoryType, components)
-        else if (groupedListing.get("<<populated>>").isDefined) toWomValue(WomMaybePopulatedFileType, components)
+        if (groupedListing.get(WomValueSimpleton.DirectoryMarker).isDefined) toWomValue(WomMaybeListedDirectoryType, components)
+        else if (groupedListing.get(WomValueSimpleton.PopulatedMarker).isDefined) toWomValue(WomMaybePopulatedFileType, components)
         else throw new IllegalArgumentException(s"Cannot build a WomFile from simpletons: ${groupedListing.toList.mkString(", ")}")
       }
     }
@@ -134,7 +134,7 @@ object WomValueBuilder {
         val map: Map[String, WomValue] = components.asMap map { case (k, ss) => k -> toWomValue(WomAnyType, ss) }
         WomObject(map)
       case WomMaybeListedDirectoryType =>
-        val directoryValues = components.asMap("<<directory>>").asMap
+        val directoryValues = components.asMap(WomValueSimpleton.DirectoryMarker).asMap
 
         val value = directoryValues.get("value").map(_.asString)
         val listing = directoryValues.get("listing")
@@ -142,7 +142,7 @@ object WomValueBuilder {
 
         WomMaybeListedDirectory(value, listing)
       case WomMaybePopulatedFileType =>
-        val populatedValues = components.asMap("<<populated>>").asMap
+        val populatedValues = components.asMap(WomValueSimpleton.PopulatedMarker).asMap
 
         val value = populatedValues.get("value").map(_.asString)
         val checksum = populatedValues.get("checksum").map(_.asString)
--- a/core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala
+++ b/core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala
@@ -55,23 +55,37 @@ object WomValueBuilder {
   // Within the noncapturing `?:` group, this looks for an escaped metacharacter OR a non-metacharacter.
   private val MapElementPattern = raw"^:((?:\\[]\[:]|[^]\[:])+)(.*)".r
 
+  // Group tuples by key using a Map with key type `K`.
+  private def group[K](tuples: Traversable[(K, SimpletonComponent)]): Map[K, Traversable[SimpletonComponent]] = {
+    tuples groupBy { case (i, _) => i } mapValues { _ map { case (_, s) => s} }
+  }
+
+  // Returns a tuple of the index into the outermost array and a `SimpletonComponent` whose path reflects the "descent"
+  // into the array.  e.g. for a component
+  // SimpletonComponent("[0][1]", v) this would return (0 -> SimpletonComponent("[1]", v)).
+  private def descendIntoArray(component: SimpletonComponent): (Int, SimpletonComponent) = {
+    component.path match { case ArrayElementPattern(index, more) => index.toInt -> component.copy(path = more)}
+  }
+
+  // Returns a tuple of the key into the outermost map and a `SimpletonComponent` whose path reflects the "descent"
+  // into the map.  e.g. for a component
+  // SimpletonComponent(":bar:baz", v) this would return ("bar" -> SimpletonComponent(":baz", v)).
+  // Map keys are treated as Strings by this method, the caller must ultimately do the appropriate coercion to the
+  // actual map key type.
+  private def descendIntoMap(component: SimpletonComponent): (String, SimpletonComponent) = {
+    component.path match { case MapElementPattern(key, more) => key.unescapeMeta -> component.copy(path = more)}
+  }
+
+  private implicit class EnhancedSimpletonComponents(val components: Traversable[SimpletonComponent]) extends AnyVal {
+    def asArray: List[Traversable[SimpletonComponent]] = group(components map descendIntoArray).toList.sortBy(_._1).map(_._2)
+    def asMap: Map[String, Traversable[SimpletonComponent]] = group(components map descendIntoMap)
+    def asPrimitive: WomValue = components.head.value
+    def asString: String = asPrimitive.valueString
+  }
+
   private def toWomValue(outputType: WomType, components: Traversable[SimpletonComponent]): WomValue = {
 
-    // Returns a tuple of the index into the outermost array and a `SimpletonComponent` whose path reflects the "descent"
-    // into the array.  e.g. for a component
-    // SimpletonComponent("[0][1]", v) this would return (0 -> SimpletonComponent("[1]", v)).
-    def descendIntoArray(component: SimpletonComponent): (Int, SimpletonComponent) = {
-      component.path match { case ArrayElementPattern(index, more) => index.toInt -> component.copy(path = more)}
-    }
 
-    // Returns a tuple of the key into the outermost map and a `SimpletonComponent` whose path reflects the "descent"
-    // into the map.  e.g. for a component
-    // SimpletonComponent(":bar:baz", v) this would return ("bar" -> SimpletonComponent(":baz", v)).
-    // Map keys are treated as Strings by this method, the caller must ultimately do the appropriate coercion to the
-    // actual map key type.
-    def descendIntoMap(component: SimpletonComponent): (String, SimpletonComponent) = {
-      component.path match { case MapElementPattern(key, more) => key.unescapeMeta -> component.copy(path = more)}
-    }
 
     // Returns a tuple of the key into the pair (i.e. left or right) and a `SimpletonComponent` whose path reflects the "descent"
     // into the pair.  e.g. for a component
@@ -85,15 +99,22 @@ object WomValueBuilder {
         case MapElementPattern("right", more) => PairRight -> component.copy(path = more)
       }
     }
-
-    // Group tuples by key using a Map with key type `K`.
-    def group[K](tuples: Traversable[(K, SimpletonComponent)]): Map[K, Traversable[SimpletonComponent]] = {
-      tuples groupBy { case (i, _) => i } mapValues { _ map { case (_, s) => s} }
+    
+    def toWomFile(components: Traversable[SimpletonComponent]) = {
+      // If there's just one simpleton, it's a single file
+      if (components.size == 1) toWomValue(WomSingleFileType, components)
+      else {
+        // Otherwise make a map of the components and detect the type of file from the marker
+        val groupedListing = components.asMap
+        if (groupedListing.get("<<directory>>").isDefined) toWomValue(WomMaybeListedDirectoryType, components)
+        else if (groupedListing.get("<<populated>>").isDefined) toWomValue(WomMaybePopulatedFileType, components)
+        else throw new IllegalArgumentException(s"Cannot build a WomFile from simpletons: ${groupedListing.toList.mkString(", ")}")
+      }
     }
-
+    
     outputType match {
       case _: WomPrimitiveType =>
-        components collectFirst { case SimpletonComponent(_, v) => v } get
+        components.asPrimitive
       case opt: WomOptionalType =>
         if (components.isEmpty) {
           WomOptionalValue(opt.memberType, None)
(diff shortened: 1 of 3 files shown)


core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala
```diff
@@ -85,36 +99,68 @@ object WomValueBuilder {
...
       }
     }
-
-    // Group tuples by key using a Map with key type `K`.
-    def group[K](tuples: Traversable[(K, SimpletonComponent)]): Map[K, Traversable[SimpletonComponent]] = {
-      tuples groupBy { case (i, _) => i } mapValues { _ map { case (_, s) => s} }
+    
+    def toWomFile(components: Traversable[SimpletonComponent]) = {
+      // If there's just one simpleton, it's a single file
+      if (components.size == 1) toWomValue(WomSingleFileType, components)
```
kshakir: Is this always true?

`WomUnlistedDirectoryType` is also a `WomPrimitiveType`, matching on the `case _: WomPrimitiveType =>` below.

kshakir: I'm still digesting all this, so happy to discuss further.

There used to only be _one_ `WomType` for _all_ files, the `WomFileType`. However, that type could be two different `WomValue` instances, either a `WomSingleFile` or a `WomGlobFile`. Now there is no concrete `WomFileType`. It's a trait instead, like `WomPrimitiveType`. The old type with a display string "File" is mapped to `WomSingleFileType`.

Before:
https://github.com/broadinstitute/cromwell/blob/cbdc3bc80736ae2ae007c8956c50d7db626af07e/wom/src/main/scala/wom/types/WomFileType.scala#L8-L9

Currently:
https://github.com/broadinstitute/cromwell/blob/063a4516487c21dc9e96292f0cf2786b15b584e1/wom/src/main/scala/wom/types/WomFileType.scala#L8

I'm wonder if...

For each of the three `WomPrimitive` types:
- `WomGlobFile`
- `WomSingleFile`
- `WomUnlistedDirectory`

...use existing code for singletons that writes the type + the-single-value (at least I _think_ it writes the types).

For the other non-primitive `WomFile`:
- `WomMaybeListedDirectory`
- `WomMaybePopulatedFile`

...write them as one would a recursive `WomMap`/`WomObject`. This shouldn't require a "marker" anymore, assuming we know via the type that instead the rows should be converted to a `WomMaybeListedDirectory`/`WomMaybePopulatedFile`?

I'm also realizing that the `toDisplayString` may be what is stored in the database and should be added to the non-primitive types. However, I also don't know how much that "display string" in WOM must be the same as it is / will be defined in the WDL spec.

Horneth: Sounds good for the primitive `WomFileType`s
However 
> assuming we know via the type that instead the rows should be converted to a WomMaybeListedDirectory/WomMaybePopulatedFile

I don't believe we do. For example `WomMaybeListedDirectory` has a `listingOption: Option[Seq[WomFile]]` attribute.
When we simpletonize those files we lose the information of what type of file exactly was it (specifically was it a WomMaybeListedDirectory or a WomMaybePopulatedFile ? - because they're both going to end up being simpletonized like objects). This is what the marker is trying to make up for.
When we want to build them back to `WomValue` the only thing we know is that they are `WomFile`s.

core/src/main/scala/cromwell/core/simpleton/WomValueSimpleton.scala
```diff
@@ -15,6 +15,9 @@ case class WomValueSimpleton(simpletonKey: String, simpletonValue: WomPrimitive)
   * `WomValueSimpleton`s are transformed back to `WomValue`s.
   */
 object WomValueSimpleton {
+  
+  val DirectoryMarker = "<<directory>>"
+  val PopulatedMarker = "<<populated>>"
```
kshakir: In this PR, new markers `<<directory>>` or `<<populated>>` are injected into the names of files.

Assuming that the WDL_TYPE stays WomMaybeListedDirectoryType for all rows (is that right?), what's not clear to me is how the WomUnlistedDirectory("innerUnlistedDirectory") is deserialized as the correct type.

| `WDL_TYPE`                    | `SIMPLETON_KEY`                                                                               | `SIMPLETON_VALUE`          |
|-------------------------------|-----------------------------------------------------------------------------------------------|----------------------------|
| `WomMaybeListedDirectoryType` | `directory:<<directory>>:value`                                                               | `outerValueName`           |
| `WomMaybeListedDirectoryType` | `directory:<<directory>>:listing[0]`                                                          | `outerSingleFile`          |
| `WomMaybeListedDirectoryType` | `directory:<<directory>>:listing[1]:<<directory>>:value`                                      | `innerValueName`           |
| `WomMaybeListedDirectoryType` | `directory:<<directory>>:listing[1]:<<directory>>:listing[0]`                                 | `innerSingleFile`          |
| `WomMaybeListedDirectoryType` | `directory:<<directory>>:listing[2]:<<populated>>:value`                                      | `populatedInnerValueName`  |
| `WomMaybeListedDirectoryType` | `directory:<<directory>>:listing[2]:<<populated>>:checksum`                                   | `innerChecksum`            |
| `WomMaybeListedDirectoryType` | `directory:<<directory>>:listing[2]:<<populated>>:size`                                       | `10`                       |
| `WomMaybeListedDirectoryType` | `directory:<<directory>>:listing[2]:<<populated>>:format`                                     | `innerFormat`              |
| `WomMaybeListedDirectoryType` | `directory:<<directory>>:listing[2]:<<populated>>:contents`                                   | `innerContents`            |
| `WomMaybeListedDirectoryType` | `directory:<<directory>>:listing[2]:<<populated>>:secondaryFiles[0]`                          | `populatedInnerSingleFile` |
| `WomMaybeListedDirectoryType` | `directory:<<directory>>:listing[2]:<<populated>>:secondaryFiles[1]:<<directory>>:value`      | `innerDirectoryValueName`  |
| `WomMaybeListedDirectoryType` | `directory:<<directory>>:listing[2]:<<populated>>:secondaryFiles[1]:<<directory>>:listing[0]` | `innerDirectorySingleFile` |
| `WomMaybeListedDirectoryType` | `directory:<<directory>>:listing[2]:<<populated>>:secondaryFiles[2]`                          | `innerUnlistedDirectory`   |
| `WomMaybeListedDirectoryType` | `directory:<<directory>>:listing[2]:<<populated>>:secondaryFiles[3]`                          | `innerGlobFile`            |
| `WomMaybeListedDirectoryType` | `directory:<<directory>>:listing[3]`                                                          | `outerUnlistedDirectory`   |
| `WomMaybeListedDirectoryType` | `directory:<<directory>>:listing[4]`                                                          | `outerGlobFile`            |

Alternatively, instead of store the type in the key name, how about storing the type as a new row/entry?		

| `WDL_TYPE`                    | `SIMPLETON_KEY`                                                 | `SIMPLETON_VALUE`             |
|-------------------------------|-----------------------------------------------------------------|-------------------------------|
| `WomMaybeListedDirectoryType` | `directory:womFileType`                                         | `WomMaybeListedDirectoryType` |
| `WomMaybeListedDirectoryType` | `directory:value`                                               | `outerValueName`              |
| `WomMaybeListedDirectoryType` | `directory:listing[0]:womFileType`                              | `WomSingleFileType`           |
| `WomMaybeListedDirectoryType` | `directory:listing[0]:value`                                    | `outerSingleFile`             |
| `WomMaybeListedDirectoryType` | `directory:listing[1]:womFileType`                              | `WomMaybeListedDirectoryType` |
| `WomMaybeListedDirectoryType` | `directory:listing[1]:value`                                    | `innerValueName`              |
| `WomMaybeListedDirectoryType` | `directory:listing[1]:listing[0]:womFileType`                   | `WomSingleFileType`           |
| `WomMaybeListedDirectoryType` | `directory:listing[1]:listing[0]:value`                         | `innerSingleFile`             |
| `WomMaybeListedDirectoryType` | `directory:listing[2]:womFileType`                              | `WomMaybePopulatedFileType`   |
| `WomMaybeListedDirectoryType` | `directory:listing[2]:value`                                    | `populatedInnerValueName`     |
| `WomMaybeListedDirectoryType` | `directory:listing[2]:checksum`                                 | `innerChecksum`               |
| `WomMaybeListedDirectoryType` | `directory:listing[2]:size`                                     | `10`                          |
| `WomMaybeListedDirectoryType` | `directory:listing[2]:format`                                   | `innerFormat`                 |
| `WomMaybeListedDirectoryType` | `directory:listing[2]:contents`                                 | `innerContents`               |
| `WomMaybeListedDirectoryType` | `directory:listing[2]:secondaryFiles[0]:womFileType`            | `WomSingleFileType`           |
| `WomMaybeListedDirectoryType` | `directory:listing[2]:secondaryFiles[0]:value`                  | `populatedInnerSingleFile`    |
| `WomMaybeListedDirectoryType` | `directory:listing[2]:secondaryFiles[1]:womFileType`            | `WomSingleFileType`           |
| `WomMaybeListedDirectoryType` | `directory:listing[2]:secondaryFiles[1]:value`                  | `innerDirectoryValueName`     |
| `WomMaybeListedDirectoryType` | `directory:listing[2]:secondaryFiles[1]:listing[0]:womFileType` | `WomSingleFileType`           |
| `WomMaybeListedDirectoryType` | `directory:listing[2]:secondaryFiles[1]:listing[0]:value`       | `innerDirectorySingleFile`    |
| `WomMaybeListedDirectoryType` | `directory:listing[2]:secondaryFiles[2]:womFileType`            | `WomUnlistedDirectoryType`    |
| `WomMaybeListedDirectoryType` | `directory:listing[2]:secondaryFiles[2]:value`                  | `innerUnlistedDirectory`      |
| `WomMaybeListedDirectoryType` | `directory:listing[2]:secondaryFiles[3]:womFileType`            | `WomGlobFile`                 |
| `WomMaybeListedDirectoryType` | `directory:listing[2]:secondaryFiles[3]:value`                  | `innerGlobFile`               |
| `WomMaybeListedDirectoryType` | `directory:listing[3]:womFileType`                              | `WomUnlistedDirectoryType`    |
| `WomMaybeListedDirectoryType` | `directory:listing[3]:value`                                    | `outerUnlistedDirectory`      |
| `WomMaybeListedDirectoryType` | `directory:listing[4]:womFileType`                              | `WomGlobFileType`             |
| `WomMaybeListedDirectoryType` | `directory:listing[4]:value`                                    | `outerGlobFile`               |

However, this is the way current root-level WomSingleFiles are stored. There is no :value sub-key																									

| `WDL_TYPE`          | `SIMPLETON_KEY` | `SIMPLETON_VALUE`   |
|---------------------|-----------------|---------------------|
| `WomSingleFileType` | `myFile`        | `WomSingleFileType` |

Under the above proposal, this is the way new files might be stored. So the deserialization code should be sure to be able to handle both types. Or, alternatively, root-level WomSingleFile can still use the non-:value type, and only when a WomSingleFile is under a directory listing or secondary files array would it include the :value + :womFileType																									

| `WDL_TYPE`          | `SIMPLETON_KEY`      | `SIMPLETON_VALUE`   |
|---------------------|----------------------|---------------------|
| `WomSingleFileType` | `myFile:womFileType` | `WomSingleFileType` |
| `WomSingleFileType` | `myFile:value`       | `/path/to/file`     |

Iff for some reason the simpletons _must be_ stored in a currently-compatible way, both the current and new types of rows could be written.

| `WDL_TYPE`          | `SIMPLETON_KEY`      | `SIMPLETON_VALUE`   |
|---------------------|----------------------|---------------------|
| `WomSingleFileType` | `myFile:womFileType` | `WomSingleFileType` |
| `WomSingleFileType` | `myFile:value`       | `/path/to/file`     |
| `WomSingleFileType` | `myFile`             | `/path/to/file`     |

Horneth: My understanding of simpletons is that they necessarily represent a primitive type. If a `WomValue` is not a `WomPrimitive`, then it gets broken down into as many simpletons as necessary that are themselves of a primitive type. The `WDL_TYPE` field is therefore necessarily a `WomPrimitiveType`
So in your example above, again as I understand it, none of the simpleton keys would be of type `WomMaybeListedDirectoryType` (which is not a primitive type) but would instead be of whatever type the value ultimately boils down to (for example `checksum` would be `WomStringType`, `size` would be `WomIntegerType`, etc...).

I didn't want to break this assumption but if it's something we're willing to consider then I think by setting the type of each individual simpleton to the `WomFileType` they belong to it could be possible to reconstruct it properly.

kshakir: Cool. I don't completely get the simpleton serialization/deserialization code yet so thanks for hammering this all out.

I guess I'm curious what a serialized form of other non-primitive types looks like, especially the new WomObjectLike(?) that have the types-of-the-values listed within a map. Whatever they use, the various WomFile types should use.

geoffjentry: If that was broken, would we start calling them `Complextons`?

Horneth: This is how I understand the serialization / deserialization of WomValues so far. I apologize you probably already know most (all) of what's below I'm just brain dumping :)

The simpleton "mini-language" that we use to flatten out arbitrarily nested `WomValue`s gives us basically the same flexibility as Json does:

primitive types:

| `WDL_TYPE`                    | `SIMPLETON_KEY`                                                                               | `SIMPLETON_VALUE`          |
|-------------------------------|-----------------------------------------------------------------------------------------------|----------------------------|
| `WomStringType` | `myStringVariable`  | `hello`    |
| `WomIntegerType` | `myIntVariable`  | `5`    |
| `WomBooleanType` | `myBoolVariable`  | `true`    |
| `WomFloatType` | `myFloatVariable`  | `5.5`    |
| `WomFileType` | `myFloatVariable`  | `myFile.txt`    |

Note that because like you mentioned `WomFileType` doesn't exist as a type anymore, the new File primitive types are instead `WomUnlistedDirectory`, `WomSingleFileType` and `WomGlobFileType`.

We can also represent arrays:

| `WDL_TYPE`                    | `SIMPLETON_KEY`                                                                               | `SIMPLETON_VALUE`          |
|-------------------------------|-----------------------------------------------------------------------------------------------|----------------------------|
| `WomStringType` | `myArrayVariable[0]`  | `hello0`    |
| `WomStringType` | `myArrayVariable[1]`  | `hello1`    |

Which let's say in WDL (for conciseness) would be `Array[String] myArrayVariable = ["hello0", "hello1"]`

And key/value pairs:

| `WDL_TYPE`                    | `SIMPLETON_KEY`                                                                               | `SIMPLETON_VALUE`          |
|-------------------------------|-----------------------------------------------------------------------------------------------|----------------------------|
| `WomStringType` | `myObjectVariable:fieldA`  | `helloA`    |
| `WomStringType` | `myObjectVariable:fieldB`  | `helloB`    |

Now if I want to show what this would be in WDL (or WOM) this is exactly where it gets tricky because there are multiple Wom structures that could produce the above 2 simpletons.

Could be 
```scala
WomMap(
WomMapType(WomStringType, WomStringType),
 Map(
   WomString("fieldA") -> WomString("helloA"),
   WomString("fieldB") -> WomString("helloB")
 )
```
or 
```scala
WomObject(
 Map(
  "fieldA" -> WomString("helloA"),
  "fieldB" -> WomString("helloB")
 )
```

Now this hasn't be too much of a problem so far because
1) the example above is (I think) the only one so far where we're not sure
2) WomObject <-> WomMap are coerceable one to the other to some extent
3) most of the time we do know exactly what target WomType we want to build the simpletons to so there's no ambiguity.

However you can see [here](https://github.com/broadinstitute/cromwell/blob/develop/core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala#L119) that there are some cases where we don't know the exact type we want and so we have to guess.

Now with the new file types, this problem becomes more visible because we need to store those complex file values that are NOT arbitrary key / value pairs (like object or map), but we store them AS key / value pairs anyway by taking each attribute and making simpleton(s) off of it (maybe that's what we should do differently ?).
In doing so we lose the information of the exact WomFileType of the value.
That would be ok if when we need to rebuild the original `WomValues` we always knew what the original `WomFileType` was, but we don't.

For example the `listing` attribute of `WomMaybeListedDirectory` just says that the values are `WomFile` values, but we don't specifically what type of `WomFile`.

When we try to re-construct the original `WomFile` all we have is a `Map[String, WomValue]`. Now we need to decide what kind of `WomFileType` this represents. 
There are multiple ways we could do this. In this PR I added some sort of special marker in the key to give a hint as to what the specific `WomFileType` was.
We could also like you suggested add a "shadow simpleton" that gives us the type.
Or store the specific `WomFileType` in the DB for each simpleton value instead of their actual type.
Or we could peak into the map and look for attributes we know correspond to one type or the other: like is there a `secondaryFiles` entry ? if yes that's probably a `WomMaybePopulatedFile`...
Or something else ?

As for strongly typed `WomObject`s with a `WomCompositeType`, they are stored exactly the same way as a map or an object.
When they're being constructed back from simpletons to WomValues, if we know we're expecting a `WomCompositeType` then we can just instantiate it from the Map and if the values match the type constraints then it'll work (which you actually made me realize I forgot to do in the other PR so thanks 😄 ). If we don't it'll be built as whatever type information we have (maybe a simple `WomObject`, which can then be coerced to a `WomCompositeType` if needed).

PR #3135, opened 2018-01-11
Document the fact that we only support v1.0 in CWL

geoffjentry:
This doesn't solve the underlying problem I ran into yesterday, but it's a start


engine/src/main/resources/swagger/cromwell.yaml
```diff
@@ -117,10 +117,10 @@ paths:
           in: formData
         - $ref: '#/parameters/workflowTypeParam'
         - name: workflowTypeVersion
-          description: The specification version for the workflow language being used. Cromwell currently supports draft-2.
+          description: The specification version for the workflow language being used. Cromwell currently supports draft-2. For CWL, Cromwell currently support v1.0.
```
mcovarr: support => supports

engine/src/main/resources/swagger/cromwell.yaml
```diff
@@ -63,10 +63,10 @@ paths:
           type: string
           in: formData
         - name: workflowTypeVersion
-          description: The specification version for the workflow language being used. For WDL, Cromwell currently supports draft-2.
+          description: The specification version for the workflow language being used. For WDL, Cromwell currently supports draft-2. For CWL, Cromwell currently support v1.0.
```
mcovarr: support => supports

gemlam3:
@geoffjentry anything left to do for this prior to merging?

geoffjentry:
@gemlam3 yes

PR #3139, opened 2018-01-12
 Simpleton and Metadata for CWL Files / Directories

Horneth:


services/src/main/scala/cromwell/services/metadata/MetadataService.scala
```diff
@@ -145,6 +153,20 @@ object MetadataService {
...
+    case populated: WomMaybePopulatedFile =>
+      import mouse.all._
+      val secondaryFiles = populated.secondaryFiles.toEvents(metadataKey.copy(key = s"${metadataKey.key}:secondaryFiles"))
+
+      List(populated.valueOption |> toPrimitiveEvent(metadataKey, "value"),
+        populated.checksumOption |> toPrimitiveEvent(metadataKey, "checksum"),
+        populated.sizeOption |> toPrimitiveEvent(metadataKey, "size"),
+        populated.formatOption |> toPrimitiveEvent(metadataKey, "format"),
+        populated.contentsOption |> toPrimitiveEvent(metadataKey, "contents")
+      ) ++ secondaryFiles
```
Horneth: Assuming this is how we want to report Files / DIrectories in metadata, we might want to add a `class` field here too to make it clear what it is ? It would also help the `OutputManipulator` figure out what type of file it's dealing with

Horneth:
FYI The conformance test that's failing can't pass until the `OutputManipulator` is also updated - I'm doing that today

core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala
```diff
@@ -55,23 +55,37 @@ object WomValueBuilder {
   // Within the noncapturing `?:` group, this looks for an escaped metacharacter OR a non-metacharacter.
   private val MapElementPattern = raw"^:((?:\\[]\[:]|[^]\[:])+)(.*)".r
 
+  // Group tuples by key using a Map with key type `K`.
+  private def group[K](tuples: Traversable[(K, SimpletonComponent)]): Map[K, Traversable[SimpletonComponent]] = {
+    tuples groupBy { case (i, _) => i } mapValues { _ map { case (_, s) => s} }
```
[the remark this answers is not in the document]
geoffjentry: IMO it absolutely needs to be proven to be ok and should have a comment explaining why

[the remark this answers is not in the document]
Horneth: I simply moved from a few lines below so it could be used in the implicit class but I'm happy to change it to a full map

core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala
```diff
@@ -85,46 +99,95 @@ object WomValueBuilder {
...
         } else {
           WomOptionalValue(toWomValue(opt.memberType, components))
         }
       case arrayType: WomArrayType =>
-        val groupedByArrayIndex: Map[Int, Traversable[SimpletonComponent]] = group(components map descendIntoArray)
-        WomArray(arrayType, groupedByArrayIndex.toList.sortBy(_._1) map { case (_, s) => toWomValue(arrayType.memberType, s) })
+        WomArray(arrayType, components.asArray map { toWomValue(arrayType.memberType, _) })
       case mapType: WomMapType =>
-        val groupedByMapKey: Map[String, Traversable[SimpletonComponent]] = group(components map descendIntoMap)
         // map keys are guaranteed by the WDL spec to be primitives, so the "coerceRawValue(..).get" is safe.
```
cjllanwarne: What about the CWL spec?

cjllanwarne: eg what if the keys are files, and the files are objects?

core/src/test/scala/cromwell/core/simpleton/WomValueBuilderSpec.scala
```diff
@@ -136,6 +136,57 @@ class WomValueBuilderSpec extends FlatSpec with Matchers with Mockito {
         WomValueSimpleton("object_with_object:b:bb[0]", WomString("bbrrrr")),
         WomValueSimpleton("object_with_object:b:bb[1]", WomString("bb gun")),
       )
+    ),
+    SimpletonConversion(
```
cjllanwarne: ToL: Only because reading the WOM object notation is so awkward, could you add a quick comment explaining what this is simpletoning?

services/src/main/scala/cromwell/services/metadata/MetadataService.scala
cjllanwarne: I agree with Khalid

services/src/test/scala/cromwell/services/metadata/MetadataServiceSpec.scala
```diff
@@ -23,41 +23,109 @@ class MetadataServiceSpec extends FlatSpec with Matchers with TableDrivenPropert
...
+      MetadataEvent(MetadataKey(workflowId, None, "root:secondaryFiles[0]:contents"), MetadataValue("innerContents")),
+      MetadataEvent(MetadataKey(workflowId, None, "root:secondaryFiles[0]:secondaryFiles[]"), None),
+      MetadataEvent(MetadataKey(workflowId, None, "root:secondaryFiles[1]"), MetadataValue("outerSingleFile"))
+    )
+  }
+
+  it should "convert a WomMaybeListedDirectory to MetadataEvents" in {
+    import MetadataService._
+    val workflowId = WorkflowId.randomId()
+    // Leave some fields out on purpose to verify that they still make it (empty) to the metadata
```
cjllanwarne: What do they look like in metadata? Just a `null`?

core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala
cjllanwarne: Side note: Should have added ToL... I don't think we should worry about this right now

services/src/test/scala/cromwell/services/metadata/MetadataServiceSpec.scala
Horneth: yes

core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala
Horneth: Good point, and yeah I feel like we can punt on this case for now but I'll make a ticket if that's ok

services/src/main/scala/cromwell/services/metadata/MetadataService.scala
Horneth: `WomSingleFile`s serialize exactly the same so this shouldn't break any backwards compatibility as long as we haven't switched from `WomSingleFile`s to `WomMaybePopulatedFile`s where they really are `WomSingleFile`s ? ( @kshakir ?)

services/src/test/scala/cromwell/services/metadata/MetadataServiceSpec.scala
Horneth: except `secondaryFiles` and `listing` I should say which being arrays will be empty arrays instead

core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala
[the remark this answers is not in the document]
danbills: `mapValues` is fine when lambda is referentially transparent AFAIK

services/src/main/scala/cromwell/services/metadata/MetadataService.scala
kshakir: At this second, afaik there are no `WomSingleFile` instance, under WDL, being converted to `WomMaybePopulatedFile` instances. Really hoping though that existing centaur tests will catch it if we do change this.

🤔 Although, now that I think about it, I don't believe we have any `metadata` file outputs in centaur, only `outputExpectations` that search-and-replace the `<<UUID>>` to get the correct dynamic path.

Horneth: Mokay, I was hoping too that something would break if it was not backward compatible but I'm less confident merging this if nothing tests it :/

Horneth: I can't find evidence that previous `WomSingleFile` from WDL workflows have been changed into `WomMaybePopulatedFile`s so I'm going to go ahead and merge this
I'll create a ticket though in the hope that we can get better testing of File output format in metadata in the future

PR #3140, opened 2018-01-16
Adding expressionLib support

danbills:
Javascript requirements have the option of declaring JS functions to be used in expressions thereafter.  One may also include a file containing functions, as a conformance tests demonstrate.

Most of the changes here are the ripple effect of passing this library around.

I was forced to disable the "strict" javascript option in order to pass a conformance test.  The library it imported used a "with" clause that is evidently verboten in strict world.


src/bin/travis/resources/conformance_expected_failures.txt
```diff
@@ -87,7 +87,7 @@
 87
 88
 89
-90
+#90
```
Horneth: 🎉

cwl/src/main/scala/cwl/ExpressionEvaluator.scala
```diff
@@ -50,12 +58,13 @@ object ExpressionEvaluator {
   private lazy val cwlJsDecoder = new CwlJsDecoder()
 
   def eval(expr: String, parameterContext: ParameterContext): ErrorOr[WomValue] = {
+    println(s"evaluating $expr")
```
Horneth: remove

cwl/src/main/scala/cwl/ExpressionEvaluator.scala
```diff
@@ -24,23 +24,31 @@ object ExpressionEvaluator {
...
+        eval(expressionFromParts(expressionLib, script), parameterContext)
       case unmatched =>
         s"Expression '$unmatched' was unable to be matched to regex '${ECMAScriptExpressionWitness.value}'".invalidNel
     }
   }
 
-  def evalFunction(function: ECMAScriptFunction)(parameterContext: ParameterContext): ErrorOr[WomValue] = {
+  def expressionFromParts(lib: ExpressionLib, script: String) = {
+    val expressionScript: String = lib.mkString(";")
+    s"$expressionScript;$script"
```
Horneth: Am I correct in saying in almost everything in the PR is wiring to get here and add this before the script ?

danbills: That is correct!  Quite a ripple effect.

danbills: I think ultimately we can mitigate the ugliness w/ parameters/dependency injection by using the `Reader` monad, but I didn't feel brave enough to introduce it in this PR.

danbills:
@kshakir I thought about making it part of `ParameterContext`, but it is not strictly the case that every time it appears, it is "next to" a Parameter Context.  Thus it is kept separate.

kshakir:
> …it is not strictly the case that every time \[expressionLib\] appears, it is "next to" a Parameter Context.

Meant the other way around, Parameter Context is always "next to" a expressionLib. If it continues to be that way as CWL settles, one can always be update the Parameter Context via a future PR.

danbills:
@kshakir You're right!   It should be changed.  (my bad, not sure why I didn't consider that angle)

PR #3146, opened 2018-01-16
IWDR: Allow inputs to be relative

cjllanwarne:
🤞 at least one new conformance test passes...


backend/src/main/scala/cromwell/backend/wdl/WriteFunctions.scala
```diff
@@ -27,8 +29,10 @@ trait WriteFunctions extends IoFunctionSet with AsyncIoFunctions {
     }
   }
 
+  def qualifyPath(pathFrom: String): String = if (new File(pathFrom).isAbsolute) pathFrom else  System.getProperty("user.dir") + "/" + pathFrom
```
cjllanwarne: Could possible replace this method with `File.currentWorkingDirectory.relativize(File("/your/absolute/path"))`?

danbills: Wouldn't that go from absolute -> relative when you intend to go relative -> absolute?

danbills: Oh I see.  Then you could standardize on relative paths.

cwl/src/main/scala/cwl/CwlWomExpression.scala
```diff
@@ -94,13 +92,15 @@ final case class InitialWorkDirFileGeneratorExpression(entry: IwdrListingArrayEn
...
-                validate(Await.result(ioFunctionSet.copyFile(source, dest), Duration.Inf))
-              }
-            newFileArray map { nfa => WomArray(WomArrayType(WomSingleFileType), nfa) }
+          case array: WomArray if WomArrayType(WomSingleFileType).coercionDefined(array) => WomArrayType(WomSingleFileType).coerceRawValue(array).toErrorOr
+//            val newFileArray: ErrorOr[List[WomFile]] =
+//              array.value.map(_.valueString).toList.traverse[ErrorOr, WomSingleFile] { source: String =>
+//                val dest = source.split('/').last
+//                validate(Await.result(ioFunctionSet.copyFile(source, dest), Duration.Inf))
+//              }
+//            newFileArray map { nfa => WomArray(WomArrayType(WomSingleFileType), nfa) }
```
mcovarr: 👁

backend/src/main/scala/cromwell/backend/wdl/WriteFunctions.scala
```diff
@@ -27,8 +32,21 @@ trait WriteFunctions extends IoFunctionSet with AsyncIoFunctions {
...
+
+  def relativeToAbsolutePath(pathFrom: String, papi: Boolean): String = if (new File(pathFrom).isAbsolute) pathFrom else {
+    val result = withSlash(if (papi) relativeToPapi else relativeToLocal) + pathFrom
+    println(s"RelativeToAbsolute of $pathFrom was $result")
+    result
+  }
+
   override def copyFile(pathFrom: String, targetName: String): Future[WomSingleFile] = {
-    val source = _writeDirectory.root / pathFrom
+    val papi = buildPath(".").toString.startsWith("gs")
```
kshakir: ToL: I will be _strongly_ encouraging each cloud vendor to also contribute tests with their code. OSS tests are still incoming...

Otherwise we'll have no idea if/when problems arise. Eventually if we don't allow the teams to pass in a method variable, eventually multiple methods like this one will contain local variables committed like:

```scala
val papi = buildPath(".").toString.startsWith("gs")
val oss = buildPath(".").toString.startsWith("oss")
val s3 = buildPath(".").toString.startsWith("s3")
val azureBlob = buildPath(".").toString.contains(".blob.core.windows.net")
…
```

Horneth: Is this only to get the conformance tests working or is it more than that ?
If it's just for the conformance test I wish there was a less disruptive way to do it :(

kshakir: afaict these inline variables are "get papi-conformance tests working". I _hope_ we eventually see PRs for "get tes-conformance tests working", "get azure-conformance tests working", "get bcs-conformance tests working", "get aws-conformance tests working", etc. otherwise nobody knows if cromwell still works with those backends. 🤷‍♀️

Horneth: totally agree, I was just saying that I would really like to see if there's really no other way than ending up with stuff like 

```
val papi = buildPath(".").toString.startsWith("gs")
val oss = buildPath(".").toString.startsWith("oss")
val s3 = buildPath(".").toString.startsWith("s3")
val azureBlob = buildPath(".").toString.contains(".blob.core.windows.net")
```

all over the place

kshakir: Also, to be clearer, these contributor-specific additions will continue to be added to this generic `cromwell.backend.wdl` namespace.

@mcovarr's use of similar jes-conformance-config-injection are isolated within `cromwell.backend.impl.jes`.

https://github.com/broadinstitute/cromwell/blob/1a429769544021c4ddb45a81e1096f52a0988c55/supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesParameters.scala#L17-L22

backend/src/main/scala/cromwell/backend/wdl/WriteFunctions.scala
```diff
@@ -27,8 +32,21 @@ trait WriteFunctions extends IoFunctionSet with AsyncIoFunctions {
...
   }
 
+  private val relativeToLocal = System.getProperty("user.dir")
+  private val relativeToPapi = ConfigFactory.load().as[Option[String]]("papi.default-input-gcs-prefix").orElse(ConfigFactory.load().as[Option[String]]("backend.providers.JES.config.root")).getOrElse("gs://")
+
+  private def withSlash(str: String) = if (str.endsWith("/")) str else str + "/"
+
+  def relativeToAbsolutePath(pathFrom: String, papi: Boolean): String = if (new File(pathFrom).isAbsolute) pathFrom else {
+    val result = withSlash(if (papi) relativeToPapi else relativeToLocal) + pathFrom
+    println(s"RelativeToAbsolute of $pathFrom was $result")
```
Horneth: ~~println~~

backend/src/main/scala/cromwell/backend/wdl/WriteFunctions.scala
cjllanwarne: FWIW this comes up because the conformance tests aren't meant for cloud platforms and server mode. If we did the fully-qualifying in the conformance runner (for JES, not for local) rather than crowbarring it here I'd be so much happier.

kshakir: What about adding a:

```scala
package wom.expression
trait IoFunctionSet {
  def toAbsolute(path: String): String
```

Backend contributors already need to extend IoFunctionSet, and could add any absolute-path in an override of this method.

backend/src/main/scala/cromwell/backend/wdl/WriteFunctions.scala
```diff
@@ -27,8 +32,21 @@ trait WriteFunctions extends IoFunctionSet with AsyncIoFunctions {
     }
   }
 
+  private val relativeToLocal = System.getProperty("user.dir")
+  private val relativeToPapi = ConfigFactory.load().as[Option[String]]("papi.default-input-gcs-prefix").orElse(ConfigFactory.load().as[Option[String]]("backend.providers.JES.config.root")).getOrElse("gs://")
```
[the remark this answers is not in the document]
mcovarr: like an uncle?

backend/src/main/scala/cromwell/backend/wdl/WriteFunctions.scala
```diff
@@ -27,8 +32,15 @@ trait WriteFunctions extends IoFunctionSet with AsyncIoFunctions {
     }
   }
 
+  private val relativeToLocal = {
+    val initial = System.getProperty("user.dir")
+    if (initial.endsWith("/")) initial else initial + "/"
```
mcovarr: I feel like I've seen 37 different versions of this line in various PRs recently

src/bin/travis/resources/conformance_expected_failures.txt
```diff
@@ -49,7 +49,7 @@
 #49
 #50
 51
-52
+#52
```
mcovarr: Interesting... I had this test go green for me too on Local by just fully qualifying the value for the `path` attribute. 🤔

backend/src/main/scala/cromwell/backend/wdl/WriteFunctions.scala
```diff
@@ -27,8 +30,15 @@ trait WriteFunctions extends IoFunctionSet with AsyncIoFunctions {
...
+  }
+
+  def relativeToAbsolutePath(pathFrom: String): String = if (new File(pathFrom).isAbsolute) pathFrom else relativeToLocal + pathFrom
```
Horneth: Could just add this as an "ioFunction" ? I need exactly this somewhere else for `ExpressionTool`s

PR #3152, opened 2018-01-17
Add support for cwl.output.json

Horneth:
- Allows for empty globs by creating a control file in the glob directory and filtering it out from the `.list` file containing the list of globbed files
- Use the above to add an optional glob file in the task definition that CWL will populate with `cwl.output.json`
- Add a "short-circuit" output evaluation method in the task definition that gives a chance to CWL to look for `cwl.output.json` and use it as the output of the tool. "short-circuit" because normally Cromwell goes over each output port of the task and evaluates their expression in order to evaluate the outputs of a call node. This will be the fallback behavior if the `cwl.output.json` happens not to be there (and will stay the current and only behavior for all WDL workflows).


cjllanwarne:
If this does what I think it does, can we now add a WDL `"works for empty globs"` test?

Horneth:
@cjllanwarne done

backend/src/main/scala/cromwell/backend/BackendJobExecutionActor.scala
```diff
@@ -107,7 +108,7 @@ trait BackendJobExecutionActor extends BackendJobLifecycleActor with ActorLoggin
   }
 
   def evaluateOutputs(wdlFunctions: IoFunctionSet,
-                      postMapper: WomValue => Try[WomValue] = v => Success(v)): EvaluatedJobOutputs = {
-    OutputEvaluator.evaluateOutputs(jobDescriptor, wdlFunctions, postMapper)
+                      postMapper: WomValue => Try[WomValue] = v => Success(v))(implicit ec: ExecutionContext): EvaluatedJobOutputs = {
+    Await.result(OutputEvaluator.evaluateOutputs(jobDescriptor, wdlFunctions, postMapper), Duration.Inf)
```
Horneth: This is unfortunate but is only used by the SparkBackend AFAICT

cwl/src/main/scala/cwl/CommandLineTool.scala
```diff
@@ -439,7 +503,8 @@ object CommandLineTool {
                                      `type`: Option[MyriadOutputType] = None)
 
   object CommandOutputParameter {
-
+    import cats.instances.list._
+    import cats.instances.option._
```
mcovarr: are these new imports being used?

backend/src/main/scala/cromwell/backend/wdl/OutputEvaluator.scala
```diff
@@ -75,10 +76,29 @@ object OutputEvaluator {
...
+    def validateCustomEvaluation(outputs: Map[OutputPort, WomValue]): EvaluatedJobOutputs = {
+      def toError(outputPort: OutputPort) = s"Missing output value for ${outputPort.identifier.fullyQualifiedName.value}"
+
+      jobDescriptor.taskCall.expressionBasedOutputPorts.diff(outputs.keySet.toList) match {
+        case Nil => ValidJobOutputs(CallOutputs(outputs))
+        case head :: tail => InvalidJobOutputs(NonEmptyList.of(toError(head), tail.map(toError): _*))
+      }
+    }
+
+    // Try the custom evaluation function
```
mcovarr: insert docs from description ☝️ 😄

mcovarr:
The last point in your description on this PR is the key to understanding what's going on here; IMHO something like that should end up in a comment in the code.

cwl/src/main/scala/cwl/CommandLineTool.scala
Horneth: yes, there was some weird ambiguity error between several cats implicits so I split them up

PR #3153, opened 2018-01-17
Backend secondary file and directory updates.

kshakir:
Updated some backend usages of WomSingleFile(womFile.value) with womFile.mapFile(_).
CWL glob creation temporarily creating WomSingleFile if the path doesn't contain '*' or '?'.
When generating secondary paths without '^' be sure to prefix the primary path.
Added directory outputs for un-nested directories.
Directory inputs are also incomplete, specifically paths don't localize to a user-specified parent directory.
IntelliJ suggested edits to BcsAsyncBackendJobExecutionActor.
Globs are now processed by CwlJsEncoder.
Added comments to CommandOutputBinding.generateOutputWomValue/coerceWomValue.
Bug fixes to CWL javascript directory encoding/decoding.
MetadataDatabaseAccessSpec uses unique names so that it can be run twice on the same db.


cwl/src/main/scala/cwl/CommandOutputBinding.scala
```diff
@@ -19,6 +19,24 @@ case class CommandOutputBinding(glob: Option[Glob] = None,
...
+   * TODO: WOM: WOMFILE: Need to support returning globs for primary and secondary.
+   *
+   * Right now, when numerous glob files are created, the backends place each glob result into a different subdirectory.
+   * Later, when the secondary files are being generated each of their globs are in different sub-directories.
+   * Unfortunately, the secondary file resolution assumes that the secondaries are _next to_ the primaries, not in
+   * sibling directories.
+   *
+   * While this is being worked on, instead of looking up globs return regular files until this can be fixed.
+   */
+  def dependingIfGlobLike[A](path: String, isGlobLike: => A, isNotGlobLike: => A): A = {
```
danbills: ToL IMO this should be `isGlobLike` and return a boolean and let user use `if`/`fold` semantics

cwl/src/main/scala/cwl/CommandOutputBinding.scala
```diff
@@ -40,22 +58,27 @@ object CommandOutputBinding {
...
-        case WomUnlistedDirectoryType => primaryPaths.map(WomUnlistedDirectory).valid
-        case other => s"Program error: $other type was not expected".invalidNel
-      }
-      secondaryWomFiles <- primaryWomFiles.flatTraverse[ErrorOr, WomFile] {
-        CommandLineTool.CommandOutputParameter.secondaryFiles(_, secondaryFilesOption, parameterContext, expressionLib)
+      primaryWomFiles <- primaryPaths
+        .map(primaryPath => WomFile(outputWomFlatType(primaryPath), primaryPath))
+        .valid
+        : ErrorOr[List[WomFile]]
+      secondaryWomFiles <- primaryWomFiles.flatTraverse[ErrorOr, WomFile] {primaryWomFile =>
```
danbills: ToL consider breaking this out into a local function as it's a bit hard to read

cwl/src/main/scala/cwl/CommandOutputBinding.scala
```diff
@@ -40,22 +58,27 @@ object CommandOutputBinding {
...
+    It turns out that secondary files can be directories. TODO: WOM: WOMFILE: Not sure what to do with that yet.
+    https://github.com/common-workflow-language/common-workflow-language/blob/master/v1.0/v1.0/search.cwl#L42
+    https://github.com/common-workflow-language/common-workflow-language/blob/master/v1.0/v1.0/index.py#L36-L37
      */
-    val outputWomFlatType = outputWomType match {
+    def outputWomFlatType(path: String) = outputWomType match {
       case WomMaybeListedDirectoryType => WomUnlistedDirectoryType
       case WomArrayType(WomMaybeListedDirectoryType) => WomUnlistedDirectoryType
-      case _ => WomGlobFileType
+      case _ => dependingIfGlobLike(path, WomGlobFileType, WomSingleFileType)
```
danbills: big fan of the pragmatic approach here.

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -148,6 +149,30 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
...
+    *
+    * @param unlistedDirectory The directory.
+    * @return The shell scripting.
+    */
+  def directoryScript(unlistedDirectory: WomUnlistedDirectory): String = {
+    val directoryPath = DirectoryFunctions.ensureUnslashed(unlistedDirectory.value)
+    val directoryList = directoryPath + ".list"
+
+    s"""|# list all the files that match the directory into a file called directory.list
+        |find $directoryPath -type f > $directoryList
```
mcovarr: `find` dependency 😮

backend/src/main/scala/cromwell/backend/io/DirectoryFunctions.scala
```diff
@@ -30,15 +42,26 @@ trait DirectoryFunctions extends IoFunctionSet with PathFactory {
...
+          pathsPerListing <-
+            uncheckedPaths.toList.traverse[ErrorOr, List[Path]](listPaths(_, newCheckedPaths).map(_.toList))
+        } yield checkedPaths ++ pathsPerListing.flatten.toSet
+      } else {
+        newCheckedPaths.valid
+      }
     }
+
+    val allPaths = listPaths(path, Set.empty).map(_.toList)
+    val allFiles = allPaths.map(_.filterNot(_.isDirectory))
```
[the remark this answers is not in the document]
kshakir: No other file types are explicitly specified for special behavior by the cwl or wdl specs. AFAIK symlinks, pipes, etc are all "undefined behavior".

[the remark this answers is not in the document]
geoffjentry: It’s worth a ping on their gitter to double check that this isn’t going to bite us

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
kshakir: `find` is in [POSIX](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/find.html) so it's still available in docker images for busybox (and hence alpine), etc.

But if we attempt to use [`globstar`](https://www.gnu.org/software/bash/manual/html_node/The-Shopt-Builtin.html) to work around deeply nested directories, globstar would be bash only.

https://github.com/common-workflow-language/common-workflow-language/blob/7b831fe0b350af27dafb50ec89fd6411922fec39/v1.0/v1.0/iwdr_with_nested_dirs.cwl#L15-L20

geoffjentry: Fwiw I’m pretty sure someone using a bashless Docker is going to have a bad time anyways, unless that’s changed and I didn’t realize?

kshakir: I don't know what's coming in PAPIv2. If they remove their bash requirements, cromwell has exactly one non-POSIX requirement left: the string `bash` at the top of our `exec.sh` scripts.

On a UX front though there is the expectation of WDL authors that they can use bash-isms within their `command` blocks. But that could be mitigated w/ yet-another-config/runtimeAttribute/knob.

geoffjentry: I believe that’s gone with v2. Re the WDL that could go to being a wdl problem instead of a Cromwell problem

PR #3158, opened 2018-01-18
Use preserveScatterIndex for inner graph OGINs

Horneth:
In a configuration like

```wdl
workflow ifs_in_scatters {
  call hello

  scatter (n in range(5)) {
    if (true) {
      call goodbye { input: i = hello.out }
    }
  }
}
```
When the conditional graph is created, all nodes outside of the scatter get "wrapped" in an `OuterGraphInputNode` that gets passed into the inner conditional graph so that nodes in the inner graph can reference nodes outside of the scatter.
The issue is that those OGINs are created with `preserveScatterIndex  = true` even though the node they're pointing to is outside of the scatter. This was preventing the `ExecutionStore` from detecting the `i = hello.out` expression as being runnable because it was looking for a `hello.out` node in `Done` state at index `n`, which doesn't exist since `call hello` is outside the scatter.
This PR changes that to use the `preserveIndexForOuterLookups ` value of the conditional / scatter instead, which in this case will be `false`, because the scatter node does set `preserveScatterIndex = false` to build its inner graph (in this case the if).

### DIFF
--- a/wdl/src/main/scala/wdl/If.scala
+++ b/wdl/src/main/scala/wdl/If.scala
@@ -56,6 +56,15 @@ object If {
       * use however many of them it needs.
       */
     val possiblyNeededNestedOgins: Map[String, OuterGraphInputNode] = outerLookup filterNot { case (name, _) => localLookup.contains(name) } map { case (name, outerPort) =>
+      /*
+        * preserveIndexForOuterLookups indicates us whether or not nodes in the outerLookup should be considered "siblings" of the conditional in term of scatter index
+        * preserveIndexForOuterLookups = false means the outerLookup nodes are outside a scatter containing this conditional node
+        * preserveIndexForOuterLookups = true means the above predicate does not hold
+        * 
+        * When creating OGINs from those outer lookup nodes for the inner graph we want to make sure we set their preserveScatterIndex to preserveIndexForOuterLookups
+        * because they effectively represent the outer lookup nodes inside the conditional. So whether the index must be preserved depends on whether this
+        * conditional node has been asked to "preserveIndexForOuterLookups".
+       */
       name -> OuterGraphInputNode(WomIdentifier(name), outerPort, preserveScatterIndex = preserveIndexForOuterLookups)
     }
     val possiblyNeededNestedOginPorts: Map[String, OutputPort] = possiblyNeededNestedOgins map { case (name: String, ogin: OuterGraphInputNode) => name -> ogin.singleOutputPort }
--- a/wdl/src/main/scala/wdl/If.scala
+++ b/wdl/src/main/scala/wdl/If.scala
@@ -56,7 +56,7 @@ object If {
       * use however many of them it needs.
       */
     val possiblyNeededNestedOgins: Map[String, OuterGraphInputNode] = outerLookup filterNot { case (name, _) => localLookup.contains(name) } map { case (name, outerPort) =>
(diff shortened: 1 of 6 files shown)


mcovarr:
:+1:

wdl/src/main/scala/wdl/If.scala
```diff
@@ -56,7 +56,16 @@ object If {
       * use however many of them it needs.
       */
     val possiblyNeededNestedOgins: Map[String, OuterGraphInputNode] = outerLookup filterNot { case (name, _) => localLookup.contains(name) } map { case (name, outerPort) =>
-      name -> OuterGraphInputNode(WomIdentifier(name), outerPort, preserveScatterIndex = true)
+      /*
+        * preserveIndexForOuterLookups indicates us whether or not nodes in the outerLookup should be considered "siblings" of the conditional in term of scatter index
```
mcovarr: not quite English

wdl/src/main/scala/wdl/If.scala
```diff
@@ -56,7 +56,16 @@ object If {
...
+        * preserveIndexForOuterLookups = false means the outerLookup nodes are outside a scatter containing this conditional node
+        * preserveIndexForOuterLookups = true means the above predicate does not hold
+        * 
+        * When creating OGINs from those outer lookup nodes for the inner graph we want to make sure we set their preserveScatterIndex to preserveIndexForOuterLookups
+        * because they effectively represent the outer lookup nodes inside the conditional. So whether the index must be preserved depends on whether this
+        * conditional node has been asked to "preserveIndexForOuterLookups".
+       */
```
mcovarr: excellent comments! 🏆

wdl/src/main/scala/wdl/Scatter.scala
```diff
@@ -60,7 +60,20 @@ object Scatter {
...
+        * 
+        * When creating OGINs from those outer lookup nodes for the inner graph we want to make sure we set their preserveScatterIndex to preserveIndexForOuterLookups
+        * because they effectively represent the outer lookup nodes inside the scatter. So whether the index must be preserved depends on whether this
+        * scatter node has been asked to "preserveIndexForOuterLookups".
+        * 
+        * Note that preserveIndexForOuterLookups will always be true here as long as scatters can't be nested. Indeed the only reason for a node to be
+        * asked to NOT preserve the index is if its outerLookup is referencing nodes outside a scatter.
+        * In this case preserveIndexForOuterLookups = false would mean this scatter node is part of the inner graph of another scatter node.
+       */
+      name -> OuterGraphInputNode(WomIdentifier(name), outerPort, preserveScatterIndex = preserveIndexForOuterLookups)
```
mcovarr: Discussed with faces, since this is a scatter it shouldn't be possible to nest inside another scatter so OGIN indices should always turn out to be -1, right @cjllanwarne?

cjllanwarne: Yes, but I think this is a more general version of that (it gives us the right answer now, and also in case we add nested scatters in the future)

Horneth: @cjllanwarne @mcovarr 
So looking into this more, regardless of what we put here it won't fully work for nested scatters anyway.
From a conceptual perspective I think the "right" thing to do is to set it to `false`, even though it turns out the actual value does not matter in practice for reasons too complicated to explain here.
Happy to whiteboard this with whoever wants.

PR #3159, opened 2018-01-18
[30_hotfix] Workflow stuck

Horneth:


wdl/src/main/scala/wdl/If.scala
```diff
@@ -56,7 +56,7 @@ object If {
       * use however many of them it needs.
       */
     val possiblyNeededNestedOgins: Map[String, OuterGraphInputNode] = outerLookup filterNot { case (name, _) => localLookup.contains(name) } map { case (name, outerPort) =>
-      name -> OuterGraphInputNode(WomIdentifier(name), outerPort, preserveScatterIndex = true)
+      name -> OuterGraphInputNode(WomIdentifier(name), outerPort, preserveScatterIndex = preserveIndexForOuterLookups)
```
mcovarr: comments plz

wdl/src/main/scala/wdl/Scatter.scala
```diff
@@ -60,7 +60,7 @@ object Scatter {
       * use however many of them it needs.
       */
     val possiblyNeededNestedOgins: Map[String, OuterGraphInputNode] = outerLookup filterNot { case (name, _) => localLookup.contains(name) } map { case (name, outerPort) =>
-      name -> OuterGraphInputNode(WomIdentifier(name), outerPort, preserveScatterIndex = false)
+      name -> OuterGraphInputNode(WomIdentifier(name), outerPort, preserveScatterIndex = preserveIndexForOuterLookups)
```
mcovarr: here too

wdl/src/main/scala/wdl/If.scala
```diff
@@ -56,7 +56,16 @@ object If {
...
+      /*
+        * preserveIndexForOuterLookups indicates us whether or not nodes in the outerLookup should be considered "siblings" of the conditional in term of scatter index
+        * preserveIndexForOuterLookups = false means the outerLookup nodes are outside a scatter containing this conditional node
+        * preserveIndexForOuterLookups = true means the above predicate does not hold
+        * 
+        * When creating OGINs from those outer lookup nodes for the inner graph we want to make sure we set their preserveScatterIndex to preserveIndexForOuterLookups
+        * because they effectively represent the outer lookup nodes inside the conditional. So whether the index must be preserved depends on whether this
+        * conditional node has been asked to "preserveIndexForOuterLookups".
+       */
+      name -> OuterGraphInputNode(WomIdentifier(name), outerPort, preserveScatterIndex = preserveIndexForOuterLookups)
```
cjllanwarne: Hmm, this change surprises me but I'd need to look again to remind myself how it all works.

If we do this then won't `preserveScatterIndex` gets fixed at false for everything inside a scatter, even if the scopes are siblings? eg

```wdl
scatter(i in [1,2,3,4,5]) {
  if(x) {
    Int a = i # we want preserveScatterIndex = true for i's ogin in if(y) even though we're inside a scatter
    if(y) {
      Int b = a + i 
    }
  }
}
```

Horneth: > we want preserveScatterIndex = true for i's ogin in if(y) even though we're inside a scatter
It will, because `preserveIndexForOuterLookups` will be `true` for `if(y)` since it was built as an inner node of `if(x)`, and `If`s build their inner nodes with  `preserveScatterIndex = true`: https://github.com/broadinstitute/cromwell/blob/develop/wdl/src/main/scala/wdl/If.scala#L67

PR #3162, opened 2018-01-18
Treat a message 13 as  premptible when the VM is preemptible

geoffjentry:


supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/StatusPolling.scala
```diff
@@ -74,15 +74,26 @@ private[statuspolling] object StatusPolling {
...
+          for {
+            request <- op.getMetadata.asScala.get("request")
+            pipelineArgs <- request.asInstanceOf[GArrayMap[String, Object]].asScala.get("pipelineArgs")
+            resources <- pipelineArgs.asInstanceOf[GArrayMap[String, Object]].asScala.get("resources")
+            preemptible <- resources.asInstanceOf[GArrayMap[String, Object]].asScala.get("preemptible")
+          } yield preemptible.asInstanceOf[Boolean]
+        } getOrElse false
+
         // If there's an error, generate an unsuccessful status. Otherwise, we were successful!
+        // metadata.pipelineArgs.resources.preemptible
```
mcovarr: it seems to say "request.pipelineArgs.resources.preemptible" above?

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/RunStatus.scala
```diff
@@ -49,11 +49,13 @@ object RunStatus {
               eventList: Seq[ExecutionEvent],
               machineType: Option[String],
               zone: Option[String],
-              instanceName: Option[String]): UnsuccessfulRunStatus = {
+              instanceName: Option[String],
+              isPreemptible: Boolean): UnsuccessfulRunStatus = {
```
cjllanwarne: ToL: `wasPreemptible`?

Horneth:
I think you could get away with just adding a new case

```scala
case failedStatus: RunStatus.UnsuccessfulRunStatus if preemptible && failedStatus.jesCode.contains(JesAsyncBackendJobExecutionActor.JesUnexpectedTermination) => handleFailedRunStatus(failedStatus, handle, returnCode)
```

here : https://github.com/broadinstitute/cromwell/blob/6fd20e70cdd5f3f2f080dbad81d2264d456fe14e/supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActor.scala#L483

But this should work too 👍

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/StatusPolling.scala
cjllanwarne: Looks like leftover copy/paste (or else it's in the wrong place)

supportedBackends/jes/src/test/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActorSpec.scala
```diff
@@ -200,7 +200,7 @@ class JesAsyncBackendJobExecutionActorSpec extends TestKitSuite("JesAsyncBackend
 
   private def runAndFail(previousPreemptions: Int, previousUnexpectedRetries: Int, preemptible: Int, errorCode: Status, innerErrorMessage: String, expectPreemptible: Boolean): BackendJobExecutionResponse = {
 
-    val runStatus = UnsuccessfulRunStatus(errorCode, Option(innerErrorMessage), Seq.empty, Option("fakeMachine"), Option("fakeZone"), Option("fakeInstance"))
+    val runStatus = UnsuccessfulRunStatus(errorCode, Option(innerErrorMessage), Seq.empty, Option("fakeMachine"), Option("fakeZone"), Option("fakeInstance"),false)
```
cjllanwarne: add space: `, false)`

supportedBackends/jes/src/test/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActorSpec.scala
```diff
@@ -286,7 +286,7 @@ class JesAsyncBackendJobExecutionActorSpec extends TestKitSuite("JesAsyncBackend
     val runId = StandardAsyncJob(UUID.randomUUID().toString)
     val handle = new JesPendingExecutionHandle(null, runId, None, None)
 
-    val failedStatus = UnsuccessfulRunStatus(Status.ABORTED, Option("14: VM XXX shut down unexpectedly."), Seq.empty, Option("fakeMachine"), Option("fakeZone"), Option("fakeInstance"))
+    val failedStatus = UnsuccessfulRunStatus(Status.ABORTED, Option("14: VM XXX shut down unexpectedly."), Seq.empty, Option("fakeMachine"), Option("fakeZone"), Option("fakeInstance"), true)
```
cjllanwarne: `preemptible` is already coming in to the test case:
```
expectations foreach { case (..., ..., preemptible, ..., ..., ..., ...) =>
```
... so I'd suggest using that (esp as it would mean we'd actually be testing this new behaviour correctly too... hooray!)

supportedBackends/jes/src/test/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActorSpec.scala
```diff
@@ -315,7 +315,22 @@ class JesAsyncBackendJobExecutionActorSpec extends TestKitSuite("JesAsyncBackend
...
```
cjllanwarne: Per comment above, could we get this as part of the table tests by using the existing `preemptible` attribute?

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/StatusPolling.scala
geoffjentry: oh, that was a note i made to myself whilst talking to ruchi

supportedBackends/jes/src/test/scala/cromwell/backend/impl/jes/JesAsyncBackendJobExecutionActorSpec.scala
geoffjentry: discussed in person, a lot of these tests don't work the way you were thinking

geoffjentry:
@Horneth I don't think so, although I'm pretty sure what would work would be to add a case similar to the one on 481, but catching `UnsuccessfulRunStatus` w/ the isPreemptible if. I didn't originally go down that road because at the time I thought it wouldn't work, but I misunderstood how things were working and I'd forgotten about it by the end. I'm going to give that a whirl as it'd be a lot easier than digging into the op metadata

geoffjentry:
@Horneth although we're already mapping something else to preemptible in the spot where i'm doing it now, so this would mean we're mapping "other" errors to preemptible in 2 different spots

geoffjentry:
@Horneth I think what I'm going to do is leave it as-is but add a comment to those mappings that these should go away w/ PAPI2 (at least that's what I was told)

cjllanwarne:
Thinking about this I'm more and more confused why none of the existing test cases are now triggering as "yes we should try to run this again"

PR #3168, opened 2018-01-19
Start is not required on query results

calbach:
Unless queryWorkflows **only** returns jobs that have already started, I think it cannot be true that this field is required.

### DIFF
--- a/docs/api/RESTAPI.md
+++ b/docs/api/RESTAPI.md
@@ -864,7 +864,7 @@ Result for an individual workflow returned by a workflow query
 |**end**  <br>*optional*|Workflow end datetime|string (date-time)|
 |**id**  <br>*required*|Workflow ID|string|
 |**name**  <br>*required*|Workflow name|string|
-|**start**  <br>*required*|Workflow start datetime|string (date-time)|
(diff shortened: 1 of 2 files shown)


docs/api/RESTAPI.md
```diff
@@ -864,7 +864,7 @@ Result for an individual workflow returned by a workflow query
 |**end**  <br>*optional*|Workflow end datetime|string (date-time)|
 |**id**  <br>*required*|Workflow ID|string|
 |**name**  <br>*required*|Workflow name|string|
-|**start**  <br>*required*|Workflow start datetime|string (date-time)|
+|**start**  <br>*optional*|Workflow start datetime|string (date-time)|
```
katevoss: @calbach Thanks for your contribution! 
For future notice, we generate the `RESTAPI.md` file from the `cromwell.yaml` file by running `sbt generateRestApiDocs` in the Cromwell top-level directory.

If you had only updated the `RESTAPI.md` without updating `cromwell.yaml`, your changes would be overwritten. But since you updated both files your edits will not be overwritten.

katevoss:
👍

PR #3170, opened 2018-01-19
Kv wom docs

katevoss:


engine/src/main/resources/swagger/cromwell.yaml
```diff
@@ -27,8 +27,7 @@ paths:
...
-          description: JSON or YAML file containing the inputs as an object. For WDL workflows a skeleton file can be generated from wom
-          tool using the "inputs" subcommand. When multiple files are specified, in case of key conflicts between multiple input JSON files, higher values of x in workflowInputs_x override lower values. For example, an input specified in workflowInputs_3 will override an input with the same name in workflowInputs or workflowInputs_2. Similarly, an input key specified in workflowInputs_5 will override an identical input key in any other input file.
+          description: JSON or YAML file containing the inputs as an object. For WDL workflows a skeleton file can be generated from womtool using the "inputs" subcommand. When multiple files are specified, in case of key conflicts between multiple input JSON files, higher values of x in workflowInputs_x override lower values. For example, an input specified in workflowInputs_3 will override an input with the same name in workflowInputs or workflowInputs_2. Similarly, an input key specified in workflowInputs_5 will override an identical input key in any other input file.
```
Horneth: Should it be `WOMTool` here too ?

katevoss: Yep, you're right. I'll fix it in the other places too.

PR #3172, opened 2018-01-20
Errant linebreak in swagger yaml busted the UI

geoffjentry:


geoffjentry:
I'm gonna just merge this as the sum total was the removal of a newline and it has the side effect of un-borking the swagger ui

PR #3175, opened 2018-01-22
WDL: Object literals, read_json and write_json. Closes #1825

### LINKED ISSUE #1825: Implement read_json, write_json (at least partially)
A user (@helgridly) was flummoxed when they tried to use `read_json` from a python task.

They had designed a method around the idea of being able to serialize and deserialize to/from JSON and had a bad experience at the last minute when these functions turned out to not be implemented. The less-than-ideal error message did not help matters.

Although the general-purpose `read_json` might be dangerous, we could at least support something like `read_json_array`.

cjllanwarne:
A big ask at BOSC:

Allows Cromwell to read `object { ... }` literals in WDL, to call `read_json(file)`, and to call `write_json(object)`.

Even if the `Object` type fades away after the structs PR, these building blocks are still going to be needed.

### DIFF
--- a/centaur/src/main/resources/standardTestCases/read_write_json.test
+++ b/centaur/src/main/resources/standardTestCases/read_write_json.test
@@ -1,0 +1,12 @@
+name: read_write_json
+testFormat: workflowsuccess
+
+files {
+  wdl: read_write_json/read_write_json.wdl
+}
+
+metadata {
+  workflowName: read_write_json
+  status: Succeeded
+  "outputs.read_write_json.result" = "27.5 was equal to 27.5"
+}
--- a/centaur/src/main/resources/standardTestCases/read_write_json/read_write_json.wdl
+++ b/centaur/src/main/resources/standardTestCases/read_write_json/read_write_json.wdl
@@ -1,0 +1,63 @@
+workflow read_write_json {
+  Object json_object = object {
+    str: "hi",
+    int: 57,
+    float: 27.5,
+    pair: (5, "hello"),
+    array: ["a", "b", "c"]
+  }
+
+  call make_some_json
+  call round_trip { input: to_jsonify = json_object }
+
+  if (round_trip.round_tripped.float == make_some_json.output_json.float) {
+    call success { input: actual = round_trip.round_tripped.float, expected = make_some_json.output_json.float }
+  }
+
+  output {
+    String? result = success.result
+  }
+}
+
+task round_trip {
+  Object to_jsonify
+  File jsonified = write_json(to_jsonify)
+  command {
+    cp ${jsonified} output.json
+  }
+  output {
+    Object round_tripped = read_json("output.json")
+  }
+}
+
+task make_some_json {
+  command <<<
+    echo '{'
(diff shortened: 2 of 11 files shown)


centaur/src/main/resources/standardTestCases/read_write_json/read_write_json.wdl
```diff
@@ -0,0 +1,69 @@
+workflow read_write_json {
+  Object json_object = object {
+    str: "hi",
+    int: 57,
+    float: 27.5,
+    pair: (5, "hello"),
+    array: ["a", "b", "c"]
```
Horneth: should we have a map or object here too ?

centaur/src/main/resources/standardTestCases/read_write_json.test
```diff
@@ -0,0 +1,12 @@
...
+testFormat: workflowsuccess
+
+files {
+  wdl: read_write_json/read_write_json.wdl
+}
+
+metadata {
+  workflowName: read_write_json
+  status: Succeeded
+  "outputs.read_write_json.result" = "28.5 was equal to 27.5"
```
Horneth: `28.5 was equal to 27.5` 😄

wdl/src/main/scala/wdl/expression/ValueEvaluator.scala
```diff
@@ -117,6 +117,15 @@ case class ValueEvaluator(override val lookup: String => WomValue, override val
             WomType.homogeneousTypeFromValues(pairs.map(_._2))
           ).coerceRawValue(pairs.toMap)
         }
+      case a: Ast if a.isObjectLiteral =>
+        val evaluatedMap = a.getAttribute("map").astListAsVector map { kv =>
+          val key = kv.asInstanceOf[Ast].getAttribute("key").sourceString
+          val value = evaluate(kv.asInstanceOf[Ast].getAttribute("value"))
+          key -> value
+        }
+        TryUtil.sequence(evaluatedMap map { tuple => tuple._2.map((tuple._1, _)) }) map { pairs =>
```
jbuabbud: What does this line do?

geoffjentry: Baaaaaaah, @jbuabbud just forget you ever saw `TryUtil` :)

cjllanwarne: @jbuabbud - feel free to ask follow ups...

- `evaluatedMap` starts as a `Map[X, Try[Y]]`
- The `map ( tuple => ...)` turns each tuple of `(X, Try[Y])` in the `Map` into a Try[(X, Y)], which turns the `Map[X, Try[Y]]` into a `Seq[Try[(X, Y)]]`
- The `TryUtil.sequence` turns the `Seq[Try[(X, Y)]]` into a `Try[Seq[(X, Y)]]`. In other words, take a list of things that might have failed and turn it into a single list that might be a failure. It's not a great thing to use (cf @geoffjentry) because if the `Seq` has more than one failure in it, the result will only carry forwards the first one.

PR #3177, opened 2018-01-23
Declare Secondary files types correctly

danbills:
fixes mutect2 fails to parse sore spot


geoffjentry:
:+1:

cjllanwarne:
The two places you've replaced with `SecondaryFiles` actually had different options originally - presumably this is what the PR is supposed to fix, rather than something that's now been lost?

danbills:
@cjllanwarne Yes those original types were incorrect (too narrow).  The declared type is correct in this PR.

PR #3178, opened 2018-01-23
Add valid hostname to DockerHubFlow

ruchim:
Previously, if a docker image was named "docker.io/<repo>/<imageName>:<tag>", there was an issue such that Cromwell failed to pull the docker hash. The job was allowed to run but the results were never saved and so one could never call-cache to jobs with that kind of docker syntax. 

This change is meant to make "docker.io" an accept hostname for the DockerHubFlow.

### DIFF
--- a/centaur/src/main/resources/standardTestCases/docker_hash/docker_hash_dockerhub.wdl
+++ b/centaur/src/main/resources/standardTestCases/docker_hash/docker_hash_dockerhub.wdl
@@ -10,10 +10,10 @@ task dockerhub {
 
 workflow docker_hash_dockerhub {
 
-    Array[String] hostnames = ["", "docker.io/", "index.docker.io/"]
+    call dockerhub as dockerWithoutHost { input: hostname = "" }
 
-    scatter (hostname in hostnames) {
-        call dockerhub { input: hostname = hostname }
-    }
+    call dockerhub as implicitRegistryHost { input: hostname = "docker.io/" }
+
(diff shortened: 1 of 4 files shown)


centaur/src/main/resources/standardTestCases/docker_hash/docker_hash_dockerhub.wdl
```diff
@@ -3,7 +3,7 @@ task dockerhub {
         echo "hello"
     }
     runtime {
-        docker: "ubuntu:precise-20161209"
+        docker: "docker.io/ubuntu:precise-20161209"
```
Horneth: Can you add a task instead to keep testing (even though I'm sure it's tested somewhere else) the case without `docker.io` ?

dockerHashing/src/main/scala/cromwell/docker/registryv2/flows/dockerhub/DockerHubFlow.scala
```diff
@@ -16,6 +16,7 @@ object DockerHubFlow {
   private val DockerHubDefaultRegistry = "index.docker.io"
   private val RegistryHostName = "registry-1.docker.io"
   private val AuthorizationServerHostName = "auth.docker.io"
+  private val DockerHubImplicitRegistryHostName = "docker.io"
```
cjllanwarne: Instead of listing these individually and chaining together `|| host == X`s, can we just make a list and say `if list.contains(host)`?

cjllanwarne:
👍 after a small commentette

PR #3179, opened 2018-01-23
Non-Dockerized standard input redirection.

mcovarr:
Part I of standard input redirection, works on non-Dockerized Local backend only. Thanks to @cjllanwarne for help in not overthinking this part of the problem and also identifying the incorrect value of `path` in CWL files as a separate problem.


backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -267,7 +268,7 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
         |cd $cwd
         |ENVIRONMENT_VARIABLES
         |INSTANTIATED_COMMAND
-        |) > >(tee $stdoutPath) 2> >(tee $stderrPath >&2)
+        |) $stdinRedirection > >(tee $stdoutPath) 2> >(tee $stderrPath >&2)
```
cjllanwarne: ToL: The `> >` surprised me but since I realised it's not part of your change... 🤷‍♂️

wom/src/main/scala/wom/package.scala
```diff
@@ -31,6 +31,13 @@ package object values {
     def md5Sum: String = DigestUtils.md5Hex(value)
     def md5SumShort: String = value.md5Sum.substring(0, 8)
   }
+
+  implicit class ShellQuoteHelper(val s: String) extends AnyVal {
+    // Escape any double quote characters in the string and surround with double quotes. The CWL spec doesn't
```
cjllanwarne: I'm curious about this relocation. This seems more `CWL` than `WOM` to me.

cjllanwarne:
👍 although I'm curious why you chose to move CWL code into WOM... :)

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
mcovarr: Yes I was surprised I didn't catch more grief for that when I actually made the change. 🙂  That syntax enables those tees to both stdout/stderr and name stdout/stderr files.

wom/src/main/scala/wom/package.scala
mcovarr: I actually didn't need to move this as part of these changes but it seemed like a good idea. As the comments say this escapes double quotes and puts double quotes around strings, which is useful for elements of a command instantiation. We should probably be using something like this more liberally to support filenames with spaces.

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -316,14 +319,19 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
...
     def evaluateEnvironmentExpression(nameAndExpression: (String, WomExpression)): ErrorOr[(String, String)] = {
       val (name, expression) = nameAndExpression
-      expression.evaluateValue(adHocFileCreationInputs, backendEngineFunctions) map { name -> _.valueString }
+      expression.evaluateValue(commandInstantiationInputs, backendEngineFunctions) map { name -> _.valueString }
     }
 
-    val environmentVariables = jobDescriptor.taskCall.callable.environmentExpressions.toList traverse evaluateEnvironmentExpression
+    val environmentVariables = callable.environmentExpressions.toList traverse evaluateEnvironmentExpression
+
+    import cats.instances.option._
```
danbills: ToL I don't feel strongly but I personally like these to live at the top of the file, despite the broader scope.

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -289,16 +290,18 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
...
         case f: WomFile => List(f).valid
         case a: WomArray => a.value.toList.traverse(validateAdHocFile).map(_.flatten)
         case other => s"Ad-hoc file creation expression invalidly created a ${other.womType.toDisplayString} result.".invalidNel
     }
 
-    val adHocFileCreations: ErrorOr[List[WomFile]] = jobDescriptor.taskCall.callable.adHocFileCreation.toList.traverse(
-      _.evaluateValue(adHocFileCreationInputs, backendEngineFunctions).flatMap(validateAdHocFile)
+    val callable = jobDescriptor.taskCall.callable
+
+    val adHocFileCreations: ErrorOr[List[WomFile]] = callable.adHocFileCreation.toList.traverse(
```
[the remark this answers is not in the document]
mcovarr: It could. I had it that way in my more aggressive set of changes but I should have kept this part in.

PR #3180, opened 2018-01-23
Improve parsing error messages

danbills:
## What it does do
Gives you the field that does not parse.  (This is better than the previous `CanBuildFrom` brick wall.)

## What it does not do
* Tell you the offending value
* Distinguish between CWL's when multiple are submitted in one file.  This is because `id` field is optional and made random by SALAD preprocessing.

Example of 2 errors combined:
```
DecodingFailure at .inputs[1].inputBinding.position: Int
DecodingFailure at .stdout: DecodingFailure at .stdout: DecodingFailure at .stdout: String
```
So not amazing (don't know why `at .stdout` is repeated thrice) but better.

### DIFF
--- a/cwl/src/main/scala/cwl/CommandLineTool.scala
+++ b/cwl/src/main/scala/cwl/CommandLineTool.scala
@@ -481,6 +481,8 @@ object CommandLineTool {
     import cats.instances.list._
     type DefaultToWomValueFunction = WomType => ErrorOr[WomValue]
 
+    type SecondaryFiles = StringOrExpression :+: Array[StringOrExpression] :+: CNil
+
     object DefaultToWomValuePoly extends Poly1 {
       implicit def caseFileOrDirectory: Case.Aux[FileOrDirectory, DefaultToWomValueFunction] = {
         at {
--- a/cwl/src/main/scala/cwl/CommandLineTool.scala
+++ b/cwl/src/main/scala/cwl/CommandLineTool.scala
@@ -481,8 +481,6 @@ object CommandLineTool {
     import cats.instances.list._
     type DefaultToWomValueFunction = WomType => ErrorOr[WomValue]
 
-    type SecondaryFiles = StringOrExpression :+: Array[StringOrExpression] :+: CNil
-
     object DefaultToWomValuePoly extends Poly1 {
       implicit def caseFileOrDirectory: Case.Aux[FileOrDirectory, DefaultToWomValueFunction] = {
         at {
--- a/cwl/src/main/scala/cwl/CwlCodecs.scala
+++ b/cwl/src/main/scala/cwl/CwlCodecs.scala
@@ -8,6 +8,12 @@ import eu.timepit.refined.string._
 import io.circe.refined._
 import io.circe.literal._
 import common.Checked
+import common.validation.ErrorOr.ErrorOr
+import cats.syntax.either._
+import cats.syntax.traverse._
+import cats.syntax.show._
+import cats.instances.vector._
+import io.circe.Error._
 
 object CwlCodecs {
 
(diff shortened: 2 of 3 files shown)


cwl/src/main/scala/cwl/CwlCodecs.scala
```diff
@@ -22,5 +31,44 @@ object CwlCodecs {
...
+      identity,
+      _ => original //this should never happen as we're gathering up the errors
+    )
+
+  private def decodeCwl(json: Json): ErrorOr[Unit] = {
+
+    val rawJson = io.circe.Printer.noSpaces.pretty(json)
+
+    // we know we are dealing with an individual CWL, so we can take some liberties like assuming "class" is defined.
+    ((json \\ "class").head.asString match {
```
mcovarr: If this is already returning `ErrorOr` would it be too painful to actually check if this JSON has no `class`?

cwl/src/main/scala/cwl/CommandLineTool.scala
```diff
@@ -478,6 +478,8 @@ object CommandLineTool {
     import cats.instances.list._
     type DefaultToWomValueFunction = WomType => ErrorOr[WomValue]
 
+    type SecondaryFiles = StringOrExpression :+: Array[StringOrExpression] :+: CNil
```
mcovarr: looks like we've got a stowaway, captain

cwl/src/main/scala/cwl/CwlCodecs.scala
```diff
@@ -22,5 +31,44 @@ object CwlCodecs {
...
+    *
+    * If it fails, attempt to parse each CWL individually and return a superset of those failures.
+    */
+  def decodeCwl(in: String): Checked[CwlFile] =
+    decodeWithErrorStrings[CwlFile](in).leftMap(decodePieces(in)).toEither
+
+  /**
+    * Drop down to a lower level and try to parse each CWL individually, returning a superset of those failures.
+    */
+  def decodePieces(in: String)(original: NonEmptyList[String]):NonEmptyList[String]  =
```
mcovarr: I think I'm getting confused with the parameter names... isn't `in` the original CWL and `original` the error strings from the first parse?

cwl/src/main/scala/cwl/CwlCodecs.scala
```diff
@@ -22,5 +31,44 @@ object CwlCodecs {
...
+  private def decodeCwl(json: Json): ErrorOr[Unit] = {
+
+    val rawJson = io.circe.Printer.noSpaces.pretty(json)
+
+    // we know we are dealing with an individual CWL, so we can take some liberties like assuming "class" is defined.
+    ((json \\ "class").head.asString match {
+      case Some("Workflow") => decodeWithErrorStrings[Workflow](rawJson)
+      case Some("CommandLineTool") => decodeWithErrorStrings[CommandLineTool](rawJson)
+      case Some("ExpressionTool") => decodeWithErrorStrings[ExpressionTool](rawJson)
+      case _ => "Class field was declared incorrectly!".invalidNel
```
mcovarr: The message could be more helpful, possibly including the actual unacceptable value.

cwl/src/main/scala/cwl/CwlCodecs.scala
```diff
@@ -22,5 +31,44 @@ object CwlCodecs {
...
+    (for {
+      raw <- parse(in).leftMap(_.show).leftMap(NonEmptyList.one)
+      _ <-
+        if (raw.isArray)
+          raw.asArray.toVector.flatten.traverse(decodeCwl).toEither
+        else
+          decodeCwl(raw).toEither
+    } yield  ()).fold(
+      identity,
+      _ => original //this should never happen as we're gathering up the errors
```
mcovarr: Yeah it seems this really shouldn't happen, maybe silently returning the original errors is not the right thing to do?

cwl/src/main/scala/cwl/CwlCodecs.scala
```diff
@@ -22,5 +31,44 @@ object CwlCodecs {
...
 
-  def decodeCwl(in: String): Checked[CwlFile] = decodeAccumulating[CwlFile](in).leftMap(_.map(_.getMessage).map(s"error parsing: $in" + _)).toEither
+  def errorToStrings : NonEmptyList[Error] => NonEmptyList[String] = _.map(_.show)
+
+  def decodeWithErrorStrings[A](in: String)(implicit d: Decoder[A]): ErrorOr[A] = decodeAccumulating[A](in).leftMap(errorToStrings)
+
+  /**
+    * Attempt to parse as an array of CWL or a single one.
+    *
+    * If it fails, attempt to parse each CWL individually and return a superset of those failures.
```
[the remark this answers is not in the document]
danbills: Yeah I wondered if I was using that correctly :fearful:.  I meant the set of all failures.

cwl/src/main/scala/cwl/CwlCodecs.scala
danbills: nope, it shall be done.

cwl/src/main/scala/cwl/CwlCodecs.scala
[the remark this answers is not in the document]
mcovarr: I think that's the union of all failures. Superset would be like you threw in some extra failures, no extra charge.

cwl/src/main/scala/cwl/CwlCodecs.scala
```diff
@@ -22,5 +31,44 @@ object CwlCodecs {
...
+
+    val rawJson = io.circe.Printer.noSpaces.pretty(json)
+
+    // we know we are dealing with an individual CWL, so we can take some liberties like assuming "class" is defined.
+    ((json \\ "class").head.asString match {
+      case Some("Workflow") => decodeWithErrorStrings[Workflow](rawJson)
+      case Some("CommandLineTool") => decodeWithErrorStrings[CommandLineTool](rawJson)
+      case Some("ExpressionTool") => decodeWithErrorStrings[ExpressionTool](rawJson)
+      case _ => "Class field was declared incorrectly!".invalidNel
+    }).map(_ => ())
```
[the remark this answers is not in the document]
danbills: I tried everything I could to make `void` work!  
i.e.
```
import cats.syntax.functor._
import cats.data.Validated._
```

cwl/src/main/scala/cwl/CwlCodecs.scala
danbills: Yes. These will be made more accurate and descriptive.

cwl/src/main/scala/cwl/CwlCodecs.scala
```diff
@@ -22,5 +31,44 @@ object CwlCodecs {
...
+    */
+  def decodePieces(in: String)(original: NonEmptyList[String]):NonEmptyList[String]  =
+    (for {
+      raw <- parse(in).leftMap(_.show).leftMap(NonEmptyList.one)
+      _ <-
+        if (raw.isArray)
+          raw.asArray.toVector.flatten.traverse(decodeCwl).toEither
+        else
+          decodeCwl(raw).toEither
+    } yield  ()).fold(
```
Horneth: a temp variable could make this a bit more readable perhaps instead of `(for { ... }).fold(...)`

cwl/src/main/scala/cwl/CwlCodecs.scala
danbills: Yeah, this is an exceptional case, and should make some noisy noise

cwl/src/main/scala/cwl/CwlCodecs.scala
```diff
@@ -22,5 +31,44 @@ object CwlCodecs {
...
+      identity,
+      _ => original //this should never happen as we're gathering up the errors
+    )
+
+  private def decodeCwl(json: Json): ErrorOr[Unit] = {
+
+    val rawJson = io.circe.Printer.noSpaces.pretty(json)
```
Horneth: do we really need to re-print the json to re-parse it ? :(

danbills: No, this should be improved.

PR #3182, opened 2018-01-24
Hotfix edition: Add valid hostname to DockerHubFlow

ruchim:


ruchim:
I heard chatter about a 30.2 release ...is there any chance this change can make it in that release? It's mostly for FC users as the current failure logs are sent to the server logs and basically the user never sees call caching fail even though the job succeeds.

Issue #3186, opened 2018-01-24
CWL salad left behind after workflow finishes


kshakir:
While running a CWL workflow, cromwell writes files to disk so that [cwltool](https://github.com/broadinstitute/cromwell/issues/2718) may salad them.

While the top-level CWL is read back into memory (& db??) the top-level CWL often refers to other relative files on disk. These relative files are required to remain for the duration of the workflow. After the workflow completes the files should be deleted. Otherwise after a large number of workflows the local disk will fill up.

Right now, the salad files are being written to a random path. For the purposes of this ticket the files could be written to a directory based on the stable `workflowId` (a `UUID`). The per-workflow logs do something similar, and are deleted at the end of the workflow in a way that the cwl files should be also.

Horneth:
Why do we need them for the duration of the workflow ? Aren't we good once we've got the full CWL object ? If so we could delete the whole thing right after it ?

kshakir:
> Why do we need them for the duration of the workflow ?

Just getting to this now and I don't 100% remember why. I would try deleting them and "see what happens". From what I remember at least one conformance test crashed due to missing files.

Horneth:
The original issue __might__ be alleviated by the fact that the conformance test runner (as well as `cromwell submit`) now salads everything in place and inlines all imported cwls into a single standalone CWL workflow

PR #3187, opened 2018-01-24
Test PAPI CWL in parallel.

kshakir:
Decode cwl into a temporary sub-directory so that unzipping doesn't collide under top level `/tmp`
BUG/TODO: cannot delete the cwl sub-directory to clean up files until after the workflow finishes.
Fixed passing cromwell inputs via `-i`.
Passing centaur google config via new environment variables listed in reference.conf instead of on command line.
Limit conformance tests to -Xmx1g per java process by default, and 2g for the cromwells running the conformance tests.


cwl/src/main/scala/cwl/CwlDecoder.scala
```diff
@@ -43,6 +51,15 @@ object CwlDecoder {
     fromEither[IO](cwlToolResult flatMap resultToEither)
   }
 
+  // TODO: WOM: During conformance testing the saladed-CWLs are referring to files in the temp directory.
+  // Thus we can't delete the temp directory until after the workflow is complete, like the workflow logs.
+  // All callers to this method should be cleaned up around at the same time.
+  def todoDeleteCwlFileParentDirectory(cwlFile: BFile): Parse[Unit] = {
+    goParse {
+      //cwlFile.parent.delete(swallowIOExceptions = true)
```
kshakir: Ticket filed: #3186

cwl/src/main/scala/cwl/CwlDecoder.scala
```diff
@@ -43,6 +51,15 @@ object CwlDecoder {
     fromEither[IO](cwlToolResult flatMap resultToEither)
   }
 
+  // TODO: WOM: During conformance testing the saladed-CWLs are referring to files in the temp directory.
+  // Thus we can't delete the temp directory until after the workflow is complete, like the workflow logs.
+  // All callers to this method should be cleaned up around at the same time.
```
mcovarr: that last sentence is a bit off

src/bin/travis/resources/cwl_conformance_test.wdl
```diff
@@ -0,0 +1,137 @@
...
+
+task get_test_count {
+    String cwl_dir
+    String centaur_cwl_runner
+
+    command {
+        (
+            cd ${cwl_dir}
+            ./run_test.sh RUNNER="${centaur_cwl_runner}" -l | grep '^\[' | wc -l
+        )
```
mcovarr: is this subshelling because our command script doesn't protect itself against user commands cd'ing?

src/bin/travis/resources/papi_conformance_expected_failures.txt
```diff
@@ -0,0 +1,114 @@
+1
```
mcovarr: TOL I probably prefer a single file ATM though that scheme certainly wouldn't scale well to N > 2

src/main/scala/cromwell/CommandLineParser.scala
```diff
@@ -64,7 +64,7 @@ object CommandLineParser extends App {
         children(
           arg[String]("workflow-source").text("Workflow source file.").required().
             action((s, c) => c.copy(workflowSource = Option(DefaultPathBuilder.get(s)))),
-          opt[String]('i', "workflow-root").text("Workflow root.").
+          opt[String]("workflow-root").text("Workflow root.").
```
mcovarr: TOL even if we did want a one char mnemonic I'm not sure why that one 😕

src/bin/travis/resources/cwl_conformance_test.wdl
kshakir: Whoops, copypasta from `run_test_index` below. `run_test_index` needs to write the exit code into a file back in the original directory. `get_test_count` here does not. Will remove this subshell.

src/main/scala/cromwell/CommandLineParser.scala
kshakir: I assumed it was copied from `opt[String]('i', "inputs")` below.

mcovarr: :man_facepalming:

src/bin/travis/resources/papi_conformance_expected_failures.txt
kshakir: Open to other ideas. I thought about using a single multi-column file for example. But eventually / in my dream world these files will not be used at all, especially as cwl-test-numbers are NOT stable.

mcovarr:
Just curious if there's a sequel in the works for parallelizing local backend conformance?

kshakir:
The urgency of this particular fix came as we starting adding more valid CWL to PAPI and the tests were creeping past 70 minutes. I'm open to parallelizing local too. The workflow is reusable and there's currently nothing technically stopping us from switching local to parallel also.

Cromwell's "randomization" (aka internally using unordered sets/hashmaps) when launching scatter jobs makes debugging the entire suite of tests wicked painful. It's hard to tell when a failure occurs what test was running. While CWL is in a state of flux I kind of like slowly working my way through the serial logs of local-conformance when I break something.

My 2¢/ToL

centaur/test_cromwell.sh
```diff
@@ -36,7 +36,7 @@ RUN_DIR=$(pwd)
 TEST_THREAD_COUNT=16
```
geoffjentry: I'm assuming you shellchecked all of this :P

geoffjentry:
👍

mcovarr:
OK, I was just curious since something seems to have changed wrt local conformance. Most builds seem to be over an hour now and multiple have gone over 90 minutes. 😦

centaur/test_cromwell.sh
kshakir: I only shellcheck other peoples PRs. 😉

PR #3191, opened 2018-01-24
Better ECMAScript WomOptionalValue encoding

kshakir:


wom/src/main/scala/wom/util/JsEncoder.scala
```diff
@@ -37,7 +36,8 @@ class JsEncoder {
     */
   def encode(value: WomValue): ErrorOr[AnyRef] = {
     value match {
-      case WomOptionalValue(WomNothingType, None) => Validated.valid(null)
+      case WomOptionalValue(_, None) => Validated.valid(null)
+      case WomOptionalValue(_, Some(innerValue)) => encode(innerValue)
```
cjllanwarne: Yup, that's better!

PR #3193, opened 2018-01-25
Reuse ogins between if condition and the inner block

cjllanwarne:
Make sure outer lookups for `if` conditions and `scatter` collections use the same set of ogins as the inner graph receives.

Most of this PR is moving code around

### DIFF
--- a/wdl/src/main/scala/wdl/If.scala
+++ b/wdl/src/main/scala/wdl/If.scala
@@ -39,13 +39,6 @@ object If {
     * @param preserveIndexForOuterLookups When we're evaluating the condition boolean, should we preserve scatter index if we have to use the outerLookup?
     */
   def womConditionalNode(ifBlock: If, localLookup: Map[String, GraphNodePort.OutputPort], outerLookup: Map[String, OutputPort], preserveIndexForOuterLookups: Boolean): ErrorOr[ConditionalNodeWithNewNodes] = {
-    val ifConditionExpression = WdlWomExpression(ifBlock.condition, ifBlock)
-    val ifConditionGraphInputExpressionValidation = WdlWomExpression.toAnonymousExpressionNode(
-      WomIdentifier("conditional"), ifConditionExpression, localLookup, outerLookup, preserveIndexForOuterLookups, ifBlock, PlainAnonymousExpressionNode.apply)
-    val ifConditionTypeValidation = ifConditionExpression.evaluateType((localLookup ++ outerLookup).map { case (k, v) => k -> v.womType }) flatMap {
-      case coerceable if WomBooleanType.isCoerceableFrom(coerceable) => Valid(())
-      case other => s"An if block must be given a boolean expression but instead got '${ifBlock.condition.toWomString}' (a ${other.toDisplayString})".invalidNel
-    }
 
     /*
       * Why? Imagine that we're building three nested levels of a innerGraph.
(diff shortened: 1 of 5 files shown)


wdl/src/main/scala/wdl/If.scala
```diff
@@ -69,6 +62,14 @@ object If {
     }
     val possiblyNeededNestedOginPorts: Map[String, OutputPort] = possiblyNeededNestedOgins map { case (name: String, ogin: OuterGraphInputNode) => name -> ogin.singleOutputPort }
 
+    val ifConditionExpression = WdlWomExpression(ifBlock.condition, ifBlock)
+    val ifConditionGraphInputExpressionValidation = WdlWomExpression.toAnonymousExpressionNode(
+      WomIdentifier("conditional"), ifConditionExpression, localLookup ++ possiblyNeededNestedOginPorts, Map.empty, preserveIndexForOuterLookups, ifBlock, PlainAnonymousExpressionNode.apply)
```
cjllanwarne: This line isn't just a copy/paste. Note the `..., localLookup ++ possiblyNeededNestedOginPorts, Map.empty, ...`

wdl/src/main/scala/wdl/Scatter.scala
```diff
@@ -68,6 +57,18 @@ object Scatter {
     }
     val possiblyNeededNestedOginPorts: Map[String, OutputPort] = possiblyNeededNestedOgins map { case (name: String, ogin: OuterGraphInputNode) => name -> ogin.singleOutputPort }
 
+    // Convert the scatter collection WdlExpression to a WdlWomExpression
+    val scatterCollectionExpression = WdlWomExpression(scatter.collection, scatter)
+    // Generate an ExpressionNode from the WdlWomExpression
+    val scatterCollectionExpressionNode =
+      WdlWomExpression.toAnonymousExpressionNode(WomIdentifier(scatter.item), scatterCollectionExpression, localLookup ++ possiblyNeededNestedOginPorts, Map.empty, preserveIndexForOuterLookups, scatter, PlainAnonymousExpressionNode.apply)
```
cjllanwarne: This line isn't just a copy/paste. Note the `..., localLookup ++ possiblyNeededNestedOginPorts, Map.empty, ...`

wdl/src/main/scala/wdl/If.scala
Horneth: I see the `possiblyNeededNestedOginPorts ` being added to the `localLookup` but I'm just wondering why the `outerLookup` has been replaced with `Map.empty` ?

Horneth: oh nevermind I think I got it

cjllanwarne: I'm actually tempted to remove the outer lookup entirely and have callers bundle in "possible needed ogins" as local inputs. 

Since that's my go-to fix to all these "duplicate ogin created" bugs, being proactive about removing the other cases might not be such a bad idea. Not in this PR though...

Horneth: Yeah as long as it doesn't mess up with `preserveIndexForOuterLookups` I agree

mcovarr: yeah this would have been a great candidate for split commits but I think I get it as much as I'm going to anytime soon...

PR #3198, opened 2018-01-25
Use processed inputs from command instantiation to write the command script.

mcovarr:
Changes are smaller than they appear, much of this is restructuring to do command instantiation and then flatMapping the rest of the logic using a half-built `InstantiatedCommand`. Slinging around a half-built `InstantiatedCommand` is IMHO a bit gross but preexisting.

This leverages the preprocessing and value mapping of inputs that takes place in command instantiation and captures those versions of the inputs. This enables the stdin redirection in particular to get the actual name of the standard input file as seen within the container.


supportedBackends/sfs/src/main/scala/cromwell/backend/impl/sfs/config/ConfigAsyncJobExecutionActor.scala
```diff
@@ -100,7 +100,7 @@ sealed trait ConfigAsyncJobExecutionActor extends SharedFileSystemAsyncJobExecut
     val allInputs = providedWomInputs ++ optionalsForciblyInitializedToNone
     val womInstantiation = taskDefinition.instantiateCommand(allInputs, NoIoFunctionSet, identity, runtimeEnvironment)
 
-    val InstantiatedCommand(command, _, _, _) = womInstantiation.toTry.get
+    val InstantiatedCommand(command, _, _, _, _, _) = womInstantiation.toTry.get
```
kshakir:
```scala
val command = womInstantiation.toTry.get.commandString
```

mcovarr: I'm sure this made more sense at one point... 😦

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -287,52 +289,59 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
...
     }
 
     val callable = jobDescriptor.taskCall.callable
+    def makeStringKeyedMap(list: List[(LocalName, WomValue)]): Map[String, WomValue] = list.toMap map { case (k, v) => k.value -> v }
 
-    val adHocFileCreations: ErrorOr[List[WomFile]] = callable.adHocFileCreation.toList.flatTraverse(
-      _.evaluateValue(commandInstantiationInputs, backendEngineFunctions).flatMap(validateAdHocFile)
-    )
+    val command = instantiatedCommandValidation flatMap { instantiatedCommand =>
+      val taskInputs = instantiatedCommand.taskInputs |> makeStringKeyedMap
```
Horneth: Don't you have access to the evaluated task inputs through the jobDescriptor already ? `jobDescritpor.evaluatedTaskInputs`

mcovarr: Evaluated inputs yes, preprocessed evaluated inputs no. I'll clean up this naming fail.

PR #3199, opened 2018-01-25
Contribution guidelines in Docs

katevoss:


docs/Contribute.md
```diff
@@ -0,0 +1,126 @@
+**Contribution Guidelines**
+
+Cromwell is an open-source project and we heartily welcome community contributions to both our code and our documentation. Here are some guidelines for adding documentation and recommendations on where we could use help the most. 
+
+First off, here are useful links:
+
+* Cromwell documentation: [cromwell.readthedocs.io/en/develop](http://cromwell.readthedocs.io/en/develop)
```
kshakir: ToL: Perhaps the link should be to the shorter `cromwell.readthedocs.io`, not `cromwell.readthedocs.io/en/develop`.

docs/Contribute.md
```diff
@@ -0,0 +1,126 @@
...
+* Absolute: `![](URLofImg.png)`
+	* _Example:_ `![](https://www.broadinstitute.org/sites/all/themes/custom/at_broad/logo.png)` _produces this image_  
+	![](https://www.broadinstitute.org/sites/all/themes/custom/at_broad/logo.png)
+
+### FAQs
+
+**_Why isn't my documentation showing up?_**
+
+* **Have you confirmed that your PR is merged?**  
+If not, [go merge it](https://github.com/broadinstitute/cromwell/pulls). This will trigger an automatic build.
```
kshakir: Only certain team members can merge. Others should wait patiently for a merge and then yell if we take too long to notice their stale PR.

docs/execution/ExecutionTwists.md
```diff
@@ -95,3 +95,43 @@ Once all jobs have been reconnected to, the workflow will keep running normally.
...
+This means that they are given some time (see below for how much and how to change it) to return to a known "consistent" state.
+For example, an actor waiting for a response from the database before sending information to the metadata will wait for that response before shutting itself down.
+4. All active connections from the REST endpoints are completed and closed. At this point any client that made a request before the shutdown process started should have received a response.
+5. All actors responsible for data persistence are in turn being asked to gracefully shutdown. 
+For example, all queued up metadata writes are executed.
+6. Database connection pools are shutdown.
+7. Actor system shuts down.
+8. JVM exits.
+    
+This multi-stage process is designed to minimize the risk of data loss during shutdown. However in order to prevent this process to last forever, each stage (called phase) has its own timeout.
```
kshakir: > However in order to prevent this process to last forever

should perhaps be:

> However in order to prevent this process from lasting forever

docs/execution/ExecutionTwists.md
```diff
@@ -95,3 +95,43 @@ Once all jobs have been reconnected to, the workflow will keep running normally.
...
+If the phase does not complete within the given timeout, actors will be forcefully stopped and the next phase will start.
+
+This logic is implemented using [Akka Coordinated Shutdown Extension](http://doc.akka.io/docs/akka/current/scala/actors.html#coordinated-shutdown).
```
kshakir: ToL: I never know what to do about linking to external-docs combined with the fact that we _rarely_ update our-docs when we update our libraries. The link here is to Akka's "[current](http://doc.akka.io/docs/akka/current/scala/actors.html#coordinated-shutdown)" floating docs URL, while we're currently [pinned](https://github.com/broadinstitute/cromwell/blob/631ec41c6303321d0d5d384061f742543782a746/project/Dependencies.scala#L5) the akka library version "2.5.4" which is close but not technically the current version. We could change our doc link for now to point to [this](https://doc.akka.io/docs/akka/2.5.4/scala/actors.html#coordinated-shutdown) permalink instead. But if someone updates the library I'm not betting they will come back and fix the permalink link too…

docs/Contribute.md
```diff
@@ -0,0 +1,126 @@
...
+After you edit the [`cromwell.yaml`](https://github.com/broadinstitute/cromwell/blob/develop/engine/src/main/resources/swagger/cromwell.yaml), run `sbt generateRestApiDocs` and commit all changes.  
+_Hint:_ Once you have regenerated the docs correctly, the hidden timestamp at the top of the [`RESTAPI.md` file](https://raw.githubusercontent.com/broadinstitute/cromwell/develop/docs/api/RESTAPI.md) will show the current time.
+
+**_How do I generate a Markdown document of the Swagger YAML?_**
+
+A library named [swagger2markup](https://github.com/Swagger2Markup/swagger2markup) will render a Swagger YAML into various formats, including Markdown. Then generated Markdown is further customized by a custom SBT task.
+
+**_Generating and Hosting ScalaDoc_**
+
+Essentially run `sbt doc` then commit the generated code into the `gh-pages` branch on this repository
```
cjllanwarne: Essentially? 🙁

docs/execution/ExecutionTwists.md
```diff
@@ -95,3 +95,43 @@ Once all jobs have been reconnected to, the workflow will keep running normally.
...
+
+You'll notice the presence of a `PhaseAbortAllWorkflows` phase. This phase is at the same level as the `PhaseServiceRequestsDone` phase which corresponds to our step #3 above.
+The reason for a specific abort phase is so that its timeout can be configured differently than the normal shutdown phase.
+
+Indeed, stopping all workflows and aborting them is very similar from an outside perspective. We send a message (resp. "Stop" and "Abort") and wait for a response.
+
+However we want to be able to give more time to abort as it will likely involve more work. This can be done by editing the value of `coordinated-shutdown.phases.abort-all-workflows.timeout` which defaults to 1 hour.
+Phases timeouts default to 5 seconds, except the stop-io-activity phase which defaults to 30 minutes. This is because depending on the Database load at the time of the shutdown, it might take a significant amount of time to flush all pending writes.
+
+All of the timeouts are configurable in the `akka.coordinated-shutdown.phases` section (see `reference.conf`).
```
kshakir: The "see `reference.conf`" should be a link. Whether floating or permalink… see prior ToL.

- 2.5.4 permalink:  
  https://github.com/akka/akka/blob/v2.5.4/akka-actor/src/main/resources/reference.conf#L1024-L1133

or

- `master`/floating link to just the current version of `reference.conf`, and user hunts through for the section they want (also assuming the file hasn't moved):  
  https://raw.githubusercontent.com/akka/akka/master/akka-actor/src/main/resources/reference.conf

docs/Contribute.md
```diff
@@ -0,0 +1,126 @@
...
+2. Put the most important information first.
+3. Use the second person, use “you” instead of “the user”.
+4. No passive verbs (everything is done by something).
+5. Link to the original source, don't repeat documentation. 
+
+### Formatting
+
+The documentation is written in Markdown. Click here for a [Github Guide on Markdown](https://guides.github.com/features/mastering-markdown/), and click here for more [tips from MkDocs](http://www.mkdocs.org/user-guide/writing-your-docs/).
+
+### Styling
```
cjllanwarne: Maybe have a link to a more complete github markdown page? Otherwise is this going to be a slightly arbitrary subset?

docs/Contribute.md
```diff
@@ -0,0 +1,126 @@
...
+		\}  
+	\`\`\`	
+_produces this block_  
+```json  
+	{
+	    "MyWorkflow.MyTask.VariableTwo": "Variable2"
+	}
+```	
+
+**REST API:**
```
cjllanwarne: If you do keep the `Styling` section, these `REST API` instructions don't belong in the middle of it

docs/Contribute.md
```diff
@@ -0,0 +1,126 @@
...
+
+* **Has the build finished?**   
+[Check build status here](https://readthedocs.org/projects/cromwell/builds/).
+
+* **Did you add the file(s) to the YAML file?**  
+If not, [add it here](https://github.com/broadinstitute/cromwell/blob/develop/mkdocs.yml).
+
+**_How do I add to the REST API documentation?_**
+
+* **Don't forget to regenerate**
```
cjllanwarne: regenerate

docs/Contribute.md
```diff
@@ -0,0 +1,126 @@
...
+
+**_How do I add to the REST API documentation?_**
+
+* **Don't forget to regenerate**  
+After you edit the [`cromwell.yaml`](https://github.com/broadinstitute/cromwell/blob/develop/engine/src/main/resources/swagger/cromwell.yaml), run `sbt generateRestApiDocs` and commit all changes.  
+_Hint:_ Once you have regenerated the docs correctly, the hidden timestamp at the top of the [`RESTAPI.md` file](https://raw.githubusercontent.com/broadinstitute/cromwell/develop/docs/api/RESTAPI.md) will show the current time.
+
+**_How do I generate a Markdown document of the Swagger YAML?_**
+
+A library named [swagger2markup](https://github.com/Swagger2Markup/swagger2markup) will render a Swagger YAML into various formats, including Markdown. Then generated Markdown is further customized by a custom SBT task.
```
cjllanwarne: I'm not sure this directly answers the question "How do **I** generate a Markdown document of the Swagger YAML?"

docs/execution/ExecutionTwists.md
```diff
@@ -95,3 +95,43 @@ Once all jobs have been reconnected to, the workflow will keep running normally.
...
+![Scaladoc](CromwellShutdownProcess.png)
+
+Pre-defined but unused phases have been omitted (cluster related phases for example that are irrelevant in Cromwell).
+
+You'll notice the presence of a `PhaseAbortAllWorkflows` phase. This phase is at the same level as the `PhaseServiceRequestsDone` phase which corresponds to our step #3 above.
+The reason for a specific abort phase is so that its timeout can be configured differently than the normal shutdown phase.
+
+Indeed, stopping all workflows and aborting them is very similar from an outside perspective. We send a message (resp. "Stop" and "Abort") and wait for a response.
+
+However we want to be able to give more time to abort as it will likely involve more work. This can be done by editing the value of `coordinated-shutdown.phases.abort-all-workflows.timeout` which defaults to 1 hour.
```
cjllanwarne: Is it just me or is this first sentence a little strangely worded?

docs/Contribute.md
katevoss: I copy-pasta'd, I'll admit. 

What is the ScalaDoc for? Do we use it? Does anyone use it?

docs/Contribute.md
katevoss: As far as I can tell, this is the complete Github Guide to Markdown page. It has all of the Markdown syntax and the Github-flavored syntax.

docs/Contribute.md
katevoss: I thought everyone could merge once it had been approved 2x?

geoffjentry: In particular, external contributions **can not** merge themselves. Since presumably they're the crowd who would care about this sort of information, that's a handy difference to note

docs/execution/ExecutionTwists.md
katevoss: That's a good point. I think I'll link to both the "current" and the "2.5.4" docs, so that if we do happen to forget to update the docs (which is impossible, I know), that users can still find the latest docs.

docs/execution/ExecutionTwists.md
katevoss: Hmm I think in this situation I'll use the current version of `reference.conf`. Thanks!

docs/Contribute.md
katevoss: ten-regenerating

docs/Contribute.md
katevoss: @cjllanwarne I ported this over from the Wiki. Why would someone want to (or need to) generate a Markdown document of the Swagger yaml?

docs/execution/ExecutionTwists.md
katevoss: How about
> In the case where you want to be able to give more time to abort, as it will likely involve more work, you can edit the value of `coordinated-shutdown.phases.abort-all-workflows.timeout` which defaults to 1 hour.

Does that still convey the same information?

docs/Contribute.md
cjllanwarne: So just OOI is there a reason for duplicating it here instead of linking?

docs/Contribute.md
cjllanwarne: Presumably to use that as the REST API docs? @kshakir ?

docs/Contribute.md
geoffjentry: We should nix it. If we ever want to resurrect we can, however since our github pages was removed a couple of years ago I know that no one is using it right now :)

cjllanwarne: I would **guess** that it lets you import cromwell or cromwell-wdl and get access to API docs. I know at least one team is DSDE does this, maybe we can ask whether they ever use the scaladocs?

cjllanwarne:
Only minor comments remain, 👍

docs/Contribute.md
katevoss: I understand you now. The reason I am duplicating here is that there is some ReadTheDocs-specific Markdown syntax for styling links, code, etc.

docs/Contribute.md
katevoss: That's not actually how we're regenerating the REST API docs, so I'll remove it.

PR #3200, opened 2018-01-25
Expression tool Part1

Horneth:
This prepares the next PR that will implement support for `ExpressionTool`
The first 2 commits are pure renames.
The 3rd one breaks `CommandLineTool` apart:
Put `InputParameter` and `OutputParameter` in their own file, and make them trait so they can be extended by specific CWL objects (`WorkflowInputParameter`, `CommandInputParameter`, `ExpressionInputParameter`)
Some logic in `CommandLineTool` is abstracted into a `Tool` trait (that will then be extended by `ExpressionTool`)
TaskDefinition is also being turned into a trait so we can have `CommandTaskDefinition` and (in the next PR) `ExpressionTaskDefinition`.

The general idea is to have `ExpressionTool` behave much like a `CommandLineTool`, including in WOM, except the actual work to be done is an expression to be evaluated by the engine (at least for now ?) instead of a command line run by the backend.


cwl/src/main/scala/cwl/InputParameter.scala
```diff
@@ -0,0 +1,88 @@
...
+  def streamable: Option[Boolean]
+  def doc: Option[Doc]
+  def inputBinding: Option[InputCommandLineBinding]
+  def default: Option[CwlAny]
+  def `type`: Option[MyriadInputType]
+
+  def loadContents = inputBinding.flatMap(_.loadContents).getOrElse(false)
+}
+
+object InputParameter {
```
Horneth: Everything (except the unapplies) here is just copied over from `CommandLineTool`

cwl/src/main/scala/cwl/OutputParameter.scala
```diff
@@ -0,0 +1,124 @@
...
+  def label: Option[String]
+  def secondaryFiles: Option[SecondaryFiles]
+  def format: Option[StringOrExpression]
+  def streamable: Option[Boolean]
+  def doc: Option[String :+: Array[String] :+: CNil]
+  def outputBinding: Option[CommandOutputBinding]
+  def `type`: Option[MyriadOutputType]
+}
+
+object OutputParameter {
```
Horneth: Everything (except the unapplies) here is just copied over from `CommandLineTool`

cwl/src/main/scala/cwl/CommandLineTool.scala
```diff
@@ -470,101 +358,12 @@ object CommandLineTool {
                                     id: String,
                                     label: Option[String] = None,
                                     secondaryFiles: Option[SecondaryFiles] = None,
-                                    format: Option[Expression :+: Array[String] :+: String :+: CNil] = None, //only valid when type: File
+                                    format: Option[Expression :+: String :+: Array[String] :+: CNil] = None, //only valid when type: File
```
mcovarr: why?

cwl/src/main/scala/cwl/CommandLineTool.scala
```diff
@@ -151,7 +105,7 @@ case class CommandLineTool private(
     }
   }
 
-
+  // This seem like it makes sense only for CommandLineTool and hence is not abstrsacted in Tool. If this assumption is wrong it could be moved up.
```
geoffjentry: seems

cwl/src/main/scala/cwl/CommandLineTool.scala
```diff
@@ -265,109 +219,51 @@ case class CommandLineTool private(
...
     }
 
-    // The try will succeed if this is a task within a step. If it's a standalone file, the ID will be the file,
-    // so the filename is the fallback.
-    def taskName = Try(FullyQualifiedName(id).id).getOrElse(Paths.get(id).getFileName.toString)
+    val stdinRedirection: Option[WomExpression] = stdin map {
+      _.fold(StringOrExpressionToWomExpression).apply(inputNames, expressionLib)
+    }
 
+    // This seem like it makes sense only for CommandLineTool and hence is not abstrsacted in Tool. If this assumption is wrong it could be moved up.
```
geoffjentry: abstracted

geoffjentry:
👍

cwl/src/main/scala/cwl/CommandLineTool.scala
Horneth: Because it now extends `InputParameter` which defines an abstract member for `format`, so the types have to be in the same order otherwise it doesn't compile.
Which actually reminds me that I created a`type Format` so I'll just use that here

PR #3203, opened 2018-01-26
Expression Tool Part 2

Horneth:
Enables `ExpressionTool`s to be run in the same way `CommandLineTool`s are run (as standalone tools or as a workflow step). They follow exactly the same transformation into WOM and then to Cromwell as `CommandLineTool`s. The only major difference being that instead of a command line to be run it's an expression.
There's a new type of `CallNode`: `ExpressionCallNode` to represent calls that encapsulate an expression instead of a command line.
When the call becomes runnable Cromwell will run the `evaluate` method on the node using its engine `IoFunctions` and the result will be aded to the value store immediately. The node will not make it to any backend.
Happy to whiteboard / tech talk / walkthrough this if needed.

Only 2 of the 6 tests enabled on local work on PAPI:
- One should fix itself once the PAPI input redirection PR is merged.
- The 3 others I plan to address in a 3rd PR


centaurCwlRunner/src/main/scala/centaur/cwl/Outputs.scala
```diff
@@ -85,7 +85,7 @@ object CwlOutputsFold extends Poly1 {
   }
 
   implicit def et: Case.Aux[cwl.ExpressionTool, Map[String, MyriadOutputType]] = at[cwl.ExpressionTool] {
-    _.outputs.map(output => output.id -> output.`type`).toMap
+    _.outputs.map(output => output.id -> output.`type`.get).toMap
```
danbills: Weird... how did this ever compile?

cwl/src/main/scala/cwl/ExpressionTool.scala
```diff
@@ -1,21 +1,105 @@
...
+        * However since the spec is not clear on this and cwltool seems to ignore entirely the outputs declaration,
+        * we'll do the same for now, meaning we'll assign a value to the output port iff the result of the expression has one for them,
+        * otherwise we'll remove the port for the final map.
+        * If there are fields in the expression's result that do not have a matching declared output port, they won't be added either,
+        * because we'd have to create an output port for them now which would be useless anyway since nothing could point to it.
+       */
+      def mapPortsToValues(values: Map[String, WomValue]): Checked[Map[OutputPort, WomValue]] = {
+        import cats.instances.list._
+        import cats.syntax.traverse._
+        import cats.syntax.either._
```
danbills: Looks like you already have this import in scope.

centaurCwlRunner/src/main/scala/centaur/cwl/OutputManipulator.scala
```diff
@@ -21,11 +21,21 @@ object OutputManipulator extends Poly1 {
...
   }
-  
+
   private def populateFileFields(pathBuilder: PathBuilder)(obj: JsonObject): JsonObject = {
-    val objAsMap = obj.toMap
-    val path = pathBuilder.build(objAsMap("location").asString.get).get
-    
+    import mouse.all._
+
+    val path = pathBuilder.build(obj.kleisli("location").get.asString.get).get
```
danbills: is there a potential to use `Option` here and thus omit the gets?

wom/src/main/scala/wom/callable/CommandTaskDefinition.scala
```diff
@@ -174,3 +173,44 @@ final case class ExecutableTaskDefinition private (callableTaskDefinition: Calla
...
 }
+
+sealed trait ExpressionTaskDefinition extends TaskDefinition {
+  def evaluate: (Map[String, WomValue], IoFunctionSet, List[OutputPort]) => Checked[Map[OutputPort, WomValue]]
+}
+
+
+/**
+  * An expression task definition only.
+  * Can be called but cannot be used in an Executable as a standalone execution.
```
danbills: > cannot be used in an Executable as a standalone execution

(Tol as I don't understand Expression Tools yet) do we enforce this?

Horneth: I actually copied this from the `CommandLineTool` version.
It's a `WOM` concept.
`Executable`s can be executed as "standalone" objects.
`Callable`s can be called in a workflow step.
`ExecutableCallable`s are both.
AFAIK It's enforced in the sense that Cromwell gives WOM basically 2 `String`a (workflow and inputs) and expects an `Executable` to come back.

centaurCwlRunner/src/main/scala/centaur/cwl/Outputs.scala
Horneth: The type for the `ExpressionTool` outputs was not optional (the spec says it should http://www.commonwl.org/v1.0/Workflow.html#ExpressionToolOutputParameter)

centaurCwlRunner/src/main/scala/centaur/cwl/OutputManipulator.scala
Horneth: yes, this is will be part of the changes that'll be coming in the next PR to fix some PAPI failures

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/keys/ExpressionKey.scala
```diff
@@ -2,48 +2,34 @@ package cromwell.engine.workflow.lifecycle.execution.keys
 
 import akka.actor.ActorRef
 import cats.syntax.validated._
-import common.validation.ErrorOr.{ErrorOr, _}
-import cromwell.core.ExecutionIndex._
+import common.validation.ErrorOr._
+import cromwell.core.ExecutionIndex.{ExecutionIndex, _}
```
kshakir: IntelliJ missed this one. Should be `import cromwell.core.ExecutionIndex._`

wom/src/main/scala-2.12/wom/executable/ExecutableValidation.scala
```diff
@@ -1,29 +0,0 @@
...
-import common.Checked
-import common.validation.ErrorOr.ErrorOr
-import common.validation.Checked._
-import wom.callable.ExecutableCallable
-import wom.executable.Executable.{DelayedCoercionFunction, InputParsingFunction, ResolvedExecutableInputs}
-import wom.graph.Graph
-
-// WARNING! Because of 2.11 vs 2.12 incompatibilities, there are two versions of this file.
-// If you're making changes here, you'll also need to update ../../../scala-2.11/wom/executable/ExecutableValidation.scala
-// (More info below on why this was necessary)
```
kshakir: This warning is delete here, but the equivalent comment still remains in the newly merged copy.

cwl/src/main/scala/cwl/Workflow.scala
```diff
@@ -7,15 +7,14 @@ import cats.instances.list._
...
 import cats.syntax.traverse._
 import common.Checked
-import common.validation.Validation.OptionValidation
 import common.validation.ErrorOr._
-import shapeless._
-import shapeless.syntax.singleton._
-import CwlVersion._
-import cwl.command.ParentName
-import common.validation.Validation._
+import common.validation.Validation.{OptionValidation, _}
```
kshakir: IntelliJ missed this one. Should be import `common.validation.Validation._`

PR #3204, opened 2018-01-26
Step Input Expressions

danbills:
This requirement boils down to declaring multiple sources of an input and providing them to a `valueFrom` expression to determine the actual value of the input.

We should pass another valueFrom-related test (\#63) when ExpressionTool support is merged, which is imminent.


cwl/src/main/scala/cwl/WorkflowStepInputExpression.scala
```diff
@@ -1,23 +1,74 @@
...
+        * evaluation of valueFrom on other parameters.
+        */
+      case (Some(StringOrExpression.Expression(expression)), Some(sources)) =>
+        //used to determine the value of "self" as expected by CWL Spec
+        def selfValue(in: List[WomValue]) = in match {
+          case single :: Nil => single
+          case multiple => WomArray(multiple)
+        }
+
+        val either =
```
Horneth: ToL: a name describing what it represents would be nice :)

cwl/src/main/scala/cwl/WorkflowStep.scala
```diff
@@ -108,13 +108,8 @@ case class WorkflowStep(
...
         fold =>
-          // The source from which we expect to satisfy this input (output from other step or workflow input)
-          // TODO: this can be None, a single source, or multiple sources. Currently assuming it's a single one
-          val inputSource: String = workflowStepInput.source.flatMap(_.select[String]).get
-
-          // Name of the step input
-
-          val accumulatedNodes = fold.generatedNodes ++ knownNodes
+          // The sources from which we expect to satisfy this input (output from other step or workflow input)
+          val inputSources: List[String] = workflowStepInput.source.toList.flatMap(_.fold(WorkflowStepInputSourceToStrings))
```
Horneth: Can we add a comment to say that we're about to go through the list and for each source make sure we either have or generate the nodes they're referencing ?

Horneth: The comment might be better somewhere else but it'll help (at least me) next time I look at this to understand what we're doing with all those sources.

cwl/src/main/scala/cwl/model.scala
```diff
@@ -25,18 +27,27 @@ case class WorkflowStepInput(
...
+        get(id).
+        toValidNel(s"couldn't find $id as derived from $source in map\n${outputTypeMapWithIDs.mkString("\n")}")
+
     (for {
-      inputType <- outputTypeMapWithIDs.get(lookupId).
-        toRight(NonEmptyList.one(s"couldn't find $lookupId as derived from $source in map\n${outputTypeMapWithIDs.mkString("\n")}"))
+      //lookup each of our source Ids, failing if any of them are missing
+      inputTypes <- sources.traverse[ErrorOr, WomType](lookupId).toEither
+      //we may have several sources, we make sure to have a type common to all of them
+      inputType = WomType.lowestCommonSubtype(inputTypes)
```
Horneth: This will yield `WomAnyType` as a last resort if there's no better common (super) type.

cwl/src/main/scala/cwl/model.scala
```diff
@@ -47,7 +47,7 @@ case class WorkflowStepInput(
       //lookup each of our source Ids, failing if any of them are missing
       inputTypes <- sources.traverse[ErrorOr, WomType](lookupId).toEither
       //we may have several sources, we make sure to have a type common to all of them
-      inputType = WomType.lowestCommonSubtype(inputTypes)
+      inputType = WomType.homogeneousTypeFromTypes(inputTypes)
```
Horneth: Do we actually want to constraint the source to being of the same type ? Or does the spec allow for sources of different types ?

danbills: This falls under MultipleInputFeatureRequirement, which I figured I would tackle next.

That being said, the spec doesn't address the case of input sources being different type, so I'm really just giving a best guess here. I can comment to that effect.

Horneth: Oh right my bad, for some reason I thought this PR was for `MultipleInputFeatureRequirement`, I didn't realize there was a Step input requirement too
I would be surprised if inputs could only be of the same type but :man_shrugging:

geoffjentry:
@danbills If you didn't spot it the Expression stuff is merged, so #63 might be unlocked as well

cwl/src/main/scala/cwl/WorkflowStep.scala
```diff
@@ -134,58 +126,69 @@ case class WorkflowStep(
...
+            * If we don't know about them, we find upstream nodes and build them (see "buildUpstreamNodes").
+             */
+
+          val inputSources: List[String] = workflowStepInput.source.toList.flatMap(_.fold(WorkflowStepInputSourceToStrings))
+
+          val baseCase = (Map.empty[String, OutputPort], fold.generatedNodes).asRight[NonEmptyList[String]]
+          val inputMappingsAndGraphNodes: Checked[(Map[String, OutputPort], Set[GraphNode])] =
+            inputSources.foldLeft(baseCase){
+              case (Right((sourceMappings, graphNodes)), inputSource) =>
+                /*
```
mcovarr: ⬅️

PR #3205, opened 2018-01-27
More cwltool-like handling of shellQuote.

mcovarr:
Now with Hello World support.


cwl/src/main/scala/cwl/CwlExpressionCommandPart.scala
```diff
@@ -82,9 +82,9 @@ abstract class CommandLineBindingCommandPart(commandLineBinding: CommandLineBind
       case _ => List.empty
     }
 
-    // Can't flatMap Either in 2.11.
-    evaluatedWomValue match {
-      case Right(womValue) => applyShellQuote(womValue).map(processValue).map(_.map(InstantiatedCommand(_))).toValidated
+    val instantiations = evaluatedWomValue map { v => processValue(v) map applyShellQuote map (InstantiatedCommand(_)) }
+    instantiations match {
```
Horneth: isn't this `toValidated` ?

mcovarr: it is now

PR #3207, opened 2018-01-29
Added CWL Any plus defaults for WomNothing

kshakir:


wom/src/main/scala/wom/expression/InputPointerToWomValue.scala
```diff
@@ -21,14 +21,22 @@ object InputPointerToWomValue extends Poly1 {
   implicit def fromOutputPort: Case.Aux[OutputPort, ToWdlValueFn] = at[OutputPort] {
     port => (knownValues: Map[String, WomValue], ioFunctions: IoFunctionSet, outputPortLookup: OutputPortLookup, inputDefinition: InputDefinition) =>
       (outputPortLookup(port), inputDefinition) match {
-        case (v: Valid[WomValue], _) => v.map(inputDefinition.valueMapper(ioFunctions))
+        case (Valid(womValue), _) if isDefined(womValue) =>
```
Horneth: Just curious about the addition of `ifDefined`

kshakir: Conformance test \#15 was failing because `outputPortLookup(port)` was returning a `Valid(WomOptionalValue(WomAnyType, None))` that is effectively `null`/`None`. Without the `isDefined`, this case statement on line 24 was matching, when instead the next case statement below should have been matching and using the `default` value provided within the CWL.

wom/src/main/scala/wom/expression/InputPointerToWomValue.scala
```diff
@@ -21,14 +21,22 @@ object InputPointerToWomValue extends Poly1 {
...
           evaluate(defaultExpression, knownValues, ioFunctions).map(valueMapper(ioFunctions))
         case (_, OptionalInputDefinition(_, optionalType, valueMapper)) => valueMapper(ioFunctions)(optionalType.none).validNel
         case _ => s"Failed to lookup input value for required input ${port.name}".invalidNel
       }
   }
 
+  private def isDefined(womValue: WomValue) = {
+    womValue match {
+      case WomOptionalValue(_, None) => false
+      case _ => true
```
cjllanwarne: Not sure if possible in CWL but generally
`case WomOptionalValue(_, Some(inner: WomOptinalValue)) => isDefined(inner)`

PR #3211, opened 2018-01-29
Pluggable language support

### LINKED ISSUE #3195: Pluggable Languages
Get languages to come in from the configuration file rather than be statically known.

cjllanwarne:
Closes #3195 

- Mostly cut and paste from `MaterializeWorkflowDescriptorActor` to the two new language projects.
- `MaterializeWorkflowDescriptorActor` now gets its language support from one of the plugged in languages and `engine` is not allowed to access language classes directly (except that it still can via `databaseMigration` but pretend you don't know that)
- Doesn't do anything to split WDL into two versions. That can come later...!
- If no version is specified, we just pick the first version from the list. I suspect we'll actually want some kind of a "decider" function based on the file contents but walking before running...


build.sbt
```diff
@@ -177,5 +194,6 @@ lazy val root = (project in file("."))
   .aggregate(wdl)
   .aggregate(wom)
   .aggregate(womtool)
+  .aggregate(languageFactoryCore)
```
kshakir: Add all the new sub-projects to aggregate. Otherwise they will not be part of the root build.

For example right now they will NOT be tested during `sbt test`.

project/Dependencies.scala
```diff
@@ -278,6 +279,15 @@ object Dependencies {
...
   )
 
+  val languageFactoryCoreDependencies = List(
+  )
+
+  val wdlD2LanguageFactoryDependencies = List(
+  )
+
+  val cwlV1LanguageFactoryDependencies = List(
+  )
```
kshakir: These variables are unnecessary. Delete them from here and where they are currently used. The dependencies are an optional parameter. See `lazy val jesBackend` in build.sbt that doesn't have any extra dependencies.

common/src/main/scala/common/validation/Parse.scala
```diff
@@ -0,0 +1,30 @@
...
+
+  // If anyone has a magic import that does exactly what these helpers do, please replace, thx!
+
+  def errorOrParse[A](f: => ErrorOr[A]): Parse[A] = fromEither[IO](f.toEither)
+
+  def tryParse[A](f: => Try[A]): Parse[A] = errorOrParse(f.toErrorOr)
+
+  def goParse[A](f: => A): Parse[A] = tryParse(Try(f))
+
+}
```
Horneth: \n

engine/src/main/scala/cromwell/engine/language/CromwellLanguages.scala
```diff
@@ -0,0 +1,44 @@
...
+/**
+  * Holds all the registered versions of a language, and a version decider for unknown/unspecified versions.
+  */
+case class LanguageVersions(allVersions: Map[CromwellLanguageVersion, LanguageFactory])
+
+object CromwellLanguages {
+  type CromwellLanguageName = String
+  type CromwellLanguageVersion = String
+
+  var Instance: CromwellLanguages = _
```
Horneth: this should probably be private and have a public accessor

engine/src/main/scala/cromwell/engine/language/CromwellLanguages.scala
```diff
@@ -0,0 +1,44 @@
+package cromwell.engine.language
+
+import cromwell.engine.language.CromwellLanguages._
+import cromwell.engine.workflow.lifecycle.materialization.LanguageFactory
+
+/**
+  * Provides a global singleton access to the instantiated language factories.
```
Horneth: `global singleton` made me think this was going to be an `object`. Could we move that part on the `Instance` below that actually represents the global singleton ?

engine/src/main/scala/cromwell/engine/language/CromwellLanguages.scala
```diff
@@ -0,0 +1,44 @@
...
+  * @param languageConfig A set of language configuration elements.
+  */
+case class CromwellLanguages(languageConfig: List[LanguageConfigurationEntry]) {
+
+  val Languages: Map[CromwellLanguageName, LanguageVersions] = (languageConfig map { lc =>
```
Horneth: Since this is not an `object` it should be lower `l`

engine/src/main/scala/cromwell/engine/language/LanguageConfiguration.scala
```diff
@@ -0,0 +1,42 @@
...
+import cromwell.engine.language.CromwellLanguages.{CromwellLanguageName, CromwellLanguageVersion}
+
+import scala.collection.JavaConverters._
+
+case class LanguageConfigurationEntry(name: CromwellLanguageName, versions: Map[CromwellLanguageVersion, String])
+
+
+object LanguageConfiguration {
+  private val LanguagesConfig = ConfigFactory.load.getConfig("languages")
+  private val LanguageNames: Set[String] = LanguagesConfig.entrySet().asScala.map(findFirstKey).toSet
```
Horneth: ToL: might be a bit easier to use ficus to browse through the config instead of the built-in config methods

engine/src/main/scala/cromwell/engine/language/LanguageConfiguration.scala
```diff
@@ -0,0 +1,42 @@
...
+
+  def findFirstKey(entry: Entry[String, _]): String = {
+    val NoQuoteFirstKey = """([^.]*)\.(.*)""".r
+    val SingleQuotedFirstKey = """"(.*)"\.(.*)""".r
+
+    val r = entry.getKey match {
+      case SingleQuotedFirstKey(firstKey, _) => firstKey
+      case NoQuoteFirstKey(firstKey, _) => firstKey
+    }
+    r
```
Horneth: unnecessary temp var

engine/src/main/scala/cromwell/engine/language/LanguageConfiguration.scala
```diff
@@ -0,0 +1,42 @@
...
+
+    val versions = (languageVersionNames.toList map { languageVersionName =>
+      val entry = languageConfig.getConfig(s""""$languageVersionName"""")
+      languageVersionName -> entry.getString("language-factory")
+    }).toMap
+
+    LanguageConfigurationEntry(languageName, versions)
+  }
+
+  def findFirstKey(entry: Entry[String, _]): String = {
```
Horneth: Can we get a comment about what this is for ?

engine/src/main/scala/cromwell/engine/language/CromwellLanguages.scala
cjllanwarne: Is that the rule? I never actually worked out when to use Upper vs lower. `val x: <some object>` works for me!

build.sbt
cjllanwarne: Done

common/src/main/scala/common/validation/Parse.scala
cjllanwarne: Done

build.sbt
```diff
@@ -144,14 +142,34 @@ lazy val womtool = project
...
+lazy val languageFactoryCore = (project in languageFactoryRoot / "language-factory-core")
+  .withExecutableSettings("language-factory-core")
+  .dependsOn(engine)
+
+lazy val wdlD2LanguageFactory = (project in languageFactoryRoot / "wdl-d2")
+  .withExecutableSettings("wdl-d2")
+  .dependsOn(languageFactoryCore)
+  .dependsOn(wdl)
+
+lazy val cwlV1LanguageFactory = (project in languageFactoryRoot / "cwl-v1-0-0")
```
geoffjentry: The intention is that we're supporting `1.0.x` CWL

core/src/main/resources/reference.conf
```diff
@@ -245,6 +245,11 @@ engine {
   }
 }
 
+languages {
```
geoffjentry: Placeholder to remind me i was going to talk to you in person, assuming my thought doesn't change after i go through the whole PR

engine/src/main/scala/cromwell/engine/language/CromwellLanguages.scala
```diff
@@ -0,0 +1,43 @@
+package cromwell.engine.language
+
+import cromwell.engine.language.CromwellLanguages._
+import cromwell.engine.workflow.lifecycle.materialization.LanguageFactory
+
+// Construct a singleton instance of this class using 'initLanguages' below.
+case class CromwellLanguages private(languageConfig: List[LanguageConfigurationEntry]) {
```
geoffjentry: final

cjllanwarne:
@geoffjentry oops, corrected, is this right?
```
languages {
  WDL {
    versions = [
      {
        version = "draft-2"
        language-factory = "languages.wdl.WdlDraft2LanguageFactory"
      }
    ]
  }
  CWL {
    versions = [
      {
        version = "v1.0"
        language-factory = "languages.cwl.CwlV1LanguageFactory"
      }
    ]
  }
}
```

cjllanwarne:
Current `languages` config style:
```

languages {
  WDL {
    versions {
      "draft-2" {
        language-factory = "languages.wdl.WdlDraft2LanguageFactory"
      }
    }
  }
  CWL {
    versions {
      "v1.0" {
        language-factory = "languages.cwl.CwlV1LanguageFactory"
      }
    }
  }
}
```

geoffjentry:
👍

engine/src/main/scala/cromwell/engine/language/LanguageConfiguration.scala
```diff
@@ -0,0 +1,50 @@
...
+  // EG
+  // foo {
+  //   "1.0" {
+  //     "x.y".z = "hello"
+  //   }
+  // }
+  // Called on the entry on line 1, returns 'foo'
+  // Called on the entry on line 2, returns '1.0'
+  // Called on the entry on line 3, returns 'x.y'
+  private def findFirstKey(entry: Entry[String, _]): String = {
```
Horneth: I'm surprised you have to do this manually and typesafe doesn't cleanly give you the unquoted key but :man_shrugging:

build.sbt
```diff
@@ -144,14 +144,34 @@ lazy val womtool = project
...
+lazy val cwlV1_0LanguageFactory = (project in languageFactoryRoot / "cwl-v1-0")
+  .withExecutableSettings("cwl-v1-0")
+  .dependsOn(languageFactoryCore)
+  .dependsOn(cwl)
+
 lazy val root = (project in file("."))
   .withExecutableSettings("cromwell", rootDependencies, rootSettings)
   // Next level of projects to include in the fat jar (their dependsOn will be transitively included)
+  // Must 'dependsOn' any projects which might be loaded at runtime from the fat jar
+  // NB: even if you 'dependsOn' a project, you still need to aggregate it (c'est la vie...)
```
mcovarr: 😕 (that's directed at sbt, not you)

engine/src/main/scala/cromwell/engine/language/LanguageConfiguration.scala
```diff
@@ -0,0 +1,50 @@
...
+  private val LanguageNames: Set[String] = LanguagesConfig.entrySet().asScala.map(findFirstKey).toSet
+
+  val AllLanguageEntries: List[LanguageConfigurationEntry] = LanguageNames.toList map { languageName =>
+
+    val languageConfig = LanguagesConfig.getConfig(languageName)
+    val versionSet = languageConfig.getConfig("versions")
+    val languageVersionNames: Set[String] = versionSet.entrySet().asScala.map(findFirstKey).toSet
+
+    val versions = (languageVersionNames.toList map { languageVersionName =>
+      val entry = versionSet.getConfig(s""""$languageVersionName"""")
```
mcovarr: am I missing something or is this an elaborately roundtripped String?

engine/src/main/scala/cromwell/engine/language/LanguageConfiguration.scala
```diff
@@ -0,0 +1,50 @@
...
+      languageVersionName -> entry.getString("language-factory")
```
mcovarr: it seems all this could be on one simpler line

engine/src/main/scala/cromwell/engine/language/LanguageConfiguration.scala
```diff
@@ -0,0 +1,50 @@
...
+  // Called on the entry on line 3, returns 'x.y'
+  private def findFirstKey(entry: Entry[String, _]): String = {
+    val NoQuoteFirstKey = """([^.]*)\.(.*)""".r
+    val SingleQuotedFirstKey = """"(.*)"\.(.*)""".r
+
+    entry.getKey match {
+      case SingleQuotedFirstKey(firstKey, _) => firstKey
+      case NoQuoteFirstKey(firstKey, _) => firstKey
+    }
+  }
```
mcovarr: couldn't this use a syntax that stays within HOCON?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/materialization/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -553,24 +349,7 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
...
-    }
-
-    NonEmptyList.fromList(failedFiles.toList) match {
-      case Some(errors) => Left(errors)
-      case None => Right(())
-    }
-  }
-
-  private def validateWorkflowOptions(workflowOptions: WorkflowJson): ErrorOr[WorkflowOptions] = {
+  private def validateWorkflowOptions(workflowOptions: String): ErrorOr[WorkflowOptions] = {
```
mcovarr: loving all this red in MWDA

build.sbt
```diff
@@ -115,20 +115,20 @@ lazy val bcsBackend = (project in backendRoot / "bcs")
...
-  .dependsOn(wdl)
-  .dependsOn(cwl)
   .dependsOn(core % "test->test")
   .dependsOn(backend % "test->test")
   // In the future we may have a dedicated test backend like the `TestLocalAsyncJobExecutionActor`.
   // For now, all the engine tests run on the "Local" backend, an implementation of an impl.sfs.config backend.
   .dependsOn(sfsBackend % "test->compile")
   .dependsOn(gcsFileSystem % "test->test")
   .dependsOn(ossFileSystem % "test->test")
+  .aggregate(CwlV1Project)
```
kshakir: What does this do? I've only ever seen [aggregate on the root sub-project](http://www.scala-sbt.org/1.0/docs/Multi-Project.html#Dependencies). You can also include that link as it documents the difference between `dependsOn` and `aggregate`.

engine/src/main/scala/cromwell/engine/language/LanguageConfiguration.scala
cjllanwarne: it's adding `"` to the start and end... elaborately

engine/src/main/scala/cromwell/engine/language/LanguageConfiguration.scala
cjllanwarne: So HOCON does support escaped keys, and if we always write keys on separate lines there's no issue. The problem comes when you try and put two keys on the same line, like
```
x {
  y.z = foo
}
```



`BackendConfigurationEntry` solves the same problem with a different string manipulation (the `.map(_.getKey.split("\\.")` which it hides by putting it inline). That doesn't consider the possibility that the first key is quoted.

build.sbt
cjllanwarne: This was an artifact me messing around trying to trick it to import the CWL code for tests only. I'm going to remove this.

PR #3228, opened 2018-01-31
Move label processing earlier in the process. Closes #3227

### LINKED ISSUE #3227: Publish labels to metadata prior to parsing workflow
I believe this is possible, at first examination it looks like code-wise label publishing depends on a successful namespace generation, but it also looks like the value it's depending on is not used. 

Assuming that statement holds, publish labels to metadata as early in the process of starting a workflow as possible

geoffjentry:


cjllanwarne:
Presumably to ensure we get the labels even when the workflow was invalid?

I think we can centaur that... I’m pretty sure we check metadata even on failure tests.

geoffjentry:
Exactly why. Since Cromiam fiddles with labels, if a workflow fails to validate that control information is lost

cjllanwarne:
Great, I'm happy to +1 but would like a centaur test to prevent regressions. 

Just a new `.labels` and verification of metadata on an existing failing test would be fine.

geoffjentry:
Yeah, I'm planning on adding it - was a good idea

geoffjentry:
@cjllanwarne added

centaur/src/main/resources/standardTestCases/labels_bad_workflow/badWorkflow.wdl
```diff
@@ -0,0 +1,44 @@
...
+    input:
+      asdf=foo
+  }
+  call hello_task2
+
+  output {
+    File foo = hello_task.foo
+    File foo2 = hello_task2.foo
+  }
+}
```
Horneth: This is a very complicated invalid workflow 😄

geoffjentry: It’s just the one which led to the initial bug report

cjllanwarne:
👍

PR #3229, opened 2018-01-31
Fully split WDL draft 3 from draft 2

### LINKED ISSUE #3214: Fork WDL draft-2 as WDL "draft-3"

cjllanwarne:
Closes #3214.

Top level view:
- Totally split `draft2` from a fresh copy of WDL code in `draft3`.
- They now inhabit separate and unconnected sbt projects.
- Any existing code references to `wdl.` are now references to `wdl.draft2.`
- Please review my `build.sbt` file, the rest is just moving, shuffling and fixing up intellij's failures to refactor packages.


build.sbt
```diff
@@ -50,7 +56,7 @@ lazy val databaseSql = (project in file("database/sql"))
 lazy val databaseMigration = (project in file("database/migration"))
   .withLibrarySettings("cromwell-database-migration", databaseMigrationDependencies)
   .dependsOn(core)
-  .dependsOn(wdl)
+  .dependsOn(wdlDraft2)
```
Horneth: I'm not sure I understand what's going to happen when draft-3 starts diverging. is this expected to depend on both draft2 and 3 ?

build.sbt
```diff
@@ -138,7 +144,7 @@ lazy val centaurCwlRunner = project
 
 lazy val womtool = project
   .withExecutableSettings("womtool", womtoolDependencies)
-  .dependsOn(wdl)
+  .dependsOn(wdlDraft2)
```
Horneth: Is there going to be a `womtool-draft2` and a `womtool-draft3` ? Or will you add a `draft3` dependency here when womtool needs to handle both ?

cjllanwarne: I'm hoping to have pluggable languages in `womtool` too, but that's probably a story for another PR

build.sbt
cjllanwarne: AFAIK this dependency is for a one-off migration which needed access to some WDL objects, as they were at the time of the migration. A nice effect of this hard split is that it allows us to freeze in time the objects which this historical migration relies on.

cjllanwarne:
Closing for now. I'll use my lessons learned here to advise the trajectory of the tech debt paydown route.

PR #3231, opened 2018-02-01
Interpolated strings.

mcovarr:


cwl/src/main/scala/cwl/ExpressionEvaluator.scala
```diff
@@ -24,6 +24,16 @@ object ExpressionEvaluator {
   type ECMAScriptFunction = String Refined MatchesRegex[ECMAScriptFunctionWitness.T]
   type MatchesECMAFunction = MatchesRegex[ECMAScriptFunctionWitness.T]
 
+  // This is not ECMAScript, just what CWL uses for interpolated expressions. A dollar sign immediately followed
+  // by a paren followed by anything that isn't a close paren followed by a close paren. Capturing groups around
+  // the fore, expression and after portions to concatenate fragments for evaluation. Obviously this will break
+  // with nested parens. It's not clear what the specification is here; this simplistic implementation may need
+  // to be revisited.
+  val InterpolatedStringWitness = Witness("""(?s)(.*)\$\(([^\)]+)\)(.*)""")
```
kshakir: Does this work for my [favorite test](https://github.com/common-workflow-language/common-workflow-language/blob/91f108df4d4ca567e567fc65f61feb0674467a84/v1.0/v1.0/search.cwl#L62)?

mcovarr: Good catch! No it won't because that has multiple expressions in the string. I wasn't previously aware of anything like that in the wild. 😦  I'll update the code and test to handle > 1.

mcovarr:
Ready for re-review, now handles >1 expression in an interpolated string.

cwl/src/main/scala/cwl/ExpressionEvaluator.scala
```diff
@@ -24,6 +25,16 @@ object ExpressionEvaluator {
   type ECMAScriptFunction = String Refined MatchesRegex[ECMAScriptFunctionWitness.T]
   type MatchesECMAFunction = MatchesRegex[ECMAScriptFunctionWitness.T]
 
+  // This is not ECMAScript, just what CWL uses for interpolated expressions. A dollar sign immediately followed
```
geoffjentry: A+ commenting

cwl/src/main/scala/cwl/ExpressionEvaluator.scala
```diff
@@ -24,6 +25,16 @@ object ExpressionEvaluator {
...
+  // by a paren followed by anything that isn't a close paren followed by a close paren. Capturing groups around
+  // the fore, expression and after portions to concatenate fragments for evaluation. Obviously this will break
+  // with nested parens. It's not clear what the specification is here; this simplistic implementation may need
```
geoffjentry: Please ask & verify what they think the spec means before merging

geoffjentry:
:+1: modulo a request to verify what they believe the spec indicates

cwl/src/main/scala/cwl/ExpressionEvaluator.scala
mcovarr: I reviewed the spec to formulate a question but it turns out I need to revise this comment since the spec is actually [quite clear](http://www.commonwl.org/v1.0/CommandLineTool.html#Parameter_references) on this. What I'm not sure about is if our expression evaluation currently does that "Objects entries are sorted by key" bit already. I can pick the relevant people's brains on this next week.

geoffjentry: Sounds good. I thought it was pretty straightforward on their side and that you had it right but my spec-fu was failing earlier tonight.

cwl/src/main/scala/cwl/ExpressionEvaluator.scala
```diff
@@ -24,6 +25,16 @@ object ExpressionEvaluator {
...
   type MatchesECMAFunction = MatchesRegex[ECMAScriptFunctionWitness.T]
 
+  // This is not ECMAScript, just what CWL uses for interpolated expressions. The pattern is 'before$(expression)after'.
+  // The regex uses a non-greedy quantifier on the 'before' to allow the expression to be processed from left to right,
+  // and there are capturing groups around each portion. The official specification for interpolated strings is in the
+  // last part of this section:
+  // http://www.commonwl.org/v1.0/CommandLineTool.html#Parameter_references
+  val InterpolatedStringWitness = Witness("""(?s)(.*?)\$\(([^\)]+)\)(.*)""")
+  val InterpolatedStringRegex = InterpolatedStringWitness.value.r
+  type InterpolatedString = String Refined MatchesRegex[InterpolatedStringWitness.T]
```
danbills: ToL could be `String Refined MatchesInterpolatedString` as you declared the type alias below

cwl/src/main/scala/cwl/ExpressionEvaluator.scala
```diff
@@ -60,6 +71,28 @@ object ExpressionEvaluator {
     }
   }
 
+  def evalInterpolatedString(string: InterpolatedString, parameterContext: ParameterContext, expressionLib: ExpressionLib): ErrorOr[WomValue] = {
+    def interpolate(remaining: String, acc: String = ""): ErrorOr[String] = {
```
danbills: ToL I'm glad this works so it's ToL, but it does feel a little like re-inventing the wheel vs. using a parser library.  It would potentially be less code, arguably easier to maintain

cwl/src/main/scala/cwl/ExpressionEvaluator.scala
mcovarr: Interesting, I essentially copy/pasted from the `ECMAScript{Function|Expression}` definitions above this and assumed there was a reason why it had to be this way. :man_shrugging:

cwl/src/main/scala/cwl/ExpressionEvaluator.scala
mcovarr: tbh I don't think I could write this using a parser library in a justifiable amount of time. 😢

PR #3234, opened 2018-02-03
Dependency injection version of a draft2/draft3 split

cjllanwarne:
Takes the proof of concept as far as:

- A core WDL package which is the base for draft 2 and draft 3.
- Answers "can I separate the common classes out of WdlParser?" I believe I can.
- Sees what it looks like to inject a new piece of functionality into the middle of WdlStandardLibrary to replace the `size` function (`size` was awkwardly implemented in full in `IoFunctions` which needed to be changed, so it was in some ways a bad example but in others it was a good one since it highlights the kinds of hurdles this draft2/draft3 splitting is going to involve in general).


wdl/wdl-core/src/main/scala/wdl/Scatter.scala
```diff
@@ -39,7 +40,10 @@ object Scatter {
   /**
     * @param preserveIndexForOuterLookups When we're evaluating the scatter collection, should we preserve scatter index when we have to use the outerLookup?
     */
-  def womScatterNode(scatter: Scatter, localLookup: Map[String, GraphNodePort.OutputPort], outerLookup: Map[String, OutputPort], preserveIndexForOuterLookups: Boolean): ErrorOr[ScatterNodeWithNewNodes] = {
+  def womScatterNode(scatter: Scatter,
+                     localLookup: Map[String, GraphNodePort.OutputPort],
+                     outerLookup: Map[String, OutputPort],
+                     preserveIndexForOuterLookups: Boolean)(implicit wdlVersionSpecifics: WdlVersionSpecifics): ErrorOr[ScatterNodeWithNewNodes] = {
```
cjllanwarne: Since almost everything has expressions, and expressions are one of the things I tried to split, it means that almost everything now has a `(implicit wdlVersionSpecifics: WdlVersionSpecifics)` on the end of its constructor

wdl/wdl-core/src/main/scala/wdl/expression/WdlStandardLibraryFunctions.scala
```diff
@@ -25,7 +29,16 @@ trait WdlStandardLibraryFunctions extends WdlFunctions[WomValue] {
...
-  def size(params: Seq[Try[WomValue]]): Try[WomFloat]
+  def singleFileSize(file: WomValue): Try[Double]
+
+  /**
+    * Gets the size of a file.
+    *
+    * @param params First parameter must be a File or File? or coerceable to one. The second is an optional string containing the size unit (eg "MB", "GiB")
+    */
+  def size(params: Seq[Try[WomValue]]): Try[WomFloat] = {
+    wdlVersionSpecifics.sizeFunctionOverride(params, singleFileSize)
```
cjllanwarne: This is the hole in `wdl-core` which relies on there being a `wdlVersionSpecifics` injected.

wdl/wdl-core/src/main/scala/wdl/versioning/WdlVersionSpecifics.scala
```diff
@@ -0,0 +1,10 @@
+package wdl.versioning
+
+import wom.values.{WomFloat, WomValue}
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.Try
+
+trait WdlVersionSpecifics {
+  def sizeFunctionOverride(params: Seq[Try[WomValue]], sizeFunc: WomValue => Try[Double]): Try[WomFloat]
```
cjllanwarne: The trait which carries around all of the functions to slot into the various holes in `wdl-core`

wdl/wdl-draft2/src/test/scala/wdl/ReadLikeFunctionsSpec.scala
```diff
@@ -0,0 +1,124 @@
...
+class ReadLikeFunctionsSpec extends FlatSpec with Matchers {
+
+  behavior of "ReadLikeFunctions.size"
+
+  it should "correctly report a 2048 byte file, in bytes by default" in {
+    val readLike = draft2Functions(Success(2048d))
+    readLike.size(Seq(Success(WomSingleFile("blah")))) should be(Success(WomFloat(2048d)))
+  }
+
+//  it should "correctly report a 2048 byte file, in bytes" taggedAs PostWomTest ignore {
```
cjllanwarne: These are commented out because 
- They're ignored already
- Fixing them wasn't the point of this PoC

If we go ahead with it, I don't see why they shouldn't be brought back as test cases

wdl/wdl-core/src/main/scala/wdl/Scatter.scala
mcovarr: If it wasn't for the `implicit` I think all this would basically be a reader monad. 😮

wdl/wdl-core/src/main/scala/wdl/WorkflowOutput.scala
```diff
@@ -1,16 +1,19 @@
...
     val expression = WdlExpression(ast.getAttribute("expression"))
     WorkflowOutput(name, womType, expression, ast, parent)
   }
 }
 
-case class WorkflowOutput(unqualifiedName: String, womType: WomType, requiredExpression: WdlExpression, ast: Ast, override val parent: Option[Scope]) extends Output
\ No newline at end of file
+case class WorkflowOutput(unqualifiedName: String, womType: WomType, requiredExpression: WdlExpression, ast: Ast, override val parent: Option[Scope])(implicit implicitVersionSpecifics: WdlVersionSpecifics) extends Output {
+  override val wdlVersionSpecifics = implicitVersionSpecifics
+}
```
mcovarr: \n

cjllanwarne:
Closing for now. Thank you all for looking. I'll use my lessons learned here to advise the trajectory of the tech debt paydown route.

PR #3236, opened 2018-02-05
Relative paths Part 1

Horneth:
Don't let the number of lines scare you, they're mostly json used in tests to assert that the pre-processed CWL looks right.
The first commit is only the pre-processing code + its tests.
The second actually uses it.
The 3rd and 4th are about abstracting code in the CromwellEndpoint and the CommandLineParser to accommodate for a "submit" command that can submit to a cromwell server + a few adjustments.
The submit thing could be isolated in its own sub-project I guess since it doesn't need Cromwell engine but just the api client.
I might do that in a part 1.1.

But there will be a part 2 needed because the input files (as the "files referenced in the input json") also need to be massaged somehow, because they're currently assumed to be relative to where cromwell is being run although they're actually relative to the input json.


cwl/src/main/scala/cwl/preprocessor/CwlPreProcessor.scala
```diff
@@ -0,0 +1,272 @@
...
+    def flatten: Parse[ProcessedJsonAndDependencies] = flattenCwlReference(CwlReference(file, cwlRoot), Map.empty, Map.empty, Set.empty)
+    
+    for {
+      original <- parseYaml(file.contentAsString)
+      flattened <- original
+        .asObject
+        .flatMap(_.kleisli("id").flatMap(_.asString).flatMap(CwlReference.fromString)) match {
+        // This by passes the pre-processing if the file already has an id for which the file doesn't match the path of the file
+        // passed to this function, as this would indicate that it has already been saladed and pre-processed.
+        case Some(contentRef) if !contentRef.file.equals(reference.file) => original.validParse
```
Horneth: This is not great and kinda hacky but the best way I could find to not pre-process files that have (very likely) already been processed.
I tried to make the whole thing idempotent to avoid this issue but turns out salading is not full idempotent and this code would also need a few adjustments to be idempotent. I'll try to give it another shot.

cwl/src/main/scala/cwl/preprocessor/CwlPreProcessor.scala
```diff
@@ -0,0 +1,272 @@
...
+        case _ => flatten.map(_.processedJson)
+      }
+    } yield flattened
+  }
+
+  /**
+    * Convenience method to get the processed workflow as a string.
+    */
+  def preProcessCwlFileToString(file: BFile, cwlRoot: Option[String]): ErrorOr[String] = {
+    preProcessCwlFile(file, cwlRoot).value.unsafeRunSync().map(_.printCompact).toValidated
```
danbills: I see you there `unsafeRunSync`!  I know it feels tedious but this is what Parse is supposed to wrap, thus I think this should return Parse[String]

cwl/src/main/scala/cwl/preprocessor/CwlPreProcessor.scala
```diff
@@ -0,0 +1,272 @@
...
+    * @return a new ProcessedReferences Map including this cwlReference processed along with all the dependencies
+    *         that might have been processed recursively.
+    */
+  private def processCwlRunReference(unProcessedReferences: UnProcessedReferences,
+                                     breadCrumbs: Set[CwlReference])
+                                    (checkedProcessedReferences: Parse[ProcessedReferences],
+                                     cwlReference: CwlReference): Parse[ProcessedReferences] = {
+    checkedProcessedReferences flatMap { processedReferences =>
+      // If the reference has already been processed, no need to do anything
+      if (processedReferences.contains(cwlReference)) processedReferences.validParse
```
danbills: ToL consider pattern matching on a tuple of `(processedReferences.contains(cwlReference), breadCrumbs.contains(cwlReference), unProcessedReferences.get(cwlReference))`

cwl/src/main/scala/cwl/preprocessor/CwlPreProcessor.scala
```diff
@@ -0,0 +1,272 @@
...
+        val result: Parse[ProcessedJsonAndDependencies] = unProcessedReferences.get(cwlReference) match {
+          case Some(unProcessedReferenceJson) =>
+            // Found the json in the unprocessed map, no need to reparse the file, just flatten this json
+            flattenJson(unProcessedReferenceJson, unProcessedReferences, processedReferences, breadCrumbs)
+          case None =>
+            // This is the first time we're seeing this reference, we need to parse its file and flatten it
+            flattenCwlReference(cwlReference, unProcessedReferences, processedReferences, breadCrumbs)
+        }
+
+        result map {
```
danbills: ToL feels like this `processCwlRunReference` function could be structured as a forcomp

cwl/src/main/scala/cwl/preprocessor/CwlPreProcessor.scala
```diff
@@ -0,0 +1,272 @@
...
+    *     This is used to be able to detect circular dependencies (if the cwlReference being processed is in that list, then we have a circular dependency) .
+    *
+    */
+  def preProcessCwlFile(file: BFile, cwlRoot: Option[String]): Parse[Json] = {
+    val reference = CwlReference(file, cwlRoot)
+    def flatten: Parse[ProcessedJsonAndDependencies] = flattenCwlReference(CwlReference(file, cwlRoot), Map.empty, Map.empty, Set.empty)
+    
+    for {
+      original <- parseYaml(file.contentAsString)
+      flattened <- original
```
danbills: ToL Consider declare a function somewhere to capture all this logic?

src/main/scala/cromwell/CromwellApp.scala
```diff
@@ -0,0 +1,49 @@
...
+  //  -i, --inputs <value>     Workflow inputs file.
+  //  -o, --options <value>    Workflow options file.
+  //  -t, --type <value>       Workflow type.
+  //  -v, --type-version <value>
+  //                           Workflow type version.
+  //  -l, --labels <value>     Workflow labels file.
+  //  -p, --imports <value>    A directory or zipfile to search for workflow imports.
+  //  -m, --metadata-output <value>
+  //                           An optional directory path to output metadata.
+  //  -h, --host               Cromwell server URL
```
mcovarr: not sure why this was moved over here?

PR #3238, opened 2018-02-05
Return host absolute paths and not ephemeral container paths.

mcovarr:


supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesJobPaths.scala
```diff
@@ -31,6 +31,7 @@ final case class JesJobPaths(override val workflowPaths: JesWorkflowPaths, jobKe
 
   val jesMonitoringScriptFilename: String = s"${JesJobPaths.JesMonitoringKey}.sh"
 
+  // This comment appears to be vestigial.
```
mcovarr: And by 'this' I mean 'that', the comment below 'this'.

backend/src/main/scala/cromwell/backend/wdl/OutputEvaluator.scala
```diff
@@ -90,7 +91,12 @@ object OutputEvaluator {
       def toError(outputPort: OutputPort) = s"Missing output value for ${outputPort.identifier.fullyQualifiedName.value}"
 
       jobDescriptor.taskCall.expressionBasedOutputPorts.diff(outputs.keySet.toList) match {
-        case Nil => ValidJobOutputs(CallOutputs(outputs))
+        case Nil =>
+          val errorMessagePrefix = "Error applying postMapper in short-circuit output evaluation"
```
Horneth: I'm not sure this will be very helpful to anyone that encounters it but on the other hand it accurately describes what happened 😄

mcovarr: Yeah that's certainly true. I wasn't sure what to say here as this code is acting like its generic but it actually exists solely for short-circuit CWL output evaluation yet lives under `cromwell.backend.wdl`. 😱

Horneth: :/ the package name definitely is a pre-wom leftover, if anything it should be `backend/wom`.

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/JesJobPaths.scala
danbills: I'd agree and since I can't readily find its home I'd assume it's orphaned.  IMO it can go.

PR #3239, opened 2018-02-05
Sanitize LanguageFactory interface

cjllanwarne:
A bit of a shuffle of classes to tidy up the pluggable language interface. In particular see `build.sbt` where this PR inverts the previous "`languageFactory` depends on `engine`" dependency.

Good for a hundred reasons but also opens the door to statically imported `ToExecutable` methods from language factories if we wanted to.


cjllanwarne:
I tried to further decompose `core` into submodules to be more exact about what I was bringing in, but I was ending up with 50 different sub-core projects and honestly I don't think super fine-grain imports are such a big deal now that the circular dependency problems are gone.

core/src/main/scala/cromwell/core/MetadataQuery.scala
```diff
@@ -1,9 +1,8 @@
-package cromwell.services.metadata
+package cromwell.core
```
Horneth: Wait how does core know about metadata stuff ?

Horneth: Oh sorry, everything it references is in this file. I'm not sure about Metadata classes being defined in core though :/
At least I would move it to a `metadata` package.

geoffjentry: Revert

cjllanwarne: This is so that the conversion from X to WOM can create a list of `MetadataEvent`s to later be sent to the metadata service.

I wonder if I can change this to instead create a list of "things that will eventually become `MetadataEvent`s" and get the MWDA to "convert and send" instead of just forwarding on the prebuilt `List[MetadataEvent]`

Horneth: It would probably end up being 
```scala
case class NotMetadataEvent { copy / paste form MetadataEvent }
``` 
but it might be better than having the real Metadata classes defined in core.

geoffjentry: I suppose I **can** envision having key metadata concepts defined outside of `services`, and then the stuff behind `services` is just responsible for how metadata events are persisted.

However, **if** that's going to happen it should be pretty wholesale & well thought out, and not just moving a single class across because it's convenient

engine/src/main/scala/cromwell/engine/workflow/lifecycle/materialization/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -227,21 +228,51 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
...
       _ <- publishLabelsToMetadata(id, labels)
       validatedNamespace <- namespaceValidation
+      _ <- pushImportedFilesToMetadata(validatedNamespace.importedFileContent)
+      _ <- pushWfInputsToMetadata(validatedNamespace.womValueInputs)
       _ <- pushWfNameMetadataService(validatedNamespace.executable.entryPoint.name)
       ewd <- fromEither[IO](buildWorkflowDescriptor(id, sourceFiles, validatedNamespace, workflowOptions, labels, conf, pathBuilders).toEither)
     } yield ewd
   }
 
+  private def pushWfInputsToMetadata(workflowInputs: Map[OutputPort, WomValue]): Parse[Unit] = {
```
danbills: ToL This function and the subsequence one could have no side effects if it just returned `PutMetadataAction` and left the actor sending to the caller.

Also no need to wrap it in the `Parse` effect.

languageFactories/wdl-draft2/src/main/scala/languages/wdl/WdlDraft2LanguageFactory.scala
```diff
@@ -69,40 +61,29 @@ class WdlDraft2LanguageFactory() extends LanguageFactory {
...
-    /* Publish `MetadataEvent`s for all imports in this `WdlNamespace`. */
-    def publishImportMetadata(wdlNamespace: WdlNamespace): Unit = {
-      def metadataEventForImportedNamespace(ns: WdlNamespace): MetadataEvent = {
-        import MetadataKey._
-        import WorkflowMetadataKeys._
-        // This should only be called on namespaces that are known to have a defined `importUri` so the .get is safe.
-        val escapedUri = ns.importUri.get.escapeMeta
-        MetadataEvent(MetadataKey(
-          workflowIdForLogging, None, SubmissionSection, SubmissionSection_Imports, escapedUri), MetadataValue(ns.sourceString))
-      }
```
mcovarr: woah why was this stuff ever here?

PR #3247, opened 2018-02-07
WDL projects refactor part I: Separate WDL model from WOM transforms

cjllanwarne:
This is a little orthogonal to the actual mechanics of sharing code between draft 2 from draft 3, but it is already pointing to a nice looking path... But whatever scheme we use, this is hopefully going to make everything a hell of a lot neater.

- Introduces type classes as an alternative to baking `toWomX` into the WDL model classes.
- Starts by assuming that all model is shared and all transformation is draft 2. This will change.
- This diff is a nightmare (`If.scala` still exists, it did *not* get renamed to `WdlDraft2WomConditionalNodeMaker.scala`). You're probably better off checking out the branch locally and looking at the new layouts in the `wdl` projects.


wdl/transforms/draft2/src/main/scala/wdl/transforms/draft2/wdlom2wom/WdlDraft2WomWorkflowDefinitionMaker.scala
```diff
@@ -0,0 +1,25 @@
...
+        wg,
+        wdlWorkflow.meta,
+        wdlWorkflow.parameterMeta,
+        List.empty)
+    }
+  }
+
+  // TODO Dan: Is adding this back as an implicit class a hideous abuse of power, or the whole point of doing this?
+  implicit class Draft2EnhancedWdlWorkflow(val wf: WdlWorkflow) extends AnyVal {
+    def toWomWorkflowDefinition = WdlDraft2WomWorkflowDefinitionMaker.toWomWorkflowDefinition(wf)
```
cjllanwarne: @danbills, I had a thought after I made the PR. I suspect the magic I've missed here is along the lines of giving this implicit class to the "WomXMaker" type classes rather than the specializations, something like (not tried yet, just want to make sure I'm not going down the wrong path):
```
trait XMaker[A] { def makeX(a: A) }
implicit class xMakerWrapper[A](a: A)(implicit val xMakerInstance: XMaker[A]) {
  def turnIntoAnX = xMakerInstance(a)
}
```

danbills: This is absolutely the correct path, and it's so common that there's a library to generate this boilerplate for you!  Even better, it's already in scope.  

Use the `@typeclass` annotation before `trait XMaker[A]`, and it gives you that `XMakerWrapper` for free.  See https://github.com/mpilquist/simulacrum for the details and how to import (creates a .ops that must be imported to use the generated stuff)

cjllanwarne: Bleargh. I just went away and added these manually. But I learned a lot by doing it and shifting the typeclasses out into their own package was a good idea regardless so not totally wasted effort.

wdl/transforms/draft2/src/main/scala/wdl/transforms/draft2/wdlom2wom/WdlDraft2WomExecutableMaker.scala
```diff
@@ -0,0 +1,16 @@
...
+import common.Checked
+import wdl.WdlNamespaceWithWorkflow
+import wom.executable.{Executable, WomExecutableMaker}
+
+object WdlDraft2WomExecutableMaker extends WomExecutableMaker[WdlNamespaceWithWorkflow] {
+  override def toWomExecutable(wdlNamespaceWithWorkflow: WdlNamespaceWithWorkflow, inputFile: Option[String] = None): Checked[Executable] = {
+    WdlInputParsing.buildWomExecutable(wdlNamespaceWithWorkflow.workflow, inputFile)
+  }
+
+  // TODO Dan: Is adding this back as an implicit class a hideous abuse of power, or the whole point of doing this?
```
danbills: See my comment regarding simulacrum.  This is done for you but for any A that has an instatnce of  WomExecutableMaker.  Thus you can delete this whole object and for any instance of `WdlNamespaceWithWorkflow` call `.toWomExecutable(inputs)`.  Notice how it moves the `namespace` out of the definition for you because you already have it!

wdl/transforms/draft2/src/main/scala/wdl/transforms/draft2/wdlom2wom/WdlDraft2WomConditionalNodeMaker.scala
```diff
@@ -70,8 +49,8 @@ object If {
       case other => s"An if block must be given a boolean expression but instead got '${ifBlock.condition.toWomString}' (a ${other.toDisplayString})".invalidNel
     }
 
-    val innerGraphValidation: ErrorOr[Graph] = WdlGraphNode.buildWomGraph(
-      ifBlock,
+    // TODO Dan: can I avoid casting to Scope before using the implicit?
+    val innerGraphValidation: ErrorOr[Graph] = (ifBlock: Scope).toWomGraph(
```
cjllanwarne: Flagging this question too

wdl/transforms/draft2/src/main/scala/wdl/transforms/draft2/wdlom2wom/WdlDraft2WomExecutableMaker.scala
cjllanwarne: I'm going to give this simulacrum a go now (alas poor newly-added implicit classes, I barely knew thee!)

wdl/transforms/draft2/src/main/scala/wdl/transforms/draft2/wdlom2wom/WdlDraft2WomConditionalNodeMaker.scala
danbills: Something's off about these WdlDraft2 - specific instances being bound to shared  types `Scope`, `If`, etc.  It wouldn't make sense to have multiple implementations bound to that common type because that would cause >1 associated implicit and that is a no-no (it won't compile).

Anyway as it is you should try simulacrum and see if it gives you some variance in types here.

I doubt it, because inheritance is not really a thing in the typeclass world AFAIK.  This maybe the best solution for now, especially since it works!

cjllanwarne: Yeah, the packages are a bit arbitrary right now, this stuff could equally well (maybe *should* even) be in `draft2`. I'm intending the targets of these conversion implementations be draft-2 classes (backed up by shared code in mixins in `shared`. Look out for tidy-up part 2.

wdl/transforms/draft2/src/main/scala/wdl/transforms/draft2/wdlom2wom/WdlDraft2WomCallNodeMaker.scala
```diff
@@ -0,0 +1,141 @@
...
+  override def toWomCallNode(taskCall: WdlCall, localLookup: Map[String, GraphNodePort.OutputPort], outerLookup: Map[String, GraphNodePort.OutputPort], preserveIndexForOuterLookups: Boolean): ErrorOr[CallNode.CallNodeAndNewNodes] = {
+    buildWomNodeAndInputs(taskCall, localLookup, outerLookup, preserveIndexForOuterLookups)
+  }
+
+  private def buildWomNodeAndInputs(wdlCall: WdlCall, localLookup: Map[String, GraphNodePort.OutputPort], outerLookup: Map[String, GraphNodePort.OutputPort], preserveIndexForOuterLookups: Boolean): ErrorOr[CallNodeAndNewNodes] = {
+    import common.validation.ErrorOr._
+
+    val callNodeBuilder = new CallNode.CallNodeBuilder()
+
+    def allInputsWereWantedValidation(callable: Callable): ErrorOr[Unit] = {
```
jbuabbud: ToL this name is a bit confusing

cjllanwarne:
@danbills (or anyone else who wants to have a look), when you check out this branch, this big changes are: 

* In `build.sbt`
* In the WOM typeclasses (`wom.transforms._`)
* In the WDL to wom conversions (eg `wdl.transforms.draft2.wdlom2wom._`, extracted from, eg,  wdlModelDraft2's `wdl.WdlCallable` class)

wom/src/main/scala/wom/transforms/WomCallNodeMaker.scala
```diff
@@ -0,0 +1,13 @@
...
+
+import common.validation.ErrorOr.ErrorOr
+import wom.graph.CallNode.CallNodeAndNewNodes
+import wom.graph.GraphNodePort
+import simulacrum._
+import scala.language.implicitConversions
+
+@typeclass
+trait WomCallNodeMaker[A] {
+  @op("toWomCallNode")
```
danbills: These `@op` annotations are used to alias these methods (looks like you are using them as a whitelist?)

I.e. if you omit this you will get the exact same output.

build.sbt
```diff
@@ -10,10 +10,30 @@ lazy val wom = project
...
 
-lazy val wdl = project
-  .withLibrarySettings("cromwell-wdl", wdlDependencies, crossCompile = true)
+lazy val wdlRoot = Path("wdl")
+
+lazy val wdlModelRoot = wdlRoot / "model"
+
+lazy val wdlSharedModel = (project in wdlModelRoot / "shared")
+  .withLibrarySettings("cromwell-wdl-model-core", wdlDependencies, crossCompile = true)
   .dependsOn(wom)
```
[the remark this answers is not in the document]
cjllanwarne: Yes, you're right, I was hoping to unset that dependency.

Right now WDL is built up with WOM types and values (well technically WOM stole WDL's types and values, but we won't get into that...)

Ultimately we may want to split WOM into two importables, something like `wom.graph` and `wom.types` in order to keep us honest with these transforms but for now I decided to leave that for a future PR

wom/src/main/scala/wom/transforms/WomCallNodeMaker.scala
cjllanwarne: ok, I thought that was required to choose the functions you wanted to expose (as opposed to other helper functions I guess), probably just an artifact of copying an example which used them to alias the name.

build.sbt
```diff
@@ -194,7 +217,10 @@ lazy val root = (project in file("."))
   .aggregate(sfsBackend)
   .aggregate(sparkBackend)
   .aggregate(tesBackend)
-  .aggregate(wdl)
+  .dependsOn(wdlSharedModel)
+  .dependsOn(wdlSharedTransforms)
+  .dependsOn(wdlModelDraft2)
+  .dependsOn(wdlTransformsDraft2)
```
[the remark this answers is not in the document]
cjllanwarne: Oops, I think they're actually supposed to be `aggregate`s

cjllanwarne:
My pullapprove.yaml change seems to have upset Dan's thumb, but I'll grandfather it in.

PR #3250, opened 2018-02-07
Set CWL repo git hash to stabilize conformance tests

danbills:
This is to keep the rug from moving out from under us as far as CWL conformance tests is concerned.


geoffjentry:
@mcovarr for now, yes. that was already suggested a few weeks ago :P

mcovarr:
OK, I was among those doing the suggesting. It was the "signing off" bit that I wasn't sure had happened. 😄

danbills:
> for now, yes.

I'm going to interpret that as a 👍  and merge

PR #3252, opened 2018-02-08
Multiple input Feature requirement

danbills:
Step Inputs have the task of tying wf inputs to their `run` command.  This is done via the `source` field of a `WorkflowStepInput`, and there may be multiple, hence the name of this requirement.

This logic falls into two buckets:

# Type

Determing the type of the element being passed to the input is a function of:
 * what type the `run` is expecting
 * whether the variable appears in a scatter 
 * whether there is an expression in the `valueFrom` field
 * How to "merge" when multiple sources are declared.
 * the "default" (On which I am punting here)

# Expression 

The way the expression presents the inputs to the run command is determined by that `LinkMergeMethod` field i.e. how the run step expects the value to appear. 

# Assumptions

The Spec says to use `merge_nested` as default, but also says:

>  If "merge_nested" is specified with a single link, the value from the link must be wrapped in a single-item list.

Thus it seemed that all values should be wrapped in a list.  this is wrong and I'm going to ask on gitter what the deal is.

# What I've punted on

the `default` field.


cwl/src/main/scala/cwl/WorkflowStepInputExpression.scala
```diff
@@ -23,18 +24,59 @@ final case class WorkflowStepInputExpression(input: WorkflowStepInput,
...
+      case (None, Some(sources), LinkMergeMethod.MergeFlattened) =>
+
+        val validatedSourceValues: Checked[List[WomValue]] = sources.
+          traverse[ErrorOr, WomValue](s => lookupValue(s)).
+          toEither
+
+        //This is the meat of "merge_flattened," where we find arrays and concatenate them to form one array
+        val flattenedValidatedSourceValues: Checked[List[WomValue]] = validatedSourceValues.map(list => list.flatMap{ womValue =>
+          womValue match {
+            case WomArray(_, value) => value
```
danbills: I wonder if I'm missing an optional case here...

cwl/src/main/scala/cwl/WorkflowStepInputExpression.scala
```diff
@@ -23,18 +24,64 @@ final case class WorkflowStepInputExpression(input: WorkflowStepInput,
...
 
-    (input.valueFrom, input.source.map(_.fold(WorkflowStepInputSourceToStrings))) match {
-      case (None, Some(List(id))) =>
-        lookupValue(id).toValidated
+    (input.valueFrom, input.source.map(_.fold(StringOrStringArrayToStringList)), input.effectiveLinkMerge) match {
+        //When we have a single source, simply look it up
+      case (None, Some(List(source)), LinkMergeMethod.MergeNested) =>
+        lookupValue(source)
+
+      //When we have a single source, validate they are all present and provide them as a nested array of maps
```
Horneth: `When we have a single source`: looks like we have several in this case ?

wom/src/main/scala/wom/types/WomType.scala
```diff
@@ -98,4 +98,13 @@ object WomType {
       case t1 if types.forall(t2 => t1.isCoerceableFrom(t2)) => t1
     } getOrElse WomAnyType
   }
+
+  object RecursiveType {
+    def unapply(in: WomType): Option[WomType] = in match {
+      case WomOptionalType(other) => Some(other)
+      case WomMaybeEmptyArrayType(other) => Some(other)
+      case WomNonEmptyArrayType(other) => Some(other)
```
Horneth: Should `WomMapType` and `WomObjectTypeLike` be there too ?

cwl/src/main/scala/cwl/WorkflowStepInput.scala
```diff
@@ -0,0 +1,145 @@
...
+        //If sink parameter is an array and merge_flattened is used, must validate input & output types are equivalent before proceeding
+      case (false, array@Some(WomArrayType(tpe)), LinkMergeMethod.MergeFlattened)  =>
+        //Collect the upstream outputs and their types,
+        stepInput.validatedSourceTypes(outputTypeMap)  match {
+          case Right(map)  if typesToItemMatch(map.values, tpe) =>  array.get.asRight
+          case Right(map) => (s"could not verify that types $map and the items type of the run's InputArraySchema $tpe were compatible" |> NonEmptyList.one).asLeft
+          case Left(invalid) => Left(invalid)
+        }
+
+        //We don't have explicit information aboutthe type, so we
```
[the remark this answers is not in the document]
mcovarr: other than that, great comment.

cwl/src/main/scala/cwl/WorkflowStepInput.scala
```diff
@@ -0,0 +1,145 @@
...
+        array.asRight
+
+        //If sink parameter is an array and merge_flattened is used, must validate input & output types are equivalent before proceeding
+      case (false, array@Some(WomArrayType(tpe)), LinkMergeMethod.MergeFlattened)  =>
+        //Collect the upstream outputs and their types,
+        stepInput.validatedSourceTypes(outputTypeMap)  match {
+          case Right(map)  if typesToItemMatch(map.values, tpe) =>  array.get.asRight
+          case Right(map) => (s"could not verify that types $map and the items type of the run's InputArraySchema $tpe were compatible" |> NonEmptyList.one).asLeft
+          case Left(invalid) => Left(invalid)
+        }
```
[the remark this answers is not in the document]
danbills: Yes, thanks!  I knew I was doing something dumb.

wom/src/main/scala/wom/types/WomType.scala
danbills: Mm I don't think so.  Those are more specific, i.e. WomMapType has two types and we want both of those, and I think a WomObject is not recursive, i.e. we want that whole object and not some piece of it.

cwl/src/main/scala/cwl/WorkflowStepInput.scala
```diff
@@ -0,0 +1,146 @@
...
+    (isScattered, expectedTypeAsWom, stepInput.effectiveLinkMerge) match {
+
+      //If scattering over this variable, we expect an array of the sink type
+      case (true, Some(tpe), _) => WomArrayType(tpe).asRight
+
+      //If sink parameter is an array, we must frame the input as an array
+      case (false, Some(array: WomArrayType), LinkMergeMethod.MergeNested) =>
+        array.asRight
+
+      //If sink parameter is an array and merge_flattened is used, must validate input & output types are equivalent before proceeding
```
mcovarr: why is there compatibility checking in this case but not in other cases?

cwl/src/main/scala/cwl/WorkflowStepInputExpression.scala
```diff
@@ -23,18 +24,65 @@ final case class WorkflowStepInputExpression(input: WorkflowStepInput,
...
-    (input.valueFrom, input.source.map(_.fold(WorkflowStepInputSourceToStrings))) match {
-      case (None, Some(List(id))) =>
-        lookupValue(id).toValidated
+    (input.valueFrom, input.source.map(_.fold(StringOrStringArrayToStringList)), input.effectiveLinkMerge) match {
+
+      //When we have a single source, simply look it up
+      case (None, Some(List(source)), LinkMergeMethod.MergeNested) =>
+        lookupValue(source)
+
+      //When we have several sources, validate they are all present and provide them as a nested array of maps
```
Horneth: Can you explain briefly why in this case we create `WomMap`s ?

cwl/src/main/scala/cwl/WorkflowStepInputExpression.scala
```diff
@@ -23,18 +24,65 @@ final case class WorkflowStepInputExpression(input: WorkflowStepInput,
...
+      case (None, Some(sources), LinkMergeMethod.MergeNested) =>
+
+        def convertTupleToWomMap: (String, WomValue) => WomMap = {
+          case (key, value) => WomMap(WomMapType(WomStringType, value.womType), Map(WomString(key) -> value))
+        }
```
mcovarr: A function of no arguments that returns a two arg function that is only used after being `.tupled` into a function of one argument at the call site?

cwl/src/main/scala/cwl/WorkflowStepInputExpression.scala
```diff
@@ -81,3 +130,5 @@ final case class WorkflowStepInputExpression(input: WorkflowStepInput,
     case WorkflowStepInputSource.StringArray(sa) => sa.map(FullyQualifiedName(_).id).toSet
   }}
 }
+
+
```
mcovarr: -\n

cwl/src/main/scala/cwl/WorkflowStepInputExpression.scala
```diff
@@ -23,18 +24,65 @@ final case class WorkflowStepInputExpression(input: WorkflowStepInput,
...
+        def flatten: WomValue => List[WomValue] = {
+          womValue =>
+            womValue match {
+              case WomArray(_, value) => value.toList
+              case WomOptionalValue(_, Some(value)) => flatten(value)
+              case other => List(other)
+            }
+        }
+        //This is the meat of "merge_flattened," where we find arrays and concatenate them to form one array
```
mcovarr: mmm merged flattened meat

cwl/src/main/scala/cwl/WorkflowStepInputExpression.scala
```diff
@@ -23,18 +24,65 @@ final case class WorkflowStepInputExpression(input: WorkflowStepInput,
...
+                  // convert each tuple into a map consisting of a single element
+                  map(convertTupleToWomMap.tupled)
+            }
+
+        validatedSourceValues.map(WomArray.apply)
+
+      case (None, Some(sources), LinkMergeMethod.MergeFlattened) =>
+
+        val validatedSourceValues: Checked[List[WomValue]] = sources.
+          traverse[ErrorOr, WomValue](s => lookupValue(s)).
```
mcovarr: can this be just `lookupValue`?

cwl/src/main/scala/cwl/WorkflowStepInputExpression.scala
```diff
@@ -23,18 +24,65 @@ final case class WorkflowStepInputExpression(input: WorkflowStepInput,
...
+        def flatten: WomValue => List[WomValue] = {
+          womValue =>
+            womValue match {
+              case WomArray(_, value) => value.toList
+              case WomOptionalValue(_, Some(value)) => flatten(value)
+              case other => List(other)
+            }
+        }
+        //This is the meat of "merge_flattened," where we find arrays and concatenate them to form one array
+        val flattenedValidatedSourceValues: Checked[List[WomValue]] = validatedSourceValues.map(list => list.flatMap{ womValue =>
```
mcovarr: `_.flatMap`

cwl/src/main/scala/cwl/WorkflowStepInputExpression.scala
danbills: Hm.  I recall thinking that this is getting around the sticky situation if they're different types, but this is indeed wrong!  I'm looking at a conformance test \#25 that is expecting just the value, not a map.  Thanks for catching this!

cwl/src/test/scala/cwl/WorkflowStepInputSpec.scala
```diff
@@ -0,0 +1,55 @@
...
+import cwl.command.ParentName
+import org.scalacheck.Properties
+import org.scalacheck.Prop._
+import org.scalacheck.Gen._
+import shapeless.Coproduct
+import wom.types._
+import wom.values.{WomArray, WomInteger, WomMap, WomString, WomValue}
+import cats.syntax.validated._
+
+object WorkflowStepInputSpec extends Properties("WorkflowStepInput") {
```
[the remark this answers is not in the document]
danbills: they can be run via SBT `test:run` though for future reference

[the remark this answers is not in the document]
Horneth: I actually lied (partially) and found a way to run them through intellij, just not through a "run me all CWL scalatest"
Feel free to leave as is if it's a burden to change

cwl/src/main/scala/cwl/WorkflowStepInput.scala
danbills: In this case, it's in the spec in the merge nested section:

>The source and sink parameters must be compatible types

I think there is a gap in the logic of this requirement which is what if there is an expression in `valueFrom`?  I believe it's not possible to predict what type will come out of the expression,  and in any case compatibility between sources going into the expression and the ultimate input type are irrelevant.   In any case, the spec does not address this issue.

I guess in general I think we'd be fine to assert compatibility across all cases where `valueFrom` == `None`, but it's kinda weird that the spec doesn't say that in general.  (I also remember some catchall clause where if the value produced doesn't match the expected type, then try to proceed anyways).

cwl/src/main/scala/cwl/WorkflowStepInputExpression.scala
```diff
@@ -47,7 +82,8 @@ final case class WorkflowStepInputExpression(input: WorkflowStepInput,
        * input parameters is undefined and the result of evaluating valueFrom on a parameter must not be visible to
        * evaluation of valueFrom on other parameters.
        */
-      case (Some(StringOrExpression.Expression(expression)), Some(sources)) =>
+      case (Some(StringOrExpression.Expression(expression)), Some(sources), _) =>
```
mcovarr: What about `Some(StringOrExpression.String(...)), Some(sources))`? Thibault actually has CWL like that in the comment of his open PR. 🙂

danbills: Well I should put this case in, but it looks like I will only be using the string value and ignoring those sources:

>If valueFrom is a constant string value, use this as the value for this input parameter.

danbills: So it seems like we should throw an error in this situation.  If someone specifies sources and has a string value in their expression, it's almost certainly a bug in their semantics (i.e. expecting to use the sources).

danbills: I will write an informative error message in this case

PR #3260, opened 2018-02-09
Abort sub workflows

Horneth:


centaur/src/main/scala/centaur/test/Test.scala
```diff
@@ -220,42 +220,7 @@ object Operations {
...
-      f match {
-        case Failure(_) if OffsetDateTime.now().isBefore(startTime.plusSeconds(timeout.toSeconds)) =>
-          blocking { Thread.sleep(5.seconds.toMillis) }
-          eventually(startTime, timeout)(f)
-        case t => t
-      }
-    }
-    
     new Test[Unit] {
-      def checkVMTerminated(): Unit = {
```
Horneth: Turns out this was never really testing the right thing. The jes instance name is not always going to be there if the VM didn't have time to start. It's already checking that JES has aborted the job which is all we can do anyway.

Horneth:
@mcovarr I know :/ I'll try to come up with something over the weekend

centaur/src/main/scala/centaur/test/Test.scala
```diff
@@ -286,18 +247,33 @@ object Operations {
     // For JES it should be fine but locally it can be quite fast
     def pollDelay() = blocking { Thread.sleep(5000) }
 
+    // Special case for sub workflow testing
+    def findJobIdInSubWorkflow(subWorkflowId: String): Option[String] = {
+      for {
+        metadata <- CentaurCromwellClient.metadata(WorkflowId.fromString(subWorkflowId)).toOption
+        jobId <- metadata.value.get("calls.inner_abort.aborted.jobId")
+      } yield jobId.asInstanceOf[JsString].value
```
danbills: ToL  wanted to point you at shapeless support for casting: https://github.com/milessabin/shapeless/wiki/Feature-overview:-shapeless-2.0.0#type-safe-cast

AFAIK it just wraps the cast and returns an `Option` so it's marginally safer.

Horneth: Oh yep that's better - because this is pre-existing I'll leave it like this for now but will use my bug rotation time to make a second pass later

PR #3263, opened 2018-02-12
Don't log cwl preprocessing in womtool

Horneth:
This allows to do write directly the output of `womtool womgraph` into a valid `.dot` file without the extra logging from cwl pre-processing

### DIFF
--- a/cwl/src/main/scala/cwl/CwlDecoder.scala
+++ b/cwl/src/main/scala/cwl/CwlDecoder.scala
@@ -36,7 +36,7 @@ object CwlDecoder {
     fromEither[IO](cwlToolResult flatMap resultToEither)
   }
 
-  lazy val cwlPreProcessor = new CwlPreProcessor()
+  private lazy val cwlPreProcessor = new CwlPreProcessor()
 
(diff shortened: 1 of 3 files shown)


cwl/src/main/scala/cwl/CwlDecoder.scala
```diff
@@ -54,9 +54,9 @@ object CwlDecoder {
     * Notice it gives you one instance of Cwl.  This has transformed all embedded files into scala object state
     */
   def decodeCwlFile(fileName: BFile,
-                    workflowRoot: Option[String] = None): Parse[Cwl] =
+                    workflowRoot: Option[String] = None)(implicit processor: CwlPreProcessor = cwlPreProcessor): Parse[Cwl] =
```
danbills: ToL I'm not crazy about implicitly importing non-typeclassy stuff  but I think the alternative, as @mcovarr alluded to in a previous PR, is the `Reader` Monad, which would make `Parse` more complicated.

I think that's ultimately the way to go, but until then I think we do it this way until we have time to plow through uncharted territory.

PR #3265, opened 2018-02-12
Initial draft 3 objects and some new (ignored) tests

cjllanwarne:


wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/ast2wdlom/Draft3FileElementFromAst.scala
```diff
@@ -0,0 +1,27 @@
...
+import wdl.draft3.parser.WdlParser.Ast
+import wdl.draft3.transforms.ast2wdlom.EnhancedDraft3Ast._
+
+object Draft3FileElementFromAst extends FromAst[FileElement] {
+  override def convert(ast: Ast): ErrorOr[FileElement] = if (ast.getName == "Draft3File") {
+
+    val validatedImportElements: ErrorOr[Vector[ImportElement]] = ast.getAttributeAsVector[ImportElement]("imports").toValidated
+
+    val validatedWorkflowAndTaskElements: ErrorOr[ (Vector[WorkflowDefinitionElement], Vector[TaskDefinitionElement]) ] =
+      ast.getAttributeAsVectors[WorkflowDefinitionElement, TaskDefinitionElement]("body", "Workflow", "Task")
```
jbuabbud: Did you mean to call `toValidated` on this?

cjllanwarne: So on imports, that's a result of `getAttributeAsVector` returning a `Checked[Vector[ImportElement]]` when we needed a `ErrorOr[Vector[ImportElement]]`.

They're sort of interchangeable and `getAttributeAsVector` could just return the ErrorOr itself, but since it doesn't we switch between the types or error-holding object here instead

PR #3273, opened 2018-02-13
Added workflow output case classes

jbuabbud:
Added case classes for workflow output elements.


wdl/model/draft3/src/main/scala/wdl/model/draft3/elements/WorkflowOutputDeclarationElement.scala
```diff
@@ -0,0 +1,16 @@
...
+
+case class WorkflowOutputDeclarationElement(name: String,
+                                            valueType: String,
+                                            value: String
+                                ) extends LanguageElement {
+  override def children: Seq[String] = Seq(name, valueType, value)
+}
+
+@typeclass
+trait WorkflowOutputDeclarationElementMaker[A] {
```
cjllanwarne: This is why the build has failed: to add this `@typeclass` annotation you need to add this line to the top of the file too (because it's doing implicit conversions behind the scenes and scala makes you declare that in your imports because they're a so-called 'advanced language feature'):
```
import scala.language.ImplicitConversion
```

But... I would actually just remove the trait altogether. I've been thinking of deleting them from the existing classes too since I've been moving towards the more general purpose (and more composable) `FromAtoB[A, B]` trait for the conversions 

Eg `object Draft3FileElementFromAst extends FromAst[FileElement]` where `FromAst[FileElement]` extends `FromAtoB[Ast, FileElement]`).

wdl/model/draft3/src/main/scala/wdl/model/draft3/elements/WorkflowOutputsElement.scala
```diff
@@ -0,0 +1,14 @@
...
+import common.validation.ErrorOr.ErrorOr
+import simulacrum.typeclass
+
+class WorkflowOutputsElement(outputs: Seq[WorkflowOutputsElement]
+                                       ) extends LanguageElement {
+  override def children: Seq[LanguageElement] = outputs
+}
+
+@typeclass
+trait WorkflowOutputsElementMaker[A] {
```
cjllanwarne: Same comment here, I'd remove this trait

wdl/model/draft3/src/main/scala/wdl/model/draft3/elements/WorkflowOutputDeclarationElement.scala
```diff
@@ -0,0 +1,16 @@
+package wdl.model.draft3.elements
+
+import common.validation.ErrorOr.ErrorOr
+import simulacrum.typeclass
+
+case class WorkflowOutputDeclarationElement(name: String,
+                                            valueType: String,
+                                            value: String
+                                ) extends LanguageElement {
+  override def children: Seq[String] = Seq(name, valueType, value)
```
cjllanwarne: You're trying to override the `LanguageElement.children` field which is a `Seq[LanguageElement]` with a `Seq[String]` so this is an error -

I think in this case the declaration has no child `LanguageElements` (since the fields are all terminals) so I think this is just `Seq.empty`

wdl/model/draft3/src/main/scala/wdl/model/draft3/elements/WorkflowOutputsElement.scala
```diff
@@ -0,0 +1,14 @@
...
+  def convert(a: A): ErrorOr[WorkflowOutputsElement]
+}
```
cjllanwarne: Add a newline

wdl/model/draft3/src/main/scala/wdl/model/draft3/elements/WorkflowOutputDeclarationElement.scala
jbuabbud: Should I go ahead and remove the typeclass trait from WorkflowDefinitionElement as well?

wdl/model/draft3/src/main/scala/wdl/model/draft3/elements/WorkflowDefinitionElement.scala
```diff
@@ -1,17 +1,7 @@
...
                                     ) extends LanguageElement {
-  override def children: Seq[LanguageElement] = List.empty // TODO: sections
-}
-
-@typeclass
-trait WorkflowDefinitionElementMaker[A] {
-  def convert(a: A): ErrorOr[WorkflowDefinitionElement]
-}
+  override def children: Seq[LanguageElement] = workflowOutputs
+}
```
cjllanwarne: \n

wdl/model/draft3/src/main/scala/wdl/model/draft3/elements/WorkflowOutputDeclarationElement.scala
```diff
@@ -0,0 +1,5 @@
+package wdl.model.draft3.elements
+
+final case class WorkflowOutputDeclarationElement(valueType: String, name: String, value: String) extends LanguageElement {
+  override def children: Seq[LanguageElement] = Seq.empty
```
cjllanwarne: Leave them for now, but as a heads up I'm going to remove this "children" field soon. 

It's extra typing for us and I'm not convinced we need it. We certainly don't need it right now!

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/ast2wdlom/EnhancedDraft3Ast.scala
```diff
@@ -44,7 +44,7 @@ object EnhancedDraft3Ast {
     }
 
     def getAttributeAsVectors[A, B](attr: String, astName1: String, astName2: String)
-                                   (implicit astNodeToA: CheckedAtoB[AstNode, A], astNodeToB:CheckedAtoB[AstNode, B]
+                                   (implicit astNodeToA: CheckedAtoB[AstNode, A], astNodeToB: CheckedAtoB[AstNode, B]
```
cjllanwarne: I was thinking about this function. If you feel up to it, I'd suggest you make this change - but maybe as a subsequent PR if you want to merge this one:

 - We should have a super type for everything we might want to be in the "body" attribute (eg `WorkflowBodyElement`)
 - Every case class for a body element extends it, eg `case class WorkflowOutputsSection(...) extends WorkflowBodyElement`
 - We can have a top level `CheckedAtoB[Ast, WorkflowBodyElement]` which delegates to (eg) `CheckedAtoB[Ast, WorkflowOutputsElement]` based on the name of the Ast it receives.
- Then in the `CheckedAtoB[Ast, WorkflowDefinitionElement]` function we'll end up with a set of `WorkflowBodyElements`. We'll need to do the filters there to get appropriate vectors for the `WorkflowDefinitionElement` fields.

jbuabbud: Sounds good to me. I think that I will wait and add that as part of my next PR, since I've already begun working on some other changes. Do you think a super type of CallableBodyElement would be more useful?

cjllanwarne: Probably best to keep workflow and task body elements as separate supertype `trait`s - the set of things that can go in a workflow is different from the set of things that can go in a task.

Although I imagine `case class OutputsSection` will have to both, ie `extends WorkflowBodyElement with TaskBodyElement`

jbuabbud: Traits are pretty cool!

PR #3275, opened 2018-02-13
Send next workflow heartbeat after it's been processed

Horneth:


engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -147,8 +149,11 @@ case class WorkflowExecutionActor(params: WorkflowExecutionActorParams)
   // Most of the Event handling is common to all states, so put it here. Specific behavior is added / overridden in each state
   whenUnhandled {
     case Event(ExecutionHeartBeat, data) if data.executionStore.needsUpdate =>
-      stay() using startRunnableNodes(data)
+      val newData =  startRunnableNodes(data)
```
mcovarr: TOL spacing slightly off

PR #3276, opened 2018-02-13
Added the ability to query cromwell by label-or.

kshakir:
Adds support for https://github.com/broadinstitute/CromIAM/pull/27


database/sql/src/main/scala/cromwell/database/slick/tables/WorkflowMetadataSummaryEntryComponent.scala
```diff
@@ -86,12 +87,24 @@ trait WorkflowMetadataSummaryEntryComponent {
         reduceLeftOption(_ || _)
       // Workflows can be queried by multiple labels, the reduceLeftOption ANDs together all workflows are associated
       // to all the labels from the query.
-      val labelsFilter = labelKeyLabelValues.map { case (labelKey, labelValue) =>
-          existsWorkflowIdLabelKeyAndValue(workflowMetadataSummaryEntry.workflowExecutionUuid, labelKey, labelValue) }.
+      val labelsAndFilter = labelAndKeyLabelValues.map { case (labelKey, labelValue) =>
+        existsWorkflowIdLabelKeyAndValue(workflowMetadataSummaryEntry.workflowExecutionUuid, labelKey, labelValue) }.
         reduceLeftOption(_ && _)
+      // Workflows can be queried by multiple labels, the reduceLeftOption ANDs together all workflows are associated
```
Horneth: This comment seems to need an update for the OR version

mcovarr: 🍝

mcovarr: most of it might be pulled out a function too (DRYed copypasta)

services/src/main/scala/cromwell/services/metadata/WorkflowQueryKey.scala
```diff
@@ -50,23 +61,30 @@ object WorkflowQueryKey {
...
         }
       }
       val nels: List[ErrorOr[Label]] = values map validateLabelRegex
       sequenceListOfValidatedNels("Label values do not match allowed pattern label-key:label-value", nels)
     }
   }
 
+
+  case object LabelAndKeyValue extends LabelLikeKeyValue {
+    override val name = "Label"
```
mcovarr: `Labeland` for symmetry or would that be breaking?

services/src/main/scala/cromwell/services/metadata/WorkflowQueryKey.scala
```diff
@@ -116,7 +134,7 @@ sealed trait WorkflowQueryKey[T] {
 sealed trait DateTimeWorkflowQueryKey extends WorkflowQueryKey[Option[OffsetDateTime]] {
   override def validate(grouped: Map[String, Seq[(String, String)]]): ErrorOr[Option[OffsetDateTime]] = {
     valuesFromMap(grouped) match {
-      case vs if vs.size > 1 =>
+      case vs if vs.lengthCompare(1) > 0 =>
```
mcovarr: why?

services/src/main/scala/cromwell/services/metadata/WorkflowQueryKey.scala
kshakir: `Labeland` instead of `Label` would be breaking unfortunately. In theory we could support both... or have a new versioned endpoint... but 🏈

services/src/main/scala/cromwell/services/metadata/WorkflowQueryKey.scala
kshakir: The primary community argument is that in general `size` shouldn't be used to compare sequence lengths. That's what `lengthCompare` is specifically designed for:
- https://pavelfatin.com/scala-collections-tips-and-tricks/#sequences-length

However, it is also acknowledged that `lengthCompare` while being faster is harder to read:
- github.com/scala/collection-strawman/issues/338

PR #3279, opened 2018-02-14
standard output and error file names can be expressions.

mcovarr:


common/src/main/scala/common/util/StringUtil.scala
```diff
@@ -0,0 +1,15 @@
...
+import org.apache.commons.lang3.StringUtils
+
+object StringUtil {
+  implicit class EnhancedString(val string: String) extends AnyVal {
+
+    def ensureSlashed: String = StringUtils.appendIfMissing(string, "/")
+
+    def ensureUnslashed: String = string.ensureSlashed |> StringUtils.chop
+
+    def removePrefix(prefix: String): String = StringUtils.removeStart(string, prefix)
```
Horneth: Is this different from http://www.scala-lang.org/api/2.12.4/scala/collection/immutable/StringOps.html#stripPrefix(prefix:String):String ?

mcovarr: My version *might* be null safe depending on how implicit classes are implemented, but the way we're using this code we'd have some pretty big problems if that argument was null. I'll remove this and use the library version.

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -236,9 +236,52 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
...
+    // references them outside a (cd "execution dir"; ...) subshell. The default names are known to be relative paths,
+    // the names from redirections may or may not be relative.
+    def absolutizeContainerPath(path: String): String = {
+      if (path.startsWith(cwd.pathAsString)) path else cwd.resolve(path).pathAsString
+    }
+
+    val List(executionStdin, executionStdout, executionStderr) = List[InstantiatedCommand => Option[String]](
+      _.evaluatedStdinRedirection,
+      _.evaluatedStdoutOverride.orElse(Option(jobPaths.defaultStdoutFilename)),
+      _.evaluatedStderrOverride.orElse(Option(jobPaths.defaultStderrFilename))) map { maybeRedirection =>
```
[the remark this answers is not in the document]
mcovarr: Actually this whole contrivance is probably more trouble than it's worth once I do that, so I'll just use simple assignments. 🙂

PR #3281, opened 2018-02-15
Wom Coproduct Type

danbills:
Introduce a new type to support possibility of various types.


danbills:
adding unit tests for WomCoproductType coercion

wom/src/test/scala/wom/types/WomCoproductSpec.scala
```diff
@@ -0,0 +1,16 @@
...
+class WomCoproductSpec extends FlatSpec with Matchers {
+
+  behavior of "Wom Coproduct Type"
+
+  it should "pick the exact type to coerce to" in {
+    val wct = WomCoproductType(List(WomStringType, WomBooleanType))
+
+    wct.coercion(WomBoolean(true)) shouldBe WomCoproductValue(wct, WomBoolean(true))
+  }
+
```
mcovarr: How about some testing of the squishy matches?

wom/src/main/scala/wom/types/WomCoproductType.scala
```diff
@@ -0,0 +1,52 @@
...
+    case wct@WomCoproductValue(tpe, _) if (tpe.equalsType(this).isSuccess) => wct
+
+    //If we can find this type exactly in our coproduct, use that type for the coercion
+    case womValue: WomValue if (types.contains(womValue.womType)) =>
+      val v: Try[WomValue] = womValue.womType.coerceRawValue(womValue)
+      v.get |> (WomCoproductValue(this, _))
+
+    //If we don't have any information, try to coerce this value one by one, stopping at the first successful try
+    case any =>
+      val triedToCoerce: Try[WomValue] = types.map(_.coerceRawValue(any)).reduce(_ orElse _)
```
Horneth: ToL: I have a vague memory that `reduce` throws on an empty list is that correct ? I don't know if it's possible for a `WomCoproductType` to have an empty list of types (definitely wouldn't make sense), but since nothing strictly prevents it just wanted to mention it.

cwl/src/main/scala/cwl/MyriadOutputTypeToWomType.scala
```diff
@@ -42,10 +42,15 @@ object MyriadOutputTypeToWomType extends Poly1{
...
+      case (Array(), array: Array[MyriadOutputInnerType]) if array.size > 1  =>
+        val types = array.map(_.fold(MyriadOutputInnerTypeToWomType))
+        WomCoproductType(types.toList)
       // If there's a null type and a single non null type, it's a WomOptionalType
       case (Array(_), Array(singleNonNullType)) =>
         WomOptionalType(singleNonNullType.fold(MyriadOutputInnerTypeToWomType))
-      // Leave other "Coproduct types" unsupported for now
+      case (Array(_), array: Array[MyriadOutputInnerType]) if array.size > 1  =>
+        val types = array.map(_.fold(MyriadOutputInnerTypeToWomType))
+        WomOptionalType(WomCoproductType(types.toList))
```
Horneth: ToL: Code coverage tells me this is not being tested. If you've got the bandwidth It seems it would be a quick win to add a couple tests for these new `case`s and in `MyriadInputTypeToWomType` too.

wom/src/main/scala/wom/types/WomCoproductType.scala
danbills: :+1:.  it does not make sense as this type only created when number of types > 2 but it is worthwhile to encode this as a type, hence I'm going to use `NonEmptyList` to represent `types`

PR #3285, opened 2018-02-16
WDLOM expression elements

cjllanwarne:
Mostly just a set of `Element` case classes.

A proof of concept `AstNodeToExpressionElement.convert` which only works for primitive literals and the `Add` function

### DIFF
--- a/wdl/model/draft3/src/main/scala/wdl/model/draft3/elements/ExpressionElement.scala
+++ b/wdl/model/draft3/src/main/scala/wdl/model/draft3/elements/ExpressionElement.scala
@@ -60,6 +60,41 @@ object ExpressionElement {
   case object StderrCall extends FunctionCall
   // TODO: and the rest...
 
-
+  /**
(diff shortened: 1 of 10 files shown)


wdl/model/draft3/src/main/scala/wdl/model/draft3/elements/ExpressionElement.scala
```diff
@@ -0,0 +1,112 @@
...
+  case class IdentifierMemberAccess(firstIdentifier: String, secondIdentifierOrFirstMemberAccess: String, memberAccessTail: Vector[String]) extends ExpressionElement
+
+  /**
+    * A member access which is based on an expression rather than an identifier.
+    *
+    * eg:
+    * (1, 2).left
+    */
+  case class ExpressionMemberAccess(expression: ExpressionElement, memberAccessTail: Vector[String]) extends ExpressionElement
+}
```
jbuabbud: newline

PR #3294, opened 2018-02-20
Ability for services to communicate with each other + a few fixes here and there

Horneth:
Turns out the batch size for write metadata was ignored (we were looking at the wrong config path)
Also turns out it doesn't matter that much when load is very high.
That is mostly because when we flush the event queue, we flush all of it, not just `dbBatchSize` of it.
This added to the fact that the flushes are controlled by messages we sent to `self` that can get lost in an ocean of metadata event messages, we can end up flushing queues of several million events at once, in one transaction, which can take quite some time for slick to do. The more time it takes, the more events we accumulate before the next flush, so the next flush takes even more time etc... 
I'll have a follow up PR to re-factor this a bit, with stats to be able to compare.

TODO:
- [x] Add some tests

### DIFF
--- a/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
@@ -427,7 +427,7 @@ case class WorkflowExecutionActor(params: WorkflowExecutionActorParams)
           else s"$tag ($shardCount shards)"
       })
     val mode = if (restarting) "Restarting" else "Starting"
-    if (runnableCalls.nonEmpty) workflowLogger.info(s"$mode " + runnableCalls.mkString(", "))
+    if (runnableCalls.nonEmpty) workflowLogger.info(s"$mode calls: " + runnableCalls.mkString(", "))
 
     val diffValidation = runnableKeys.traverse[ErrorOr, WorkflowExecutionDiff]({
       case key: BackendJobDescriptorKey => processRunnableJob(key, data)
--- a/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
@@ -416,11 +416,18 @@ case class WorkflowExecutionActor(params: WorkflowExecutionActorParams)
     import keys._
 
     val DataStoreUpdate(runnableKeys, updatedData) = data.executionStoreUpdate
-    val runnableCalls = runnableKeys.view collect { case k if k.node.isInstanceOf[CallNode] => k } sortBy { k =>
-      (k.node.fullyQualifiedName, k.index.getOrElse(-1)) } map { _.tag }
-
+    val runnableCalls = runnableKeys.view
+      .collect({ case k: BackendJobDescriptorKey => k })
+      .groupBy(_.node)
+      .map({
+        case (node, keys) => 
+          val tag = node.fullyQualifiedName
+          val shardCount = keys.map(_.index).distinct.size
+          if (shardCount == 1) tag
+          else s"$tag ($shardCount shards)"
+      })
     val mode = if (restarting) "Restarting" else "Starting"
-    if (runnableCalls.nonEmpty) workflowLogger.info(s"$mode calls: " + runnableCalls.mkString(", "))
+    if (runnableCalls.nonEmpty) workflowLogger.info(s"$mode " + runnableCalls.mkString(", "))
 
     val diffValidation = runnableKeys.traverse[ErrorOr, WorkflowExecutionDiff]({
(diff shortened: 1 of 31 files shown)


cjllanwarne:
ToL: 

The title made me a little nervous that this was adding more coupling that would make things hard in a world where these services are different implementations or instantiated on remote servers.

But since it looks like it's all being directed through the standard interfaces in the service registry it all looks OK. And I assume we can still give the `serviceRegistryActor: ActorRef` argument to a new service that's created remotely.

geoffjentry:
I had the same initial terror as @cjllanwarne but this makes a lot of sense

mcovarr:
This may have been PR click bait. I may have to try the same thing. 😛

Horneth:
@mcovarr it does seem to have worked 😄
@cjllanwarne @geoffjentry Good point, my main goal was to make it possible for the metadata service to report statsd metrics, which is done through another service. But like you said since it's just making available the service registry actor to the services I don't think it introduces any additional coupling.

services/src/main/scala/cromwell/services/metadata/impl/MetadataServiceActor.scala
```diff
@@ -49,9 +50,9 @@ final case class MetadataServiceActor(serviceConfig: Config, globalConfig: Confi
 
   val readActor = context.actorOf(ReadMetadataActor.props(), "read-metadata-actor")
 
-  val dbFlushRate = serviceConfig.as[Option[FiniteDuration]]("services.MetadataService.db-flush-rate").getOrElse(5 seconds)
-  val dbBatchSize = serviceConfig.as[Option[Int]]("services.MetadataService.db-batch-size").getOrElse(200)
-  val writeActor = context.actorOf(WriteMetadataActor.props(dbBatchSize, dbFlushRate), "WriteMetadataActor")
+  val dbFlushRate = serviceConfig.as[Option[FiniteDuration]]("db-flush-rate").getOrElse(5 seconds)
+  val dbBatchSize = serviceConfig.as[Option[Int]]("db-batch-size").getOrElse(200)
```
mcovarr: 😦

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -416,11 +416,18 @@ case class WorkflowExecutionActor(params: WorkflowExecutionActorParams)
...
+      .map({
+        case (node, keys) => 
+          val tag = node.fullyQualifiedName
+          val shardCount = keys.map(_.index).distinct.size
+          if (shardCount == 1) tag
+          else s"$tag ($shardCount shards)"
+      })
     val mode = if (restarting) "Restarting" else "Starting"
-    if (runnableCalls.nonEmpty) workflowLogger.info(s"$mode calls: " + runnableCalls.mkString(", "))
+    if (runnableCalls.nonEmpty) workflowLogger.info(s"$mode " + runnableCalls.mkString(", "))
```
mcovarr: Interesting that the old code was sorting and the new code is not and no tests seem to mind. I guess we got rid of that yucky integrationesque test from the early days?

services/src/main/scala/cromwell/services/metadata/impl/WriteMetadataActor.scala
```diff
@@ -87,7 +91,10 @@ class WriteMetadataActor(override val dbBatchSize: Int, override val dbFlushRate
 }
 
 object WriteMetadataActor {
-  def props(dbBatchSize: Int, flushRate: FiniteDuration): Props = Props(new WriteMetadataActor(dbBatchSize, flushRate)).withDispatcher(ServiceDispatcher)
+  val MetadataWritePath = MetadataServiceActor.MetadataInstrumentationPrefix.::("writes")
```
mcovarr: is there a more Englishy version of that method that might be more readable?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
Horneth: yeah I was expecting tests to fail to with this change but none seemed to care. I actually wanted to draw some attention on this, we were previously logging every single call that was started here, including every shard and its number. This just give the number of shards instead, also do not differentiate attempts.
Are we ok with losing some of this info in the logs ?

services/src/main/scala/cromwell/services/metadata/impl/WriteMetadataActor.scala
cjllanwarne: Two suggestions:

`.doubleColon`?
`.colonColon`?

Horneth: Not that I see, I could do `MetadataServiceActor.MetadataInstrumentationPrefix.++(NonEmptyList.one("writes"))` but that seems unnecessary.

Issue #3271, opened 2018-02-13
CRON test mutect2 is broken


kshakir:
Last [CRON errors](https://travis-ci.org/broadinstitute/cromwell/jobs/340487274):

```scala
java.lang.Exception: Task Mutect2.oncotate_m2:NA:1 failed. Job exit code 1. Check gs://cloud-cromwell-dev/cromwell_execution/travis/Mutect2/a5d2b104-647b-45fa-8149-7bddf6bc2094/call-oncotate_m2/oncotate_m2-stderr.log for more information. PAPI error code 5.  Message: 10: Failed to delocalize files: failed to copy the following files: "/mnt/local-disk/HCC1143.maf.annotated -> gs://cloud-cromwell-dev/cromwell_execution/travis/Mutect2/a5d2b104-647b-45fa-8149-7bddf6bc2094/call-oncotate_m2/HCC1143.maf.annotated (cp failed: gsutil -q -m cp -L /var/log/google-genomics/out.log /mnt/local-disk/HCC1143.maf.annotated gs://cloud-cromwell-dev/cromwell_execution/travis/Mutect2/a5d2b104-647b-45fa-8149-7bddf6bc2094/call-oncotate_m2/HCC1143.maf.annotated, command failed: CommandException: No URLs matched: /mnt/local-disk/HCC1143.maf.annotated\nCommandException: 1 file/object could not be transferred.\n)"
```

geoffjentry:
@kshakir this is fixed, right?

kshakir:
Fixed by https://github.com/broadinstitute/cromwell/pull/3313

PR #3295, opened 2018-02-20
Update CNV/M2 wdls & docker input

ruchim:
Successful CRON Build: https://travis-ci.org/broadinstitute/cromwell/builds/342402374


ruchim:
Also connected to issue #3271

PR #3298, opened 2018-02-20
Copy all workflow output files

Horneth:


geoffjentry:
Can you explain a bit in not-code what the problem was here?

Horneth:
The `findFiles` function in `CopyWorkflowOutputsActor.scala` that was supposed to extract all the files references in a `WomValue` was missing a few `case`s.
More generally there's a `collectAsSeq` method that can be used to traverse a `WomValue` (recursively if necessary) and collect objects based on a partial function. I used that extract `WomFiles`

geoffjentry:
@Horneth gotcha, thanks

PR #3300, opened 2018-02-20
Fix scala 2.11 build

jbuabbud:


wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/ast2wdlom/AstToOutputsSectionElement.scala
```diff
@@ -1,13 +1,17 @@
...
 import wdl.model.draft3.elements.{OutputDeclarationElement, OutputsSectionElement}
 import wdl.draft3.transforms.ast2wdlom.EnhancedDraft3Ast._
 import common.Checked
 
-
 object AstToOutputsSectionElement {
   def convert(a: WdlParser.Ast): Checked[OutputsSectionElement] = {
-      a.getAttributeAsVector[OutputDeclarationElement]("outputs") map OutputsSectionElement
+      a.getAttributeAsVector[OutputDeclarationElement]("outputs") map { declarations =>
+        OutputsSectionElement(declarations)
```
cjllanwarne: Is this line change still necessary when you add the import above?

jbuabbud: It is not! Proceeding with the import alone.

PR #3301, opened 2018-02-20
Breaking JSON Stringify for the greater good

danbills:
Does what the title says, but we have solutions(read: hacks) for JSON stringify that we do not have for Nashorn in general.

I don't have those hacks in hand as I'm working on the ramifications of it working as intended, but rest assured that PR is coming soon.


cjllanwarne:
The greater good

PR #3299, opened 2018-02-20
AST to Expression logic

cjllanwarne:
Logic to convert draft-3 AstNodes into wdlom expression elements

PR #3302, opened 2018-02-20
Split Declaration types and give them expressions

cjllanwarne:
* Includes changes in grammar from https://github.com/cjllanwarne/wdl/pull/3
* Makes InputDeclarations different from SetDeclarations (based on whether they have a setter or not)
* ~~Removes the distinction between OutputDeclarations and SetDeclarations~~
* EDIT: **Formalizes** the distinction between OutputDeclarations and SetDeclarations!
* Includes `SetDeclarations` as part of the workflow graph

- [x] Rebase on develop after #3299


wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/ast2wdlom/AstToDeclarationElement.scala
```diff
@@ -2,20 +2,20 @@ package wdl.draft3.transforms.ast2wdlom
...
 import wdl.draft3.parser.WdlParser
-import wdl.model.draft3.elements.{DeclarationElement, ExpressionElement, TypeElement}
+import wdl.model.draft3.elements.{SetDeclarationElement, ExpressionElement, TypeElement}
 import cats.syntax.apply._
 import cats.syntax.either._
 import wdl.draft3.transforms.ast2wdlom.EnhancedDraft3Ast._
 
-object AstToOutputDeclarationElement {
-  def convert(a: WdlParser.Ast): ErrorOr[DeclarationElement] = {
+object AstToDeclarationElement {
```
jbuabbud: ToL maybe this should be renamed as well for continuity? I'm sort of unclear on the reasoning for changing "Declaration" -> "SetDeclaration". I think InputDeclaration being its own separate entity makes sense because the expression parameter is optional, but a Declaration must by definition have an expression, so it's sort of implied that it's setting something in the parameters. "Set" seems superfluous.

cjllanwarne: Yeah, I'd go along with that. I'll make the change back

wdl/model/draft3/src/main/scala/wdl/model/draft3/elements/DeclarationElement.scala
```diff
@@ -1,6 +1,29 @@
...
-  * A Declaration which has a setter
+  * Content of an intermediate or output declaration:
   */
-final case class DeclarationElement(typeElement: TypeElement, name: String, expression: ExpressionElement) extends TaskBodyElement with WorkflowGraphNodeElement
+final case class DeclarationContent(typeElement: TypeElement, name: String, expression: ExpressionElement)
+
+/**
+  * A Declaration outside of an input or output block
+  */
+final case class IntermediateValueDeclarationElement(typeElement: TypeElement, name: String, expression: ExpressionElement) extends WorkflowGraphNodeElement with TaskBodyElement
```
jbuabbud: Is it possible for one of these to exist within a workflow body?

cjllanwarne:
Travis `centaur-local` is having troubles unrelated to this PR, I'm going to merge

PR #3304, opened 2018-02-20
FileBodyElement supertype and FileElement conversion

jbuabbud:


wdl/model/draft3/src/main/scala/wdl/model/draft3/elements/ImportElement.scala
```diff
@@ -1,4 +1,4 @@
 package wdl.model.draft3.elements
 
 case class ImportElement(importUrl: String,
-                         alias: Option[String]) extends LanguageElement
+                         alias: Option[String]) extends FileBodyElement
```
cjllanwarne: I'd revert this since `import`s are already separated from the `body` attribute

PR #3307, opened 2018-02-21
Declarations and outputs wired up as a linked graph

cjllanwarne:
See [declaration_chain.wdl](https://github.com/broadinstitute/cromwell/compare/cjl_first_wdlgraph?expand=1#diff-444c3d4a82de93c01c7cc237702c565f) for the motivating WDL for this PR.

Allows us to make the correct WOM graph from these Declaration elements, and tests that centaur is able to run the workflow.


wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wom/WdlomWomExpression.scala
```diff
@@ -0,0 +1,35 @@
...
+    case id: IdentifierLookup => inputMap((id: ExpressionElement).consumedValues.head).outputPort.womType.validNel
+    case _ => ??? // TODO other expression elements
+  }
+
+  override def evaluateFiles(inputTypes: Map[String, WomValue], ioFunctionSet: IoFunctionSet, coerceTo: WomType): ErrorOr[Set[WomFile]] = {
+    Set.empty[WomFile].validNel
+  }
+}
+
+final case class FoundConsumedValue(outputPort: OutputPort, inputName: String)
```
cjllanwarne: Looks like the port is only used to get the type, so maybe just store the type

wdl/model/draft3/src/main/scala/wdl/model/draft3/elements/DeclarationElement.scala
```diff
@@ -27,3 +27,11 @@ object OutputDeclarationElement {
...
   */
 final case class InputDeclarationElement(typeElement: TypeElement, name: String, expression: Option[ExpressionElement]) extends LanguageElement with WorkflowGraphNodeElement
+
+object DeclarationElement {
+  def unapply(languageElement: LanguageElement): Option[(TypeElement, String, Option[ExpressionElement])] = languageElement match {
+    case IntermediateValueDeclarationElement(typeElement, name, expr) => Option((typeElement, name, Option(expr)))
+    case OutputDeclarationElement(typeElement, name, expr) => Option((typeElement, name, Option(expr)))
+    case InputDeclarationElement(typeElement, name, expr) => Option((typeElement, name, expr))
+  }
+}
```
jbuabbud: \n

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wom/WorkflowDefinitionElementToWomWorkflowDefinition.scala
```diff
@@ -1,19 +1,108 @@
...
+  }
+
+  private def getOrdering(unlinkedGraphNodes: Set[UnlinkedGraphNode])
+                         (dependencies: List[(UnlinkedGraphNode, UnlinkedGraphNode)]): ErrorOr[List[UnlinkedGraphNode]] = {
+    // Find the topological order in which we must create the graph nodes:
+    val edges = dependencies map { case (from, to) => DiEdge(from, to) }
+
+    Graph.from[UnlinkedGraphNode, DiEdge](unlinkedGraphNodes, edges).topologicalSort match {
+      case Left(cycleNode) => s"This workflow contains a cyclic dependency on ${cycleNode.value}".invalidNel
+      case Right(topologicalOrder) => topologicalOrder.toList.map(_.value).validNel
```
jbuabbud: I'm getting a compilation error here due to type mismatch.

cjllanwarne: A compile error or an intelliJ squiggle? I get a red highlight but the compiler is happy. Maybe a comment would be useful though

jbuabbud: It's like a Heisenberg compile error. It only fails to compile if I'm looking at that line.

wdl/model/draft3/src/main/scala/wdl/model/draft3/elements/DeclarationElement.scala
```diff
@@ -27,3 +27,11 @@ object OutputDeclarationElement {
   * A declaration in an input block
   */
 final case class InputDeclarationElement(typeElement: TypeElement, name: String, expression: Option[ExpressionElement]) extends LanguageElement with WorkflowGraphNodeElement
+
+object DeclarationElement {
+  def unapply(languageElement: LanguageElement): Option[(TypeElement, String, Option[ExpressionElement])] = languageElement match {
```
jbuabbud: What is the intended use for this object? To me it seems to be used to conveniently access the members of one of the DeclarationElement types, but the LanguageElement type of the parameter is sort of confusing. I think a comment would be helpful.

wdl/model/draft3/src/main/scala/wdl/model/draft3/elements/DeclarationElement.scala
```diff
@@ -27,3 +27,11 @@ object OutputDeclarationElement {
...
+    case IntermediateValueDeclarationElement(typeElement, name, expr) => Option((typeElement, name, Option(expr)))
+    case OutputDeclarationElement(typeElement, name, expr) => Option((typeElement, name, Option(expr)))
+    case InputDeclarationElement(typeElement, name, expr) => Option((typeElement, name, expr))
+  }
```
cjllanwarne: Should have a `case _ => None`

wdl/model/draft3/src/main/scala/wdl/model/draft3/elements/DeclarationElement.scala
cjllanwarne: I'll definitely add a comment. This is a scala extractor (full description here: http://danielwestheide.com/blog/2012/11/21/the-neophytes-guide-to-scala-part-1-extractors.html)

TL;DR: 
* This is the function that allows the case match `case DeclarationElement(typeElement, name, Some(expr)) => ...` that's used [here](https://github.com/broadinstitute/cromwell/pull/3307/files#diff-3255c6301890ce6c0d4bb261ef95b340R21). 
* It's useful because it lets me treat all three of the types the same way in that match without having to retype the `=> ...` for each individual case class. 
* It takes a `LanguageElement` parameter because that's the type I'm matching against [here](https://github.com/broadinstitute/cromwell/pull/3307/files#diff-3255c6301890ce6c0d4bb261ef95b340R16).
* I made a note to myself to add a `case _ => None` to this unapply otherwise I'll get a MatchException thrown every time I try to run this against a `LanguageElement` which doesn't match one of these three cases.

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wom/ValueGeneratorAndConsumer.scala
```diff
@@ -0,0 +1,22 @@
+package wdl.draft3.transforms.wdlom2wom
+
+import simulacrum.typeclass
+import scala.language.implicitConversions
+
+@typeclass
+trait ValueGenerator[A] {
+  def generatedValues(a: A): Set[String]
```
danbills: Any chance of being stricter than `String`?   If it's really a string consider using Refined to enforce a structure.

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wom/ValueGeneratorAndConsumer.scala
```diff
@@ -0,0 +1,22 @@
+package wdl.draft3.transforms.wdlom2wom
+
+import simulacrum.typeclass
+import scala.language.implicitConversions
+
+@typeclass
+trait ValueGenerator[A] {
```
danbills: ToL Any way to be more specific?   By name it looks like it could create any value but by type sig looks like it only produces values of type string, seems like there's a potential for clarification.

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wom/WdlomWomExpression.scala
```diff
@@ -0,0 +1,34 @@
...
+
+  override def evaluateValue(inputValues: Map[String, WomValue], ioFunctionSet: IoFunctionSet): ErrorOr[WomValue] = expressionElement match {
+    case PrimitiveLiteralExpressionElement(v) => v.validNel
+    case IdentifierLookup(identifier) => inputValues.get(identifier).map(_.validNel).getOrElse(s"No value supplied for variable lookup: $identifier".invalidNel)
+    case _ => ??? // TODO other expression elements
+  }
+
+  override def evaluateType(inputTypes: Map[String, WomType]): ErrorOr[WomType] = expressionElement match {
+    case PrimitiveLiteralExpressionElement(v) => v.womType.validNel
+    case id: IdentifierLookup => inputMap((id: ExpressionElement).consumedValues.head).foundType.validNel
```
danbills: ToL `head` -> `headOption`

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wom/WdlomWomExpression.scala
```diff
@@ -0,0 +1,34 @@
...
+    case _ => ??? // TODO other expression elements
+  }
+
+  override def evaluateFiles(inputTypes: Map[String, WomValue], ioFunctionSet: IoFunctionSet, coerceTo: WomType): ErrorOr[Set[WomFile]] = {
+    Set.empty[WomFile].validNel
+  }
+}
+
+final case class FoundConsumedValue(inputName: String, foundType: WomType)
```
danbills: is this `inputName` the same as `ConsumedValue.name`?  If so perhaps there's a way this can be structured so that this info is only stated once

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wom/WdlomWomExpression.scala
danbills: why not match to `ExpressionElement` and avoid castiness (don't know what it's called to use `:`)

cjllanwarne: I need to be specific about the `IdentifierLookup` here because other expression elements will interact with the types of their subexpressions in different ways.

So `IdentifierLookup extends ExpressionElement` but the `@typeclass` magic doesn't add the `consumedValues` on subtypes, only on `ExpressionElement` itself. I dunno if there's a way to tell the `@typeclass` that subtypes are also OK?

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wom/WdlomWomExpression.scala
cjllanwarne: Sort of... `ConsumedValue` is unlinked, so it doesn't know how much of an expression like `Int x = a.b.c.d` is the value to lookup, and how much is member accesses on the looked up value (eg is there an object called `a` or a task output from `a` called `a.b`).

The `FoundConsumedValue` is the result of looking through all the generated values and finding one that can satisfy some prefix of the expression lookup, so that `inputName` might be `a.b` if it was a call output, or `a` if it was an object.

cjllanwarne: Maybe `UnlinkedConsumedValue` and `LinkedConsumedValue` would be a good idea for a rename?

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wom/WdlomWomExpression.scala
cjllanwarne: I did this because `IdentifierLookup` expressions will always have exactly one value they need to consume, but I guess safety first is never a bad idea.

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wom/ValueGeneratorAndConsumer.scala
cjllanwarne: Ah, yeah, it's actually finding the *names* of values that this `A` will generate. Maybe I could rename the def to `def generatedValueNames`.

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wom/ValueGeneratorAndConsumer.scala
cjllanwarne: See below comment, it's really `generatedValueNames` for the sake of linking different elements together.

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wom/WdlomWomExpression.scala
danbills: re: `@typeclass` Dunno.   AFAIK typeclasses and subtypes don't usually mix.

So it looks like your logic is correct given your constraints

cjllanwarne: Hmm I wonder whether that fact is indicating that I should actually have the logic in some new `ValueConsumer[IdentifierLookup]` so that I can call it directly from here, or indirectly from `ValueConsumer[ExpressionElement]`?

PR #3312, opened 2018-02-23
Allow placeholders in all String expressions

cjllanwarne:
Based on changes in https://github.com/cjllanwarne/wdl/pull/5
Updates String literal elements to maybe contain expression placeholders


wdl/model/draft3/src/main/scala/wdl/model/draft3/elements/ExpressionElement.scala
```diff
@@ -7,10 +7,16 @@ sealed trait ExpressionElement
 object ExpressionElement {
   final case class PrimitiveLiteralExpressionElement(value: WomPrimitive) extends ExpressionElement
 
+  final case class PlaceholderedStringLiteral(pieces: Seq[StringPiece]) extends ExpressionElement
```
jbuabbud: ToL "Placeholdered" could probably just be "Placeholder"

Issue #3270, opened 2018-02-13
CRON test somatic_cnv_panel/cnv_somatic_panel is broken


kshakir:
Last [CRON errors](https://travis-ci.org/broadinstitute/cromwell/jobs/340487274):

```scala
2018-02-12 13:50:46,018 cromwell-system-akka.dispatchers.engine-dispatcher-29 ERROR - WorkflowManagerActor Workflow 4c983def-4f9c-4e62-ae38-4543da9922de failed (during MaterializingWorkflowDescriptorState): Workflow input processing failed:
Required workflow input 'CNVSomaticPanelWorkflow.CollectCounts.ref_fasta_fai' not specified
Required workflow input 'CNVSomaticPanelWorkflow.PreprocessIntervals.ref_fasta' not specified
Required workflow input 'CNVSomaticPanelWorkflow.PreprocessIntervals.ref_fasta_fai' not specified
Required workflow input 'CNVSomaticPanelWorkflow.CollectCounts.ref_fasta' not specified
Required workflow input 'CNVSomaticPanelWorkflow.CollectCounts.ref_fasta_dict' not specified
cromwell.engine.workflow.lifecycle.materialization.MaterializeWorkflowDescriptorActor$$anon$1: Workflow input processing failed:
Required workflow input 'CNVSomaticPanelWorkflow.CollectCounts.ref_fasta_fai' not specified
Required workflow input 'CNVSomaticPanelWorkflow.PreprocessIntervals.ref_fasta' not specified
Required workflow input 'CNVSomaticPanelWorkflow.PreprocessIntervals.ref_fasta_fai' not specified
Required workflow input 'CNVSomaticPanelWorkflow.CollectCounts.ref_fasta' not specified
Required workflow input 'CNVSomaticPanelWorkflow.CollectCounts.ref_fasta_dict' not specified
	at cromwell.engine.workflow.lifecycle.materialization.MaterializeWorkflowDescriptorActor.cromwell$engine$workflow$lifecycle$materialization$MaterializeWorkflowDescriptorActor$$workflowInitializationFailed(MaterializeWorkflowDescriptorActor.scala:203)
```

PR #3313, opened 2018-02-23
Update IntegrationTestCases

ruchim:
CRON Job: https://travis-ci.org/broadinstitute/cromwell/jobs/344821732

TIL: The Mutect2 workflow contains a task where the WDL author had defined the `HOME` variable upon Docker image creation and used it inside the command block. Cromwell seems to be setting its own `HOME` variable in the exec.sh and thus caused the task to fail.


ruchim:
Linked to #3270 #3271

PR #3322, opened 2018-02-26
output_kv -> input_kv with '='

jbuabbud:
For expression assignment within input sections.


jbuabbud:
I killed the travis builds for this as the only change is to the WdlParser.

cjllanwarne:
For a change like this I'd prefer to see a test that proves that it works (eg, a call with one of these inputs being correctly converted into WDLOM)

Even for a change like this I'd still let sbt tests run to make sure this change hasn't regressed any draft-3 parsing code elsewhere.

jbuabbud:
Closing this and including the update to WdlParser in the call element PR that I'm working on.

PR #3324, opened 2018-02-27
fix expensive metadata sorting by not sorting

mcovarr:
ftfy the metadata sort by deleting it. Batching the summary updates didn't work out since the metadata summary "last updated id" scheme assumes rows are processed in batches where each batch has IDs greater than those of the preceding batch. AFAICT that wouldn't be possible without sorting all rows with IDs greater than the last processed row. Instead this just takes all committed rows with IDs greater than the last processed row and assumes that rows are being committed with monotonically increasing IDs.

### DIFF
--- a/database/sql/src/main/scala/cromwell/database/slick/MetadataSlickDatabase.scala
+++ b/database/sql/src/main/scala/cromwell/database/slick/MetadataSlickDatabase.scala
@@ -109,8 +109,8 @@ class MetadataSlickDatabase(originalDatabaseConfig: Config)
     } yield ()
   }
 
-  private def upsertCustomLabelEntry(metadataEntry: MetadataEntry)
-                                    (implicit ec: ExecutionContext): DBIO[Unit] = {
+  private def createCustomLabelEntryIfNecessary(metadataEntry: MetadataEntry)
+                                               (implicit ec: ExecutionContext): DBIO[Unit] = {
     //Extracting the label key from the MetadataEntry key
     val labelKey = metadataEntry.metadataKey.split("\\:", 2)(1)
     val labelValue = metadataEntry.metadataValue.toRawString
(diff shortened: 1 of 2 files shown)


database/sql/src/main/scala/cromwell/database/slick/MetadataSlickDatabase.scala
```diff
@@ -109,8 +109,8 @@ class MetadataSlickDatabase(originalDatabaseConfig: Config)
     } yield ()
   }
 
-  private def upsertCustomLabelEntry(metadataEntry: MetadataEntry)
-                                    (implicit ec: ExecutionContext): DBIO[Unit] = {
+  private def createCustomLabelEntryIfNecessary(metadataEntry: MetadataEntry)
```
mcovarr: unrelated naming fix

database/sql/src/main/scala/cromwell/database/slick/tables/MetadataEntryComponent.scala
```diff
@@ -109,17 +109,20 @@ trait MetadataEntryComponent {
...
+     statusMetadataKey: Rep[String], likeLabelMetadataKey: Rep[String]) => for {
       metadataEntry <- metadataEntries
       if metadataEntry.metadataEntryId >= metadataEntryId
       if (metadataEntry.metadataKey === startMetadataKey || metadataEntry.metadataKey === endMetadataKey ||
         metadataEntry.metadataKey === nameMetadataKey || metadataEntry.metadataKey === statusMetadataKey ||
         metadataEntry.metadataKey.like(likeLabelMetadataKey)) &&
         (metadataEntry.callFullyQualifiedName.isEmpty && metadataEntry.jobIndex.isEmpty &&
           metadataEntry.jobAttempt.isEmpty)
-    } yield metadataEntry).sortBy(_.metadataTimestamp)
+    } yield metadataEntry
```
cjllanwarne: ToL: Looks like this was sorting by timestamp not insertion ID. Is this safe against situations where status updates get re-ordered on the wire?

geoffjentry: I believe we're robust to out of order events for everything, but definitely for status

Horneth: The type of events being summarized are not aggregated based on "most recent" so we don't need to sort them by timestamp.
But for the rest of metadata we definitely DO want them to be sorted by timestamp.

mcovarr: We should be safe to reordering on the wire without sorting as long as we're not limiting the size of batches to process. And status is definitely safe thanks to the magic of CRDTs, while the other metadata only appears to be published from one spot for each.

cjllanwarne: 💯 - sorting a "journal"-style table does indeed seem like an antipattern to me, and especially so at insert-time

PR #3327, opened 2018-02-27
Throttle kv store an job store reads

Horneth:
Throttles the JobStore read queries and KV queries.
This imposes a hard limit that might be too low: not more than 1 query at a time. It might be too low in which case we can simply put it behind a router to increase the number of concurrent queries.
The important part is that we control how many of those requests are being done compared to "just send them to slick as fast as you possibly can".
Because it uses the same queue mechanism as batching it also makes it easy to turn the queue size into a metric that we can use to control the system's load.

TODO:

- [ ] Assess impact on scale

### DIFF
--- a/centaur/src/main/scala/centaur/test/Test.scala
+++ b/centaur/src/main/scala/centaur/test/Test.scala
@@ -191,7 +191,7 @@ object Operations {
           case _ =>
             // Re-add this to discover which test is running forever
             // But beware!! The stdout will break CWL conformance tests so you can't leave it in once the debugging is done:
-            //println(s"Waiting for completion of test '${testDefinition.testName}' as ${workflow.id}")
+            println(s"Waiting for completion of test '${testDefinition.testName}' as ${workflow.id}")
             pollDelay()
             doPerform()
         }
--- a/centaur/src/main/scala/centaur/test/Test.scala
+++ b/centaur/src/main/scala/centaur/test/Test.scala
@@ -191,7 +191,7 @@ object Operations {
           case _ =>
             // Re-add this to discover which test is running forever
             // But beware!! The stdout will break CWL conformance tests so you can't leave it in once the debugging is done:
-            println(s"Waiting for completion of test '${testDefinition.testName}' as ${workflow.id}")
+//            println(s"Waiting for completion of test '${testDefinition.testName}' as ${workflow.id}")
             pollDelay()
             doPerform()
         }
--- a/centaur/src/main/scala/centaur/test/Test.scala
+++ b/centaur/src/main/scala/centaur/test/Test.scala
@@ -191,7 +191,7 @@ object Operations {
           case _ =>
             // Re-add this to discover which test is running forever
             // But beware!! The stdout will break CWL conformance tests so you can't leave it in once the debugging is done:
-//            println(s"Waiting for completion of test '${testDefinition.testName}' as ${workflow.id}")
+            // println(s"Waiting for completion of test '${testDefinition.testName}' as ${workflow.id}")
             pollDelay()
             doPerform()
         }
--- a/centaur/src/main/scala/centaur/test/Test.scala
+++ b/centaur/src/main/scala/centaur/test/Test.scala
@@ -191,7 +191,7 @@ object Operations {
           case _ =>
             // Re-add this to discover which test is running forever
             // But beware!! The stdout will break CWL conformance tests so you can't leave it in once the debugging is done:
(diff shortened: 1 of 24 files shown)


core/src/main/scala/cromwell/core/actor/ThrottlerActor.scala
```diff
@@ -0,0 +1,25 @@
...
+
+import cats.syntax.traverse._
+import cats.instances.vector._
+import cats.instances.future._
+import scala.concurrent.Future
+import scala.concurrent.duration.Duration
+
+/**
+  * By removing the periodic flush and setting the batch size to 1,
+  * This transforms the BatchActor into throttler processing commands one at a time.
```
geoffjentry: `this`

core/src/main/scala/cromwell/core/actor/ThrottlerActor.scala
```diff
@@ -0,0 +1,25 @@
...
+  */
+abstract class ThrottlerActor[C] extends BatchActor[C](Duration.Zero, 1) {
+  override def weightFunction(command: C) = 1
+  override final def process(data: NonEmptyVector[C]): Future[Int] = {
+    // This ShouldNotBePossible™ but in case it happens, instead of dropping elements process them all anyway
```
geoffjentry: I'm assuming you're right here, but can you explain why finding yourself in an impossible state means that you should go ahead and proceed normally? That seems like a bad thing?

geoffjentry:
👍

core/src/main/scala/cromwell/core/actor/ThrottlerActor.scala
Horneth: With a batch size of 1 and each element having a weight of 1, the assumption is that only single element vectors will be passed to this function. That is because of the logic implemented in the parent `BatchActor` which calls the `process` function with a maxWeight of batchSize.
That is good because the goal of this class is to make the actor process one element at a time.
However if for some reason the `process` function is being called with more than one element (which would mean the parent `BatchActor` has a bug - which obviously is possible but hopefully not the case),
we don't want to just process the first one and drop the rest as they might be important operations (DB queries most likely). So instead it logs that something's wrong and process them all anyway.

centaur/src/main/scala/centaur/test/Test.scala
```diff
@@ -191,7 +191,7 @@ object Operations {
           case _ =>
             // Re-add this to discover which test is running forever
             // But beware!! The stdout will break CWL conformance tests so you can't leave it in once the debugging is done:
-            //println(s"Waiting for completion of test '${testDefinition.testName}' as ${workflow.id}")
+            // println(s"Waiting for completion of test '${testDefinition.testName}' as ${workflow.id}")
```
mcovarr: um

core/src/main/scala/cromwell/core/actor/BatchActor.scala
```diff
@@ -45,19 +46,21 @@ object BatchActor {
...
-  
+
   implicit val ec = context.dispatcher
   private val name = self.path.name
 
   override def preStart(): Unit = {
-    log.info("{} configured to flush with batch size {} and processHead rate {}.", name, batchSize, flushRate)
-    timers.startPeriodicTimer(ScheduledFlushKey, ScheduledProcessAction, flushRate)
+    log.info("{} configured to flush with batch size {} and process rate {}.", name, batchSize, flushRate)
+    if (flushRate != Duration.Zero) {
```
mcovarr: I'm afraid to ask

core/src/main/scala/cromwell/core/actor/ThrottlerActor.scala
```diff
@@ -0,0 +1,25 @@
...
+
+import cats.syntax.traverse._
+import cats.instances.vector._
+import cats.instances.future._
+import scala.concurrent.Future
+import scala.concurrent.duration.Duration
+
+/**
+  * By removing the periodic flush and setting the batch size to 1,
+  * this transforms the BatchActor into throttler processing commands one at a time.
```
mcovarr: That sentence doesn't quite English.

core/src/main/scala/cromwell/core/actor/BatchActor.scala
Horneth: I wasn't sure what would happen if you start a timer every 0 second so in case I `if`ed it.. :/

mcovarr: It certainly wouldn't be good; I was just worried you had actually seen this happening. 😄

PR #3334, opened 2018-03-01
If at first you fail try try check-failure-retry-count times again

mcovarr:


services/src/main/scala/cromwell/services/healthmonitor/HealthMonitorServiceActor.scala
```diff
@@ -32,49 +34,71 @@ trait HealthMonitorServiceActor extends Actor with LazyLogging with Timers {
...
+    case Check(subsystem, retriesLeft) =>
+      checkSubsystem(subsystem, retriesLeft)
+    case Store(subsystem, status) =>
+      store(subsystem, status)
+      scheduleSweepCheck(subsystem)
     case GetCurrentStatus => sender ! getCurrentStatus
     case ShutdownCommand => context.stop(self) // Not necessary but service registry requires it. See #2575
     case Status.Failure(f) => throw f
-    case _ => throw new RuntimeException("wtf")
+    case e => throw new RuntimeException(e.toString)
```
Horneth: ToL: is this really wanted ? Any unrecognized message should kill the actor ?

services/src/main/scala/cromwell/services/healthmonitor/HealthMonitorServiceActor.scala
```diff
@@ -32,49 +34,71 @@ trait HealthMonitorServiceActor extends Actor with LazyLogging with Timers {
...
+      case Success(status) =>
+        // This is also a success in getting the status but the status is a failure and there are no more retries available.
+        // Womp womp.
+        self ! Store(subsystem, status)
+      case Failure(NonFatal(_)) if retriesLeft > 0 =>
+        scheduleFailedRetryCheck(subsystem, retriesLeft)
+      case Failure(NonFatal(e)) =>
+        self ! Store(subsystem, SubsystemStatus(ok = false, messages = Option(List(e.getMessage))))
+      case f =>
+        self ! f
```
Horneth: ToL: maybe wrap that into a `Status.Failure(f)` or explicitly handle the case of a `scala.Failure` in the `receive` method ?

services/src/test/scala/cromwell/services/healthmonitor/HealthMonitorServiceActorSpec.scala
```diff
@@ -68,50 +63,91 @@ class HealthMonitorServiceActorSpec extends TestKitSuite with AsyncFlatSpecLike
...
+    var storeInvocationCount = 0
+    // Check the status that's being recorded, incrementing a counter each time it is invoked.
+    var statusStores = List.empty[SubsystemStatus]
+    def storeChecker(monitoredSubsystem: MonitoredSubsystem, subsystemStatus: SubsystemStatus): Unit = {
+      storeInvocationCount = storeInvocationCount + 1
+      assert(subsystemStatus.ok ^ (statusCheckCount == 4))
+      statusStores = statusStores :+ subsystemStatus
+    }
+
+    val FailAndCountSubsystem = MonitoredSubsystem("FailAndCountSubsystem", okThenFailStatusChecker _)
```
Horneth: lower case `f` ?

services/src/main/scala/cromwell/services/healthmonitor/HealthMonitorServiceActor.scala
mcovarr: Good question but preexisting, I only made the message possibly less useless.

services/src/main/scala/cromwell/services/healthmonitor/HealthMonitorServiceActor.scala
mcovarr: At this point in the match `f` could only be a `scala.Failure` already and is explicitly "handled" in the `receive` method:

```
    case Status.Failure(f) => throw f
```
🙄

Horneth: Ah right 😄

services/src/main/scala/cromwell/services/healthmonitor/HealthMonitorServiceActor.scala
Horneth: Okay, I'd lean towards logging an error rather than crashing the whole thing but 🤷‍♂️

PR #3332, opened 2018-02-28
Refactored token dispenser to accommodate for rate limiting

Horneth:

PR #3335, opened 2018-03-01
Load controller service and rate limiter

Horneth:
This is not wired to actually do anything yet but with https://github.com/broadinstitute/cromwell/pull/3332 it'll make it possible to start controlling the rate of job start based on metadata queue size

### DIFF
--- a/engine/src/main/scala/cromwell/server/CromwellSystem.scala
+++ b/engine/src/main/scala/cromwell/server/CromwellSystem.scala
@@ -5,7 +5,7 @@ import akka.http.scaladsl.Http
 import akka.stream.ActorMaterializer
 import com.typesafe.config.ConfigFactory
 import cromwell.engine.backend.{BackendConfiguration, CromwellBackends}
-import cromwell.engine.language.{CromwellLanguages, LanguageConfiguration, LanguageConfigurationEntry}
+import cromwell.engine.language.{CromwellLanguages, LanguageConfiguration}
 import cromwell.services.{EngineServicesStore, MetadataServicesStore}
 
 import scala.concurrent.Future
@@ -34,11 +34,15 @@ trait CromwellSystem {
   implicit private final lazy val ec = actorSystem.dispatcher
 
   def shutdownActorSystem(): Future[Terminated] = {
-    Http().shutdownAllConnectionPools() flatMap { _ =>
-      shutdownMaterializerAndActorSystem()
-    } recoverWith {
-      case _ => shutdownMaterializerAndActorSystem()
-    }
+    // If the actor system is already terminated it's already too late for a clean shutdown
+    if (!actorSystem.whenTerminated.isCompleted) {
+      Http().shutdownAllConnectionPools() flatMap { _ =>
+        shutdownMaterializerAndActorSystem()
+      } recoverWith {
+        case _ =>
+          shutdownMaterializerAndActorSystem()
+      }
+    } else actorSystem.whenTerminated
   }
   
   private def shutdownMaterializerAndActorSystem() = {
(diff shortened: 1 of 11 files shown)


services/src/main/scala/cromwell/services/loadcontroller/impl/LoadControllerServiceActor.scala
```diff
@@ -0,0 +1,69 @@
...
+  private val metadataQueueThreshold = serviceConfig.as[Option[Int]]("metadata-queue-threshold").getOrElse(1 * 1000 * 1000)
+  private val metadataQueueMetric = MetadataQueueMetric(metadataQueueThreshold)
+
+  private var loadLevel: LoadLevel = NormalLoad
+  private var loadMetrics: Map[LoadMetric, LoadLevel] = Map.empty
+
+  override def receive = listenerManagement.orElse(controlReceive)
+
+  override def preStart() = {
+    serviceRegistryActor ! ListenToMetadataWriteActor
```
Horneth: I think I'm going to change this from "this actor needs to ask the right actors to send it updates" to "whoever wants can send `LoadMetric`s to this actor and it will aggregate them and take action if necessary"

engine/src/main/scala/cromwell/server/CromwellSystem.scala
```diff
@@ -34,11 +34,15 @@ trait CromwellSystem {
...
-    } recoverWith {
-      case _ => shutdownMaterializerAndActorSystem()
-    }
+    // If the actor system is already terminated it's already too late for a clean shutdown
+    if (!actorSystem.whenTerminated.isCompleted) {
+      Http().shutdownAllConnectionPools() flatMap { _ =>
+        shutdownMaterializerAndActorSystem()
+      } recoverWith {
+        case _ =>
+          shutdownMaterializerAndActorSystem()
```
mcovarr: I see it's preexisting but is this `recover` in the right place? Try to run `shutdownMaterializerAndActorSystem` and if that doesn't work run it again? If so and you understand why it's like that a comment would be appreciated. 🙂

services/src/main/scala/cromwell/services/loadcontroller/impl/LoadControllerServiceActor.scala
mcovarr: Is something guaranteed to notice if one of those "right actors" is AWOL and therefore its `LoadMetric`s never arrive?

Horneth: Good question, presumably if one of the actors goes AWOL it's going to be outside the scope of what this actor can do to fix it (at least for now).
But we could have it actor could watch every new actor that starts sending metrics to it and if one of them dies it could automatically raise the load level ?

engine/src/main/scala/cromwell/server/CromwellSystem.scala
```diff
@@ -34,11 +34,15 @@ trait CromwellSystem {
   implicit private final lazy val ec = actorSystem.dispatcher
 
   def shutdownActorSystem(): Future[Terminated] = {
-    Http().shutdownAllConnectionPools() flatMap { _ =>
-      shutdownMaterializerAndActorSystem()
-    } recoverWith {
-      case _ => shutdownMaterializerAndActorSystem()
-    }
+    // If the actor system is already terminated it's already too late for a clean shutdown
+    if (!actorSystem.whenTerminated.isCompleted) {
```
kshakir: ToL: This does seem better that what was there before, but checking the `Future.isCompleted` is false then trying to run another future seems like a race condition. `!isCompleted` doesn't mean that a shutdown hasn't already started.

"ToL" because off the top of my head I don't have anything better at the moment. Someday we should audit all the code and ensure no one is terminating the actor system without going through `def shutdownActorSystem()`, such that this check should _in theory_ never be necessary.

Horneth: Good point, I think I added this because in the event where the root actor crashes at initialization, it escalates to the guardian which itself dies which shuts down the actor system, but we still end up here trying to shut it down again and it can get stuck for some reason.
So in this case it's not even another part of the code it's just a side effect of supervision.

PR #3350, opened 2018-03-03
Cwl secondary files

kshakir:


kshakir:
Via "the docker":

```
[2018-03-03 01:11:21,96] [info] SingleWorkflowRunnerActor workflow finished with status 'Succeeded'.
{ 
  "outputs": {
    "workflow.cwl.mark_duplicates_metrics_file": {
      "format": null,
      "location": "/root/cromwell-executions/workflow.cwl/b745d4f8-aa09-402c-b612-fb112fe8d983/call-mark_duplicates_and_sort/execution/mark_dups_metrics.txt",
      "size": null,
      "secondaryFiles": [],
      "contents": null,
      "checksum": null,
      "class": "File"
    },
    "workflow.cwl.final_cram": {
      "format": null,
      "location": "/root/cromwell-executions/workflow.cwl/b745d4f8-aa09-402c-b612-fb112fe8d983/call-index_cram/execution/final.cram",
      "size": null,
      "secondaryFiles": ["/root/cromwell-executions/workflow.cwl/b745d4f8-aa09-402c-b612-fb112fe8d983/call-index_cram/execution/final.cram.crai", "/root/cromwell-executions/workflow.cwl/b745d4f8-aa09-402c-b612-fb112fe8d983/call-index_cram/execution/final.crai"],
      "contents": null,
      "checksum": null,
      "class": "File"
    }
  },
  "id": "b745d4f8-aa09-402c-b612-fb112fe8d983"
}
```

cwl/src/main/scala/cwl/CwlWomExpression.scala
```diff
@@ -49,65 +50,160 @@ final case class InitialWorkDirFileGeneratorExpression(entry: IwdrListingArrayEn
...
+  }
+
+  override def evaluateFiles(inputTypes: Map[String, WomValue],
+                             ioFunctionSet: IoFunctionSet,
+                             coerceTo: WomType): ErrorOr[Set[WomFile]] = {
+    "Programmer error: Shouldn't use InitialWorkDirRequirement listing to find output files.".invalidNel
+  }
+
+  /**
+    * We already get all of the task inputs when evaluating, and we don't need to highlight anything else
```
mcovarr: this comment seems oddly womtooly

mcovarr:
Could you give a high-level overview of the changes here? Obviously the state machine is the star of this PR but there are quite a few changes in IWD and other spots where I wasn't expecting them. It looks like it's mostly a refactor but I'm not completely sure.

geoffjentry:
Likewise. I chalked my confusion up to being not embroiled in the CWL code but since Captain Seagull is saying something similar perhaps it ain't just me

centaur/src/main/resources/standardTestCases/cwl_secondary_files/cwl_secondary_files.cwl
```diff
@@ -0,0 +1,30 @@
...
+        items: File
+        inputBinding:
+          position: 3
+      secondaryFiles: [.also]
+  outputs:
+    ignored:
+      type: int
+      outputBinding:
+        glob: "*.txt"
+        outputEval: $(self.length)
```
mcovarr: this is a goofy bit of vestigial code and could maybe just return a constant integer expression or something

mcovarr: Note: I wrote this bit and am the source of the goofiness.

cwl/src/main/scala/cwl/ExpressionEvaluator.scala
```diff
@@ -13,46 +12,31 @@ import wom.values.{WomFloat, WomInteger, WomString, WomValue}
...
   // A code fragment wrapped in the $(...) syntax must be evaluated as a ECMAScript expression.
-  val ECMAScriptExpressionWitness = Witness("""(?s)\s*\$\((.*)\)\s*""")
+  /*
+  There are several places in cwltool where a simple contains check is used:
+  - https://github.com/common-workflow-language/cwltool/blob/353dbed/cwltool/builder.py#L157
+  - https://github.com/common-workflow-language/cwltool/blob/353dbed/cwltool/command_line_tool.py#L669
+  - https://github.com/common-workflow-language/cwltool/blob/353dbed/cwltool/expression.py#L219
+  - https://github.com/common-workflow-language/cwltool/blob/353dbed/cwltool/process.py#L645
+   */
+  val ECMAScriptExpressionWitness = Witness("""(?s).*\$\(.*""")
```
mcovarr: Dare I ask what happens if that's in a quoted string?

kshakir:
In addition to the commit comments.

Specifically for "the docker":
- InputParameters weren't wired in for secondary files. Now they are, with Polys and a bit of borrowing from the OutputParameters. The common code was refactored into FileParameter.
- CWL spec says InputArraySchema's can't have secondary files, and that they should be specified on the InputBinding. The CWL via "the docker" wants them on the IAS anyway. The IAS secondary files are now wired in through a Poly.
- "The docker" was using strings like `$("foo")/$("bar")` that was tripping up our various expression regex patterns. This PR replaces them with the (latest as of Friday) copy-port of cwltool's state machine.

The state machine works on _any_ string, even those without `"$("`. But the spec for secondary files says that only expressions should be returned-as-rendered, while plain strings should be instead appended instead of used as literals. It turns out this discrimination is done in cwltool `"$(" in expr` (see numerous links in the scala comments). So the presence of `"$("` is now used for our `ECMAScriptExpression` to similarly guess if a string is a interpolated string or a regular string. `ECMAScriptExpression` and `InterpolatedString` were pretty much the same thing and were merged back together.

There was also a bit of code where if a malicious (or errant) CWL was submitted a `java.lang.Error` would be thrown possibly exiting the JVM. The code around this section was refactored to use a compile-time-checked Poly instead of hope-we-got-them-all pattern matching.

cwl/src/main/scala/cwl/CwlWomExpression.scala
kshakir: Copied from below. Happy to update if there's better text.

cwl/src/main/scala/cwl/ExpressionEvaluator.scala
kshakir: Thanks for the comment. I reversed (and tested) that `${return "$(";}` matches a `Function` and not an `Expression`, as the wrong order was matching the wrong type.

The copy-ported state machine can handle plain strings, so a quoted `"$("` will work just fine passing through the machine.

However, secondary files seem to handle expressions and strings differently, so with this regex match we should be effectively doing the same thing as cwltool. Now, if someone is using `$(` inside a quoted string that should not be an expression, I think they'll have a bad time in general.

cwl/src/main/scala/cwl/InputParameter.scala
```diff
@@ -84,4 +87,59 @@ object InputParameter {
...
+
+      womValue => populateFiles(womValue).toTry(s"loading $womValue for ${inputParameter.id}").get
+    }
+  }
+
+  private def maybeLoadContents(womMaybePopulatedFile: WomMaybePopulatedFile,
+                                ioFunctionSet: IoFunctionSet,
+                                loadContents: Boolean): ErrorOr[WomMaybePopulatedFile] = {
+    if (loadContents) {
+      FileParameter.load64KiB(womMaybePopulatedFile.value, ioFunctionSet) map { contents =>
```
Horneth: I wonder if we should check if `womMaybePopulatedFile` already has its content loaded for some reason, and if that's the case skip re-loading it here ?

cwl/src/main/scala/cwl/MyriadInputTypeToSortedCommandParts.scala
```diff
@@ -80,8 +97,11 @@ object MyriadInputInnerTypeToSortedCommandParts extends Poly1 {
       case (inputBinding, WomOptionalValue(_, Some(value)), sortingKey, expressionLib) =>
         go(inputBinding, value, sortingKey, expressionLib)
 
-        //If it's optional and there's no value, do nothing
-      case (_, WomOptionalValue(_, None), _, _) => List.empty
+      // If it's optional and there's no value, do nothing
+      case (_, WomOptionalValue(_, None), _, _) => CommandPartsList.empty
+
+      // If there's no input binding and no input bindings within the irs, do nothing
+      case (None, _, _, _) if !irs.fields.exists(_.exists(_.inputBinding.isDefined)) => CommandPartsList.empty
```
Horneth: Thanks for fixing this

cwl/src/main/scala/cwl/InputParameter.scala
kshakir: At some point, absolutely... I'll file a new ticket for our missing functionality on "File literals" where the content is already provided. In "File literals" there is no file path, we're supposed to make up a temp path. That special handling could also overlap here with skipping re-loading.

cwl/src/main/scala/cwl/CwlWomExpression.scala
```diff
@@ -49,65 +50,160 @@ final case class InitialWorkDirFileGeneratorExpression(entry: IwdrListingArrayEn
...
+              }
+            case other => for {
+              coerced <- WomStringType.coerceRawValue(other).toErrorOr
+              contentString = coerced.asInstanceOf[WomString].value
+              // We force the entryname to be specified, and then evaluate it:
+              entryNameStringOrExpression <- expressionDirent.entryname.toErrorOr(
+                "Invalid dirent: Entry was a string but no file name was supplied")
+              entryName <- entryNameStringOrExpression.fold(EntryNamePoly).apply(parameterContext, expressionLib)
+              writeFile = ioFunctionSet.writeFile(entryName, contentString)
+              writtenFile <- validate(Await.result(writeFile, Duration.Inf))
```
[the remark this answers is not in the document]
kshakir: Probably? This PR was primarily aimed at fixing "the docker". This section here is a Poly refactoring of the previous pattern-matching code aimed at removing the `???`. No existing conformance tests were injured or harmed during the copy/paste of the former code.

I will be the owner under the "last touched" rule, but I've left a comment above that points out that the Dirent logic _doesn't look quite right_ and to change as necessary.

Re: your specific comment, yep `writeFile` would fail on a directory. CWL Directories are each effectively a recursive `listing: Option[Array[FileOrDirectory]]`. If there's `Some()` listing, one is supposed to write the supplied listing. If the listing is `None` one is supposed to query the file-system to find out what files are under that path. Our expression code will decode into a `Directory` object, but we'll likely need more wiring for handling Dirent / directories-in-general for InitialWorkDir.

kshakir:
Merging while we figure out centaur issues.

PR #3356, opened 2018-03-05
Limit number of queued jobs per workflow

Horneth:
This might be a bit controversial so feel free to seagull.

_What it does_: When looking for new nodes to run in a workflow, if there are more than 1000 call nodes in "Queued" state, don't start new call nodes (still execute other ones like ExpressionNodes etc...).
_Pro_: Reduce load by not starting too many jobs that won't be able to run for the moment anyway since there's already 1000+ queued (waiting for a token)
_Con_: Jobs stay in `NotStarted` state longer (this status is sent to metadata and is visible by users), even if they could technically be started as far as dependencies are concerned.

### DIFF
--- a/core/src/main/scala/cromwell/core/ExecutionStatus.scala
+++ b/core/src/main/scala/cromwell/core/ExecutionStatus.scala
@@ -2,7 +2,7 @@ package cromwell.core
 
 object ExecutionStatus extends Enumeration {
   type ExecutionStatus = Value
-  val NotStarted, QueuedInCromwell, Starting, Running, Aborting, Failed, RetryableFailure, Done, Bypassed, Aborted, Unstartable = Value
+  val NotStarted, WaitingForQueueSpace, QueuedInCromwell, Starting, Running, Aborting, Failed, RetryableFailure, Done, Bypassed, Aborted, Unstartable = Value
   val TerminalStatuses = Set(Failed, Done, Aborted, Bypassed, Unstartable)
   val TerminalOrRetryableStatuses = TerminalStatuses + RetryableFailure
   val NonTerminalStatuses = values.diff(TerminalOrRetryableStatuses)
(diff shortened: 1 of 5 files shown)


mcovarr:
Is this kind of redundant to the token system?

Horneth:
In a sense yes, but it's upstream of it.
Currently if you scatter 10 million wide, Cromwell still creates 10 million EJEAs that are going to ask for a token.
This stops it from even creating EJEAs if it knows they won't be able to run anyway (yet).
To be perfectly honest I just wanted to kinda float the idea as I haven't gotten to precisely measure if it indeed reduces cpu / memory load.
I wanted to get people's opinion and if it's a no go already I won't bother measuring.

Horneth:
To clarify after discussion with @mcovarr:
This is a per-workflow limitation: each workflow chooses not to start new jobs until it sees its number of "queued jobs" go under a certain limit.
It doesn't protect against someone starting 10 million workflows with 1 job. That's what the token dispenser does.
It would make it more difficult to distinguish between a job that can't be started because its dependencies are not fulfilled and a job that is not being started because there are too many queued jobs already: they would both have the same `NotStarted` status. We could introduce a new status but this has a cost in terms of educating users etc...
This seems TechTalk worthy

geoffjentry:
I think the effect is fine. We often tell people that once a job is runnable that Cromwell fires it off, but that's always used as a way to help them understand that Cromwell isn't partaking in true scheduling (ie resource based negotiation via SGE, PAPI, etc). IMO it's absolutely ok for jobs which are runnable to have not started if that's the limitation the system imposes.

Further, I think it's a good move in the resiliency front. Infinite scalability is a great goal, but from a practical perspective a limit is always going to be reached, so finding ways to make the system manage to keep on ticking ok when that happens is a good thing. That's generally going to involve slowing things down.

Horneth:
After tech talk discussion this is in standby until after we release

mcovarr:
mergeable now that we've (you've) released?

Horneth:
I think so, if you and @kshakir or someone else don't mind giving it a second look that'd be nice though because I updated it a bit last week after realizing it was broken in some case.
I should also probably update the changelog to advertise the introduction of `WaitingForQueueSpace` status

PR #3357, opened 2018-03-05
PAPI Updates

Horneth:
- Send abort requests through the `JesAPIQueryManager`. This is wanted because currently each JABJEA aborts on its own which has undesired consequences, like flooding the backend thread pool with blocking requests.
- Lift the "1 second" maximum limit of the `JesPollingActor` by switching to milliseconds (new limit is 1 millisecond)
- Add a second `JesPollingActor` to help with throughput of PAPI requests

### DIFF
--- a/supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesApiQueryManager.scala
+++ b/supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesApiQueryManager.scala
@@ -84,7 +84,7 @@ class JesApiQueryManager(val qps: Int Refined Positive, override val serviceRegi
   private def statusPollerProps = JesPollingActor.props(self, workerQps, serviceRegistryActor)
 
   // statusPollers is protected for the unit tests, not intended to be generally overridden
-  protected[statuspolling] val statusPollers: Array[ActorRef] = resetAllWorkers()
+  protected[statuspolling] var statusPollers: Vector[ActorRef] = resetAllWorkers()
 
   override def preStart() = {
     timers.startSingleTimer(QueueMonitoringTimerKey, QueueMonitoringTimerAction, 10.seconds)
@@ -100,8 +100,7 @@ class JesApiQueryManager(val qps: Int Refined Positive, override val serviceRegi
       if (creationQuery.contentLength > maxBatchRequestSize) {
         creationQuery.requester ! JesApiRunCreationQueryFailed(creationQuery, requestTooLargeException)
       } else workQueue :+= creationQuery
-    case DoAbortRun(workflowId, run) =>
-      workQueue :+= makeAbortQuery(workflowId, sender, run)
+    case DoAbortRun(workflowId, run) => workQueue :+= makeAbortQuery(workflowId, sender, run)
     case q: JesApiQuery => workQueue :+= q
     case RequestJesPollingWork(maxBatchSize) =>
       log.debug("Request for JES Polling Work received (max batch: {}, current queue size is {})", maxBatchSize, workQueue.size)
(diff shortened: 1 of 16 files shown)


supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesApiQueryManager.scala
```diff
@@ -65,6 +68,10 @@ class JesApiQueryManager(val qps: Int Refined Positive, override val serviceRegi
       "and localize the files yourself."
   ))
 
+  private[statuspolling] lazy val nbWorkers = 2
+  private lazy val workerQps = refineV[Positive](Math.max(qps.value / nbWorkers, 1))
+    .getOrElse(throw new IllegalArgumentException("Programmer error ! Keep the min qps at 1 !"))
```
mcovarr: Isn't the "or else" after the `Math.max` unreachable? It might not be a bad idea to actually throw on a nonsensical qps value assuming the failure mode is reasonable.

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesApiQueryManager.scala
```diff
@@ -219,7 +243,12 @@ object JesApiQueryManager {
     */
   final case class DoCreateRun(workflowId: WorkflowId, genomicsInterface: Genomics, rpr: RunPipelineRequest) extends JesApiQueryManagerRequest
 
-  private[statuspolling] trait JesApiQuery {
+  /**
+    * Abort a pipeline.
+    */
+  final case class DoAbortRun(workflowId: WorkflowId, run: Run) extends JesApiQueryManagerRequest
+
+  private[statuspolling] sealed trait JesApiQuery {
```
mcovarr: TOLish and obviously preexisting but the use of "Query" terminology in this API for state-changing operations is confusing, particularly when the more accurate "request" is used internally

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/PapiInstrumentation.scala
```diff
@@ -14,12 +14,15 @@ object PapiInstrumentation {
   private val PapiKey = NonEmptyList.of("papi")
   private val PapiPollKey = PapiKey.concat("poll")
   private val PapiRunKey = PapiKey.concat("run")
-  
+  private val PapiAbortKey = PapiKey.concat("run")
```
mcovarr: 👁

supportedBackends/jes/src/main/scala/cromwell/backend/impl/jes/statuspolling/JesPollingActor.scala
```diff
@@ -107,13 +108,10 @@ object JesPollingActor {
...
-    * easier
     */
   def determineBatchInterval(qps: Int Refined Positive): FiniteDuration = {
-    val maxInterval = MaxBatchSize / qps.value
-    val interval = Math.max(maxInterval / 0.9, 1).toInt
-    interval.seconds
+    val maxInterval = MaxBatchSize.toDouble / qps.value.toDouble
+    val interval = ((maxInterval / 0.9) * 1000).toInt
+    interval.milliseconds
   }
```
mcovarr: From the review comments, three orders of magnitude seems like kind of a big change in the frequency limit all at once. 😬

mcovarr: TOL even if we stick with the `Refined Positive` qps after my earlier gulling, it's now less clear here that the result of this expression is constrained to be positive.

Horneth: About your first comment I don't think it matters too much, because this time value is actually a delay we wait for before doing the next request, and after we finish our current one. So we won't go faster than we can anyway, we might go as fast as we can + 1ms instead of +1s.

PR #3355, opened 2018-03-05
Member access, binary, unary and if/then/else expression evaluation

cjllanwarne:

PR #3358, opened 2018-03-05
Allow nested structs and member accesses

cjllanwarne:
- [x] Needs rebase onto develop after #3355 
- Motivating example: https://github.com/broadinstitute/cromwell/pull/3358/files#diff-6994a325db20e7eee1b98e63c6c685b3

### DIFF
--- a/wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/linking/expression/package.scala
+++ b/wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/linking/expression/package.scala
@@ -9,7 +9,7 @@ import wdl.model.draft3.graph.UnlinkedValueConsumer.ops._
 import wdl.model.draft3.elements.ExpressionElement
 import wdl.model.draft3.elements.ExpressionElement._
 import wdl.model.draft3.graph.expression.WomExpressionMaker
-import wdl.model.draft3.graph.{GeneratedValueHandle, UnlinkedConsumedValueHook, UnlinkedIdentifierHook, UnlinkedValueConsumer}
+import wdl.model.draft3.graph._
 import wom.expression.WomExpression
 
 package object expression {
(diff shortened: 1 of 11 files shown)


wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/linking/expression/types/package.scala
```diff
@@ -2,28 +2,66 @@ package wdl.draft3.transforms.linking.expression
...
+//        case a: LessThanOrEquals => a.evaluateType(linkedValues)
+//        case a: GreaterThan => a.evaluateType(linkedValues)
+//        case a: GreaterThanOrEquals => a.evaluateType(linkedValues)
+//        case a: Add => a.evaluateType(linkedValues)
+//        case a: Subtract => a.evaluateType(linkedValues)
+//        case a: Multiply => a.evaluateType(linkedValues)
+//        case a: Divide => a.evaluateType(linkedValues)
+//        case a: Remainder => a.evaluateType(linkedValues)
+
+//        case a: TernaryIf => a.evaluateType(linkedValues)
```
mcovarr: that's a lot of commented out code

mcovarr:
Why is this Red Thumb? It looks pure WDLy.

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/linking/expression/package.scala
```diff
@@ -38,14 +38,21 @@ package object expression {
...
-      case o: ObjectLiteral => o.consumedValueHooks
-      case p: PairLiteral => p.consumedValueHooks
+      case a: ObjectLiteral => a.consumedValueHooks
+      case a: PairLiteral => a.consumedValueHooks
       case a: ArrayLiteral => a.consumedValueHooks
-      case m: MapLiteral => m.consumedValueHooks
+      case a: MapLiteral => a.consumedValueHooks
+
+      case a: IdentifierLookup => a.consumedValueHooks
+      case a: IdentifierMemberAccess => a.consumedValueHooks
```
mcovarr: TOL I don't have enough background to judge in this particular case, but it seems that while typeclasses are a nice way to add type-specific handling for nonessential concerns, they do introduce quite a bit of boilerplate as evidenced above. If these actually are essential concerns they might be less noisily modeled in a supertrait, but as I said I can't really judge how essential this stuff is to the `*Literal`s.

cjllanwarne: It's hard to tell with an `n` of 1 whether this will really work... but I'm hoping that this scheme will make switching out evaluators for different language versions easier (because they're not baked into the case classes but implemented by the specific language version).

PR #3361, opened 2018-03-06
Updated a quay tag's hash & manipulate null outputs

kshakir:


centaur/src/main/resources/standardTestCases/docker_hash_quay.test
```diff
@@ -7,5 +7,7 @@ files {
 
 metadata {
   "calls.docker_hash_quay.quay.runtimeAttributes.docker": "quay.io/aptible/ubuntu:12.04",
-  "calls.docker_hash_quay.quay.dockerImageUsed": "quay.io/aptible/ubuntu@sha256:cf53a62c50cd1ec4e48b1c995451b94005d3e85527d06ce9151b51b0f1b6dd87"
+  # NOTE: Whenever aptible updates their tagged images then the sha256's will change.
```
Horneth: I'm getting us a `cromwell-docker-test` repo on quay so when we have it we can use that

src/bin/travis/resources/local_conformance_expected_failures.txt
```diff
@@ -1,4 +1,4 @@
-1
+#1
```
Horneth: :D

centaur/src/main/resources/standardTestCases/docker_hash_quay.test
kshakir: Discussed. Will merge now and then when our quay repo is available another PR will be submitted.

PR #3366, opened 2018-03-06
Load Control

Horneth:
This sets up a mechanism to:
1) Collect information about "load" from various part of the system
2) Summarize this information and calculate a global load
3) Notify parts of the system of changes in the global load

Current implementation is simple:
Only 2 load levels: `NormalLoad` and `HighLoad`
Actors reporting their load are:
- WriteMetadataActor
- JobStoreReadActor
- JobStoreWriteActor
- CallCacheWriteActor
- CallCacheReadActor
- KeyValueReadActor
- KeyValueWriteActor
- IoActor
- JesAPIQueryManagerActor

Additionally free memory is also being monitored and will go to `HighLoad` if going below a certain threshold

Global load == max(all load levels). So if one actor or more say their load is high, the global load will be high, otherwise normal.
The only actor listening to changes on the global load is the job token dispenser. It will stop dispensing tokens when load is high and start again when load is back to normal.

At the exception of the IoActor, all the above mentioned actors have a queue in which they store work to be done. Their load is determined by comparing the size of this queue to a threshold.
The IoActor's queue is not easily accessible because hidden in the stream implementation and its size cannot easily be known. However we know when its full because we can't add to it anymore (this is when backpressure messages are sent). When that happens the IO actor reports its load to be `High`. When it hasn't had to backpressure for 10 seconds, the load returns to normal.

There are many ways this could be made smarter but it already yields improvements in terms of stability and robustness.

TODO: 

- [x] Add Changelog
- [x]  Configuration ? Lots of thresholds and values in this PR that could be configurable, how much and how do we want to configure ?

### DIFF
--- a/database/sql/src/main/scala/cromwell/database/slick/tables/JobKeyValueEntryComponent.scala
+++ b/database/sql/src/main/scala/cromwell/database/slick/tables/JobKeyValueEntryComponent.scala
@@ -1,6 +1,7 @@
 package cromwell.database.slick.tables
 
 import cromwell.database.sql.tables.JobKeyValueEntry
+import slick.jdbc.MySQLProfile
 
 trait JobKeyValueEntryComponent {
 
@@ -31,6 +32,7 @@ trait JobKeyValueEntryComponent {
   }
 
   protected val jobKeyValueEntries = TableQuery[JobKeyValueEntries]
+  lazy val jobKeyValueTableQueryCompiled = MySQLProfile.compileInsert(jobKeyValueEntries.toNode)
 
   val jobKeyValueEntryIdsAutoInc = jobKeyValueEntries returning jobKeyValueEntries.map(_.jobKeyValueEntryId)
 
--- a/database/sql/src/main/scala/cromwell/database/slick/tables/JobKeyValueEntryComponent.scala
+++ b/database/sql/src/main/scala/cromwell/database/slick/tables/JobKeyValueEntryComponent.scala
@@ -1,7 +1,6 @@
 package cromwell.database.slick.tables
 
 import cromwell.database.sql.tables.JobKeyValueEntry
-import slick.jdbc.MySQLProfile
 
 trait JobKeyValueEntryComponent {
 
@@ -32,7 +31,7 @@ trait JobKeyValueEntryComponent {
   }
 
   protected val jobKeyValueEntries = TableQuery[JobKeyValueEntries]
-  lazy val jobKeyValueTableQueryCompiled = MySQLProfile.compileInsert(jobKeyValueEntries.toNode)
+  lazy val jobKeyValueTableQueryCompiled = driver.compileInsert(jobKeyValueEntries.toNode)
 
   val jobKeyValueEntryIdsAutoInc = jobKeyValueEntries returning jobKeyValueEntries.map(_.jobKeyValueEntryId)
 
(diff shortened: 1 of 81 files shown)


cjllanwarne:
ToL: (you can never get enough of a good thing...): being able to see a graph of components' "load level" over time would be awesome...!

Horneth:
screen shot 2018-03-06 at 3 55 19 pm

cjllanwarne:
So... I guess Cromwell is really just a fancy PAPI querying engine...

Horneth:
@cjllanwarne yep ! 😄 That was an earlier run I've been trying to improve PAPI throughput but it's one of the highest "load generators" for now

Horneth:
It also depends on the situation Cromwell is dealing with, in other circumstances I/O was the main limiter for instance

services/src/main/scala/cromwell/services/keyvalue/KeyValueReadActor.scala
```diff
@@ -0,0 +1,33 @@
...
+  extends EnhancedThrottlerActor[CommandAndReplyTo[KvGet]] {
+  override def receive = enhancedReceive.orElse(super.receive)
+  
+  override def processHead(head: CommandAndReplyTo[KvGet]) = instrumentedProcess {
+    val processed = processGet(head.command)
+    processed onComplete {
+      case Success(response) => head.replyTo ! response
+      case Failure(f) => head.replyTo ! KvFailure(head.command, f)
+    }
+    processed.map(_ => 1)
```
danbills: would you please provide a comment here?

services/src/main/scala/cromwell/services/keyvalue/KeyValueWriteActor.scala
```diff
@@ -0,0 +1,41 @@
...
+    val processed = processPut(data.map(_.command).toVector)
+    processed onComplete {
+      case Success(_) => data.toVector.foreach({
+        case CommandAndReplyTo(command, replyTo) => replyTo ! KvPutSuccess(command)
+      })
+      case Failure(f) => data.toVector.foreach({
+        case CommandAndReplyTo(command, replyTo) => replyTo ! KvFailure(command, f)
+      })
+    }
+    processed.map(_ => data.length)
```
danbills: as above please and thank you

services/src/test/scala/cromwell/services/loadcontroller/impl/LoadControllerServiceActorSpec.scala
```diff
@@ -0,0 +1,94 @@
...
+import org.scalatest.{FlatSpecLike, Matchers}
+
+import scala.concurrent.duration._
+
+object LoadControllerServiceActorSpec {
+  val Config = ConfigFactory.parseString(
+    """
+      |control-frequency = 1 second
+    """.stripMargin
+  )
```
kshakir: `val Config = ConfigFactory.parseString("control-frequency = 1 second")`

database/sql/src/main/scala/cromwell/database/slick/tables/JobKeyValueEntryComponent.scala
```diff
@@ -31,6 +32,7 @@ trait JobKeyValueEntryComponent {
   }
 
   protected val jobKeyValueEntries = TableQuery[JobKeyValueEntries]
+  lazy val jobKeyValueTableQueryCompiled = MySQLProfile.compileInsert(jobKeyValueEntries.toNode)
```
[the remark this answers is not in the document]
Horneth: Cool thanks ! I'm trying it

PR #3369, opened 2018-03-07
Partial wiring for Sam whitelist checking.

kshakir:
Made GoogleAuthMode scopes optional via a param/conf, still defaulting to the pipelines api scopes.
Removed unused scopes and data-dirs from google auth modes that do not uses them.
Moved access token TTL refresher from GcrAbstractFlow to GoogleAuthMode.
Fixed queryPostRoute logging method as "GET" instead of "POST".

### DIFF
--- a/CromIAM/src/main/scala/cromiam/server/config/CromIamServerConfig.scala
+++ b/CromIAM/src/main/scala/cromiam/server/config/CromIamServerConfig.scala
@@ -1,18 +1,25 @@
 package cromiam.server.config
 
 import akka.http.scaladsl.settings.ServerSettings
+import cats.instances.option._
 import cats.syntax.apply._
+import cats.syntax.traverse._
 import cats.syntax.validated._
 import com.typesafe.config.{Config, ConfigFactory}
-import common.validation.ErrorOr.ErrorOr
+import common.validation.ErrorOr._
+import common.validation.Validation._
+import cromiam.server.config.CromIamServerConfig._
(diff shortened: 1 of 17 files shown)


geoffjentry:
NB: I'm not going to bother github commenting on things we already talked about on slack

CromIAM/src/main/scala/cromiam/server/config/CromIamServerConfig.scala
```diff
@@ -1,28 +1,37 @@
...
+import net.ceedubs.ficus.Ficus._
 
 import scala.collection.JavaConverters._
 import scala.util.{Failure, Success, Try}
-import cromiam.server.config.CromIamServerConfig._
 
 final case class CromIamServerConfig(cromIamConfig: CromIamConfig,
                                      cromwellConfig: ServiceConfig,
                                      samConfig: ServiceConfig,
+                                     whitelistAuthModeOption: Option[GoogleAuthMode],
```
geoffjentry: I see why you were asking about this now. This (and the relevant portion in the conf) seems like it should be part of the sam config block*. Making that change would imply that `ServiceConfig` is no longer the right type for `samConfig`, but instead an e.g. `SamConfig` type which contains a `ServiceConfig`

* if I'm correct in believing that it's only used by the sam client in cromiam

geoffjentry:
:+1: modulo the one comment here and the stuff we chatted about earlier

kshakir:
WRONG.

PR #3373, opened 2018-03-07
Re-organize the running state of EJEA and tests

Horneth:
The bug I mentioned this morning came from a change in the EJEA (it didn't get caught by unit tests because they're not being run, but that's another story..)
They're still not being run in this PR so this is on hold until we fix this (@kshakir is on it)
In the meantime I re-factored a bit the "Running" state of the EJEA to be able to understand what's going on and added more tests.


engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/job/EngineJobExecutionActor.scala
```diff
@@ -760,16 +803,28 @@ object EngineJobExecutionActor {
     def dockerImageUsed: Option[String]
     def withHashes(hashes: Option[Try[CallCacheHashes]]): ResponseData
   }
+  
+  // Only Successes and Failures are saved to the job store, not Aborts. Why ? Because. This could be an improvement AFAICT.
```
mcovarr: The advantage to this being you would know on restart to not even bother aborting?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/job/EngineJobExecutionActor.scala
```diff
@@ -263,47 +263,94 @@ class EngineJobExecutionActor(replyTo: ActorRef,
       // Can't write hashes for this job, but continue to wait for the copy response.
       stay using data.copy(hashes = Option(Failure(t)))
   }
+
```
mcovarr: all that follows is just a logical reorganization of the various EJEA cases or are there new cases or ?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/job/EngineJobExecutionActor.scala
```diff
@@ -760,16 +803,28 @@ object EngineJobExecutionActor {
...
     override def withHashes(hashes: Option[Try[CallCacheHashes]]) = copy(hashes = hashes)
   }
 
-  private[execution] case class NotSucceededResponseData(response: BackendJobExecutionResponse,
-                                                         hashes: Option[Try[CallCacheHashes]] = None) extends ResponseData {
+  private[execution] case class FailedResponseData(failedResponse: BackendJobFailedResponse,
+                                                      hashes: Option[Try[CallCacheHashes]] = None) extends ShouldBeSavedToJobStoreResponseData {
+    override def response = failedResponse
+    // Seems like we should be able to get the docker image used even if the job failed
+    override def dockerImageUsed = None
```
mcovarr: hmm yeah

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/job/EngineJobExecutionActor.scala
```diff
@@ -263,47 +263,94 @@ class EngineJobExecutionActor(replyTo: ActorRef,
...
+    // publish the hashes and terminate
+    case Event(hashes: CallCacheHashes, data: AbortedResponseData) =>
+      publishHashesToMetadata(Option(Success(hashes)))
+      // Our work is done, aborted jobs don't get saved to the job store
+      forwardAndStop(data.response)
+    // We're getting the hashes and the job has not completed yet, save them and stay where we are
     case Event(hashes: CallCacheHashes, data) =>
       addHashesAndStay(data, hashes)
+      
+    // Not sure why this is here but pre-existing so leaving it, there's nothing we can do with it anyway at this point
```
mcovarr: 😬

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/job/EngineJobExecutionActor.scala
Horneth: There's 1 new case that will wait for the hashes if `writeToCache` is on and the job failed. Before it would stop without waiting for the hashes. All the rest is the same, just more explicitly stated, specifically for the aborted case.

PR #3379, opened 2018-03-08
horizontalizing the db

mcovarr:


engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreEngineActor.scala
```diff
@@ -14,14 +15,23 @@ import cromwell.engine.workflow.workflowstore.WorkflowStoreActor._
...
 
 final case class WorkflowStoreEngineActor private(store: WorkflowStore, serviceRegistryActor: ActorRef, abortAllJobsOnTerminate: Boolean)
   extends LoggingFSM[WorkflowStoreActorState, WorkflowStoreActorData] with ActorLogging with WorkflowInstrumentation with CromwellInstrumentationScheduler with WorkflowMetadataHelper {
 
+  val cromwellId: Option[String] = ConfigFactory.load().as[Option[String]]("system.cromwell_id")
+
+  // Entries in the workflow store with a Cromwell ID different from this instance and whose heartbeat timestamps are
+  // older than now - heartbeatTtl are presumed abandoned and up for grabs.
+  val DefaultWorkflowStoreHeartbeatTtl = 1.hour
+  val heartbeatTtl: Duration = ConfigFactory.load().as[Option[Duration]]("system.workflow_heartbeat_ttl").getOrElse(DefaultWorkflowStoreHeartbeatTtl)
```
kshakir:
```scala
val heartbeatTtl: FiniteDuration = ConfigFactory.load()
  .getOrElse("system.workflow_heartbeat_ttl", DefaultWorkflowStoreHeartbeatTtl)
```

Also very ToL: Embedded use of `ConfigFactory.load()` works (the load is even cached by ConfigFactory so it's pretty efficient), but makes it hard(er) for specs to pass in their own test configuration. For the sub-project running the test it requires a custom `src/test/resources/application.conf`. In a lot of places I thought we were passing around global and "local" `Config` objects, but I don't see those around here. Food for future thought.

cjllanwarne:
> codecov/patch — 100% of diff hit (target 50%)

... nice!

database/sql/src/main/scala/cromwell/database/slick/tables/WorkflowStoreEntryComponent.scala
```diff
@@ -74,14 +76,29 @@ trait WorkflowStoreEntryComponent {
...
+        // locked such that readers are blocked since we will do an update subsequent to this select in the same
+        // transaction that we know will impact those readers.
+        //
+        // The current code only writes heartbeats on initial pickup, so any other Cromwell's workflows will appear
+        // to be abandoned if they were picked up before `heartbeatThreshold`.
+        if (row.cromwellId === cromwellId && row.heartbeatTimestamp.isEmpty) || // 1
+          (row.cromwellId.isEmpty && row.heartbeatTimestamp.isEmpty) || // 2
+          (row.cromwellId =!= cromwellId && row.heartbeatTimestamp < heartbeatThreshold) // 3
+        // This logic leaves a hole where if another Cromwell comes up, nulls out the heartbeats of the workflows
+        // it had in flight, then goes down and stays down without picking up some of those workflows, no other
```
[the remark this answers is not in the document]
mcovarr: The startup code nulls out the heartbeat so that this code will know the workflow is eligible for pickup. If the startup code wrote a current heartbeat then this code would wait `heartbeatTtl` before picking it up.

database/sql/src/main/scala/cromwell/database/slick/WorkflowStoreSlickDatabase.scala
```diff
@@ -49,29 +54,38 @@ trait WorkflowStoreSlickDatabase extends WorkflowStoreSqlDatabase {
...
     }
+
+    val displayNow = DateFormat.getDateTimeInstance.format(now)
+
     for {
-      // When fetched, the restarted flag is set back to false so we don't pick it up next time.
-      updateCount <- dataAccess.workflowStateAndRestartedForWorkflowExecutionUuid(workflowExecutionUuid).update((updateState, false))
-      _ <- assertUpdateCount(s"Update $workflowExecutionUuid to $updateState", updateCount, 1)
+      // When fetched, the heartbeat timestamp is set to now so we don't pick it up next time.
+      updateCount <- dataAccess.workflowStateAndHeartbeatTimestampForWorkflowExecutionUuid(workflowExecutionUuid).update((updateState, Option(now)))
```
Horneth: shouldn't this set the `cromwellId` too ?

mcovarr: yes it should. 😬

PR #3389, opened 2018-03-09
Keep timers running

Horneth:
Forgot to kick off a new timer when the last one arrives


cjllanwarne:
The travis failure looks unrelated but sort of scary... perhaps one for the spreadsheet...

Horneth:
@cjllanwarne I added it to the doc

PR #3390, opened 2018-03-09
 Short-circuit boolean OR and AND

### LINKED ISSUE #3384: && operator doesn't short-circuit, possibly || as well
Reported by @asmirnov239

Cromwell 30.2 doesn't seem to short-circuit for logical operators so that the following fails to evaluate when `optional_int` is undefined:

```wdl 
Integer? optional_int
if (defined(optional_int) && select_first([optional_int]) == 2 ) {
#expresssions to do if optional_int is defined and equal to 2
}
```

expected behavior: if `optional_int` is undefined, if statement evaluates to false. 
actual behavior: cromwell fails and stops with the error: 

>Evaluating defined(gc_low_high_filter_params) && select_first([gc_low_high_filter_params]) == 2 failed: select_first failed. All provided values were empty.

While it hasn't been tested, I presume that cromwell doesn't short-circut for the `||` operator...but it should!

cjllanwarne:
Closes #3384


cjllanwarne:
Merging despite known sbt failure.

PR #3393, opened 2018-03-09
Sam submit whitelist checking.

kshakir:
Switched the SamDenialResponse from "I don't know you" to "I know who your are and do not like you".

### DIFF
--- a/CromIAM/src/main/scala/cromiam/sam/SamClient.scala
+++ b/CromIAM/src/main/scala/cromiam/sam/SamClient.scala
@@ -3,15 +3,18 @@ package cromiam.sam
 import akka.actor.ActorSystem
 import akka.event.LoggingAdapter
 import akka.http.scaladsl.Http
+import akka.http.scaladsl.marshallers.sprayjson.SprayJsonSupport._
 import akka.http.scaladsl.model._
+import akka.http.scaladsl.unmarshalling.Unmarshal
 import akka.stream.ActorMaterializer
 import akka.util.ByteString
-
-import scala.concurrent.{ExecutionContextExecutor, Future}
 import com.softwaremill.sttp._
 import cromiam.auth.{Collection, User}
 import cromiam.sam.SamClient._
+import cromiam.sam.SamResourceJsonSupport._
 import cromiam.server.status.StatusCheckedSubsystem
+
+import scala.concurrent.{ExecutionContextExecutor, Future}
 
 /*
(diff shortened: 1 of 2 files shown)


kshakir:
**DO NOT MERGE** before https://github.com/broadinstitute/sam/pull/113 and related changes documented in the team drive.

CromIAM/src/main/scala/cromiam/sam/SamClient.scala
```diff
@@ -24,11 +27,24 @@ class SamClient(scheme: String, interface: String, port: Int, log: LoggingAdapte
...
+      headers = List[HttpHeader](user.authorization)
+    )
 
+    for {
+      response <- Http().singleRequest(request)
+      whitelisted <- response.status match {
+        case StatusCodes.OK => Unmarshal(response.entity).to[String].map(_.toBoolean)
+        case _ => Future.successful(false)
+      }
+      _ = if (!whitelisted) log.error("Submit Access Denied for user {}", user.userId)
```
cjllanwarne: ToL: `user.userId` is information that they gave us, right? Not that we've dug up from SAM for them?

CromIAM/src/main/scala/cromiam/sam/SamClient.scala
```diff
@@ -98,15 +114,16 @@ class SamClient(scheme: String, interface: String, port: Int, log: LoggingAdapte
...
   private lazy val samBaseCollectionUri = s"$samBaseResourceUri/workflow-collection"
+  private lazy val samSubmitWhitelistUri = s"$samBaseResourceUri/caas/submit/action/get_whitelist"
 
 }
 
 object SamClient {
   case object SamDenialException extends Exception("Access Denied")
-  val SamDenialResponse = HttpResponse(status = StatusCodes.Unauthorized, entity = SamDenialException.getMessage)
+
+  val SamDenialResponse = HttpResponse(status = StatusCodes.Forbidden, entity = SamDenialException.getMessage)
```
cjllanwarne: ToL: Oops, that comment above should probably have been here: we should make sure we don't leak anything from SAM in the exception message that the bad user didn't already tell us.

CromIAM/src/main/scala/cromiam/sam/SamClient.scala
geoffjentry: Look at how it is being called

PR #3396, opened 2018-03-12
A few more I/O configurable values

Horneth:


core/src/main/resources/reference.conf
```diff
@@ -434,7 +442,11 @@ load-control {
   call-cache-write = 10000
   key-value-read = 10000
   key-value-write = 10000
-  io = 10000
+  # The I/O queue has the specificity to be bounde. This sets its size
```
mcovarr: However this misspelling is not bounded.

core/src/main/resources/reference.conf
```diff
@@ -434,7 +442,11 @@ load-control {
...
+  io-queue-size = 10000
+  # If the I/O queue is full, subsequent requests are rejeced and to be retried later.
```
mcovarr: The spelling of `rejeced` has also rejected and should be retried.

PR #3404, opened 2018-03-13
Increase gauge processing throughput

Horneth:
While testing C31 perf I noticed the statsd service can get very slow which translates in metrics being largely out of dates.
I traced this to an issue with how gauges are being updated.


geoffjentry:
:+1:

PR #3407, opened 2018-03-14
Victory by default

mcovarr:
Apply defaults where appropriate for WorkflowStepInputs, straighten out some handling of optionals.


cwl/src/main/scala/cwl/Workflow.scala
```diff
@@ -87,29 +86,33 @@ case class Workflow private(
...
-        val womType = womTypeForInputParameter(inputParameter).get
-
-        val womValue = inputParameter.default.get.fold(InputParameter.DefaultToWomValuePoly).apply(womType).toTry.get
-        OptionalGraphInputNodeWithDefault(WomIdentifier(parsedInputId, inputParameter.id), womType, ValueAsAnExpression(womValue), parsedInputId)
-      case input =>
-        val parsedInputId = FileAndId(input.id).id
-
-        RequiredGraphInputNode(WomIdentifier(parsedInputId, input.id), womTypeForInputParameter(input).get, parsedInputId)
+    val externalGraphInputNodes: Set[ExternalGraphInputNode] = inputs.map { wip =>
+      val womType = wip.`type`.head.fold(MyriadInputTypeToWomType)
```
danbills: Curious why no longer using `womTypeForInputParameter` ?

cwl/src/main/scala/cwl/Workflow.scala
```diff
@@ -87,29 +86,33 @@ case class Workflow private(
...
+      val parsedInputId = FileAndId(wip.id).id
+      val womId = WomIdentifier(parsedInputId, wip.id)
+
+      def optionalWithDefault(memberType: WomType): OptionalGraphInputNodeWithDefault = {
+        val defaultValue = wip.default.get.fold(InputParameter.DefaultToWomValuePoly).apply(womType).toTry.get
```
danbills: I see that default is always defined but is `DefaultToWomValuePoly` result always going to be there?  If not can this whole thing become a `Checked`?

cwl/src/main/scala/cwl/Workflow.scala
mcovarr: uh not sure... going to try to put that back

PR #3411, opened 2018-03-14
Explicitly wait for a result from docker run

danbills:
This approach allows docker to re-connect to container that it had been detached from in order to retrieve the exit status. 

See attached issue for background.


Horneth:
Hmm on second thought I think I'm wrong, there's the `run-in-background` conf that probably makes it ok

danbills:
@Horneth Per our discussion this morning, it was already a blocking call, this just makes retrieving the return code more robust in case of disconnect.

PR #3413, opened 2018-03-14
Taskless engine functions

cjllanwarne:
- Ports an existing draft 2 test of taskless engine functions into draft 3.
- Also adds implementation or placeholders for all of the type evaluation and value evaluation for engine functions (so apologies about the high line count!)

- [x] Rebase on develop after #3403 

Red thumb requirement triggered by the addition of a `WomTypeCoercer` typeclass. I added it so that I didn't have to keep casting results but I think this could be even useful-er if CWL and WDL coercions start to (or need to) diverge.

### DIFF
--- a/wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/linking/expression/values/EngineFunctionEvaluators.scala
+++ b/wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/linking/expression/values/EngineFunctionEvaluators.scala
@@ -1,22 +1,229 @@
 package wdl.draft3.transforms.linking.expression.values
 
+import cats.syntax.traverse._
 import cats.syntax.validated._
+import cats.instances.list._
 import common.validation.ErrorOr._
 import common.validation.ErrorOr.ErrorOr
+import common.validation.Validation._
 import wdl.model.draft3.elements.ExpressionElement._
 import wdl.model.draft3.graph.expression.ValueEvaluator
 import wdl.model.draft3.graph.expression.ValueEvaluator.ops._
+import wdl.shared.transforms.evaluation.values.EngineFunctions
 import wom.expression.IoFunctionSet
-import wom.types.{WomArrayType, WomIntegerType}
-import wom.values.{WomArray, WomInteger, WomValue}
+import wom.types._
+import wom.values.WomArray.WomArrayLike
+import wom.values.{WomArray, WomFloat, WomInteger, WomPair, WomString, WomValue}
+import wom.types.coercion.ops._
+import wom.types.coercion.defaults._
+import wom.types.coercion.WomTypeCoercer
 
 object EngineFunctionEvaluators {
+  implicit val readLinesFunctionEvaluator: ValueEvaluator[ReadLines] = new ValueEvaluator[ReadLines] {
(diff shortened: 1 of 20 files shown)


wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/linking/expression/values/EngineFunctionEvaluators.scala
```diff
@@ -1,22 +1,229 @@
...
+  implicit val writeJsonFunctionEvaluator: ValueEvaluator[WriteJson] = new ValueEvaluator[WriteJson] {
+    override def evaluateValue(a: WriteJson, inputs: Map[String, WomValue], ioFunctionSet: IoFunctionSet): ErrorOr[WomValue] = ???
+  }
+
   implicit val rangeFunctionEvaluator: ValueEvaluator[Range] = new ValueEvaluator[Range] {
     override def evaluateValue(a: Range, inputs: Map[String, WomValue], ioFunctionSet: IoFunctionSet): ErrorOr[WomValue] = {
-      a.param.evaluateValue(inputs, ioFunctionSet) flatMap {
-        case WomInteger(i) => WomArray(WomArrayType(WomIntegerType, guaranteedNonEmpty = i > 0), (0 until i).map(WomInteger)).validNel
-        case other => s"Could not evaluate $a: Expected integer argument but got ${other.womType}".invalidNel
+      processValidatedSingleValue[WomInteger](a.param.evaluateValue(inputs, ioFunctionSet)) { integer =>
```
cjllanwarne: This is an example of using the coercion typeclass.

We want a `WomInteger` and there's a `WomTypeCoercer[WomInteger]` in scope, so the magic can happen.

If we wanted different coercions in different languages (or language versions), we just need to import the coercers from different scopes.

wom/src/main/scala/wom/types/coercion/defaults/package.scala
```diff
@@ -0,0 +1,29 @@
...
+import wom.types._
+import wom.values._
+
+import scala.language.implicitConversions
+import scala.reflect.ClassTag
+
+package object defaults {
+  implicit val womIntegerCoercer: WomTypeCoercer[WomInteger] = defaultCoercionForType[WomInteger](WomIntegerType)
+  implicit val womFloatCoercer: WomTypeCoercer[WomFloat] = defaultCoercionForType[WomFloat](WomFloatType)
+  implicit val womStringCoercer: WomTypeCoercer[WomString] = defaultCoercionForType[WomString](WomStringType)
```
cjllanwarne: The default coercers tie together the `WomValue` type and the coercion function on their respective `WomType`s

danbills:
Cool!  I've been thinking about how to approach this too.  I looked at [shapeless' Typeable](https://github.com/milessabin/shapeless/blob/master/core/src/main/scala/shapeless/typeable.scala#L28) (and [docs](https://github.com/milessabin/shapeless/wiki/Feature-overview:-shapeless-2.0.0#type-safe-cast)) which co-exists with [ValueTypeable](https://github.com/milessabin/shapeless/blob/master/core/src/main/scala/shapeless/typeable.scala#L53) (declare from -> to types).

Dunno if it has any advantages over rolling your own but it's worth a look to see if there are any.

mcovarr:
re [diverging coercions](https://github.com/broadinstitute/cromwell/pull/3430)

PR #3419, opened 2018-03-15
Orderly retreat

danbills:
The reasons for doing this are chiefly:

* We now have centaur which tests much of this for us
* Doing a giant diff of this vs centaur is untenable
* It doesn't test JES
* We plan to reproduce the valuable single workflow runner aspect of this in a travis build (coming soon)
* It is flaky as hell


geoffjentry:
Why doesn't this get rid of `CromwellTestKitSpec` itself, or did I just miss it?

1o9ejc

danbills:
It does, you missed it

geoffjentry:
Well then, I go back to the sentiment of the fine image I added

server/src/test/scala/cromwell/engine/workflow/lifecycle/execution/ejea/EjeaFetchingCachedOutputsFromDatabaseSpec.scala
```diff
@@ -1,85 +0,0 @@
-package cromwell.engine.workflow.lifecycle.execution.ejea
```
Horneth: Do we have to get rid of those too ? Those EJEA unit test despite being hard to read are valuable I think because they test a lot of corner cases that can't be tested through centaur.

geoffjentry: @Horneth I don't think there should be any reason for a test spec to have the "run a workflow with full cromwell" though, so those tests should be representable via some other mechanism. Do you disagree in this case, or is your complaint more like w/ the single workflow where we need to find another way to test that?

Horneth: Those `EjeaDoingSomething` tests are not running workflows, they're testing how the EJEA responds to different messages while being in specific states. It tests that we didn't forget a `case ...` in the FSM and that things won't stay stuck.

cjllanwarne: it's possible that methods unrelated to "run a WDL" got bundled into `CromwellTestKitSpec` because it sounded a bit like `AkkaTestKit`

server/src/test/scala/cromwell/MultipleFilesWithSameNameWorkflowSpec.scala
```diff
@@ -1,21 +0,0 @@
-package cromwell
-
-import akka.testkit._
-import cromwell.util.SampleWdl
-import wom.values.WomString
-
-
-class MultipleFilesWithSameNameWorkflowSpec extends CromwellTestKitWordSpec {
```
cjllanwarne: Flagging this as "might not be in centaur"

server/src/test/scala/cromwell/engine/WorkflowStoreActorSpec.scala
```diff
@@ -1,184 +0,0 @@
...
-import cromwell.util.SampleWdl.HelloWorld
-import cromwell.{CromwellTestKitSpec, CromwellTestKitWordSpec}
-import org.scalatest.concurrent.Eventually
-import org.scalatest.{BeforeAndAfter, Matchers}
-import org.specs2.mock.Mockito
-
-import scala.concurrent.duration._
-import scala.language.postfixOps
-
-class WorkflowStoreActorSpec extends CromwellTestKitWordSpec with Matchers with BeforeAndAfter with Mockito with Eventually {
```
cjllanwarne: This looks like another case of "uses functions" rather than "runs a workflow"

server/src/test/scala/cromwell/jobstore/JobStoreServiceSpec.scala
```diff
@@ -1,71 +0,0 @@
...
-
-import scala.concurrent.duration._
-import scala.language.postfixOps
-
-object JobStoreServiceSpec {
-  val MaxWait = 5 seconds
-  val EmptyExpression = PlaceholderWomExpression(Set.empty, WomStringType)
-}
-
-class JobStoreServiceSpec extends CromwellTestKitWordSpec with Matchers with Mockito {
```
cjllanwarne: This one also looks more Akka TestKit-y

server/src/test/scala/cromwell/jobstore/JobStoreWriterSpec.scala
```diff
@@ -1,138 +0,0 @@
...
-import cromwell.jobstore.JobStore.{JobCompletion, WorkflowCompletion}
-import cromwell.jobstore.JobStoreActor.{JobStoreWriteSuccess, JobStoreWriterCommand, RegisterJobCompleted, RegisterWorkflowCompleted}
-import org.scalatest.{BeforeAndAfter, Matchers}
-import wom.graph.GraphNodePort.OutputPort
-
-import scala.concurrent.duration._
-import scala.concurrent.{ExecutionContext, Future, Promise}
-import scala.language.postfixOps
-
-class JobStoreWriterSpec extends CromwellTestKitWordSpec with Matchers with BeforeAndAfter {
```
cjllanwarne: Another `AkkaTestKit`?

server/src/test/scala/cromwell/subworkflowstore/SubWorkflowStoreSpec.scala
```diff
@@ -1,102 +0,0 @@
...
-
-import scala.concurrent.duration._
-import scala.language.postfixOps
-
-object SubWorkflowStoreSpec {
-  val MaxWait = 5 seconds
-  val EmptyExpression = WdlExpression.fromString(""" "" """)
-}
-
-class SubWorkflowStoreSpec extends CromwellTestKitWordSpec with Matchers with Mockito {
```
cjllanwarne: This one too?

server/src/test/scala/cromwell/engine/workflow/lifecycle/execution/ejea/EjeaFetchingCachedOutputsFromDatabaseSpec.scala
geoffjentry: How about we sit down and go through this, come up with a scheme for preserving anything needing preserving via a means other than CTKS and then decide on a case by case basis if it makes sense to do that now or bit by bit

danbills:
closing this reckless PR in favor of an even more orderly, measured retreat

PR #3422, opened 2018-03-15
Memory coercion and coproduct simpletonization

Horneth:


backend/src/main/scala/cromwell/backend/validation/MemoryValidation.scala
```diff
@@ -83,7 +83,7 @@ class MemoryValidation(attributeName: String = RuntimeAttributesKeys.MemoryKey)
 
   override def key = attributeName
 
-  override def coercion = Seq(WomIntegerType, WomStringType)
+  override def coercion = Seq(WomIntegerType, WomStringType, WomLongType)
```
cjllanwarne: Is the order important here? If so, I'd put Integer first, then Long, then String

PR #3428, opened 2018-03-16
Fix Home environment variable

danbills:
A bug was producing export HOME="$HOME$HOME$HOME..." as seen in issue


geoffjentry:
Can you also put this on hot fix?1

PR #3430, opened 2018-03-16
Parse scientific notation on numbers, some coercions from filled optionals.

mcovarr:
Still driving on the Wash U workflow, this is a pregull checkpoint with some ~hacks~ work I did for blockers:

We round trip inputs YAML through Circe which uses its own "Big" number types that stringify with scientific notation. Our existing WomInteger / WomLong code didn't deal with that so the changes here add some flexibility.

CWL explicitly allows filled optional to non-optional assignments with warnings, with an error at runtime if it turns out the optional wasn't actually filled. I expect some discussion on this 🙂

### DIFF
--- a/cwl/src/main/scala/cwl/CwlExpressionCommandPart.scala
+++ b/cwl/src/main/scala/cwl/CwlExpressionCommandPart.scala
@@ -71,8 +71,9 @@ abstract class CommandLineBindingCommandPart(commandLineBinding: CommandLineBind
     }
     
     def processValue(womValue: WomValue): List[String] = womValue match {
-      case WomOptionalValue(_, Some(womInnerValue)) => processValue(valueMapper(womInnerValue))
-      case _: WomString | _: WomInteger | _: WomFile => handlePrefix(valueMapper(womValue).valueString)
+      case WomOptionalValue(_, Some(value)) => processValue(valueMapper(value))
+      case WomOptionalValue(_, None) => List.empty
+      case _: WomString | _: WomInteger | _: WomFile | _: WomLong | _: WomFloat => handlePrefix(valueMapper(womValue).valueString)
       // For boolean values, use the value of the boolean to choose whether to print the prefix or not
       case WomBoolean(false) => List.empty
       case WomBoolean(true) => prefixAsList
@@ -99,7 +100,7 @@ abstract class CommandLineBindingCommandPart(commandLineBinding: CommandLineBind
         case _ => prefixAsList
       }
       case _: WomObjectLike => prefixAsList
-      case _ => List.empty
+      case w => throw new RuntimeException(s"Unhandled CwlExpressionCommandPart value '$w' of type ${w.womType.toDisplayString}")
     }
 
     evaluatedWomValue map { v => processValue(v) map applyShellQuote map (InstantiatedCommand(_)) } toValidated
--- a/cwl/src/main/scala/cwl/preprocessor/CwlPreProcessor.scala
+++ b/cwl/src/main/scala/cwl/preprocessor/CwlPreProcessor.scala
@@ -10,8 +10,9 @@ import common.validation.Validation._
 import cwl.command.ParentName
 import cwl.preprocessor.CwlPreProcessor._
 import cwl.{CwlDecoder, FileAndId, FullyQualifiedName}
-import io.circe.{Json, JsonObject}
+import io.circe.Json.JNumber
 import io.circe.optics.JsonPath._
+import io.circe.{Json, JsonNumber, JsonObject}
 import mouse.all._
 import org.slf4j.LoggerFactory
 
@@ -138,10 +139,33 @@ class CwlPreProcessor(saladFunction: BFile => Parse[String] = saladCwlFile) {
     jsonObject = prefixObject(mappingFunction),
     jsonArray = arr => Json.arr(arr.map(mapFilesAndDirectories(mappingFunction)): _*)
   )
+
+  private def mapNumbers(json: Json): Json = {
+    // Circumvent Circe's scientific format for numbers: convert to a JSON String without exponential notation.
+    def nonScientificNumberFormatting(jsonNumber: JsonNumber): Json = {
+      val conversions = Stream[JsonNumber => Option[Any]](
+        _.toBigInt.map(_.longValue()),
+        _.toBigDecimal.map(_.doubleValue()),
+        Function.const(Option("null")))
+
+      // The `get` is safe because `Option("null")` guarantees a match even if the other two Stream elements
+      // do not satisfy the predicate.
+      conversions.map(_.apply(jsonNumber)).find(_.isDefined).flatten.get.toString |> Json.fromString
+    }
+
+    json.fold(
+      jsonNull = json,
+      jsonBoolean = _ => json,
+      jsonNumber = nonScientificNumberFormatting,
+      jsonString = _ => json,
+      jsonObject = _.mapValues(mapNumbers) |> Json.fromJsonObject,
+      jsonArray = _.map(mapNumbers) |> Json.fromValues
+    )
+  }
   
   def preProcessInputFiles(inputContent: String, mappingFunction: String => String): Parse[String] = for {
     parsed <- parseYaml(inputContent)
-    mapped = mapFilesAndDirectories(mappingFunction)(parsed)
+    mapped = parsed |> mapFilesAndDirectories(mappingFunction) |> mapNumbers
   } yield mapped.printCompact
 
   /**
--- a/wom/src/main/scala/wom/types/WomBooleanType.scala
+++ b/wom/src/main/scala/wom/types/WomBooleanType.scala
@@ -1,7 +1,7 @@
 package wom.types
 
 import spray.json.{JsBoolean, JsString}
-import wom.values.{WomBoolean, WomValue}
+import wom.values.WomBoolean
 
 import scala.util.{Success, Try}
 
--- a/wom/src/main/scala/wom/types/WomIntegerLike.scala
+++ b/wom/src/main/scala/wom/types/WomIntegerLike.scala
@@ -1,0 +1,8 @@
+package wom.types
+
+
+object WomIntegerLike {
+  implicit class EnhancedLong(val long: Long) extends AnyVal {
+    def inIntRange: Boolean = long >= Int.MinValue && long <= Int.MaxValue
+  }
+}
--- a/wom/src/main/scala/wom/types/WomIntegerType.scala
+++ b/wom/src/main/scala/wom/types/WomIntegerType.scala
@@ -1,6 +1,7 @@
 package wom.types
 
 import spray.json.{JsNumber, JsString}
+import wom.types.WomIntegerLike._
 import wom.values.{WomInteger, WomLong, WomString}
 
 import scala.util.{Success, Try}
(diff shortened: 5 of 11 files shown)


wom/src/main/scala/wom/types/WomBooleanType.scala
```diff
@@ -10,6 +10,9 @@ case object WomBooleanType extends WomPrimitiveType {
 
   override protected def coercion = {
     case b: Boolean => WomBoolean(b)
+    case WomOptionalValue(WomBooleanType, Some(WomBoolean(b))) =>
+      println("Warning: you just coerced a filled Boolean optional to a non-optional Boolean and you should feel bad about yourself")
```
geoffjentry: 🤔

mcovarr: Similar in spirit to the cwltool error, minus the helpful context ("This happened with input 'foo'") and the use of a proper logging facility since there isn't one in scope here.

wom/src/main/scala/wom/types/WomBooleanType.scala
```diff
@@ -10,6 +10,9 @@ case object WomBooleanType extends WomPrimitiveType {
 
   override protected def coercion = {
     case b: Boolean => WomBoolean(b)
+    case WomOptionalValue(WomBooleanType, Some(WomBoolean(b))) =>
```
Horneth: If we're going to do this should it be everywhere and not just WomBoolean ?
We should check with Commodore that this is not going to break WDL behavior too. The typeclass coercion thingy does seem like the right thing to do

cjllanwarne:
I'm anti this if this breaks WDL's typesafety, but I suspect it won't actually since you didn't touch the `isCoerceableFrom` method - but that also makes me wonder if this really works?

cjllanwarne:
Actually, if this works (because WOM doesn't ever check `isCoerceableFrom`), maybe all we need is the set of coercion functions and leave it up to individual languages to implement their own `isCoerceableFrom` checks at graph-building time.

Because although they might have different coersions allowed, I can't imagine the coercion methods themselves being different (is that too naive?)

Horneth:
@cjllanwarne are we talking about coercing from `WomValue of WomTypeA` to `WomTypeB` or `Any` to `WomTypeB`
I can see the first one being fairly uncontroversial but maybe the second one could need to be language specific ?

mcovarr:
@cjllanwarne I don't think this broke WDL's typesafety for the reasons you state. Also it sounds like we're all thinking similar things re dynamic versus static typechecks and the "coercing possibly anything to a WomType" versus "coercing one WomType to another".  The filled-optional to non-optional coercion does work, the workflow doesn't even get started without it. 🙂

wom/src/main/scala/wom/types/WomBooleanType.scala
mcovarr: ...and actually the lack of at least the scientific notation parsing bit of this stuff in float-land seems to be responsible for those broken command lines so I'll be fixing that rn 😄

cwl/src/main/scala/cwl/CwlExpressionCommandPart.scala
```diff
@@ -69,8 +69,8 @@ abstract class CommandLineBindingCommandPart(commandLineBinding: CommandLineBind
     }
     
     def processValue(womValue: WomValue): List[String] = womValue match {
-      case WomOptionalValue(_, Some(womValue)) => processValue(valueMapper(womValue))
-      case _: WomString | _: WomInteger | _: WomFile => handlePrefix(valueMapper(womValue).valueString)
+      case WomOptionalValue(_, Some(value)) => processValue(valueMapper(value))
+      case _: WomString | _: WomInteger | _: WomFile | _: WomLong | _: WomFloat => handlePrefix(valueMapper(womValue).valueString)
```
mcovarr: long and float inputs were being silently omitted from the command line

wom/src/main/scala/wom/types/WomFloatLike.scala
```diff
@@ -0,0 +1,14 @@
+package wom.types
+
+/** WomFloat actually wraps a double. Mimicking the structure of WomIntegerLike JIC we need to make the
+  * distinction between float and double that CWL seems to be suggesting we should make. */
+object WomFloatLike {
+  private def parseString(string: String): Double = {
+    // More flexible floating point parsing that can handle the sort of scientific notation Circe likes to produce.
+    java.lang.Double.valueOf(string)
```
mcovarr: needed to add this for `WomFloat`s too

wom/src/main/scala/wom/types/WomFloatType.scala
```diff
@@ -13,9 +14,12 @@ case object WomFloatType extends WomPrimitiveType {
...
     case n: JsNumber => WomFloat(n.value.doubleValue())
     case f: WomFloat => f
-    case s: String => WomFloat(s.toDouble)
-    case s: JsString => WomFloat(s.value.toDouble)
-    case s: WomString => WomFloat(s.value.toDouble)
+    case s: String => WomFloat(s.asDouble)
+    case s: JsString => WomFloat(s.value.asDouble)
+    case s: WomString => WomFloat(s.value.asDouble)
+    case WomOptionalValue(WomFloatType, Some(d)) =>
+      println(s"My stars! That was a filled optional Float you just converted to a Float!")
```
mcovarr: didn't actually need the filled-optional thing for floats yet but as long as I was in here...

wom/src/main/scala/wom/types/WomBooleanType.scala
cjllanwarne: If this PR goes beyond "DO NOT MERGE", I probably would tone down the judgy-ness though...

mcovarr: gotcha, `LogLevel.JUDGY`.

cwl/src/main/scala/cwl/CwlExpressionCommandPart.scala
cjllanwarne: TOL is there any way to make that non-silent for future omissions of the same sort?

wom/src/main/scala/wom/types/WomFloatLike.scala
```diff
@@ -0,0 +1,19 @@
+package wom.types
+
+/** WomFloat actually wraps a double. Mimicking the structure of WomIntegerLike JIC we need to make the
+  * distinction between float and double that CWL seems to be suggesting we should make. */
+object WomFloatLike {
+  private def parseString(string: String): Double = {
```
cjllanwarne: Not `ErrorOr[Double]`?

wom/src/main/scala/wom/types/WomIntegerType.scala
```diff
@@ -12,11 +13,19 @@ case object WomIntegerType extends WomPrimitiveType {
...
     case i: WomInteger => i
-    case WomLong(i) if (i >= Int.MinValue && i <= Int.MaxValue) => WomInteger(i.toInt)
-    case WomLong(i) => throw new RuntimeException(s"Tried to convert a long value $i into an integer but it was outside the bounds of acceptable Integers, namely ${Int.MinValue} <-> ${Int.MaxValue}")
-    case s: WomString => WomInteger(s.value.toInt)
-    case s: String => WomInteger(s.toInt)
-    case s: JsString => WomInteger(s.value.toInt)
+    case WomLong(i) if i.inIntRange => WomInteger(i.toInt)
+    case WomLong(i) => throw new RuntimeException(
+      s"Tried to convert a Long value $i into an Int but it was outside the bounds of acceptable Ints, namely ${Int.MinValue} <-> ${Int.MaxValue}")
+    case WomOptionalValue(WomIntegerType, Some(i)) => i
```
cjllanwarne: Could you recurse in the case of a filled optional, rather than enumerating all of the possible cases? (it should be impossible but what if we accidentally end up with a `Int??`)

cjllanwarne: Also I'd rather ignore what type the optional thinks it is, and match on the type of the value, eg
```
case WomOptionalValue(_, Some(i: WomInteger)) => i
```

cwl/src/main/scala/cwl/CwlExpressionCommandPart.scala
mcovarr: good idea

wom/src/main/scala/wom/types/WomIntegerType.scala
mcovarr: works for me

mcovarr: I can try

wom/src/main/scala/wom/types/WomFloatLike.scala
mcovarr: If I understand what you're saying correctly, `ErrorOr`ing `coerce` would be a pretty significant change from the existing structure

cwl/src/main/scala/cwl/CwlExpressionCommandPart.scala
```diff
@@ -80,7 +81,7 @@ abstract class CommandLineBindingCommandPart(commandLineBinding: CommandLineBind
         case _ => prefixAsList
       }
       case _: WomObjectLike => prefixAsList
-      case _ => List.empty
+      case w => throw new RuntimeException(s"Unhandled value: $w")
```
mcovarr: now explody on unhandled values rather than silently failing

wom/src/main/scala/wom/types/WomIntegerLike.scala
```diff
@@ -0,0 +1,24 @@
...
+
+object WomIntegerLike {
+  private def parseString(string: String): Long = {
+    if (string.contains("e") || string.contains("E")) {
+      // More flexible integer parsing that can handle the sort of scientific notation Circe likes to produce
+      // but is not as precise as `toLong` so only use if needed.
+      java.lang.Double.valueOf(string).longValue()
+    } else {
+      string.toLong
+    }
```
mcovarr: Yeah the gross got grosser here. There was a loss of precision in `Double.valueOf`ing a String unnecessarily that did not escape the notice of one of our unit tests. Consequently this is now more defensive and only invokes `Double.valueOf` if it appears scientific notation might be in use.

wom/src/main/scala/wom/types/WomType.scala
```diff
@@ -33,6 +33,7 @@ trait WomType {
       case _ if !coercion.isDefinedAt(any) => Failure(new IllegalArgumentException(
         s"No coercion defined from '${ScalaRunTime.stringOf(any, 3)}' of type" +
           s" '${Option(any.getClass.getCanonicalName).getOrElse(any.getClass.getName)}' to '$toDisplayString'."))
+      case WomOptionalValue(_, Some(v)) => coerceRawValue(v)
```
mcovarr: Pulled all the optional handling into `WomType.coerceRawValue` so it didn't need to be duplicated in the subtypes.

cwl/src/main/scala/cwl/CwlExpressionCommandPart.scala
cjllanwarne: Use either `.toDisplayString` or `.womType.toDisplayString`. I'd prefer the second since this is a "you gave me the wrong type" problem

wom/src/main/scala/wom/types/WomIntegerLike.scala
mcovarr: This is bugging me too. The way this code has "evolved" I now think it would be cleaner to do this with Circe's `fold` back in in the `CwlPreProcessor`; let me see if I can get that to work.

cwl/src/main/scala/cwl/preprocessor/CwlPreProcessor.scala
```diff
@@ -138,10 +139,33 @@ class CwlPreProcessor(saladFunction: BFile => Parse[String] = saladCwlFile) {
     jsonObject = prefixObject(mappingFunction),
     jsonArray = arr => Json.arr(arr.map(mapFilesAndDirectories(mappingFunction)): _*)
   )
+
```
mcovarr: @cjllanwarne moved the number reformatting back to the CwlPreProcessor

PR #3433, opened 2018-03-19
Don't duplicate scatter nodes

Horneth:


cwl/src/main/scala/cwl/WorkflowStep.scala
```diff
@@ -197,12 +197,16 @@ case class WorkflowStep(
 
     def typedRunInputs: Map[String, Option[MyriadInputType]] = run.fold(RunToInputTypeMap).apply(parentName)
 
+    def allIdentifiersRecursively(nodes: Set[GraphNode]): Set[WomIdentifier] = nodes.flatMap({
+      case WorkflowCallNode(identifier, _, _, _, _) => Set(identifier)
+      case CommandCallNode(identifier, _, _, _) => Set(identifier)
+      case ExpressionCallNode(identifier, _, _, _) => Set(identifier)
+      case scatter: ScatterNode => allIdentifiersRecursively(scatter.innerGraph.nodes)
```
cjllanwarne: Maybe a comment along the lines of "valid because CWL tasks can force us to wrap them into a scatter during construction, but the lookups are to the original task name"?

cjllanwarne:
👍

PR #3434, opened 2018-03-19
Don't do the weird unzip thing

### LINKED ISSUE #3297: Fix import zip heuristic to determine root directory
When being passed an import zip file, Cromwell behavior is currently as follows:

Lists the content of the zip, if the first element encountered is a directory, this directory is considered the root of imports from which imported files in the workflow are to be resolved.
This means that in the following case:

```
Archive:  imports.zip
  Length     Date   Time    Name
 --------    ----   ----    ----
        0  02-20-18 11:35   tasks/
     4793  02-20-18 11:35   tasks/alignment.wdl
    10530  02-20-18 11:35   tasks/bam_processing.wdl
        6  02-22-18 14:11   lost_import.wdl
```

`lost_import.wdl` will not be found because the root for import resolving will be `tasks`.

I believe this behavior exists to help users that inadvertently zipped the directory containing their imports instead of placing them at the root of the zip.
While maybe convenient this is IMO an awkward special case:

```
Archive:  imports.zip
  Length     Date   Time    Name
 --------    ----   ----    ----
        0  02-20-18 11:35   tasks/
     4793  02-20-18 11:35   tasks/alignment.wdl
    10530  02-20-18 11:35   tasks/bam_processing.wdl
```

incorrect: 
```wdl
  import "tasks/alignment.wdl" as my_tasks
```

correct:
```wdl
  import "alignment.wdl" as my_tasks
```

If we want to keep this behavior I think we should at least change the root if and only if the root of the zip is 1 directory.
I'd advocate to remove this behavior entirely but it would potentially be backward incompatible.

@geoffjentry you might have opinions on this

cjllanwarne:
Closes #3297.

If you unzip a zip file containing a directory, don't assume the user wants to reference things from within that directory.

Especially don't ignore the fact that there might be *other* directories in the zip file as well, which now cannot be accessed at all

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Cromwell Change Log
 
+## 32 Release Notes
+
+### Bug Fixes
+
+The imports zip no longer unpacks a single (arbitrary) internal directory if it finds one (or more). Instead, import statements should now be made relative to the base of the import zip root.
+
 ## 31 Release Notes
 
 * **Cromwell server**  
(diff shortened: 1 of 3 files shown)


Horneth:
Maybe a changelog entry for this ?

PR #3437, opened 2018-03-20
Conformance 108: Always shell quote in the absence of ShellCommandRequirement

mcovarr:


mcovarr:
Hopefully one day Travis will find true happiness. In the meantime I'll keep clicking restart.

PR #3438, opened 2018-03-20
Populate self for command input and fix optional secondary files

Horneth:


mcovarr:
No conformance tests enabled? I'm in a similar predicament over on my Wash U branch where I want to refactor a bunch of stuff but there are no tests to keep me honest. 😢

Horneth:
Yeah those seem like things I would have expected to be failing some conformance tests but apparently not... :/

cwl/src/main/scala/cwl/InputParameter.scala
```diff
@@ -122,6 +123,19 @@ object InputParameter {
...
+            obj.values.toList.traverse[ErrorOr, (String, WomValue)]({
+              case (key, value) => populateFiles(value).map(key -> _)
+            })
+              .map(_.toMap)
+              // transform to Either so we can flatMap
+              .toEither
+              // Validate new types are still valid w.r.t the object
+              .flatMap(WomObject.withTypeChecked(_, obj.womObjectTypeLike))
+              // re-transform to ErrorOr
+              .toValidated
```
kshakir: I don't think the method `withTypeChecked` is required.

```scala
// transform to Either so we can flatMap
.toEither
// Validate new types are still valid w.r.t the object
.flatMap(WomObject.withTypeChecked(_, obj.womObjectTypeLike))
// re-transform to ErrorOr
.toValidated
```

should be the same as

```scala
.flatMap(WomObject.withTypeErrorOr(_, obj.womObjectTypeLike))
```

cjllanwarne:
@mcovarr could we make a simple centaur test? Or even a PR into CWL’s suite?

cjllanwarne:
Because otherwise it’s going to be so easy to have accidental regressions on these more surprising edge cases in the future

mcovarr:
@cjllanwarne For my Wash U stuff I've added now added some tests. I'm not sure what's involved for a test on this.

Horneth:
I can add a test.
I'm all for making PRs to CWL conformance tests but it's going to increase the merge time of our PRs if we want to wait for it to be in the CWL repo. Also we'd need to unpin the hash for conformance test or update it every time..

cwl/src/main/scala/cwl/InputParameter.scala
Horneth: flatMapping `ErrorOr` requires the special `ShortCircuitingFlatMap` import whereas `Either` can naturally flatMap since 2.12 so I prefer using that even though it's a bit more verbose

cjllanwarne:
Thanks @Horneth!

src/bin/travis/resources/papi_conformance_expected_failures.txt
```diff
@@ -114,6 +114,6 @@
 114
 115
 116
-#117
+117
```
cjllanwarne: 😢

Horneth: It's not as bad as you think, it's a pre-processing conformance test problem where we don't prefix default values in the CWL with gcs paths.
We'll fix it at some point it's just not in the set of #1 priorities now 😄

Horneth:
Looks like I need a formal approval from `one_reviewer` for this one. I assume it means @cjllanwarne or @jbuabbud ?

PR #3443, opened 2018-03-21
Pass workflow inputs through the input value mapper

Horneth:
This properly processes workflow inputs by applying the value mapper to them once they have been bound to their `WomValue`.
For instance it enables support for the case where secondaryFiles are declared at the workflow input level but not the tool input level.
The real change is the addition of the `InputValueMapper` to `ExternalGraphInputNode`s and its use in `Executable.scala`
The rest of it is wiring the io functions through.

### DIFF
--- a/cwl/src/main/scala/cwl/CwlExecutableValidation.scala
+++ b/cwl/src/main/scala/cwl/CwlExecutableValidation.scala
@@ -10,9 +10,6 @@ import wom.executable.Executable
 import wom.executable.Executable.{InputParsingFunction, ParsedInputMap}
 import wom.expression.IoFunctionSet
 
-// WARNING! Because of 2.11 vs 2.12 incompatibilities, there are two versions of this file.
-// If you're making changes here, you'll also need to update ../../scala-2.12/cwl/CwlExecutableValidation.scala
-// (ExecutableValidation.scala has more info on why this was necessary)
 object CwlExecutableValidation {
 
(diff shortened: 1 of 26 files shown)


cjllanwarne:
I'm not sure I understand the description...

Can't we just have a task input fed by a workflow input (even if the CWL file doesn't say it, we know it's true)?

Horneth:
Discussed in person, a better way could be to do this work as a separate graph node. It would make the wiring a bit more complicated and the immediate gain isn't clear so I'll leave as is for now and we can revisit later when we refactor everything to achieve Cromwell singularity.

geoffjentry:
:+1:

cwl/src/main/scala/cwl/CwlExecutableValidation.scala
```diff
@@ -8,6 +8,7 @@ import io.circe.yaml
 import wom.callable.{ExecutableCallable, TaskDefinition}
 import wom.executable.Executable
 import wom.executable.Executable.{InputParsingFunction, ParsedInputMap}
+import wom.expression.IoFunctionSet
 
 // WARNING! Because of 2.11 vs 2.12 incompatibilities, there are two versions of this file.
 // If you're making changes here, you'll also need to update ../../scala-2.12/cwl/CwlExecutableValidation.scala
```
mcovarr: Preexisting but this comment appears to now be out of place?

PR #3439, opened 2018-03-20
Draft 3 size engine function

cjllanwarne:
Demonstrates the working of openWDL PR https://github.com/openwdl/wdl/pull/169

- [x] Rebase onto develop after 3432

PR #3444, opened 2018-03-22
Default input overriding

cjllanwarne:
- [x] Rebase on develop after #3439

Implements the ability to override inputs, even if they depend on upstream nodes.

Another trick on the conversion into WOM. In this case we simplify/expand this:
```wdl
input {
  Int b = a
}
Int a = 55
```

Into this:
```wdl
input {
  Int __b
}
Int a = 55
Int b = select_first([_b, a])
```

But preserve the fact that the input being looked for in the input file is still just `b`.

Also brings in a fix from the hermes grammar to fix the string regex (it wasn't allowing `(` or `)`)


cjllanwarne:
NB this demonstrates the changes in https://github.com/openwdl/wdl/pull/161

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wom/graph/InputDeclarationElementToGraphNode.scala
```diff
@@ -0,0 +1,67 @@
...
+    // - An expression node representing select_first([optional_input, original_expression])
+    case InputDeclarationElement(typeElement, name, Some(expr)) =>
+      val womTypeValidation: ErrorOr[WomType] = typeElement.determineWomType(a.availableTypeAliases)
+
+      val nameInInputSet = s"${a.workflowName}.$name"
+
+      womTypeValidation flatMap { womType =>
+        val optionalInputIdentifier = s"__$name"
+        val optionalInputType = WomOptionalType(womType).flatOptionalType
+        val optionalInputNode = OptionalGraphInputNode(WomIdentifier(optionalInputIdentifier), WomOptionalType(womType).flatOptionalType, nameInInputSet)
```
jbuabbud: Could replace WomOptionalType(womType).flatOptionalType with optionalInputType

PR #3445, opened 2018-03-22
Stabilize HealthServiceMonitorSpec.

mcovarr:


services/src/test/scala/cromwell/services/healthmonitor/HealthMonitorServiceActorSpec.scala
```diff
@@ -1,160 +1,210 @@
...
+      actual.zip(expected) map { case (a, e) =>
+        assert(a.name == e.name)
+        assert(a.status.ok == e.status.ok)
+        val isEmpty = a.status.messages.isEmpty && e.status.messages.isEmpty
+        lazy val actualPrefixedByExpected = for {
+          am <- a.status.messages.toList.flatten.headOption
+          em <- e.status.messages.toList.flatten.headOption
+        } yield am.startsWith(em)
+        assert(isEmpty || actualPrefixedByExpected.contains(true))
+      } head
```
Horneth: OOC why do you need the `head` here ?

services/src/test/scala/cromwell/services/healthmonitor/HealthMonitorServiceActorSpec.scala
```diff
@@ -1,160 +1,210 @@
...
+    }
+  }
+
   "HealthMonitor" should "start with unknown status for all subsystems" in {
-    newHealthMonitorActor().ask(GetCurrentStatus).mapTo[StatusCheckResponse] map { resp =>
-      assert(!resp.ok)
-      assert(resp.systems.filterNot(_._2.ok).filter(_._2 == UnknownStatus).keySet == MultipleSubsystems.map(_.name))
+    // Indirection to the constructor since "new" can't be invoked on an Actor class outside Props().
+    val actor = () => new TestHealthMonitorActor {
```
Horneth: I think you can use `TestActorRef` for this

services/src/test/scala/cromwell/services/healthmonitor/HealthMonitorServiceActorSpec.scala
```diff
@@ -1,160 +1,210 @@
...
   implicit val timeout = Timeout(5.seconds)
 
+  override val patienceConfig = PatienceConfig(timeout = scaled(20 seconds), interval = scaled(1 second))
+  implicit val patience = patienceConfig
+
+  private def eventualStatus(actorRef: ActorRef, ok: Boolean, systems: (MonitoredSubsystem, SubsystemStatus)*): Assertion = {
+    case class NameAndStatus(name: String, status: SubsystemStatus)
+
+    eventually {
+      val resp = Await.result(actorRef.ask(GetCurrentStatus).mapTo[StatusCheckResponse], Duration.Inf)
```
kshakir: ToL: I'm thinking of reducing (eliminating?) the usage hardcoded `Duration.Inf` except when necessary. Perhaps reuse the variable `implicit val timeout` defined above, since the `eventually` is using that anyway.

services/src/test/scala/cromwell/services/healthmonitor/HealthMonitorServiceActorSpec.scala
mcovarr: `eventually` wants an `Assertion` result not a `List[Assertion]`

PR #3451, opened 2018-03-26
Early feedback requested, not a real PR

geoffjentry:
@elerch


filesystems/aws/src/main/scala/cromwell/filesystems/aws/S3BatchIoCommand.scala
```diff
@@ -0,0 +1,150 @@
...
+                         IoExistsCommand,
+                         IoCopyCommand}
+import cromwell.filesystems.aws.{S3Path}
+
+/**
+  * Io commands with S3 paths and some logic enabling batching of request.
+  * @tparam T Return type of the IoCommand
+  * @tparam U Return type of the response
+  */
+sealed trait S3BatchIoCommand[T, U] extends IoCommand[T] {
```
kshakir: ToL:

Like the original OSS implementation, it looks like this S3 implementation is extending `cromwell.core.io.IoCommand` as a means of bypassing implementing a `FileSystemProvider`.

This will work in cromwell, but will be orders of magnitudes slower and more expensive when running GATK4 workflows.

If the goal is just to get a basic proof-of-concept this will work. But there's no way in hell I'm waiting (nor paying for temporary block storage) for 300+GB to localize when I only want a GATK4 job to use 0.2% of the file.

Horneth: I would agree that having an S3 NIO implementation would be very valuable inside and outside of Cromwell.
We do make use of nio capabilities inside Cromwell too to stream in parts of files.
Moreover, I don't know if S3 allows it (which I'd doubt) but GCS won't accept batching of read / write requests. If you do want to take advantage of batching optimizations some S3 requests you'll have to implement an equivalent of `GcsBatchFlow`.

cjllanwarne: @kshakir the time/cost comment on localizing (or not) files is important but unrelated to this specific PR?

* In GATK's NIO workflow versions, the file is never localized. One of my goals next quarter is to make that a Cromwell feature rather requiring dual-workflow-implementations.
* In Cromwell, we specify how to achieve IO operations using NIO in preference to rolling our own, but we don't change what IO we actually end up doing

kshakir: Yes, ToL because it is unrelated to sliding past centaur. It's fine to release this backend without NIO and call it "done", but I personally would not recommend anyone use non-NIO with BAM files-- which is kind of the point unless one is only running very, very small samples.

> In GATK's NIO workflow versions, the file is never localized.

Exactly this. This PR is not using S3 NIO as afaik it doesn't exist (yet?).

geoffjentry: I’ll note that “tol” is meaningless to the target audience of this pr

adamstruck: Rather than creating your own S3 NIO implementation you could consider using/forking https://github.com/Upplication/Amazon-S3-FileSystem-NIO2 

I've had success using it; however, I did end up having to fork it to make a few changes to get it working with Bunny.

geoffjentry: Hi everyone - just to be clear, @elerch is intending to use Amazon's incoming NIO implementation, so we're good here :)

cjllanwarne: My point was that I think @kshakir is talking about not localizing files to the VM at all and letting GATK's NIO (set up from within the VM) access the file. Unless I misunderstand, that is completely unrelated to anything Cromwell might or might not be doing with NIO (eg for things like `read_boolean` or `write_lines`)?

PR #3452, opened 2018-03-26
Call element WDLOM without inputs or outputs to WOM

jbuabbud:


wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/linking/graph/package.scala
```diff
@@ -48,17 +50,36 @@ package object graph {
...
+      case a: CallElement => a.graphElementConsumedValueHooks(typeAliases)
       // TODO fill in other expression types
       case other => throw new Exception(s"Cannot generate consumed values for WorkflowGraphNodeElement $other")
     }
   }
 
+  implicit val callElementUnlinkedValueConsumer: GraphElementValueConsumer[CallElement] = new GraphElementValueConsumer[CallElement] {
+    override def graphElementConsumedValueHooks(a: CallElement, typeAliases: Map[String, WomType]): ErrorOr[Set[UnlinkedConsumedValueHook]] = {
+      a.body match {
+        case Some(callBodyElement: CallBodyElement) => callBodyElement.inputs.toSet.flatten { input: ExpressionElement.KvPair =>
```
cjllanwarne: I suspect that a `traverse` followed by a `.map(_.flatten)`  is better here. I think that'll mean we get all of the errors if there's more than one

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/linking/graph/package.scala
```diff
@@ -48,17 +50,36 @@ package object graph {
     }
   }
 
+  implicit val callElementUnlinkedValueGenerator: UnlinkedValueGenerator[CallElement] = new UnlinkedValueGenerator[CallElement] {
+    override def generatedValueHandles(a: CallElement, typeAliases: Map[String, WomType]): ErrorOr[Set[GeneratedValueHandle]] = {
```
cjllanwarne: Thinking about including outputs... I think we're going to need to change this method to include a set of WOM `Callable`s as an argument.

~~Which is probably also going to mean we'll need to build up all of the `TaskDefinition` WOMs before we even begin to look at the workflow structure. Which actually, I think, probably does make sense (i.e. we need to know what our tasks are before we can start to build the workflow)~~ 
EDIT: I just looked at the rest of this PR, and in fact this seems to be what happens already, so it might just be as easy as changing the method signature

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wom/WorkflowDefinitionElementToWomWorkflowDefinition.scala
```diff
@@ -49,19 +45,19 @@ object WorkflowDefinitionElementToWomWorkflowDefinition {
 
     for {
       linkedGraph <- LinkedGraphMaker.make(nodes = a.graphElements, seedGeneratedValueHandles, typeAliases = a.typeAliases)
-      womGraph <- makeWomGraph(linkedGraph, a.seedNodes, a.workflowName, a.insideAScatter)
+      womGraph <- makeWomGraph(linkedGraph, a.seedNodes, a.workflowName, a.insideAScatter, a.tasks)
     } yield womGraph
   }
 
-  private def makeWomGraph(linkedGraph: LinkedGraph, seedNodes: Set[GraphNode], workflowName: String, insideAScatter: Boolean): ErrorOr[WomGraph] = {
+  private def makeWomGraph(linkedGraph: LinkedGraph, seedNodes: Set[GraphNode], workflowName: String, insideAScatter: Boolean, tasks: ErrorOr[Set[CallableTaskDefinition]]): ErrorOr[WomGraph] = {
```
cjllanwarne: I'd suggest making this `tasks: Set[Callable]` and doing the `ErrorOr` handling in the method that calls this

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wom/graph/CallElementToGraphNode.scala
```diff
@@ -0,0 +1,27 @@
...
+
+object CallElementToGraphNode extends Object {
+  def convert(a: CallableNodeMakerInputs): ErrorOr[Set[GraphNode]] = {
+    val task = a.tasks.map[CallableTaskDefinition](_.filter(_.name == a.node.callableName).head)
+    task map { task =>  Set(CommandCallNode.apply(WomIdentifier(a.node.callableName), task, Set.empty, List.empty)) }
+  }
+}
+
+final case class CallableNodeMakerInputs(node: CallElement,
+                                         tasks: ErrorOr[Set[CallableTaskDefinition]],
```
cjllanwarne: I'd like to make this just `Callable` (here and throughout the PR), since hopefully it shouldn't matter whether it's a call or a workflow that's being called here

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wom/graph/WorkflowGraphElementToGraphNode.scala
```diff
@@ -46,18 +47,17 @@ object WorkflowGraphElementToGraphNode {
...
   }
 }
 
-final case class GraphNodeMakerInputs(node: WorkflowGraphElement,
-                                      linkableValues: Map[UnlinkedConsumedValueHook, GeneratedValueHandle],
-                                      linkablePorts: Map[String, OutputPort],
-                                      availableTypeAliases: Map[String, WomType],
-                                      workflowName: String,
-                                      insideAScatter: Boolean)
+final case class GraphNodeMakerInputs(node: WorkflowGraphElement, linkableValues: Map[UnlinkedConsumedValueHook, GeneratedValueHandle], linkablePorts: Map[String, OutputPort], availableTypeAliases: Map[String, WomType], workflowName: String, insideAScatter: Boolean, tasks: ErrorOr[Set[CallableTaskDefinition]])
```
cjllanwarne: Personal preference I guess, but when the lists are this long I do tend to prefer splitting them over many lines for readability

cjllanwarne:
@jbuabbud this already looks like it's based off `develop`?

PR #3455, opened 2018-03-26
Possibly removes some flakiness in sbt test

Horneth:


mcovarr:
Assuming the 0% coverage is an artifact of this class being loaded through reflection or something?

Horneth:
@mcovarr hmm I guess technically this `NoopServiceActor` is 100% untested...
I'll move it to the test folder or test that it indeed does nothing

cjllanwarne:
@Horneth I'm pretty sure this is one of the existing AkkaTestkit actors

cjllanwarne:
(as in, a standard Akka thing, not one of our "roll our own"s

mcovarr:
I can't imagine the standard TestKit actor responds to `cromwell.util.GracefulShutdownHelper.ShutdownCommand` though...

Horneth:
The constructor signature needs to be exactly that since it's built reflectively, and like Miguel said it's better if it shuts itself down with the ShutdownCommand. Plus it's literally 5 lines so.. 😄

PR #3457, opened 2018-03-27
Support ConsString

Horneth:


Horneth:
I don't think so but it's the first layer of the multi-failures test I was trying to fix. The rest is not ready yet but I though I'd PR that already in case other tests fail because of it

geoffjentry:
👍

PR #3460, opened 2018-03-27
Stage files from expressions into the initial work dir.

mcovarr:
In full disclosure this was meant to be an interim commit toward fixing conformance 91 but these changes unintentionally got 84 to pass so hey everybody here's a PR!


cwl/src/main/scala/cwl/CwlWomExpression.scala
```diff
@@ -122,14 +121,17 @@ object InitialWorkDirFileGeneratorExpression {
...
           // A single expression which must evaluate to an array of Files
           val expressionEvaluation = ExpressionEvaluator.eval(expression, unmappedParameterContext, expressionLib)
 
           expressionEvaluation flatMap {
             case array: WomArray if WomArrayType(WomSingleFileType).coercionDefined(array) =>
               WomArrayType(WomSingleFileType).coerceRawValue(array).toErrorOr
-            case file: WomFile => file.valid
+            case file: WomFile =>
+              // The file should be staged to the initial work dir using the base filename.
+              val baseFileName = file.value.substring(file.value.lastIndexOf('/') + 1)
```
mcovarr: Hopefully Thibault's forthcoming IoFunctionSet changes can hide this from view 😄

wom/src/main/scala/wom/callable/ContainerizedInputExpression.scala
```diff
@@ -5,7 +5,7 @@ import wom.expression.IoFunctionSet
 import wom.values.WomValue
 
 /**
-  * This is an expression that
+  * This is an expression that is all I can say.
```
geoffjentry: Not strictly necessary but I'd be 👍 on fellow reviewers trying to fill in this blank with something else :)

geoffjentry:
👍

cwl/src/main/scala/cwl/CwlWomExpression.scala
cjllanwarne: ToL: since you've been looking at it most recently, could you comment what this function, and the outer `Poly1`, is actually doing? I've looked at it a few times now and it's no less opaque with this change...

cjllanwarne: You presumably also need to do the same transform to the elements of the array in the `case` above this one?

wom/src/main/scala/wom/callable/ContainerizedInputExpression.scala
cjllanwarne: At the very least. please put a TODO here to fix this this comment before handover

mcovarr: If the original author would like to suggest some text I'd be happy to include it in this PR. 😛

cwl/src/main/scala/cwl/CwlWomExpression.scala
mcovarr: I thought you wrote this?

wom/src/main/scala/wom/callable/ContainerizedInputExpression.scala
danbills: Original author here!  Could you please use `that uses "containerized" (a.k.a. "mapped" input values) to evalute its expression`.

cwl/src/main/scala/cwl/CwlWomExpression.scala
mcovarr: Maybe not for this conformance test though perhaps for my conscience.

mcovarr:
Lots of changes here for reviewer comments: this now works for expression dirents that evaluate to arrays of `WomSingleFile`s and not just "scalar" `WomSingleFile`s, although there isn't currently a test for arrays. We could possibly PR such a test into CWL (it's the same as the scalar test with the JS expression wrapped in square brackets)? We definitely shouldn't PR any tests for expression dirents that evaluate to `WomMaybePopulatedFile`s (scalar or array) since we definitely wouldn't pass such a test atm. 😦

mcovarr:
Error message from this failed test on PAPI 🤔 ```Could not copy gs://cloud-cromwell-dev/cromwell_execution/travis/linkfile.cwl/5c134cd8-80d5-47d1-a635-e4dd5df2356d/call-linkfile.cwl/home/travis/build/broadinstitute/cromwell/common-workflow-language/v1.0/v1.0/gs://centaur-cwl-conformance/cwl-inputs/Hello.java to gs://cloud-cromwell-dev/cromwell_execution/travis/linkfile.cwl/5c134cd8-80d5-47d1-a635-e4dd5df2356d/call-linkfile.cwl/Hello.java```

cwl/src/main/scala/cwl/CwlWomExpression.scala
```diff
@@ -122,16 +122,40 @@ object InitialWorkDirFileGeneratorExpression {
...
       at { expression =>
-        (_, unmappedParameterContext, mappedInputValues, expressionLib) => {
-          // A single expression which must evaluate to an array of Files
+        (ioFunctionSet, unmappedParameterContext, _, expressionLib) => {
+          // A single expression which should evaluate to a File or an array of Files.
           val expressionEvaluation = ExpressionEvaluator.eval(expression, unmappedParameterContext, expressionLib)
 
+          def stageFile(file: WomFile): Future[WomSingleFile] = {
+            // TODO WomFile could be a WomMaybePopulatedFile with secondary files but this code only stages in
+            // the primary file.
```
Horneth: OOC: Is this because it's super annoying to do or because we haven't seen a use case for it yet ?

cwl/src/main/scala/cwl/CwlWomExpression.scala
```diff
@@ -122,16 +122,40 @@ object InitialWorkDirFileGeneratorExpression {
...
+            // The file should be staged to the initial work dir using the base filename.
+            val baseFileName = file.value.substring(file.value.lastIndexOf('/') + 1)
+            ioFunctionSet.copyFile(file.value, baseFileName)
```
Horneth: ToL: I wonder how this relates to the code in `SharedFileSystem` that historically was doing "localization" of input files in the call directory (and still is I think).
This is for InitialWorkDir so a bit different but we have now "localization code" both deep in CWL territory and in backend territory.
Also I wonder how this plays in PAPI ? If those files are in GCS we can't really `copyFile` them to the VM.

cwl/src/main/scala/cwl/CwlWomExpression.scala
mcovarr: We definitely haven't seen a use case for it yet and I'm not 100% sure how to deal with directories tbh.

cwl/src/main/scala/cwl/CwlWomExpression.scala
mcovarr: I agree that it is weird having localization code here but this was [already](https://github.com/broadinstitute/cromwell/blob/develop/cwl/src/main/scala/cwl/CwlWomExpression.scala#L101) [the](https://github.com/broadinstitute/cromwell/blob/develop/cwl/src/main/scala/cwl/CwlWomExpression.scala#L116) [case](https://github.com/broadinstitute/cromwell/blob/develop/cwl/src/main/scala/cwl/CwlWomExpression.scala#L92) before these changes. These changes only add coverage for one type of directory entry that wasn't doing its localizations previously.

mcovarr: As for PAPI, this system does return the new paths to the files in GCS and I believe they will get localized in to PAPI as inputs. However the copy here is broken for reasons I don't currently understand.

PR #3464, opened 2018-03-28
Only alter permissions when running on docker

Horneth:
Thanks @mcovarr !


geoffjentry:
Apologies if it's there & i misunderstood what was going on, but would it be feasible to add a test somehow checking that it's working as intended?

I know, I know, Jeff is actually asking for tests? Up is down, cats and dogs living together in sin ....

geoffjentry:
👍

PR #3467, opened 2018-03-29
CWL Grab Bag!

mcovarr:
* Re-pin conformance test hash to the current HEAD of master.
* Allow workflow inputs to be recycled back as outputs (part 1/2 fixing conformance 20).
* Allow arrays to be coerced to Anys (part 2/2 fixing conformance 20).
* Fix prefix handling with empty arrays (fix conformance 125).
* Adjust expected conformance failures for all of the above.


geoffjentry:
👍

PR #3468, opened 2018-03-29
Path functions and secondary files / directory improvements

Horneth:
- Cleans up the `IoFunctionSet` a little and add `PathFunctionSet` in it to deal with path manipulation that doesn't involve I/O
- When secondary files are actually secondary directories, list their content and return a `WomMaybeListedDirectory` instead of a `WomSingleFile` or `WomMaybePopulatedFile`
- Make paths absolute as much and as early as possible to avoid ambiguity and having to guess later where they are relative to
- Adjust the `OutputManipulator` to deal with secondary listed directories


mcovarr:
Could you please give a very brief overview of the conceptual changes here?

centaurCwlRunner/src/main/scala/centaur/cwl/OutputManipulator.scala
```diff
@@ -85,39 +101,39 @@ object OutputManipulator extends Poly1 {
         * 2) the size calculated from the value of the "contents" field of the metadata
         * 3) the size calculated from the file itself
         */
-      val defaultSize = sizeContent.orElse(sizeFile).map(Json.fromLong).getOrElse(Json.Null)
+      val defaultSize = sizeContent.orElse(sizeFile(path)).map(Json.fromLong).getOrElse(Json.Null)
       val size = valueOrNull("size", defaultSize)
 
+      val basename: Option[Json] = if (isInsideDirectory) {
+        Option(valueOrNull("basename", path.exists.option(path.nameWithoutExtension).map(Json.fromString).getOrElse(Json.Null)))
+      } else None
```
mcovarr: I'm not getting the "inside directory" logic here

cwl/src/main/scala/cwl/CwlType.scala
```diff
@@ -107,18 +110,44 @@ object File {
...
 
   def secondaryExpressionFiles(primaryWomFile: WomFile,
                                stringWomFileType: WomFileType,
                                expression: Expression,
                                parameterContext: ParameterContext,
-                               expressionLib: ExpressionLib): ErrorOr[List[WomFile]] = {
+                               expressionLib: ExpressionLib,
+                               ioFunctions: IoFunctionSet): ErrorOr[List[WomFile]] = {
+    
+    def relativize(path: String) = ioFunctions.pathFunctions.absoluteSibling(primaryWomFile.value, path)
```
mcovarr: `relativize` returns the result of `absoluteSibling`?

src/bin/travis/resources/local_conformance_expected_failures.txt
```diff
@@ -88,10 +88,10 @@
...
 #90
-91
+#91
 92
 #93
-94
+#94
 #95
 96
 #97
```
mcovarr: For rebase these are 55, 91 and 99

src/bin/travis/resources/papi_conformance_expected_failures.txt
```diff
@@ -79,7 +79,7 @@
 79
 80
 81
-82
+#82
 83
 84
 #85
```
mcovarr: 56 and 87

backend/src/main/scala/cromwell/backend/wdl/WriteFunctions.scala
```diff
@@ -33,7 +30,7 @@ trait WriteFunctions extends PathFactory with IoFunctionSet with AsyncIoFunction
 
   private val relativeToLocal = System.getProperty("user.dir").ensureSlashed
 
-  def relativeToAbsolutePath(pathFrom: String): String = if (new File(pathFrom).isAbsolute) pathFrom else relativeToLocal + pathFrom
+  def relativeToAbsolutePath(pathFrom: String): String = if (buildPath(pathFrom).isAbsolute) pathFrom else relativeToLocal + pathFrom
```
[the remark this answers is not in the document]
Horneth: yes, and that's why we ended up with weird concatenated gcs paths

centaurCwlRunner/src/main/scala/centaur/cwl/OutputManipulator.scala
Horneth: Yeah I meant to add a comment and forgot. Apparently the conformance tests expect the files listed in the `listing` field for directories to have `basename` filled in, but only there and not for other top level files...
So this flag is there to decide if we should add basename

cwl/src/main/scala/cwl/CwlType.scala
Horneth: Yeah the naming is confusing -_-
What it's doing is taking a "relative" path A and turning it into an absolute path B. It's doing that by saying, A is actually relative to absolute path C, so the absolute version of A is C/A.
Not sure how to describe that in terms of function names 😄

backend/src/main/scala/cromwell/backend/io/DirectoryFunctions.scala
```diff
@@ -41,25 +46,7 @@ trait DirectoryFunctions extends IoFunctionSet with PathFactory {
...
-      } else {
-        newCheckedPaths.valid
-      }
-    }
-
-    val allPaths = listPaths(path, Set.empty).map(_.toList)
-    val allFiles = allPaths.map(_.filterNot(_.isDirectory))
-    allFiles
-  }
+  def listFiles(path: Path): ErrorOr[List[Path]] = path.listRecursively.filterNot(_.isDirectory).toList.validNel
```
danbills: :100:

PR #3470, opened 2018-03-31
CWL Restart

mcovarr:
~~Pulling this back to a pregull, I want to think about and possibly robustify simpleton key handling.~~  Have at it!

### DIFF
--- a/core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala
+++ b/core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala
@@ -15,16 +15,23 @@ import scala.language.postfixOps
 object WomValueBuilder {
 
   /**
-    * Looks for a WOM identifier possibly followed by a metacharacter and more stuff.
+    * Looks for a WOM identifier possibly followed by a metacharacter and more stuff. A 'WOM identifier' is kind of a
+    * made up term that encompasses anything preceding an open square brace or colon (used for encoding simpleton paths).
     *
     * Capture groups:
     *
     * <ol>
-    * <li>A WOM identifier</li>
-    * <li>Possibly a metacharacter and more stuff after the WOM identifier</li>
+    * <li>A WOM identifier (nongreedy match of one or more of any character)</li>
+    * <li>Either:
+    *   <ol>
+    *     <li>An open square bracket or colon followed by anything.</li>
+    *     <li>Nothing at all, forcing the initial nongreedy match to consume everything.</li>
+    *   </ol>
+    * </li>
     * </ol>
+    *
     */
-  private val IdentifierAndPathPattern = "^([a-zA-Z][a-zA-Z0-9_]*)(.*)".r
+  private val IdentifierAndPathPattern = raw"^(.+?)([\[:].*|)".r
 
   /**
     * Looks for an array element reference: square braces surrounding digits, possibly followed by a metacharacter and
--- a/src/bin/travis/resources/jes_centaur.conf.ctmpl
+++ b/src/bin/travis/resources/jes_centaur.conf.ctmpl
@@ -96,5 +96,9 @@ database {
   profile = "slick.jdbc.MySQLProfile$"
 }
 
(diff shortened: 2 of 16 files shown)


core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala
```diff
@@ -15,16 +15,16 @@ import scala.language.postfixOps
 object WomValueBuilder {
 
   /**
-    * Looks for a WDL identifier possibly followed by a metacharacter and more stuff.
+    * Looks for a WOM identifier possibly followed by a colon metacharacter and more stuff.
```
mcovarr: fix

core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala
```diff
@@ -15,16 +15,16 @@ import scala.language.postfixOps
...
     *
     * <ol>
-    * <li>A WDL identifier</li>
-    * <li>Possibly a metacharacter and more stuff after the WDL identifier</li>
+    * <li>A WOM identifier</li>
+    * <li>Possibly a metacharacter and more stuff after the WOM identifier</li>
     * </ol>
     */
-  private val IdentifierAndPathPattern = "^([a-zA-Z][a-zA-Z0-9_]*)(.*)".r
+  private val IdentifierAndPathPattern = "^([a-zA-Z][-a-zA-Z0-9_]*)(.*)".r
```
mcovarr: Currently this expands the simpleton key character space to include dash characters. The use of a dash in 'ps-stdOut' was terminating identifier parsing prematurely and causing that file not to be populated the way downstream code expected.

But AFAICT there aren't any restrictions on CWL identifier naming at all so our code should either be more robust than this or at least generate a non-useless fail-fast error message if we find ourselves in this situation which is most definitely not what we're doing now.

core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala
```diff
@@ -143,13 +151,13 @@ object WomValueBuilder {
       case arrayType: WomArrayType =>
         WomArray(arrayType, components.asArray map { toWomValue(arrayType.memberType, _) })
       case mapType: WomMapType =>
-        // map keys are guaranteed by the WDL spec to be primitives, so the "coerceRawValue(..).get" is safe.
+        // map keys are guaranteed by the WDL spec to be primitives, so the "coerceRawValue(..).get" is safe if this is WDL.
```
geoffjentry: What happens if this is not WDL?

src/bin/travis/resources/jes_centaur.conf.ctmpl
```diff
@@ -96,5 +96,9 @@ database {
   profile = "slick.jdbc.MySQLProfile$"
 }
 
+backend.providers.Jes.config.concurrent-job-limit = 20
```
geoffjentry: what's up w/ all these config changes?

mcovarr: Since all the Centaur CWLs are now eligible to run concurrently they were overwhelming Cromwell on Travis at the default job limit (10 million I think).

core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala
mcovarr: Yeah good question. This file was renamed from `WdlValueBuilder` and much of it wasn't updated for our brave new WOMmy world.

centaur/src/main/resources/standardTestCases/three_step/three_step.cwl
```diff
@@ -19,11 +19,11 @@ steps:
 - id: ps
   in: []
   out:
-  - id: ps-stdOut
+  - id: "ps-stdOut"
```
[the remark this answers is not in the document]
mcovarr: Oh I can probably revert that, I had been experimenting with even more bizarre identifiers but backed away.

PR #3474, opened 2018-04-03
Shorten inputs dir

danbills:
# To be discussed

I would like to delete the test for  "A task which contains a parameter " from Mapworkflowspec:

* it was constructed by making assumptions about the path that are no longer the case.
* It is basic and almost certainly covered in centaur tests
* It was written in the late Cretaceous period


supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystem.scala
```diff
@@ -188,7 +188,7 @@ trait SharedFileSystem extends PathFactory {
       val dest = if (inputsRoot.isParentOf(localInputPath)) localInputPath
       else {
         // Concatenate call directory with absolute input path
-        DefaultPathBuilder.get(inputsRoot.pathAsString, localInputPath.pathAsString)
+        DefaultPathBuilder.get(inputsRoot.pathAsString, localInputPath.parent.pathAsString.hashCode.toString + "/" + localInputPath.name)
```
Horneth: `DefaultPathBuilder.get(inputsRoot.pathAsString, localInputPath.parent.pathAsString.hashCode.toString, localInputPath.name)`

Horneth: I imagine the answer is yes but if 2 strings have the same value do they necessarily have the same `hashCode` ?
We want it to be the case so that files that were originally in the same directory always end up in the same "shorten" directory.

geoffjentry:
@danbills Can you provide a few before & after examples for paths? I'm concerned about overhashing, i.e. removing parts that are actually providing semantic value as people due go spelunking in these directory trees.

geoffjentry:
Agree w/ @Horneth - we should have consistent pathing across all platforms.

danbills:
Minutes from recent in-person discussions:

* other teams are not using prior knowledge of these paths (namely mint team is looking at outputs, not input paths).
* other backends do nothing during "preProcess" stage, thus do not have to conform to this new arrangement.

also since I've heard nothing regarding the test deletion I proposed I'm going to proceed as planned

supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystem.scala
danbills: yes, strings have the same hashcode when they are effectively the same value (even if a different instance)

cjllanwarne:
@danbills it makes me nervous that we're changing how something works in a way that enables one test, and then deleting an old test which was no longer working.

Not saying this is bad or wrong (and the fewer of those old "run a workflow" tests the better)... this is just a request to check through the centaur tests and make sure they really do cover the case which you're assuming they do

danbills:
@cjllanwarne Indeed `read_map` and `write_map` was not being tested in centaur.  I will migrate this test.

geoffjentry:
@mcovarr i retracted my request at standup :)

danbills:
@mcovarr 
before:
image

After:
image

mcovarr:
😍
so if there were secondary files that were siblings to these files they would go in the same directory?

danbills:
yep!

geoffjentry:
That last exchange just made me wonder, is there any possibility of a file name collision which wouldn't have happened before? I don't think so but wanted to double check.

danbills:
@geoffjentry no, in fact the test I took screenshots from is a filename collision test, notice they're both name `out.txt`

Issue #3074, opened 2017-12-17
No call caching on restart edge case


Horneth:
Because writing to the call caching store and the job store is not atomic, the following chain of events is possible and not necessarily desirable:

- A job start
- A cache hit is found
- The outputs are copied
- The hashes / simpletons are written to the DB
- ** Cromwell Stops **: This is after the hashes are written successfully but before the EJEA had a chance to write the outputs to the job store and mark the job as complete.
- Cromwell starts
- The workflow is restarted
- The job is not found in the job store
- At this point the EJEA has a state to check if there are hashes existing for this job already. If there is, it disables call caching (so that the EJEA doesn't try to call cache to himself, and that we don't write to the hash store again - which would fail because of the unique index in the call cache table).
- However since we've disabled call caching we then proceed to try and recover the job, which fails because it was never run (since we found a cache hit the first time), and then falls back to running the job for reals.

This is not great because this job already has all the outputs it needs, files have been copied already, but we run the job on top of it, which seems to increase the likelihood of having empty files at least locally when trying to read outputs and cause `cannot create an Int from ""` types of failures.

Maybe a better way would be to re-use the outputs that have been written to the cache to make the job succeed and bypass all the rest.

Relevant code in the EJEA: https://github.com/broadinstitute/cromwell/blob/develop/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/job/EngineJobExecutionActor.scala#L153

Horneth:
Thinking about it more, what's probably happening is because we're trying to run a job where all the files that it's going to produce have already been copied over, including the return code file, the first poll immediately sees an RC file and thinks the job is done while it's actually still running and overwriting the output files.

This is definitely an edge case but if it does happen it could have pretty bad consequences.

Horneth:
This also causes flakiness in the call cache capoeira centaur test

PR #3476, opened 2018-04-04
Fix edge case on restart in EJEA

Horneth:
See https://github.com/broadinstitute/cromwell/issues/3074 for a description of the problem

This re-uses the values from the cache instead of re-running the job.
The call cache capoeira test was failing in centaur intermittently because a side effect of the issue was that the hashes were not published to metadata, which was failing the test.

### DIFF
--- a/core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala
+++ b/core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala
@@ -15,13 +15,13 @@ import scala.language.postfixOps
 object WomValueBuilder {
 
   /**
-    * Looks for a WDL identifier possibly followed by a metacharacter and more stuff.
+    * Looks for a WOM identifier possibly followed by a metacharacter and more stuff.
     *
     * Capture groups:
     *
     * <ol>
-    * <li>A WDL identifier</li>
-    * <li>Possibly a metacharacter and more stuff after the WDL identifier</li>
+    * <li>A WOM identifier</li>
+    * <li>Possibly a metacharacter and more stuff after the WOM identifier</li>
     * </ol>
     */
   private val IdentifierAndPathPattern = "^([a-zA-Z][a-zA-Z0-9_]*)(.*)".r
@@ -143,13 +143,13 @@ object WomValueBuilder {
       case arrayType: WomArrayType =>
         WomArray(arrayType, components.asArray map { toWomValue(arrayType.memberType, _) })
       case mapType: WomMapType =>
-        // map keys are guaranteed by the WDL spec to be primitives, so the "coerceRawValue(..).get" is safe.
+        // map keys are guaranteed by WOM to be primitives, so the "coerceRawValue(..).get" is safe.
         WomMap(mapType, components.asMap map { case (k, ss) => mapType.keyType.coerceRawValue(k).get -> toWomValue(mapType.valueType, ss) })
       case pairType: WomPairType =>
         val groupedByLeftOrRight: Map[PairLeftOrRight, Traversable[SimpletonComponent]] = group(components map descendIntoPair)
         WomPair(toWomValue(pairType.leftType, groupedByLeftOrRight(PairLeft)), toWomValue(pairType.rightType, groupedByLeftOrRight(PairRight)))
       case WomObjectType =>
-        // map keys are guaranteed by the WDL spec to be primitives, so the "coerceRawValue(..).get" is safe.
+        // map keys are guaranteed by WOM to be primitives, so the "coerceRawValue(..).get" is safe.
         val map: Map[String, WomValue] = components.asMap map { case (k, ss) => k -> toWomValue(WomAnyType, ss) }
         WomObject(map)
       case composite: WomCompositeType =>
@@ -212,13 +212,13 @@ object WomValueBuilder {
     * path to the element.  e.g. for a `WomValueSimpleton` of
     *
     * {{{
-    * WomValueSimpleton("foo:bar[0]", WdlString("baz"))
+    * WomValueSimpleton("foo:bar[0]", WomString("baz"))
     * }}}
     *
     * the corresponding `SimpletonComponent` would be
     *
     * {{{
-    * SimpletonComponent(":bar[0]", WdlString("baz"))
+    * SimpletonComponent(":bar[0]", WomString("baz"))
     * }}}
     */
   private case class SimpletonComponent(path: String, value: WomValue)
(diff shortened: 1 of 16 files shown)


core/src/main/scala/cromwell/core/simpleton/WomValueBuilder.scala
```diff
@@ -224,10 +224,10 @@ object WomValueBuilder {
   private case class SimpletonComponent(path: String, value: WomValue)
 
   def toJobOutputs(taskOutputs: Traversable[OutputPort], simpletons: Traversable[WomValueSimpleton]): CallOutputs = {
-    CallOutputs(toWdlValues(taskOutputs, simpletons))
+    CallOutputs(toWomValues(taskOutputs, simpletons))
   }
 
-  def toWdlValues(taskOutputs: Traversable[OutputPort], simpletons: Traversable[WomValueSimpleton]): Map[OutputPort, WomValue] = {
+  def toWomValues(taskOutputs: Traversable[OutputPort], simpletons: Traversable[WomValueSimpleton]): Map[OutputPort, WomValue] = {
```
mcovarr: As long as you're in here doing this there are many more opportunities to remove inappropriate references to `WDL` in this file. 😄

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/job/EngineJobExecutionActor.scala
```diff
@@ -141,17 +142,23 @@ class EngineJobExecutionActor(replyTo: ActorRef,
 
   // When we're restarting but the job store says the job is not complete.
   // This is to cover for the case where Cromwell was stopped after writing Cache Info to the DB but before
-  // writing to the JobStore. In that case, we don't want to cache to ourselves (turn cache read off), nor do we want to 
-  // try and write the cache info again, which would fail (turn cache write off).
-  // This means call caching should be disabled.
-  // Note that we check that there is not a Cache entry for *this* current job. It's still technically possible
-  // to call cache to another job that finished while this one was running (before the restart).
+  // writing to the JobStore. In that case, we can re-use the cached results and save them to the job store directly.
+  // Note that this checks that *this* particular job has a cache entry, not that there is *a* cache hit (possibly from another jobs)
```
mcovarr: jobs => job

database/sql/src/main/scala/cromwell/database/slick/CallCachingSlickDatabase.scala
```diff
@@ -92,12 +92,31 @@ trait CallCachingSlickDatabase extends CallCachingSqlDatabase {
...
+    for {
+      callCachingSimpletonEntries <- dataAccess.
+        callCachingSimpletonEntriesForCallCachingEntryId(callCachingEntryId).result
+      callCachingDetritusEntries <- dataAccess.
+        callCachingDetritusEntriesForCallCachingEntryId(callCachingEntryId).result
+      callCachingHashEntries <- dataAccess.
+        callCachingHashEntriesForCallCachingEntryId(callCachingEntryId).result
+      callCachingAggregationEntries <- dataAccess.
+        callCachingAggregationForCacheEntryId(callCachingEntryId).result.headOption
+    } yield CallCachingJoin(callCachingEntry, callCachingHashEntries, callCachingAggregationEntries, callCachingSimpletonEntries, callCachingDetritusEntries)
```
danbills: ToL could be a (x,y,z) mapN ?

database/sql/src/main/scala/cromwell/database/slick/CallCachingSlickDatabase.scala
```diff
@@ -92,12 +92,31 @@ trait CallCachingSlickDatabase extends CallCachingSqlDatabase {
...
+  }
+
+  override def callCacheJoinForCall(workflowExecutionUuid: String, callFqn: String, index: Int)
+                                   (implicit ec: ExecutionContext): Future[Option[CallCachingJoin]] = {
     val action = for {
       callCachingEntryOption <- dataAccess.
         callCachingEntriesForWorkflowFqnIndex((workflowExecutionUuid, callFqn, index)).result.headOption
-    } yield callCachingEntryOption.nonEmpty
+      callCacheJoin <- callCachingEntryOption match {
+        case Some(entry) => callCacheJoinFromEntryQuery(entry).map(Option.apply)
```
danbills: ToL I've been spreading the good word about `cats.syntax.option._`'s `fold` method which basically does this match but with less boilerplate

geoffjentry: Leaving aside the "what is this, amateur hour?" meme, I've been personally swayed back to the `match` structure via vigorous support from Odersky (who also notes that it's the fastest scheme out there, but premature optimization and all of that).

I don't really care either way but throwing it out there

danbills: Style over speed IMO.  

actually now that I think about it, this is a `map(callCacheJoinFromEntryQuery).getOrElse(DBIO.successful(None))` in disguise

mcovarr: matcheroni

geoffjentry: I want to like fold for options, but never have. I did live squarely in team flatmap for a while though.

database/sql/src/main/scala/cromwell/database/slick/CallCachingSlickDatabase.scala
Horneth: Actually I'm not sure if / how this can be done with slick DBIO actions. I'm probably going to need your FP expertise

mcovarr: oh the DBIO stuff is not Cats so unless you want to write all the instances... 😬

Horneth: I found this https://github.com/RMSone/slick-cats but it doesn't have an Applicative for DBIO AFAICT so it still doesn't work

cjllanwarne: If you can make a `toErrorOr` for `DBIO` then that it might be easier/less stressful than implementing the typeclass (at least upfront).

If there's a fundamental reason why `DBIO.toErrorOr` wouldn't work, you'd probably run into the same problem trying to make the applicative.

So, I'd probably start with trying to write that conversion method

danbills: slick-cats' DBIO Monad is all you need for cats to figure out the Applicative ( as it is derived from the `flatMap` in the monad instance)

cjllanwarne: @danbills wait, so why can't we have applicative for `Checked[A]`?

danbills: @cjllanwarne We most certainly can (and do) have an applicative for it (i.e. `Either`), but what we usually want is the applicative for `Validated`, thus the converting back and forth all the time

PR #3489, opened 2018-04-06
Docs: Explicity invoke bash in SGE, instead of relying on #!

EvanTheB:
SGE defaults to csh, and at least in my environment, even with #!/bin/bash bash was not launched in the job. This could well just be a quirk of my environment, and I am happy if you decide to not take this change. However it seems like a more generally robust default for different environments.

### DIFF
--- a/docs/tutorials/HPCIntro.md
+++ b/docs/tutorials/HPCIntro.md
@@ -146,7 +146,7 @@ qsub -terse -V -b y -N my_job_name \
   -o /path/to/stdout \
   -e /path/to/stderr \
   -pe smp 1 -l mem_free=0.5g -q short \
-  bash myScript.bash
+  /usr/bin/env bash myScript.bash
 ```
 
 For this particular SGE cluster, the above sets the working directory, stdout and stderr paths, the number of cpus to 1, the memory to half a gigabyte, and runs on the short queue.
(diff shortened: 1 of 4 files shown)


cjllanwarne:
I like this:
* it makes our dependency on bash explicit
* users can always turn in back if it doesn't work for them

I'm curous whether the `qsub` lets you specify what shell to use as a separate parameter rather than having to revert to full "binary command submission" mode?

docs/tutorials/HPCIntro.md
```diff
@@ -141,12 +141,12 @@ backend.providers.SGE.config {
...
 
 ```bash
-qsub -terse -V -b n -N my_job_name \
+qsub -terse -V -b y -N my_job_name \
   -wd /path/to/working_directory \
   -o /path/to/stdout \
   -e /path/to/stderr \
   -pe smp 1 -l mem_free=0.5g -q short \
-  myScript.bash
+  bash myScript.bash
```
Horneth: I would be consistent with either `bash` or `/bin/bash` everywhere, unless there's a reason not to ?

EvanTheB: Yes I think you are right, I know some arguments call for '/usr/bin/env bash', which allows a separate version of bash to be specified with environment variables. I think that might be the best way.

Issue #3161, opened 2018-01-18
Workflow option to retry any task failure


dshiga:
# Summary

Currently Cromwell will retry tasks that fail in Pipelines API due to preemption, with the number of retries configurable on a task by task basis.

It would be very helpful if this could be generalized, so that I could tell Cromwell to retry all tasks that fail -- for any reason, not just preemption.

I imagine this being configured via a workflow option like "failed_task_retries: 3", which would tell Cromwell to run each task in the workflow up to 3 times if any type of failure is encountered.

# Why it would be valuable

For people running many instances of a well-tested workflow, such as Green Team and Mint Team production at Broad, the vast majority of failures are due to transient problems in the cloud, and it is very time consuming to deal with them.

Having this auto-retry capability in Cromwell would be a huge help in making these workflows more robust and would greatly reduce the amount of manual work required to relaunch failed workflows (or save people from having to write their own bespoke scripts to auto-retry failed workflows). Having retries at the task level (rather than having to resubmit the whole workflow) would also be more efficient, especially when call caching is not in use.

# Difference from existing issue

I believe this feature would satisfy the use cases of many (but not all) of the commenters on #1991, but in a simpler way. In contrast to that issue, no error messages need to be parsed here and there is no added functionality around auto increasing memory or disk. (For Mint Team produciton, we're interested in something like #1991, too, especially the stderr pattern matching, but I am guessing it would take longer to make happen given the wdl changes required, etc. The issue I'm filing here is the low hanging fruit for us.)

# Combining with preemptibles

There is a question to resolve about what to do for a preemptible task in a workflow where failed_task_retries has also been set.

My preference would be to make them additive. If the task says "preemptible: 5" and the workflow says "failed_task_retries: 3", then Cromwell will retry that task up to 8 times. The first 3 retries will use a preemptible machine -- regardless of the reason for the error. Additional retries beyond that will not use a preemptible machine.

This approach would let failed_task_retries act as a floor for the entire workflow that guarantees all tasks, preemptible or not, will retry at least failed_task_retries times upon failure.

cjllanwarne:
I remember we added this for certain JES failures so the backend plumbing might already be there (although IIRC we called it "unexpected failure" and it was hard coded to 3)

dshiga:
**Update: Non-additive retry counts**

If failures due to preemption can be clearly distinguished from failures due to other causes, I would prefer the failed_task_retries count to be independent of the preemptible count, rather than additive.

[...]

PR #3490, opened 2018-04-06
Workflow Option to retry jobs in JES

ruchim:
Taking advantage of the existing framework of "unexpectedRetryCount" and to implement https://github.com/broadinstitute/cromwell/issues/3161.

### DIFF
--- a/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/job/preparation/JobPreparationActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/job/preparation/JobPreparationActor.scala
@@ -187,7 +187,10 @@ class JobPreparationActor(workflowDescriptor: EngineWorkflowDescriptor,
                                                     runtimeAttributes: Map[LocallyQualifiedName, WomValue],
                                                     maybeCallCachingEligible: MaybeCallCachingEligible,
                                                     prefetchedJobStoreEntries: Map[String, KvResponse]): BackendJobPreparationSucceeded = {
-    val jobDescriptor = BackendJobDescriptor(workflowDescriptor.backendDescriptor, jobKey, runtimeAttributes, inputEvaluation, maybeCallCachingEligible, prefetchedJobStoreEntries)
+    val backendWorkflowDescriptor = workflowDescriptor.backendDescriptor
+    val jobDescriptor = BackendJobDescriptor(backendWorkflowDescriptor, jobKey, runtimeAttributes, inputEvaluation,
+      maybeCallCachingEligible, prefetchedJobStoreEntries, backendWorkflowDescriptor.jobRetries)
+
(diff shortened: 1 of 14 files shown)


engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/job/preparation/JobPreparationActor.scala
```diff
@@ -187,7 +187,10 @@ class JobPreparationActor(workflowDescriptor: EngineWorkflowDescriptor,
                                                     runtimeAttributes: Map[LocallyQualifiedName, WomValue],
                                                     maybeCallCachingEligible: MaybeCallCachingEligible,
                                                     prefetchedJobStoreEntries: Map[String, KvResponse]): BackendJobPreparationSucceeded = {
-    val jobDescriptor = BackendJobDescriptor(workflowDescriptor.backendDescriptor, jobKey, runtimeAttributes, inputEvaluation, maybeCallCachingEligible, prefetchedJobStoreEntries)
+    val backendWorkflowDescriptor = workflowDescriptor.backendDescriptor
+    val jobDescriptor = BackendJobDescriptor(backendWorkflowDescriptor, jobKey, runtimeAttributes, inputEvaluation,
```
cjllanwarne: ToL: IMO if you're going to separate lines, separate everything onto their own lines...

cjllanwarne:
I'm not sure the exact scenario people would want to use this in, but it seems to replace the previous behavior "retry a few times for known flakinesses, otherwise fail" - with a less tuned "retry a custom number of times for all codes".

* Is there any possibility that users would want to continue retrying on known flakinesses (hint: I really think they do) but not blindly restart on other error codes (which could be expensive for typos!)?
---
One other thought:
* This seems to only catch JES telling us the task has failed. If the task on JES "succeeds" but Cromwell later finds out that it cannot download or evaluate a result, does this still retry?

geoffjentry:
@cjlanwarne re discussing granularity of retries, note that this exactly satisfies a request from mint (see the ticket linked in the pr description)

geoffjentry:
Or at least that was the intention, I haven’t looked at the code to see if it succeeds but I have faith in @ruchim :)

geoffjentry:
Deleted a few comments as I misread what @cola Warner said. Sorry I’m leaving that autocorrect typo as it’s awesome

We should discuss what this does and does not do. I viewed the request to be an enhancement on current behavior, so should ensure that’s the case

geoffjentry:
@cjllanwarne @ruchim I've been going back and forth as to whether or not this retry count should replace or add on to the "we know that's flakey so retried it for ya" situation vs. adding on top of it. Chris, you seemed pretty sure it should be the latter, why do you view it that way? Ruchi, I know this was spec'd out to go the former route, but I think the person speccing it out didn't know that we already did some stuff - do you actually have an opinion here or is this just how you did it?

ruchim:
@geoffjentry @cjllanwarne I've already made changes after a conversation with Chris on Friday--so I'll close this out for now, and re-open when we've all had a chance to talk about specifics.

PR #3493, opened 2018-04-08
Write workflow heartbeats.

mcovarr:
The current heartbeat-scanning code is incorrect and will attempt to relaunch workflows an hour after they start. Fortunately there is an error check that prevents workflows from actually being restarted but this situation is still less than ideal since these workflows cycle back to the top of the runnable queue an hour later (running workflows will have submission times earlier than submitted-but-never-started workflows and the sort is by submission time). Also some semi-alarming warning messages are generated for these failed starts.

These changes should fix this problem by actually writing workflow heartbeats for running workflows. The current vitality logic is simply "is this workflow actor not dead".

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -173,12 +173,19 @@ system {
     object = 128000
   }
 
-  # Rate at which Cromwell updates it's instrumentation gauge metrics (e.g: Number of workflows running, queued, etc..)
+  # Rate at which Cromwell updates its instrumentation gauge metrics (e.g: Number of workflows running, queued, etc..)
   instrumentation-rate = 5 seconds
   
   job-rate-control {
     jobs = 50
     per = 1 second
+  }
+
+  workflow-heartbeats {
+    heartbeat-interval: 2 minutes
+    ttl: 10 minutes
+    write-batch-size: 10000
+    write-threshold: 10000
   }
 }
 
--- a/engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowHeartbeatConfig.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowHeartbeatConfig.scala
@@ -1,0 +1,64 @@
+package cromwell.engine.workflow.workflowstore
+
+import java.util.UUID
+
+import com.typesafe.config.{Config, ConfigFactory}
+import io.circe.generic.auto._
+import io.circe.literal._
+import io.circe.syntax._
+import io.circe._
+import net.ceedubs.ficus.Ficus._
+
+import scala.concurrent.duration._
+import scala.language.postfixOps
+
+
+/**
+  *
+  * @param cromwellId         Identifier for this Cromwell for horizontaling purposes.
+  * @param heartbeatInterval  How long between `WorkflowActor` heartbeats and `WorkflowStoreHeartbeatWriteActor` database flushes.
+  * @param ttl                Entries in the workflow store with heartbeats older than `ttl` ago are presumed abandoned and up for grabs.
+  * @param writeBatchSize     The maximum size of a write batch.
+  * @param writeThreshold     The threshold of heartbeat writes above which load is considered high.
+  */
+case class WorkflowHeartbeatConfig(
+                                    cromwellId: String,
+                                    heartbeatInterval: FiniteDuration,
+                                    ttl: FiniteDuration,
+                                    writeBatchSize: Int,
+                                    writeThreshold: Int)
+{
+
+  // An `Encoder[FiniteDuration]` is needed for the `.asJson.toString()` shenanigans below. Unfortunately the compiler
+  // appears to get confused about this encoder actually being used. If this is made fully private or a local the compiler
+  // wrongly emits a warning about it being unused, which our compiler flag settings then promote to an error.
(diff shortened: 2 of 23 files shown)


engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowHeartbeatConfig.scala
```diff
@@ -0,0 +1,59 @@
...
+/**
+  *
+  * @param cromwellId         Identifier for this Cromwell for horizontaling purposes.
+  * @param ttl                Entries in the workflow store with heartbeats older than `ttl` ago are presumed abandoned and up for grabs.
+  * @param sendsPerTtl        How many heartbeats each `WorkflowActor` sends within an interval of duration `ttl`.
+  * @param writeBatchSize     The maximum size of a write batch.
+  * @param writeBatchesPerTtl How many write batches to attempt within an interval of duration `ttl`.
+  * @param writeThreshold     The threshold of heartbeat writes above which load is considered high.
+  */
+case class WorkflowHeartbeatConfig(
```
geoffjentry: Dan is going to be so happy to see this

engine/src/main/scala/cromwell/server/CromwellRootActor.scala
```diff
@@ -27,13 +25,14 @@ import cromwell.engine.workflow.WorkflowManagerActor.AbortAllWorkflowsCommand
...
 import cromwell.engine.workflow.tokens.{DynamicRateLimiter, JobExecutionTokenDispenserActor}
-import cromwell.engine.workflow.workflowstore.{SqlWorkflowStore, WorkflowStore, WorkflowStoreActor}
+import cromwell.engine.workflow.workflowstore.{SqlWorkflowStore, WorkflowHeartbeatConfig, WorkflowStore, WorkflowStoreActor}
 import cromwell.jobstore.{JobStore, JobStoreActor, SqlJobStore}
 import cromwell.services.{EngineServicesStore, ServiceRegistryActor}
 import cromwell.subworkflowstore.{SqlSubWorkflowStore, SubWorkflowStoreActor}
 import cromwell.util.GracefulShutdownHelper
 import cromwell.util.GracefulShutdownHelper.ShutdownCommand
 import net.ceedubs.ficus.Ficus._
+import sext._
```
geoffjentry: What is this being used for?

mcovarr: Line 63. The output looks like:

```
[INFO] [04/07/2018 20:47:33.646] [cromwell-system-akka.dispatchers.engine-dispatcher-5] [akka://cromwell-system/user/cromwell-service] Workflow heartbeat configuation:
- cromwellId:
| cromid-286ccb0
- ttl:
| 10 minutes
- sendsPerTtl:
| 60
- writeBatchSize:
| 100
- writeBatchesPerTtl:
| 60
- writeThreshold:
| 100
```

mcovarr: Though actually with an `Encoder[FiniteDuration]` Circe can derive an `Encoder[WorkflowHeartbeatConfig]` that produces this which is probably just as good:

```
{
  "cromwellId" : "cromid-80eb68c",
  "ttl" : "10 minutes",
  "sendsPerTtl" : 3,
  "writeBatchesPerTtl" : 3,
  "writeBatchSize" : 100,
  "writeThreshold" : 100
}
```

core/src/main/resources/reference.conf
```diff
@@ -173,13 +173,20 @@ system {
...
   instrumentation-rate = 5 seconds
   
   job-rate-control {
     jobs = 50
     per = 1 second
   }
+
+  workflow-heartbeats {
+    ttl: 10 minutes
+    heartbeats-per-ttl: 3
```
Horneth: OOC do you think there's a use case where someone would want to change this or is it only so that it's not hardcoded ?

mcovarr: I was using it a lot during testing, though I could have accomplished something similar by tweaking `ttl` instead. 🙂 I was making knobs for everything else so I made one for this too. shrug

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowHeartbeatConfig.scala
```diff
@@ -0,0 +1,64 @@
...
+
+import scala.concurrent.duration._
+import scala.language.postfixOps
+
+
+/**
+  *
+  * @param cromwellId         Identifier for this Cromwell for horizontaling purposes.
+  * @param ttl                Entries in the workflow store with heartbeats older than `ttl` ago are presumed abandoned and up for grabs.
+  * @param heartbeatsPerTtl   How many heartbeats each `WorkflowActor` sends within an interval of duration `ttl` and how frequently
```
kcibul: Why not just define this as heartbeatInterval?  It looks like everywhere it is used, it's in that form after doing the math.  Also might eliminate problems where someone extends the TTL, expecting the heartbeat to be the same rate and not realizing that they have do up that as well.

mcovarr: I thought it was more intuitive to think of it as the number of heartbeats per TTL but it could certainly be done as an interval instead. You're certainly right it's an interval everywhere internally. If this does go to an interval I'd want to keep similar sanity-bounding as exists now (the heartbeat interval could not be longer than say 1/3 of the TTL).

PR #3498, opened 2018-04-09
Return all of a workflow's labels when updated via PATCH. Closes #2953

### LINKED ISSUE #2953: Return all labels when using the PATCH labels endpoint
currently it just returns the updated ones.

this is needed for job manager

https://www.youtube.com/watch?v=1Isjgc0oX0s

geoffjentry:


engine/src/main/scala/cromwell/webservice/LabelsManagerActor.scala
```diff
@@ -50,19 +46,49 @@ class LabelsManagerActor(serviceRegistryActor: ActorRef) extends Actor with Acto
...
+      /*
+        There's some trickery going on here. We've updated the labels in the metadata store but almost certainly when
+        the store received the GetLabels request above the summarizer will not have been run so our new values are
+        not present. Instead we'll fake it by manually merging what the metadata store has, as well as the new updated
+        values.
+
+        There's a potential flaw here in that a second PATCH request to the labels endpoint could come in prior to
+        summarization and thus the response to the user would lose the previous updates, but we can cross that bridge
+        if it becomes a problem. So if you find yourself debugging why previously updated labels aren't showing up in
+        the return packet, this is a likely cause.
```
rexwangcc: Thanks for the docstring here, they would be really helpful if we run into this kind of issues.

PR #3512, opened 2018-04-12
Submit workflow without starting it

salonishah11:
This will allow the user to submit a workflow but in a state where it will not automatically be picked up for execution. This new state is called 'On Hold'. A new API end point 'release' is added which will allow the user to send a signal to Cromwell to allow a workflow to be startable, at which point it will be picked up by normal execution schemes.


geoffjentry:
You'll need to also add your API changes to `cromiam.yaml` as well as `cromwell.yaml`, they're sadly not kept in synch automatically

engine/src/main/resources/swagger/cromwell.yaml
```diff
@@ -229,6 +239,29 @@ paths:
       security:
         - google_oauth:
             - openid
+  '/api/workflows/{version}/{id}/release':
```
cjllanwarne: On reflection I'm not 100% sure on `release` - it almost sounds like we're asking Cromwell to release the workflow (rather than release the hold)

FWIW from thesaurus.com...:
screen shot 2018-04-13 at 11 29 04 am

geoffjentry: `/api/workflows/{version}/{id}/let%20go` FTW :P

.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -1,6 +0,0 @@
-<component name="InspectionProjectProfileManager">
-  <profile version="1.0">
-    <option name="myName" value="Project Default" />
-    <inspection_tool class="TypeAnnotation" enabled="false" level="WARNING" enabled_by_default="false" />
-  </profile>
-</component>
```
Horneth: You might want to put this back

engine/src/main/resources/swagger/cromwell.yaml
```diff
@@ -105,6 +110,11 @@ paths:
           required: true
           type: file
           in: formData
+        - name: workflowOnHold
+          description: Put workflow on hold upon submission?
+          required: false
```
cjllanwarne: Does swagger allow us to say what the default is (presumably `false`?)

database/sql/src/main/scala/cromwell/database/slick/tables/WorkflowStoreEntryComponent.scala
```diff
@@ -98,7 +98,8 @@ trait WorkflowStoreEntryComponent {
         // The current code only writes heartbeats on initial pickup; this needs to be fixed by instrumenting workflow
         // heartbeats into Cromwell appropriately.  But in the meantime any other Cromwell's workflows will appear
         // to be abandoned if they were picked up before `heartbeatThreshold`.
-        if row.heartbeatTimestamp.isEmpty || row.heartbeatTimestamp < heartbeatThreshold
+//        _ = println(row.heartbeatTimestamp.toString + row.workflowState.toString)
```
Horneth: remove

engine/src/main/resources/swagger/cromwell.yaml
```diff
@@ -26,6 +26,11 @@ paths:
           required: true
           type: file
           in: formData
+        - name: workflowOnHold
+          description: Put workflow on hold upon submission?
```
Horneth: There's no need for `?` I think

engine/src/main/resources/swagger/cromwell.yaml
geoffjentry: FWIW I've had bad luck w/ defaults & enums in swagger, so if this route is tried make sure that it actually works as expected

engine/src/main/resources/swagger/cromwell.yaml
cjllanwarne: I think `releaseHold` would also work

engine/src/main/resources/swagger/cromwell.yaml
```diff
@@ -105,6 +110,11 @@ paths:
...
```
Horneth: Same here for `?`

engine/src/main/scala/cromwell/engine/workflow/workflowstore/InMemoryWorkflowStore.scala
```diff
@@ -16,10 +17,15 @@ class InMemoryWorkflowStore extends WorkflowStore {
     * Adds the requested WorkflowSourceFiles to the store and returns a WorkflowId for each one (in order)
     * for tracking purposes.
     */
-  override def add(sources: NonEmptyList[WorkflowSourceFilesCollection])(implicit ec: ExecutionContext): Future[NonEmptyList[WorkflowId]] = {
-    val submittedWorkflows = sources map { SubmittedWorkflow(WorkflowId.randomId(), _) -> WorkflowStoreState.Submitted }
+  override def add(sources: NonEmptyList[WorkflowSourceFilesCollection])(implicit ec: ExecutionContext): Future[NonEmptyList[WorkflowSubmissionResponse]] = {
+    val actualWorkflowState = if (sources.head.workflowOnHold)
+      WorkflowStoreState.OnHold
+    else WorkflowStoreState.Submitted
```
Horneth: I'd put the whole `if` `else` on one line or have them both on separate lines:
```
if (sources.head.workflowOnHold)
   WorkflowStoreState.OnHold
else
   WorkflowStoreState.Submitted
```

engine/src/main/resources/swagger/cromwell.yaml
mcovarr: I'd probably take `release` over any of these suggestions except possibly `releaseHold`

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreSubmitActor.scala
```diff
@@ -71,11 +74,20 @@ final case class WorkflowStoreSubmitActor(store: WorkflowStore, serviceRegistryA
   
   override def receive = workflowStoreReceive.orElse(monitoringReceive)
 
-  private def storeWorkflowSources(sources: NonEmptyList[WorkflowSourceFilesCollection]): Future[NonEmptyList[WorkflowId]] = {
+  private def getWorkflowState(workflowStoreState: WorkflowStoreState): WorkflowState ={
```
Horneth: Maybe rename this something like `databaseStateToApiState`

engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
```diff
@@ -58,6 +60,9 @@ object PartialWorkflowSources {
 
     val partialSources: ErrorOr[PartialWorkflowSources] = {
       def getStringValue(key: String) = formData.get(key).map(_.utf8String)
+      def getBooleanValue(key: String) = getStringValue(key) map { b =>
+        Try(b.toBoolean).toErrorOr
+      }
```
Horneth: I think this could be `getStringValue(key).map(Try(_.toBoolean).toErrorOr)`

database/sql/src/main/scala/cromwell/database/slick/WorkflowStoreSlickDatabase.scala
```diff
@@ -110,4 +110,13 @@ trait WorkflowStoreSlickDatabase extends WorkflowStoreSqlDatabase {
     val action = dataAccess.workflowStoreStats.result
     runTransaction(action) map { _.toMap }
   }
+
+  override def setOnHoldToSubmitted(workflowId: String)(implicit ec: ExecutionContext): Future[Unit] = {
+    val action = for {
+        updated <- dataAccess.workflowForIdAndOnHold(workflowId).update(WorkflowStoreState.Submitted)
+        _ <- if (updated == 0) DBIO.failed(new Exception(s"Couldn't change status of workflow $workflowId to 'Submitted' because the workflow is not in 'On Hold' state")) else DBIO.successful(())
```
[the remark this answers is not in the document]
Horneth: Technically true but the API validates the workflow ID first and check it exists before sending it here

.idea/inspectionProfiles/Project_Default.xml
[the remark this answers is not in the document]
salonishah11: I did not change it. Somehow it was automatically deleted when I build the project for the first time. I will put it back.

engine/src/main/resources/swagger/cromwell.yaml
salonishah11: I will rename it to 'releaseHold'

database/sql/src/main/scala/cromwell/database/slick/tables/WorkflowStoreEntryComponent.scala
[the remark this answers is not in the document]
salonishah11: Missed this while cleaning up the code. Will remove it.

.idea/inspectionProfiles/Project_Default.xml
salonishah11: Will put it back. It was somehow removed when I built the project first time.

database/sql/src/main/scala/cromwell/database/slick/tables/WorkflowStoreEntryComponent.scala
salonishah11: Missed it while cleaning up. Will remove it.

engine/src/main/resources/swagger/cromwell.yaml
salonishah11: Removed '?' 👍

engine/src/main/resources/swagger/cromwell.yaml
salonishah11: Removed '?' 👍

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreSubmitActor.scala
salonishah11: Renamed it to 'convertDatabaseStateToApiState'

engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
salonishah11: I tried the way you suggested. It could not resolve 'toBoolean' function. So I have changed it to getStringValue(key).map(b => Try(b.toBoolean).toErrorOr)

engine/src/main/resources/swagger/cromwell.yaml
salonishah11: I tried adding a default value, but it doesn't work. I also tried using enum along with default but that doesn't work too. So I have added comment in description of parameter (in Swagger UI) saying that default behavior is false.

docs/api/RESTAPI.md
```diff
@@ -554,6 +556,48 @@ Retrieve the outputs for the specified workflow. Cromwell will return any output
...
+|**Path**|**version**  <br>*required*|Cromwell API Version|string|`"v1"`|
+
+
+#### Responses
+
+|HTTP Code|Description|Schema|
+|---|---|---|
+|**200**|Successful Request|[WorkflowIdAndStatus](#workflowidandstatus)|
+|**400**|Malformed Workflow ID|No Content|
+|**403**|Workflow in terminal status|No Content|
```
Horneth: Is this true ? I think it's returning 400 now if the workflow is not `onHold`
Returning 403 might actually be better

database/sql/src/main/scala/cromwell/database/slick/WorkflowStoreSlickDatabase.scala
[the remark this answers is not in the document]
mcovarr: a very small window for a race condition 🙂

docs/api/RESTAPI.md
salonishah11: Makes sense. Changed it to 403 and pushed the changes.

PR #3513, opened 2018-04-13
Pipelines API Version split

Horneth:
This is almost only moving around / splitting / renaming stuff.
The pieces to review are the `build.sbt` and maybe `PipelinesApiBatchHandler`, `PipelinesApiFactoryInterface`, and `PipelinesApiRequestFactory` and their implementation under `v1alpha2`. Those are the interfaces for which implementation will be different in v1 and v2


core/src/main/scala/cromwell/util/StopAndLogSupervisor.scala
```diff
@@ -10,6 +10,7 @@ trait StopAndLogSupervisor { this: Actor =>
   final val stopAndLogStrategy: SupervisorStrategy = {
     def stoppingDecider: Decider = {
       case e: Exception =>
+        e.printStackTrace()
```
geoffjentry: I'm a little confused by this whole class. Despite the name it doesn't seem to have been doing any degree of logging previously. And now there's the `printStackTrace`, but does that actually get captured by logback, or only to the console?

Horneth: Oh right yeah this is me debugging. I'm going to try to find a way to log it properly though because when the worker actor crashes it doesn't say why which is annoying

supportedBackends/google/pipelines/common/src/main/scala/cromwell/backend/google/pipelines/common/api/clients/PipelinesApiAbortClient.scala
```diff
@@ -0,0 +1,39 @@
...
+  
+  def abortJob(jobId: StandardAsyncJob) = {
+    papiApiActor ! PAPIAbortRequest(workflowId, self, requestFactory.abortRequest(jobId), jobId)
+  }
+
+  def abortActorClientReceive: Actor.Receive = {
+    case PAPIAbortRequestSuccessful(jobId) =>
+      abortSuccess()
+      jobLogger.info(s"Successfully requested cancellation of $jobId")
+    // In this case we could immediately return an aborted handle and spare ourselves a d round of polling
```
mcovarr: not sure what "a d round" is but it's here and below

supportedBackends/google/pipelines/v1alpha2/src/main/scala/cromwell/backend/google/pipelines/v1alpha2/api/StatusRequestBatchHandler.scala
```diff
@@ -146,4 +145,4 @@ private[statuspolling] object StatusPolling {
   private def eventIfExists(name: String, metadata: Map[String, AnyRef], eventName: String): Option[ExecutionEvent] = {
     metadata.get(name) map { time => ExecutionEvent(eventName, OffsetDateTime.parse(time.toString)) }
   }
-}
+}
```
mcovarr: \n

cromwell.examples.conf
```diff
@@ -566,7 +566,7 @@ backend {
     # }
 
     #JES {
```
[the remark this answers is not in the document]
Horneth: Yeah there's still a lot of "JES" around. I renamed another bunch in the next PR

Issue #3092, opened 2018-01-02
Detritus in execution directory


Horneth:
The CWL conformance tests have a new addition that is globbing everything in the current directory and expecting files based on what the command produced. It fails because in addition to those files Cromwell also has the script file, rc file etc...

geoffjentry:
@Horneth I did see them talking about adding it in gitter, it was to provide a test for a bug which was found in toil. we should probably ask what the bug was and see if perhaps we're failing it for the same reason or if this was just an unfortunate side effect.

mr-c:
Huh, I thought we were explicit about the execution directory being empty in http://www.commonwl.org/v1.0/CommandLineTool.html#Runtime_environment 

I've made an issue on our end to document this assumption: https://github.com/common-workflow-language/common-workflow-language/issues/579

`OutputBinding: { glob: * }` is fairly common in CWL

ruchim:
@mcovarr did this work get completed? If yes, would you mind closing the issue? if no, any idea on where this last stands? Thanks!

mcovarr:
No this is not completed.

PR #3516, opened 2018-04-17
Enumerations, SchemaDefRequirement and fix for nested Inputbindings

danbills:
Borne of a requirement on GDC Dnaseq using Enumerations.

Represents a spike on Test \#3 which was the only one that tested Enumerations at the time of this writing.

A list of features/fixes

* implementation of SchemaDefRequirement.  This produced a lot of noise as any time `Myriad` types were used this must be provided for reference.
* nested input bindings sort properly. 
* impl of EnumSchema for input & output
* fix for coproduct of types -> command parts ( previously was using values from fields that matched any field of any coproduct type.  this fix regards each type holistically and thus rejects unless all fields match)
* Introduced a `RequirementsAndHints` class to handle common operations on `List[Requirement]`.

## ~TODO~

~I broke conf test \#71 and centaur test dealing w/ secondary files.  I expect fixes to be minor.~


cwl/src/main/scala/cwl/MyriadInputTypeToSortedCommandParts.scala
```diff
@@ -72,18 +88,20 @@ object MyriadInputTypeToSortedCommandParts extends Poly1 {
...
 
   // Primitive type: we just need to create a command part from the binding if there's one here.
   implicit def ct: Aux[CwlType, CommandPartBuilder] = {
     at {
       cwlType => {
-        (inputBinding, womValue, key, hasShellCommandRequirement, expressionLib) => {
-          cwlType match {
-            case CwlType.Null => CommandPartsList.empty
-            case _ => inputBinding.toList.map(_.toCommandPart(key, womValue, hasShellCommandRequirement, expressionLib))
+        case (_, WomOptionalValue(_, None), _, _, _, _) => CommandPartsList.empty.some
```
mcovarr:
```
scala> import cats.syntax.option._
import cats.syntax.option._

scala> val s: String = null
s: String = null

scala> s.some
res0: Option[String] = Some(null)

scala> Option(s)
res1: Option[String] = None
```

😞

cwl/src/main/scala/cwl/MyriadInputTypeToSortedCommandParts.scala
```diff
@@ -94,111 +112,144 @@ object MyriadInputInnerTypeToSortedCommandParts extends Poly1 {
...
+      case (None, _, _, _, _, _) if !irs.fields.exists(_.exists(_.inputBinding.isDefined)) => CommandPartsList.empty.some
 
-      case (inputBinding, objectLike: WomObjectLike, sortingKey, hasShellCommandRequirement, expressionLib) =>
+      case (inputBinding, objectLike: WomObjectLike, sortingKey, hasShellCommandRequirement, expressionLib, schemaDefRequirement) =>
         // If there's an input binding, make a SortKeyAndCommandPart for it
-        val fromInputBinding: Option[SortKeyAndCommandPart] =
+        val sortingKeyFromInputBindingFromInputParameter: Option[SortKeyAndCommandPart] =
           inputBinding.map(_.toCommandPart(sortingKey, objectLike, hasShellCommandRequirement, expressionLib))
 
         // Go over the fields an fold over their type
```
mcovarr: and fold

cwl/src/main/scala/cwl/internal/CommandPartSortingAlgorithm.scala
```diff
@@ -0,0 +1,71 @@
+package cwl.internal
+
+import cats.data.Kleisli._
+import cats.data.ReaderT
+import cats.data.Validated.{Invalid, Valid, _}
```
mcovarr: aka _

centaur/src/main/resources/standardTestCases/cwl_secondary_files/cwl_secondary_files.cwl
```diff
@@ -29,6 +29,7 @@ $graph:
       secondaryFiles: [.also]
       type:
         type: record
+        name: foo
```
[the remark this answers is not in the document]
danbills: Hm, I guess this should be optional in `InputRecordSchema`.  If you look there you'll see I added a `name` field.  This is contentious as `name` field appeared in conformance tests but not the spec!  I'm off to add a comment in the code.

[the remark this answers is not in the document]
danbills: So it is non-optional in v1.1 so I've kept it as such. Thus this data remains to conform to the (pseudo-)spec.

cwl/src/main/scala/cwl/MyriadInputTypeToSortedCommandParts.scala
danbills: I get your point, but this instance is guaranteed non-null.  I'm working on a `.option` PR for cats for general usage.

danbills: https://github.com/typelevel/cats/pull/2235

cwl/src/main/scala/cwl/CommandOutputBinding.scala
```diff
@@ -220,7 +220,13 @@ object CommandOutputBinding {
         val globs = List(cwlPath)
         globs traverse loadFileWithContents(ioFunctionSet, commandOutputBinding)
       case WomMaybePopulatedFileType =>
-        val globs = Await.result(ioFunctionSet.glob(cwlPath), Duration.Inf)
+        val globs: Seq[String] = Await.result(ioFunctionSet.glob(cwlPath), Duration.Inf).
+
+        //TODO: HACK ALERT - DB: I am starting on ticket https://github.com/broadinstitute/cromwell/issues/3472 which will redeem me of this mortal sin.
+          filterNot{s =>
+            s.endsWith("rc.tmp") || s.endsWith("docker_cid") || s.endsWith("script") || s.endsWith("script.background") || s.endsWith("script.submit") || s.endsWith("stderr") || s.endsWith("stderr.background") || s.endsWith("stdout") || s.endsWith("stdout.background")
```
danbills: I did in fact push this code on purpose to see if it causes damage or even passes the detritus test.  It got me over the hump on GDC bamfastq_align. I am starting ticket #3092 to provide an actual solution.

PR #3527, opened 2018-04-17
Annotate implicits in slow-building classes

aednichols:


danbills:
It's hard to overstate how happy this PR makes me

aednichols:
Unfortunately it looks like removing those implicits reversed the speed gain
```
Total time: 179 s, completed Apr 18, 2018 11:29:35 AM
```

danbills:
looks like we're hitting this: https://github.com/circe/circe/issues/575

aednichols:
How incredibly specific, nice find

aednichols:
I tried the `semiauto` + `deriveDecoder` + type annotation suggestion in the Circe link - it compiles and passes tests, but the speed improvement goes away

I am going to put those particular implicits aside for now and measure whether the other (easy) changes make a worthwhile difference - I suspect they do

danbills:
Summarizing my points from this morning:

* The great gains were had because the derived encoders weren't being derived.
* The type annotations satisfy the implicit search before values are initialized, resulting in `null` values.
* The arrangement we have currently is almost certainly the best we can do.

aednichols:
Yes, it's amazing what you can accomplish when you accidentally skip a big part of the binary 😄

PR #3505, opened 2018-04-11
Remaining value evaluation inc. read/write functions

cjllanwarne:
* Wiring in the remaining draft 3 engine functions' value evaluators
* Modified value evaluation to keep a running track of its side-effect file generation
* Also allowed value evaluation to run value mapping if used in placeholder interpolation

Note: red thumb required because there's some changes to WomIdentifier undoing some of the changes in the forkjoin PR and making them again, but correcter this time. 

This PR ended up with far wider-ranging changes than I originally expected (even requiring changes in draft 2 and CWL) so if you see something that smells wrong... ⚠️ ❗️ ❗️ ⚠️ 

~~NB: this PR will go down by about 700 lines once the `forkjoin` PR merges~~ DONE!

PR #3530, opened 2018-04-18
Make sure callable importing works

cjllanwarne:
- [ ] Needs a rebase on `develop` after #3505 

Wires in and checks imported tasks, workflows, namespaces and aliases


centaur/src/main/resources/standardTestCases/wdl_draft3/callable_imports/my_workflow.wdl
```diff
@@ -0,0 +1,10 @@
+version draft-3
+
+workflow my_workflow {
+  input{
```
aednichols: [optional / pedantic] missing space messes up github highlighting (or maybe github highlighting is what needs an update?)

cjllanwarne: A little bit of both I think but github highlighting is based on our sublime syntax - https://github.com/broadinstitute/wdl-sublime-syntax-highlighter - and probably needs an update (it also needs an update for draft-3 now that I think of it...)

centaur/src/main/resources/standardTestCases/wdl_draft3/callable_imports/callable_imports.wdl
```diff
@@ -0,0 +1,19 @@
+version draft-3
+
+import "my_task.wdl"
+import "my_workflow.wdl" as mai_workflau
+
+workflow callable_imports {
```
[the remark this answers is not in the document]
cjllanwarne: I like the idea but I'd prefer to do this as a full scale renaming of all the draft 3 tests.

Right now if they workflow is called `x` then the test name is always `draft3_x` so fixing only this one would actually make things less consistent short-term

PR #3531, opened 2018-04-18
Scala two evel-gone

cjllanwarne:


project/Settings.scala
```diff
@@ -104,13 +102,10 @@ object Settings {
...
-      case Some((2, 11)) =>
-        // Scala 2.11 takes a simplified set of options
-        baseSettings
       case _ =>
         throw new NotImplementedError(
           s"Found unsupported Scala version '${scalaVersion.value}'." +
-            s" ${name.value} does not support versions of Scala other than 2.11 or 2.12.")
+            s" ${name.value} does not support versions of Scala other than 2.12.")
     }),
     // http://stackoverflow.com/questions/31488335/scaladoc-2-11-6-fails-on-throws-tag-with-unable-to-find-any-member-to-link#31497874
```
cjllanwarne: Wasn't sure if this was 2.11 specific?

project/Settings.scala
```diff
@@ -130,17 +125,16 @@ object Settings {
      */
     coverageEnabled := (CrossVersion.partialVersion(scalaVersion.value) match {
       case Some((2, 12)) => sys.env.get("ENABLE_COVERAGE").exists(_.toBoolean)
```
cjllanwarne: There's a comment here about "once 2.11 is gone we can..." - but I wasn't sure how to action that

geoffjentry:
Make a note in the changelog that 2.11 support is gone

project/Settings.scala
kshakir: The comment is referring to any line in the codebase using `ENABLE_COVERAGE`. Each case should be cleaned up to remove this code.

FYI we will likely be putting this all back shortly due to 2.13 coming out very soon, but that can just be a revert of this PR.

cjllanwarne: So... you want me to remove this line and with it the option to not generating coverage with sbt?

Eg, right now my laptop won't generate coverage because that env variables isn't set - isn't that what we want?

kshakir: Nah, don't remove it on my account. I was only referring to "if one wanted to remove this code..." As I said I think this will all brb.

If you run `sbt test` it won't generate coverage for you. But if you run `test.sh` coverage will be generated as that script has the `ENABLE_COVERAGE` embedded within its layers.

PR #3532, opened 2018-04-18
Fix SFS hard linking

mcovarr:
🙈


cjllanwarne:
ToL: is it possible to test that this doesn't unfix itself in the future?

cjllanwarne:
Looks pretty show-stopping... how did we not hit this before (because we don't usually hard link?)

mcovarr:
Jeff and I have been hitting this nonstop, we just didn't realize what it was saying. 😬 It's SFS only so that limits the visibility.

geoffjentry:
Yeah, I didn't notice it until I started having trouble due to our VM shenanigans over the weekend and at that point I was thinking it was an artifact of all that

mcovarr:
Detritus still errors on hard linking, I think we may not have created the execution directory first.

mcovarr:
It looks like Chris is going to get his wish for tests since this fix seems to have broken some of the existing SFS hard linking tests. 😮

mcovarr:
Needs more investigation. Something is definitely not right here but not sure this is it.

mcovarr:
So it appears my proposed changes here fixed call cache hit copy hard linking (except for detritus) but broke input localization hard linking. More investigation needed...

PR #3536, opened 2018-04-19
Disallow unwanted inputs specified in call body

cjllanwarne:


wdl/model/draft2/src/main/scala/wdl/draft2/model/Declaration.scala
```diff
@@ -149,6 +149,7 @@ object Declaration {
       case RequiredInputDefinition(_, womType, _) => RequiredGraphInputNode(decl.womIdentifier, womType, decl.womIdentifier.fullyQualifiedName.value)
       case OptionalInputDefinition(_, optionalType, _) => OptionalGraphInputNode(decl.womIdentifier, optionalType, decl.womIdentifier.fullyQualifiedName.value)
       case InputDefinitionWithDefault(_, womType, default, _) => OptionalGraphInputNodeWithDefault(decl.womIdentifier, womType, default, decl.womIdentifier.fullyQualifiedName.value)
+      case other => throw new RuntimeException(s"Programmer Error! Draft 2 isn't set up to produce or handle ${other.getClass.getSimpleName}")
```
aednichols: [ToL] Could we craft a marginally more helpful message? It seems possible, perhaps even likely, that the user is inadvertently attempting to use a draft-3 feature, so maybe we should say that

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wom/graph/CallElementToGraphNode.scala
```diff
@@ -38,10 +38,14 @@ object CallElementToGraphNode {
...
     def expressionNodeMappings(callable: Callable): ErrorOr[Map[LocalName, AnonymousExpressionNode]] = {
+      def validInput(name: String, definition: Callable.InputDefinition): Boolean = {
+        definition.name == name && !definition.isInstanceOf[FixedInputDefinition]
+      }
+
       a.node.body match {
         case Some(body) =>
           body.inputs.map(input => input.key -> input.value).toMap.traverse {
-            case (name, expression) =>
+            case (name, expression) if callable.inputs.exists(validInput(name, _)) =>
```
aednichols: [ToL] It wasn't immediately clear to me what value this `_` binds to - the result of the `traverse`? It has some competing `_`'s in close proximity, so it might be nice to give this one a name

cjllanwarne: Sure! In fact it was the `.exists` rather than the `traverse` - which IMO reinforces your point that it was non-obvious!

PR #3539, opened 2018-04-20
Decoupling custom labels and Google Labels; Removing restrictions on custom labels

salonishah11:
Custom labels will not be pushed to Google Cloud. Also, the restrictions on custom labels are removed, except the size which is now increased to 255. Now user can enter anything as key and value.


geoffjentry:
Make sure to make a note of this in the `CHANGELOG.md` file as well as modify `Labels.md` as appropriate

salonishah11:
@Horneth I did rebase it. But I think the problem was that initially I had created my branch based off your branch for PAPI, so I think it somehow added all those commits.

core/src/test/scala/cromwell/core/labels/LabelSpec.scala
```diff
@@ -13,21 +13,30 @@ class LabelSpec extends FlatSpec with Matchers {
...
+    "now valid 255 character key-at vero eos et accusamus et iusto odio dignissimos ducimus qui blanditiis praesentium voluptatum deleniti atque corrupti quos dolores et quas molestias excepturi sint occaecati cupiditate non provident, similique sunt in culpa",
+    "!@#$%^&*()_+={}[]:;'<>?,./`~"
   )
 
   val goodLabelValues = List(
     "11f2468c-39d6-4be3-85c8-32735c01e66b",
-    ""
+    "",
+    "!@#$%^&*()_+={}[]:;'<>?,./`~",
+    "now valid 255 character value-at vero eosd accusamus et iusto odio dignissimos ducimus qui blanditiis praesentium voluptatum deleniti atque corrupti quos dolores et quas molestias excepturi sint occaecati cupiditate non provident, similique sunt in culpa",
```
mcovarr: Should be "eos" (no 'd' at the end) - though if this is only a long string test and not a match test then that isn't really an issue, except for my Ancient Roman OCD.

supportedBackends/google/pipelines/v1alpha2/src/main/scala/cromwell/backend/google/pipelines/v1alpha2/api/StatusRequestBatchHandler.scala
```diff
@@ -145,4 +145,4 @@ private[api] object StatusRequestBatchHandler {
   private def eventIfExists(name: String, metadata: Map[String, AnyRef], eventName: String): Option[ExecutionEvent] = {
     metadata.get(name) map { time => ExecutionEvent(eventName, OffsetDateTime.parse(time.toString)) }
   }
-}
+}
```
mcovarr: revert change

PR #3541, opened 2018-04-21
CWL Call Caching

mcovarr:
Fully caches Genome in a Bottle chm to run in ~2 hours. Still waiting on results from GIAB Joint because that takes much longer and has to be restarted for sporadic failures, but all the calls so far will cache.


cwl/src/main/scala/cwl/CommandLineTool.scala
```diff
@@ -381,7 +381,29 @@ object CommandLineTool {
...
+    //
+    // CommandOutputParameter(file:///var/folders/qh/vvrlr2q92mvb9bb45sttxll4y3gg2g/T/e17762d1-9921-46c5-833e-cb47e3c3bdfd.temp.1197611185902619026/e17762d1-9921-46c5-833e-cb47e3c3bdfd.cwl#ps/ea5165fc-8948-43ae-8dec-3c0468b56bcb/ps-stdOut,None,None,None,None,None,Some(CommandOutputBinding(Some(Inl(Inr(Inl(ps-stdOut.txt)))),None,None)),Some(Inl(Inl(File))))
+    //
+    // after:
+    //
+    // CommandOutputParameter(ps-stdOut,None,None,None,None,None,Some(CommandOutputBinding(Some(Inl(Inr(Inl(ps-stdOut.txt)))),None,None)),Some(Inl(Inl(File))))
+    //
+    // This is possibly too strict (i.e. some of these fields may be irrelevant for cache hit determination), but preferable
+    // to having false positives.
+    // Also two coproduct types that can be either single or Arrays have custom stringifying folds for arrays.
```
kshakir: Use `/* */` instead of multiple `//`, or even better yet for this method `/** */`.

cwl/src/main/scala/cwl/model.scala
```diff
@@ -90,6 +91,60 @@ case class ArgumentCommandLineBinding(
...
+
+object SecondaryFilesCacheableString extends Poly1 {
+  import Case._
+
+  implicit def one: Aux[StringOrExpression, String] = at[StringOrExpression] {
+    _.toString
+  }
+
+  implicit def array: Aux[Array[StringOrExpression], String] = at[Array[StringOrExpression]] {
+    "Array(" + _.mkString(",") + ")"
```
kshakir: I'll leave it to your preference, but there's also: 

```scala
_.mkString("Array(", ",", ")")
```

wom/src/main/scala/wom/types/WomFloatType.scala
```diff
@@ -11,8 +11,11 @@ case object WomFloatType extends WomPrimitiveType {
   override protected def coercion = {
     case f: Float => WomFloat(f.toDouble)
     case d: Double => WomFloat(d)
+    case i: Integer => WomFloat(i.toDouble)
     case n: JsNumber => WomFloat(n.value.doubleValue())
     case f: WomFloat => f
+    case i: WomInteger => WomFloat(i.value.toDouble)
+    case l: WomLong => WomFloat(l.value.toDouble)
```
kshakir: I personally love this, but I think the above breaks the WDL draft-2 spec as ints cannot be cast to floats. (And it doesn't even mention "long" so...)

https://github.com/openwdl/wdl/blob/master/versions/draft-2/SPEC.md#type-coercion

cc: @cjllanwarne @aednichols for their thoughts.

PR #3545, opened 2018-04-23
Wire in draft3 globs

cjllanwarne:


wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/linking/expression/values/EngineFunctionEvaluators.scala
```diff
@@ -443,6 +443,16 @@ object EngineFunctionEvaluators {
     }
   }
 
+  implicit val globFunctionValueEvaluator: ValueEvaluator[Glob] = (a, inputs, ioFunctionSet, forCommandInstantiationOptions) => {
+    processValidatedSingleValue[WomString, WomArray](a.param.evaluateValue(inputs, ioFunctionSet, forCommandInstantiationOptions)) { globString =>
+      for {
+        globbed <- Try(Await.result(ioFunctionSet.glob(globString.valueString), ReadWaitTimeout)).toErrorOr
```
kshakir: `Try(block).toErrorOr` can be written more compactly as `validate(block)`.

https://github.com/broadinstitute/cromwell/blob/31/common/src/main/scala/common/validation/Validation.scala#L29-L32

PR #3547, opened 2018-04-23
Test struct localization and delocalization

cjllanwarne:


centaur/src/main/resources/standardTestCases/wdl_draft3/struct_localize_delocalize/struct_localize_delocalize.wdl
```diff
@@ -0,0 +1,51 @@
...
+struct FileContainer {
+  File bam
+  File index
+}
+
+workflow struct_localize_delocalize {
+  call struct_delocalize
+  call struct_localize { input: container = struct_delocalize.fc }
+
+  scatter(s in struct_localize.found_files) {
```
mcovarr: highlighting's not right

cjllanwarne: Huh, that’s the second time I’ve seen this recently. Either something changed or I just haven’t submitted much WDL recently...

PR #61, opened 2015-06-23
Adding in first slick implementation

kshakir:

Issue #2423, opened 2017-07-06
Convert to using Circe


geoffjentry:
We seem to be gravitating towards circe in general, convert Cromwell from spray-json to circe

katevoss:
@geoffjentry what is Circe? What effort would this take?

geoffjentry:
https://github.com/circe/circe

it wouldn't be a huge project but it's also not a huge need. since we're using circe everywhere else now it just means we have two different ways of doing the same thing

kshakir:
Related to switching to circe, and printing JSON in general, pretty-vs-compact printing has been up for debate.

Here is an example of how to pretty print in circe: https://github.com/broadinstitute/clio/commit/4b78f9a5d2f5a0aea389dc404859d5f91e07321c#diff-4543387a548c3521400bdaa006943757

The above always pretty-prints. It may be the case that one wants the pretty-vs-compact printing to be an [option](https://www.elastic.co/guide/en/elasticsearch/reference/current/common-options.html). Because this particular solution uses implicits, one _could_ turn this

```scala
implicit val akkaHttpJsonPrinter: Printer
```

into

```scala
implicit def akkaHttpJsonPrinter()(implicit cromwellPrintType: CromwellPrintType): Printer
```

and then implicitly provide an enum set by a query string parameter.

PR #3551, opened 2018-04-24
Conformance tests #60 and #61 - Myriad Javascript assertions

danbills:
These tests assert ~ 50 different nuances of JS evaluation.


centaurCwlRunner/src/main/scala/centaur/cwl/OutputManipulator.scala
```diff
@@ -144,6 +144,17 @@ object OutputManipulator extends Poly1 {
...
       case (JsString(metadata), Inl(CwlType.String)) => metadata.asJson
+
+      //The Anys.  They have to be done for each type so that the asJson can use this type information when going back to Json representation
+      case (JsString(metadata), Inl(CwlType.Any)) => metadata.asJson
+      case (JsNumber(metadata), Inl(CwlType.Any)) => metadata.asJson
+      case (obj: JsObject, Inl(CwlType.Any)) =>
+        import io.circe.parser._
+        parse(obj.compactPrint).right.getOrElse(throw new Exception("Failed to parse Json output as Json... something is very wrong"))
+      case (JsBoolean(metadata), Inl(CwlType.Any)) => metadata.asJson
+      case (JsArray(metadata), Inl(CwlType.Any)) => metadata.asJson
```
mcovarr: it seems like most of these just `unapply` from specific `JsFoo` types and then re-JSONify them? I think the comment is trying to tell me otherwise but I'm not getting it... 😢

centaurCwlRunner/src/main/scala/centaur/cwl/OutputManipulator.scala
```diff
@@ -162,9 +173,10 @@ object OutputManipulator extends Poly1 {
           fields <- schema.fields
           typeMap = fields.flatMap(processField).toMap
           outputJson = metadata.map({
-            case (k, v) => k -> resolveOutputViaInnerType(typeMap(k))(v, pathBuilder)
+            case (k, v) =>
+              k -> resolveOutputViaInnerType(typeMap(k))(v, pathBuilder)
```
mcovarr: 🔙

centaurCwlRunner/src/main/scala/centaur/cwl/OutputManipulator.scala
kshakir: If you stare too closely at the ~sun~ codebase you'll find a number of places with "SprayJson <-> Circe" conversions like this. In this case it's because of `case class centaur.test.metadata.WorkflowMetadata(value: Map[String, JsValue])`.

New code like this OutputManipulator has been using Circe, while we still have SprayJson sprinkled inside classes all over. Someday perhaps a SOC refactoring should pull the SprayJson _and_ Circe code out into wrappers, otherwise I foresee "SprayJson <-> Circe <-> (insert next scala-json hotness here)" conversions in our future.

geoffjentry: Or someone does this ticket :)

https://github.com/broadinstitute/cromwell/issues/2423

PR #3558, opened 2018-04-25
Create temp dirs for WomMaybePopulatedDirectories otherwise lacking values.

mcovarr:
This would fix 116 too if that weren't using `alpine` for its `dockerPull`; I'm looking at that separately.


cwl/src/main/scala/cwl/CwlType.scala
```diff
@@ -235,12 +235,8 @@ case class Directory private
 
   lazy val asWomValue: ErrorOr[WomFile] = {
     errorOrListingOption flatMap { listingOption =>
-      val valueOption = path.orElse(location).orElse(basename)
-      (valueOption, listingOption) match {
-        case (None, None) =>
-          "Cannot convert CWL File to WomValue without either a location, a path, a basename, or a listing".invalidNel
-        case (_, _) => WomMaybeListedDirectory(valueOption, listingOption).valid
-      }
+      val valueOption = path.orElse(location).orElse(basename).orElse(Option(better.files.File.newTemporaryDirectory().pathAsString))
```
kshakir: This leaks files (technically directories), and is works as a temporary patch for SFS only AFAICT.

Again, similar to Dan's PR I'm fine with it as long as it's tracked.

geoffjentry: “Tracking” generally leads to “forgotten”

cjllanwarne: Is this final `Option` intended to always be `Some`? Could the last `orElse` be a `getOrElse`?

mcovarr: Conformance 115/116 won't be happy on PAPI until this is fixed there which will warrant a more general and hopefully less horrible solution.

mcovarr: Not sure if I can de-`Option` the `valueOption`, I'll look at that.

mcovarr: I tried this and hit hanging tests in sbt that I couldn't resolve in a timely manner so leaving as is for now.

PR #3559, opened 2018-04-26
Small bundle of new centaur and unit tests for draft 3

cjllanwarne:
Red thumb required because of changes in `womtool`


wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wom/graph/CallElementToGraphNode.scala
```diff
@@ -57,12 +57,16 @@ object CallElementToGraphNode {
...
                 LocalName(name) -> _
               }
-            case (name, _) if callable.inputs.exists(i => i.name == name) =>
+            case (name, _) =>
               val callNameAlias = a.node.alias match {
                 case Some(alias) => s" (as '$alias')"
                 case None => ""
               }
-              s"Invalid call to '${callable.name}'$callNameAlias: Didn't expect the input '$name'".invalidNel
+              if (callable.inputs.exists(i => i.name == name)) {
```
kshakir: Can this be `if (callable.inputs.exists(_.name == name))`?

cjllanwarne: I'll let you fight it out with @aednichols 😄 

Funnily enough this is the exact inverse of the second comment here: https://github.com/broadinstitute/cromwell/pull/3536

kshakir: Heh. The whole reason we chose an even number of reviewers is exactly for this. In the event of a reviewer-tie the original author gets to pick what they want to do.

So change it or don't, but my 2¢ is that I don't think a single letter variable really informs the reader more that `_`, in this particular case.

geoffjentry: `_` plz

cjllanwarne: FWIW the comment here wasn't "give the variable a name", it was "make it obvious which function this `_` belongs to (is it the `traverse`)".

Since it was *not* the `traverse`, I lean towards agreeing and wanting to make it more obvious - any ideas other than naming the variable?

PR #3561, opened 2018-04-27
Configurable job shell.

mcovarr:


geoffjentry:
This seems worthy of both winding up in ReadTheFineDocs as well as the changelog

geoffjentry:
Some might argue that the birds have already struck :P

src/bin/travis/testCentaurCwlConformancePAPI.sh
```diff
@@ -85,6 +85,7 @@ java \
   -Xmx2g \
   -Dconfig.file="$JES_CONF" \
   -Dsystem.new-workflow-poll-rate=1 \
+  -Dsystem.job-shell=/bin/sh \
```
kshakir: As there is already a `-Dconfig.file` used on this command line, this configuration should be kept together with the rest inside the file `$JES_CONF` aka `src/bin/travis/resources/jes_centaur.conf.ctmpl`.

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -290,18 +290,18 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
     lazy val environmentVariables = instantiatedCommand.environmentVariables map { case (k, v) => s"""export $k="$v"""" } mkString("", "\n", "\n")
 
     val home = jobDescriptor.taskCall.callable.homeOverride.map { _ (runtimeEnvironment) }.getOrElse("$HOME")
+    val shortId = jobDescriptor.workflowDescriptor.id.shortString
+    // Give the out and error FIFO variables names that are unlikely to conflict with anything the user is doing.
+    val (out, err) = (s"out$shortId", s"err$shortId")
+    val jobShell = configurationDescriptor.globalConfig.as[Option[String]]("system.job-shell").getOrElse("/bin/bash")
```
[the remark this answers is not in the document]
mcovarr: What's a good place for this variable? In the `StandardJobExecutionActorParams` perhaps?

src/bin/travis/resources/local_conformance_expected_failures.txt
```diff
@@ -113,7 +113,7 @@
 #113
 #114
 #115
-116
+#116
```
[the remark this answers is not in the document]
mcovarr: Not sure we can do this without significantly expanding the scope of this work. Currently this is set up as a system-wide option via config and we must have `/bin/bash` for at least some of the existing Centaur WDL tests that explicitly use bashisms in the command block.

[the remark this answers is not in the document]
kshakir: Makes sense... someday... Still glad to see this positive direction!

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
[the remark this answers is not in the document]
kshakir: No, the `*ActorParams` classes were because I was sick of having to update multiple `Actor` class constructors whenever we wired in a new parameter.

I think it can be yet[[0]](https://github.com/broadinstitute/cromwell/blob/31/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala#L153) another[[1]](https://github.com/broadinstitute/cromwell/blob/31/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala#L290) lazy[[2]](https://github.com/broadinstitute/cromwell/blob/31/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala#L336) val[[3]](https://github.com/broadinstitute/cromwell/blob/31/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala#L417) in the `StandardAsyncExecutionActor`.

(Side note, every time I'm back in this code makes we wish we had more time for developer's choice especially after FP book club)

centaur/src/main/resources/standardTestCases/local_bourne/local_bourne.wdl
```diff
@@ -0,0 +1,20 @@
+task shell {
+  command {
+    # The SHELL variable isn't set in /bin/sh on alpine, busybox or ubuntu so there is this abomination instead.
+    ps auxwww | head -n 2 | tail -n 1 | cut -c23-50
```
mcovarr: @kshakir I expanded the original code a bit to be able to take `job-shell` from the backend config, and with some Centaur whispering I got an alpine/`/bin/sh` test running. The one bit that's still irksome is the "what shell is this" logic since `$SHELL` isn't being set on `/bin/sh`. Suggestions appreciated. 🙂

kshakir: Woo!! Nice to see WDL getting some bourne love. ❤️ 

I think your command is overkill though. The images are intentionally tiny and provide very little functionality beyond the POSIX spec, so they won't have your `SHELL` variable etc. But because the busybox images (including alpine) don't have bash, the fact that this runs at all is amazing. A simple `echo hello world` would suffice.

If you really want to push the limits of stress testing our POSIX support... add another call to the same wdl with [`busybox:1.28.3`](https://hub.docker.com/r/library/busybox/tags/) that's half the size of [`alpine:3.6`](https://hub.docker.com/r/library/alpine/tags/)!

centaur/src/main/resources/standardTestCases/local_bourne/local_bourne.wdl
```diff
@@ -0,0 +1,20 @@
+task shell {
+  command {
+    # The SHELL variable isn't set in /bin/sh on alpine, busybox or ubuntu so there is this abomination instead.
+    ps auxwww | head -n 2 | tail -n 1 | cut -c22-28
```
EvanTheB: Is this to discover which shell is running? I beleive this is more reliable: 
sh -c 'ps -p $$ -o ppid=' | xargs -i readlink -f /proc/\{\}/exe
https://askubuntu.com/a/1022440/382239

kshakir: Thanks for the link! Turns out `busybox` is so slim that `ps` doesn't even support `-p`. https://busybox.net/BusyBox.html#ps

Out of curiosity I tried scanning through one of the POSIX standard versions and apparently `ps -p` might technically be required: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/ps.html

I guess this [comment](https://busybox.net/BusyBox.html#description) leaves a little wiggle room?
> BusyBox provides a fairly complete POSIX environment

kshakir:
cheers and re-👍

centaur/src/main/resources/standardTestCases/local_bourne/local_bourne.wdl
EvanTheB: Ah it is in a busybox environment I did not realise. Ah well. So we just need a solution more portable that posix...

BEEDELLROKEJULIANLOCKHART: @EvanTheB, I doubt that such a solution exists. If you were unable to locate one, do you mind filing an issue with Busybox? If you were, are you able to share it with me?

PR #3562, opened 2018-04-27
Patch 3

EvanTheB:


geoffjentry:
@EvanTheB just to double check, you did this by autogenerating the docs, right? Thanks for the catch.

EvanTheB:
Hi Geoff,
I do not have a java dev environment, and did not know that the docs were generated! I just grepped for the help string. Please feel free to modify this branch

kshakir:
FYI there's a hidden watermark at the top of the file that one can use in PRs to tell if the RESTAPI.md was manually or automatically updated. Example: https://github.com/broadinstitute/cromwell/blame/31/docs/api/RESTAPI.md#L1-L8

Also if one doesn't have a dev environment locally they can still use any public sbt docker. It will take a while as it downloads ~the entire internet~ all of the un-cached cromwell dependencies, but something like this will work:

```shell
docker \
  run \
  --rm \
  -v $PWD:$PWD \
  -w $PWD \
  hseeberger/scala-sbt \
  sbt generateRestApiDocs
```

geoffjentry:
Thanks @evantheb I’ll update woth the aigogenerated version

geoffjentry:
Closing this in favor of #3573

PR #3566, opened 2018-04-27
Honor thy trailing newlines

mcovarr:
Mix one line of super obvious fix with 45 lines of conformance hash repinning noise.


geoffjentry:
👍

PR #3572, opened 2018-04-30
Placeholder attributes

cjllanwarne:
Tests validation of placeholder attributes and adds handling of the `true=` and `false=` attributes


cjllanwarne:
Another red thumb required for the `womtool` change (the final 2 files in this PR)

PR #3576, opened 2018-05-01
Publish labels upon workflow submission

salonishah11:
Labels will now be published upon workflow submission even if the workflow is On Hold or not started yet. If the labels are invalid, submission request will fail.


engine/src/main/scala/cromwell/engine/workflow/lifecycle/materialization/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -283,6 +280,15 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
     MetadataEvent(MetadataKey(workflowIdForLogging, None, WorkflowMetadataKeys.Name), MetadataValue(name))
   }
 
+  private def convertJsonToLabels(json: String): Labels = {
+    json.parseJson match {
+      case JsObject(inputs) => Labels(inputs.toVector.collect({
+        case (key, JsString(value)) => Label(key, value)
+      }))
+      case _ => Labels(Vector.empty)
```
mcovarr: what is this case supposed to represent?

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreSubmitActor.scala
```diff
@@ -98,6 +105,26 @@ final case class WorkflowStoreSubmitActor(store: WorkflowStore, serviceRegistryA
     futureLists.map(seq => NonEmptyList.fromList(seq).get)
   }
 
+  private def convertJsonToLabelsMap(json: String): Map[String, String] = {
+    json.parseJson match {
+      case JsObject(inputs) => inputs.collect({
+        case (key, JsString(value)) => key -> value
+      })
+      case _ => Map.empty
```
mcovarr: same question here as above

engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
```diff
@@ -259,7 +259,7 @@ class CromwellApiServiceSpec extends AsyncFlatSpec with ScalatestRouteTest with
       }
   }
 
-  it should "return 201 with warnings for a successful v1 workflow submission still using wdlSource" in {
+    it should "return 201 with warnings for a successful v1 workflow submission still using wdlSource" in {
```
mcovarr: not sure why these indents changed, it looks like they were more correct before

engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
```diff
@@ -347,6 +347,25 @@ class CromwellApiServiceSpec extends AsyncFlatSpec with ScalatestRouteTest with
         }
     }
 
+    it should "return 400 for a workflow submission with invalid workflow custom labels" in {
+      val labels = s"""
+                      |{"key with characters more than 255-at vero eos et accusamus et iusto odio dignissimos ducimus qui blanditiis praesentium voluptatum deleniti atque corrupti quos dolores et quas molestias excepturi sint occaecati cupiditate non provident, similique sunt in culpas":"value with characters more than 255-at vero eos et accusamus et iusto odio dignissimos ducimus qui blanditiis praesentium voluptatum deleniti atque corrupti quos dolores et quas molestias excepturi sint occaecati cupiditate non provident, similique sunt in culpa"}
```
mcovarr: more than 255 characters

engine/src/main/scala/cromwell/engine/workflow/lifecycle/materialization/MaterializeWorkflowDescriptorActor.scala
salonishah11: case _ represents all the cases other than JsObjects (including failure) after doing json.parseJson. Since now we validate labels during submission time inside PartialWorkflowSources, it is unlikely that parsing the json string of labels will fail here. But we need to parse again as we only have access to the json string and not actual labels.

Same thing happens in WorkflowStoreSubmitActor.

engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
salonishah11: They might have been indented somehow, I will revert them.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Labels
 * Cromwell has removed most of the formatting restrictions from custom labels. Please check the [README](README.md#label-format) for more detailed documentation.
 * Custom labels won't be submitted to Google backend as they are now decoupled from Google's default labels.
+* Cromwell now publishes the labels as soon as the workflow is submitted (whether started or on hold). If the labels are invalid, workflow will not be submitted and request will fail.
```
geoffjentry: Maybe "the workflow will not be submitted"?

geoffjentry: oh no, i meant to add the "the". I think. You shouldn't take grammar advice from me. That's @mcovarr 's bag

engine/src/main/scala/cromwell/engine/workflow/lifecycle/materialization/MaterializeWorkflowDescriptorActor.scala
mcovarr: I'm not sure about representing failure. From a quick scan of the JSON parsing code it looks like this would throw a `ParseException` if it's unhappy. If this code is reparsing something that parsed successfully once perhaps it doesn't matter, but maybe the first-pass parser might need to be more robust if it's doing something like this.

salonishah11: Yes this code does re-parse the labels. We need to re-parse it as we only have access to json string rather than actual Label. The labels are initially parsed in PartialWorkflowSources, and if they are invalid the submission is rejected. So when the control reaches here (in MaterializeWorkflowDescriptorActor) the labels are already valid.

mcovarr: so if it can't be an error, should we be concerned if what comes out of the parse is not a `JsObject`?

salonishah11: That case (not being JsObject) is also covered in PartialWorkflowSources. If the labels are not JsObject it will result in error stating 'Invalid workflow labels JSON. Expected a JsObject of "labelKey": "labelValue" values.' and the workflow submission is rejected.

kbergin:
Hooray!! Thank you @salonishah11 !

kbergin:
Linking to issue #3576

PR #3578, opened 2018-05-02
Workflow options google_project output in metadata

kshakir:


supportedBackends/google/pipelines/common/src/main/scala/cromwell/backend/google/pipelines/common/PipelinesApiJobCachingActorHelper.scala
```diff
@@ -69,8 +69,8 @@ trait PipelinesApiJobCachingActorHelper extends StandardCachingActorHelper {
       .workflowPaths
       .workflowDescriptor
       .workflowOptions
-      .get(PipelinesApiMetadataKeys.GoogleProject)
-      .getOrElse(jesAttributes.project) 
+      .get(WorkflowOptionKeys.GoogleProject)
```
Horneth: So it was there already, just not where they wanted it ?
Seems like the only diff between `PipelinesApiMetadataKeys.GoogleProject` and `WorkflowOptionKeys.GoogleProject` is that the first one is under the `jes` section of metadata.
Should it be in both then ?

kshakir: This is editing the read from the workflow options json.

There is no such workflow option key `jes:google_project`. So the original line was always returning `None`. The key name under a workflow options json is `google_project`.

Horneth: Oh ok my bad I read that wrong, thanks

PR #3580, opened 2018-05-02
Partial Unification Compiler Flag

danbills:
This compiler flags allows the compiler to take types of kind `F[_,_]` and infers `F[_]` by parameterizing the right most argument and fixing the other ones.

This allows us to omit type args when traversing to `Either` and `ErrorOr`, because scala can now see them as shape `F[_]` instead of "you gave me a type w/ 2 type args and I was expecting one"

**I did a find/replace on all our traverses, this may be controversial**, chime in if you disagree.

[Longer explanation](https://gist.github.com/djspiewak/7a81a395c461fd3a09a6941d4cd040f2)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 .artifactory
 .idea/*
 .ensime_cache/*
+.config/*
 .ensime
 tags
 !/.idea/inspectionProfiles/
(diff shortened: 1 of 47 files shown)


cjllanwarne:
Why might it be controversial?

Although, as a contrary data point, 9 times out of 10, I only have to put the type annotation on traverse for intelliJ's benefit rather than the compiler

danbills:
@cjllanwarne it may be that people like the type annotation on the traverse.

I have seen Intellij struggle at times, but I find that a restart-and-invalidate fixes the issue for good.  I know it's not ideal, but it does work and should only need to be done once for this PR.

danbills:
@cjllanwarne also the redwood flag is more of a "heads-up you can do this now" than a concern that this is really controversial.

PR #3591, opened 2018-05-04
Link dockerOutputDirectory to execution directory

Horneth:
Fixes conformance test 102 on local and PAPI


centaur/src/main/scala/centaur/api/CentaurCromwellClient.scala
```diff
@@ -83,13 +83,12 @@ object CentaurCromwellClient {
       case _: TimeoutException |
            _: StreamTcpException |
            _: IOException |
-           _: UnsupportedContentTypeException if !CromwellManager.isReady && attempt < awaitMaxAttempts =>
+           _: UnsupportedContentTypeException if attempt < awaitMaxAttempts =>
```
Horneth: It seems that removing this unflakies (at least partly) centaur tests.
We can discuss about why it's here and what removing it does if people care.

Horneth: This comes from the fact that centaur starts and stops Cromwell to be able to test restarts.
Now the individual centaur tests that are polling Cromwell for status / metadata to see if their workflow finished are not aware that Cromwell is being restarted, which means they keep polling even when Cromwell is down.
To avoid the tests failing because of that, the code is looking for IO errors / timeouts that occur **when Cromwell is considered "not ready"**. If that's the case, then the request is being retried. The idea is to differentiate between Cromwell not responding because it's being restarted (which is normal), or not responding for unknown reason while it's running (which is not normal).
It could be that the heuristic in `CromwellManager.isReady` is not very accurate and / or there's some sort of race condition that makes the retry not happening.
By removing this we don't differentiate between "Cromwell being restarted" and "normal execution" anymore - which would not cache transient unresponsiveness from Cromwell anymore even when it's running.
It's probably not the jobs of centaur and there should be another category of tests under the scalability umbrella responsible to test and measure Cromwell's responsiveness under load though so that's why I'm AOK with removing this.

PR #3593, opened 2018-05-04
minor change to cromiam's swagger to be less weird looking

geoffjentry:


CromIAM/src/main/resources/swagger/cromiam.yaml
```diff
@@ -427,40 +427,7 @@ paths:
             - openid
             - email
             - profile
-  '/api/workflows/{version}/{id}/releaseHold':
```
Horneth: Why is this one deleted ?

PR #3594, opened 2018-05-04
Use directory basename when localizing

Horneth:
Fixes conformance test 86 on local


centaurCwlRunner/src/main/scala/centaur/cwl/OutputManipulator.scala
```diff
@@ -95,8 +99,17 @@ object OutputManipulator extends Poly1 {
     // We need this because some task return a "File" object that actually does not exist on disk but has its content provided directly instead
     def sizeContent: Option[Long] = obj.kleisli("contents").flatMap(_.asString.map(_.length.toLong))
 
-    // The cwl test runner expects only the name, not the full path
-    val updatedLocation = obj.add("location", Json.fromString(path.name))
+    val cleanName = callRootPaths.find(path.startsWith) match {
+      case Some(localRoot: DefaultPath) =>
+        // For local paths, the root doesn't have "execution" in it but that's where the outputs are
```
mcovarr: 🙈 
I'm just about to write the same thing to localize IWDR content to the execution directory 😢

centaurCwlRunner/src/main/scala/centaur/cwl/Outputs.scala
```diff
@@ -2,25 +2,37 @@ package centaur.cwl
...
+import common.validation.ErrorOr.ErrorOr
+import common.validation.Validation._
 
 object Outputs {
 
   //When the string returned is not valid JSON, it is effectively an exception as CWL runner expects JSON to be returned
   def handleOutput(submittedWorkflow: SubmittedWorkflow, pathBuilder: PathBuilder): String = {
     val metadata: Map[String, JsValue] = CentaurCromwellClient.metadata(submittedWorkflow).get.value
 
+    val callRoots = metadata.toList.flatTraverse[ErrorOr, Path]({
```
mcovarr: are you having type parameter issues too? 😄

centaurCwlRunner/src/main/scala/centaur/cwl/OutputManipulator.scala
Horneth: Oh this is actually not related to test 86 I didn't mean to push this here I must have mixed up my branches. This is for 106.
I might as well leave it here though..
And yeah this is far from great. Maybe we could have a `callExecutionRoot` on top of the `callRoot` published to metadata ?

mcovarr: Yeah we might want to publish that. Currently I'm assuming in about 50 various places that the execution directory is immediately under the call directory but it's possible that could change. 😄

PR #3601, opened 2018-05-07
WDL 1.0 upgrade script Phase I: WDL writer

aednichols:
Takes WDL 1.0 case classes (`wdl.model.draft3.elements.FileElement` and its children) and writes them out as WDL source code.

Test scheme:
1. Load the WDL as case classes
2. Write them out as source
3. Load up the source from step 2 as case classes
4. Verify the case classes from (1) and (3) are identical

I also uncovered some gaps in the set of WDLs we use for unit tests - I had gotten all the tests to pass, but there were still a handful of `???`'s in my code. Added test cases such that all of them were hit (and then implemented).


wdl/transforms/draft3/src/test/cases/command_syntaxes.wdl
```diff
@@ -9,6 +9,7 @@ task a {
   command {
     echo ~{"hello"} ${world1}
     echo goodbye ~{world2}
+    echo ~{sep=", " true="--yes" false="--no" default="foo" world1}
```
cjllanwarne: Hmm, I'm actually a bit disappointed that the `toWom` test hasn't complained about invalid attributes for this `String` placeholder value....

wdl/model/draft3/src/main/scala/wdl/model/draft3/elements/ScatterElement.scala
```diff
@@ -3,4 +3,21 @@ package wdl.model.draft3.elements
...
                                 scatterVariableName: String,
-                                graphElements: Seq[WorkflowGraphElement]) extends WorkflowGraphElement
+                                graphElements: Seq[WorkflowGraphElement]) extends WorkflowGraphElement {
+
+  // Scatter names do not contain intrinsic information about the scatter; rather they are a sort
+  // of hash based on the declarations's physical location in the source.
+  override def equals(other: scala.Any): Boolean = {
+    other match {
+      case otherScatter: ScatterElement =>
+        this.scatterExpression == otherScatter.scatterExpression &&
```
cjllanwarne: I forget the exact interplay between `.equals` and `==` in scala but since you're overriding `equals`, does it make sense to use `.equals` in these sub-comparisons?

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wdl/WdlWriter.scala
```diff
@@ -0,0 +1,450 @@
...
+        "." + a.memberAccessTail.mkString(".")
+      } else {
+        ""
+      })
+    }
+  }
+
+  implicit val binaryOperationWriter: WdlWriter[BinaryOperation] = new WdlWriter[BinaryOperation] {
+    override def toWdl(a: BinaryOperation) = {
+      def sandwich(opString: String) = s"${a.left.toWdl} $opString ${a.right.toWdl}"
```
cjllanwarne: I wonder whether it's necessary to add parens here to avoid accidentally undoing order of precedence?
eg does `(1+2)*3` get converted to `1+2*3`?

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wdl/WdlWriter.scala
```diff
@@ -0,0 +1,450 @@
...
+import wdl.model.draft3.elements.ExpressionElement._
+import wdl.model.draft3.elements.MetaValueElement._
+import wdl.model.draft3.elements._
+import wom.types._
+
+import scala.language.implicitConversions
+
+@typeclass
+trait WdlWriter[A] {
+  def toWdl(a: A): String
```
cjllanwarne: Specifically, `toWdlV1`

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wdl/WdlWriter.scala
```diff
@@ -0,0 +1,450 @@
...
+
+  // Stolen from WomGraph.scala
+  def indent(s: String) = s.lines.map(x => s"  $x").mkString(System.lineSeparator)
+  def combine(ss: Iterable[String]) = ss.mkString(start="", sep=System.lineSeparator, end=System.lineSeparator)
+  def indentAndCombine(ss: Iterable[String]) = combine(ss.map(indent))
+}
+
+object WdlWriter {
```
cjllanwarne: I *think* I'd prefer to have this implementation in a different file to the typeclass, to make it clear that "here is a typeclass that anyone can implement" and "here is one implementation"

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wdl/WdlWriter.scala
```diff
@@ -0,0 +1,450 @@
...
+      def sandwich(opString: String) = s"${a.left.toWdl} $opString ${a.right.toWdl}"
+
+      a match {
+        case _: LogicalOr           => sandwich("||")
+        case _: LogicalAnd          => sandwich("&&")
+        case _: Equals              => sandwich("==")
+        case _: NotEquals           => sandwich("!=")
+        case _: LessThan            => sandwich("< ")
+        case _: LessThanOrEquals    => sandwich("<=")
+        case _: GreaterThan         => sandwich("> ")
```
cjllanwarne: Just OOI, is the extra space here deliberate?

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wdl/WdlWriter.scala
```diff
@@ -0,0 +1,450 @@
...
+  }
+
+  implicit val scatterElementWriter: WdlWriter[ScatterElement] = new WdlWriter[ScatterElement] {
+    override def toWdl(a: ScatterElement): String =
+      s"""scatter (${a.scatterVariableName} in ${a.scatterExpression.toWdl}) {
+         |${indentAndCombine(a.graphElements.map(_.toWdl))}}""".stripMargin
+  }
+
+  implicit val callBodyElement: WdlWriter[CallBodyElement] = new WdlWriter[CallBodyElement] {
+    override def toWdl(a: CallBodyElement): String = {
```
cjllanwarne: ToL: There's a new short-form syntax for these in scala 2.12 (which we can now use - hooray!). I'll let you decide whether you want to retype everything just for the sake of saving a few characters:
```scala
implicit val callBodyElement: WdlWriter[CallBodyElement] = a => {
  // method contents
}
```

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wdl/WdlWriter.scala
```diff
@@ -0,0 +1,450 @@
...
+          case None => ""
+        }
+      }).filterNot(_.length == 0).mkString(sep = " ")
+    }
+  }
+
+  implicit val inputsSectionElementWriter: WdlWriter[InputsSectionElement] = new WdlWriter[InputsSectionElement] {
+    override def toWdl(a: InputsSectionElement): String = {
+      s"""input {
+         |${indentAndCombine(a.inputDeclarations.map(_.toWdl))}}""".stripMargin
```
cjllanwarne: I think I'd prefer that closing `}` to be on a newline

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wdl/WdlWriter.scala
```diff
@@ -0,0 +1,450 @@
...
+      }
+
+      s"${a.typeElement.toWdl} ${a.name}$expression"
+    }
+  }
+
+  implicit val outputsSectionElementWriter: WdlWriter[OutputsSectionElement] = new WdlWriter[OutputsSectionElement] {
+    override def toWdl(a: OutputsSectionElement): String = {
+      s"""output {
+         |${indentAndCombine(a.outputs.map(_.toWdl))}}""".stripMargin
```
cjllanwarne: Same here re the closing `}`

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wdl/WdlWriter.scala
```diff
@@ -0,0 +1,450 @@
...
+         |${indent(runtime)}
+         |${indent(meta)}
+         |${indent(parameterMeta)}}""".stripMargin
+    }
+  }
+
+  implicit val commandSectionElementWriter: WdlWriter[CommandSectionElement] = new WdlWriter[CommandSectionElement] {
+    override def toWdl(a: CommandSectionElement): String = {
+      s"""command <<<
+         |${combine(a.parts.map(_.toWdl))}>>>""".stripMargin
```
cjllanwarne: Same comment about moving the trailing `>>>` to a new line

cjllanwarne:
ToL:

Looking at this, I wonder whether an easier route to the upgrade script is to make another implementation of this `toWdlVersion1` typeclass for draft 2 `WdlNamespace`.

It leaves us further away from funneling draft 2 and draft 3 through the same object mode (and the massive code deletion we'd get from that)l, but it might be a much more expedient (and maybe safer?) way of achieving the upgrade script?

wdl/model/draft3/src/main/scala/wdl/model/draft3/elements/ScatterElement.scala
aednichols: Wasn't sure so [looked it up](https://stackoverflow.com/a/7681243/818054):
>You normally use `==`, it routes to `equals`, except that it treats `null`s properly. Reference equality (rarely used) is `eq`.

So we could do so on a style basis only, which I'm 🤷‍♂️ on

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wdl/WdlWriter.scala
aednichols: Quite possibly. On the other hand, we end up with beauties like
```
Int j = (((a + b) + c) + x)
```
so I have been wondering if there is a smart way to insert parens only as needed

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wdl/WdlWriter.scala
aednichols: Nope, removed

wdl/model/draft3/src/main/scala/wdl/model/draft3/elements/ScatterElement.scala
cjllanwarne: oh, it's `eq` that's different...

Ok, so, never mind, looks good to me

aednichols:
^^ that commit message is a lie, I accidentally accepted the default in IntelliJ. It's just about all of your change requests that hadn't spawned a question/discussion

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wdl/WdlWriter.scala
aednichols: Thanks, good to know - in this particular case, I like having `toWdlV1` written out explicitly so people know what method we're implementing, the shortening seems to necessitate making it anonymous

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wdl/WdlWriter.scala
aednichols: Are you referring to the source code or the output?

With the brace in its current position, the output is
```
  Array[Pair[String, Int]?] lotsa_nesting_array
}
```
and on its own line
```
  Array[Pair[String, Int]?] lotsa_nesting_array

}
```

aednichols: We could have it both ways by modifying how `combine` works, looking into that for consistency

cjllanwarne:
Red thumb required for the `womtool` changes

wdl/model/draft3/src/main/scala/wdl/model/draft3/elements/ScatterElement.scala
```diff
@@ -3,4 +3,21 @@ package wdl.model.draft3.elements
...
+        this.scatterVariableName == otherScatter.scatterVariableName &&
+        this.graphElements == otherScatter.graphElements
+      case _ => false
+    }
+  }
+
+  // Shorthand to ignore specific members for hashing purposes
+  // https://stackoverflow.com/a/31915429/818054
+  override def hashCode(): Int = (scatterExpression, scatterVariableName, graphElements).##
```
mcovarr: 😮 droppin knowledge. Though it might be more clear (if less magical) to say it only includes certain members.

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wdl/WdlWriterImpl.scala
```diff
@@ -0,0 +1,438 @@
...
+        case _: LessThan            => sandwich("<")
+        case _: LessThanOrEquals    => sandwich("<=")
+        case _: GreaterThan         => sandwich(">")
+        case _: GreaterThanOrEquals => sandwich(">=")
+        case _: Add                 => sandwich("+")
+        case _: Subtract            => sandwich("-")
+        case _: Multiply            => sandwich("*")
+        case _: Divide              => sandwich("/")
+        case _: Remainder           => sandwich("%")
+      }
```
mcovarr: You should be able to DRY this out a bunch by only returning the String from the `match` and then `sandwich`ing that. But then less sandwiches.

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wdl/WdlWriterImpl.scala
```diff
@@ -0,0 +1,438 @@
...
+        case None => ""
+      }
+      val meta = a.metaSection match {
+        case Some(m) => m.toWdlV1
+        case None => ""
+      }
+      val parameterMeta = a.parameterMetaSection match {
+        case Some(p) => p.toWdlV1
+        case None => ""
+      }
```
mcovarr: don't know how you feel about folding Options but it would be a lot less lines

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wdl/WdlWriterImpl.scala
```diff
@@ -0,0 +1,438 @@
...
+        case _: Transpose =>    functionCall("transpose")
+        case _: Length =>       functionCall("length")
+        case _: Flatten =>      functionCall("flatten")
+        case _: SelectFirst =>  functionCall("select_first")
+        case _: SelectAll =>    functionCall("select_all")
+        case _: Defined =>      functionCall("defined")
+        case _: Floor =>        functionCall("floor")
+        case _: Ceil =>         functionCall("ceil")
+        case _: Round =>        functionCall("round")
+        case _: Glob =>         functionCall("glob")
```
mcovarr: similar DRYing comment as above, minus the sandwich

wdl/transforms/draft3/src/test/cases/command_syntaxes.wdl
mcovarr: WOMp WOMp

wdl/transforms/draft3/src/test/scala/wdl/draft3/transforms/wdlom2wdl/WdlomToWdlFileSpec.scala
```diff
@@ -0,0 +1,37 @@
...
+    it should s"write a file that re-evaluates to the same case classes for ${file.name}" in {
+
+      val model: Checked[FileElement] = (fileToAst andThen astToFileElement).run(file)
+        model match {
+        case Right(wdlModel) =>
+
+          val newModel = (stringToAst andThen astToFileElement).run(FileStringParserInput(wdlModel.toWdlV1, file.name))
+
+          // Scala case class deep equality is so nice here
+          newModel shouldEqual model
```
mcovarr: 🍠

cjllanwarne: In case anyone else has to reverse search this one... it's https://emojipedia.org/roasted-sweet-potato/ apparently

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wdl/WdlWriterImpl.scala
cjllanwarne: My objection to folding `Option` is that it doesn't actually work in analogy to folding lists, so I just find I have to relearn what it's doing each time. I don't know why they had to override `fold` as a method name rather than calling it `mapOrDefault` or some such. OTOH it makes it terser so up to you...

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wdl/WdlWriterImpl.scala
aednichols: The big block of same-named functions did feel wrong to me aesthetically. Yours is the obvious solution, which apparently eluded me on two separate occasions

PR #3609, opened 2018-05-09
The non-detritus parts of CWL conformance test "Test dynamic initial work dir".

mcovarr:
Includes a Centaur smoke test. Pay no mind to the additional entry in `skipped_tests.csv`, nothing to see there.


centaur/src/main/resources/standardTestCases/cwl_dynamic_initial_workdir.test
```diff
@@ -0,0 +1,17 @@
...
+backends: [Local, LocalNoDocker]
+tags: [localdockertest]
+
+files {
+  workflow: cwl_dynamic_initial_workdir/cwl_dynamic_initial_workdir.cwl
+  inputs: cwl_dynamic_initial_workdir/cwl_dynamic_initial_workdir.inputs
+}
+
+metadata {
+  status: Succeeded
```
kshakir: Is there anything else this test can/should validate?

mcovarr: If nothing exploded that's a big upgrade from what would happen on current develop, but let me see if I can do even better.

mcovarr:
This seems to have been broken by a rebase. Closed for repairs.

PR #3616, opened 2018-05-11
IWDR + inputs directory listing expression - detritus = almost 87

mcovarr:
This is essentially the same as yesterday's PR except:

* `find` with a very restrictive predicate | `sort` | `cut` | `sha1sum` to make sure the directory contents are what we expect.
* Input paths in IWDR expressions are relativized to the Cromwell server so they don't errantly get relativized to the call directory before the IWDR files have been localized.


geoffjentry:
👍

PR #3620, opened 2018-05-14
Silence error messages on shutdown

Horneth:
This should help with error messages when Cromwell shuts down


engine/src/main/scala/cromwell/server/CromwellAkkaLogFilter.scala
```diff
@@ -0,0 +1,20 @@
+package cromwell.server
+
+import akka.actor.ActorSystem
+import akka.event.EventStream
+import akka.event.slf4j.Slf4jLoggingFilter
+
+class CromwellAkkaLogFilter(settings: ActorSystem.Settings, eventStream: EventStream) extends Slf4jLoggingFilter(settings, eventStream) {
+  override def isErrorEnabled(logClass: Class[_], logSource: String) = {
+    /*
+     * This might filter out too much but it's the finest granularity we have hear
```
mcovarr: here

mcovarr: are you aware of anything specifically being filtered other than the bad stuff?

Horneth: Not specifically no

Horneth:
Hmm I don't think this will do anything for the noop but I can look into it too, it looks like it's coming from sentry.

geoffjentry:
Yeah that’s a sentry error. TBH I thought it was just on my home laptop, apparently not

kshakir:
Aha. So maybe we can just default in our own Noop DSN to silence the error.

ToL: The lack-of-a-DSN-message is also [Logback adjacent](https://docs.sentry.io/clients/java/modules/logback/#usage). Someday I'll figure out how the hell to use logback/Joran. On first glance it looks a lot like HOCON's embedded default `application.conf` that can be overriden via `-Dconfig.file=…` except one is supposed to use [`-Dlogback.configurationFile=…`](https://logback.qos.ch/manual/configuration.html#configFileProperty). But while I "get" HOCON's mechanics I do not yet "get" best practices for logback [overrides/includes](https://stackoverflow.com/a/23737143/3320205).

engine/src/main/scala/cromwell/server/CromwellDeadLetterListener.scala
```diff
@@ -0,0 +1,11 @@
+package cromwell.server
+
+import akka.event.DeadLetterListener
+
+class CromwellDeadLetterListener extends DeadLetterListener {
+  def shutdownReceive: Receive = {
+    // This silences the dead letter messages when Cromwell is shutting down
+    case _ if CromwellShutdown.shutdownInProgress() =>
```
danbills: I'm curious what these messages are and who's sending to whom.

Can we do some gentler printing (maybe don't mention the word "dead", perhaps "unhandled" or even "bonus" :) ).

I had actually started a branch to do this myself: https://github.com/broadinstitute/cromwell/compare/db_dead_letters#diff-ef805a742940b60e9d750303111fbb71R64

Horneth: Would logging them at debug level be ok ? I fear that even if we re-word the message people will get confused..

danbills: sure!

PR #3627, opened 2018-05-15
Update CWL conformance test git hash

Horneth:


mcovarr:
Tests being fixed with hash bumps I can get behind. 👍

kshakir:
Just noticed, this PR uses different hashes for conformance tests for Local / PapiV1 / PapiV2. I'm assuming that was not intentional. I have an incoming PR (as soon as PRs quiet down + I get travis to pass for once) that refactors this into reusable includes. That will hopefully help making CI changes in the future.

Horneth:
@kshakir good point thanks, I'll update local too

kshakir:
Your latest commit _might_ be failing because it's logging to stdout while conformance tests only allow logging to stderr. There are lots of `Extra data: line 1 column 3 - line 18 column 1 (char 2 - 1091)` in the travis logs for Local conformance tests.

Someday one of us will get the hang of logback and we can just "easily" [switch from stdout to stderr](https://stackoverflow.com/questions/25935326/how-can-i-configure-logback-conf-to-send-all-messages-to-stderr). For now I don't have any quick fixes for logging w/ our existing slf4j framework. `Console.err.println()` would work, but isn't pretty either.

Horneth:
@kshakir yeah I was afraid of that.. `Console.err.println()` sounds good to me since this is mostly for debugging + in centaur 🙈

PR #3629, opened 2018-05-15
CWL optional outputs + PAPI 2 = conformance 3

mcovarr:


Horneth:
> I haven't fully thought it through yet, but I'll keep something sorta like this in mind when we're dealing with tracking the WomFile.value to (cloud, vm, container) path split.

I think we could use a mini brainstorm session on this. I'm finding other areas where it would be very nice to clarify this cloud / vm /container separation.
Namely `JobPaths` has become a huge messy melting pot over time and it would be awesome to have a clean `cloudCallRoot`, `vmCloudRoot`, `containerCloudRoot` for example (some of those that could be the same depending on the backend, w/ docker or not etc..)

PR #3642, opened 2018-05-16
 Cromwell CI refactoring.

kshakir:
Updates:
- Restores cron build.
- Moved chunks of duplicated scripts and confs into includes.
- Scripts and confs are now each under 100 lines.
- Switch Local backend CWL conformance testing from serial to parallel.
- With minimal prep (ex: vault auth) run test scripts on your laptop.

Future:
- More refactoring might possible but at some point conditionals-in-common-code grows exponentially.


src/bin/ci/resources/funnel.conf
```diff
@@ -12,6 +12,7 @@ LocalStorage:
     - /home/
     - /cromwell-executions
     - /tmp/
+    - /Users/
```
mcovarr: Thank you!

src/bin/ci/testConformanceLocal.sh
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env bash
+
+set -e
+# import in shellcheck / CI / IntelliJ compatible ways
+# shellcheck source=/dev/null
+source "${BASH_SOURCE%/*}/test.inc.sh" || source test.inc.sh
+
+cromwell::build::setup_everyone_environment
```
mcovarr: `everyone_environment`?

src/bin/ci/testConformanceLocal.sh
```diff
@@ -0,0 +1,63 @@
...
+cromwell::build::assemble_jars
+
+CENTAUR_CWL_RUNNER_MODE="local"
+
+# Export variables used in conf files and commands
+export CENTAUR_CWL_RUNNER_MODE
+
+shutdown_cromwell() {
+    if [ -n "${CROMWELL_PID+set}" ]; then
+        cromwell::build::kill_children "${CROMWELL_PID}"
```
mcovarr: not sure about the name 😬

src/bin/ci/testConformanceLocal.sh
kshakir: I needed an word to describe "this does not need private information from Vault because it doesn't access the cloud and can run on any person's laptop". The term "public" seemed very overloaded already and "everyone" was chosen. Open to other suggestions.

Its sibling is called `cromwell::build::setup_secure_environment` if that should be renamed too.

src/bin/ci/testConformanceLocal.sh
mcovarr: `kill_cromwell_server` maybe...

kshakir: The command kills a unix process ID and any of the child UNIX process IDs, especially those created from subshells. It's used for killing `funnel` and its `nohup`, `tail`, `cromwell`, etc.

kshakir: Going to try out `kill_tree`.

PR #3651, opened 2018-05-18
Requester pays functionality

salonishah11:
Requester pays is now added to Cromwell. The user can set the billing project id using one of the methods listed below:

- It can be added as `'google_project':'project-id'` as part of workflow options during workflow submission
- It can be included inside configuration file as shown in Getting started on Google Pipelines API where you need to replace the <google-billing-project-id> with the project id
- If it is not mentioned using above 2 ways, Cromwell will use the default project that has been configured with gcloud

More information about Requester pays can be found [here](https://cloud.google.com/storage/docs/requester-pays)

### DIFF
--- a/centaur/src/main/resources/standardTestCases/globbing_behavior_with_requester_pays/globbing_behavior_with_requester_pays.wdl
+++ b/centaur/src/main/resources/standardTestCases/globbing_behavior_with_requester_pays/globbing_behavior_with_requester_pays.wdl
@@ -1,0 +1,26 @@
+task globbing_behavior_task{
+    command {
+        echo "staticFile" > staticFile.txt
+        echo "staticArray" > staticArray.txt
+        echo "globFile" > globFile.txt
+        echo "globArray" > globArray.txt
+    }
+
+    output {
+        Array[File] globArray = glob("*.txt")
+        String globArrayContent = read_string(globArray[0])
+    }
+
+    runtime {
+        docker: "us.gcr.io/google-containers/ubuntu-slim:0.14"
+    }
+}
+
+
+workflow globbing_behavior_with_requester_pays{
+    call globbing_behavior_task
+
+    output {
+        String globArrayContent = globbing_behavior_task.globArrayContent
+    }
+}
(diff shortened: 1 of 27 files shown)


centaur/src/main/resources/standardTestCases/globbing_behavior_with_requester_pays/globbing_behavior_with_requester_pays.wdl
```diff
@@ -0,0 +1,26 @@
+task globbing_behavior_task{
```
cjllanwarne: ToL (because this comment is a little unfair) - would you consider making these all WDL 1.0s?

If you're using intelliJ you can just put `version 1.0` at the top and the highlighter will tell you what else to change (though looking at the tasks, I don't think it needs anything else changed)

centaur/src/main/resources/standardTestCases/read_file_from_bucket_with_requester_pays/read_file_from_bucket_with_requester_pays.wdl
```diff
@@ -0,0 +1,21 @@
+task concat_task{
+    Array[String] array
+
+    command {
+        echo $${sep = ' ' array} > concat
```
cjllanwarne: is the double `$` deliberate?

docs/backends/Google.md
```diff
@@ -167,6 +167,19 @@ precedence over the `root` specified at `backend.providers.JES.config.root` in t
 workflows using the Google backend.
 
 
+#### Requester Pays
```
cjllanwarne: thanks!

centaur/src/main/resources/standardTestCases/read_file_from_bucket_with_requester_pays/read_file_from_bucket_with_requester_pays.wdl
salonishah11: No. I just saw in some other WDL so added it. Removed it now.

centaur/src/main/resources/standardTestCases/globbing_behavior_with_requester_pays/globbing_behavior_with_requester_pays.wdl
salonishah11: Added `version 1.0` to 2 WDLs 👍

PR #3653, opened 2018-05-18
Call cache capoeira for 1.0 (plus ensure task hashes match draft-2)

cjllanwarne:
Hint: these are barely changed from the draft-2 versions. 
You probably only *need* to review the `.scala` changes, although obviously another pair of eyes on the call cache capoeira tests never hurt.

I made a mistake - no red thumb is required now but a red thumb is of course always welcome...

### DIFF
--- a/wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wdl/WdlWriterImpl.scala
+++ b/wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wdl/WdlWriterImpl.scala
@@ -300,7 +300,6 @@ object WdlWriterImpl {
         "false" -> a.falseAttribute,
         "default" -> a.defaultAttribute
       ).collect({ case (attrKey: String, Some(value)) => attrKey + "=\"" + value + "\"" })
-        .filterNot(_.length == 0)
 
       if (attrStrings.isEmpty) "" else attrStrings.mkString(start = "", sep = " ", end = " ")
     }
--- a/wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wdl/WdlWriterImpl.scala
+++ b/wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wdl/WdlWriterImpl.scala
@@ -294,17 +294,15 @@ object WdlWriterImpl {
 
   implicit val placeholderAttributeSetWriter: WdlWriter[PlaceholderAttributeSet] = new WdlWriter[PlaceholderAttributeSet] {
     override def toWdlV1(a: PlaceholderAttributeSet): String = {
-      Map(
+      val attrStrings = Map(
         "sep" -> a.sepAttribute,
         "true" -> a.trueAttribute,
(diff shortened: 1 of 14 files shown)


wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wdl/WdlWriterImpl.scala
```diff
@@ -294,17 +294,15 @@ object WdlWriterImpl {
...
         "false" -> a.falseAttribute,
         "default" -> a.defaultAttribute
-      ).map({ case (attrKey: String, maybeValue: Option[String]) =>
-        maybeValue match {
-          case Some(value) => attrKey + "=\"" + value + "\""
-          case None => ""
-        }
-      }).filterNot(_.length == 0).mkString(sep = " ")
+      ).collect({ case (attrKey: String, Some(value)) => attrKey + "=\"" + value + "\"" })
+        .filterNot(_.length == 0)
```
cjllanwarne: Actually, I think this is no longer needed

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wdl/WdlWriterImpl.scala
```diff
@@ -294,17 +294,15 @@ object WdlWriterImpl {
...
+
+      if (attrStrings.isEmpty) "" else attrStrings.mkString(start = "", sep = " ", end = " ")
```
cjllanwarne: I ended up refactoring this so that I could get this extra `" "` at the end, which turned out to be important in our call caching algorithm (but only because we have to lay it out the same as the draft 2 printer did. Spaces between attributes shouldn't matter in the original source 🤞)

wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/wdlom2wdl/WdlWriterImpl.scala
```diff
@@ -294,17 +294,14 @@ object WdlWriterImpl {
...
         "default" -> a.defaultAttribute
-      ).map({ case (attrKey: String, maybeValue: Option[String]) =>
-        maybeValue match {
-          case Some(value) => attrKey + "=\"" + value + "\""
-          case None => ""
-        }
-      }).filterNot(_.length == 0).mkString(sep = " ")
+      ).collect({ case (attrKey: String, Some(value)) => attrKey + "=\"" + value + "\"" })
+
+      if (attrStrings.isEmpty) "" else attrStrings.mkString(start = "", sep = " ", end = " ")
```
cjllanwarne: Resurrecting lost comment:

I ended up refactoring this so that I could get this extra " " at the end, which turned out to be important in our call caching algorithm (but only because we have to lay it out the same as the draft 2 printer did for hashing. Spaces between attributes shouldn't matter in the original source 🤞)

cjllanwarne:
A red thumb is required after all because of `-e draft3_call_cache_capoeira_local \`

PR #3661, opened 2018-05-22
More CI updates/fixes.

kshakir:
Give Centaur-managed Cromwell more time to restart and a custom exit code.
Publish artifacts again on each build tag.
Login to docker before trying to push images.
Functions using secure variables ensure that xtrace is not enabled, thus no longer need a subshell, thus do not need to be exported.
Artifactory and Docker Hub credentials added to vault.
Docker login can use environment variables or vault once dsde-toolbox is public.
Split setup_secure_environment into setup_common_environment and setup_secure_resources.
Sbt environment variables prefixed as CROMWELL_SBT_*.
Print out a warning instead of exiting when vault resources cannot be rendered when testing locally.
Minor updates for more consistent shell variable usage.

### DIFF
--- a/project/Settings.scala
+++ b/project/Settings.scala
@@ -93,7 +93,7 @@ object Settings {
     test in assembly := {},
     assemblyMergeStrategy in assembly := customMergeStrategy.value,
     logLevel in assembly :=
-      sys.env.get("ASSEMBLY_LOG_LEVEL").flatMap(Level.apply).getOrElse((logLevel in assembly).value)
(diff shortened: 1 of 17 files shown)


project/Settings.scala
```diff
@@ -127,12 +127,12 @@ object Settings {
     NOTE: Like below, gave up coming with an SBT setting. Using an environment variable instead.
 
     Once 2.11 is gone, instead of
-      `ENABLE_COVERAGE=true sbt +test coverageReport`
+      `CROMWELL_SBT_COVERAGE=true sbt +test coverageReport`
```
danbills: `+test` -> `test` and rm 2.11 comments

PR #3667, opened 2018-05-22
Allow language factories to infer appropriateness

cjllanwarne:
I think this'll be a nice quality of life improvement for people who aren't already in the habit of supplying a "workflowTypeVersion" in their submissions (ie everyone)


cjllanwarne:
This will return when it can work out language as well as language version

PR #3668, opened 2018-05-22
conformance 67 - Test optional output file and optional secondaryFile on output

mcovarr:
Issues addressed:

- Custom-named stdout or stderr files were not being delocalized.
- The wrong stdout and stderr files were being delocalized (user action instead of user command).
- Existence check for optional files was a `-a` test not actually supported by Alpine's `/bin/sh`. This failed with a zero error code and only minor noise to stderr.
- The fallback non-link chasing symlink that fails noisily for `cwl.output.json` and other no-match globs should now fail silently.
- Optional output files that were not produced were being reported as if they had been produced.
- The output manipulator didn't handle cases of > 1 type (optionals, coproducts) correctly in most cases. This is only a partial fix since it turned out not to be required for this test.
- Several non-WDL classes were found in a `wdl` package.


backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -234,7 +234,7 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
         |echo "${controlFileContent.trim}" > $globDirectory/$controlFileName
         |
         |# symlink all the files into the glob directory
-        |( ln -L ${globFile.value} $globDirectory 2> /dev/null ) || ( ln ${globFile.value} $globDirectory )
+        |( ln -L ${globFile.value} $globDirectory 2> /dev/null ) || ( ln ${globFile.value} $globDirectory 2> /dev/null)
```
Horneth: Thanks :)

mcovarr:
NOT MERGED closing temporarily for repairs as there are at least a couple of legit broken PAPI Centaur tests

PR #3669, opened 2018-05-22
Re-wire read limits for WDL

Horneth:
@danbills I ended up using the global conf in the interest of time.
I'm sure there's a better way to do it (@cjllanwarne suggested through the language factory configuration which seems a good idea). This is what the original ticket actually intended: https://github.com/broadinstitute/cromwell/issues/2611


centaur/src/main/resources/standardTestCases/read_file_limits/read_file_limits.wdl
```diff
@@ -0,0 +1,133 @@
+task readString {
+  command {
+    dd if=/dev/zero of=file_256k bs=256k count=1
```
mcovarr: I mean sure this file is large but to be fair it would compress REALLY well

wdl/model/shared/src/main/scala/wdl/shared/FileSizeLimitationConfig.scala
```diff
@@ -1,8 +1,8 @@
-package cromwell.backend.wdl
+package wdl.shared
```
mcovarr: oh I moved this to `cromwell.backend` in my PR but now it's looking a little more WDLy

wdl/model/draft2/src/test/scala/wdl/expression/WdlStandardLibraryFunctionsSpec.scala
```diff
@@ -64,6 +65,8 @@ case object TestableFunctions extends WdlStandardLibraryFunctions {
   override def stdout(params: Seq[Try[WomValue]]): Try[WomFile] = throw new NotImplementedError
   override def globHelper(pattern: String): Seq[String] = throw new NotImplementedError
   override def stderr(params: Seq[Try[WomValue]]): Try[WomFile] = throw new NotImplementedError
+
+  override protected def fileSizeLimitationConfig: FileSizeLimitationConfig = ???
```
kshakir: This should `throw new java.lang.UnsupportedOperationException`.

1) The original meaning of `???` was "I haven't finished coding and this code is not fit to be merged yet". However `throw new NotImplementedError` means "I really want this to throw a `java.lang.Error` because this isn't implemented".
2) `throw <? extends java.lang.Error>` means "Shut down the JVM!" It's almost akin to `System.exit()` except tries to print out the error while dying and _then_ shuts down.<sup>1</sup> If something goes wrong in only one of our tests we still want to run other tests.

It's outstanding tech-debt issue to clean up our prevalent usage of `Error` including `Predef.assert()` but we shouldn't add more. 

---

<sup>1</sup> The JVM dev community has so abused [`Error`](https://docs.oracle.com/javase/8/docs/api/java/lang/Error.html) over the years that the behavior catching errors is now a whole [copypasta](https://www.scala-lang.org/api/current/scala/util/control/NonFatal$.html) of maybe-shutdown-maybe-not. 🤦‍♂️

PR #3671, opened 2018-05-23
Remove cross compilation.

kshakir:


project/Dependencies.scala
```diff
@@ -420,7 +420,7 @@ object Dependencies {
   ) ++ slf4jBindingDependencies // During testing, add an slf4j binding for _all_ libraries.
 
   val kindProjectorPlugin = "org.spire-math" %% "kind-projector" % kindProjectorV
-  val paradisePlugin = "org.scalamacros" % "paradise" % paradiseV
+  val paradisePlugin = "org.scalamacros" % "paradise" % paradiseV cross CrossVersion.full
```
danbills: I'm surprised this is necessary if we're not doing the cross-build.

project/Settings.scala
```diff
@@ -105,44 +103,14 @@ object Settings {
...
-        // The default scalacOptions includes console-hostile options.  These options are overridden specifically below
-        // for the `console` target.
-        baseSettings ++ warningSettings ++ consoleHostileSettings
-      case _ =>
-        throw new NotImplementedError(
-          s"Found unsupported Scala version '${scalaVersion.value}'." +
-            s" ${name.value} does not support versions of Scala other than 2.12.")
-    }),
+    scalacOptions ++= baseSettings ++ warningSettings ++ consoleHostileSettings,
+    scalacOptions in(Compile, doc) ++= baseSettings,
```
danbills: I think this is duplicated 2 lines below

project/Dependencies.scala
kshakir: (Yesterday) i Learned:


- `CrossVersion.binary` (the default for scala libs) means "this library has specific builds for `2.11`, `2.12`, `2.13`, etc." Libraries using this default can use `%%` as the [artifact separator](https://www.scala-sbt.org/1.0/docs/Cross-Build.html#Cross-Building+a+Project).
- `CrossVersion.full` means "the library has specific builds for `2.12.0`, `2.12.1`, `2.12.2`, etc." Checkout the ton of cross versions for https://mvnrepository.com/artifact/org.scalamacros/paradise

Issue #3673, opened 2018-05-23
Workflows stuck in submitted status for 45 minutes in CaaS


danbills:
Copying chat log below:

---

dshiga [4:46 PM]
hi @kshakir, we're seeing a lot of workflows that have been stuck in submitted status for the past ~45 minutes. is something wrong in caas?
for example this workflow is stuck: 8de76a93-6b66-4c29-a2fe-31e6cd1f969e

kshakir [4:48 PM]
```
Workflow 8de76a93-6b66-4c29-a2fe-31e6cd1f969e in state Running and restarted = false cannot be started and should not have been fetched.
        at common.validation.Validation$ValidationTry$.toTry$extension1(Validation.scala:60)
        at common.validation.Validation$ValidationTry$.toTry$extension0(Validation.scala:56)
        at cromwell.engine.workflow.workflowstore.SqlWorkflowStore.$anonfun$fetchStartableWorkflows$1(SqlWorkflowStore.scala:57)
        at scala.util.Success.$anonfun$map$1(Try.scala:251)
        at scala.util.Success.map(Try.scala:209)
        at scala.concurrent.Future.$anonfun$map$1(Future.scala:288)
        at scala.concurrent.impl.Promise.liftedTree1$1(Promise.scala:29)
        at scala.concurrent.impl.Promise.$anonfun$transform$1(Promise.scala:29)
        at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:60)
        at akka.dispatch.BatchingExecutor$AbstractBatch.processBatch(BatchingExecutor.scala:55)
        at akka.dispatch.BatchingExecutor$BlockableBatch.$anonfun$run$1(BatchingExecutor.scala:91)
        at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:12)
        at scala.concurrent.BlockContext$.withBlockContext(BlockContext.scala:81)
        at akka.dispatch.BatchingExecutor$BlockableBatch.run(BatchingExecutor.scala:91)
        at akka.dispatch.TaskInvocation.run(AbstractDispatcher.scala:40)
        at akka.dispatch.ForkJoinExecutorConfigurator$AkkaForkJoinTask.exec(ForkJoinExecutorConfigurator.scala:43)
        at akka.dispatch.forkjoin.ForkJoinTask.doExec(ForkJoinTask.java:260)
[...]

PR #3675, opened 2018-05-23
Unclean shutdown support.

mcovarr:
A cleanly shut down Cromwell instance cleans up the workflow store to null out the Cromwell instance ID field for the workflow store entries it was running. When a new Cromwell instance comes up it will consider those workflow store entries to be fair game for pickup because those instance ID fields are null.

However an uncleanly shut down Cromwell does not null out the Cromwell instance ID field of its running workflows before it exits. When a new Cromwell instance comes up it will see that those workflow store entries appear to be owned by another Cromwell and will only pick them up if the heartbeats on those rows are older than the workflow heartbeat TTL (default 10 minutes). 

The problem here was some vestigial logic for the way restarts used to work that no longer makes sense in the 2/3-implemented horizontal Cromwell system. It is now entirely reasonable to see workflows in Running or Aborted state with or without a heartbeat timestamp depending on whether the Cromwell that was previously running those workflows was shut down cleanly or not.


mcovarr:
Addresses #3673

PR #3676, opened 2018-05-23
Language and Version inference

cjllanwarne:
Redo of #3667 with language and version inference. Also logs and metadata's the *actual* language factory we end up using


engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
```diff
@@ -212,14 +202,14 @@ object PartialWorkflowSources {
     partialSources match {
       case Valid(partialSource) =>
         (validateInputs(partialSource),
-          validateOptions(partialSource.workflowOptions), validateWorkflowType(partialSource),
-          validateWorkflowTypeVersion(partialSource), validateLabels(partialSource.customLabels.getOrElse("{}"))) mapN {
-          case (wfInputs, wfOptions, workflowType, workflowTypeVersion, workflowLabels) =>
+          validateOptions(partialSource.workflowOptions),
+          validateLabels(partialSource.customLabels.getOrElse("{}"))) mapN {
```
danbills: this `getOrElse` looks like a code smell.

can this `getOrElse` be put into a constructor for partial source, then the type of customLabels just be `WorkflowJson`?  Or maybe this validate code could be put into that class?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/materialization/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -218,19 +217,30 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
...
-      case None => fromEither[IO]("Need a workflow type here !".invalidNelCheck[ValidatedWomNamespace])
+      case Some(other) => s"Unknown workflow type: $other".invalidNel[LanguageFactory]
+      case None =>
+        val allFactories = CromwellLanguages.instance.languages.values.flatMap(_.allVersions.values)
+        chooseFactory(allFactories.toList).getOrElse(CromwellLanguages.instance.default.default).validNel
+    }
+
+    factory foreach { validFactory =>
+      workflowLogger.info(s"Parsing workflow as ${validFactory.languageName} ${validFactory.languageVersionName}")
+      pushLanguageToMetadata(validFactory.languageName, validFactory.languageVersionName)
```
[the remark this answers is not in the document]
cjllanwarne: true, although personally I think it's reassuring to log "you said this", and "you actually got this" separately. 

It also makes this code, and the code of any metadata reader, less complex since they can always rely on this field existing and not have to do `if (supplied.workflowType.exists) { ...} else { actualWorkflowType ... }`

engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
cjllanwarne: Having re-looked at it only briefly, the `PartialWorkflowSources` class seems to accumulate fields (eg from a REST request) which may or may not be incomplete/inconsistent, and then this `validate` method turns it into a guaranteed complete version that the rest of the process uses.

Personally, I'm sort of inclined to think that this "validate that this partial set is good enough and return a validated set" function is a good enough place to inject a default in case the REST request didn't include a `customLabels` field?

cjllanwarne: It might be (it definitely is) super-paranoid - but if the partial sources had `{}` as the default on that field, we couldn't tell whether we incorrectly got two copies of `customLabels` in the request - though I'm not sure that's actually checked anyway...

engine/src/main/scala/cromwell/engine/workflow/lifecycle/materialization/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -218,19 +217,30 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
                                       workflowOptions: WorkflowOptions,
                                       pathBuilders: List[PathBuilder],
                                       engineIoFunctions: EngineIoFunctions): Parse[EngineWorkflowDescriptor] = {
-    val namespaceValidation: Parse[ValidatedWomNamespace] = sourceFiles.workflowType match {
+    def chooseFactory(factories: List[LanguageFactory]): Option[LanguageFactory] = factories.find(_.looksParsable(sourceFiles.workflowSource))
```
aednichols: You missed a golden opportunity to create a `LanguageFactoryFactory` 😈

cjllanwarne: Ha! ... I have made a *terrible* mistake...

PR #3678, opened 2018-05-23
Requester pays clarification doc

Horneth:


docs/backends/Google.md
```diff
@@ -169,16 +169,60 @@ workflows using the Google backend.
...
+}
+workflow my_workflow {
+    call cat
+}
+```
+
+We assume here the task `cat` will run on a Pipelines API Cromwell backend.
+Consequently, Pipelines API will allocate a Virtual Machine for the task `cat`, which will in turn "localize" the file down from the GCS to a local disk.
+This operation is done on the VM itself, not by the Cromwell instance.
+Due to limitations on Pipelines API v1, localizing an object located in a bucket with requester pays enabled **can only be achieve on PAPI v2**.
```
mcovarr: achieved

CHANGELOG.md
```diff
@@ -2,6 +2,18 @@
 
 ## 32 Release Notes
 
+### Important Notes
+
+Cromwell instances configured to interact with **Google Cloud Storage** in any fashion, including but not limited to configurations with a **Pipelines API backend** (formerly JES backend) should **read the following section carefully**:
+
+As described in more details below, Cromwell 32 introduces initial support for PAPI v2, as well as Requester Pays.
+A direct consequence of those features is that user credentials and service accounts used in Cromwell now **require** a role which has the following permission: `serviceusage.services.use`. See [Requester Pays Requirements](https://cloud.google.com/storage/docs/requester-pays#requirements).
+This statement is **true regardless of the backend you are using, and only depends on whether Cromwell needs to communicate with GCS** (which is the case for **PAPI V1 and PAPI V2** backends).
```
mcovarr: do we ever say that Pipelines API == PAPI?

CHANGELOG.md
```diff
@@ -2,6 +2,18 @@
 
 ## 32 Release Notes
 
+### Important Notes
```
ruchim: I think this announcement header should qualify as "Breaking Changes"

CHANGELOG.md
```diff
@@ -2,6 +2,18 @@
...
+
+Cromwell instances configured to interact with **Google Cloud Storage** in any fashion, including but not limited to configurations with a **Pipelines API (PAPI) backend** (formerly JES backend) should **read the following section carefully**:
```
ruchim: Is it possible that in Changelog we basically limit the main points to  
*If* you access files in GCS *then* you now have to do ...

And then abstract away most of the rest to the new sections you've written up in the `Google.md` file? The explanation of "why" this is required is more of an implementation detail. I do think certain users would like to understand the reason behind this better --  but most people will need to know *what* and *how* to make the required changes.

Horneth: Ok

PR #3682, opened 2018-05-24
conformance 67 take 3

mcovarr:
Issues addressed:

- ~~Custom-named stdout or stderr files were not being delocalized.~~ Handled in #3695 instead
- ~~The wrong stdout and stderr files were being delocalized (user action instead of user command).~~ Handled in #3695 instead
- ~~Optional output files that were not produced were being published to metadata as if they had been produced.~~ Revert fixes for this due to existence check performance issues
- Existence check for optional files was a `-a` test not actually supported by the delocalization Docker's `/bin/sh`. This failed with a zero error code and only minor noise to stderr.
- ~~The fallback non-link chasing symlink that fails noisily for `cwl.output.json` and other no-match globs should now fail silently.~~ Reverted by reviewer request.
- The output manipulator didn't handle cases of > 1 type (optionals, coproducts) correctly in most cases. This is only a partial fix since it turned out not to be required for this test.
- Several non-WDL classes were found in a `wdl` package. Now correctly moved thanks to reviewer input. 🙂 
- The `read_tsv` Centaur test was falsely claiming success when the files referenced in the generated TSV were never delocalized from the VM to their stated cloud locations. With the existence checks removed from Cromwell proper this has reverted back to "failing to fail" which is mistaken for success.
- File literals now actually create a file, needed because the OutputManipulator does existence checks on outputs.


backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -234,7 +234,7 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
         |echo "${controlFileContent.trim}" > $globDirectory/$controlFileName
         |
         |# symlink all the files into the glob directory
-        |( ln -L ${globFile.value} $globDirectory 2> /dev/null ) || ( ln ${globFile.value} $globDirectory )
+        |( ln -L ${globFile.value} $globDirectory 2> /dev/null ) || ( ln ${globFile.value} $globDirectory 2> /dev/null)
```
kshakir: Curious why this error is being silenced?

If I end up with an empty array I generally want to know why. The prior `2> /dev/null` was only there because not all platforms support `-L`.

```wdl
Array[File] out = glob("oops_wrong_dir/*")
```

Produced previously:
```
ln: oops_wrong_dir/*: No such file or directory
```

In the case where one is using globs-for-optionals, then I guess there will be a "No such file or directory" generated, but a) it would only be in the stderr and b) would still help debug why one ended up with `Array[File] out = []`.

kshakir:
👍 when above issues are addressed w/ comments/updates.

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
mcovarr: With the disclaimer that this change is a stowaway not required for the rest of the PR to work, I thought  it could safely be assumed that if the array of files returned from the glob was empty it was because the glob did not match any files. An explicit message on stderr to that effect didn't seem to be adding value and looks scary in the logs. And due to the way the cwl.output.json is currently globbed we see that error message with nearly all CWL workflows, though we could probably handle cwl.output.json differently if we really wanted to keep this message for globs in general.

supportedBackends/google/pipelines/common/src/main/scala/cromwell/backend/google/pipelines/common/PipelinesApiAsyncBackendJobExecutionActor.scala
```diff
@@ -470,7 +476,16 @@ class PipelinesApiAsyncBackendJobExecutionActor(override val standardParams: Sta
...
-    womFileToGcsPath(generateJesOutputs(jobDescriptor))(womFile)
+    val cloudOutputWomFile: WomFile = womFileToGcsPath(generateJesOutputs(jobDescriptor))(womFile)
+    val cloudOutputPath: Path = backendEngineFunctions.buildPath(cloudOutputWomFile.value)
+
+    // It's not really exceptional for an optional output file to be missing but this is how that fact is communicated
+    // to the caller.
+    // .exists() on the file itself is fast but not 100% compatible with the gs:// URLs supported by gsutil.
+    // AsyncIo appears to be more compatible but a lot slower when used synchronously for individual files like this,
+    // so only use that if exists() returns false.
+    if (!cloudOutputPath.exists() && !asyncIo.exists(cloudOutputPath)) throw new FileNotFoundException(s"Could not process output, file not found: ${cloudOutputPath.pathAsString}")
```
Horneth: I'm surprised that it takes +1h with the asyncIo. I'm wondering if it's not the await result that makes it so much slower rather than the fact that it's going through the Io actor (which makes it slower but shouldn't be that much). My main concern is that the `.exists` on the path calls directly NIO which amongst other things doesn't have our custom retries which prevent known transient failures. I'd be curious to see what a Futurized version (without Await.result) would do performance wise, but I imagine that it would be extremely annoying to Futurize all this so... MaybeLaterTM

Horneth: I think you explained it to me already but can you re-explain why we need here to check that the files indeed do exist ? Doesn't the delocalization fail if they don't ?

mcovarr: This follows the behavior of the SFS backend which checks if an output file exists and if it does not throws a FNFE. From the context of evaluating an optional File output this gets turned into a None, which then shows up as a null-valued output in the metadata JSON. This conformance test does not expect to see the nonexistent optional output file in the output JSON, which I guess the OutputManipulator is helpful enough to remove for null values? This existence check was also helpful in pointing out that the `read_tsv` Centaur test doesn't actually work on PAPI since the files in the generated TSV are never delocalized from the VM.

mcovarr: I'll kick off another build with just async IO, maybe yesterday was just a bad cloud day.

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
kshakir: If no other reviewers have a problem changing this warning then I guess commit it as is and see if we get anyone complaining. Ultimately I could see how the UX could be debated as an empty _accidental_ glob is due to PEBKAC, but an empty _intentional_ glob could maybe be used in some circumstances.

Also if you're reading this: right now `OutputEvaluatorSpec` is still in the wrong package/directory. Should be under `cromwell/backend`.

supportedBackends/google/pipelines/common/src/main/scala/cromwell/backend/google/pipelines/common/PipelinesApiAsyncBackendJobExecutionActor.scala
Horneth: Oh ok I see what you're saying with the `read_tsv` test. I just think it's potentially expensive to add a request for every job output file (I'm not saying it's bad but we should be considerate of the performance impact, especially when we see how much Cromwell struggles these days).
If anything IMO the test is not thorough enough in the first place and should do a `read_string` or try to access the files in the array to prove that they are indeed delocalized.
If we can get away with relying on the PAPI delocalization to determine file existence without having to double check all of them in Cromwell afterwards it would be a big win.
But I don't think I understand all the intricacies with optional outputs so if you think it's necessary I trust your judgment 😄

mcovarr: Happy to talk this over more because tbh I'm a little apprehensive about the performance implications of adding this check as well, I just don't know another way to do it atm. 🙂 I'm not sure what you mean by having the delocalization determine file existence (outside the context of the VM)?

Horneth: The case of the `read_tsv` test is actually similar to a `cwl.output.json` containing file paths. The file paths in the tsv as well as the ones in the `cwl.output.json` are what we could call "dynamic outputs" in that they're unknown when the job starts. They have to be "discovered" once the job is complete and delocalized appropriately. I guess what I'm saying is that for non-dynamic outputs (the majority of them), we'll have an explicit delocalization action (as they're known when we create the job). For those, we can rely on the fact that `gsutil cp` will fail if the file is not there after the command runs (unless it's optional). If we assume that a successful `gsutil cp` guarantees that the file is in GCS, we shouldn't need to test their existence in the output evaluation.
For dynamic outputs, it's true that if we don't do anything special they will not be delocalized. However, provided that they are actually used in later tasks (or accessed in any way via `read_string` or whatever), the workflow will fail because they were left behind on the VM and subsequent access or localization will fail. Checking for their existence when we evaluate the outputs allows to fail a bit earlier but it has a potential performance cost like we said.
I think we should consider dynamic outputs as a feature that is currently partially implemented (works for `cwl.output.json` but not for this tsv case in WDL as you caught it).

For example we could update the test and add a workflow output like

```wdl
String out = read_string(output_file_table.table[0][0]
```

which would probably fail the test although it's valid WDL.

mcovarr:
The third and hopefully final version of this PR, could use a fresh thumbing. Much smaller than before since #3695 took care of the standard streams more universally leaving less to do here. Existence checks have shifted from Cromwell to the Centaur CWL runner and File literals have become real.

centaur/src/main/resources/standardTestCases/read_tsv.test
```diff
@@ -1,5 +1,8 @@
 name: read_tsv
 testFormat: workflowsuccess
+backendsMode: "only"
+backends: [Local, LocalNoDocker]
+tags: [localdockertest]
```
Horneth: I think this warrants a ticket so that this sort of WDL feature works on PAPI as we discussed with @cjllanwarne last week.

cwl/src/main/scala/cwl/CwlType.scala
```diff
@@ -31,6 +32,16 @@ object CwlType extends Enumeration {
   val String = Value("string")
   val File = Value("File")
   val Directory = Value("Directory")
+
+  // Callers expect the last component to match the basename if a basename is specified.
+  private [cwl] def tempFile(basename: Option[String], asDirectory: Boolean): files.File = {
```
Horneth: As a heads up this _might_ not be necessary anymore with https://github.com/broadinstitute/cromwell/pull/3697, I'll check after yours is merge though.

PR #3684, opened 2018-05-24
Typo

denis-yuen:
Ran into this while experimenting
https://github.com/ga4gh/task-execution-schemas/blob/master/task_execution.swagger.json#L18


kshakir:
Thanks!

PR #3685, opened 2018-05-24
Wire through HTTP resolver for WDL 1.0

cjllanwarne:


centaur/src/main/resources/standardTestCases/wdl_draft3/draft3_http_imports/draft3_http_imports.wdl
```diff
@@ -0,0 +1,11 @@
+version 1.0
+
+import "https://raw.githubusercontent.com/broadinstitute/cromwell/develop/centaur/src/main/resources/standardTestCases/wdl_draft3/draft3_infer_version/draft3_infer_version.wdl" as infered
```
aednichols: It would be nice to do this with a mock server instead of github for stability, but not a blocker by any means

languageFactories/language-factory-core/src/main/scala/cromwell/languages/util/ImportResolver.scala
```diff
@@ -61,4 +70,28 @@ object ImportResolver {
...
+  def httpResolverWithHeaders(headers: Map[String,String]): ImportResolver = CheckedAtoB.fromCheck { str: String =>
+
+    implicit val sttpBackend = AsyncHttpClientCatsBackend[IO]()
+
+    val responseIO : IO[Response[String]] =
+      sttp.get(uri"$str").headers(headers).send()
+
+    // temporary situation to get functionality working before
+    // starting in on async-ifying the entire WdlNamespace flow
+    val result: Checked[String] = Await.result(responseIO.unsafeToFuture, 15.seconds).body.leftMap(NonEmptyList(_, List.empty))
```
danbills: ToL possible to return IO[ImportResolver] and leave the dirty work to the caller? I know we're doing this all over the place but worth an ask.

centaur/src/main/resources/standardTestCases/wdl_draft3/draft3_http_imports/draft3_http_imports.wdl
kshakir: Don't think this is a blocker either but `develop` is also a floating tag. Links to github should [use a permalink](https://help.github.com/articles/getting-permanent-links-to-files/).

PR #3687, opened 2018-05-24
Oops. Restore insecure test skipping.

kshakir:


kshakir:
Fixes the false build errors from contributors, such as during https://github.com/broadinstitute/cromwell/pull/3684

PR #3697, opened 2018-05-28
PAPI Directories and Literals

Horneth:
- Better localization and delocalization of directories in PAPI2 using hidden files to cover for empty directories
- IWDR localization is not baked in the CWL code anymore but left to the backend. This allows for the PAPI backend to opt out of it since localization is done directly on the VM.
- ~~Use configurable `job-shell` instead of hardcoded `/bin/bash`~~ It fixes 117 but also makes a bunch of centaur tests fail, so leaving as is for now.
- Refactors Pipelines conversions in v2 (w/ typeclasses !)
- Allow for lazy evaluation of file and directory literals so that they can be written when the backend and the appropriate IoFunctions are known. This only partially covers the possible cases. It needs a deeper tech talk discussion.

This is orthogonal to the above and only here to avoid a later rebase (the files changed overlap with the refactoring mentioned).
- Partially replaces the custom `MemorySize` with [squants](https://github.com/typelevel/squants)
- Turns the CPU runtime validation from an `Int` to a `Int Refined Positive`
- Automatically fits the resources requirements in the task to the [GCE constraints](https://cloud.google.com/compute/docs/instances/creating-instance-with-custom-machine-type#specifications)


mcovarr:
You don't really *need* to worry about 117 here, but just FYI it's expected that the Centaur PAPI tests would require `/bin/bash` while the PAPI conformance tests should be able to get by with `/bin/sh`.

backend/src/main/scala/cromwell/backend/standard/package.scala
```diff
@@ -0,0 +1,18 @@
...
+
+package object standard {
+  object StandardAdHocValue {
+    object AsAdHocValue {
+      def unapply(arg: StandardAdHocValue): Option[AdHocValue] = arg.select[AdHocValue]
+    }
+    object AsLocalizedAdHocValue {
+      def unapply(arg: StandardAdHocValue): Option[LocalizedAdHocValue] = arg.select[LocalizedAdHocValue]
+    }
+  }
```
mcovarr: pity these unapplys need to be written out, seems like a common thing one would want to do with coproducts

centaurCwlRunner/src/bin/centaur-cwl-runner.bash
```diff
@@ -5,4 +5,4 @@ BUILD_ROOT="$( dirname "${BASH_SOURCE[0]}" )/../../.."
 CENTAUR_CWL_JAR="${CENTAUR_CWL_JAR:-"$( find "${BUILD_ROOT}/centaurCwlRunner/target/scala-2.12" -name 'centaur-cwl-runner-*.jar' | head -n 1 )"}"
 CENTAUR_CWL_SKIP_FILE="${BUILD_ROOT}/centaurCwlRunner/src/main/resources/skipped_tests.csv"
 
-java ${CENTAUR_CWL_JAVA_ARGS-"-Xmx1g"} -jar "${CENTAUR_CWL_JAR}" --skip-file "${CENTAUR_CWL_SKIP_FILE}" "$@"
+java -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005 ${CENTAUR_CWL_JAVA_ARGS-"-Xmx1g"} -jar "${CENTAUR_CWL_JAR}" --skip-file "${CENTAUR_CWL_SKIP_FILE}" "$@"
```
mcovarr: 🔙

centaurCwlRunner/src/main/scala/centaur/cwl/OutputManipulator.scala
```diff
@@ -183,7 +184,8 @@ object OutputManipulator extends Poly1 {
           o <- json.asObject
           l <- o.kleisli("location")
           s <- l.asString
-          p <- pathBuilder.build(s).toOption
+          c <- if (a.contains(Inl(CwlType.Directory))) s.ensureSlashed else s
```
mcovarr: should this be ```c = ...```?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -311,14 +312,14 @@ case class WorkflowExecutionActor(params: WorkflowExecutionActorParams)
 
     // Output ports for the workflow
     val workflowOutputNodes: Set[GraphOutputNode] = workflowDescriptor.callable.graph.outputNodes
-
+
```
mcovarr: 🔙

src/bin/ci/testCentaurPapiV2.sh
```diff
@@ -39,7 +39,6 @@ centaur/test_cromwell.sh \
     -g \
     -e localdockertest \
     -e docker_hash_dockerhub_private \
-    -e gpu_on_papi \
```
mcovarr: 📈

supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/PipelinesParameterConversions.scala
```diff
@@ -0,0 +1,65 @@
...
+trait PipelinesParameterConversions {
+  implicit val fileInputToParameter = new ToParameter[PipelinesApiFileInput] {
+    override def toActions(fileInput: PipelinesApiFileInput, mounts: List[Mount], projectId: String) = {
+      List(gsutil("cp", fileInput.cloudPath.pathAsString, fileInput.containerPath.pathAsString)(mounts, description = Option("localizing")))
+    }
+  }
+
+  implicit val directoryInputToParameter = new ToParameter[PipelinesApiDirectoryInput] {
+    override def toActions(directoryInput: PipelinesApiDirectoryInput, mounts: List[Mount], projectId: String) = {
+      // rsync need the traget directory to exist already, so create it beforehand
```
mcovarr: target

supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/api/ActionBuilder.scala
```diff
@@ -34,11 +34,11 @@ object ActionBuilder {
...
     .setImageUri(image)
 
-  def userAction(docker: String, scriptContainerPath: String, mounts: List[Mount]): Action = {
+  def userAction(docker: String, scriptContainerPath: String, mounts: List[Mount], jobShell: String): Action = {
     new Action()
       .setImageUri(docker)
-      // TODO shouldn't this be using the job shell?
-      .setCommands(List("/bin/bash", scriptContainerPath).asJava)
+      // TODO this should use jobShell instead of "/bin/bash"
+      .setCommands(List("/bin//bash", scriptContainerPath).asJava)
```
mcovarr: two slashes are unlikely to be better than one

Horneth:
> it's expected that the Centaur PAPI tests would require /bin/bash while the PAPI conformance tests should be able to get by with /bin/sh.

@mcovarr Oh I didn't realize that. I'd be curious to know why. Maybe we can tech talk this tomorrow too

Horneth:
@mcovarr oh I see thanks. Well I only saw 4 failure IIRC, so it's not that bad. In the meantime we could always keep `/bin/bash` as the default and set it to`/bin/sh` for CWL conf tests which would make 117 pass

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -362,30 +365,58 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
...
+      // If the original file is not already under callExecutionRoot, then copy it there
+      case (adHoc @ AdHocValue(_, None, _), file) if !file.isChildOf(jobPaths.callExecutionRoot) =>
+        asyncIo.copyAsync(file, jobPaths.callExecutionRoot / file.name) as {
+          Coproduct[StandardAdHocValue](LocalizedAdHocValue(adHoc, jobPaths.callExecutionRoot / file.name))
+        }
+      // If the file should be available as a different name than its own, copy it as that
+      case (adHoc @ AdHocValue(_, Some(alternateName), _), file) =>
+        val finalPath = jobPaths.callExecutionRoot / alternateName
+        // First check that it's not already there under execution root
+        asyncIo.existsAsync(finalPath) flatMap {
```
mcovarr: is there a reason why the previous case uses a blocking `isChildOf` and this does `existsAsync`?

Horneth: Hmm I was assuming that `isChildOf` was purely a path thing to answer the question "is this file located in this path hierarchy ?" and did not involve any I/O. But drilling into it it looks like it's calling `isDirectory` which does involve reading the attributes of the directory and would correspond to an http request for gcs paths.. I would be tempted to say that since the PAPI Backend overrides this to a noop it's likely not going to be an issue for now.
The reason for `isChildOf` is that since we mount the whole call directory in docker, if it's already in there for whatever reason there's no need to re-localize it.

supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/PipelinesUtilityConversions.scala
```diff
@@ -0,0 +1,50 @@
...
+import cromwell.backend.google.pipelines.common.io.PipelinesApiAttachedDisk
+import cromwell.backend.google.pipelines.common.{GpuResource, PipelinesApiRuntimeAttributes}
+import cromwell.backend.google.pipelines.v2alpha1.PipelinesConversions._
+import cromwell.core.ExecutionEvent
+import cromwell.core.logging.JobLogger
+import simulacrum.typeclass
+
+import scala.language.implicitConversions
+
+@typeclass trait EventConversion[A <: Event] {
```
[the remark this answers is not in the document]
danbills: > better than methods on the core types

Agree!  I wouldn't put the method on the type itself, just a one-off function in an object here or maybe in a package object.

supportedBackends/google/pipelines/v1alpha2/src/main/scala/cromwell/backend/google/pipelines/v1alpha2/PipelinesConversions.scala
```diff
@@ -30,22 +31,38 @@ object PipelinesConversions {
...
 
-    override def toGoogleRunParameter(output: PipelinesApiFileOutput): String = output.cloudPath
+    override def toGoogleRunParameter(output: PipelinesApiFileOutput): String = output.cloudPath.pathAsString
+  }
+
+  implicit val outputTo: ToParameter[PipelinesApiOutput] = new ToParameter[PipelinesApiOutput] {
+    override def toGooglePipelineParameter(output: PipelinesApiOutput): PipelineParameter = output match {
+      case fileOutput: PipelinesApiFileOutput => fileOutputTo.toGooglePipelineParameter(fileOutput)
+      case _: PipelinesApiDirectoryOutput =>
+        throw new UnsupportedOperationException("Pipelines API V1 backend does not support PipelinesApiDirectoryOutput parameters. This is a programmer error.")
```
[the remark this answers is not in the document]
Horneth: I don't think so, it would mean that the PAPI common code created a `PipelinesApiDirectoryOutput` which it does not right now, only the V2 specific code does that.
This is due to the mix and match between common PAPI code and version specific code. A LOT of this should be cleaner when we remove V1 support.

supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/PipelinesUtilityConversions.scala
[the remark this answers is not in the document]
Horneth: Yeah I agree I got caught in a typeclass frenziness, will revert to methds

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -340,6 +338,11 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
         |) $stdinRedirection > "$$$out" 2> "$$$err"
         |echo $$? > $rcTmpPath
         |(
+        |# add a .file in every empty directory to facilitate directory delocalization on the cloud
+        |cd $cwd
+        |find . -type d -empty -print | xargs -I % touch %/.file
```
kshakir: 🙈

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -362,30 +365,58 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
...
+    
+    adHocValues.traverse[ErrorOr, (AdHocValue, Path)]({ adHocValue =>
+      // Build an actionable Path from the ad hoc file
+      getPath(adHocValue.womValue.value).toErrorOr.map(adHocValue -> _)
+    })
+      // Localize the values if necessary
+      .map(_.traverse[Future, StandardAdHocValue](localize.tupled)).toEither
+      // TODO: Asynchronify
+      // This is obviously sad but turning it into a Future has earth-shattering consequences, so synchronizing it for now
+      .flatMap(future => Try(Await.result(future, 1.hour)).toChecked)
```
kshakir: Wait, no this is the real 🙈

cromwell.examples.conf
```diff
@@ -649,7 +649,7 @@ backend {
     #      cpu: 1
     #      failOnStderr: false
     #      continueOnReturnCode: 0
-    #      memory: "2 GB"
+    #      memory: "2048 MB"
```
kshakir: ToL: Curious, why the change?

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -362,30 +365,58 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
...
+      .toValidated
+  }
+  
+  def adHocValueToCommandSetupSideEffectFile(adHocValue: StandardAdHocValue) = adHocValue match {
+    case AsAdHocValue(AdHocValue(womValue, alternativeName, _)) =>
+      CommandSetupSideEffectFile(womValue, alternativeName)
+    case AsLocalizedAdHocValue(LocalizedAdHocValue(AdHocValue(womValue, alternativeName, _), _)) =>
+      CommandSetupSideEffectFile(womValue, alternativeName)
+  }
```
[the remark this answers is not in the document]
Horneth: If we have one I don't know about it, but it would definitely be a good thing to have !

cwl/src/main/scala/cwl/CwlType.scala
```diff
@@ -67,10 +56,19 @@ case class File private
 
   lazy val asWomValue: ErrorOr[WomMaybePopulatedFile] = {
     errorOrSecondaryFiles flatMap { secondaryFiles =>
-      val valueOption = path.orElse(location).orElse(basename)
+      val valueOption = path.orElse(location).orElse(path)
```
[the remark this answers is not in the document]
Horneth: 🤦‍♂️

wom/src/main/scala/wom/values/WomObject.scala
```diff
@@ -23,6 +27,18 @@ trait WomObjectLike extends WomValue {
...
   }
+
+  def traverse[R <: WomValue, G[_]](f: WomValue => G[R])(implicit applicative: Applicative[G]): G[WomObjectLike] = {
+    if (values.isEmpty) applicative.pure(this)
+    else {
+      val traverseFunction: (String, WomValue) => G[(String, R)] = {
+        case (key, value) => applicative.map(f(value)) { key -> _ }
+      }
+      applicative.map(values.toList.traverse[G, (String, R)](traverseFunction.tupled)) { mapped =>
+        copyWith(mapped.toMap)
```
kshakir: ToL: Uh... I've got lot of FP homework to do.

wom/src/main/scala/wom/values/WomArray.scala
```diff
@@ -59,7 +62,16 @@ sealed abstract case class WomArray(womType: WomArrayType, value: Seq[WomValue])
...
-  
+
+  def traverse[R <: WomValue, G[_]](f: WomValue => G[R])(implicit applicative: Applicative[G]): G[WomArray] = {
+    if (value.isEmpty) applicative.pure(this)
+    else {
+      applicative.map(value.toList.traverse(f)) { mapped =>
+        WomArray(WomArrayType(mapped.head.womType), mapped)
+      }
+    }
+  }
```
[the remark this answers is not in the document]
Horneth: oh you mean like writing a `Traverse[WomArray]` ? That'd be cool I'll give it a shot
For the type I copied what `map` is doing just above but I suspect you're right

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
Horneth: 😭

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
mcovarr: you might want to save up your 🙈 for my next PR

wom/src/main/scala/wom/values/WomArray.scala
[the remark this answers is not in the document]
Horneth: Unfortunately it doesn't seem possible now because traversable WomTypes are not truly generic. If `WomArray(WomArrayType(WomStringType), List(WomString("hello"))` was instead `WomArray[WomString](WomString("hello"))` I think we could make an instance of `Traverse[WomArray]` and use cats.traverse directly on a `WomArray` value.

wom/src/main/scala/wom/values/WomObject.scala
Horneth: There might be a simpler way to write this tbh

backend/src/main/scala/cromwell/backend/standard/package.scala
Horneth: The upside of this is it can be re-used for different purposes, whereas we need a different Poly1 for each sort of "map" we want to do

PR #3715, opened 2018-05-31
Input expressions with IoFunctions

Horneth:
The `validateWomNamespace` method was using `NoIoFunctionSet` instead of the available ioFunctions, causing the workflow in the centaur test to fail


danbills:
@ruchim @kshakir Yes, @Horneth could you please add to hotfix branch?  They are getting 32 ready for FC, would like to get this in there.

PR #3721, opened 2018-06-04
temporary directory should not be world readable

cpavanrun:
Setting the tmpdir to be word accessible poses a security risk at some high-performance compute clusters.

Original idea seems to make it world writabled: https://github.com/broadinstitute/cromwell/pull/2053

I would even prefer to set it to be *only* accessable for the `cromwell` user (mod: `700`)... or just leave it as it was provided...

### DIFF
--- a/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
+++ b/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
@@ -314,7 +314,7 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
         |DOCKER_OUTPUT_DIR_LINK
         |cd $cwd
         |tmpDir=`$temporaryDirectory`
-        |chmod 777 "$$tmpDir"
+        |chmod 711 "$$tmpDir"
         |export _JAVA_OPTIONS=-Djava.io.tmpdir="$$tmpDir"
         |export TMPDIR="$$tmpDir"
         |export HOME="$home"


danbills:
Hi @cpavanrun , thanks for the contribution!

Are you using docker?  Not having write permission would mean files could only be removed via root access, as docker is executing these services as root.

As such, I think this should be `733`

geoffjentry:
Right. When these issues come up they're part of a tension between people using docker and people using HPC on shared filesystems (who almost always are **not** using docker). 

What we've done in the past has been to detect if the context is docker - if so, stick with something more permissive and if not lock it down. YMMV and all of that

cpavanrun:
If it's a recurring issue whould it not make sense to have it configurable?
1. Patch it on every cromwell release
1. Hard-code a chmod reset in every WDL command

Horneth:
The same way we have a configurable `SCRIPT_EPILOGUE` we could introduce a configurable `SCRIPT_PREFACE` 📖 that would execute before the script. If we export the `TMPDIR` variable before `SCRIPT_PREFACE` is run then backend authors could set whatever permissions they want (or none) on the `TMPDIR`.

cpavanrun:
@Horneth , what would be the difference between the `SCRIPT_PREFACE` and the `SCRIPT_PREAMBLE` be? Is the `SCRIPT_PREAMBLE` used for something different?

Horneth:
@cpavanrun That's a good point, although it looks like the `SCRIPT_PREAMBLE` is  not configurable, only overridable in code by backend implementations.
We could make it configurable too (while making sure that backend implementations still get to add their bit), or create a different one for just that purpose

mcovarr:
I've merged #3735 that only `chmod`s the tmpdir when running on Docker. If it turns out we need a more flexible solution we can revisit the suggestions here.

PR #3725, opened 2018-06-04
Quotation marks when quoting source in error msg

aednichols:
I was just going to ignore this, but then @abaumann mentioned he was confused by the same thing... so why not.

Used to be confusing because it could be referring to a value of type Boolean, rather than literally "Boolean":
```
Expected equal, got "Boolean".

      Boolean conditional = incremented != 0
      ^
```
Used to be confusing because integers could suggest you're using the wrong number of identifiers:
```
Expected identifier, got "1".

    echo "~{one.1}" > ~{one.1}.txt
                ^
```


cjllanwarne:
I think the main reason this is confusing is because it's saying `"Expected <TOKENTYPE> but got <SOURCESTRING>"`.

Maybe it'd be less confusing to say `"Expected <TOKENTYPE> but got <TOKENTYPE> (<SOURCESTRING>)"`?

ie in the above example, `Expected identifier, got integer ('1')`

abaumann:
any additional information helps, so i like the type (value)

aednichols:
Ah, that sounds really nice, but I don't think it's possible: the `actual` is of type `Terminal`, which does not include type information (which makes sense, because once parsing the grammar has broken down you can't make any guarantees that a symbol you find will fit into a defined type)

cjllanwarne:
@aednichols did you try `getTerminalStr`?

aednichols:
Hmm, that comes out to
```
Expected identifier, got type ("Boolean").

    String s Boolean Boolean
                     ^
```
Providing the grammatical "part of speech" (`type`) is useful for people who know what it means, but maybe not for others. Will sleep on it.

aednichols:
Tests are in, waiting for green!

cjllanwarne:
I think the change to `womtool` has triggered a red thumb requirement

PR #3727, opened 2018-06-04
Conformance 97: Delocalize secondary files on PAPI

mcovarr:
Doesn't handle directories yet, that's in conformance 55.


cwl/src/main/scala/cwl/OutputParameterExpression.scala
```diff
@@ -77,22 +78,30 @@ case class OutputParameterExpression(parameter: OutputParameter,
...
-    def fromOutputBinding: ErrorOr[Set[WomFile]] = parameter
+
+    // Ignore the supplied ioFunctionSet and use a custom stubbed IoFunctionSet. This is better than a real I/O function set
+    // because in the context of file evaluation we don't care about the results of these operations. The NoIoFunctionSet
+    // that otherwise would be used throws for all of its operations which doesn't fly for the way our CWL evaluation works.
+    val stubbedIoFunctionSet = new EmptyIoFunctionSet {
+      override def size(path: String): Future[Long] = Future.successful(0L)
+      override def isDirectory(path: String): Future[Boolean] = Future.successful(false)
+      override def ec: ExecutionContext = scala.concurrent.ExecutionContext.global
+    }
```
Horneth: ToL and definitely not to block out this PR:
IIRC the major reason for the existence of the `evaluateFiles` function is to tell PAPI (maybe TES too ?) what files need to be delocalized.
In PAPIv2 at least we still have to do that to some extent but we have a lot more flexibility over the delocalization code that can run on the VM and I would love it if one day we could think of a better way to deal with this than "fake evaluating" all expressions. It could also help making "dynamic outputs" like a tsv of file paths or even something like `read_string(read_string("out"))` work.

supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/api/ActionBuilder.scala
```diff
@@ -63,7 +63,7 @@ object ActionBuilder {
     lazy val copyOnlyIfExists = s"if [[ -e $containerPath ]]; then $copy; fi"
 
     cloudSdkAction
-      .withCommand("/bin/sh", "-c", if (fileOutput.optional) copyOnlyIfExists else copy)
+      .withCommand("/bin/sh", "-c", if (fileOutput.optional || fileOutput.secondary) copyOnlyIfExists else copy)
```
Horneth: ToL: just wondering why we need a `secondary` boolean different from `optional` if eventually we just `||` them.
I was thinking of `optional` as meaning "this is an optional output" but not necessarily tied to being strictly of "optional type".
Again I'm fine with this, just curious if there was another reason I missed :)

cwl/src/main/scala/cwl/OutputParameterExpression.scala
mcovarr: agree 💯there's some serious tech debt that could be made a lot better here.

cjllanwarne: ToL: we could probably refactor `evaluateFiles` to return something other than `WomFile` since that forced us to implement `WomGlobFile` as though it's a `WomFile` but really it isn't - and maybe there are more cases like `OptionallyDelocalizedFile`, `DirectoryToDownload`, `FoFN`, etc - which would then end up being convertible into appropriate bash snippets in the PAPI2 delocalization script

supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/api/ActionBuilder.scala
mcovarr: discussed in person

cwl/src/main/scala/cwl/OutputParameterExpression.scala
mcovarr: Also agree, FileEvaluation was a stab in that general direction but it still requires a `WomFile` parameter which as you point out is not really the right abstraction.

PR #3729, opened 2018-06-05
correct reference to papi config in workbench health actor

danbills:
copying over from hotfix branch,

fix for looking for "config" in the papi provider config


mcovarr:
Per Joel this seems to be working now and I'm certainly a fan of this PR so 👍

PR #3735, opened 2018-06-06
only adjust tmpdir permissions under Docker

mcovarr:
In the spirit of #3721, only adjusts permissions if running under Docker.


backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -306,15 +306,15 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
       s"ln -s $cwd $d"
     } getOrElse ""
 
-    // The `tee` trickery below is to be able to redirect to known filenames for CWL while also streaming
-    // stdout and stderr for PAPI to periodically upload to cloud storage.
-    // https://stackoverflow.com/questions/692000/how-do-i-write-stderr-to-a-file-while-using-tee-with-a-pipe
```
mcovarr: this is an unrelated comment I should have removed in a previous PR

PR #3756, opened 2018-06-08
Query endpoint: Submission time

### LINKED ISSUE #3167: FR: Include submission time in QueryWorkflow responses
It is not currently documented as being returned:

https://github.com/broadinstitute/cromwell/blob/fa4f2bd77557a8d46ff7171c980d0da93ebf9b47/engine/src/main/resources/swagger/cromwell.yaml#L816

### LINKED ISSUE #3693: Query workflows by submitted time
Be able to search for workflows "submitted" after X time (similar to how one can query for "started" after X time). This is to be able to find workflows submitted but not started (ie onHold)

Linked to https://github.com/broadinstitute/cromwell/issues/3167

salonishah11:
**Query workflow by submission time**
This adds supports querying workflows by submission time. This will help find workflows that are submitted but not started yet (i.e. workflows which are in On Hold state). Similar to the start and end timestamps, you can include `submission:<timestamp>` to fetch workflows who have submission time greater or equal to specified timestamp.

**Submission time in WorkflowQueryResponse**
Submission time of a workflow is now included in WorkflowQueryResult, which is part of the response for workflow query.

Closes #3693 
Closes #3167


database/migration/src/main/resources/metadata_changesets/add_submission_timestamp_metadata_summary.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.3.xsd">
+
+    <changeSet author="sshah" id="add_submission_timestamp_in_metadata_entry">
+        <addColumn tableName="WORKFLOW_METADATA_SUMMARY_ENTRY">
+            <column name="SUBMISSION_TIMESTAMP" type="DATETIME(6)">
+                <constraints nullable="true" />
```
kshakir: This isn't needed. Columns are null by default.

database/sql/src/main/scala/cromwell/database/slick/MetadataSlickDatabase.scala
```diff
@@ -157,17 +157,17 @@ class MetadataSlickDatabase(originalDatabaseConfig: Config)
...
+                                             statusMetadataKey: String, labelMetadataKey: String, submissionMetadataKey: String,
+                                             buildUpdatedSummary: (Option[WorkflowMetadataSummaryEntry], Seq[MetadataEntry]) =>
                                                WorkflowMetadataSummaryEntry)
                                             (implicit ec: ExecutionContext): Future[Long] = {
     val likeMetadataLabelKey = labelMetadataKey + "%"
     val action = for {
       previousMetadataEntryIdOption <- getSummaryStatusEntryMaximumId(
         "WORKFLOW_METADATA_SUMMARY_ENTRY", "METADATA_ENTRY")
-      previousMetadataEntryId = previousMetadataEntryIdOption.getOrElse(0L)
+      previousMetadataEntryId = previousMetadataEntryIdOption.getOrElse(-1L)
```
kshakir: Thanks!!

database/sql/src/main/scala/cromwell/database/slick/MetadataSlickDatabase.scala
```diff
@@ -201,10 +201,11 @@ class MetadataSlickDatabase(originalDatabaseConfig: Config)
                                       startTimestampOption: Option[Timestamp],
                                       endTimestampOption: Option[Timestamp],
                                       page: Option[Int],
-                                      pageSize: Option[Int])
+                                      pageSize: Option[Int],
+                                      submissionTimestampOption: Option[Timestamp])
                                      (implicit ec: ExecutionContext): Future[Seq[WorkflowMetadataSummaryEntry]] = {
     val action = dataAccess.queryWorkflowMetadataSummaryEntries(workflowStatuses, workflowNames, workflowExecutionUuids,
-      labelAndKeyLabelValues, labelOrKeyLabelValues, startTimestampOption, endTimestampOption, page, pageSize).result
+      labelAndKeyLabelValues, labelOrKeyLabelValues, startTimestampOption, endTimestampOption, page, pageSize, submissionTimestampOption).result
```
mcovarr: TOL nitpick but in parameter lists that include submission, start, and end timestamps I'd lean toward putting all three next to each other in that (chronological) order.

docs/api/RESTAPI.md
```diff
@@ -886,6 +887,7 @@ Workflow query parameters
 |**name**  <br>*optional*|Returns only workflows with the specified name.  If specified multiple times, returns workflows with any of the specified names.  <br>**Pattern** : `"^[a-zA-Z][a-zA-Z0-9_]*$"`|string|
 |**start**  <br>*optional*|Returns only workflows with an equal or later start datetime.  Can be specified at most once. If both start and end date are specified, start date must be before or equal to end date.|string (date-time)|
 |**status**  <br>*optional*|Returns only workflows with the specified status.  If specified multiple times, returns workflows in any of the specified statuses.|enum (Submitted, Running, Aborting, Failed, Succeeded, Aborted)|
+|**submission**  <br>*optional*|Returns only workflows with an equal or later submission time. Can be specified at most once. If both submission time and start date are specified, submission time should be before or equal to start date.|string (date-time)|
```
cjllanwarne: ToL: also @ruchim - are we planning to make it possible to search for submission times between X and Y? In that case, maybe it makes more sense to call this `submitted_after`?

CHANGELOG.md
```diff
@@ -2,6 +2,18 @@
 
 ## 33 Release Notes
 
+### Query endpoint
+
+#### Query workflows by Submission time
+
+Cromwell now supports querying workflows by submission time. This will help find workflows that are submitted but not started yet (i.e. workflows which are
+in On Hold state). Similar to the start and end timestamps, you can include `submission:<timestamp>` to fetch workflows who have submission time greater or equal to specified timestamp.
```
cjllanwarne: ToL: to keep these short, I would stop at `On Hold state).` and then link directly to the docs for the rest.

docs/api/RESTAPI.md
mcovarr: Presumably start and end would need to change too if we're adding that general capability.

cjllanwarne: I don't know about change, I assume that the least disruptive thing would be to add the rightly named ones (eg `end_before` and `end_after`) but leave in the unqualified ones (eg `end`) as an unfortunate, deprecated option for at least a few releases so that callers get a chance to update)?

ruchim: @cjllanwarne  We definitely have use cases for supporting time ranges. I don't think its required as a part of this issue, and Miguels totally right that we'd likely need to support time range queries for start/end timestamps as well.

Issue #3753, opened 2018-06-08
batch abort


ruchim:
A batch endpoint to abort workflows

mcovarr:
Batched access to the workflow store will likely touch on #3757

PR #3761, opened 2018-06-11
Synchronize access to workflow store 

mcovarr:
The batched heartbeat writer and workflow picker upper both try to lock multiple rows in the workflow store table inside a transaction and were often observed to deadlock. These two workflow store accesses are now routed through an actor that effectively serializes access to the workflow store table (other accesses are not affected). If this manages to run the gauntlet of gulls [batch abort](https://github.com/broadinstitute/cromwell/issues/3753) would likely need to be added to this system.

Known shortcomings:
- ~~Should probably give more thought as to the thread on which the blocking happens.~~ now on the IO dispatcher
- ~~Should consider actor supervision because if this one actor ever dies that will be bad times.~~ default Akka supervision is reasonable here
- ~~May keep one writer Cromwell from tripping over itself but wouldn't keep multiple writer Cromwells from tripping over each other~~ ticketed in #3795

### DIFF
--- a/database/sql/src/main/scala/cromwell/database/slick/WorkflowStoreSlickDatabase.scala
+++ b/database/sql/src/main/scala/cromwell/database/slick/WorkflowStoreSlickDatabase.scala
@@ -72,12 +72,12 @@ trait WorkflowStoreSlickDatabase extends WorkflowStoreSqlDatabase {
     runTransaction(action)
   }
 
-  override def writeWorkflowHeartbeats(workflowExecutionUuids: List[String])(implicit ec: ExecutionContext): Future[Int] = {
+  override def writeWorkflowHeartbeats(workflowExecutionUuids: Set[String])(implicit ec: ExecutionContext): Future[Int] = {
     val optionNow = Option(now)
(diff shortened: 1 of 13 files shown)


database/sql/src/main/scala/cromwell/database/slick/WorkflowStoreSlickDatabase.scala
```diff
@@ -72,12 +72,12 @@ trait WorkflowStoreSlickDatabase extends WorkflowStoreSqlDatabase {
     runTransaction(action)
   }
 
-  override def writeWorkflowHeartbeats(workflowExecutionUuids: List[String])(implicit ec: ExecutionContext): Future[Int] = {
+  override def writeWorkflowHeartbeats(workflowExecutionUuids: Set[String])(implicit ec: ExecutionContext): Future[Int] = {
```
mcovarr: Unrelated to the deadlock issue, redundant heartbeats for a single workflow were not being squashed out. Now they are.

mcovarr:
After a little research it seems the default Akka supervision decider and strategy looks [pretty reasonable](https://github.com/akka/akka/blob/master/akka-actor/src/main/scala/akka/actor/FaultHandling.scala#L156) for this.

Issue #3766, opened 2018-06-12
Strange behavior with optionality and nested conditionals in WDL 1.0


aednichols:
In WDL 1.0, variables defined in nested if-statements are "doubly optional" (`Int??`) and break `select_first()`.

Note that the degree of optionality maxes out at two, regardless of how deep the nesting is.

This workflow is fine
```
workflow Test {

  Int a = 5

  if (true) {
    if (true) {
      if (true) {
        Int b = 5
      }
    }
  }

  Int c = select_first([a, b])
}
```
but this one fails
```
version 1.0

workflow Test {

  Int a = 5

  if (true) {
    if (true) {
      if (true) {
        Int b = 5
      }
    }
  }

  Int c = select_first([a, b])
}
```
with error
```
Failed to process workflow definition 'Test' (reason 1 of 1):
  Failed to process declaration 'Int c = select_first([a, b])' (reason 1 of 1):
    Cannot coerce expression of type 'Int??' to 'Int'
```

PR #3768, opened 2018-06-12
Prevent nested if-statements from creating nested optionals

aednichols:
https://github.com/broadinstitute/cromwell/issues/3766


wdl/transforms/draft3/src/test/scala/wdl/draft3/transforms/wdlom2wom/WdlFileToWomSpec.scala
```diff
@@ -57,6 +57,7 @@ class WdlFileToWomSpec extends FlatSpec with Matchers {
     "input_values" -> anyWomWillDo,
     "passthrough_workflow" -> anyWomWillDo,
     "scatter_var_member_access" -> anyWomWillDo,
+    "nested_conditionals" -> anyWomWillDo,
```
cjllanwarne: If all you want to do is check that a workflow will validate correctly, I'd consider putting this as a valid`WomtoolValidateSpec` case

PR #3769, opened 2018-06-12
Fix path manipulation issue in adhoc file evaluation

Horneth:


Horneth:
This still doesn't work on TES for some reason. I'll make an issue to fix it there too.

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -479,7 +488,7 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
     // Gets the inputs that will be mutated by instantiating the command.
     def mutatingPreProcessor(in: WomEvaluatedCallInputs): Try[WomEvaluatedCallInputs] = {
```
danbills: I think this param can be removed

PR #3770, opened 2018-06-12
plug workflow log filehandle leaks

mcovarr:
Prevent workflow log file handle leakage. There are two distinct workflow log shutdown paths depending on whether the `final_workflow_log_dir` workflow option is set, this addresses both.


engine/src/main/scala/cromwell/engine/workflow/lifecycle/finalization/CopyWorkflowLogsActor.scala
```diff
@@ -42,7 +42,7 @@ class CopyWorkflowLogsActor(override val serviceRegistryActor: ActorRef, overrid
   }
 
   def deleteLog(src: Path) = if (WorkflowLogger.isTemporary) {
-    sendIoCommand(GcsBatchCommandBuilder.deleteCommand(src))
+    sendIoCommand(DefaultIoCommandBuilder.deleteCommand(src))
```
mcovarr: This was actually doing the right thing of deleting the local log file, though for unintuitive reasons. 🙂

PR #3772, opened 2018-06-13
Allow task-only validation in draft 2 WDL

### LINKED ISSUE #3762: Womtool validate does not acept wdl files with only tasks
Since version 32 Womtool validate give this error on wdl file without a workflow: 
`Namespace does not have a local workflow to run`

Is this required by spec? Would be better to still accept this for reusable tasks like this: https://github.com/biowdl/tasks

cjllanwarne:
Allows for and tests `womtool validate draft2_task.wdl`

Fixes #3762


cjllanwarne:
Ignoring the PAPI2 failures resulting from their API change

PR #3777, opened 2018-06-13
Validate variable lookups in draft 2 runtime attributes

### LINKED ISSUE #3326: Task validation only happens at job execution time
This currently only throws an exception when it tries to call the task. Instead, it should not be able to generate a validated WOM graph in the first place:
```wdl
workflow oops {
  call oopsie
}

task oopsie {
  String str
  command { echo ${str} }
  runtime { docker: docker_image }
}
```

cjllanwarne:
Fixes #3326

### DIFF
--- a/womtool/src/test/resources/validate/wdl_draft3/invalid/bad_task/error.txt
+++ b/womtool/src/test/resources/validate/wdl_draft3/invalid/bad_task/error.txt
@@ -1,0 +1,1 @@
+Failed to process task definition 'oopsie' (reason 1 of 1): No generated value handle found for consumed value 'UnlinkedIdentifierHook(docker_image)'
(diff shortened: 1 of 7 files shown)


womtool/src/test/resources/validate/wdl_draft3/invalid/bad_task/error.txt
```diff
@@ -0,0 +1 @@
+Failed to process task definition 'oopsie' (reason 1 of 1): No generated value handle found for consumed value 'UnlinkedIdentifierHook(docker_image)'
```
mcovarr: the v3 error message... 🤕

cjllanwarne: Hmm, I bet I can make this a tiny bit friendlier...

womtool/src/test/resources/validate/wdl_draft3/invalid/bad_task/error.txt
```diff
@@ -1 +1 @@
-Failed to process task definition 'oopsie' (reason 1 of 1): No generated value handle found for consumed value 'UnlinkedIdentifierHook(docker_image)'
+Failed to process task definition 'oopsie' (reason 1 of 1): Value 'docker_image' is never declared. Available values are: ['str']
```
mcovarr: 👌

PR #3778, opened 2018-06-14
Update to papi2 disk type

Horneth:


danbills:
isn't papi v1 using the old enumerated values?

Horneth:
It is, pushed a new commit, it's hacky but should take care of it

PR #3780, opened 2018-06-14
Query Endpoint: Include/Exclude Subworkflows

### LINKED ISSUE #3240: Add ability to include/exclude subworkflows in the query endpoint
The query endpoint currently returns a list of all workflows and subworkflows matching the query parameters. For the job manager UI, we want to show parent workflows in the list view so we include the `parentWorkflowId` in `additionalQueryResultFields` and then filter out any workflows that have a `parentWorkflowId`. This complicates paginating the response in the UI, since you might request a page of 100 workflows where half of them are subworkflows. 

It would be useful to add a parameter to exclude subworkflows from the response so that additional processing is not required after retrieving those results.

salonishah11:
(Affects /query endpoint)

This will support excluding subworkflows from workflow query results using `includeSubworkflows` parameter. By default they are included in the results.

Closes #3240


services/src/main/scala/cromwell/services/metadata/WorkflowQueryKey.scala
```diff
@@ -127,9 +128,14 @@ object WorkflowQueryKey {
       val nels: List[ErrorOr[String]] = values map { v => {
         allowedValues.contains(v).fold(v.validNel[String], v.invalidNel[String])
       }}
-      sequenceListOfValidatedNels("Unrecognized values", nels)
+      sequenceListOfValidatedNels(s"Keys should be from ${allowedValues.toString}. Unrecognized values", nels)
     }
   }
+
+  case object IncludeSubworkflows extends BooleanWorkflowQueryKey[Boolean] {
+    override val name = "Includesubworkflows"
```
cjllanwarne: `Includesubworkflows` or `includeSubworkflows`? Or does it not matter?

services/src/main/scala/cromwell/services/metadata/WorkflowQueryKey.scala
```diff
@@ -182,3 +188,19 @@ sealed trait IntWorkflowQueryKey extends WorkflowQueryKey[Option[Int]] {
   def displayName: String
 }
 
+sealed trait BooleanWorkflowQueryKey[A] extends WorkflowQueryKey[Boolean] {
+  override def validate(grouped: Map[String, Seq[(String, String)]]): ErrorOr[Boolean] = {
+    valuesFromMap(grouped) match {
+      case vs if vs.lengthCompare(1) > 0 => s"Found ${vs.size} values for key '$name' but at most one is allowed.".invalidNel[Boolean]
+      case Nil => true.validNel
+      case v :: Nil => {
```
cjllanwarne: Be careful here - I think `valuesFromMap` will give you a `Seq` whereas (I believe) `:: Nil` only applies to `List` (which is a subtype of `Seq`, so this will work *most of the time*).

I think it'll be safer to, eg:
```scala
case vs if vs.size == 1 => 
  val v = vs.head 
  ...
```

services/src/main/scala/cromwell/services/metadata/WorkflowQueryKey.scala
```diff
@@ -182,3 +188,19 @@ sealed trait IntWorkflowQueryKey extends WorkflowQueryKey[Option[Int]] {
   def displayName: String
 }
 
+sealed trait BooleanWorkflowQueryKey[A] extends WorkflowQueryKey[Boolean] {
+  override def validate(grouped: Map[String, Seq[(String, String)]]): ErrorOr[Boolean] = {
+    valuesFromMap(grouped) match {
+      case vs if vs.lengthCompare(1) > 0 => s"Found ${vs.size} values for key '$name' but at most one is allowed.".invalidNel[Boolean]
+      case Nil => true.validNel
```
cjllanwarne: ToL: Maybe have `def defaultValue` be a trait method on `BooleanWorkflowQueryKey` to allow the default to change depending on query key?

services/src/main/scala/cromwell/services/metadata/WorkflowQueryKey.scala
```diff
@@ -182,3 +188,19 @@ sealed trait IntWorkflowQueryKey extends WorkflowQueryKey[Option[Int]] {
   def displayName: String
 }
 
+sealed trait BooleanWorkflowQueryKey[A] extends WorkflowQueryKey[Boolean] {
```
cjllanwarne: I don't think the `A` is needed here

services/src/main/scala/cromwell/services/metadata/WorkflowQueryKey.scala
geoffjentry: "most of the time" - we have definitely had bugs over the years of this nature, so emphasis on "most". Someone uses the cons operator in a pattern match and someone else passes a non-`List` `Seq` into the function, which is itself specifying a `Seq` input.

services/src/main/scala/cromwell/services/metadata/WorkflowQueryKey.scala
salonishah11: It needs to be `Includesubworkflows` as there is validation logic for recognizing the valid keys (inside WorkflowQueryParameters class) where it converts them to lowercase and capitalizes the first letter

services/src/main/scala/cromwell/services/metadata/WorkflowQueryKey.scala
salonishah11: Oh, I didn't know! Thanks!

Will it then make sense to change similar logic inside DateTimeWorkflowQueryKey and IntWorkflowQueryKey as well?

salonishah11: I converted it to List and kept the original `v :: Nil`, so it's cleaner and consistent with similar code in other traits.

cjllanwarne:
regarding:
> Will it then make sense to change similar logic inside ...

... yes, I think it does

services/src/main/scala/cromwell/services/metadata/WorkflowQueryKey.scala
```diff
@@ -168,17 +175,35 @@ sealed trait SeqWorkflowQueryKey[A] extends WorkflowQueryKey[Seq[A]] {
...
-    valuesFromMap(grouped) match {
+    valuesFromMap(grouped).toList match {
       case vs if vs.lengthCompare(1) > 0 =>
         s"Found ${vs.size} values for key '$name' but at most one is allowed.".invalidNel[Option[Int]]
       case Nil => None.validNel
       case v :: Nil =>
         Try(v.toInt) match {
           case Success(intVal) => if (intVal > 0) Option(intVal).validNel else s"Integer value not greater than 0".invalidNel[Option[Int]]
-          case _ => s"Value given for $displayName does not parse as a integer: $v".invalidNel[Option[Int]]
+          case _ => s"Value given  for $displayName does not parse as a integer: $v".invalidNel[Option[Int]]
```
mcovarr: 🔙

services/src/test/scala/cromwell/services/metadata/impl/MetadataDatabaseAccessSpec.scala
```diff
@@ -318,6 +334,35 @@ class MetadataDatabaseAccessSpec extends FlatSpec with Matchers with ScalaFuture
...
+        // include subworkflows
+        _ <- dataAccess.queryWorkflowSummaries(WorkflowQueryParameters(Seq(WorkflowQueryKey.IncludeSubworkflows.name -> true.toString))) map { case (resp, _) =>
+          val resultByName = resp.results groupBy (_.name)
+          resultByName.keys.toSet.flatten should contain(SubworkflowName)
+        }
+        // exclude subworkflows
+        _ <- dataAccess.queryWorkflowSummaries(WorkflowQueryParameters(Seq(WorkflowQueryKey.IncludeSubworkflows.name -> false.toString))) map {
+          case (resp, _) =>
+            val resultByName = resp.results groupBy (_.name)
+            resultByName.keys.toSet.flatten should not contain SubworkflowName
```
mcovarr: nice test!

salonishah11: Thanks! :)

services/src/main/scala/cromwell/services/metadata/WorkflowQueryKey.scala
salonishah11: This one is 'oops!'. Thanks for pointing it out :)

services/src/main/scala/cromwell/services/metadata/WorkflowQueryKey.scala
```diff
@@ -127,9 +128,15 @@ object WorkflowQueryKey {
       val nels: List[ErrorOr[String]] = values map { v => {
         allowedValues.contains(v).fold(v.validNel[String], v.invalidNel[String])
       }}
-      sequenceListOfValidatedNels("Unrecognized values", nels)
+      sequenceListOfValidatedNels(s"Keys should be from ${allowedValues.toString}. Unrecognized values", nels)
```
danbills: think you can just do `$allowedValues` without braces and `toString`.

salonishah11: 💯 correct! Removed them. Thanks! :)

PR #3788, opened 2018-06-15
Add Cromwell versioning note for localization_optional

cjllanwarne:
Since ReadTheDocs follows develop, not releases, we should be careful not to over-advertise any as-yet-unreleased features

### DIFF
--- a/docs/optimizations/FileLocalization.md
+++ b/docs/optimizations/FileLocalization.md
@@ -28,6 +28,10 @@ None!
 
 Files marked for optional localization are still treated in exactly the same way as other `File` inputs for call caching.
 
+## Cromwell Support
+
+This optimization will be active in Cromwell as of version 33 and higher.
+
 ## Language Support
 
 ### WDL 1.0 (or later)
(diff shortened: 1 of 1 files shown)


docs/optimizations/FileLocalization.md
```diff
@@ -28,6 +28,10 @@ None!
 
 Files marked for optional localization are still treated in exactly the same way as other `File` inputs for call caching.
 
+## Cromwell Support
+
+This optimization will be active in Cromwell as of version 33 and higher.
```
katevoss: I recommend putting this and all notes about Cromwell version support at the top of the file. This way a user sees the availability before reading through the rest of the doc.

Something like:
> Available in Cromwell version 33 and higher.

docs/optimizations/FileLocalization.md
```diff
@@ -1,5 +1,7 @@
 # The 'localization_optional' Optimization
 
+Available in Cromwell version 33 and higher.
```
katevoss: 👍

cjllanwarne: Could you make this again as a top-level comment (or make a "approve" review)?

Looks like pullapprove doesn't count review-comment thumbs :'(

katevoss:
👍 as requested!

PR #3789, opened 2018-06-15
Restore configurable tmpDir behavior

mcovarr:


mcovarr:
I'd also like to cherry pick this onto 32 hotfix if the shorebirds allow.

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -119,7 +119,7 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
 
   lazy val temporaryDirectory = configurationDescriptor.backendConfig.getOrElse(
       path = "temporary-directory",
-      default = s"""mkdir -p "${runtimeEnvironment.tempPath}" && echo "${runtimeEnvironment.tempPath}""""
+      default = s"""`mkdir -p "${runtimeEnvironment.tempPath}" && echo "${runtimeEnvironment.tempPath}"`"""
```
kshakir: ToL-ish: Realize that this is just a refactor, by while you're here, "$()" should also work instead of the deprecated "\`\`"

mcovarr: We used to use "$()" and at one point during my job shell work I convinced myself that it didn't work under one of the Bourne shells I tried, but now I can't find a shell where this doesn't work. 🤔 I'll revert, if an incompatible shell rears its head we can revert the reversion.

PR #3797, opened 2018-06-19
Removed security tags from cromwell and fixed tags in cromiam.

kshakir:


CromIAM/src/main/resources/swagger/cromiam.yaml
```diff
@@ -424,43 +424,45 @@ paths:
...
+          description: Malformed Request
+        '404':
+          description: Workflow ID Not Found
+        '500':
+          description: Internal Error
+      security:
+        - googleoauth:
+          - openid
+          - email
+          - profile
```
kshakir: For anyone checking this is the real change beside fixing the indentation.

"Diff settings" > "Hide whitespace changes" also helps too.

PR #3800, opened 2018-06-20
Query endpoint: Ability to exclude workflows based on labels

### LINKED ISSUE #3608: Ability to do NOT(keyX:ValueY) queries
A key feature for job manager is an ability to 'archive' a job so that it no longer shows in the UI when you don't want to see it anymore. Say it was a failure and you have dealt with the failure. You don't need to see it every time you look at recently failed jobs if you've dealt with it. The idea for implementing this was to have a label key called "flag" and have a value "archived" that we could apply to jobs a user wants to archive. The part we are struggling with is how to filter those out of view in the UI. If it's not difficult, the best way would be to be able to do queries where we say "show me all the jobs that match this query and do not have the label `flag:archived` attached to them"

Is this something that would be possible?

salonishah11:
This gives the ability to filter out workflows based on labels. Two new parameters called `excludeLabelAnd` and `excludeLabelOr` are added.

* excludeLabelAnd: excludes workflows with the specified label. If specified multiple times, excludes workflows with all of the specified label keys
* excludeLabelOr: excludes workflows with the specified label. If specified multiple times, excludes workflows with any of the specified label keys.

Closes #3608


cjllanwarne:
ToL: (sorry for the seagulling...) I found the names of these new filters kinda confusing and it started me thinking - these combinations of the various label filters, their meanings, how they interact if more than one is set, etc, etc... seems likely to explode in number and become harder and harder to manage as more filter types get dreamed up.

At the risk of "if you have a hammer every problem starts to look like a nail"... I wonder whether a simple expression style syntax would make this easier for us to explain and easier for callers to reason about?

I'm imagining something like a labelFilterExpression which could be, e.g.:

("Label1" && "Label2" && !"Label3") || "Label4"

I think that's much easier to read and write, more flexible and more extensible than the current scheme which requires adding new fields for every filter type and a lot of confusion over exact meaning.

There's even a (previously core Scala) library which should make reading those expressions (relatively) painless: https://github.com/scala/scala-parser-combinators

docs/api/RESTAPI.md
```diff
@@ -779,9 +783,13 @@ Workflow query parameters
...
+|**excludeLabelOr**  <br>*optional*|Excludes workflows with the specified label.  If specified multiple times, excludes workflows with any of the specified label keys. Specify the label key and label value pair as separated with "label-key:label-value"  <br>**Pattern** : `"^([a-z][-a-z0-9]*[a-z0-9])?[:]([a-z][-a-z0-9]*[a-z0-9])?$"`|string (array)|
 |**id**  <br>*optional*|Returns only workflows with the specified workflow id.  If specified multiple times, returns workflows with any of the specified workflow ids.  <br>**Pattern** : `"^[0-9A-Fa-f]{8}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{4}-[0-9A-Fa-f]{12}$"`|string|
 |**includeSubworkflows**  <br>*optional*|Include subworkflows in results. By default, it is taken as true.|string (boolean)|
 |**name**  <br>*optional*|Returns only workflows with the specified name.  If specified multiple times, returns workflows with any of the specified names.  <br>**Pattern** : `"^[a-zA-Z][a-zA-Z0-9_]*$"`|string|
+|**page**  <br>*optional*|Page number for results (for paginated query)|string (integer)|
+|**pagesize**  <br>*optional*|Number of results per page (for paginated query)|string (integer)|
```
mcovarr: why are these being added to the doc and swagger now?

salonishah11: I am not sure how long these 2 parameters have been included in query endpoint, but I was just trying to update the Swagger doc (for query endpoint). 

Since I always use Swagger to know data models of request/response and make API requests, it was a bit of surprise for me when I found out last week that there exists page and pagesize parameters in query and I never saw them Swagger UI.

I can remove them if it seems unnecessary to add it here.

cjllanwarne: I think "because they weren't here before" is a totally valid answer!

As long as the fields work when used in swagger, I'd 100% keep them (unless we were trying to hide their existence for some reason?)

salonishah11: The fields do work when used in Swagger. I did test it. Though, I am not sure why it was not included before!

services/src/test/scala/cromwell/services/metadata/impl/MetadataDatabaseAccessSpec.scala
```diff
@@ -262,6 +262,36 @@ class MetadataDatabaseAccessSpec extends FlatSpec with Matchers with ScalaFuture
           val resultByName = response.results groupBy (_.name)
           resultByName.keys.toSet.flatten should equal(Set(Workflow1Name, Workflow2Name))
         }
+        // Filter by exclude label using AND
+        _ <- dataAccess.queryWorkflowSummaries(WorkflowQueryParameters(Seq(
+          WorkflowQueryKey.ExcludeLabelAndKeyValue.name -> s"${testLabel2.key}:${testLabel2.value}"))) map { case (response, _) =>
+          val resultByName = response.results groupBy (_.name)
+          resultByName.keys.toSet.flatten should equal(Set(Workflow1Name))
```
cjllanwarne: Could we have a test which specifies two labels so that we get a different answer for the AND and the OR cases?

salonishah11:
@cjllanwarne Yes, I am removing the pagination from Swagger docs :)

services/src/test/scala/cromwell/services/metadata/impl/MetadataDatabaseAccessSpec.scala
salonishah11: I do have that test, but surprisingly they both return workflows with same `WorkflowName = Workflow1Name (= 'test1')` but different number of `totalResultsCount`. So I am adding that condition check as well inside the same tests.

cjllanwarne: oh, right, because the same workflow can be submitted with different names? Sounds good to me

salonishah11: Awesome! Thanks! :)

PR #3801, opened 2018-06-21
This might help Cromwell endpoints being overwhelmed during centaur testing

Horneth:


geoffjentry:
👍

centaur/src/main/scala/centaur/api/CentaurCromwellClient.scala
```diff
@@ -27,7 +27,7 @@ object CentaurCromwellClient {
   // See https://github.com/akka/akka-http/issues/602
   // And https://github.com/viktorklang/blog/blob/master/Futures-in-Scala-2.12-part-7.md
   final implicit val blockingEc: ExecutionContextExecutor = ExecutionContext.fromExecutor(
-    Executors.newCachedThreadPool(DaemonizedDefaultThreadFactory))
+    Executors.newFixedThreadPool(50, DaemonizedDefaultThreadFactory))
```
mcovarr: TOL There's an existing `-p` parallelism option in `test_cromwell.sh` that seems intended to achieve a similar effect through different means and that's currently set to a value of 100.

Horneth:
Well nevermind, this does not seem to help, still seeing 
```
Exceeded configured max-open-requests value of [128]. This means that the request queue of this pool (HostConnectionPoolSetup(localhost,...
```

We might want to increase the `max-open-requests` conf setting instead ?

PR #3802, opened 2018-06-21
Fix delocalization path issue

Horneth:
This fixes the issue in PAPI2 and revert it to what it was before for PAPI1. Since this change was targeted at CWL and we don't plan to support CWL on PAPI1, it seems safer to revert it entirely in case it still breaks something else

### DIFF
--- a/centaur/src/main/resources/standardTestCases/custom_mount_point/custom_mount_point.wdl
+++ b/centaur/src/main/resources/standardTestCases/custom_mount_point/custom_mount_point.wdl
@@ -1,30 +1,35 @@
-#  1) Mounting a SSD to a custom location
+#  1) Mounting a SSD to a custom location and default location with custom disk size
 #  2) Write a file to that mount
 #  3) Changing directory within a command shouldn't break Cromwell
 #  4) Use the file that was written on the mount as an output
 
+# ChrisTM
 task t {
-  String version
+  String v
 
   command {
+    echo "bazqux" > some_file
+
     cd /some/mnt
     echo "foobar" > some_file
   }
 
   output {
+    String out1 = read_string("some_file")
     String out2 = read_string("/some/mnt/some_file")
   }
 
   runtime {
-    docker: "ubuntu:" + version
+    docker: "ubuntu:" + v
     disks: "local-disk 20 SSD, /some/mnt 20 SSD"
   }
 }
 
 workflow custom_mount_point {
-  call t {input: version="latest"}
+  call t {input: v="latest"}
 
   output {
-     t.out2
+     String o1 = t.out1
+     String o2 = t.out2
    }
 }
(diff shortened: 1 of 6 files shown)


src/bin/ci/resources/build_application.inc.conf
```diff
@@ -1,3 +1,5 @@
+akka.http.host-connection-pool.max-open-requests = 1024
```
mcovarr: stowaway?

centaur/src/main/resources/standardTestCases/custom_mount_point/custom_mount_point.wdl
```diff
@@ -1,30 +1,35 @@
...
   }
 }
 
 workflow custom_mount_point {
-  call t {input: version="latest"}
+  call t {input: v="latest"}
 
   output {
-     t.out2
+     String o1 = t.out1
```
cjllanwarne: Can we also check these outputs in the `.test`:
```
metadata {
  "calls.custom_mount_point.t.backend": "Papi"
  "calls.custom_mount_point.t.backendStatus": "Success"

  "outputs.custom_mount_point.out1.": "bazqux"
  "outputs.custom_mount_point.out1.": "foobar"
}
```

src/bin/ci/resources/build_application.inc.conf
cjllanwarne: Is this related?

supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/PipelinesApiAsyncBackendJobExecutionActor.scala
```diff
@@ -86,4 +87,16 @@ class PipelinesApiAsyncBackendJobExecutionActor(standardParams: StandardAsyncExe
       })
     case _ => List.empty
   }
+
+  override def generateSingleFileOutputs(womFile: WomSingleFile, fileEvaluation: FileEvaluation) = {
```
cjllanwarne: Do we have a test for the original problem that this change was intended to fix when it was made?

Horneth: conformance test 55 (on PAPI2) is the test

src/bin/ci/resources/build_application.inc.conf
Horneth: No but all my PRs keep failing with this `max-open-requests` error so I'm trying to ~~sneak this in~~ fix it

centaur/src/main/resources/standardTestCases/custom_mount_point/custom_mount_point.wdl
cjllanwarne: uh, oops - that's a typo, should be `out2` on the final line!

Issue #3784, opened 2018-06-15
Cromwell tmpdir 31 -> 32 behaviour change


EvanTheB:
In version 31 I had in my configuration: 
```temporary-directory = "$(mkdir -p $TMPDIR && echo $TMPDIR)"```

Which resulted in:
```
tmpDir=$(
  set -e
  cd /share/ScratchGeneral/evaben/cromwell/cromwell-executions/Happy_Workflow/f7b9ac1e-994c-46bd-bad5-e11ac6696165/call-happy/execution
  tmpDir="$(mkdir -p $TMPDIR && echo $TMPDIR)"
  echo "$tmpDir"
)
chmod 777 "$tmpDir"
```

In version 32 the same config results in:
```
cd /share/ScratchGeneral/evaben/cromwell/cromwell-executions/Happy/356aa17f-6276-44e0-9859-391c6c58cf49/call-happy/execution
tmpDir=`$(mkdir -p $TMPDIR && echo $TMPDIR)`
chmod 777 "$tmpDir"
```

which executes using my $() as well as the cromwell provided ``, causing an error (which is no longer caught by set -e). Then many subsequent errors as the script tries to write to / (lucky I did not rm -r!). I thought there was documentation on readthedocs but I cannot find it with the inbuilt search or google.

If I just remove my $(), it should work, but as the change was not announced and there does not seem to be documentation, I wanted to check what the actual contract is.

I would also prefer cromwell did not chmod my tmpdir, but that is a separate issue.

mcovarr:
Hi Evan thanks for reporting this. That change in the way `tmpDir` is calculated actually wasn't actually intentional, let me see if I can move things around so your old config value still works. Also `tmpDir` chmoding will be [turned off](https://github.com/broadinstitute/cromwell/pull/3735) in Dockerless environments with Cromwell 33.

PR #3803, opened 2018-06-21
fix tempdir examples

mcovarr:


cromwell.examples.conf
```diff
@@ -277,13 +277,13 @@ backend {
         # `temporary-directory` creates the temporary directory for commands.
         #
         # If this value is not set explicitly, the default value creates a unique temporary directory, equivalent to:
-        # temporary-directory = "mktemp -d \"$PWD\"/tmp.XXXXXX"
+        # temporary-directory = "$(mktemp -d \"$PWD\"/tmp.XXXXXX)"
```
cjllanwarne: I think this example should be below the "what does this value mean" explanation below

cromwell.examples.conf
```diff
@@ -277,13 +277,13 @@ backend {
...
         #
         # The expression is run from the execution directory for the script. The expression must create the directory
         # if it does not exist, and then return the full path to the directory.
         #
         # To create and return a non-random temporary directory, use something like:
-        # temporary-directory = "mkdir -p /tmp/mydir && echo /tmp/mydir"
+        # temporary-directory = "$(mkdir -p /tmp/mydir && echo /tmp/mydir)"
```
cjllanwarne: ToL is it not possible for Cromwell to just auto-wrap these values as it reads them in? Right now this just feels like a weird thing you have to to remember to do (differently from before) in the config file

cromwell.examples.conf
cjllanwarne: Does this count as a breaking change?

mcovarr: This counts as unbreaking an [unintentionally breaking change](https://github.com/broadinstitute/cromwell/issues/3784) 😛

PR #3810, opened 2018-06-22
Update to release WDL

Horneth:
- Update to WDL 1.0 syntax
- Automate PR to homebrew
- Can be used to release minor versions

The brew task supposes `brew` is installed on the machine where Cromwell runs. There is a [Brew Linux docker image](https://hub.docker.com/r/linuxbrew/linuxbrew/) that maybe could be used but the `git` commands don't seem to be authenticated properly in the container (probably because the ssh keys aren't there). I'm sure this can be worked around, maybe in the next iteration..

This can be fully tested with a different `organization` as long as it has a fork of Cromwell and homebrew-core.


release/release_workflow.wdl
```diff
@@ -42,133 +35,312 @@ task do_release {
...
 
     command <<<
-      curl -o versionFile https://raw.githubusercontent.com/${organization}/${repo}/develop/${file}
-      perl -ne '/\s+cromwellVersion\s*=\s*"(.*)"/ && print $1' versionFile > currentVersion
-      echo $((`cat currentVersion` + 1)) > nextVersion
+      curl -s https://api.github.com/repos/~{organization}/cromwell/releases/latest | jq --raw-output '.tag_name'
     >>>
+    
+    runtime {
+        docker: "broadinstitute/cromwell:cromwell-release-image"
```
kshakir: Alternative to above assuming you still want to use a docker: After the [literal fires](https://twitter.com/lukwam/status/1010286524523565056) are out, let's get a real repo for this image.

I'm all for different versions/tags containing different flavors of some library/executable/artifact.

But this is like someone going to download https://github.com/broadinstitute/cromwell/releases/download/kshakir-special/cromwell-kshakir-special.jar and discovering the jar only contains Minecraft.

release/release_workflow.wdl
```diff
@@ -42,133 +35,312 @@ task do_release {
...
+    input {
+        String organization
+        Boolean majorRelease
+    }
 
     command <<<
-      curl -o versionFile https://raw.githubusercontent.com/${organization}/${repo}/develop/${file}
-      perl -ne '/\s+cromwellVersion\s*=\s*"(.*)"/ && print $1' versionFile > currentVersion
-      echo $((`cat currentVersion` + 1)) > nextVersion
+      curl -s https://api.github.com/repos/~{organization}/cromwell/releases/latest | jq --raw-output '.tag_name'
```
kshakir: Alternative to below: Instead of using a `docker` image, before this line put `which jq | brew install jq`.

While this is possibly installing something that someone doesn't want:
1. `brew` is already installing / uninstalling the `cromwell` cask below.
2. EVERYONE should be using `jq`!

release/release_workflow.wdl
Horneth: >  let's get a real repo for this image.
Agreed, I think it'd be nice to have this as portable as possible and therefore use docker images (eventually maybe even for all the tasks). But until we can get a proper repo we can do what you suggested below and install jq on the machine directly.

Horneth:
@kshakir removed the docker image for now. I'll get us a dockerhub repo for it. I wonder if we could use GCR instead an put it in `broad-dsde-cromwell-dev` ? It would require to be `gcloud auth`ed probably though.

kshakir:
Privately hosted GCR works for me. If devops doesn't care I'm fine putting the image under `broad-dsde-cromwell-dev`

release/release_workflow.wdl
```diff
@@ -42,133 +35,309 @@ task do_release {
...
+    input {
+        String organization
+        Boolean majorRelease
+    }
 
     command <<<
-      curl -o versionFile https://raw.githubusercontent.com/${organization}/${repo}/develop/${file}
-      perl -ne '/\s+cromwellVersion\s*=\s*"(.*)"/ && print $1' versionFile > currentVersion
-      echo $((`cat currentVersion` + 1)) > nextVersion
+      which jq | brew install jq
```
mcovarr: TIL single `|` also works for logical or in bash

cjllanwarne: uh, is that not a pipe?

release/release_workflow.wdl
```diff
@@ -1,34 +1,27 @@
+version 1.0
+
 # Cromwell Release WDL
 # Part 1 of the release process.
 # See also: https://docs.google.com/document/d/1khCqpOYpkCE95pE4a6VfBIxd2zHPWuMILgpisNKCF6c
```
cjllanwarne: This could be `meta` info

release/release_workflow.wdl
```diff
@@ -1,34 +1,27 @@
...
-   # Next version
-   String nextV
-   # Command that will update the appropriate file for the current release
-   String updateVersionCommand
+task do_major_release {
+   input {
+      # Current version being released
+      Int releaseVersion
+   
+      # Can be swapped out to try this on a fork
```
cjllanwarne: Could be `parameter_meta`. Also, does this comment imply a default?

release/release_workflow.wdl
Horneth: I think that's a typo

release/release_workflow.wdl
```diff
@@ -42,133 +35,309 @@ task do_release {
...
+        
+        File cromwellJar
+        File womtoolJar
+    }
+    
+    String branchName = "cromwell-${releaseVersion}"
+    
+    # This is to allow for testing. If the organization is not broadinstitute, meaning we're doing a test,
+    # instead of creating the PR in the official homebrew repo make a PR in the test repo.
+    String headBranch = if (organization == "broadinstitute") then "broadinstitute:${branchName}" else branchName
```
cjllanwarne: I'd encourage `~{}` in `String` interpolations too

Horneth: 👍 Would be awesome if the intelliJ plugin detected those as variables btw :)

release/release_workflow.wdl
```diff
@@ -1,34 +1,25 @@
-# Cromwell Release WDL
-# Part 1 of the release process.
-# See also: https://docs.google.com/document/d/1khCqpOYpkCE95pE4a6VfBIxd2zHPWuMILgpisNKCF6c
```
mcovarr: 👁

Horneth: I put it in  the `meta` section of the workflow, but yeah it's less visible now :/

release/release_workflow.wdl
```diff
@@ -42,137 +33,324 @@ task do_release {
...
+          # Download a template for the homebrew PR
+          curl -o template.md https://raw.githubusercontent.com/brewsci/homebrew-bio/master/.github/PULL_REQUEST_TEMPLATE.md
+          
+          # Check the boxes in the template. White list the boxes to be checked so that if new boxes are added they're not 
+          # automatically checked and should be reviewed manually instead
+          sed -i '' -e '/guidelines for contributing/s/\[[[:space:]]\]/[x]/' \
+          -e "/checked that there aren't other open/s/\[[[:space:]]\]/[x]/" \
+          -e "/built your formula locally/s/\[[[:space:]]\]/[x]/" \
+          -e "/your build pass/s/\[[[:space:]]\]/[x]/" \
+          template.md
```
mcovarr: 😄

release/release_workflow.wdl
mcovarr: ah ok that's cool, I just noticed the deletion but didn't read to the end where it is now 🙂

release/release_workflow.wdl
```diff
@@ -42,137 +33,324 @@ task do_release {
...
+    String baseBranch = "master"
+    String pullRepo = if (organization == "broadinstitute") then "Homebrew" else organization
 
-  String cromwellTemplate = "sed -i '' \"s/cromwellVersion[[:space:]]=.*/cromwellVersion = \\\"<<VERSION>>\\\"/g\" project/Version.scala"
+    command <<<
+        set -e
+        set -x 
+
+        # Clone the homebrew fork
+        git clone git@github.com:~{organization}/homebrew-core.git --depth=100
```
mcovarr: can there be some docs as to what's going on here with cloning a fork of homebrew?

Horneth: Now that I'm looking at the homebrew docs, there is a brew command that does all this automatically -_-
https://docs.brew.sh/How-To-Open-a-Homebrew-Pull-Request

geoffjentry: download

Horneth: It doesn't seem like it could handle the update of womtool in the same formula, so I'm going to leave like this for this release and added a comment in the WDL to look into it.

PR #3830, opened 2018-06-28
Annotate job name with task name

elerch:
This adds more information to the aws batch job name to more easily identify jobs as they move through AWS batch.

### DIFF
--- a/supportedBackends/aws/src/main/scala/cromwell/backend/impl/aws/AwsBatchJob.scala
+++ b/supportedBackends/aws/src/main/scala/cromwell/backend/impl/aws/AwsBatchJob.scala
@@ -197,7 +197,7 @@ final case class AwsBatchJob(jobDescriptor: BackendJobDescriptor,           // W
     val definitionArn = createDefinition(s"""${workflow.callable.name}-${jobDescriptor.taskCall.callable.name}""", taskId)
 
     val job = client.submitJob(SubmitJobRequest.builder()
-                .jobName(sanitize(s"""${workflow.callable.name}-${jobDescriptor.taskCall.callable.name}"""))
+                .jobName(sanitize(jobDescriptor.taskCall.fullyQualifiedName))
                 .parameters(parameters.collect({ case i: AwsBatchInput => i.toStringString }).toMap.asJava)
                 .jobQueue(runtimeAttributes.queueArn)
                 .jobDefinition(definitionArn).build)


supportedBackends/aws/src/main/scala/cromwell/backend/impl/aws/AwsBatchJob.scala
```diff
@@ -194,7 +194,7 @@ final case class AwsBatchJob(jobDescriptor: BackendJobDescriptor,           // W
     val definitionArn = createDefinition(workflow.callable.name)
 
     val job = client.submitJob(SubmitJobRequest.builder()
-                .jobName(sanitize(workflow.callable.name))
+                .jobName(sanitize(s"""${workflow.callable.name}-${jobDescriptor.taskCall.callable.name}"""))
```
cjllanwarne: The callable name isn't necessary unique per call, eg
```
call my_task as a
call my_task as b
```

The callable name is `my_task` both times, you probably want the call name

delagoya: Confused. Do you mean `workflow.callable.name` is not unique or `${workflow.callable.name}-${jobDescriptor.taskCall.callable.name}` is not unique?

cjllanwarne: I would suggest `.jobName(sanitize(jobDescriptor.taskCall.fullyQualifiedName))` to replace the workflow and task names here, since that will be filled in with the full path to the call (but double check that it gives you something sensible).

Eg if this is a call to a task 5 subworkflows deep you want the full context not just the current workflow and the current task name.

delagoya:
I see what you mean. In this case, AWS Batch Job names do not have to be unique, just descriptive enough to point to other sources of information. We are also under a character limit, so the FQN may be truncated more times than the above spec.

cjllanwarne:
Didn't realize that I was blocking this by withholding a thumb.

IMO the (unqualified) call name is strictly better for this purpose than the task name:

```wdl
workflow foo {
  call x as a
  call x as b
  call x as c
}
```

In your scheme you'd have three batch jobs called `x`. Using call name you'd have three jobs called `a`, `b` and `c`. 

On the other hand I don't want to hold you up here. Happy to give this my approval if you see my point and would rather just carry on as-is 😄 .

delagoya:
Over the weekend I’ve actually come around to your way of thinking. May as well fix for that before merge.

delagoya:
While we are at it, it may make sense to add in the workflow UUID to the job.name, since this is really an instantiation of a JobDefinition. What is the best place to get that wf uuid?

Issue #3831, opened 2018-06-28
WOMtool fails to resolve local imports


tlangs:
When WOMtool validates a WDL with local imports at relative directory paths, it fails:
```
$ java -jar womtool-33.1.jar validate WdlWithLocalRelativeImports.wdl 
Failed to import workflow ../../other_dir/TasksToImport.wdl.:
Bad import ../../other_dir/TasksToImport.wdl: ../../other_dir/TasksToImport.wdl is not a valid import
Bad import ../../other_dir/TasksToImport.wdl: ../../other_dir/TasksToImport.wdl is not a valid import
```

cjllanwarne:
I believe this was fixed in #3832

PR #3832, opened 2018-06-28
Added localFileResolver to womtool

tlangs:
Fix for issue https://github.com/broadinstitute/cromwell/issues/3831

### DIFF
--- a/languageFactories/language-factory-core/src/main/scala/cromwell/languages/util/ImportResolver.scala
+++ b/languageFactories/language-factory-core/src/main/scala/cromwell/languages/util/ImportResolver.scala
@@ -30,7 +30,7 @@ object ImportResolver {
   def directoryResolver(directory: Path, allowEscapingDirectory: Boolean = false): ImportResolver = CheckedAtoB.fromErrorOr { path =>
     Try(Paths.get(directory.resolve(path).toFile.getCanonicalPath)).toErrorOr flatMap { absolutePathToFile =>
       val absolutePathToImports = Paths.get(directory.toJava.getCanonicalPath)
-      if ((!allowEscapingDirectory && absolutePathToFile.startsWith(absolutePathToImports)) || allowEscapingDirectory) {
+      if (allowEscapingDirectory || absolutePathToFile.startsWith(absolutePathToImports)) {
         val file = File(absolutePathToFile)
         if (file.exists) {
           File(absolutePathToFile).contentAsString.validNel
--- a/languageFactories/language-factory-core/src/main/scala/cromwell/languages/util/ImportResolver.scala
+++ b/languageFactories/language-factory-core/src/main/scala/cromwell/languages/util/ImportResolver.scala
@@ -13,10 +13,10 @@ import common.Checked
 import common.transforms.CheckedAtoB
 import common.validation.ErrorOr._
 import common.validation.Validation._
-import cromwell.core.path.Path
+import cromwell.core.path.{DefaultPathBuilder, Path}
 import wom.core.WorkflowSource
-import scala.concurrent.duration._
 
+import scala.concurrent.duration._
 import scala.concurrent.Await
 import scala.util.Try
 
@@ -47,16 +47,18 @@ object ImportResolver {
     directoryValidation flatMap { directoryResolver(_).run(path) }
   }
 
-  lazy val localFileResolver: ImportResolver = CheckedAtoB.fromErrorOr { path =>
-    Try(Paths.get(Paths.get(".").resolve(path).toFile.getCanonicalPath)).toErrorOr flatMap { absolutePathToFile =>
+  def relativeFileResolver(wdlPath: Path): ImportResolver = CheckedAtoB.fromErrorOr { relativePath =>
+    Try(Paths.get(wdlPath.getParent.resolve(relativePath).toFile.getCanonicalPath)).toErrorOr flatMap { absolutePathToFile =>
       val file = File(absolutePathToFile)
       if (file.exists) {
         File(absolutePathToFile).contentAsString.validNel
       } else {
-        s"Import file not found: $path".invalidNel
+        s"Import file not found: $relativePath".invalidNel
       }
     }
   }
+
+  lazy val localFileResolver: ImportResolver = relativeFileResolver(DefaultPathBuilder.build(Paths.get(".")))
 
   def specificFileResolver(filePath: Path): ImportResolver = CheckedAtoB.fromErrorOr { path =>
     if (path == filePath.toAbsolutePath.toString || path == filePath.name) {
--- a/languageFactories/language-factory-core/src/main/scala/cromwell/languages/util/ImportResolver.scala
+++ b/languageFactories/language-factory-core/src/main/scala/cromwell/languages/util/ImportResolver.scala
@@ -13,7 +13,7 @@ import common.Checked
 import common.transforms.CheckedAtoB
 import common.validation.ErrorOr._
 import common.validation.Validation._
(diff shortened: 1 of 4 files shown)


cjllanwarne:
Thanks for fixing this!

Would you be willing to also add a test case for draft-2 and/or 1.0 to https://github.com/broadinstitute/cromwell/tree/develop/womtool/src/test/resources/validate to make sure this doesn't regress again?

You shouldn't need to do any wiring other than dropping a new test case into an appropriate `valid/testname` directory.

womtool/src/test/resources/validate/wdl_draft2/valid/relative_local_import/relative_local_import.wdl
```diff
@@ -0,0 +1,11 @@
+import "./womtool/src/test/resources/validate/wdl_draft2/valid/task_only/task_only.wdl" as imports
```
cjllanwarne: I suspect this would have worked regardless of the change you made, since it was incorrectly validating that the WDL being imported was in a sub-directory of the current WDL's location.

I think making the path relative to the current file would be a good test since it lets you backtrack a little, eg
```
import ../task_only/task_only.wdl
```

womtool/src/test/resources/validate/wdl_draft2/valid/relative_local_import/relative_local_import.wdl
```diff
@@ -0,0 +1,11 @@
...
+
+workflow http_import {
+
+  File in_file
+
+  call imports.task_only {
+    input:
+      in_file = in_file
+  }
+}
```
cjllanwarne: Request to please add a newline to these two files!

womtool/src/test/resources/validate/wdl_draft2/valid/relative_local_import/relative_local_import.wdl
```diff
@@ -0,0 +1,11 @@
+import "./womtool/src/test/resources/validate/wdl_draft2/valid/task_only/task_only.wdl" as imports
+
+workflow http_import {
```
Horneth: Nitpick: This workflow name doesn't seem to match what the test is

womtool/src/test/resources/validate/wdl_draft2/valid/relative_local_import/relative_local_import.wdl
tlangs: I tried that, but it fails because the tool is looking for files relative to womtool rather than the WDL itself. I'm just restoring the previous behaviour. I've verified that the new test actually tests `localFileResolver` through the debugger. Should I make it so that WOMtool looks for files relative to the WDL file?

womtool/src/test/resources/validate/wdl_draft2/valid/relative_local_import/relative_local_import.wdl
tlangs: Done!

tlangs:
I've made it so WOMtool will resolve files relative to the main WDL file. The tests now test that instead. However, this is a new behaviour and I'm not sure if its what you guys want.

womtool/src/test/resources/validate/wdl_draft2/valid/relative_local_import/relative_local_import.wdl
cjllanwarne: Oh right, because the `localFileResolver` is fixed relative to PWD...

On consideration I think the best answer is to give the DirectoryResolver a `checkWhetherThisFileIsInASubdirectory: Boolean` parameter so that womtool can allow escaping their directory but server mode cannot

cjllanwarne:
@tlangs I think the best answer may be to change the [directoryResolver](https://github.com/broadinstitute/cromwell/blob/develop/languageFactories/language-factory-core/src/main/scala/cromwell/languages/util/ImportResolver.scala#L33) to allow us to customize the "don't escape the directory" validation when womtool calls it only, ie:
```scala
def directoryResolver(directory: Path, allowEscapingDirectory: Boolean = false) = { 
...
  if (absolutePathToFile.startsWith(absolutePathToImports) || allowEscapingDirectory) {
    ...
  } else {
    ...
  }
}
```

cjllanwarne:
(because that will allow us to import relative to PWD *and* the WDL file)

languageFactories/language-factory-core/src/main/scala/cromwell/languages/util/ImportResolver.scala
```diff
@@ -27,10 +27,10 @@ object ImportResolver {
     directoryResolver(LanguageFactoryUtil.validateImportsDirectory(zippedImports).toEither)
   }
 
-  def directoryResolver(directory: Path): ImportResolver = CheckedAtoB.fromErrorOr { path =>
+  def directoryResolver(directory: Path, allowEscapingDirectory: Boolean = false): ImportResolver = CheckedAtoB.fromErrorOr { path =>
     Try(Paths.get(directory.resolve(path).toFile.getCanonicalPath)).toErrorOr flatMap { absolutePathToFile =>
       val absolutePathToImports = Paths.get(directory.toJava.getCanonicalPath)
-      if (absolutePathToFile.startsWith(absolutePathToImports)) {
+      if ((!allowEscapingDirectory && absolutePathToFile.startsWith(absolutePathToImports)) || allowEscapingDirectory) {
```
cjllanwarne: I'm pretty sure that the `(!allowEscapingDirectory &&` at the start is redundant. Though if you want it for the short circuiting, maybe make the whole expression `if (allowEscapingDirectory || absolutePathToFile...) {`

womtool/src/main/scala/womtool/input/WomGraphMaker.scala
```diff
@@ -27,7 +27,10 @@ object WomGraphMaker {
     // - Where the file is
     lazy val importResolvers = List(
       directoryResolver(DefaultPathBuilder.build(Paths.get("."))),
```
cjllanwarne: Allow this one to escape directory too

womtool/src/test/resources/validate/wdl_draft2/valid/relative_local_import/relative_local_import.wdl
```diff
@@ -0,0 +1,11 @@
+import "../task_only/task_only.wdl" as imports
```
cjllanwarne: Nice!

languageFactories/language-factory-core/src/main/scala/cromwell/languages/util/ImportResolver.scala
tlangs: It was for the short-circuiting, but this is better.

cjllanwarne: | x | y | x \|\| y | (!y && x) \|\| y |  
|---|---|:--------:|:----------------:|
| F | F | F      | F              |  
| F | T | T      | T              |   
| T | F | T      | T              |   
| T | T | T      | T              |

tlangs: good 'ol boolean algebra

cjllanwarne:
@Horneth since you already had a look, could you pass over this PR again for thumb-worthiness?

tlangs:
Do I merge this, or is that your teams job?

gemmalam:
@tlangs Let's leave this merged PR to @cjllanwarne 

I'll make sure it gets merged ASAP

PR #3837, opened 2018-06-30
Wired a Demo DOS file system into the PAPI V2 backend.

kshakir:


kshakir:
Brain dump on how this PR would be used:
https://drive.google.com/open?id=1MBsgFHhMtSZPK1Egfr4NCkH3C4zN5NDW

Happy to move the docs into `/docs/` if that's more appropriate?

filesystems/demo-dos/src/main/scala/cromwell/filesystems/demo/dos/DemoDosPathBuilder.scala
```diff
@@ -0,0 +1,73 @@
...
+import com.google.cloud.storage.contrib.nio.{CloudStorageConfiguration, CloudStoragePath}
+import cromwell.core.path.{NioPath, Path, PathBuilder}
+import cromwell.filesystems.demo.dos.DemoDosPathBuilder._
+import cromwell.filesystems.gcs.{GcsPath, GcsPathBuilder}
+
+import scala.util.{Failure, Success, Try}
+
+class DemoDosPathBuilder extends PathBuilder {
+  private val mockStorage: Storage =
+    new Storage.Builder(new MockHttpTransport, new MockJsonFactory, null)
```
Horneth: OOC why a mock storage and not a real one here ?
Is it going to work with all types of Credentials passed from workflow options etc.. ?

kshakir: The Demo DOS file system implementation cannot and does not actually talk to the cloud for accessing file contents/listing. Plus, it only supports a subset of path manipulation. Both path-without-scheme & path-as-string are supported. Unsupported operations include resolve-sibling & get-parent.

Back to accessing DOS file data, the mocks should never be invoked by Cromwell. This demo implementation wraps GCS as it is the most stable file system, but could have wrapped S3 or OSS. Or I could have copy/pasted yet another filesystem beyond the three copies we have already. Ideally woodard will halt the need to blindly copy `java.nio.Path` or `cromwell.core.path.Path` implementations in the future. 😉

supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/PipelinesParameterConversions.scala
```diff
@@ -26,7 +27,25 @@ trait PipelinesParameterConversions {
         Key.Tag -> Value.Localization,
         Key.InputName -> fileInput.name
       )
-      cloudSdkBashAction(localizeFile(fileInput.cloudPath, fileInput.containerPath))(mounts, labels = labels)
+
+      if (fileInput.cloudPath.pathAsString.startsWith("dos://")) {
+        import cromwell.backend.google.pipelines.v2alpha1.api.ActionCommands.ShellPath
+        val config = ConfigFactory.load
+        val demoDosDockerImage = config.getString("demo.dos.docker-image")
+        val demoDosCommandTemplate = config.getString("demo.dos.command-template")
```
danbills: Would you consider making these abstract strings on the trait and filled in elsewhere?  It'd be nice to not add static(-ish, I realize it's in a def) config loads, and if inevitable, make the caller do it and provide these values.

kshakir: Sure, will go over a plan tomorrow. Happy to implement whatever for the demo.

I changed course and tried to keep this PR as isolated as technically possible. I'm keeping one eye on what will be needed for `read_lines` support etc post demo. I foresee possibly (probably?) refactoring a bit of this code to something more permanent... something... if only... 🤔

Will likely pull in @Horneth for his expertise too.

supportedBackends/google/pipelines/v2alpha1/src/test/resources/reference.conf
```diff
@@ -0,0 +1,2 @@
+demo.dos.docker-image = "somerepo/dos-downloader:tagged"
```
salonishah11: So this will be the docker image link containing the environment for running the Scala script?

supportedBackends/google/pipelines/v2alpha1/src/test/resources/reference.conf
```diff
@@ -0,0 +1,2 @@
...
+demo.dos.command-template = "/path/to/some_executable before args ${cloudPath} middle args ${containerPath} ends args"
```
salonishah11: And this runs the script along with passing the dos url and download location values?

supportedBackends/google/pipelines/v2alpha1/src/test/resources/reference.conf
salonishah11: Nevermind. I found answer to both my questions in the link mentioned in comment! Should have taken a look at it first!

supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/PipelinesParameterConversions.scala
kshakir: Created a ticket for following up on this comment: https://github.com/broadinstitute/cromwell/issues/3858

Merging as is for now.

PR #3839, opened 2018-07-01
Correct Mysql version in Persistent Server Doc

danbills:
pointing to version `latest` doesn't work, fixing to `5.5` does

### DIFF
--- a/docs/tutorials/PersistentServer.md
+++ b/docs/tutorials/PersistentServer.md
@@ -13,12 +13,10 @@ Cromwell remembers everything it knows!
 
 ### Let's get started!
 
-- Pull the MySQL docker image from dockerhub:
-`docker pull mysql:5.5`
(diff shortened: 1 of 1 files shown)


docs/tutorials/PersistentServer.md
```diff
@@ -13,12 +13,10 @@ Cromwell remembers everything it knows!
 
 ### Let's get started!
 
-- Pull the MySQL docker image from dockerhub:
-`docker pull mysql:5.5`
```
mcovarr: definitely don't need this?

kshakir: Not needed. `docker run` pulls images if they aren't local already

mcovarr:
Sounds like we need a ticket to not recommend/require a soon to be EOL'd version of MySQL for persistence.

PR #3852, opened 2018-07-03
Share new base transforms between WDL 1.0 and 1.1

cjllanwarne:
Allows reading of WDL 1.0 and 1.1 `Ast`s through a shared set of `CheckedAtoB` functions, with the flexibility to inject different transform behavior into each usage of the instantiations of the transforms.

Note: of the 9013 added lines, 7484 are the new 1.1 WDL parser and presumably about 810 are files which got moved out of draft-3/transforms and into base/transforms


Horneth:
@cjllanwarne Do you need a red review on this ? If so could you elaborate just a little on what 
> Allows reading of WDL 1.0 and 1.1 Asts through a shared set of CheckedAtoB functions, with the flexibility to inject different transform behavior into each usage of the instantiations of the transforms.

means for me poor mortal and / or point to relevant code that I should look at ? 😄

cjllanwarne:
The WDL map might help:

wdlmap

This PR splits the top two circles into two parallel paths: `Draft3 file -> Draft3 ASTs` and `Biscayne file -> Biscayne ASTs` (I'm using `Biscayne` as my codename for "whatever WDL version is next, probably 1.1", because national parks are cool). The ASTs can then go through a single transform function into a shared WDL Object Model.

The clever part is that the transform function can (but doesn't yet) be customised by replacing specific "AST to WDLOM" transformations with version specific functions (those are the `CheckedAtoB`s, and that's why the two lists of `CheckedAtoB`s each version uses are identical... for now)

wdl/transforms/biscayne/src/main/scala/wdl/transforms/biscayne/ast2wdlom/BiscayneGenericAstNode.scala
```diff
@@ -0,0 +1,34 @@
+package wdl.transforms.biscayne.ast2wdlom
```
aednichols: Could we use type parameters to reduce duplication between this file and `Draft3GenericAstNode.scala`? The only difference seems to be the type of the `Ast` (`wdl.draft3.parser.WdlParser.Ast` vs. `wdl.biscayne.parser.WdlParser.Ast`).

Incomplete but illustrates what I'm getting at:
```
case class BaseGenericAst[T <: GenericAstNode](ast: T) extends GenericAst {
  override def getAttribute(attr: String): GenericAstNode = Option(ast.getAttribute(attr)).map(T.apply).orNull
  override def getAttributes: Map[String, GenericAstNode] = ast.getAttributes.asScala.toMap collect {
    case (key, value) if value != null => key -> T(value)
  }
  override def getName: String = ast.getName
}
```

wdl/transforms/biscayne/src/main/scala/wdl/transforms/biscayne/ast2wdlom/ast2wdlom.scala
```diff
@@ -0,0 +1,62 @@
+package wdl.transforms.biscayne
```
aednichols: For consistency and ease of navigation, it probably makes sense to unify the file name between Biscayne and draft-3:

`wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/ast2wdlom/package.scala`
`wdl/transforms/biscayne/src/main/scala/wdl/transforms/biscayne/ast2wdlom/ast2wdlom.scala`

wdl/transforms/new-base/src/main/scala/wdl/transforms/base/linking/expression/values/EngineFunctionEvaluators.scala
```diff
@@ -205,7 +205,7 @@ object EngineFunctionEvaluators {
           read <- readFile(fileToRead, ioFunctionSet, fileSizeLimitationConfig.readBoolLimit)
           asBool <- Try(read.trim.toBoolean)
         } yield WomBoolean(asBool)
-        tryResult.map(EvaluatedValue(_, Seq.empty)).toErrorOr.contextualizeErrors(s"""read_boolean("${fileToRead.value}")""")
+        tryResult.map(EvaluatedValue(_, Seq.empty)).toErrorOr.contextualizeErrors(s"""read_int("${fileToRead.value}")""")
```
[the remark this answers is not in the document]
cjllanwarne: yeah... very good catch! I have no idea what caused this but it's definitely not right...

PR #3856, opened 2018-07-03
Make localization retries configurable

Horneth:


cromwell.examples.conf
```diff
@@ -624,6 +624,11 @@ backend {
     #      # Restrict access to VM metadata. Useful in cases when untrusted containers are running under a service
     #      # account not owned by the submitting user
     #      restrict-metadata-access = false
+    #      
+    #      # Pipelines v2 only: specify the number of times localization and delocalization operations should be attempted
+    #      # There is no logic to determine if the error was transient or not, everything is retried upon failure
+    #      # Defaults to 3
```
kshakir: I would put this into somewhere in docs too. From briefly looking, maybe "Let's get started!" on [/docs/tutorials/PipelinesApi101.md](https://github.com/broadinstitute/cromwell/blob/b830603f77eeb6af7c7721049a1d683efb3bdbbc/docs/tutorials/PipelinesApi101.md)?

supportedBackends/google/pipelines/common/src/test/scala/cromwell/backend/google/pipelines/common/PipelinesApiAttributesSpec.scala
```diff
@@ -55,6 +55,14 @@ class PipelinesApiAttributesSpec extends FlatSpec with Matchers {
 
   }
 
+  it should "parse localization-attempts" taggedAs IntegrationTest in {
+    val backendConfig = ConfigFactory.parseString(configString(genomics = "localization-attempts = 31380"))
+
+    val jesAttributes = PipelinesApiAttributes(googleConfig, backendConfig)
+    jesAttributes.localizationConfiguration.localizationAttempts.value should be(31380)
```
kshakir: ToL: This is copy/paste, but we should stop using "JES". Same goes for other instances in this PR. I'd be ok w/ separate ticket/PR to clean up the remaining references.

supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/GenomicsFactory.scala
```diff
@@ -16,7 +17,7 @@ import mouse.all._
 
 import scala.collection.JavaConverters._
 
-case class GenomicsFactory(applicationName: String, authMode: GoogleAuthMode, endpointUrl: URL) extends PipelinesApiFactoryInterface
+case class GenomicsFactory(applicationName: String, authMode: GoogleAuthMode, endpointUrl: URL)(implicit localizationConfiguration: LocalizationConfiguration) extends PipelinesApiFactoryInterface
```
kshakir: Very-ToL: I've been burned so may times by implicits, including compile time, that I cringe at them. However this implementation is very "Scala" and matches multiple other libraries that I've seen, so LGTM.

Horneth: Yeah I thought this was going to be controversial, you're not the first one to bring this up 😄 . We do this in other places where people don't seem to have a problem with it (like all our database methods have an implicit execution context), and this seemed similar to me. I can see that it could potentially cause trouble if we had to choose between multiple configurations and the wrong one would implicitly be picked up (is that what's burned you before ?). Here there is only one of those per backend instance so we really should not be in that case.
But again people always have stronger opinions than me about this so happy to make it non-implicit.

mcovarr: I share Khalid's disease with implicits in general but that doesn't rise to the level of a 'strong opinion'. 🙂

PR #3857, opened 2018-07-03
Set endpoint url on v1

Horneth:


Horneth:
Should always have been there it's a bug, I just happened to see that it was set for v2 but not v1 today

PR #3864, opened 2018-07-05
Make find ' ' in filename safe

EvanTheB:
I cannot test this change as I do not have the build environment set up. Is is intentional that this 'cloud' feature is inherited by the SGE backend? I am fixing an issue caused by a feature I do not need. I wonder if it could be placed in the epilogue for user configurations. Isn't that the purpose of epilogue and configuration?

### DIFF
--- a/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
+++ b/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
@@ -351,7 +351,7 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
         |(
         |# add a .file in every empty directory to facilitate directory delocalization on the cloud
         |cd $cwd
-        |find . -type d -empty -print0 | xargs --null -I % touch %/.file
+        |find . -type d -empty -print0 | xargs -0 -I % touch %/.file
         |)
         |(
         |cd $cwd
--- a/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
+++ b/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
@@ -351,7 +351,7 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
         |(
         |# add a .file in every empty directory to facilitate directory delocalization on the cloud
         |cd $cwd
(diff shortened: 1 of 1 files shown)


cjllanwarne:
This does indeed feel PAPI specific - @Horneth would you agree with "this could be in a shared PAPI epilogue"?

kshakir:
PAPI, AWS, Alibaba Cloud, etc.

Horneth:
I've been seeing the configurable epilogue more as a "hey user, here's a place for you to add stuff for your specific setup" rather than "you absolutely need your epilogue config to have this if you want your backend to work", but maybe they're not that far apart after all.
I'd be ok with making it a "required" epilogue for cloud backends as long as we make it pretty clear in the changelog that it's required and that a config update is necessary. Also I'm not too worried that anyone has been relying on this anyway since it's a hack to support empty directories which hopefully is not wildly used.

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -351,7 +351,7 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
         |(
         |# add a .file in every empty directory to facilitate directory delocalization on the cloud
         |cd $cwd
-        |find . -type d -empty -print | xargs -I % touch %/.file
+        |find . -type d -empty -print0 | xargs --null -I % touch %/.file
```
mcovarr: The `--null` version of the "Input file names are terminated by a null character" option is a GNU nicety that isn't supported by some versions of `xargs` like that in `alpine:latest`. The `-0` equivalent appears to be more portable.

mcovarr:
Agree with Thibault's description of how the epilogue was intended to allow for tweaks for specific environments.

EvanTheB:
Given that it is a hack that turns empty directories into non-empty directories presumably there exists some user who will be surprised by the default behaviour one day.

geoffjentry:
@EvanTheB FWIW **we** were surprised :) When we were fighting CWL's `Directory` type and object stores it turned out that this was an ubiquitous hack to handle the situation (one could argue that if everyone needs to make a hack, perhaps the problem should be fixed at the source, but ....). However my point here is that it seems to be an increasingly common thing to do so perhaps not as surprising as one would think.

Or perhaps that's just a rationalization :)

PR #3870, opened 2018-07-06
Pulled wes2cromwell shim into Cromwell repo

geoffjentry:
**REVIEWERS: NB!!!!**  With a few minor exceptions in order to make our stricter compiler checks pass this was brought in as-is. **All** comments on the code shall be treated as ToL unless you're volunteering to clean it up yourself. However, feel free to seagull the SBT structure and such.


cjllanwarne:
ToL: it looks like there's a *lot* of overlap between this and the `cromwell.api.CromwellClient` classes (imported by Centaur, for example) - especially around the ADT definitions and the JSON tranforms - which would ideally be reconciled before either one is developed much further

Horneth:
Re cromwell client: FWIW I really like the idea of having this code auto-generated from the swagger definition. Or even have another format than swagger if for whatever reason we don't want to generate it from that.

geoffjentry:
@Horneth See #3860 - someone needs to fix our swagger :)

geoffjentry:
Closing this - was pointed out to me that I shouldn't have copied `master` over

geoffjentry:
Now with correct code

PR #3875, opened 2018-07-09
Fix read_json coercion into Maps

### LINKED ISSUE #3306: reads_json does not work
Whenever I try to use read_json it does not work.
For instance, when I make a simple wdl like:
```wdl
workflow quantification {

    File references
    Map[String, String] indexes = read_json(references)
}
where test.json is:
```json
{
  "quantification.references": "/home/antonkulaga/rna-seq/workflows/rna-seq/quantification/inputs/test.json"
}
```
```
it tells
```
Unable to build WOM node for Declaration 'indexes': Return type of read_json() can't be known statically
	at cromwell.engine.workflow.lifecycle.materialization.MaterializeWorkflowDescriptorActor.cromwell$engine$workflow$lifecycle$materialization$MaterializeWorkflowDescriptorActor$$workflowInitializationFailed(MaterializeWorkflowDescriptorActor.scala:211)
	at cromwell.engine.workflow.lifecycle.materialization.MaterializeWorkflowDescriptorActor$$anonfun$2.applyOrElse(MaterializeWorkflowDescriptorActor.scala:181)
	at cromwell.engine.workflow.lifecycle.materialization.MaterializeWorkflowDescriptorActor$$anonfun$2.applyOrElse(MaterializeWorkflowDescriptorActor.scala:176)
	at scala.runtime.AbstractPartialFunction.apply(AbstractPartialFunction.scala:34)
```
That means I cannot just load a file and parse it, so how else can I read JSON then?

### PR DESCRIPTION
**Pull request by cjllanwarne**
Closes #3306 

- [x] Requires rebase after merging #3895 into develop

### DIFF
--- a/centaur/src/main/resources/standardTestCases/draft3_read_and_coerce_json.test
+++ b/centaur/src/main/resources/standardTestCases/draft3_read_and_coerce_json.test
@@ -1,0 +1,16 @@
+name: draft3_read_and_coerce_json
+testFormat: workflowsuccess
+workflowType: WDL
+workflowTypeVersion: 1.0
+tags: ["wdl_1.0"]
+
+files {
+  workflow: wdl_draft3/draft3_read_and_coerce_json/draft3_read_and_coerce_json.wdl
+}
+
+metadata {
+  workflowName: draft3_read_and_coerce_json
+  status: Succeeded
+  "outputs.draft3_read_and_coerce_json.from_coerced_map": "value1"
+  "outputs.draft3_read_and_coerce_json.from_coerced_foo": "value2"
+}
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: wdl/transforms/draft3/src/main/scala/wdl/draft3/transforms/linking/expression/values/LookupEvaluators.scala [hunk 1]
```diff
@@ -112,6 +114,11 @@ object LookupEvaluators {
       case WomObject(_, _) => s"'Object'-type value did not contain the field '$key' at runtime".invalidNel
       case p: WomPair if key == "left" => p.left.validNel
       case p: WomPair if key == "right" => p.right.validNel
+      case WomMap(_, value) => Validated.fromOption(
+        o = value.collectFirst {
```
aednichols: Would there be any value in checking for duplicate keys here, or is there an earlier verification that renders them impossible?

aednichols: i.e. use `collect` instead of `collectFirst` and assert size is 1

PR #3890, opened 2018-07-12
PAPI v2 Private Docker

mcovarr:


server/src/main/scala/cromwell/CromwellEntryPoint.scala
```diff
@@ -93,6 +94,8 @@ object CromwellEntryPoint extends GracefulStopSupport {
       case t: Throwable =>
         Log.error("Failed to instantiate Cromwell System. Shutting down Cromwell.")
         Log.error(t.getMessage)
+        val stackTrace = ExceptionUtils.getStackTrace(t)
+        Log.error(stackTrace)
```
kshakir: I haven't tried it, but these four lines feel like `Log.error(s"Failed to instantiate Cromwell System. Shutting down Cromwell.", t)`. If however the four lines output something specific that the one line does not then this change is fine.

mcovarr: I suspect you're right. 🙂

PR #3897, opened 2018-07-13
Note an IntelliJ gotcha

jmthibault79:


docs/Building.md
```diff
@@ -37,3 +37,17 @@ $ sbt server/docker
 ```
 
 This will build and tag a Docker image with a name like `broadinstitute/cromwell:<VERSION>-SNAP`.
+
+###IntelliJ gotchas
+
+If you develop Cromwell using IntelliJ IDEA, you may the following error message on startup:
```
mcovarr: verb AWOL

docs/Building.md
```diff
@@ -37,3 +37,17 @@ $ sbt server/docker
...
+
+```bash
```
mcovarr: Why `bash`?

docs/Building.md
aednichols: Maybe specify "when starting SBT" to make clear that it's not Cromwell itself that's missing adequate memory

kshakir: `4100 MB` is a lot. That much memory should only be required for `sbt clean test …`<sup>1</sup>. I believe the IDE shouldn't need that much for just dumping our complex sbt project. I've been able to get away (so far) with `1536`, which is just double IntelliJ IDEA's default `768`.

---
<sup>1</sup> In addition to the project and compilation using a lot of memory the tests run within the sbt process. Once they're done the tests do not seem to be cleaning up after themselves correctly. When other sbt tasks run after `test` there often isn't enough memory "left", thus `4g`.

docs/Building.md
```diff
@@ -37,3 +37,17 @@ $ sbt server/docker
...
+Error occurred during initialization of VM
+Initial heap size set to a larger value than the maximum heap size
+```
+
+This is caused by a conflict between our custom `.sbtopts` file and IntelliJ's defaults.
+Go to **Preferences** -> **Build, Execution, Deployment** -> **sbt** -> **JVM** and increase the value of `Maximum heap size, MB` above 
+that of the `Xmx` option in `.sbtopts`.  Since that value is currently `4g`, a `Maximum heap size` value of `4100 MB` will be sufficient.
```
kshakir: `4100 MB` is a lot. That much memory should only be required for `sbt clean test …`<sup>1</sup>. I believe the IDE shouldn't need that much for just dumping our complex sbt project. I've been able to get away (so far) with `1536`, which is just double IntelliJ IDEA's default `768`.

---
<sup>1</sup> In addition to the project and compilation using a lot of memory the tests run within the sbt process. Once they're done the tests do not seem to be cleaning up after themselves correctly. When other sbt tasks run after `test` there often isn't enough memory "left", thus `4g`.

aednichols: @kshakir for me `1536` causes the `Initial heap size set to a larger value than the maximum heap size` error and SBT shell does fails to start

aednichols: `2048` works fine though.

docs/Building.md
jmthibault79: no good reason.  removing.

docs/Building.md
jmthibault79: thx

docs/Building.md
jmthibault79: Hmmm.  My interpretation of the error was that it was *necessary* to increase it above 4GB, due to the Xmx setting.  But I can confirm 1600 (1536 + some extra) works also.  We should rewrite this after we determine what is going on and what is actually required.

gemmalam:
@jmthibault79 please merge this when it is ready :)

jmthibault79:
Looked at the issue more carefully and changed the recommendation.  Please re-review.

mcovarr:
:+1:

Issue #3873, opened 2018-07-06
Query endpoint `includeSubworkflows` parameter database error


rexwangcc:
On Cromwell version, "34-5156b78-SNAP", when making a request to the query endpoint that includes the `"includeSubworkflows": "false"` parameter, the database will throw an error if the query matches ~1000 results. 

Example POST request:
```
curl -X POST "https://cromwell.caas-dev.broadinstitute.org/api/workflows/v1/query" -H "accept: application/json" -H "authorization: Bearer XXXXX" -H "Content-Type: application/json" -d "[ { \"status\": \"Failed\", \"includeSubworkflows\": \"false\" }]"
```
Error message:
```
{
  "status": "fail",
  "message": "Task slick.basic.BasicBackend$DatabaseDef$$anon$2@27386688 rejected from slick.util.AsyncExecutor$$anon$2$$anon$1@2e52a800[Running, pool size = 200, active threads = 200, queued tasks = 999, completed tasks = 16375]"
}
```
This error is very similar to the other issue: https://github.com/broadinstitute/cromwell/issues/3115
but `includeSubworkflows` shouldn't limit the number of results the `/query` endpoint can handle.

cjllanwarne:
This problem no longer happens but the related issues #3115 and #4561 may still be problems.

rexwangcc:
Sorry @cjllanwarne , I just tested with:
```
[
  {
    "includeSubworkflows": "false"
  }
]
``` on 
https://cromwell.caas-prod.broadinstitute.org which has version `36-fde91e6`, the problem appeared to me again:
```
{
  "status": "fail",
  "message": "Task slick.basic.BasicBackend$DatabaseDef$$anon$2@23304acb rejected from slick.util.AsyncExecutor$$anon$2$$anon$1@233013e3[Running, pool size = 200, active threads = 200, queued tasks = 1000, completed tasks = 2178366]"
}
```
Maybe the query is too ambitious? I should use pagination or more restrictive query instead?

PR #3903, opened 2018-07-17
Fix 'includeSubworkflow' behavior and change query return order

cjllanwarne:
Choose your poison:

1. "FTFY" the query return order for people, making it "newest first" rather than "oldest first"
2. Add another parameter to the query endpoint
3. Add a config option 

I chose option (1) for this PR, but I'm quite happy to change that.


build.sbt
```diff
@@ -94,6 +94,7 @@ lazy val ossFileSystem = (project in file("filesystems/oss"))
 
 lazy val databaseSql = (project in file("database/sql"))
   .withLibrarySettings("cromwell-database-sql", databaseSqlDependencies)
+  .dependsOn(core)
```
kshakir: Pass in the key(s) you need through the database interface. Do not depend on core. Think of this artifact as if you were already in the MySQL/HSQLDB database and had no access to Cromwell business logic.

database/sql/src/main/scala/cromwell/database/slick/tables/WorkflowMetadataSummaryEntryComponent.scala
```diff
@@ -133,6 +137,13 @@ trait WorkflowMetadataSummaryEntryComponent {
       .reduceLeftOption(op)
   }
 
+  /**
+    * Determines whether a workflow is a subworkflow by checking whether it has metadata for a parent.
+    */
+  private def isASubworkflow(workflowMetadataSummaryEntry: WorkflowMetadataSummaryEntries): Rep[Boolean] = {
+    metadataEntryExistsForWorkflowExecutionUuid(workflowMetadataSummaryEntry.workflowExecutionUuid, WorkflowMetadataKeys.ParentWorkflowId)
```
kshakir: Key should continue to be passed through as a `String` argument here.

database/sql/src/main/scala/cromwell/database/slick/tables/WorkflowMetadataSummaryEntryComponent.scala
```diff
@@ -180,11 +195,13 @@ trait WorkflowMetadataSummaryEntryComponent {
...
       submissionTimestampOption,
       startTimestampOption,
-      endTimestampOption
+      endTimestampOption,
+      includeSubworkflows
     )
     val query = workflowMetadataSummaryEntries.filter(filter)
     (page, pageSize) match {
       case (Some(p), Some(ps)) => query.drop((p - 1) * ps).take(ps)
+      case (None, Some(ps)) => query.take(ps)
```
kshakir: Warning: Scanning through the code there are other cases that expect BOTH params. I would move this logic out of the database layer. At a higher level *outside* of the database there should be business logic that sets the default page size if `page` is selected, and selects the default page if only `pageSize` is selected.

Here is another case that uses both as `Some` (this code is broken anyway and should be deleted):

https://github.com/broadinstitute/cromwell/blob/c70505bc0b8e4cb35a88d1cb89a6917f57f835ca/engine/src/main/scala/cromwell/webservice/metadata/WorkflowQueryPagination.scala#L39

And here's one that triggers just on `page` instead of `pageSize`:

https://github.com/broadinstitute/cromwell/blob/c70505bc0b8e4cb35a88d1cb89a6917f57f835ca/services/src/main/scala/cromwell/services/metadata/impl/MetadataDatabaseAccess.scala#L214

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## 34 Release Notes
 
+### Query API
+
+* Fixes a bug which stopped `includeSubworkflow=false` from paging correctly and subworkflows from being discounted correctly from `totalResultsCount`.
```
kshakir: ToL: Either don't mention paging yet (as it's still unsupported!), or finally euthanize the bad RFC 5988 pagination headers and add pagination to cromwell.yaml, /docs, etc.

geoffjentry: We're going to need to [modify our pagination scheme](https://github.com/ga4gh/workflow-execution-service-schemas/blob/develop/openapi/workflow_execution_service.swagger.yaml#L780) soon anyways, FWIW

build.sbt
cjllanwarne: Necessary to let `WorkflowMetadataSummaryEntryComponent`  see the `WorkflowMetadataKeys`

database/sql/src/main/scala/cromwell/database/slick/tables/WorkflowMetadataSummaryEntryComponent.scala
cjllanwarne: I've removed the bad headers and made `page` and `pageSize` public because it's useful, we rely on it in JMUI, and so we don't want it to accidentally get "nobody uses this"'d.

If it's wrong, that's cause for a ticket, not a "hide it under the rug" IMO and whatever query does end up looking like, it needs to include pagination.

CHANGELOG.md
cjllanwarne: I've removed the bad headers and made page and pageSize public because it's useful, we rely on it in JMUI, and we don't want it to get "nobody uses this"'d.

If it's wrong, that's cause for a ticket, not a "hide it under the rug" IMO and whatever query does end up looking like, it needs to include pagination.

cjllanwarne:
Also requesting a "review" from @rsasch - maybe more from a "does this unblock your JMUI changes" rather than a "does the database scheme look good?" perspective

database/sql/src/main/scala/cromwell/database/slick/tables/WorkflowMetadataSummaryEntryComponent.scala
cjllanwarne: This behavior also matches what WES looks like it would expect

database/sql/src/main/scala/cromwell/database/slick/tables/WorkflowMetadataSummaryEntryComponent.scala
```diff
@@ -96,6 +98,9 @@ trait WorkflowMetadataSummaryEntryComponent {
       val labelsOrFilter = existsWorkflowLabels(workflowMetadataSummaryEntry, labelOrKeyValues, _ || _)
       val excludeLabelsAndFilter = existsWorkflowLabels(workflowMetadataSummaryEntry, excludeLabelAndValues, _ && _).map(v => !v)
       val excludeLabelsOrFilter = existsWorkflowLabels(workflowMetadataSummaryEntry, excludeLabelOrValues, _ || _).map(v => !v)
+      val notASubworkflowFilter: Option[Rep[Boolean]] =
+        if (includeSubworkflows) None else Some(metadataEntryExistsForWorkflowExecutionUuid(workflowMetadataSummaryEntry.workflowExecutionUuid, parentIdWorkflowMetadataKey))
+
```
salonishah11: @cjllanwarne Do you think adding the filtering of subworkflows when `includeSubworkflows = false` here, might resolve the issue: #3873? Or is it unrelated?

cjllanwarne: Great question! @rexwangcc do you still have the database of >1000 jobs to test this version against to see if anything changed, got better, or got worse?

salonishah11: I feel it might make a difference, because before all the workflows were fetched which would match other filters, and then each of them would be checked for subworkflow condition if `includeSubworkflows = false` was set. Now, the subworkflows condition is also checked along with other filters, and if it is a subworkflow it won't be included in the database result itself (when `includeSubworkflows = false`). It might resolve the issue!

database/sql/src/main/scala/cromwell/database/slick/tables/WorkflowMetadataSummaryEntryComponent.scala
kshakir: ToL: There may be others places in the code with biz logic embedded in data access code. I'm not sure at the moment. I still believe this biz logic should be concentrated in _one_ place higher up, not sprinkled throughout DAOs.

engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
```diff
@@ -345,7 +345,7 @@ trait CromwellApiService extends HttpInstrumentation {
 
     onComplete(response) {
       case Success(w: WorkflowQuerySuccess) =>
-        val headers = WorkflowQueryPagination.generateLinkHeaders(uri, w.meta)
+        val headers = List.empty
```
kshakir: Just `complete()`. Don't add empty `headers`

Issue #3744, opened 2018-06-07
Add support for volumes specification


elerch:
Add volumes runtime attribute that matches a restricted Docker volumes specification

tom-dyar:
Hi! Is any of this related to #3746? How might I get started now with storage -- I have some permissions problems making a custom AMI right now. Is there any way to use EFS or some other file system that could work well with command line tools like kraken / bowtie2, etc...?

Thanks for any tips!

delagoya:
Current proposal is to support the [`disks` runtime attribute](http://cromwell.readthedocs.io/en/develop/RuntimeAttributes/#disks) using the following rules:

1. For all tasks, provide a predictable bind mount for `local-disk`. Specifications for disk size and disk type will be ignored, as they are not needed or configurable at runtime for AWS Batch. 
2. Other mount points that are defined (e.g. `disks: "/mnt/my_mnt 3 SSD, /mnt/my_mnt2 500 HDD"`) will result in additional bind mounts from the host container to the running docker container for the task. The disk size and disk type are ignored.

The AWS Batch reference deployment for Cromwell will provide a mount point for each task which the `disks` will be structured under. As an example, assume the following runtime attribute definition:

```
runtime {
  disks: "local-disk 100 SSD, /mnt/my_mnt 3 SSD, /mnt/my_mnt2 500 HDD"
}
```

Will result in a filesystem tree structure on the host:

```
/mnt/cromwell_io_mountpoint/
    ├── $CROMWELL_TASK_ID
              ├── /cromwell_root
              └── /mnt/
                    ├── /my_mnt
                    └── /my_mnt2
```

And the running container will see the `/cromwell_root`, `/mnt/my_mnt` and `/mnt/my_mnt2` directories.

delagoya:
Also worth noting that `/mnt/cromwell_io_mountpoint` is not to be taken literally. 

It should be a configurable option to Cromwell server, but not a runtime attribute. 

Proposed  name "CROMWELL_HOST_ROOT_MOUNTPOINT" but open to better suggestion.

ruchim:
What are thoughts about the cleanup of directory `$CROMWELL_TASK_ID` after the task finishes and the appropriate outputs are staged back to an S3 bucket?

delagoya:
@ruchim that is better discussed in issue #3804

delagoya:
[...]

PR #3914, opened 2018-07-20
Add unique host path for storing task data on AWS Batch instances

elerch:
This creates a host path that should be suitable for the use of multiple containers in #3744.


supportedBackends/aws/src/main/scala/cromwell/backend/impl/aws/AwsBatchJob.scala
```diff
@@ -179,22 +179,23 @@ final case class AwsBatchJob(jobDescriptor: BackendJobDescriptor,           // W
   }
   def submitJob(): Try[SubmitJobResponse] = Try {
     val taskId = jobDescriptor.key.call.fullyQualifiedName + "-" + jobDescriptor.key.index + "-" + jobDescriptor.key.attempt
+    val workflow = jobDescriptor.workflowDescriptor
+    val uniquePath = workflow.callable.name + "/" +
+                     jobDescriptor.taskCall.callable.name + "/" +
+                     workflow.id + "/" +
```
mcovarr: TOL I'd flip the order of these two to keep the workflow and call bits together

PR #3916, opened 2018-07-21
Relatively important

### LINKED ISSUE #3868: Relative WDL imports
- [ ] Implement https://github.com/openwdl/wdl/pull/220 in Cromwell
- [ ] Cycle back to the OpenWDL issue to notify the existence of an implementation

cjllanwarne:
- Keeps us up to date with OpenWDL PR https://github.com/openwdl/wdl/pull/220 (relative imports)
  - [ ] Notify OpenWDL of an existing implementation when this PR merges.
- Closes #3868 
- Since the WDL spec was previously not opinionated on this (and it turned out to be easier to share this with 1.0 than to sabotage the 1.0 version...) this functionality also works in WDL 1.0!
  - This PR doesn't implement relative imports in draft-2. 
  - This PR doesn't implement relative imports in CWL. 
- I'm particularly interested in thoughts on how to test HTTP relative imports in centaur. Right now "raw github URLs to my feature branch" doesn't feel super great 😨.
  - Thought: fake the internet and have a fake `HttpImportResolver` just for the centaur Cromwell?

### DIFF
--- a/centaur/src/main/resources/standardTestCases/wdl_biscayne/biscayne_http_relative_imports/biscayne_http_relative_imports.wdl
+++ b/centaur/src/main/resources/standardTestCases/wdl_biscayne/biscayne_http_relative_imports/biscayne_http_relative_imports.wdl
@@ -1,0 +1,12 @@
+version 1.0
+
+# Is there a better way to test this?
+import "https://raw.githubusercontent.com/broadinstitute/cromwell/cjl_relative_imports/womtool/src/test/resources/validate/biscayne/valid/relative_imports/sub_wfs/foo.wdl"
+
+workflow biscayne_http_relative_imports {
+  call foo.foo_wf
+
+  output {
+    Int result = foo_wf.unpacked
+  }
+}
--- a/centaur/src/main/resources/standardTestCases/wdl_biscayne/biscayne_http_relative_imports/biscayne_http_relative_imports.wdl
+++ b/centaur/src/main/resources/standardTestCases/wdl_biscayne/biscayne_http_relative_imports/biscayne_http_relative_imports.wdl
@@ -1,4 +1,4 @@
-version 1.0
+version biscayne
 
 # Is there a better way to test this?
 import "https://raw.githubusercontent.com/broadinstitute/cromwell/cjl_relative_imports/womtool/src/test/resources/validate/biscayne/valid/relative_imports/sub_wfs/foo.wdl"
--- a/languageFactories/language-factory-core/src/main/scala/cromwell/languages/util/ImportResolver.scala
+++ b/languageFactories/language-factory-core/src/main/scala/cromwell/languages/util/ImportResolver.scala
@@ -12,86 +12,120 @@ import com.softwaremill.sttp.asynchttpclient.cats.AsyncHttpClientCatsBackend
 import common.Checked
 import common.transforms.CheckedAtoB
 import common.validation.ErrorOr._
+import common.validation.Checked._
 import common.validation.Validation._
 import cromwell.core.path.Path
 import wom.core.WorkflowSource
 
 import scala.concurrent.duration._
 import scala.concurrent.Await
-import scala.util.Try
+import scala.util.{Failure, Success, Try}
 
 object ImportResolver {
-  type ImportResolver = CheckedAtoB[String, WorkflowSource]
-
-  def zippedImportsResolver(zippedImports: Array[Byte]): ImportResolver = {
-    directoryResolver(LanguageFactoryUtil.validateImportsDirectory(zippedImports).toEither)
+  trait ImportResolver2 {
+    def name: String
+    protected def innerResolver(str: String): Checked[WorkflowSource]
+    def updateForNewRoot(newRoot: String): ImportResolver2
+    def resolver: CheckedAtoB[String, WorkflowSource] = CheckedAtoB.fromCheck { str =>
+      innerResolver(str).contextualizeErrors(s"resolve '$str' using resolver: '$name'")
+    }
   }
 
-  def directoryResolver(directory: Path, allowEscapingDirectory: Boolean = false): ImportResolver = CheckedAtoB.fromErrorOr { path =>
-    Try(Paths.get(directory.resolve(path).toFile.getCanonicalPath)).toErrorOr flatMap { absolutePathToFile =>
-      val absolutePathToImports = Paths.get(directory.toJava.getCanonicalPath)
-      if (allowEscapingDirectory || absolutePathToFile.startsWith(absolutePathToImports)) {
+  object DirectoryResolver2 {
+    def apply(directory: Path, allowEscapingDirectory: Boolean): DirectoryResolver2 = {
+      val dontEscapeFrom = if (allowEscapingDirectory) None else Option(directory.toJava.getCanonicalPath)
+      DirectoryResolver2(directory, dontEscapeFrom)
+    }
+  }
+
+  case class DirectoryResolver2(directory: Path, dontEscapeFrom: Option[String] = None) extends ImportResolver2 {
+    lazy val absolutePathToDirectory = directory.toJava.getCanonicalPath
+
+    override def innerResolver(path: String): Checked[WorkflowSource] = {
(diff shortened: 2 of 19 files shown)


centaur/src/main/resources/standardTestCases/wdl_biscayne/biscayne_http_relative_imports/biscayne_http_relative_imports.wdl
```diff
@@ -0,0 +1,12 @@
+version 1.0
+
+# Is there a better way to test this?
+import "https://raw.githubusercontent.com/broadinstitute/cromwell/cjl_relative_imports/womtool/src/test/resources/validate/biscayne/valid/relative_imports/sub_wfs/foo.wdl"
```
cjllanwarne: Change to `develop` before merging?

womtool/src/test/resources/validate/biscayne/valid/http_relative_imports/http_relative_imports.wdl
```diff
@@ -0,0 +1,12 @@
+version 1.0
+
+# Is there a better way to test this?
+import "https://raw.githubusercontent.com/broadinstitute/cromwell/cjl_relative_imports/womtool/src/test/resources/validate/biscayne/valid/relative_imports/sub_wfs/foo.wdl"
```
cjllanwarne: Change to `develop` before merging?

kshakir:
> thoughts on how to test HTTP relative imports in centaur

No need to use fakes. Use an http link to a public object in GCS. Ex: https://storage.googleapis.com/gcp-public-data-landsat/index.csv.gz

cjllanwarne:
Also requesting a review by @salonishah11 because I think these resolvers might potentially be useful for downloading content from the `workflowUrl` as well

languageFactories/language-factory-core/src/main/scala/cromwell/languages/util/ImportResolver.scala
```diff
@@ -12,86 +13,136 @@ import com.softwaremill.sttp.asynchttpclient.cats.AsyncHttpClientCatsBackend
...
+        val rootWithoutFilename = newRoot.split('/').init.mkString("", "/", "/")
+        List(
+          HttpResolver(relativeTo = Some(canonicalize(rootWithoutFilename)), headers)
+        )
+      }
 
-    implicit val sttpBackend = AsyncHttpClientCatsBackend[IO]()
+      def canonicalize(str: String): String = {
+        val url = new URL(str)
+        val uri = new URI(url.getProtocol, url.getUserInfo, url.getHost, url.getPort, url.getPath, url.getQuery, url.getRef)
```
kshakir: `val uri = url.toURI`

The URI constructor you're using still goes through the same parser.

https://github.com/openjdk-mirror/jdk/blob/jdk8u/jdk8u/master/src/share/classes/java/net/URI.java#L645-L648
https://github.com/openjdk-mirror/jdk/blob/jdk8u/jdk8u/master/src/share/classes/java/net/URL.java#L126-L127

languageFactories/language-factory-core/src/main/scala/cromwell/languages/util/ImportResolver.scala
```diff
@@ -12,86 +13,136 @@ import com.softwaremill.sttp.asynchttpclient.cats.AsyncHttpClientCatsBackend
...
+        uri.toString
+      }
 
-    val responseIO : IO[Response[String]] =
-      sttp.get(uri"$str").headers(headers).send()
+      def root(str: String): String = {
+        val url = new URL(str)
+        val uri = new URI(url.getProtocol, url.getUserInfo, url.getHost, url.getPort, "", "", "")
```
kshakir: Perhaps add a test to cover this line and make sure this method is doing exactly what you want.

I'm pretty sure those empty strings `""` should each be `null`. As is, the `toString` below returns strings ending with `?#`.

If this is expected behavior then please leave a comment explaining why.

centaur/src/main/resources/standardTestCases/wdl_biscayne/biscayne_http_relative_imports/biscayne_http_relative_imports.wdl
```diff
@@ -0,0 +1,12 @@
+version biscayne
+
+# Is there a better way to test this?
+import "https://raw.githubusercontent.com/broadinstitute/cromwell/cjl_relative_imports/womtool/src/test/resources/validate/biscayne/valid/relative_imports/sub_wfs/foo.wdl"
```
aednichols: You could look into using a mock server to serve a specific URL off of localhost

Example of how we test Rawls pulling from Dockstore without actually calling Dockstore:
https://github.com/broadinstitute/rawls/blob/develop/core/src/test/scala/org/broadinstitute/dsde/rawls/mock/RemoteServicesMockServer.scala#L259

aednichols: Or right in Cromwell for that matter: https://github.com/broadinstitute/cromwell/blob/develop/languageFactories/wdl-draft2/src/test/scala/languages.wdl.draft2/WdlWorkflowHttpImportSpec.scala#L41

languageFactories/language-factory-core/src/main/scala/cromwell/languages/util/ImportResolver.scala
```diff
@@ -12,86 +13,151 @@ import com.softwaremill.sttp.asynchttpclient.cats.AsyncHttpClientCatsBackend
...
+import java.nio.file.{Path => NioPath}
+
 import wom.core.WorkflowSource
 
 import scala.concurrent.duration._
 import scala.concurrent.Await
-import scala.util.Try
+import scala.util.{Failure, Success, Try}
 
 object ImportResolver {
```
aednichols: (optional, nice to have) I feel like this file would benefit from a breakup into trait + http + directory files

The distinctions between sections are not super clear

languageFactories/language-factory-core/src/main/scala/cromwell/languages/util/ImportResolver.scala
```diff
@@ -12,86 +13,151 @@ import com.softwaremill.sttp.asynchttpclient.cats.AsyncHttpClientCatsBackend
...
+        }).contextualizeErrors(s"download $toLookup")
+      }
+    }
+  }
 
-    // temporary situation to get functionality working before
-    // starting in on async-ifying the entire WdlNamespace flow
-    val result: Checked[String] = Await.result(responseIO.unsafeToFuture, 15.seconds).body.leftMap(NonEmptyList(_, List.empty))
+  object HttpResolver {
+    def canonicalize(str: String): String = {
```
aednichols: FWIW, I had to look up what canonical means when applied to filepaths

Perhaps I am the last person on earth who didn't know, or perhaps this could use a brief comment

I like this explanation...
>For example, both ../../here/bar/x and ./test/../../bar/x may refer to the same location, but you can't do a textual comparison on the two paths. However, if you turn them into their canonical representation, they both become ../bar/x, and we see that they actually refer to the same thing.

...because it makes clear how they differ from absolute paths (which would also allow textual comparison)

[Source.](https://stackoverflow.com/a/12100371/818054)

geoffjentry: To be fair, it **is** part of the [Java `File` API](https://docs.oracle.com/javase/7/docs/api/java/io/File.html#getCanonicalFile())

centaur/src/main/resources/standardTestCases/wdl_biscayne/biscayne_http_relative_imports/biscayne_http_relative_imports.wdl
cjllanwarne: I think this will definitely work in the unit tests.

I wonder if centaur can be changed to host its own set of http files in some non case-by-case manner? As we move more towards HTTP based files, I think that'll become more and more useful

PR #3918, opened 2018-07-23
Bug Fix: Response error codes in releaseHold endpoint 

salonishah11:
This fixes the response error codes. It will now return `404 Not Found` for an unrecognized workflow ID and `400 Bad Request` for a malformed or invalid workflow ID.

### DIFF
--- a/engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
+++ b/engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
@@ -250,6 +250,7 @@ trait CromwellApiService extends HttpInstrumentation {
           case Success(WorkflowStoreEngineActor.WorkflowOnHoldToSubmittedFailure(_, e: NotInOnHoldStateException)) => e.errorRequest(StatusCodes.Forbidden)
           case Success(WorkflowStoreEngineActor.WorkflowOnHoldToSubmittedFailure(_, e)) => e.errorRequest(StatusCodes.InternalServerError)
           case Success(r: WorkflowStoreEngineActor.WorkflowOnHoldToSubmittedSuccess) => completeResponse(StatusCodes.OK, toResponse(r.workflowId, WorkflowSubmitted), Seq.empty)
+          case Failure(e: UnrecognizedWorkflowException) => e.failRequest(StatusCodes.NotFound)
           case Failure(e) => e.errorRequest(StatusCodes.InternalServerError)
         }
       }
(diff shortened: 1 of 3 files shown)


engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
```diff
@@ -250,6 +250,8 @@ trait CromwellApiService extends HttpInstrumentation {
           case Success(WorkflowStoreEngineActor.WorkflowOnHoldToSubmittedFailure(_, e: NotInOnHoldStateException)) => e.errorRequest(StatusCodes.Forbidden)
           case Success(WorkflowStoreEngineActor.WorkflowOnHoldToSubmittedFailure(_, e)) => e.errorRequest(StatusCodes.InternalServerError)
           case Success(r: WorkflowStoreEngineActor.WorkflowOnHoldToSubmittedSuccess) => completeResponse(StatusCodes.OK, toResponse(r.workflowId, WorkflowSubmitted), Seq.empty)
+          case Failure(e: UnrecognizedWorkflowException) => e.failRequest(StatusCodes.NotFound)
+          case Failure(e: InvalidWorkflowException) => e.failRequest(StatusCodes.BadRequest)
```
cjllanwarne: ToL: Since these are presumably also relevant in other response handlers, could we wrap these "standard error catchers" into a `PartialFunction` and use the `orElse` operator to include them in all of the relevant handlers?

engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
```diff
@@ -240,19 +240,19 @@ class CromwellApiServiceSpec extends AsyncFlatSpec with ScalatestRouteTest with
...
-  it should "return 500 when invalid workflow id is submitted to onHoldToSubmitted API end point" in {
+  it should "return 404 when invalid workflow id is submitted to onHoldToSubmitted API end point" in {
     val id = UnrecognizedWorkflowId
     Post(s"/workflows/$version/$id/releaseHold") ~>
       akkaHttpService.workflowRoutes ~>
       check {
         assertResult(
           s"""{
-             |  "status": "error",
+             |  "status": "fail",
```
cjllanwarne: ToL: is there a big difference between `error` and `fail`? It seems slightly awkward for users to have to handle two synonymous but different `status` strings?

salonishah11: So AFAIK, fail indicates that the data submitted in the request is incorrect, while error is when some kind of exception is thrown while processing the request.

mcovarr: Ah this means FAIL in the all-caps sense.

PR #3922, opened 2018-07-23
 Disallow array autoboxing

### LINKED ISSUE #3478: Unexpected String -> Array[String] coercion
Cromwell version: 30-9a7de06

Minimized from one of our WDLs:
```wdl
workflow Test {
  Boolean do
  Int n

  if (do) {
    call Optional
  }

  scatter (i in range(n)) {
    call Scattered
  }

  call Gather {
    input:
      # HERE: select_first returns String, and Scattered.out is an Array[String]
      ins = if defined(Optional.out) then select_first([Optional.out]) else Scattered.out
  }
  output {
    Gather.out
  }
}

task Optional {
  command {
    echo "Hey!"
  }
  output {
    String out = read_string(stdout())
  }
}

task Scattered {
  command {
    echo "Hello!"
  }
  output {
    String out = read_string(stdout())
  }
}

task Gather {
  Array[String] ins

  command {
    cat ${write_lines(ins)}
  }
  output {
    String out = read_string(stdout())
  }
}
```

This WDL runs successfully, but in code review I noticed the weird type mismatch between the branches. I asked @cjllanwarne about it and he thought it was an old "feature" that had been purged to avoid bugs / confusion. I'd expect something like this to be rejected.

cjllanwarne:
Closes #3478 

Best explained by the new test case:

```wdl
workflow bad_autobox {
  Int i = 5

  # Bad autobox!
  Array[Int] bad_is = i

  # Good manual-box!
  Array[Int] good_is = [ i ]
}

```

### DIFF
--- a/wom/src/main/scala/wom/types/WomArrayType.scala
+++ b/wom/src/main/scala/wom/types/WomArrayType.scala
@@ -41,11 +41,6 @@ sealed trait WomArrayType extends WomType {
     case womArray: WomArray if (allowEmpty || womArray.nonEmpty) && memberType.isCoerceableFrom(womArray.womType.memberType) =>
       womArray.map(v => memberType.coerceRawValue(v).get) // .get because isCoerceableFrom should make it safe
     case WomArrayLike(womArray) if this.isCoerceableFrom(womArray.womType) => coercion.apply(womArray)
-    case womValue: WomValue if memberType.isCoerceableFrom(womValue.womType) =>
-      memberType.coerceRawValue(womValue) match {
-        case Success(coercedValue) => WomArray(this, Seq(coercedValue))
-        case Failure(ex) => throw ex
-      }
   }
 
   override def typeSpecificIsCoerceableFrom(otherType: WomType): Boolean = otherType match {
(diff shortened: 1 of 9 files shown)


wom/src/main/scala/wom/types/WomArrayType.scala
```diff
@@ -41,11 +41,6 @@ sealed trait WomArrayType extends WomType {
     case womArray: WomArray if (allowEmpty || womArray.nonEmpty) && memberType.isCoerceableFrom(womArray.womType.memberType) =>
       womArray.map(v => memberType.coerceRawValue(v).get) // .get because isCoerceableFrom should make it safe
     case WomArrayLike(womArray) if this.isCoerceableFrom(womArray.womType) => coercion.apply(womArray)
-    case womValue: WomValue if memberType.isCoerceableFrom(womValue.womType) =>
```
aednichols: How is this match still exhaustive with this case removed? The `WomValue` type presumably still exists, so what happens when you get one?

cjllanwarne: It isn't, but that's ok because it's a `PartialFunction[Any, WomValue]`.

In fact it never was complete - the `case womValue: WomValue` wasn't a catch-all because it still had a `if memberType.isCoerceableFrom(womValue.womType)` clause

PR #3924, opened 2018-07-23
Dos Script: Martha url can be passed as environment variable

salonishah11:
This PR will allow Martha's url to be passed as an environment variable during `docker run`. The variable can be set as `--env MARTHA_URL=<url>`.

PR #3930, opened 2018-07-24
Allow for passing Martha URL to DOS localization script.

mcovarr:
Cromwell side of #3924


supportedBackends/google/pipelines/v2alpha1/src/test/resources/reference.conf
```diff
@@ -1,2 +1,2 @@
 demo.dos.localization.docker-image = "somerepo/dos-downloader:tagged"
-demo.dos.localization.command-template = "/path/to/some_executable before args ${dosPath} middle args ${containerPath} ends args"
+demo.dos.localization.command-template = "/path/to/some_executable before args ${marthaUrl} ${dosPath} middle args ${containerPath} ends args"
```
salonishah11: I think the command template should have `--env MARTHA_URL=<url>`. So, it can be something like `--env MARTHA_URL=${marthaUrl} /path/to/some_executable before args ${dosPath} middle args ${containerPath} ends args`

geoffjentry:
👍

mcovarr:
This shall return.

PR #3933, opened 2018-07-25
Basic HTTP input support

mcovarr:
Needs docs if this is not just a checkpoint.


backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -99,7 +99,7 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
 
   override lazy val ioActor = standardParams.ioActor
 
-  /** Backend initialization data created by the a factory initializer. */
+  /** Backend initialization data created by the factory initializer. */
```
geoffjentry: primo work

supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystem.scala
```diff
@@ -198,8 +205,25 @@ trait SharedFileSystem extends PathFactory {
...
+      pathBuilders.collectFirst({ case http: HttpPathBuilder if HttpPathBuilder.accepts(input) => http }) match {
+        case Some(http) =>
+          implicit val materializer = ActorMaterializer()
+          implicit val ec = actorContext.dispatcher
+          val stagedFile = better.files.File.newTemporaryFile()
+          val futurePath = http.contentAsByteIterator(input) map { bytes =>
+            stagedFile.writeBytes(bytes)
+            stagedFile.pathAsString
+          }
+          Await.result(futurePath, Duration.Inf)
```
geoffjentry: Should I be giving the side eye here or is there a reason it's AOK? (including "good enough for now!"

mcovarr: The other network path builders also block on localization, they just hide that better behind their NIO interfaces. 🙂

geoffjentry: 🙈

core/src/main/resources/reference.conf
```diff
@@ -249,6 +249,9 @@ filesystems {
   s3 {
     class = "cromwell.filesystems.s3.S3PathBuilderFactory"
   }
+  http {
+    class = "cromwell.filesystems.http.HttpPathBuilderFactory"
```
cjllanwarne: Given that this is in `reference.conf` do we want a way to turn this off that doesn't equal "sabotage with an invalid class path"?

supportedBackends/google/pipelines/common/src/main/scala/cromwell/backend/google/pipelines/common/PipelinesApiAsyncBackendJobExecutionActor.scala
```diff
@@ -658,20 +659,20 @@ class PipelinesApiAsyncBackendJobExecutionActor(override val standardParams: Sta
...
-    womFile.mapFile(value =>
+    womFile.mapFile { value =>
       (getPath(value), asAdHocFile(womFile)) match {
         case (Success(gcsPath: GcsPath), Some(adHocFile)) =>
           // Ad hoc files will be placed directly at the root ("/cromwell_root/ad_hoc_file.txt") unlike other input files
           // for which the full path is being propagated ("/cromwell_root/path/to/input_file.txt")
           workingDisk.mountPoint.resolve(adHocFile.alternativeName.getOrElse(gcsPath.name)).pathAsString
-        case (Success(gcsPath: GcsPath), _) =>
-          workingDisk.mountPoint.resolve(gcsPath.pathWithoutScheme).pathAsString
+        case (Success(path @ ( _: GcsPath | _: HttpPath)), _) =>
```
cjllanwarne: Is this pattern suggesting that `GcsPath` and `HttpPath` should inherit some shared `trait`?

Otherwise we're going to have to remember to append to all of these cases whenever a new FS like this is added (S3Path...?)

supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystem.scala
```diff
@@ -198,8 +205,25 @@ trait SharedFileSystem extends PathFactory {
       PairOfFiles(src, dest)
     }
 
+    // A possibly staged version of the input file suitable for downstream processing, or just the original input
+    // file if no staging was required.
+    val staged: WomFile = value.mapFile { input =>
+      pathBuilders.collectFirst({ case http: HttpPathBuilder if HttpPathBuilder.accepts(input) => http }) match {
```
cjllanwarne: Could we multi-line this `case`?

supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystem.scala
```diff
@@ -198,8 +205,25 @@ trait SharedFileSystem extends PathFactory {
...
+        case Some(http) =>
```
cjllanwarne: ToL for clarity, can `http` => `httpResolver`?

supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystem.scala
cjllanwarne: In fact, this doesn't seem particularly specific to `Http` - if *any* path builder is willing to localize this path for us, shouldn't we let it?

supportedBackends/google/pipelines/common/src/main/scala/cromwell/backend/google/pipelines/common/PipelinesApiAsyncBackendJobExecutionActor.scala
cjllanwarne: (or `DosPath`...)

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
mcovarr: thx

centaur/src/main/resources/standardTestCases/http_inputs/http_inputs.wdl
```diff
@@ -0,0 +1,16 @@
...
+  }
+  output {
+    String sum = read_string(stdout())
+  }
+  runtime {
+    docker: "ubuntu:latest"
+  }
+}
+
+workflow http_inputs {
```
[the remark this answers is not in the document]
mcovarr: sure

core/src/main/resources/reference.conf
mcovarr: Yeah that's a good point.

filesystems/demo-dos/src/main/scala/cromwell/filesystems/demo/dos/DemoDosPathBuilder.scala
```diff
@@ -24,7 +24,7 @@ class DemoDosPathBuilder extends PathBuilder {
   override def name: String = "Demo Dos"
 
   override def build(pathAsString: String): Try[Path] = {
-    if (pathAsString.startsWith("dos://")) {
+    if (DemoDosPathBuilder.accepts(pathAsString)) {
```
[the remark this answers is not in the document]
mcovarr: I'm not really up on the long term plans for DOS so I don't really know.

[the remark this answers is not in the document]
geoffjentry: The scope hasn't changed

supportedBackends/google/pipelines/common/src/main/scala/cromwell/backend/google/pipelines/common/PipelinesApiAsyncBackendJobExecutionActor.scala
geoffjentry: Or FTP in the very near future (@Horneth )

filesystems/http/src/main/scala/cromwell/filesystems/http/HttpPathBuilder.scala
```diff
@@ -0,0 +1,46 @@
...
+      response <- Http().singleRequest(HttpRequest(uri = input))
+      byteString <- response.entity.dataBytes.runFold(ByteString(""))(_ ++ _)
+    } yield byteString.iterator
+  }
+}
+
+case class HttpPath(nioPath: NioPath, httpPathBuilder: HttpPathBuilder) extends Path {
+  override protected def newPath(nioPath: NioPath): Path = httpPathBuilder.build(pathAsString).get
+
+  override def pathAsString: String = nioPath.toString.replaceFirst("/", "//")
```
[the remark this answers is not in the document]
mcovarr: Discussed in person.

supportedBackends/google/pipelines/common/src/main/scala/cromwell/backend/google/pipelines/common/PipelinesApiAsyncBackendJobExecutionActor.scala
mcovarr: If N &gt; 2 I would agree. *Demo*DosPath below is currently doing something a little different and I'm not sure what's brewing for FTP.

supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystem.scala
mcovarr: Right now it's a method on two of the path builder companion objects. I tried making it an instance method but IIRC the complexity of the rules on what's acceptable in GCS paths made me rethink that.

supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystem.scala
mcovarr: sure

supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystem.scala
mcovarr: https://github.com/broadinstitute/cromwell/blob/develop/core/src/main/scala/cromwell/core/path/PathBuilderFactory.scala#L15

mcovarr: Oops this reply probably makes more sense attached to your comment below. 🙂

core/src/main/resources/reference.conf
mcovarr: Also discussed in person, this is defining a filesystem named `http` but isn't configuring the default `Local` backend to use it. So using the HTTP filesystem would be opt-in for the default Local backend which is what we want.

Horneth: Adding a FS here makes it "available for referencing" but they still need to be explicitly referenced it in the engine or backend `filesystems` stanza to really be usable

supportedBackends/google/pipelines/common/src/main/scala/cromwell/backend/google/pipelines/common/PipelinesApiAsyncBackendJobExecutionActor.scala
Horneth: Hmm are we talking about something like `trait PathSupportedByPipelinesAPI` and have GcsPath etc extend it ?
I'm not totally convinced that would make things easier especially if filesystems impl and their respective `Path` are in separate isolated subprojects.

core/src/main/resources/reference.conf
mcovarr: Per discussion yesterday we decided the http filesystem should be on by default in the default `Local` backend but that it should also be possible to turn it off. I've added code that allows setting `-Dbackend.providers.Local.config.filesystems.http.enabled=false` to turn the http filesystem off in `Local`.

cjllanwarne:
@mcovarr does "needs docs if not a checkpoint" still apply?

filesystems/http/src/main/scala/cromwell/filesystems/http/HttpPathBuilder.scala
```diff
@@ -0,0 +1,46 @@
...
+    }
+  }
+
+  def contentAsByteIterator(input: String)(implicit actorContext: ActorContext, actorMaterializer: ActorMaterializer): Future[ByteIterator] = {
+    implicit val actorSystem: ActorSystem = actorContext.system
+    implicit val executionContext: ExecutionContext = actorContext.dispatcher
+    for {
+      response <- Http().singleRequest(HttpRequest(uri = input))
+      byteString <- response.entity.dataBytes.runFold(ByteString(""))(_ ++ _)
+    } yield byteString.iterator
```
cjllanwarne: Just to double check, I may be wrong but this looks like it downloads the entire file before creating an iterator out of it?

ie if someone does `read_string(oops.bam)` will we download the whole thing to memory, produce an iterator for it, and then hit the "file too big" only while iterating through the in-memory iterator?

supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystem.scala
```diff
@@ -198,8 +205,25 @@ trait SharedFileSystem extends PathFactory {
...
+    // file if no staging was required.
+    val staged: WomFile = value.mapFile { input =>
+      pathBuilders.collectFirst({ case h: HttpPathBuilder if HttpPathBuilder.accepts(input) => h }) match {
+        case Some(httpPathBuilder) =>
+          implicit val materializer = ActorMaterializer()
+          implicit val ec = actorContext.dispatcher
+          val stagedFile = better.files.File.newTemporaryFile()
+          val futurePath = httpPathBuilder.contentAsByteIterator(input) map { bytes =>
+            stagedFile.writeBytes(bytes)
+            stagedFile.pathAsString
```
cjllanwarne: ToL: it smells a little fishy to me that it's the Cromwell server itself rather than the target system that's doing the downloading.

eg if 100,000 jobs are submitted we will open 100,000 connections and stream 100,000 files before adding jobs to the SGE queue (for example). If the backend was doing it as part of the job (like in the PAPI2 case), we could leverage the existing queuing and breadth of computational resource to take care of all that

geoffjentry: It's a good point, OTOH submitting 100,000 SGE jobs is also not going to go so well on most systems

cjllanwarne: Maybe I'm just especially sensitive right now to HTTP-related things that happen "in-JVM" if we don't know how well they scale...

Issue #3804, opened 2018-06-21
Add support for staging INPUTS/OUTPUTS from/to AWS S3


delagoya:
Functionality is needed to stage data files to/from S3 on a per-task level.

For the example WDL file:
```
task copy_file {
  String output_file
  File input_file

  command {
    cp  ${input_file} ${output_file}
  }
  runtime {
    docker: "ubuntu:latest"
  }
}

workflow wf_copy_file {
  call copy_file
}
```
and corresponding `inputs.json`

```json
{
  "wf_cop_file.copy_file.input_file": "s3://myBucket/hello.txt",
  "wf_cop_file.copy_file.output_file": "greetings.txt"
}
```

The workflow execution should be able to copy the input file from S3 to the working task directory, and copy the output file "greetings.txt" to the configured S3 bucket for writing logs and outputs. An example of files written to the output S3 bucket would be:

```
# $WF_ID is the workflow identifier (e.g. "E6D5143C-89BC-4823-AED7-2A6AE00A1C2B")
s3://cromwell-output-bucket/$WF_ID/copy_file/outputs/greetings.txt
s3://cromwell-output-bucket/$WF_ID/copy_file/wf_copy_file-rc.txt
s3://cromwell-output-bucket/$WF_ID/copy_file/wf_copy_file-stdout.txt
s3://cromwell-output-bucket/$WF_ID/copy_file/wf_copy_file-stderr.txt
```

danbills:
@delagoya what about the privileged docker that's running other docker containers?  I.e. the phases of localization - command - delocalization.  Would be nice to spell that out in more detail

danbills:
Just brain-dumping my understanding of the approach to get the convo started:

The idea as I understood it was to run an orchestrating docker with elevated privileges, which I'll call Orchestrator in a fit of creativity.

Orchestrator would kick off N number of localization commands to localize files onto the host.
  (N could be 1)  

Orchestrator runs exec.sh with files existing on host.

Orchestrator runs N de-localization commands.

--

## Idle Thoughts

I don't immediately recall why this had to be separate docker commands and not just scripts running before and after the main script.  

This is how GCP Papi v2 works, by chaining dockers together (I think some concept of fork() is supported too)

delagoya:
@danbills The Orchestrator pattern as described above is what we discussed. 

[...]

PR #3934, opened 2018-07-26
Cromwell agent

elerch:
This implements the cromwell agent (process that monitors docker for Cromwell tasks) as part of #3804. I will include the sidecar process later as a separate PR. After building the agent and running the agent with the build-agent/run-agent scripts, a "cromwell-agent" container will be shown in docker ps. This agent will be reading the docker event stream from /events, filter for starts with grep, and pass any start events to the small go program "spawnoninput".

This program simply spawns a new process when it sees input. In this case, the process is a shell script that pauses the container, checks to see if it matches the target condition, then either resumes/exits (when not meeting the target condition) or leaves the container paused and launches a sidecar, which is still in development.

If the target condition does not match, this has an effect of slowing startup for the launched container of approximately 100ms for the inspection process. This image should ideally be under the Broad's account on Dockerhub or ECR - it's based on alpine Linux and is relatively small. It should also be run automatically in the AWS Batch/Cromwell AMI.

More documentation will be forthcoming in a later PR, though some inline comments are in the code here.

### DIFF
--- a/supportedBackends/aws/src/main/resources/agent/launch_sidecar
+++ b/supportedBackends/aws/src/main/resources/agent/launch_sidecar
@@ -1,0 +1,103 @@
+#!/bin/sh
+# Copyright 2018 Amazon.com, Inc. or its affiliates.
+#
+#  Redistribution and use in source and binary forms, with or without
+#  modification, are permitted provided that the following conditions are met:
+#
+#  1. Redistributions of source code must retain the above copyright notice,
+#  this list of conditions and the following disclaimer.
+#
+#  2. Redistributions in binary form must reproduce the above copyright
+#  notice, this list of conditions and the following disclaimer in the
+#  documentation and/or other materials provided with the distribution.
+#
+#  3. Neither the name of the copyright holder nor the names of its
+#  contributors may be used to endorse or promote products derived from
+#  this software without specific prior written permission.
+#
+#  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+#  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING,
+#  BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
+#  FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+#  THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT,
+#  INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+#  (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+#  SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+#  HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+#  STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING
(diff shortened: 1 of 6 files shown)


supportedBackends/aws/src/main/resources/agent/launch_sidecar
```diff
@@ -0,0 +1,92 @@
...
+docker pause $dockerid
+# Docker pause can happen based on the "create" event from docker
+#
+# The infrastructure around this seems to take between 10 and 20ms in my tests.
+# This is likely due to the processing time needed for jq to parse the json.
+#
+# It is possible, though unlikely, that some dependent work starts happening
+# within that time. If so, we may need a sleep 0.02 in the script preamble
+# to accomodate this scenario and/or find something more performant than jq.
+# Because of this, we are using cut above rather than jq, which should be slightly faster
```
Horneth: Have you measured how multiple jobs being scheduled in parallel on the same host would affect the sidecar performance ?

supportedBackends/aws/src/main/resources/agent/launch_sidecar
```diff
@@ -0,0 +1,92 @@
...
+containerInfo=$(docker inspect $dockerid)
+
+if [ -z $(environmentVariable "AWS_CRMWLL_PROCESS_MONITOR_MARKER") ]; then
```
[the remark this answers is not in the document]
elerch: No - the other variables are named CROMWELL - this one is special so it doesn't get bundled in with the rest, creating a recursive monitoring explosion of activity.

supportedBackends/aws/src/main/resources/agent/launch_sidecar
elerch: The biggest risk on sidecar performance is in regards to the limitations of inotify, especially wrt large directory trees with lots of small files. But then, that's not an ideal situation for s3 either. I haven't done much measurement of these limitations, but I don't expect number of containers to be an issue as much as the aggregate number of files involved.

Horneth: Oh I see, when I said sidecar I actually meant the agent. But if all it does is spawning a new process it should not be too much of a bottleneck.
As for the actual sidecar, how will it know what files to localize and delocalize ? For delocalization you seem to imply it will watch the filesystem and push everything to S3 ?

supportedBackends/aws/src/main/resources/agent/launch_sidecar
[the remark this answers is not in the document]
Horneth: Do you mean that something is looking for `CROMWELL` in the name of env variables ?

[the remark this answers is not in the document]
elerch: Yes - I reserve the AWS_CROMWELL prefix for variables that I need to pass through to enable this process. The variables are used for the sidecar to properly localize/delocalize (Call root, workflow root, rc file - stuff like that).

supportedBackends/aws/src/main/resources/agent/launch_sidecar
elerch: The environment variables I pass through to the task (the AWS_CROMWELL stuff) have the necessary information the sidecar needs to localize/delocalize. I'll have a new PR for that shortly.

supportedBackends/aws/src/main/resources/agent/launch_sidecar
```diff
@@ -0,0 +1,92 @@
...
+  echo $containerInfo |jq -r "$query" | cut -d= -f2
+}
+
+# $1 is passed in by spawnoninput, which originally will come from a single
+# line from the docker event stream (endpoint /events)
+line=$1
+
+# These commands are for "start", "create", "attach"
+# While jq is functionally preferred, cut should be slightly faster. We're
+# somewhat against the clock here as the container is going to do some work
```
mcovarr: IMHO it would be preferable to have have proper coordination between the task and localization rather than creating a race.

gemmalam:
Hi Emil,
Reopening this ticket to connect with the associated PR. Once the PR is merged, this issue will automatically close.

PR #3935, opened 2018-07-27
[32] Fix performance issues with concurrent HTTP import validation

aednichols:
I.e. don't kill Rawls during FireCloud workshops.

( This is all @cjllanwarne, I just made the branch. )

### DIFF
--- a/languageFactories/language-factory-core/src/main/scala/cromwell/languages/util/ImportResolver.scala
+++ b/languageFactories/language-factory-core/src/main/scala/cromwell/languages/util/ImportResolver.scala
@@ -91,6 +91,8 @@ object ImportResolver {
     // starting in on async-ifying the entire WdlNamespace flow
     val result: Checked[String] = Await.result(responseIO.unsafeToFuture, 15.seconds).body.leftMap(NonEmptyList(_, List.empty))
 
+    sttpBackend.close()
+
     result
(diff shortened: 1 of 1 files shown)


cjllanwarne:
I feel like I have to recuse myself from this PR since it was my change. So "red thumbs required" please

languageFactories/language-factory-core/src/main/scala/cromwell/languages/util/ImportResolver.scala
```diff
@@ -94,4 +94,8 @@ object ImportResolver {
     result
   }
 
+  object HttpResolver {
+    lazy val sttpBackend = AsyncHttpClientCatsBackend[IO]()
```
Horneth: ToL: This could be implicit already + `import HttpResolver._` and the `implicit val sttpBackend = HttpResolver.sttpBackend` wouldn't be needed (I think) but 🤷‍♂️

cjllanwarne: If it's implicit in the object then anyone importing HttpResolver would get an implicit `AsyncHttpClientCatsBackend` in their scope.

Probably not a problem, but it felt nicer to force people to opt-in in this case

PR #3938, opened 2018-07-27
Even More Important

aednichols:
~Same code change as the hotfixes.~

The code change from the hotfixes, plus a shutdown hook to cleanly close the HTTP backend as suggested by their docs

### DIFF
--- a/common/src/main/scala/common/util/IntrospectableLazy.scala
+++ b/common/src/main/scala/common/util/IntrospectableLazy.scala
@@ -1,0 +1,27 @@
+package common.util
+
+import scala.language.implicitConversions
+
+// Adapted https://stackoverflow.com/a/17057490/818054
+// Not thread-safe!
+
+object IntrospectableLazy {
+
+  def lazily[A](f: => A): IntrospectableLazy[A] = new IntrospectableLazy(f)
+
+  implicit def evalLazy[A](l: IntrospectableLazy[A]): A = l.create()
+
+}
+
+class IntrospectableLazy[A] private(f: => A) {
+
+  private var option: Option[A] = None
+
+  def create(): A = option match {
+    case Some(a) => a
+    case None => val a = f; option = Some(a); a
+  }
+
+  def exists: Boolean = option.isDefined
+
+}
--- a/common/src/main/scala/common/util/IntrospectableLazy.scala
+++ b/common/src/main/scala/common/util/IntrospectableLazy.scala
@@ -17,11 +17,13 @@ class IntrospectableLazy[A] private(f: => A) {
   private var option: Option[A] = None
 
   def apply(): A = {
-    this.synchronized {
-      option match {
-        case Some(a) => a
-        case None => val a = f; option = Some(a); a
-      }
+    option match {
+      case Some(a) => a
+      case None =>
+        synchronized { option match {
+          case Some(a) => a
+          case None => val a = f; option = Some(a); a
+        }}
     }
   }
 
--- a/common/src/main/scala/common/util/IntrospectableLazy.scala
+++ b/common/src/main/scala/common/util/IntrospectableLazy.scala
@@ -2,14 +2,13 @@ package common.util
 
 import scala.language.implicitConversions
 
-// Adapted https://stackoverflow.com/a/17057490/818054
-// Not thread-safe!
+// Adapted from https://stackoverflow.com/a/17057490/818054
 
 object IntrospectableLazy {
 
   def lazily[A](f: => A): IntrospectableLazy[A] = new IntrospectableLazy(f)
 
-  implicit def evalLazy[A](l: IntrospectableLazy[A]): A = l.create()
+  implicit def evalLazy[A](l: IntrospectableLazy[A]): A = l()
 
 }
 
@@ -17,9 +16,13 @@ class IntrospectableLazy[A] private(f: => A) {
 
   private var option: Option[A] = None
 
-  def create(): A = option match {
-    case Some(a) => a
-    case None => val a = f; option = Some(a); a
+  def apply(): A = {
+    this.synchronized {
+      option match {
+        case Some(a) => a
+        case None => val a = f; option = Some(a); a
+      }
+    }
   }
 
   def exists: Boolean = option.isDefined
--- a/common/src/main/scala/common/util/IntrospectableLazy.scala
+++ b/common/src/main/scala/common/util/IntrospectableLazy.scala
@@ -2,6 +2,15 @@ package common.util
 
 import scala.language.implicitConversions
 
+// Designed to hold resources that may or may not be initialized during the lifetime
+// of the application, much like lazy.
+//
+// The difference is that e.g. a shutdown routine can check whether an IntrospectableLazy
+// actually exists() and skip cleaning it up if it doesn't.
+//
+// Otherwise, the routine may actually be the first reference to the resource, causing
+// it to be initialized immediately before cleaning it up.
+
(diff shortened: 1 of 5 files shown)


geoffjentry:
:+1:

This appears to fit the model propposed by [the docs](http://sttp.readthedocs.io/en/latest/backends/start_stop.html). Stated in person, but if not a giant pain in the butt it'd be nice to adhere to their suggestion to call `close()`.

cjllanwarne:
We should probably do the `close()` when the `womtool` CLI exits too, if possible?

aednichols:
Good idea @cjllanwarne, committed.

womtool/src/main/scala/womtool/WomtoolMain.scala
```diff
@@ -139,6 +140,9 @@ object WomtoolMain extends App {
   val termination = runWomtool(args)
   termination.stdout foreach Console.out.println
   termination.stderr foreach Console.err.println
+
+  HttpResolver.sttpBackend.close()
```
cjllanwarne: ToL: Sorry for turning a throwaway comment into a much bigger refactor, but if `sttpBackend` was not started naturally during processing then this line will have to create it (which will involve starting up an entire actor system) and then immediately shut it down again.

If that's not a negligible process, then this could be an annoying extra overhead on womtool processing time.

aednichols: That does sound quite silly.

It would be nice to somehow capture the startup / shutdown steps into `HttpResolver` itself so that its users don't have to remember (i.e. occasionally forget) to close the backend.

Is there a reasonable factoring for the above? I know destructors and the like are very much not a thing on the JVM

aednichols: This is a thing that looks like it could work: https://docs.oracle.com/javase/7/docs/api/java/lang/AutoCloseable.html

cjllanwarne: I believe that that `AutoCloseable` is part of the try-with-resources syntax?

aednichols: Yeah, this would not be trivial

cjllanwarne: I wonder if there's a nice scala way to check whether a `lazy val` has been instantiated yet to let us short-circuit the closing if it hasn't been instantiated yet

cjllanwarne: Some interesting thoughts on https://stackoverflow.com/questions/17057255/how-to-check-if-a-lazy-val-is-initialized-without-initializing-it

aednichols: Was reading this too, love it:
>This isn't really an answer to your question, and I hate when people do this, but I'm going to do it anyway. I think the best response is: A lazy val isn't appropriate for this, so define a type that does support what you need.

engine/src/main/scala/cromwell/server/CromwellShutdown.scala
```diff
@@ -217,5 +218,13 @@ object CromwellShutdown extends GracefulStopSupport {
       logger.info("Stream materializer shut down")
       Future.successful(Done)
     }
+
+    // 7) Close out the backend used for WDL HTTP import resolution
+    // http://sttp.readthedocs.io/en/latest/backends/start_stop.html
+    coordinatedShutdown.addTask(CoordinatedShutdown.PhaseBeforeActorSystemTerminate, "wdlHttpImportResolverBackend") { () =>
+      HttpResolver.sttpBackend.close()
```
Horneth: @aednichols looks good to me :)
Although the same comment @cjllanwarne made below applies here too of course.

common/src/main/scala/common/util/IntrospectableLazy.scala
```diff
@@ -0,0 +1,30 @@
...
+  implicit def evalLazy[A](l: IntrospectableLazy[A]): A = l()
+
+}
+
+class IntrospectableLazy[A] private(f: => A) {
+
+  private var option: Option[A] = None
+
+  def apply(): A = {
+    this.synchronized {
```
cjllanwarne: I do think it'll be nice to optimize for the (by far most common) common case by putting this `synchronized` only around the `None` case (and re-check whether it's still `None` inside the `synchronized`)

common/src/test/scala/common/util/IntrospectableLazySpec.scala
```diff
@@ -0,0 +1,32 @@
...
+      lazyInstantiations += 1
+      4
+    }
+
+    val myLazy = lazily { lazyContents }
+
+    assert(lazyInstantiations == 0)
+    assert(!myLazy.exists)
+
+    assert(myLazy() == 4)
```
cjllanwarne: Maybe wrap this in something concurrent? eg
```scala
100 times {
  Future {
    assert(myLazy() == 4)
  }
}
```

common/src/main/scala/common/util/IntrospectableLazy.scala
```diff
@@ -0,0 +1,32 @@
...
+
+object IntrospectableLazy {
+
+  def lazily[A](f: => A): IntrospectableLazy[A] = new IntrospectableLazy(f)
+
+  implicit def evalLazy[A](l: IntrospectableLazy[A]): A = l()
+
+}
+
+class IntrospectableLazy[A] private(f: => A) {
```
cjllanwarne: Maybe add some docs here on:

 - When you would anticipate this being used
 - An examples of using it
 - Why you had to do it like this (when it looks like you just way went out of your way just to reimplement `lazy val`)

engine/src/main/scala/cromwell/server/CromwellShutdown.scala
```diff
@@ -217,5 +218,13 @@ object CromwellShutdown extends GracefulStopSupport {
       logger.info("Stream materializer shut down")
       Future.successful(Done)
     }
+
+    // 7) Close out the backend used for WDL HTTP import resolution
+    // http://sttp.readthedocs.io/en/latest/backends/start_stop.html
+    coordinatedShutdown.addTask(CoordinatedShutdown.PhaseBeforeActorSystemTerminate, "wdlHttpImportResolverBackend") { () =>
+      HttpResolver.closeBackendIfNecessary()
+      logger.info("WDL HTTP import resolver closed")
+      Future.successful(Done)
```
cjllanwarne: Is the fact that this wants a `Future` back indicating that this should really be non-blocking, ie:
```scala
Future {
  HttpResolver.closeBackendIfNecessary()
  logger.info("WDL HTTP import resolver closed")
}
```

cjllanwarne:
@geoffjentry do you want to claim re-thumbing rights this given that your approval was prior to the `IntrospectableLazy`ness? Or happy to let your previous thumb be grandfathered in?

cjllanwarne:
(or @Horneth or @mcovarr as also prior-thumbers)

geoffjentry:
@cjllanwarne provided you're fully satisfied that it's safe and is well tested I'm good

engine/src/main/scala/cromwell/server/CromwellShutdown.scala
aednichols: The shutdown tasks seem to bucket into "takes a while" which is implemented like your example and "pretty fast" which do their work synchronously.

If closing the backend is a tossup, I will conservatively move it to async.

PR #3948, opened 2018-07-31
Initial sidecar implementation

### LINKED ISSUE #3744: Add support for volumes specification

elerch:
These changes implement the sidecar and together with #3934 close #3744 and #3804

### DIFF
--- a/supportedBackends/aws/src/main/resources/sidecar/sidecar
+++ b/supportedBackends/aws/src/main/resources/sidecar/sidecar
@@ -1,0 +1,117 @@
+#!/bin/sh
+# Copyright 2018 Amazon.com, Inc. or its affiliates.
+#
+#  Redistribution and use in source and binary forms, with or without
+#  modification, are permitted provided that the following conditions are met:
+#
+#  1. Redistributions of source code must retain the above copyright notice,
+#  this list of conditions and the following disclaimer.
+#
+#  2. Redistributions in binary form must reproduce the above copyright
+#  notice, this list of conditions and the following disclaimer in the
+#  documentation and/or other materials provided with the distribution.
+#
+#  3. Neither the name of the copyright holder nor the names of its
+#  contributors may be used to endorse or promote products derived from
+#  this software without specific prior written permission.
+#
+#  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+#  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING,
+#  BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
+#  FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL
+#  THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT,
+#  INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+#  (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+#  SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
+#  HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+#  STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING
+#  IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+#  POSSIBILITY OF SUCH DAMAGE.
+
+# We expect the following environment variables:
+#
+# AWS_CROMWELL_TASK_DOCKER_ID
+# AWS_CROMWELL_RC_FILE
+# AWS_CROMWELL_PATH
+# AWS_CROMWELL_LOCAL_DISK
+# AWS_CROMWELL_CALL_ROOT
+
+waitfor() {
+  echo Watching "$1". Will return when "$2" is found
+  inotifywait -r -e close_write,moved_to,moved_from,delete -m "$1" | while read -r path action file; do
+    if [ "$file" = "$2" ]; then
+      echo Target file found - killing inotifywait and returning
+      killall inotifywait # inotifywait will simply hang out with a return like this
+      return
+    fi
+    localDiskRelativePath=${path#"${AWS_CROMWELL_LOCAL_DISK}/"}
+    if [ "$action" = "MOVED_TO" ] || [ "$action" = "CLOSE_WRITE,CLOSE" ]; then
+      case $path in
+        ${AWS_CROMWELL_LOCAL_DISK}/${exclude}*)
+          echo skipping change in excluded directory
+          ;;
+        *)
+          aws s3 cp "${path}${file}" "$AWS_CROMWELL_CALL_ROOT/${localDiskRelativePath}${file}"
+      esac
+    elif [ "$action" = "MOVED_FROM" ]; then
+      case $path in
+        ${AWS_CROMWELL_LOCAL_DISK}/${exclude}*)
+          echo skipping change in excluded directory
+          ;;
+        *)
+          # The file was removed from the local disk, so we'll remove it
+          # from s3 as well.
+          aws s3 rm "$AWS_CROMWELL_CALL_ROOT/${localDiskRelativePath}${file}"
+      esac
+    fi
+  done
+}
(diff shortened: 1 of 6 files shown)


supportedBackends/aws/src/main/resources/sidecar/sidecar
```diff
@@ -0,0 +1,117 @@
+#!/bin/sh
```
geoffjentry: If you haven't already, we generally prefer that all shell scripts are run through [shellcheck](https://www.shellcheck.net/) and all warnings corrected

supportedBackends/aws/src/main/resources/sidecar/sidecar
```diff
@@ -0,0 +1,117 @@
...
+# localRoot example: /cromwell_root/cromwell-aws/cromwell-execution...
+# AWS_CROMWELL_PATH example: ...
+localCallRoot="$AWS_CROMWELL_LOCAL_DISK"/"${AWS_CROMWELL_CALL_ROOT#"s3://"}"
+localWorkflowRoot="$AWS_CROMWELL_LOCAL_DISK"/"${AWS_CROMWELL_WORKFLOW_ROOT#"s3://"}"
+suffix="/*"
+exclude=${AWS_CROMWELL_CALL_ROOT#"s3://"}
+exclude=${exclude%%$suffix}
+echo "syncing files from s3"
+echo "source ${AWS_CROMWELL_WORKFLOW_ROOT}, dest $localWorkflowRoot"
+aws s3 sync "$AWS_CROMWELL_WORKFLOW_ROOT" "$localWorkflowRoot"
```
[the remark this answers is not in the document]
Horneth: > Is this also localizing inputs to this call that are the outputs from upstream calls?

Hmm if that's true the performance and storage costs of localization for large workflows are going to be very heavy IMO

Also, is this assuming that all inputs to the call will be in the s3 workflow root ?
Because even though a lot of time this will be true, it won't always be, specifically the inputs to the workflow itself will be on the user's bucket or even some public bucket, wherever the input data is.

supportedBackends/aws/src/main/resources/sidecar/sidecar
```diff
@@ -0,0 +1,117 @@
...
+waitfor "$AWS_CROMWELL_LOCAL_DISK" "${AWS_CROMWELL_RC_FILE#"${AWS_CROMWELL_LOCAL_DISK}/"}"
+
+# We'll pause the task container while we do our work. This will prevent
+# the container from exiting before files are out on S3. Also, the ecs agent
+# has some cleanup, and we don't want a race condition with that process
+docker pause "$AWS_CROMWELL_TASK_DOCKER_ID"
+
+###########################################################################
+# Now we will do the final sync. If we've done the job right, this sync
+# should only copy the rc file in place to s3, since all the other files
```
Horneth: Why do we need to re-sync everything if we know that the only thing left is the rc file ?
In what circumstance could we have not "done the job right" ?

supportedBackends/aws/src/main/resources/sidecar/sidecar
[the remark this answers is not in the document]
elerch: Is there a centaur test that shows this outside the s3 workflow root scenario @Horneth? I was assuming I'd need to pass some external input requests as metadata into the container but I didn't see any examples.

supportedBackends/aws/src/main/resources/sidecar/sidecar
elerch: Yup - I run with shellcheck. There was a missed warning from a last minute change that I've since cleaned up.

supportedBackends/aws/src/main/resources/sidecar/sidecar
elerch: In my testing, all files are synced properly by the time this call is run, so it's effectively a no-op, with the exception of the rc file which I'm currently not copying prior to the exit from waitfor(). There is, however, a possibility of missed files due to issues with inotifywait that are documented in the man page: https://linux.die.net/man/1/inotifywait

supportedBackends/aws/src/main/resources/sidecar/sidecar
```diff
@@ -0,0 +1,117 @@
...
+# We expect the following environment variables:
+#
+# AWS_CROMWELL_TASK_DOCKER_ID
+# AWS_CROMWELL_RC_FILE
+# AWS_CROMWELL_PATH
+# AWS_CROMWELL_LOCAL_DISK
+# AWS_CROMWELL_CALL_ROOT
+
+waitfor() {
+  echo Watching "$1". Will return when "$2" is found
```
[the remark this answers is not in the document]
elerch: The script epilogue default set by StandardAsyncExecutionActor is a sync call: https://github.com/broadinstitute/cromwell/blob/519708d9cafd4c4886d089f3cdf545879d81812c/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala#L119

IIf a user overrides the default epilogue I believe it would be up to them to make sure all files are quiesced prior to exit.

supportedBackends/aws/src/main/resources/sidecar/sidecar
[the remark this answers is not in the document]
Horneth: I found [this one](https://github.com/broadinstitute/cromwell/blob/develop/centaur/src/main/resources/standardTestCases/missing_sub_inputs/missing_sub_inputs.inputs) for GCS/PAPI for example.
But more generally workflow inputs *will be* outside the workflow root for any real workflow, and a job should be able to take inputs from any bucket (modulo permissions of course).
When you say "pass some external input requests as metadata", can you elaborate ? Could we do it for all job inputs instead of just the ones that are outside the workflow root, and avoid having to localize the whole thing ?

supportedBackends/aws/src/main/resources/sidecar/sidecar
Horneth: To take the problem the other way around then, why not _only_ do a sync of everything at the end instead of copying files individually with `inotifywait` if it has issues ? Just trying to see if the redundancy is really needed :)

supportedBackends/aws/src/main/resources/sidecar/sidecar
```diff
@@ -0,0 +1,117 @@
...
+docker pause "$AWS_CROMWELL_TASK_DOCKER_ID"
+
+###########################################################################
+# Now we will do the final sync. If we've done the job right, this sync
+# should only copy the rc file in place to s3, since all the other files
+# should have been caught during the task run with our inotifywatch magic(tm)
+###########################################################################
+echo "syncing files with s3"
+echo aws s3 sync "$AWS_CROMWELL_LOCAL_DISK" "$AWS_CROMWELL_CALL_ROOT" --exclude "${exclude}/*"
+aws s3 sync "$AWS_CROMWELL_LOCAL_DISK" "$AWS_CROMWELL_CALL_ROOT" --exclude "${exclude}/*"
```
Horneth: Could you use this exclude variable to also exclude everything that has not been declared a job output by the author ? This would prevent from delocalizing a potentially large number of unwanted temporary files to S3.

supportedBackends/aws/src/main/resources/sidecar/sidecar
elerch: That was the original design, but I was told I should be doing a sync-as-you-go. I'm happy to pull out the inotify stuff and just to the pre and post sync steps though if that's all that's needed. It would have the advantage of not syncing temp files.

elerch: No - the files being processed by the current running task.

delocalizer:
:clap: (please forgive random excited cheering on the sidelines from a user super-keen to have a working AWS backend)

supportedBackends/aws/src/main/resources/sidecar/sidecar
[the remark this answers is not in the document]
ruchim: @elerch  the HaplotypeCaller test also has external inputs -- Khalid posted the inputs file with those examples in the aws cromwell slack.

supportedBackends/aws/src/main/resources/sidecar/sidecar
elerch: That's correct - I didn't want to sync to S3 during file writes, so I'm doing it after close. The theory being that in a days-long workflow, files written would be available even if the full processed dataset is not.

supportedBackends/aws/src/main/resources/sidecar/sidecar
elerch: As the design sits today I don't have that information in the sidecar, so I can't exclude things not explicitly declared. I'm going to take a look today and see if I can get more specific with both inputs and outputs.

elerch:
Closing this PR and will open a new one shortly. Moving from sidecar to proxy implementation.

PR #3952, opened 2018-08-01
Go back to MemorySize

kshakir:


mcovarr:
here's the thumb from the other PR :+1:

kshakir:
Got a verbal second thumb. Merging.

PR #3959, opened 2018-08-02
correct typo in Slurm configuration

rhpvorderman:
The number of cpus per task should be designated with the `-c` flag. The `-n` flag is for the number of tasks. Due to the nature of cromwell the number of tasks is always 1. So we need to adjust the number of cpus by setting `--cpus-per-task=` or `-c `.


ruchim:
:+1:  Great catch!

cjllanwarne:
Thanks @rhpvorderman! This will officially become part of Cromwell in version 35!

PR #3963, opened 2018-08-02
Do not build a wes2cromwell docker.

kshakir:


cjllanwarne:
I think merging should probably be gated on @geoffjentry's thumb, even though there are already two approvals

geoffjentry:
👍

PR #3965, opened 2018-08-02
Reduce hash cost of WdlSyntaxErrorFormatter

aednichols:
I don't have numbers for this yet, but it seems to run way faster.

### DIFF
--- a/wdl/model/draft2/src/main/scala/wdl/draft2/model/WdlSyntaxErrorFormatter.scala
+++ b/wdl/model/draft2/src/main/scala/wdl/draft2/model/WdlSyntaxErrorFormatter.scala
@@ -13,6 +13,14 @@ case class WdlSyntaxErrorFormatter(terminalMap: Map[Terminal, WorkflowSource]) e
   // To approximate, just use Java's case-class-ignorant implementation (hash functions are allowed to be imperfect)
   override def hashCode(): Int = System.identityHashCode(this)
 
+  override def equals(other: Any): Boolean = {
+    other match {
+      case otherFormatter: WdlSyntaxErrorFormatter =>
+        this.terminalMap == otherFormatter.terminalMap
+      case _ => false
+    }
+  }
+
   private def pointToSource(t: Terminal): String = s"${line(t)}\n${" " * (t.getColumn - 1)}^"
   private def getTerminal(t: Terminal) = t match {
     case interpolated: InterpolatedTerminal => terminalMap.get(interpolated.rootTerminal)
--- a/wdl/model/draft2/src/main/scala/wdl/draft2/model/WdlSyntaxErrorFormatter.scala
+++ b/wdl/model/draft2/src/main/scala/wdl/draft2/model/WdlSyntaxErrorFormatter.scala
@@ -13,6 +13,8 @@ case class WdlSyntaxErrorFormatter(terminalMap: Map[Terminal, WorkflowSource]) e
   // To approximate, just use Java's case-class-ignorant implementation (hash functions are allowed to be imperfect)
   override def hashCode(): Int = System.identityHashCode(this)
 
+  override def equals(obj: Any): Boolean = this eq obj.asInstanceOf[AnyRef]
+
   private def pointToSource(t: Terminal): String = s"${line(t)}\n${" " * (t.getColumn - 1)}^"
   private def getTerminal(t: Terminal) = t match {
     case interpolated: InterpolatedTerminal => terminalMap.get(interpolated.rootTerminal)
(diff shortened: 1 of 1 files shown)


wdl/model/draft2/src/main/scala/wdl/draft2/model/WdlSyntaxErrorFormatter.scala
```diff
@@ -9,6 +9,10 @@ import scala.collection.JavaConverters._
 
 case class WdlSyntaxErrorFormatter(terminalMap: Map[Terminal, WorkflowSource]) extends SyntaxErrorFormatter {
 
+  // Hashing the terminalMap is very expensive because we re-hash the entire workflow once for every terminal in it (exponential with size of workflow)
+  // To approximate, let's just say that two workflows with the same number of terminals are likely identical (hash functions are allowed to be imperfect)
+  override def hashCode(): Int = terminalMap.size
```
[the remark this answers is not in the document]
aednichols: Computing the hash even once is very expensive with the current implementation.

If you `womtool validate` Chet's WDL, my branch takes about 5 seconds, `develop` about 25.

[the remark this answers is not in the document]
mcovarr: Reference equality would be at least as cheap, though now I'm wondering why this thing is even involved in a hashCode computation at all.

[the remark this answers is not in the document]
aednichols: Case classes `WdlNamespaceWithWorkflow` and `WdlWorkflow` both have a `WdlSyntaxErrorFormatter` as a member, and it looks like they are being added to hash-based data structures further up the stack.

wdl/model/draft2/src/main/scala/wdl/draft2/model/WdlSyntaxErrorFormatter.scala
```diff
@@ -9,6 +9,10 @@ import scala.collection.JavaConverters._
 
 case class WdlSyntaxErrorFormatter(terminalMap: Map[Terminal, WorkflowSource]) extends SyntaxErrorFormatter {
 
+  // Hashing the terminalMap is very expensive because we re-hash the entire workflow once for every terminal in it (exponential with size of workflow)
+  // To approximate, just use Java's case-class-ignorant implementation (hash functions are allowed to be imperfect)
+  override def hashCode(): Int = System.identityHashCode(this)
```
cjllanwarne: Remember to also override `equals` when overriding `hashCode` and vice versa

aednichols: I'm not sure how to implement this subject to the constraint that [equal objects must have equal hashcodes](https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#equals(java.lang.Object)) - now that the hashcode no longer considers the case class fields

cjllanwarne: I think object equality in scala is just ‘x eq y’

aednichols: Ah, OK - have put that in.

My understanding is that hashcode is supposed to be computed based on a subset of the state used for equals - which is now true here, because both use identity.

However, I feel like `equals` is now wrong / less useful because a deep comparison of two WDL case class hierarchies may now unexpectedly fail. I'm wondering if a `hashCode` consistent with the default `equals` wouldn't serve us better.

wdl/model/draft2/src/main/scala/wdl/draft2/model/WdlSyntaxErrorFormatter.scala
```diff
@@ -9,6 +9,12 @@ import scala.collection.JavaConverters._
...
+
+  override def equals(obj: Any): Boolean = this eq obj.asInstanceOf[AnyRef]
```
cjllanwarne: Was the `.asInstanceOf[AnyRef]` necessary? It seems a tiny bit dangerous in case it gets passed a `null` (I can't see *why* it ever would, but just in case?)

cjllanwarne:
👍

wdl/model/draft2/src/main/scala/wdl/draft2/model/WdlSyntaxErrorFormatter.scala
aednichols: Necessary because one cannot compare `Any` to `AnyRef`; `null.asInstanceOf[AnyRef]` seems to just politely return `null`

PR #3980, opened 2018-08-06
WDL Directory

cjllanwarne:
Prototype to inform conversation on WDL directories in OpenWDL (https://github.com/openwdl/wdl/pull/241)

TODO:
- [x] Add a test to ensure I didn't accidentally add this to WDL 1.0 (where `Directory` would be treated like a `struct` name)

Limits:
- Pipelines API v1 cannot handle sub-directories.


cjllanwarne:
@antonkulaga no, this PR would merge the changes into `develop`, but they aren't there yet.

To try this out you'd need to check out this branch of the source code and then build from source (`sbt assembly`, to get the appropriate Cromwell `jar` file generated in `target`)

PR #3982, opened 2018-08-06
[34] Disable broken /stats endpoint.

mcovarr:


engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
```diff
@@ -60,11 +59,7 @@ trait CromwellApiService extends HttpInstrumentation {
   val engineRoutes = concat(
     path("engine" / Segment / "stats") { _ =>
       get {
-        onComplete(workflowManagerActor.ask(WorkflowManagerActor.EngineStatsCommand).mapTo[EngineStatsActor.EngineStats]) {
-          case Success(stats) => complete(ToResponseMarshallable(stats))
-          case Failure(e: TimeoutException) => e.failRequest(StatusCodes.ServiceUnavailable)
-          case Failure(_) => new RuntimeException("Unable to gather engine stats").failRequest(StatusCodes.InternalServerError)
-        }
+        completeResponse(StatusCodes.Forbidden, APIResponse.fail(new RuntimeException("The /stats endpoint is currently disabled.")), warnings = Seq.empty)
```
cjllanwarne: Do you really need the overhead of creating an instantly-unpacked exception here? (ie is there not just a "use this message" version of the method?)

Also, it's not really a `RuntimeException`...

engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
```diff
@@ -36,7 +36,7 @@ class CromwellApiServiceSpec extends AsyncFlatSpec with ScalatestRouteTest with
 
   implicit def default = RouteTestTimeout(5.seconds)
 
-  "REST ENGINE /stats endpoint" should "return 200 for stats" in {
+  "REST ENGINE /stats endpoint" should "return 200 for stats" ignore {
```
danbills: could delete instead?

ruchim:
👍

engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
ruchim: do we ever want to bring this back?

danbills: mm yeah, we do.  nm

PR #3985, opened 2018-08-06
Disable broken /stats endpoint.

mcovarr:


engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
```diff
@@ -60,11 +59,7 @@ trait CromwellApiService extends HttpInstrumentation {
   val engineRoutes = concat(
     path("engine" / Segment / "stats") { _ =>
       get {
-        onComplete(workflowManagerActor.ask(WorkflowManagerActor.EngineStatsCommand).mapTo[EngineStatsActor.EngineStats]) {
-          case Success(stats) => complete(ToResponseMarshallable(stats))
-          case Failure(e: TimeoutException) => e.failRequest(StatusCodes.ServiceUnavailable)
-          case Failure(_) => new RuntimeException("Unable to gather engine stats").failRequest(StatusCodes.InternalServerError)
-        }
+        completeResponse(StatusCodes.Forbidden, APIResponse.fail(new RuntimeException("The /stats endpoint is currently disabled.")), warnings = Seq.empty)
```
cjllanwarne: Do you really need the overhead of creating an instantly-unpacked exception here? (ie is there not just a "use this message" version of the method?)

Also, it's not really a RuntimeException...

mcovarr: This method works fine with a String but then returns `text/plain` rather than `application/json`. If you have specific suggestions for what this should be instead I'd be open to trying them.

engine/src/test/scala/cromwell/webservice/CromwellApiServiceSpec.scala
```diff
@@ -36,7 +36,7 @@ class CromwellApiServiceSpec extends AsyncFlatSpec with ScalatestRouteTest with
 
   implicit def default = RouteTestTimeout(5.seconds)
 
-  "REST ENGINE /stats endpoint" should "return 200 for stats" in {
+  "REST ENGINE /stats endpoint" should "return 200 for stats" ignore {
```
kshakir: IMHO the test should be updated to match the new expectation, and not leave this cruft behind. That is unless `/engine/v1/stats` is coming back, in which case there should be a link to the ticket with acceptance criteria.

That said, maybe a future PR can clean this up.

mcovarr:
Per discussion with PO merging this as is and possibly remove altogether for the release after next.

PR #3987, opened 2018-08-07
Better import failure messages and importLocalFilesystem resolver set

### LINKED ISSUE #3986: imported files in develop version expected to be at system root
Imports of local files seem to be broken in the development version of cromwell. The paths to the imported wdl files seem to be expected to be absolute paths. Any relative path is treated as relative to the root directory of the system. 

For example:
In the following workflow, test.wdl is imported. If test.wdl is located in the same directory as this workflow it can't be found (`/home/<various directories>/test/test.wdl`), however, if test.wdl is located in the root directory it will be found (`/test.wdl`).
```
version 1.0

import "test.wdl" as test

workflow test2 {
    call test.sayHello as blah {
        input:
            name="Grog"
    }

    output {
        String out = blah.blah
    }
}
```
test.wdl looks like this:
```
version 1.0

task sayHello {
    input {
        String name
    }

    command {
        echo Hello, ~{name}
    }

    output {
        String blah = read_string(stdout())
    }
}
```
The following is mentioned in the printed output:
```
Failed to import 'test.wdl' (reason 1 of 2): Failed to resolve 'test.wdl' using resolver: 'relative to directory / (without escaping None)' (reason 1 of 1): Import file not found: test.wdl
Failed to import 'test.wdl' (reason 2 of 2): Failed to resolve 'test.wdl' using resolver: 'http importer' (reason 1 of 1): Cannot import 'test.wdl' relative to nothing
```
Looking at the cromwell source code I suspect the problem lies with the directory path being given to `DirectoryResolver` in `localFilesystemResolvers` ([this line](https://github.com/broadinstitute/cromwell/blob/develop/engine/src/main/scala/cromwell/engine/workflow/lifecycle/materialization/MaterializeWorkflowDescriptorActor.scala#L271)).

cjllanwarne:
Closes #3986


womtool/src/main/scala/womtool/input/WomGraphMaker.scala
```diff
@@ -26,16 +26,8 @@ object WomGraphMaker {
...
-        DefaultPathBuilder.build(Paths.get(".")),
-        allowEscapingDirectory = false),
-      DirectoryResolver(
-        DefaultPathBuilder.build(Paths.get(mainFile.toAbsolutePath.toFile.getParent)),
-        allowEscapingDirectory = true
-      ),
-      HttpResolver()
-    )
+    lazy val importResolvers: List[ImportResolver] =
+      DirectoryResolver.localFilesystemResolvers(Some(mainFile)) :+ HttpResolver(relativeTo = None)
```
aednichols: Comment seems to be out of date - we don't resolve relative to where we run from anymore

I think the code is pretty self explanatory, so it could just go away

languageFactories/language-factory-core/src/main/scala/cromwell/languages/util/ImportResolver.scala
```diff
@@ -122,7 +157,7 @@ object ImportResolver {
         else canonicalize(s"${relativeToValue.stripSuffix("/")}/$str").validNelCheck
       case None =>
         if (str.startsWith("http")) canonicalize(str).validNelCheck
-        else s"Cannot import '$str' relative to nothing".invalidNelCheck
+        else "Relative path".invalidNelCheck
```
aednichols: Is this a complete thought?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/materialization/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -268,7 +263,9 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
       errorOrParse(factory).flatMap(_.validateNamespace(sourceFiles, workflowSource, workflowOptions, importLocalFilesystem, workflowIdForLogging, engineIoFunctions, importResolvers))
     }
 
-    val localFilesystemResolvers = if (importLocalFilesystem) List(DirectoryResolver(DefaultPath(Paths.get("/")))) else List.empty
+    val localFilesystemResolvers =
+      if (importLocalFilesystem) DirectoryResolver.localFilesystemResolvers(None) :+ HttpResolver(relativeTo = None)
```
aednichols: Why would `importLocalFilesystem` determine whether we include the `HttpResolver(relativeTo = None)`? Should HTTP be on both branches

languageFactories/language-factory-core/src/main/scala/cromwell/languages/util/ImportResolver.scala
cjllanwarne: It ends up looking ok (see the test errors). But I agree, it might be a candidate for improving at some point in the future

womtool/src/main/scala/womtool/input/WomGraphMaker.scala
cjllanwarne: Changed

aednichols:
Deliberately leaving branch since it's not mine in the first place!

PR #3988, opened 2018-08-08
Workflow url in run mode

### LINKED ISSUE #3849: Be able to take a workflow by URL, not workflow descriptor
The cromwell API currently only allows users to submit the actual workflow file when they'd like to run a workflow. Also allow for a user to submit the URL pointing to the workflow file, which will then be consumed by Cromwell. 

Edit: While the end goal is to not need a zip bundle for dependencies (as they can now be worked out via relative paths - cc @cjllanwarne ) -- that work is part of [3868](https://github.com/broadinstitute/cromwell/issues/3868).

NB: TBD (@ruchim ?) what should we do with `file://` URLs? Perhaps a config option defaulting to off?

### LINKED ISSUE #3986: imported files in develop version expected to be at system root

salonishah11:
This PR adds the remaining functionality of submitting workflow url to run a workflow in `Run` mode.

Closes #3849 
Closes #3986


server/src/main/scala/cromwell/CommandLineArguments.scala
```diff
@@ -60,26 +60,34 @@ case class CommandLineArguments(command: Option[Command] = None,
...
       }
-    } else readOptionContent("Workflow source", workflowSource).map((_, imports.map(p => File(p.pathAsString)), workflowRoot))
+    }
 
-    val workflowSourceFinal: ErrorOr[String] = (workflowSource, workflowUrl) match {
-      case (Some(path), None) => readContent("Workflow source", path)
-      case (None, Some(url)) =>  PartialWorkflowSources.convertStringToUrl(url)
-      case (Some(_), Some(_)) => "Both Workflow source and Workflow url can't be supplied".invalidNel
-      case (None, None) => "Workflow source and Workflow url needs to be supplied".invalidNel
+    def getWorkflowSourceFromPath(workflowPath: Path): ErrorOr[(Option[String], Option[String])] = {
```
mcovarr: TOL This 2-tuple of optional Strings would probably be better as a proper type, something like

```
case class SourceOrUrl(source: Option[String], url: Option[String])
```

but there are other options for representing this as well.

server/src/main/scala/cromwell/CromwellEntryPoint.scala
```diff
@@ -175,17 +175,24 @@ object CromwellEntryPoint extends GracefulStopSupport {
...
+          if (w.isDefined) (w,u)
+          else if (u.get.startsWith("http")) (w, u)
+          else (Option(DefaultPathBuilder.get(u.get).contentAsString), None) //case where url is a WDL file
+        }
+
         WorkflowSingleSubmission(
-          workflowSource = w,
-          workflowUrl = u,
+          workflowSource = finalWorkflowSourceAndUrl._1,
+          workflowUrl = finalWorkflowSourceAndUrl._2,
```
mcovarr: if you do the `case class` thing above you can unapply:

```
val SourceOrUrl(source, url) = {
 if (w.isDefined) (w, u)
.
.
.
 }
.
.
.
WorkflowSingleSubmission(
  workflowSource = source,
  workflowUrl = url,
.
.
.
)
```

docs/CommandLine.md
```diff
@@ -40,7 +41,7 @@ The Cromwell jar file can be built as described in [Building](Building).
 `run` mode executes a single workflow in Cromwell and then exits.
 
 * **`workflow-source`**  
-The single required argument for the workflow source file.
+The single required argument. It can either be a workflow source file or a url pointing to the workflow (The maximum length of url can be 2000 characters).
```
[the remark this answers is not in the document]
salonishah11: Makes sense. Will remove it 👍

engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
```diff
@@ -239,11 +234,17 @@ object PartialWorkflowSources {
...
+  def validateWorkflowUrl(workflowUrl: String): ErrorOr[WorkflowUrl] = {
+    def convertStringToUrl(workflowUrl: String): ErrorOr[WorkflowUrl] = {
+      Try(new URL(workflowUrl)) match {
+        case Success(_) => workflowUrl.validNel
+        case Failure(e) => s"Error while validating workflow url: ${e.getMessage}".invalidNel
+      }
+    }
+
+    workflowUrl.length match {
+      case l if l > 2000 => s"Invalid workflow url. Length of the url:$l is more than 2000 characters".invalidNel
```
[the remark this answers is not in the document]
salonishah11: Good point!

server/src/main/scala/cromwell/CommandLineArguments.scala
```diff
@@ -60,26 +60,34 @@ case class CommandLineArguments(command: Option[Command] = None,
...
+      val workflowPath = File(workflowSourcePath.pathAsString)
 
-    val workflowAndDependencies: ErrorOr[(String, Option[File], Option[String])] = if (isCwl) {
       logger.info("Pre Processing Workflow...")
       lazy val preProcessedCwl = cwlPreProcessor.preProcessCwlFileToString(workflowPath, None)
 
       imports match {
-        case Some(explicitImports) => readOptionContent("Workflow source", workflowSource).map((_, Option(File(explicitImports.pathAsString)), workflowRoot))
-        case None => Try(preProcessedCwl.map((_, None, None)).value.unsafeRunSync())
+        case Some(_) => readContent("Workflow source", workflowSourcePath)
```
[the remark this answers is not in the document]
salonishah11: I have tried to keep this CWL part of logic same as it was before workflow url was introduced. I thought it was better to keep it the same way as before.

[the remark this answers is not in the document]
salonishah11: Talked with @Horneth, and it does make sense to send CWL file path as workflow url if imports are present (in Run mode). Will change it. :)

docs/CommandLine.md
```diff
@@ -40,7 +41,7 @@ The Cromwell jar file can be built as described in [Building](Building).
 `run` mode executes a single workflow in Cromwell and then exits.
 
 * **`workflow-source`**  
-The single required argument for the workflow source file.
+The single required argument. It can either be a workflow source file or a url pointing to the workflow.
```
cjllanwarne: I would say "it can be either a local path or a remote URL pointing to the workflow source file"

engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
```diff
@@ -239,11 +236,17 @@ object PartialWorkflowSources {
...
+  def validateWorkflowUrl(workflowUrl: String): ErrorOr[WorkflowUrl] = {
+    def convertStringToUrl(workflowUrl: String): ErrorOr[WorkflowUrl] = {
+      Try(new URL(workflowUrl)) match {
+        case Success(_) => workflowUrl.validNel
+        case Failure(e) => s"Error while validating workflow url: ${e.getMessage}".invalidNel
+      }
+    }
+
+    workflowUrl.length match {
+      case l if l > MaxWorkflowUrlLength => s"Invalid workflow url: url has length $l, longer than the maximum allowed $MaxWorkflowUrlLength characters".invalidNel
```
cjllanwarne: Minor point but this whole `match`/`case` could be restructured as a simpler `if (l > MaxWorkflowUrlLength) {} else {}`

server/src/main/scala/cromwell/CromwellEntryPoint.scala
```diff
@@ -175,17 +176,24 @@ object CromwellEntryPoint extends GracefulStopSupport {
     import spray.json._
 
     val validation = args.validateSubmission(EntryPointLogger) map {
-      case ValidSubmission(w, u, r, i, o, l, z) =>
+      case ValidSubmission(w, u, r, i, o, l, z) =>{
```
cjllanwarne: The `{` and `}` are redundant on a `case` (but maybe you want them to make things easier to read?)

server/src/test/scala/cromwell/CromwellCommandLineSpec.scala
```diff
@@ -47,37 +47,114 @@ class CromwellCommandLineSpec extends FlatSpec with Matchers with BeforeAndAfter
...
+    optionsFirst.workflowSource.get shouldBe threeStep.wdl
+    optionsFirst.workflowInputs.get.pathAsString shouldBe threeStep.inputs
+
+    val validation = Try(CromwellEntryPoint.validateRunArguments(optionsFirst))
+    validation.isSuccess shouldBe true
+    validation.get.workflowSource shouldBe None
+    validation.get.workflowUrl shouldBe Some(threeStep.wdl)
+  }
+
+  it should "run single when supplying workflow using url" in {
```
cjllanwarne: what does "run single" mean?

engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
salonishah11: It sure can! 👍

server/src/main/scala/cromwell/CromwellEntryPoint.scala
salonishah11: Yeah I like to add braces if it make things easier to read. But, here it does not matter as there is just one case! So, will remove it.

server/src/test/scala/cromwell/CromwellCommandLineSpec.scala
salonishah11: Run single workflow/submission. The test is meant for `Run` mode.

PR #3992, opened 2018-08-09
Pin cwltool version

cjllanwarne:
Should unblock our CI but may not be necessary if the short-term fix in https://github.com/common-workflow-language/cwltool/pull/865 or the longer-term update in https://github.com/common-workflow-language/cwltool/pull/864 get released soon.

On the other hand, as Peter says re ruamel, having open ended dependencies is a bad idea generally, so maybe pinning to a known good version is a good idea regardless?


kshakir:
This PR was squashed into #4000

PR #3994, opened 2018-08-10
Conditional predicates don't need to be logged.

mcovarr:


cjllanwarne:
This feels to my like "as a developer, I don't care whether an `if` is executed in somebody's workflow". 
- What about admins and users who look at logs to work out what happened in a workflow?
  - When I was trying to work out what was happening in a workflow with conditionals, I *had* to add these logs to work out what on earth was going on.
- Do we know that the logs are really a cause of problems vs all the other things that have to happen when a 50,000-wide scatter happens and processes conditionals in parallel?
  - Maybe hiding this log just loses visibility that lots and lots of things are happening that we are no longer seeing?

In other words...
- Maybe server logs shouldn't contain individual workflow information *at all*? If we don't log these, why do we log command lines, inputs, tasks completing, etc? Shouldn't that *all* be in metadata?
- Maybe we should sort of run mode so that it can log workflow information to console and *not* server logs...

cjllanwarne:
Tl;DR: I think "our one-size-fits-all" logging is the real problem here. 

Hiding one message to help one user type (and I'm suspicious that this is more signal than cause) might cause other problems for other user types.

geoffjentry:
I think one needs to be mindful of the current logging situation however when they’re adding logging messages.

Horneth:
What if we logged it in the workflow log (file) but not in the Cromwell server log ?

PR #3996, opened 2018-08-10
Catch other message type when cache copy fails

cjllanwarne:


aednichols:
Merging on the basis that tests pass, this is the same code as https://github.com/broadinstitute/cromwell/pull/3997, and https://github.com/broadinstitute/cromwell/pull/3997 is approved

PR #3997, opened 2018-08-10
[34] Catch other message type when cache copy fails

cjllanwarne:


engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/job/EngineJobExecutionActor.scala
```diff
@@ -217,6 +217,10 @@ class EngineJobExecutionActor(replyTo: ActorRef,
     case Event(CachedOutputLookupFailed(_, error), data: ResponsePendingData) =>
       log.warning("Can't make a copy of the cached job outputs for {} due to {}. Running job.", jobTag, error)
       runJob(data)
+    case Event(JobFailedNonRetryableResponse(_, error, _), data: ResponsePendingData) =>
```
danbills: Whoa.  What happens today when this message is dropped?

danbills: nm I can't read diffs

aednichols:
Closing on the basis that https://github.com/broadinstitute/cromwell/pull/3996 is on dev and no one seems to think this is important enough to hotfix - in fact, we forgot all about it for a while

PR #4000, opened 2018-08-13
Retry centaur tests after logging to Sentry.

kshakir:
Sending error info to sentry during centaur testing before retrying.
Encrypting sensitive variables using a random key during centaur tests, jic they are sent to sentry.
Rendering secure resources during _all_ tests.
When secure variables cannot be rendered, only fail when secure variables are required, otherwise producing only info/warning messages.
Disabled caches during tests that read `backendStatus` call metadata.
Allow `test_cromwell.sh` to use a centaur config file.
Enable GcsPathBuilderFactory to retry more than zero times.
Lazy load centaur `*.inputs` & `*.options` so that they aren't required to load `*.test` files.
Relatedly, so that one doesn't (try to) accidentally commit the changes, `git rm` the options file that was being rendered.
Moved logback.xml out of transitive core library and into executables, next to application.conf files.
Pin `cwltool` version.
Use a workaround to pass `--timeout` through `run_test.sh` to `cwltest`.
Using `better.files` instead of `java.nio.Path`, and passing `IO` monads further up.


src/ci/bin/test.inc.sh
```diff
@@ -110,12 +113,13 @@ cromwell::private::create_build_variables() {
         *)
             CROMWELL_BUILD_IS_CI=false
             CROMWELL_BUILD_IS_CRON="${CROMWELL_BUILD_IS_CRON-false}"
-            CROMWELL_BUILD_IS_SECURE=false
+            CROMWELL_BUILD_IS_SECURE=true
```
mcovarr: 👁

src/ci/resources/cwl_conformance_test.wdl
```diff
@@ -63,10 +63,13 @@ task run_test_index {
         Int timeout
     }
 
+    # Weird -n/--timout format is because ./run_test.sh doesn't pass through the timeout parameter to cwltest.
```
mcovarr: --timeout

src/ci/resources/cwl_conformance_test.wdl
```diff
@@ -63,10 +63,13 @@ task run_test_index {
...
+    # Test 55 often runs over 10 minutes in Travis/PapiV2 and requires a longer timeout.
+    # So, we use the fact that run_test.sh conveniently doesn't sanitize -n to wire the --timeout _inside_ the -n arg.
```
mcovarr: 🤕

centaur/src/it/scala/centaur/AbstractCentaurTestCaseSpec.scala
```diff
@@ -92,31 +102,68 @@ abstract class AbstractCentaurTestCaseSpec(cromwellBackends: List[String]) exten
...
+    * @param retries Total number of attempts to retry.
+    * @param attempt Current zero based attempt.
+    * @return IO effect that will run the test, possibly retrying.
+    */
+  private def tryTryAgain(testName: String, runTest: => IO[Unit], retries: Int, attempt: Int = 0): IO[Unit] = {
+    def maybeRetry(centaurTestException: CentaurTestException): IO[Unit] = {
+      val testEnvironment = TestEnvironment(testName, retries, attempt)
+      for {
+        _ <- ErrorReporters.logCentaurFailure(testEnvironment, centaurTestException)
+        _ <- if (attempt < retries) {
```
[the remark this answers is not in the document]
kshakir: `attempt`/`retryAttempt` is zero based in all places in this PR for easier arithmetic. Perhaps `attempt` should be renamed to `retryAttempt` here too?

centaur/src/it/scala/centaur/reporting/AggregatedIo.scala
```diff
@@ -0,0 +1,56 @@
...
+    val ioListThrowables: IO[List[ThrowableNelOr[A]]] =
+      listIo.traverse[IO, ThrowableNelOr[A]](_.attempt.map(_.toValidatedNel))
+    val ioThrowablesList: IO[ThrowableNelOr[List[A]]] = ioListThrowables.map(_.sequence[ThrowableNelOr, A])
+    val ioEither: IO[Either[Throwable, List[A]]] = ioThrowablesList map {
+      case Valid(list) => Either.right(list)
+      case Invalid(nel) if nel.tail.isEmpty => Either.left(nel.head)
+      case Invalid(nel) => Either.left(new AggregatedException(multipleExceptionContext, nel.toList))
+    }
+    val ioList: IO[List[A]] = ioEither.flatMap(IO.fromEither)
+    ioList
```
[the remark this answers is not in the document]
kshakir: Sorta, when my eyes read through this FP-algebra I didn't like that I had to jump back up to the method signature to determine the return type of the last line.

I can squish them though.

src/ci/bin/test.inc.sh
kshakir: If auto-vault-rendering of secrets fails during a run on one's laptop, `test.inc.sh` prints a warning and continues, letting the user know that they should be manually setting up the confs using their own secure creds. As far as the rest of the logic goes, the tests are still "secure" in this case and may continue.

src/ci/resources/cwl_conformance_test.wdl
kshakir: <sup>Hunh... this isn't the only `timout` in the code...</sup>

mcovarr: I'm surprised / disappointed if `--timeout` wasn't being passed through. Are you sure it wasn't a `--timout`?

Issue #2177, opened 2017-04-17
add singularity support


antonkulaga:
In many use-cases (exp. HPC and systems when you cannot run from root) Singularity ( http://singularity.lbl.gov/about ) is better than docker, it would be nice to see its support in Cromwell

vsoch:
@katevoss I'm one of the developers of Singularity and I would like to +1 this request! I don't know scala, but if it comes down to making an equivalent folder [like this one for Docker](https://github.com/broadinstitute/cromwell/tree/9aff9f2957d303a4789801d6a482777faf47d48f/dockerHashing/src/main/scala/cromwell/docker) I can give a first stab at it. Or if it's more helpful I can give complete examples for all the steps to working with singularity images. We have both a registry ([Singularity Hub](https://singularity-hub.org) that is hooked up to the singularity command line client to work with images. So - to integrate into cromwell you could either just run the container via a singularity command, or implement your own connection to our API to download the image. Please let me know how I might be helpful, and I'd gladly help. If you want me to give a go at scala I would just ask for your general workflow to compile and test functionality.

katevoss:
Hi @vsoch, thanks for chiming in! Supporting Singularity is definitely on our to-do list but it will take some exploration so it would be great to have your assistance as we plan our approach. I can't be sure when we'll start but we'll definitely keep you in the loop. Thanks again!

vsoch:
awesome! Yeah just ping on here when you are ready, and I'll be glad to help :)

katevoss:
@geoffjentry I know I've heard Singularity come up fairly regularly, do you know if there are users who are insisting on using Singularity in order to use Cromwell?

geoffjentry:
@katevoss I don't know of anyone who has said "I need this" or "If it existed I'd use Cromwell". Rather it's a topic which is building more steam across the space and I'm suggesting it'd be nice to be leaders and not followers here.

katevoss:
As a **user with images in Singularity**, I want **Cromwell to support using Singularity images (either via Singularity Hub and the command line, or connecting via API)**, so that I can **use Singularity images and not have to duplicate them in Docker**.
- Effort: ** @geoffjentry ? **
- Risk: **Small**
- Business value: **Medium**

vsoch:
[...]

PR #4002, opened 2018-08-14
Adding Dockerfile for sbt/scala development build

vsoch:
This is a quick PR to add a Dockerfile (and instructions) so that scala/sbt don't need to be installed on the user system. This is a nice step to have, and I decided to do before working on the issues addressed in https://github.com/broadinstitute/cromwell/issues/2177#issuecomment-412505720

@geoffjentry could you point me to the udocker example you mentioned? If there isn't a clear documentation for it, what I'm generally interested in to develop the singularity bit is:

 - where is the entrypoint for the scala application (when the user interacts to do something with docker, for example)
 - what is the general flow
 - is what we want to do for singularity comparable to the udocker example, so mirroring that code would do the trick?

Here are some quick links, if they are helpful!

 - https://hub.docker.com/r/vanessa/cromwell-dev/
 - https://github.com/vsoch/cromwell/tree/add/singularity/scripts/docker-develop

For the first, the container is still pushing... on my insanely great crappy wireless :)

### DIFF
--- a/docs/Building.md
+++ b/docs/Building.md
@@ -8,6 +8,13 @@ features or fixes, the following are required to build Cromwell from source:
 * [Java 8](http://www.oracle.com/technetwork/java/javase/overview/java8-2100321.html)
 * [Git](https://git-scm.com/)
 
+You can also use the [development image](https://github.com/broadinstitute/cromwell/tree/develop/scripts/docker-develop), and build a development container to work inside:
+
+```bash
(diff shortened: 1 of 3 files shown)


danbills:
:+1: this is really cool, thanks!

vsoch:
The power of laziness compels me! :P

geoffjentry:
:+1: thanks @vsoch !

PR #4003, opened 2018-08-14
Disaggregate "Unexpected terminal status" errors in Sentry

cjllanwarne:
Otherwise we won't be able to distinguish "Unexpected terminal status Failed but was waiting for Succeeded"


cjllanwarne:
I'm going to close this since I found out that it's possible to drill down into the exception types and get a sorted frequency list within the Sentry UI itself.

PR #4012, opened 2018-08-16
Added link to Google Docs ToC.

kshakir:


kshakir:
> Are there any other public google docs that should be added to the table of contents?

I'm sure there are, but on a quick scan I didn't come across anything obvious.

PR #4013, opened 2018-08-16
Hog factor, and concurrent jobs

cjllanwarne:
An early look at the hog factor:

Adds:
- A hog factor in the configuration file representing "how many greedy users would it take to use up all our resources"
  - Higher values protect Cromwell's resources to keep them available for small-scale users
  - Lower values let power users get their stuff done as fast as possible
  - 1 is equivalent to "no hog factor" (and is the default)
  - Idea: it would be awesome to be able to dynamically scale this up and down based on "we need to run a workflow" or "person X really needs to run their stuff at *full* speed for the next 4 hours"
- The ability to identify a workflow option as indicating hog group 
- A hog group is assigned to every `BackendWorkflowDescriptor` (using the specified workflow option if possible, or 'root workflow ID' if the specified option is not provided)
- An update to the `TokenPool` to make it hog-group aware

TODOs:
- [x] Enhance `RoundRobinQueueIteratorSpec` with hog-limit tests
- [x] Enhance `TokenQueueSpec` with hog-limit tests
- [ ] Add full-system tests demonstrating the hog limit in action


mcovarr:
image

engine/src/main/scala/cromwell/engine/workflow/tokens/TokenQueue.scala
```diff
@@ -31,23 +34,41 @@ final case class TokenQueue(queue: Queue[ActorRef], private [tokens] val pool: T
...
+        shuffleQueueAndGetLease(poolAcquisitionResult, placeholder, queue, newQueue) match {
+          case ShuffledQueueAndLease(Some(lease), nq) => DequeueResult(Option(LeasedActor(placeholder.actor, lease)), copy(queue = nq))
+          case ShuffledQueueAndLease(None, nq) => DequeueResult(None, copy(queue = nq))
+        }
+      case None =>
+        DequeueResult(None, this)
+    }
+  }
+
+  private def shuffleQueueAndGetLease(result: UnhoggableTokenPoolResult,
```
Horneth: Can you explain this ?

Horneth: Specifically the shuffle part, why does the hog get moved at the end ?

cjllanwarne: This is a pretty bad name - it's really "try to unpack a lease from this result, and depending on the result, choose what the new queue should look like"

cjllanwarne: The "hogs go to the back" rule is intended to be an approximation of "you stay at the back of the queue for as long as you're still using all your available tokens"

engine/src/main/scala/cromwell/engine/workflow/tokens/TokenQueue.scala
```diff
@@ -31,23 +34,41 @@ final case class TokenQueue(queue: Queue[ActorRef], private [tokens] val pool: T
...
+                              oldQueue: Queue[TokenQueuePlaceholder],
+                              newQueue: Queue[TokenQueuePlaceholder]): ShuffledQueueAndLease = result match {
+    case thl: TokenHoggingLease => ShuffledQueueAndLease(Some(thl), newQueue)
+    case ComeBackLater => ShuffledQueueAndLease(None, oldQueue)
+    // Hog response - move this placeholder to the end of the queue!
+    case Oink => ShuffledQueueAndLease(None, newQueue :+ thisPlaceholder )
   }
 
   /**
     * Returns true if there's at least on element that can be dequeued, false otherwise.
```
mcovarr: on => one

mcovarr:
Despite the image I pasted above this LGTM.

engine/src/main/scala/cromwell/engine/workflow/tokens/TokenQueue.scala
```diff
@@ -31,23 +34,41 @@ final case class TokenQueue(queue: Queue[ActorRef], private [tokens] val pool: T
...
     */
-  def available: Boolean = queue.nonEmpty && pool.leased() < pool.capacity
+  def available: Boolean =
+    queue.nonEmpty && queue.map(_.hogGroup).iterator.exists(hg => pool.available(hg))
```
Horneth: I have a vague reminiscence of something like this being very slow before, because having to loop through (potentially) the entire queue to determine if an element is available. Maybe a `toSet` would be help so we only check for unique hog groups.
If it's possible to have some sort of "perf unit test" with a large queue (maybe 100k) and the only available hog group at the end of the queue it could be interesting to see if it has any impact when trying to dequeue.

engine/src/main/scala/cromwell/engine/workflow/tokens/UnhoggableTokenPool.scala
```diff
@@ -0,0 +1,112 @@
...
+  referenceType = ReferenceType.Strong,
+  _factory = () => JobExecutionToken(tokenType, UUID.randomUUID()),
+  _reset = Function.const(()),
+  _dispose = Function.const(()),
+  _healthCheck = Function.const(true)) {
+
+  lazy val hogLimitOption: Option[Int] = tokenType match {
+    case JobExecutionTokenType(_, None, _) => None
+    case JobExecutionTokenType(_, _, 1) => None
+    case JobExecutionTokenType(_, Some(limit), hogFactor) => Some(math.max(1, math.round(limit.floatValue() / hogFactor.floatValue())))
```
Horneth: Will this effectively blow up if someone configured a hogFactor of 0 or a negative one ?

engine/src/main/scala/cromwell/engine/workflow/tokens/RoundRobinQueueIterator.scala
```diff
@@ -22,8 +23,16 @@ case class RoundRobinQueueIterator(initialTokenQueue: List[TokenQueue]) extends
...
-  override def next() = findFirst.getOrElse(throw new IllegalStateException("Token iterator is empty"))
+  /**
+    * And the same w.r.t the current pointer:
+    */
+  def updatedPointer = pointer
+
+  def hasNext = tokenQueues.exists(_.available)
+  def next() = findFirst.getOrElse(unexpectedlyEmpty)
+
+  def unexpectedlyEmpty: LeasedActor =
```
mcovarr: inline

engine/src/main/scala/cromwell/engine/workflow/tokens/TokenQueue.scala
```diff
@@ -31,23 +34,42 @@ final case class TokenQueue(queue: Queue[ActorRef], private [tokens] val pool: T
...
+                                                    oldQueue: Queue[TokenQueuePlaceholder],
+                                                    newQueue: Queue[TokenQueuePlaceholder]): UnpackedLeaseAndUpdatedQueue = result match {
+    case thl: TokenHoggingLease => UnpackedLeaseAndUpdatedQueue(Some(thl), copy(queue = newQueue))
+    case ComeBackLater => UnpackedLeaseAndUpdatedQueue(None, copy(queue = oldQueue))
+    // Hog response - move this placeholder to the end of the queue!
+    case Oink => UnpackedLeaseAndUpdatedQueue(None, copy(queue = newQueue :+ thisPlaceholder) )
   }
 
   /**
     * Returns true if there's at least on element that can be dequeued, false otherwise.
```
mcovarr: on => one

engine/src/main/scala/cromwell/engine/workflow/tokens/TokenQueue.scala
```diff
@@ -31,23 +34,42 @@ final case class TokenQueue(queue: Queue[ActorRef], private [tokens] val pool: T
...
+          case UnpackedLeaseAndUpdatedQueue(Some(lease), nq) => DequeueResult(Option(LeasedActor(placeholder.actor, lease)), nq)
+          case UnpackedLeaseAndUpdatedQueue(None, nq) if !(this.queue eq nq.queue) => nq.dequeue
+          case UnpackedLeaseAndUpdatedQueue(None, _) => DequeueResult(None, this)
+        }
+      case None =>
+        DequeueResult(None, this)
+    }
+  }
+
+  private def unpackLeaseOptionAndDetermineNewQueue(result: UnhoggableTokenPoolResult,
```
mcovarr: This is only called from the method above. If it were a function local to that method you could kill a lot of the noisy parameter passing.

engine/src/main/scala/cromwell/engine/workflow/tokens/UnhoggableTokenPool.scala
```diff
@@ -0,0 +1,112 @@
...
+          } else {
+            Oink
+          }
+        }
+      case None =>
+        super.tryAcquire() match {
+          case Some(lease) =>
+            val hoggingLease = new TokenHoggingLease(lease, hogGroup, this)
+            hoggingLease
+          case None => ComeBackLater
```
mcovarr: TOL This feels repetitive wrt the above though I realize it's not exactly the same...

engine/src/main/scala/cromwell/engine/workflow/tokens/UnhoggableTokenPool.scala
```diff
@@ -0,0 +1,112 @@
...
+  val MaxCapacity = 10 * 1000 * 1000
+
+  trait UnhoggableTokenPoolResult
+
+  final class TokenHoggingLease(lease: Lease[JobExecutionToken], hogGroup: String, pool: UnhoggableTokenPool) extends Lease[JobExecutionToken] with UnhoggableTokenPoolResult {
+    private[this] val dirty = new AtomicBoolean(false)
+    override protected[this] def a: JobExecutionToken = lease.get
+
+    override protected[this] def handleRelease(): Unit = {
+      if (dirty.compareAndSet(false, true)) {
```
mcovarr: why the atomicity concerns here?

PR #4015, opened 2018-08-17
adding circleci config stuffs

vsoch:
This pull request will trigger CI to run for the newly added configuration. 

## What does it do?
Specifically, this test builds and deploys a docker development container with sbt and scala to build cromwell. 

## How is it tested?
The user is instructed to bind the source code with "sbt assembly" in the `/code` directory in the container after binding the root of this repository to `/code`. Since we cannot have volumes in circle, we instead just test the added code to the container, also located at `/code`. Since the purpose of this container is to be a clean slate with sbt, scala to build cromwell, the primary test that is important is ensuring that `sbt assembly` runs successfully without a hitch. Any other "docker" tests for the actual cromwell (not building it) would not belong here, but with Docker containers meant to deploy cromwell proper.

# Where does it deploy?
The container will deploy to the `CONTAINER_NAME` defined in the circle environment settings (or in the circle config at `.circleci/config.yml`. By default, it will be tagged with the commit first 10 characters, and then latest, and you can change this behavior by defining `DOCKER_TAG` either in the config or circle environment (I don't see a reason to do this). Note that deploy is ONLY set up to happen on pushes to master (and you can change this to also be develop, if you choose, or to be both and then to deploy to tags `<branch>-<commit>` or something like that.

## Background
This was first done at the repo [vsoch/cromwell](https://github.com/vsoch/cromwell/pull/1) to test since I can't set it up for the broadinstitute. The (finally) working test is at [https://circleci.com/gh/vsoch/cromwell/11](https://circleci.com/gh/vsoch/cromwell/11). I forgot that I can't have volumes, so it took me many tries to remember this, derp :P 

When adding to the repository here, the following additional work will be needed for setup:

 - Turn on the repository to build at circleci. The first build, since there is no `.circici/config.yml` will probably just yell at you for having "Version 1.0" or not finding a config.
 - You will want to turn on building forked pull requests in the settings
 - Under environment variables, define the following:
    - `DOCKER_USER` should be the user to authenticate pushing
    - `DOCKER_PASS` password for that user (**important** do not turn on also testing of forked pull requests on their branch (different setting from above) as this could compromise these credentials.
    - `CONTAINER_NAME` should be something like `broadinstiutute/cromwell-dev`

 - The tag will always build the commit id, and then latest. If you want to change this behavior, define `DOCKER_TAG`.
 - ensure the branch logic (when things are triggered) is to your liking.
 - update the repo badge to be cromwell here and not on vsoch (after you connect the two!)

I noticed that there is no sbt version set (in some config file) - would this make sense to do?

```bash
[warn] No sbt.version set in project/build.properties, base directory: /
[info] Set current project to root (in build file:/)
[info] 1.2.1
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -98,11 +98,10 @@ dockerdeploy: &dockerdeploy
             docker push ${CONTAINER_NAME}:latest
       fi
 
-
 dockerbuild: &dockerbuild
     name: Build cromwell development Docker container
     command: |
-      docker build --cache-from=${CONTAINER_BUILD_NAME} -t ${CONTAINER_NAME}:${DOCKER_TAG} .
+      docker build --cache-from=${CONTAINER_BUILD_NAME} -f scripts/docker-develop/Dockerfile -t ${CONTAINER_NAME}:${DOCKER_TAG} .
 
 
 
@@ -134,8 +133,12 @@ jobs:
           name: Run tests
           command: |
             echo "What tests would you like, my dear McMuffins?"
+            echo "1. Testing for install of sbt and scala"
             docker run --it ${CONTAINER_NAME}:${DOCKER_TAG} which scala
+            docker run --it ${CONTAINER_NAME}:${DOCKER_TAG} scala --version
             docker run --it ${CONTAINER_NAME}:${DOCKER_TAG} which sbt
+            docker run --it ${CONTAINER_NAME}:${DOCKER_TAG} sbt --version
+            docker run -it ${CONTAINER_NAME}:${DOCKER_TAG} sbt assembly
       - run: *dockersave
 
   deploy:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -139,8 +139,8 @@ jobs:
             echo "docker run -it ${CONTAINER_NAME}:${DOCKER_TAG} which sbt"
             docker run -it ${CONTAINER_NAME}:${DOCKER_TAG} which sbt
             echo "1. Testing sbt assembly"
-            echo "docker run -it -v $PWD:/code ${CONTAINER_NAME}:${DOCKER_TAG} sbt assembly"
-            docker run -it -v $PWD:/code ${CONTAINER_NAME}:${DOCKER_TAG} sbt assembly
+            echo "docker run -v /tmp/src:/code ${CONTAINER_NAME}:${DOCKER_TAG} sbt assembly"
+            docker run -v $PWD:/code ${CONTAINER_NAME}:${DOCKER_TAG} sbt assembly
       - run: *dockersave
 
   deploy:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -101,7 +101,8 @@ dockerdeploy: &dockerdeploy
 dockerbuild: &dockerbuild
     name: Build cromwell development Docker container
     command: |
-      docker build --cache-from=${CONTAINER_BUILD_NAME} -f scripts/docker-develop/Dockerfile -t ${CONTAINER_NAME}:${DOCKER_TAG} .
+      source ${BASH_ENV}
+      docker build -f scripts/docker-develop/Dockerfile -t ${CONTAINER_NAME}:${DOCKER_TAG} .
 
 
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -135,10 +135,11 @@ jobs:
           command: |
             echo "What tests would you like, my dear McMuffins?"
             echo "1. Testing for install of sbt and scala"
-            docker run --it ${CONTAINER_NAME}:${DOCKER_TAG} which scala
-            docker run --it ${CONTAINER_NAME}:${DOCKER_TAG} scala --version
-            docker run --it ${CONTAINER_NAME}:${DOCKER_TAG} which sbt
-            docker run --it ${CONTAINER_NAME}:${DOCKER_TAG} sbt --version
+            docker run -it ${CONTAINER_NAME}:${DOCKER_TAG} which scala
+            docker run -it ${CONTAINER_NAME}:${DOCKER_TAG} scala --version
+            docker run -it ${CONTAINER_NAME}:${DOCKER_TAG} which sbt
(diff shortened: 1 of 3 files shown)


vsoch:
oups wrong base, disregard! lol.

vsoch:
okay I've just updated the circle badge to be cromwell on broadinstitute (which doesn't technically exist yet) but the purpose was to add all the commits where I failed like a struggling panda to get the testing setup :) All is working now!

vsoch:
I'm not sure why this failed? I didn't change any of the testing here... :/ Is the CI base broken or inconsistent?

kshakir:
Hi @vsoch,

Lot's of good stuff here on first glance. I'll dive deeper over the weekend.

For better or worse, depending on pricing, support, reliability, etc. etc. etc. we like to move around our CI. I personally also like being able to test scripts on my laptop as much as possible.

To that end, I'm trying to advocate for bash scripts that are then invoked from whatever CI we choose.

I haven't RTFM'ed enough of this PR nor CircleCI's manual yet to fully grasp what specific Circle features are being used here. Could a lot of the logic be separated from the `.circleci/config.yml` into a script, or multiple scripts if necessary?

On a related note, based on your expertise I may want to pick your brain to go over our [existing CI scripts](https://github.com/broadinstitute/cromwell/blob/develop/src/ci/bin/test.inc.sh#L38-L39) too as we move to Circle, or perhaps something even ~shinier~ [newer](https://news.ycombinator.com/item?id=17602838).

Re your build failing: it wasn't anything in your PR. Based on the logs there was a weird connection issue between Travis and Github returning HTTP 5xx errors during the tests.

.circleci/config.yml
```diff
@@ -0,0 +1,183 @@
...
+      - restore_cache:
+          key: dependency-cache
+      - setup_remote_docker
+      - run: *dockerenv
+      - run: *install
+      - run: *dockerload
+      - run: *dockerbuild
+      - run:
+          name: Run tests
+          command: |
```
vsoch: hey @kshakir I'd love to help! To answer your question, the entire workflow yaml can stay as is, and you can generalize to having them work locally by writing the commands below this comment into a shell script (and then just running the script for a command). You would also want to nicely manage the variables for the container name, etc, basically another copy paste from the `dockerenv` function here :)

vsoch:
> Lot's of good stuff here on first glance. I'll dive deeper over the weekend.

ok!

> For better or worse, depending on pricing, support, reliability, etc. etc. etc. we like to move around our CI. I personally also like being able to test scripts on my laptop as much as possible.

Yes we definitely can! See my comment above - it just is above moving the little snippet where the test actually happens from a command block to running a script from that same command block.

> To that end, I'm trying to advocate for bash scripts that are then invoked from whatever CI we choose.

+1!

> I haven't RTFM'ed enough of this PR nor CircleCI's manual yet to fully grasp what specific Circle features are being used here. Could a lot of the logic be separated from the .circleci/config.yml into a script, or multiple scripts if necessary?

I think the part we would want to take out are the testing commands, just executed via some primary file (that calls the individual ones, and which could be run on a host).

> On a related note, based on your expertise I may want to pick your brain to go over our existing CI scripts too as we move to Circle, or perhaps something even shinier newer.

Sure! I'm always around :)

> Re your build failing: it wasn't anything in your PR. Based on the logs there was a weird connection issue between Travis and Github returning HTTP 5xx errors during the tests.

Yeah, I've unfortunately been there :P Good luck this weekend!

vsoch:
ping! Just wanted to check here to see if you need anything from me?

kshakir:
@vsoch Sorry, our devops team has asked us to be especially thorough with PRs affecting our CI environments/dependencies. I've been bouncing between reading up on the [CircleCI docs](https://circleci.com/docs/2.0/configuration-reference/) and checking on several 🔥events this week.

If you have time, perhaps we can setup a remote session next week where you can give me a tour of yml and everything that's going on? If you propose three times that work for you I'll pick one. Feel free to msg me here or email if that's easier.

Otherwise I'll continue looking through the docs and get back to you once the flames die down. 🤞

kshakir:
Thanks for the brain dump. Chatting w/ some of our devops folks, they will work with us to move to CircleCI at some point. For now the team doesn't have the expertise nor bandwidth to evaluate how to do so securely. For example, we have several Hashicorp Vault rendered-secrets in our CI builds that should stay in the CI and not get vacuumed up into a docker image.

I still want to ensure your code lives on, so for now [I submitted a PR](https://github.com/broadinstitute/cromwell/pull/4038) that takes takes your work above, wraps the `docker build` in a portable script, then adds it as [a parallel regression test](https://travis-ci.org/broadinstitute/cromwell/builds/420635707) under our common CI scripts. Whenever we move over from Travis to Circle it should move with the other scripts.

vsoch:
okay, that works for me! To answer your questions about CircleCI:

 - environment variables used in the project are [encrypted](https://circleci.com/docs/2.0/security/#encryption) also using hashicorp vault! So, same thing or if not very similar deal as what you have.
 - once you set them in the interface, you can't change or see them
 - if the environment variables aren't set in the container with ENV or as flags with --env then they won't be saved. You would likely want to have them be [ARGS](https://vsupalov.com/docker-arg-env-variable-guide/) instead to be used and available in the container during build, but then not persisted in the container.

So, as long as:
 - you set secrets in the project and not the circle.yml
 -  you don't allow the CI to pass on secrets to other forked build requests (you would have to turn it on in settings are there are a lot of **warning don't do this!** prompts before you get there and
 - you use ARGS to expose needed variables from the environment to the container for building (that don't get saved). 

I think you'd be ok :) But sure, I'm definitely not a security expert. Anyway, since it's a single file, please feel free to grab the commit from here if/when you are ready.

vsoch:
Sure, but generally I would advocate to leave open PRs that you might want other contributors to see (that they have not been fully addressed / might be integrated in the future). There isn't any harm in leaving it open, until it's irrelevant and has no chance of being looked over again!

This is of course my 0.02 and just an opinion, so if you feel strongly about closing it, that is up to you.

geoffjentry:
Hi @vsoch - we do hope to move to circleci but need to tie in to other internal things before that happens, which won't be i nthe immediate future. Thanks for poking at this, but closing for now.

PR #4016, opened 2018-08-17
POC call cache copy fail blacklisting.

mcovarr:


backend/src/main/scala/cromwell/backend/standard/callcaching/StandardCacheHitCopyingActor.scala
```diff
@@ -178,6 +196,13 @@ abstract class StandardCacheHitCopyingActor(val standardParams: StandardCacheHit
           stay() using Option(newData)
       }
     case Event(IoFailure(command: IoCommand[_], failure), Some(data)) =>
+      val pattern = ".*does not have storage.objects.get access to ([^/]+).*".r.pattern
```
Horneth: ToL for when out of POC state: I think this could be pushed up to the IoActor and have a more generic "Forbidden" kind of exception that we could return for all filesystems when appropriate.

mcovarr:
temporarily closing

PR #4018, opened 2018-08-20
Embedded cwltool/salad via heterodon.

kshakir:


src/ci/bin/test.inc.sh
```diff
@@ -312,14 +311,11 @@ cromwell::private::pull_common_docker_images() {
     docker pull ubuntu
 }
 
-cromwell::private::install_cwltool() {
+cromwell::private::install_cwltest() {
     # TODO: No clue why these are needed for cwltool. If you know please update this comment.
```
mcovarr: is this the time to update this comment?

mcovarr:
and another :+1: for the `heterodon` name

src/ci/bin/test.inc.sh
kshakir: I still don't know why this is required. 🤔 So I'm leaving it until someone gives an answer.

cwl/src/main/scala/cwl/CwltoolRunner.scala
```diff
@@ -0,0 +1,71 @@
...
+sealed trait CwltoolRunner {
+  def salad(file: BFile): String
+}
+
+object CwltoolRunner {
+  private lazy val config = ConfigFactory.load
+
+  lazy val instance: CwltoolRunner = {
+    val runnerClass = config.getString("cwltool-runner.class")
+    Class.forName(runnerClass).newInstance().asInstanceOf[CwltoolRunner]
```
danbills: ToL Could just import this lib/class and `new` it up, no need for jvm-level instantiation IMO

PR #4019, opened 2018-08-20
Execution bucket call caching hint support.

mcovarr:
OK slight change of plan: execution bucket call cache hints will be opt-in for now via workflow options with no changes required in Cromwell config. This actually does not have to be  PAPI specific so the proposed nomenclature would be more general:

```json
{
  call_cache_hit_path_prefixes: [ "gs://bucket-full-o-cache-hits", "gs://another-bucket-o-cache-hits" ]
}
```

"opt-in" here means that for now the caller would be responsible for specifying the prefixes by which to filter cache hits. If no prefixes are specified Cromwell will do no cache hit filtering. Backends have the ability to specify a root execution path to prepend to this list. For now this root execution path would only be prepended by PAPI since that's the only backend for which a sensible root could be determined by inspection (i.e. the GCS bucket).

The number of prefixes in this list would be limited, for now I'm arbitrarily choosing 3 total.

@ruchim

### DIFF
--- a/core/src/main/scala/cromwell/core/WorkflowOptions.scala
+++ b/core/src/main/scala/cromwell/core/WorkflowOptions.scala
@@ -1,7 +1,9 @@
 package cromwell.core
 
+import cats.syntax.validated._
 import com.typesafe.config.ConfigFactory
 import common.util.TryUtil
+import common.validation.ErrorOr.ErrorOr
 import spray.json._
 
 import scala.collection.JavaConverters._
@@ -47,6 +49,7 @@ object WorkflowOptions {
   // Caching
   case object WriteToCache extends WorkflowOption("write_to_cache")
   case object ReadFromCache extends WorkflowOption("read_from_cache")
+  case object CallCacheHitPathPrefixes extends WorkflowOption("call_cache_hit_path_prefixes")
 
   // Copying
   case object FinalWorkflowLogDir extends WorkflowOption("final_workflow_log_dir")
@@ -93,6 +96,7 @@ object WorkflowOptions {
       case (k, v: JsObject) if defaultRuntimeOptionKey.equals(k) => k -> Success(v)
       case (k, v: JsNumber) => k -> Success(v)
       case (k, v) if isEncryptedField(v) => k -> Success(v)
+      case (k, v: JsArray) => k -> Success(v)
       case (k, v) => k -> Failure(new UnsupportedOperationException(s"Unsupported key/value pair in WorkflowOptions: $k -> $v"))
     }
 
@@ -145,6 +149,12 @@ case class WorkflowOptions(jsObject: JsObject) {
     case Some(jsBool: JsBoolean) => Success(jsBool.value)
     case Some(jsVal: JsValue) => Failure(new IllegalArgumentException(s"Unsupported JsValue as JsBoolean: $jsVal"))
     case None => Failure(OptionNotFoundException(s"Field not found: $key"))
+  }
+
+  def getVectorOfStrings(key: String): ErrorOr[Option[Vector[String]]] = jsObject.fields.get(key) match {
+    case Some(jsArr: JsArray) => Option(jsArr.elements collect { case e: JsString => e.value } ).validNel
+    case Some(jsVal: JsValue) => s"Unsupported JsValue as JsArray: $jsVal".invalidNel
+    case _ => None.validNel
   }
 
   def getDefaultRuntimeOption(key: String): Try[JsValue] = defaultRuntimeOptions map { attributes =>
--- a/core/src/main/scala/cromwell/core/callcaching/CallCachingMode.scala
+++ b/core/src/main/scala/cromwell/core/callcaching/CallCachingMode.scala
@@ -19,7 +19,7 @@ case object CallCachingOff extends CallCachingMode {
   override val withoutWrite = this
 }
 
-case class CallCachingActivity(readWriteMode: ReadWriteMode, options: CallCachingOptions = CallCachingOptions(invalidateBadCacheResults = true)) extends CallCachingMode {
+case class CallCachingActivity(readWriteMode: ReadWriteMode, options: CallCachingOptions = CallCachingOptions()) extends CallCachingMode {
   override val readFromCache = readWriteMode.r
   override val writeToCache = readWriteMode.w
   override lazy val withoutRead: CallCachingMode = if (!writeToCache) CallCachingOff else this.copy(readWriteMode = WriteCache)
@@ -35,4 +35,4 @@ case object ReadCache extends ReadWriteMode { override val w = false }
 case object WriteCache extends ReadWriteMode { override val r = false }
 case object ReadAndWriteCache extends ReadWriteMode
 
-final case class CallCachingOptions(invalidateBadCacheResults: Boolean = true)
+final case class CallCachingOptions(invalidateBadCacheResults: Boolean = true, workflowOptionCallCachePrefixes: Option[Vector[String]] = None)
--- a/database/sql/src/main/scala/cromwell/database/slick/CallCachingSlickDatabase.scala
+++ b/database/sql/src/main/scala/cromwell/database/slick/CallCachingSlickDatabase.scala
@@ -58,10 +58,25 @@ trait CallCachingSlickDatabase extends CallCachingSqlDatabase {
     runTransaction(action)
   }
 
-  override def hasMatchingCallCachingEntriesForBaseAggregation(baseAggregationHash: String)
+  override def hasMatchingCallCachingEntriesForBaseAggregation(baseAggregationHash: String, callCachePrefixes: Option[List[String]] = None)
                                                               (implicit ec: ExecutionContext): Future[Boolean] = {
-    val action = dataAccess.existsCallCachingEntriesForBaseAggregationHash(baseAggregationHash).result
-
+    val action = callCachePrefixes match {
+      case None => dataAccess.existsCallCachingEntriesForBaseAggregationHash(baseAggregationHash).result
+      case Some(ps) =>
+        case class PrefixAndLength(prefix: String, length: Int)
+        // Trying to match a prefix of length 1 with an empty string is (intentionally) doomed to fail.
+        val doNotMatch = PrefixAndLength("", 1)
+        // `total` is ps.apply as a total function which returns an Option[String] instead of a String.
+        // `total` returns `None` if there is no element at the specified index.
+        val total = ps.lift
+        // Take the first three prefixes that have been supplied or fallback values that will intentionally fail to match anything.
+        val one :: two :: three :: _ = (0 to 2).toList map { total(_) map { p => PrefixAndLength(p, p.length) } getOrElse doNotMatch }
+        dataAccess.existsCallCachingEntriesForBaseAggregationHashWithCallCachePrefix(
+          (baseAggregationHash,
+            one.prefix, one.length,
+            two.prefix, two.length,
+            three.prefix, three.length)).result
+    }
     runTransaction(action)
   }
 
--- a/database/sql/src/main/scala/cromwell/database/slick/tables/CallCachingAggregationEntryComponent.scala
+++ b/database/sql/src/main/scala/cromwell/database/slick/tables/CallCachingAggregationEntryComponent.scala
@@ -5,7 +5,7 @@ import cromwell.database.sql.tables.CallCachingAggregationEntry
 
 trait CallCachingAggregationEntryComponent {
 
-  this: DriverComponent with CallCachingEntryComponent =>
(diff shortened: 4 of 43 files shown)


database/sql/src/main/scala/cromwell/database/slick/tables/CallCachingAggregationEntryComponent.scala
```diff
@@ -50,6 +50,20 @@ trait CallCachingAggregationEntryComponent {
...
   )
 
+  val existsCallCachingEntriesForBaseAggregationHashWithExecutionBucket = Compiled(
+    (baseAggregation: Rep[String], executionBucket: Rep[String], executionBucketLength: Rep[Int]) => (for {
+      callCachingEntry <- callCachingEntries
+      if callCachingEntry.allowResultReuse
+      callCachingAggregationEntry <- callCachingAggregationEntries
+      if callCachingEntry.callCachingEntryId === callCachingAggregationEntry.callCachingEntryId
+      if callCachingAggregationEntry.baseAggregation === baseAggregation
+      detritus <- callCachingDetritusEntries
```
Horneth: Is this checking that all detrituses have the bucket prefix ?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheHashingJobActor.scala
```diff
@@ -137,7 +141,9 @@ class CallCacheHashingJobActor(jobDescriptor: BackendJobDescriptor,
     val hashingJobActorData = CallCacheHashingJobActorData(fileHashRequests.toList, callCacheReadingJobActor)
     startWith(WaitingForHashFileRequest, hashingJobActorData)
 
-    val initialHashingResult = InitialHashingResult(initialHashes, calculateHashAggregation(initialHashes, MessageDigest.getInstance("MD5")))
+    val aggregatedBaseHash = calculateHashAggregation(initialHashes, MessageDigest.getInstance("MD5"))
+    val executionBucketHint = initializationData collect { case d: StandardInitializationData => ExecutionBucketHint(d.workflowPaths.executionRootString) } toList
```
Horneth: This is not limited to gcs buckets is it ? Won't it also filter on SFS ?

mcovarr: Yes which is why it fails for non-GCS, I need to loop back and fix this particularly POCy bit. :)

database/sql/src/main/scala/cromwell/database/slick/tables/CallCachingAggregationEntryComponent.scala
mcovarr: This is meant to check that there exists a detritus file that belongs to this call caching entry whose GCS bucket prefix matches the desired value.

database/sql/src/main/scala/cromwell/database/slick/tables/CallCachingAggregationEntryComponent.scala
```diff
@@ -50,6 +50,20 @@ trait CallCachingAggregationEntryComponent {
     } yield ()).exists
   )
 
+  val existsCallCachingEntriesForBaseAggregationHashWithExecutionBucket = Compiled(
```
Horneth: ToL: the "buckets" in the name and variable names might be overly specific as this really just checks that the detritus starts with `executionBucket` but it could be anything

database/sql/src/main/scala/cromwell/database/slick/tables/CallCachingAggregationEntryComponent.scala
Horneth: It'll be interesting to see if this extra join has any effect on the DB performance of reading cache hits. Even if it does the overall gain will probably be enormous anyway but would be good to know with the incoming perf testing

database/sql/src/main/scala/cromwell/database/slick/tables/CallCachingAggregationEntryComponent.scala
```diff
@@ -50,6 +50,25 @@ trait CallCachingAggregationEntryComponent {
...
+      if callCachingEntry.allowResultReuse
+      callCachingAggregationEntry <- callCachingAggregationEntries
+      if callCachingEntry.callCachingEntryId === callCachingAggregationEntry.callCachingEntryId
+      if callCachingAggregationEntry.baseAggregation === baseAggregation
+      detritus <- callCachingDetritusEntries
+      if detritus.callCachingEntryId === callCachingEntry.callCachingEntryId
+      detritusPath = detritus.detritusValue.map { x => x.asColumnOf[String] }
+      if (detritusPath.substring(0, prefix1Length) === prefix1) ||
+        (detritusPath.substring(0, prefix2Length) === prefix2) ||
+        (detritusPath.substring(0, prefix3Length) === prefix3)} yield ()).exists
```
Horneth: ToL: I wonder if we shouldn't start hashing the output bucket and store it along the other cache hashes so we can use it in the future directly in the hash lookup instead of having to look here, but maybe for later

database/sql/src/main/scala/cromwell/database/slick/CallCachingSlickDatabase.scala
```diff
@@ -58,10 +58,25 @@ trait CallCachingSlickDatabase extends CallCachingSqlDatabase {
...
+      case None => dataAccess.existsCallCachingEntriesForBaseAggregationHash(baseAggregationHash).result
+      case Some(ps) =>
+        case class PrefixAndLength(prefix: String, length: Int)
+        // Trying to match a prefix of length 1 with an empty string is (intentionally) doomed to fail.
+        val doNotMatch = PrefixAndLength("", 1)
+        // `total` is ps.apply as a total function which returns an Option[String] instead of a String.
+        // `total` returns `None` if there is no element at the specified index.
+        val total = ps.lift
+        // Take the first three prefixes that have been supplied or fallback values that will intentionally fail to match anything.
+        val one :: two :: three :: _ = (0 to 2).toList map { total(_) map { p => PrefixAndLength(p, p.length) } getOrElse doNotMatch }
```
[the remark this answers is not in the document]
mcovarr: One certainly could but I don't think I see a formulation that would be significantly simpler than this and achieve the same effect? 🤔

database/sql/src/main/scala/cromwell/database/slick/tables/CallCachingAggregationEntryComponent.scala
mcovarr: So we would add an element to our set of hashes? Wouldn't this cause misses on the existing cache hits?

Horneth: Yeah that's why I was saying "in the future" but you're right it would require a migration for existing hits.
Currently we have the `CALL_CACHING_AGGREGATION_ENTRY` table which has a "base hash" (which is everything but input files) and "input files hash".
One way would _maybe_  to add a third nullable "hint hash", that we would start populating now with an md5 of the root bucket.
We could use it by hashing the "hint buckets" we get from workflow options and filter rows for which the "hint hash" is either null or matches one of the hint buckets we have.
Again I'm really just ToL and this is likely an over optimization for now, and not a request to change this PR :)
I just think we should generally try keeping the answer to the "is this a cache hit ?" question as efficient as possible, since it has proven to be pathologically bad in the past (which this is of course very far from being).

database/sql/src/main/scala/cromwell/database/slick/CallCachingSlickDatabase.scala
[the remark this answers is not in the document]
Horneth: Oh yeah since you always need 3 of them and the list could have less I guess `take` would not make it simpler, Disregard :)

core/src/main/scala/cromwell/core/WorkflowOptions.scala
```diff
@@ -147,6 +151,12 @@ case class WorkflowOptions(jsObject: JsObject) {
     case None => Failure(OptionNotFoundException(s"Field not found: $key"))
   }
 
+  def getVectorOfStrings(key: String): ErrorOr[Option[Vector[String]]] = jsObject.fields.get(key) match {
+    case Some(jsArr: JsArray) => Option(jsArr.elements collect { case e: JsString => e.value } ).validNel
```
danbills: ToL why the `Option`?

database/sql/src/main/scala/cromwell/database/slick/tables/CallCachingAggregationEntryComponent.scala
```diff
@@ -50,6 +50,25 @@ trait CallCachingAggregationEntryComponent {
...
+    ) => (for {callCachingEntry <- callCachingEntries
+      callCachingEntry <- callCachingEntries
+      if callCachingEntry.allowResultReuse
+      callCachingAggregationEntry <- callCachingAggregationEntries
+      if callCachingEntry.callCachingEntryId === callCachingAggregationEntry.callCachingEntryId
+      if callCachingAggregationEntry.baseAggregation === baseAggregation
+      detritus <- callCachingDetritusEntries
+      if detritus.callCachingEntryId === callCachingEntry.callCachingEntryId
+      detritusPath = detritus.detritusValue.map { x => x.asColumnOf[String] }
+      if (detritusPath.substring(0, prefix1Length) === prefix1) ||
```
danbills: ToL thinking you could do something with `Option[String]` args,  `flatten` a list of them, and filter by `startsWith` but not 100% sure it would be cleaner.  could get rid of the prefixlength and noMatch stuff.  Will checkout your PR and attempt

core/src/main/scala/cromwell/core/WorkflowOptions.scala
```diff
@@ -147,6 +151,12 @@ case class WorkflowOptions(jsObject: JsObject) {
...
+    case Some(jsVal: JsValue) => s"Unsupported JsValue as JsArray: $jsVal".invalidNel
+    case _ => None.validNel
```
mcovarr: to be able to return `None` here 🙂

database/sql/src/main/scala/cromwell/database/slick/tables/CallCachingAggregationEntryComponent.scala
```diff
@@ -50,6 +50,25 @@ trait CallCachingAggregationEntryComponent {
     } yield ()).exists
   )
 
+  val existsCallCachingEntriesForBaseAggregationHashWithCallCachePrefix = Compiled(
+    (baseAggregation: Rep[String],
+     prefix1: Rep[String], prefix1Length: Rep[Int],
+     prefix2: Rep[String], prefix2Length: Rep[Int],
+     prefix3: Rep[String], prefix3Length: Rep[Int]
+    ) => (for {callCachingEntry <- callCachingEntries
```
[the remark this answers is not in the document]
mcovarr: uh whoops good catch

core/src/main/scala/cromwell/core/callcaching/CallCachingMode.scala
```diff
@@ -35,4 +35,8 @@ case object ReadCache extends ReadWriteMode { override val w = false }
 case object WriteCache extends ReadWriteMode { override val r = false }
 case object ReadAndWriteCache extends ReadWriteMode
 
-final case class CallCachingOptions(invalidateBadCacheResults: Boolean = true)
+final case class CallCachingOptions(invalidateBadCacheResults: Boolean, workflowOptionCallCachePrefixes: Option[Vector[String]])
+
+object CallCachingOptions {
+  def default: CallCachingOptions = CallCachingOptions(invalidateBadCacheResults = true, workflowOptionCallCachePrefixes = None)
```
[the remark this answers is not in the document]
mcovarr: IMHO callers should be making a conscious choice about these arguments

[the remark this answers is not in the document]
mcovarr: Having said that I might get rid of this particular method because it's weird and only used in one place.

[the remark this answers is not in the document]
mcovarr: Hmm let me see if I can clean this up a bit.

database/sql/src/main/scala/cromwell/database/slick/tables/CallCachingAggregationEntryComponent.scala
mcovarr: I tried a few things before I ended up with this yukkiness and I think mostly due to the restrictions of `Compiled` ended up with this. At least once I had something that only blew up at runtime, so watch out for that. 🙂

Issue #3748, opened 2018-06-07
MIME multipart parsing is inefficient and brittle


elerch:
The parseOutput function https://github.com/broadinstitute/cromwell/blob/aws_backend/supportedBackends/aws/src/main/scala/cromwell/backend/impl/aws/AwsBatchJob.scala#L55, is brittle, but at least it's slow and not idiomatic scala. On the plus side, it is working in simple cases and has a unit test around it.

This should be processing output line by line and handle proper MIME multipart format. Speaking of which, we should be outputting to CloudWatch logs in MIME multipart format as well. We're missing the empty line between headers and body in both cases.

As an edge case, if a task outputs "Content-Disposition: attachment; filename=blahblah" on it's own line anywhere in the body of stdout or stderr, the brittle code here will happily process it as a header. At least it should throw in these circumstances.

Bonus points if negative unit tests for invalid logs output are also added.

This last second change (d81b3e9) has also apparently has also broken a bunch of Centaur tests, probably due to the hacking it's doing to the script output (https://github.com/broadinstitute/cromwell/blob/aws_backend/supportedBackends/aws/src/main/scala/cromwell/backend/impl/aws/AwsBatchJob.scala#L150). The idea, at least, appears solid.

Issue #3750, opened 2018-06-07
Reuse job definitions


elerch:
Currently a new job definition is created per job submitted to AWS Batch. There should be built into AwsBatchJob.scala a mechanism to reuse definitions.

One thought here is to perform a hash of the parameters used to create a job definition, then setting the job definition name to that hash. It is then relatively easy to perform a describeJobDefinitions call against aws batch to look for that name, and create it if it does not exist.

PR #4020, opened 2018-08-20
Initial ecs-agent proxy implementation

### LINKED ISSUE #3804: Add support for staging INPUTS/OUTPUTS from/to AWS S3

elerch:
This PR removes the agent/sidecar approach for a lighter proxy to be used in conjunction with a patched ecs agent running on the host. It removes the potential race caused by the agent watching the docker stream, as the ecs agent will launch the proxy directly.  It resolves #3804 in all cases and allows the Haplotypecaller test to succeed.


elerch:
FYI for reviewers. I've added the following additional changes to this commit:

* Added retry handling for a few transient errors I saw during test runs
* Moved stdout/stderr/rc file writes to s3 into the proxy to resolve an intermittent timing issue seen in testing
* Removed the MIME parsing, as this is no longer needed due to the above change. This resolves #3748
* Addressed a TODO in the code regarding a large number of input parameters
* Removed broken code in the proxy (since the shell would continue, this ultimately is a cosmetic change)
* Moved the actual command text from the job definition (RegisterJobDefinition) to the job submission (SubmitJob) call. This balances out the payload to allow more work in the job before hitting AWS Batch payload limits: https://docs.aws.amazon.com/batch/latest/userguide/service_limits.html. Additionally it should allow easier consolidation of job definitions in the future. #3750

supportedBackends/aws/src/main/scala/cromwell/backend/impl/aws/AwsBatchJobDefinition.scala
```diff
@@ -32,13 +32,18 @@
...
+import scala.collection.mutable.ListBuffer
 import cromwell.backend.BackendJobDescriptor
-import cromwell.core.path.Path
+import cromwell.backend.io.JobPaths
 import software.amazon.awssdk.services.batch.model.{ContainerProperties, KeyValuePair}
 import wdl4s.parser.MemoryUnit
 
 import scala.collection.JavaConverters._
 
+// GZip
```
mcovarr: TOL not really necessary to comment this, when we optimize the imports with IntelliJ the meaning will be lost anyway 🙂

danbills:
I submitted [this PR](https://github.com/broadinstitute/cromwell/pull/4076) to address a couple things

PR #4036, opened 2018-08-24
Bump file read timeout to maybe reduce test failures

aednichols:
We've seen a few times the IO error `Failed to evaluate job outputs [...] Futures timed out after [10 seconds]` in tests.

I hypothesize that file read times from buckets may suffer from occasional outliers due to 🌩

I know this is the right timeout to change thanks to [this branch](https://github.com/broadinstitute/cromwell/compare/aen_make_it_timeout?expand=1) where I induced error `Failed to evaluate job outputs [...] Futures timed out after [10 microseconds]`


aednichols:
`WriteWaitTimeout` is already `10.minutes`

Which actually makes me wonder whether `ReadWaitTimeout` shouldn't be the same.

PR #4037, opened 2018-08-24
Backstop Resources: TotalMaxJobsPerRootWf and MaxScatterWidth

salonishah11:
This PR introduces two new config settings: `max-scatter-width-per-scatter` and `total-max-jobs-per-root-workflow`. Their purpose is:

- max-scatter-width-per-scatter: Maximum scatter width per scatter node. If a workflow has a scatter width more than this number, Cromwell will fail the workflow
- total-max-jobs-per-root-workflow: Total max. jobs that can be created per root workflow. If workflow tries to create jobs more than this number, Cromwell will fail the workflow by: no longer creating new jobs and let the jobs that have already been started finish, and then fail the workflow

Remaining: Update the default values of these two configurations in `reference.conf` based on threshold determined from Performance Tests.

### DIFF
--- a/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
@@ -76,9 +76,8 @@ case class WorkflowExecutionActor(params: WorkflowExecutionActorParams)
 
   val executionStore: ErrorOr[ActiveExecutionStore] = ExecutionStore(workflowDescriptor.callable, params.totalJobsByRootWf)
 
-  //TODO: Saloni- update the below comment? Ask Thibault?
-  // This is done to avoid race condition. If executionStore returns a Failure, the WEA will fail by sending WorkflowExecutionFailedResponse
-  // to it's parent and goto WorkflowExecutionFailedState state directly.
+  // If executionStore returns a Failure about root workflow creating jobs more than total jobs per root workflow limit, the WEA will fail by sending WorkflowExecutionFailedResponse
+  // to it's parent and kill itself.
   executionStore match {
     case Valid(validExecutionStore) => {
       startWith(
@@ -88,7 +87,13 @@ case class WorkflowExecutionActor(params: WorkflowExecutionActorParams)
     }
     case Invalid(e) => {
       context.parent ! WorkflowExecutionFailedResponse(Map.empty, new Exception(s"Failed to initialize WorkflowExecutionActor. Error: $e"))
-      goto(WorkflowExecutionFailedState)
+
+      // The actor is started with some state before killing itself, otherwise FSM.goto throws NullPointerException as it doesn't find currentState
+      startWith(
+        WorkflowExecutionFailedState,
+        WorkflowExecutionActorData(workflowDescriptor, ioEc, new AsyncIo(params.ioActor, GcsBatchCommandBuilder), params.totalJobsByRootWf, ExecutionStore.empty)
+      )
+      context.stop(self)
     }
   }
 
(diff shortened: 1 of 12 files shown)


engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -63,10 +72,34 @@ case class WorkflowExecutionActor(params: WorkflowExecutionActorParams)
...
 
-  startWith(
-    WorkflowExecutionPendingState,
-    WorkflowExecutionActorData(workflowDescriptor, ioEc, new AsyncIo(params.ioActor, GcsBatchCommandBuilder))
-  )
+
+  val executionStore: ErrorOr[ActiveExecutionStore] = ExecutionStore(workflowDescriptor.callable, params.totalJobsByRootWf)
+
+  // If executionStore returns a Failure about root workflow creating jobs more than total jobs per root workflow limit,
+  // the WEA will fail by sending WorkflowExecutionFailedResponse to it's parent and kill itself
```
mcovarr: its

mcovarr: its

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala
```diff
@@ -52,6 +58,9 @@ case class WorkflowExecutionActor(params: WorkflowExecutionActorParams)
   private val restarting = params.startState.restarted
   private val tag = s"WorkflowExecutionActor [UUID(${workflowDescriptor.id.shortString})]"
 
+  private val DefaultTotalMaxJobsPerRootWf = 1000000
+  private val TotalMaxJobsPerRootWf = ConfigFactory.load.getConfig("system").as[Option[Int]]("total-max-jobs-per-root-workflow").getOrElse(DefaultTotalMaxJobsPerRootWf)
```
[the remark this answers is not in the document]
salonishah11: Makes sense! Will make parent pass it to child actors.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/keys/ScatterKey.scala
```diff
@@ -17,6 +21,9 @@ import wom.values.WomValue
 import scala.language.postfixOps
 
 private [execution] case class ScatterKey(node: ScatterNode) extends JobKey {
+  private val DefaultMaxScatterSize = 1000000
+  private val MaxScatterWidth = ConfigFactory.load.getConfig("system").as[Option[Int]]("max-scatter-width-per-scatter").getOrElse(DefaultMaxScatterSize)
```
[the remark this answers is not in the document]
salonishah11: Sweet! Will make that change.

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/keys/ScatterKey.scala
```diff
@@ -89,17 +96,33 @@ private [execution] case class ScatterKey(node: ScatterNode) extends JobKey {
...
     }
 
+    // Checks the scatter width of a scatter node and builds WorkflowExecutionDiff accordingly
+    // If scatter width is more than max allowed limit, it fails the ScatterNode key
+    def buildExecutionDiff(scatterSize: Int, arrays: List[ScatterVariableAndValue]): WorkflowExecutionDiff = {
+      if(scatterSize > MaxScatterWidth) {
+        workflowExecutionActor ! JobFailedNonRetryableResponse(this, new Exception(s"Workflow has scatter width $scatterSize, which is more than the max scatter width $MaxScatterWidth allowed per scatter!"), None)
+
+        // even though the status is updated to Running, it will eventually be changed to Failed when WEA receives JobFailedNonRetryableResponse event
+        WorkflowExecutionDiff(Map(this -> ExecutionStatus.Running))
```
mcovarr: Is there harm in setting this to Failed instead? It might be kind of confusing to debug this (not that it would ever need debugging of course 😛 ) if one sees the job in Running state when a job is started after the execution limit has been hit.

salonishah11: I don't see any harm in setting it to Failed state. Will make that change. 👍

Issue #1499, opened 2016-09-29
Fix polling when jobs are killed by HPC scheduler


delocalizer:
This is an issue to track the same problem that @ffinfo has raised on Gitter and PR #1346. The scenario for us is an HPC cluster environment with a scheduler (SGE or PBS) that strictly enforces walltimes; having jobs be killed by the scheduler (without producing rc file) when they exceed the requested walltime is, if not exactly common, not that rare either.

The hacky non-async solution I have been using (up until the nice configurable backends of release 0.21 changed everything in backend-land) was to have two check cycles, a frequent cheap one to see if rc existed and occasional expensive one to poll the scheduler itself: https://github.com/delocalizer/cromwell/blob/develop/engine/src/main/scala/cromwell/engine/backend/pbs/PbsBackend.scala#L128-L166

katevoss:
@delocalizer Hi Conrad, I'm going through the backlog and came across your issue. How often are you still experiencing this problem?

delocalizer:
@katevoss Hi Kate. I think there are two aspects to the issue worth considering - the first being how often we hit this problem in practice (I'll get back with you after I ask the production team) and the second being whether the underlying cause has been addressed - which is that relying only on the creation of a file to detect task completion is not robust at least for SGE/PBS type backends where jobs may be killed by the scheduler out-of-process without creating a file. Based only on the release changelog I suspect that the answer to the second is no. I suggest re-using the "check-alive" configuration value that's documented as currently used only on cromwell restart, for periodic (but infrequent) polling of the scheduler.

delocalizer:
@katevoss #2281 is a dupe of this

delocalizer:
#2315 (closed) also a dupe

katevoss:
Good catch! I closed #2281. 

Transferring conversation here:
From @MatthewMah 
> When running jobs on backends with job runtime limits such as LSF or SLURM, jobs reaching the runtime limits are killed by the backend. [Cromwell never detects that this occurs](http://gatkforums.broadinstitute.org/wdl/discussion/9542/does-cromwell-detect-task-failures-based-on-check-alive), and will wait forever for a job that is already dead. It would be helpful to configure periodic checks for whether tasks are still alive, and enter failure modes for non-zero return codes when unfinished tasks are no longer alive. 

From @geoffjentry 
[...]

PR #4039, opened 2018-08-27
Adding example "backends" for singularity

vsoch:
hey cromwell! I know the other PR is still being reviewed (and you are busy, no worries!) but I just wanted to put this example here of (one idea) for running a singularity container, as an instance or binary, just via a configuration. I don't think any special changes are needed here to the backend of cromwell, it's really just executing commands to a container with a particular set of inputs (container name, etc.) Likely we would want to add a general variable to plug in any user specific arguments (e.g., custom binds) and singularity arguments? (e.g., debug?) I don't know if this is "how it's supposed to look" but here is what a test run spits out for me:

```bash
$ java -jar -Dconfig.file=backends/backend.conf -Dbackend.default=singularity cromwell-34.jar run runners/test.wdl -i data/TEST-YEAST/inputs.json -o workflow_opts/singularity.json
[2018-08-27 02:03:57,30] [info] Running with database db.url = jdbc:hsqldb:mem:3cd4a928-7743-4252-8bd9-e905ebe70980;shutdown=false;hsqldb.tx=mvcc
[2018-08-27 02:04:05,58] [info] Running migration RenameWorkflowOptionsInMetadata with a read batch size of 100000 and a write batch size of 100000
[2018-08-27 02:04:05,60] [info] [RenameWorkflowOptionsInMetadata] 100%
[2018-08-27 02:04:05,75] [info] Running with database db.url = jdbc:hsqldb:mem:c850e4aa-3449-4d7e-bf04-4593fe287777;shutdown=false;hsqldb.tx=mvcc
[2018-08-27 02:04:06,15] [warn] This actor factory is deprecated. Please use cromwell.backend.google.pipelines.v1alpha2.PipelinesApiLifecycleActorFactory for PAPI v1 or cromwell.backend.google.pipelines.v2alpha1.PipelinesApiLifecycleActorFactory for PAPI v2
[2018-08-27 02:04:06,16] [warn] Couldn't find a suitable DSN, defaulting to a Noop one.
[2018-08-27 02:04:06,16] [info] Using noop to send events.
[2018-08-27 02:04:06,43] [info] Slf4jLogger started
[2018-08-27 02:04:06,64] [info] Workflow heartbeat configuration:
{
  "cromwellId" : "cromid-be06fbc",
  "heartbeatInterval" : "2 minutes",
  "ttl" : "10 minutes",
  "writeBatchSize" : 10000,
  "writeThreshold" : 10000
}
[2018-08-27 02:04:06,71] [info] Metadata summary refreshing every 2 seconds.
[2018-08-27 02:04:06,81] [info] KvWriteActor configured to flush with batch size 200 and process rate 5 seconds.
[2018-08-27 02:04:06,81] [info] WriteMetadataActor configured to flush with batch size 200 and process rate 5 seconds.
[2018-08-27 02:04:06,91] [info] CallCacheWriteActor configured to flush with batch size 100 and process rate 3 seconds.
[2018-08-27 02:04:07,85] [info] JobExecutionTokenDispenser - Distribution rate: 50 per 1 seconds.
[2018-08-27 02:04:07,88] [info] SingleWorkflowRunnerActor: Version 34
[2018-08-27 02:04:07,90] [info] SingleWorkflowRunnerActor: Submitting workflow
[2018-08-27 02:04:07,91] [info] PAPIQueryManager Running with 3 workers
[2018-08-27 02:04:07,91] [info] JES batch polling interval is 33333 milliseconds
[2018-08-27 02:04:07,92] [info] JES batch polling interval is 33333 milliseconds
[2018-08-27 02:04:07,93] [info] JES batch polling interval is 33333 milliseconds
[2018-08-27 02:04:07,98] [info] Unspecified type (Unspecified version) workflow 967af8b6-0d68-44c4-b04e-204674333468 submitted
[2018-08-27 02:04:08,05] [info] SingleWorkflowRunnerActor: Workflow submitted 967af8b6-0d68-44c4-b04e-204674333468
[2018-08-27 02:04:08,05] [info] 1 new workflows fetched
[2018-08-27 02:04:08,05] [info] WorkflowManagerActor Starting workflow 967af8b6-0d68-44c4-b04e-204674333468
[2018-08-27 02:04:08,06] [warn] SingleWorkflowRunnerActor: received unexpected message: Done in state RunningSwraData
[2018-08-27 02:04:08,07] [info] WorkflowManagerActor Successfully started WorkflowActor-967af8b6-0d68-44c4-b04e-204674333468
[2018-08-27 02:04:08,07] [info] Retrieved 1 workflows from the WorkflowStoreActor
[2018-08-27 02:04:08,09] [info] WorkflowStoreHeartbeatWriteActor configured to flush with batch size 10000 and process rate 2 minutes.
[2018-08-27 02:04:08,17] [info] MaterializeWorkflowDescriptorActor [967af8b6]: Parsing workflow as WDL draft-2
[2018-08-27 02:04:08,86] [info] MaterializeWorkflowDescriptorActor [967af8b6]: Call-to-Backend assignments: wgbs.flatten_ -> singularity
[2018-08-27 02:04:12,30] [info] WorkflowExecutionActor-967af8b6-0d68-44c4-b04e-204674333468 [967af8b6]: Starting wgbs.flatten_
[2018-08-27 02:04:13,48] [info] DispatchedConfigAsyncJobExecutionActor [967af8b6wgbs.flatten_:NA:1]: 
mkdir -p mapping
cat /home/vanessa/Documents/Dropbox/Code/labs/cherry/pipelines/wgbs-pipeline/cromwell-executions/wgbs/967af8b6-0d68-44c4-b04e-204674333468/call-flatten_/execution/write_lines_8f61fd340a04ccd930e243709dfb1bed.tmp | xargs -I % ln -s % mapping
ls mapping
[2018-08-27 02:04:13,50] [info] DispatchedConfigAsyncJobExecutionActor [967af8b6wgbs.flatten_:NA:1]: executing: chmod u+x /home/vanessa/Documents/Dropbox/Code/labs/cherry/pipelines/wgbs-pipeline/cromwell-executions/wgbs/967af8b6-0d68-44c4-b04e-204674333468/call-flatten_/execution/script && \
singularity \
exec \
gemBS.simg \
/home/vanessa/Documents/Dropbox/Code/labs/cherry/pipelines/wgbs-pipeline/cromwell-executions/wgbs/967af8b6-0d68-44c4-b04e-204674333468/call-flatten_/execution/script &
echo $?
[2018-08-27 02:04:16,88] [info] DispatchedConfigAsyncJobExecutionActor [967af8b6wgbs.flatten_:NA:1]: job id: 0
[2018-08-27 02:04:16,88] [info] DispatchedConfigAsyncJobExecutionActor [967af8b6wgbs.flatten_:NA:1]: Status change from - to Done
[2018-08-27 02:04:19,50] [info] WorkflowExecutionActor-967af8b6-0d68-44c4-b04e-204674333468 [967af8b6]: Workflow wgbs complete. Final Outputs:
{

}
[2018-08-27 02:04:19,53] [info] WorkflowManagerActor WorkflowActor-967af8b6-0d68-44c4-b04e-204674333468 is in a terminal state: WorkflowSucceededState
[2018-08-27 02:04:22,18] [info] SingleWorkflowRunnerActor workflow finished with status 'Succeeded'.
{
  "outputs": {

  },
  "id": "967af8b6-0d68-44c4-b04e-204674333468"
}
[2018-08-27 02:04:26,91] [info] Workflow polling stopped
[2018-08-27 02:04:26,91] [info] Shutting down WorkflowStoreActor - Timeout = 5 seconds
[2018-08-27 02:04:26,92] [info] Aborting all running workflows.
[2018-08-27 02:04:26,92] [info] WorkflowStoreActor stopped
[2018-08-27 02:04:26,93] [info] Shutting down WorkflowLogCopyRouter - Timeout = 5 seconds
[2018-08-27 02:04:26,93] [info] Shutting down JobExecutionTokenDispenser - Timeout = 5 seconds
[2018-08-27 02:04:26,93] [info] JobExecutionTokenDispenser stopped
[2018-08-27 02:04:26,93] [info] Shutting down WorkflowManagerActor - Timeout = 3600 seconds
[2018-08-27 02:04:26,93] [info] WorkflowLogCopyRouter stopped
[2018-08-27 02:04:26,94] [info] WorkflowManagerActor stopped
[2018-08-27 02:04:26,94] [info] WorkflowManagerActor All workflows finished
[2018-08-27 02:04:26,94] [info] Connection pools shut down
[2018-08-27 02:04:26,94] [info] Shutting down SubWorkflowStoreActor - Timeout = 1800 seconds
[2018-08-27 02:04:26,95] [info] Shutting down JobStoreActor - Timeout = 1800 seconds
[2018-08-27 02:04:26,95] [info] Shutting down CallCacheWriteActor - Timeout = 1800 seconds
[2018-08-27 02:04:26,96] [info] CallCacheWriteActor Shutting down: 0 queued messages to process
[2018-08-27 02:04:26,96] [info] Shutting down ServiceRegistryActor - Timeout = 1800 seconds
[2018-08-27 02:04:26,96] [info] WriteMetadataActor Shutting down: 0 queued messages to process
[2018-08-27 02:04:26,96] [info] KvWriteActor Shutting down: 0 queued messages to process
[2018-08-27 02:04:26,96] [info] Shutting down DockerHashActor - Timeout = 1800 seconds
[2018-08-27 02:04:26,96] [info] ServiceRegistryActor stopped
[2018-08-27 02:04:26,96] [info] Shutting down IoProxy - Timeout = 1800 seconds
[2018-08-27 02:04:26,96] [info] SubWorkflowStoreActor stopped
[2018-08-27 02:04:26,96] [info] DockerHashActor stopped
[2018-08-27 02:04:26,97] [info] IoProxy stopped
[2018-08-27 02:04:26,97] [info] JobStoreActor stopped
[2018-08-27 02:04:26,97] [info] CallCacheWriteActor stopped
[2018-08-27 02:04:27,00] [info] Database closed
[2018-08-27 02:04:27,00] [info] Stream materializer shut down
[2018-08-27 02:04:27,06] [info] Automatic shutdown of the async connection
[2018-08-27 02:04:27,06] [info] Gracefully shutdown sentry threads.
[2018-08-27 02:04:27,06] [info] Shutdown finished.
```
I don't see any strings in the outputs above, but I do see files in the "mapping" folder of execution:

```bash
$ tree cromwell-executions/wgbs/e56c969e-e2fe-4ba1-89d5-c12faf3139d4/call-flatten_/execution
cromwell-executions/wgbs/e56c969e-e2fe-4ba1-89d5-c12faf3139d4/call-flatten_/execution
├── mapping
│   ├── flowcell_1_1_1.fastq.gz -> /home/vanessa/Documents/Dropbox/Code/labs/cherry/pipelines/wgbs-pipeline/cromwell-executions/wgbs/e56c969e-e2fe-4ba1-89d5-c12faf3139d4/call-flatten_/inputs/-1629611677/flowcell_1_1_1.fastq.gz
│   └── flowcell_1_1_2.fastq.gz -> /home/vanessa/Documents/Dropbox/Code/labs/cherry/pipelines/wgbs-pipeline/cromwell-executions/wgbs/e56c969e-e2fe-4ba1-89d5-c12faf3139d4/call-flatten_/inputs/-1629611677/flowcell_1_1_2.fastq.gz
├── rc
├── script
├── script.submit
├── stderr
├── stderr.submit
├── stdout
├── stdout.submit
└── write_lines_b474737d24db6afd91c22a47b2b69e4f.tmp

1 directory, 10 files
```
and the stderr files are totally empty, and then the one stdout (without extension) shows those two mapping files:

```
$ cat cromwell-executions/wgbs/e56c969e-e2fe-4ba1-89d5-c12faf3139d4/call-flatten_/execution/stdout
flowcell_1_1_1.fastq.gz
flowcell_1_1_2.fastq.gz
```
Let me know if this looks correct? What you are looking for? Completely off base? I think this will be easier with respect to getting the process ids for singularity instances with the next version, when there (I think) will be better control commands.

Also - any reason to have all capitals vs. lowercase for the backend examples? (e.g. SLURM vs slurm)


geoffjentry:
Hey @vsoch - to the extent that I understand Singularity this is the direction I was picturing.

Can you post your `test.wdl` as well as the inputs and options files?

kshakir:
> Likely we would want to add a general variable to plug in any user specific arguments (e.g., custom binds) and singularity arguments? (e.g., debug?) I don't know if this is "how it's supposed to look"

The easiest way to verify a run is to use one (or more) of the [standard test cases](https://github.com/broadinstitute/cromwell/tree/develop/centaur/src/main/resources/standardTestCases). For example you can try to run 

```
java \
  -Dconfig.file=my.conf \
  -jar cromwell-34.jar \
  run centaur/src/main/resources/standardTestCases/hello/hello.wdl \
  -i centaur/src/main/resources/standardTestCases/hello/hello.inputs
```

The accompanying `.test` file lists the CI expectations of the workflow run, ex: `centaur/src/main/resources/standardTestCases/hello.test`.

> the stderr files are totally empty, and then the one stdout (without extension) shows those two mapping files … Let me know if this looks correct? What you are looking for? Completely off base?

Based on the WDL you linked to, this output looks like what was expected :+1:

> Also - any reason to have all capitals vs. lowercase for the backend examples? (e.g. SLURM vs slurm)

no reasoN.

---

On a related note I personally would love to see cromwell+singularity running under our  CI, so that we could all a) point others at the working example and b) be sure the examples continue to work in the future.

Most Broadies I know are even greener on Singularity than CircleCI, but I would be keen to learn sometime. Google turned up your earlier work on installing (parts-of?) [Singularity on a Travis VM](https://github.com/singularityhub/singularity-ci). That combined with these commented out configs could be a fantastic starting point to getting singularity+cromwell regularly tested together.

For a similar example, with cromwell+TES, here is where that CI script installs and runs `funnel`:

https://github.com/broadinstitute/cromwell/blob/9f33e2a867fe20924e4f24e0cba8774f7d6d3132/src/ci/bin/testCentaurTes.sh#L14-L36

A similar script that installs the singularity binaries plus a small cluster(?) and then uses a working config file to run our Centaur test suite would be amazing for users.

After it's all working, users are being pointed to docs under https://cromwell.readthedocs.io/, such as https://cromwell.readthedocs.io/en/stable/backends/TES/. A similar entry should be added for a working/tested singularity setup.

vsoch:
okay, well I just updated the linked repo to have this "hello" example. I've concluded that singularity makes absolutely no sense as a backend. it's more logical as just a binary that is run (Local) or with other backend (e.g., slurm). See here --> https://github.com/vsoch/wgbs-pipeline/blob/add/singularity/docs/pages/docs/tutorial/getting-started/index.md#step-4-run-test-case for the test case, and then the example wdl I'll just show you here, because it's stupid. It's just running a command, lol.

```
task dinosaur {
  String singularity_container = "shub://vsoch/hello-world"
  output {
    String roar = read_string(stdout())
  }
  command {
    singularity --silent \
    run \
    ${singularity_container}
  }
}

workflow wf_hello {
  call dinosaur
  output {
     dinosaur.roar
  }
}
```
I think trying to use cromwell just makes it a lot **more** complicated. But I guess it's useful / helping users in other ways, so it's worth it to show how to run a singularity container as a command? /shrug.

I don't totally follow what you are saying, to be honest. cromwell running singularity is just the same as cromwell running anything else, and I'm convinced (after this testing) there is no special use case. Docker has a backend and workflow runtime (because you run a docker container, the thing in a workflow) Singularity you are trying to shove into a backend or a workflow, and it has to be one or the other. It's not really making sense as a backend because it's more like an executable. I think before coming up with tools around something, it would be nice to have a clear definition of how singularity even fits in here to be special or different from any other binary.

vsoch:
So basically:

```
singularity run docker://vanessa/pokemon run catch
```
instead of

```
docker run vanessa/pokemon run catch
```
Isn't that just a shuffling of commands? E.g., if the user asks for docker, but it's not found (or incorrect permissions) you could just look for singularity instead, and then fill that pattern. 

It makes sense, and it seems like someone best to implement it would be familiar not with primarily singularity (myself) but with how Cromwell works with Docker, and then I can offer to help when a specific singularity question is asked. Thanks for the clarification!

vsoch:
Just for some interesting history, I've toyed with implementing the same workflow across container technologies, and the commands are eerily similar, here is the workflows / scripts:

https://github.com/sci-f/snakemake.scif

and the command breakdown - check out starting around [these slides](https://docs.google.com/presentation/d/e/2PACX-1vTPzgIfvmDFKA32bGfeWmmYmeujRmfHwLquvcKZp_3gMfkEe5hfOj2cyJ95IYjM944JBlM9bkz9nibq/pub?start=false&loop=false&delayms=3000&slide=id.g34c959ef7d_0_261)

vsoch:
So whomever solves the problem of the rosetta stone for containers, they win :)

vsoch:
what is "User Joy" ?

vsoch:
Thanks @gemmalam... will let you know when I run into that user joy... all I've found thus far are flying pigs :joy: 

@leepc12 and I had discussion yesterday, and I've updated the recipe to be more general, meaning that it would allow the pipeline creator to set (general) strings of singularity commands (e.g., `--debug --vvv`) or the command itself (`exec` vs `run`) and then the command options (`--bind /host:/container --cleanenv`) etc. This will work to run a docker:// uri for a singularity container, or a file path, which I understand to be a use case that is wanted? I think providing this example is a good solution (for the time being) until Singularity can be added as a backend proper.

geoffjentry:
Hi @vsoch - it appears that @rhpvorderman and @davycats have managed to get the use case of "run a standard docker container using Singularity" working for them (as per #4088 and #4084). I know you gave that a whirl and weren't able to make it work, I wonder if one of them can share what they're doing and perhaps see if you can confirm it's working?

DavyCats:
@vsoch @geoffjentry We did manage to get it working, with some caveats. We also haven't really tested it very extensively yet.
These are the relevant lines from the backend configuration:
```
submit-docker = """
        echo ' \
        CROMWELLROOT=$(echo ${cwd} | sed "s/cromwell-executions\\/.*/cromwell-executions/") && \
        sed -i "s/\\/exports\\//\\/data\\//g" ${cwd}/execution/script && \
        chmod 775 ${cwd}/execution/script && \
        singularity exec --bind /exports:/data/,$CROMWELLROOT:/config docker://${docker} ${script}' | \
        qsub \
            -terse \
            -V \
            -b n \
            -wd ${cwd} \
            -N ${job_name} \
            ${'-pe BWA ' + cpu} \
            ${'-l h_vmem=' + memory + "G"} \
            -o ${cwd}/execution/stdout \
            -e ${cwd}/execution/stderr
        """
        dockerRoot = "/config"
```
> This only works if your container has both a /data and /config mount point. I tested this (very shallowly) using biocontainers.

Line by line:
1.  `CROMWELLROOT=$(echo ${cwd} | sed "s/cromwell-executions\\/.*/cromwell-executions/")` 
    1. If dockerRoot is `/cromwell-executions`
    2. The script will contains paths like: `/cromwell-executions/test/<hash>/call-task/execution/rc`
    3. Therefore we need to have the entire structure under the root of the execution folder mounted, as such, we need to bind the entire execution folder.
    4. This gets the path to the root of the execution folder.
    - I also tried setting dockerRoot to be the same as `cwd`: `dockerRoot = "${cwd}"`, but this resulted in `${cwd}` being placed literally in the execution script. If this had been an option we wouldn't have to bind the execution directory separately (I think), but since it isn't we do have to do so.
1. `sed -i "s/\\/exports\\//\\/data\\//g" ${cwd}/execution/script` 
    - This a bit of a nasty workaround to convert absolute paths used in the commands to what their path would be in the container. This is necessary if you have (eg.) a String type output directory in a command. There are other ways of dealing with this, you could make a /data directory which links to /exports for example.
1. `chmod 775 ${cwd}/execution/script`
    - Make the script executable.
1. `singularity exec --bind /exports:/data/,$CROMWELLROOT:/config docker://${docker} ${script}'`
    - Run the script in singularity, binding the cromwell execution folder to /config and /exports to /data.

DavyCats:
@vsoch @geoffjentry I just wanted to come back to this since singularity 3.0.1 was released a few weeks ago. The backend configuration can now be made a lot more simplistic:
```
submit-docker = """
    echo ' \
    singularity exec --bind /run,/exports,${cwd}:${docker_cwd} docker://${docker} bash ${script}' | \
    qsub \
        -terse \
        -V \
        -b n \
        -wd ${cwd} \
        -N ${job_name} \
        ${'-pe BWA ' + cpu} \
        ${'-l h_vmem=' + memory + "G"} \
        -o ${cwd}/execution/stdout \
        -e ${cwd}/execution/stderr
    """
```
The bind to `/run` was neccessary on our SGE cluster to make python multiprocessing work, as in [this issue](https://github.com/sylabs/singularity/issues/455). The bind to `/exports` is also specific to our cluster.

vsoch:
Just curious, why are you giving the qsub command to singularity? You might consider launching a job that runs singularity, not the other way around. If the above works for you then ok, but it seems likely to lead to error to me. If you use this method you will likely find you don't need all of those binds.

The need for multiprocessing with python was also a requirement that was pre-3.0 - after 3.0 python is totally removed. So if you are adding extra binds for the benefit of the Singularity python, you definitely don't need to do this.

DavyCats:
@vsoch Actually I'm piping the singularity command to qsub, so I'm actually doing exactly what you're suggesting :stuck_out_tongue:  (Notice the `echo ' \` and `' | \` surrounding the singularity command.)

For the binds: `/exports` is needed because output is written to this directory and the `${cwd}:${docker_cwd}` is needed because the script under `${script}` (generated by cromwell) uses this path.

As for the multiprocessing thing, I get an error: `OSError: [Errno 30] Read-only file system` when I try to run a python script using multiprocessing inside a container. This seems to be caused by the fact that `/dev/shm` is a symlink to `/run/shm` (as in that issue I linked), so it can't be found unless I mount the `/run` directory. So singularity works just fine without it, but this is needed for certain python based programs. This is also specific to the cluster, as on my own pc I don't have this problem.

vsoch:
Cool, then you are probably good to go :) Did you want feedback on something in particular? I don't know why you pinged me but I'm glad to see you have a new command!

DavyCats:
I figured it might be useful information for you since geoffjentry had asked us to share with you how we had configured this before. If it's not, please excuse the spam :sweat_smile:

vsoch:
Oh ok great!! Yes thank you, apologies I take things very literally and miss the subtle social details sometimes. This will be good as a reference for others using Cromwell that want to achieve similar.

geoffjentry:
Hi - thanks to both of you. In this case my procrastination paid off, I've been meaning to find a way to work this example into something like the examples conf or the docs, so having this updated example is great.

TMiguelT:
Can this be merged?

geoffjentry:
@TMiguelT It's been on my todo list for ages (well, since 11/27 apparently) to loop back around and make a couple of small modifications prior to merging, but it's not been a high priority.

As this is a purely documentation PR it's not preventing any functionality - are you looking to use Singularity? And if so have you tried this, could you verify that it did/did not work for you?

TMiguelT:
Yes, I am looking to use singularity (my guess is that lots of others are as well!). I'm currently playing around with udocker and singularity with Cromwell on HPC. If you'd like, I can report my findings here or somewhere else

geoffjentry:
For Singularity it'd be super helpful to post it here just so that the information is in one place. Also, it'll serve a dual purpose of letting me ~procrastinate~ let this mature a bit more as well as give me a good ~kick in the backside~ reminder to finish this off once it's done.

For udocker IIRC we don't have explicit docs like this but there definitely **are** configs floating around which have worked for most, but not all, people. If my statement is correct and you haven't dug one up let me know and I can send one to you, we can go from there.

TMiguelT:
If you can find an example config that would be very helpful. It should be almost as simple as tweaking the `submit-docker` flag in the config to use `udocker`, no? 

I have a feeling that `udocker` will be easier to use, and will be a better recommendation for Cromwell than directly using `singularity`, because it provides a very docker-like CLI that should make the switch fairly painless.

geoffjentry:
@TMiguelT Will follow up w/ config in email just to not muddy the waters on this issue. I haven't had enough first hand experience with either to opine here, unfortunately.

illusional:
@TMiguelT @geoffjentry I've been following the conversation and we're pretty keen to use some container system with Cromwell on our cluster. At the moment I'm trying to use udocker with Cromwell with the following conf, but the docker param [is looked up](https://cromwell.readthedocs.io/en/stable/cromwell_features/CallCaching/#Docker+Tags) and injected as a [digest](https://docs.docker.com/engine/reference/commandline/pull/#pull-an-image-by-digest-immutable-identifier) which udocker [doesn't appear to support](https://github.com/indigo-dc/udocker/issues/112). 

```
backend {
    default: udocker
    providers: {
        udocker {
            # The backend custom configuration.

            config {

                # The list of possible runtime custom attributes.
                runtime-attributes = """
                String? docker
                String? docker_user
                """

                # Submit string when there is a "docker" runtime attribute.
                submit-docker = """
                udocker run \
                  --rm -i \
                  ${"--user " + docker_user} \
                  --entrypoint ${job_shell} \
                  -v ${cwd}:${docker_cwd} \
                  ${docker} ${script}
                """
            }
        }
    }
}
```

which results in the script.submit:
```
udocker run \
  --rm -i \
  --entrypoint /bin/bash \
  -v /path/to/call-untar:/cromwell-executions/path/to/call-untar \
  ubuntu@sha256:868fd30a0e47b8d8ac485df174795b5e2fe8a6c8f056cc707b232d65b8a1ab68 \
  /cromwell-executions/path/to/call-untar/execution/script
```

and fails with the error:
```
Error: invalid repo name syntax
Error: must specify image:tag or repository/image:tag
```

I can't find some way to disable the docker lookup by Cromwell, nor some non-digest runtime variable that Cromwell exposes. Just wondering how you're achieving this on docker or singularity.

vsoch:
Based on the documentation it (at least seems like) there are ways to get this working. For example:

> if the docker image uses a hash, all call caching settings apply normally

You could try a full uri WITH a hash, e.g,.

```
library/ubuntu:latest@sha256:868fd30a0e47b8d8ac485df174795b5e2fe8a6c8f056cc707b232d65b8a1ab68
```
Then in the case that you provide a tag (e.g., library/ubuntu:14.04):

> Cromwell will attempt to look up the immutable digest of the image with this floating tag. Upon success it will pass both the floating tag and this digest value to the backend.

So if you try providing the example above, does it still reduce it to `ubuntu@sha256:868fd30a0e47b8d8ac485df174795b5e2fe8a6c8f056cc707b232d65b8a1ab68`?

> If Cromwell fails to lookup the digest (for instance an unsupported docker registry, wrong credentials,  it will run the job with the user provided floating tag.

So couldn't you just give it something you know will fail, and then have it use the tag you did provide?

If you wanted udocker to work, it seems more like a bug that cromwell is considering an incomplete uri (just ubuntu and the hash) as "the correct way." That tells us nothing about the registry, the start of the namespace (you could consider this like the collection, library) or a tag. Minimally the entire namespace should be honored, and the hash can still be looked up.

illusional:
I tried to run it through with a few different broken tags, but Cromwell doesn't parse it and just passes everything, ie:
- `ubuntu:latest` → `ubuntu@sha256:868fd30a0e47b8d8ac485df174795b5e2fe8a6c8f056cc707b232d65b8a1ab68`
- `library/ubuntu@sha256:868fd30a0e47b8d8ac485df174795b5e2fe8a6c8f056cc707b232d65b8a1ab68` → `<SAME>`
- `ubuntu@sha256:868fd30a0e47b8d8ac485df174795b5e2fe8a6c8f056cc707b232d65b8a1ab68` → `<SAME>`
- `ubuntu@sha256:868fd30a0e47b8d8ac485df174795b5e2fe8a6c8f056cc707b232d65b8a1ab68̶ ` → `<SAME>`

Unfortunately none of which udocker supports. I don't think it feels misplaced to mark this as a deficiency in udocker given the general trend towards reproducibility. I could somewhat get this working by changing the .conf to include some other runtime var (eg called udocker) and including a docker and udocker runtime var, though udocker doesn't seem to correctly signal back to say the job has finished.

vsoch:
Not `library/ubuntu:latest@sha256:868fd30a0e47b8d8ac485df174795b5e2fe8a6c8f056cc707b232d65b8a1ab68` ?

vsoch:
Agreed! I would check with the udocker folks about what's up :)

TMiguelT:
Hi all, I encountered this bug myself, which led me to this issue: https://github.com/indigo-dc/udocker/issues/112. In other words, its a known issue but it doesn't look like it will be solved in the immediate future. I was just going to wait until I had finished my investigation to chat about this but currently I think singularity is looking more promising.

geoffjentry:
@illusional Ah interesting. That makes sense, we made some changes to how docker is handled internally which postdate that udocker project (which was a one off proof of concept). 

As you intimated this behavior on Cromwell's part ties back to reproducibility, but also the question of "what happens if a container changes in the middle of a workflow?". As an example, imagine there is a scatter job of 1000 identical tasks, all calling `ubuntu:latest`. And suppose that after processing 750/1000, `latest` is changed. Since we already were capturing the info for provenance we also decided to lock in a hash throughout a workflow.

I have a vague memory that you might be able to disable this behavior but I can't find it at the moment. I'll ask the team tomorrow.

geoffjentry:
@illusional Setting `docker.hash-lookup.enabled` to `false` **might** work here in terms of the docker hash translations going on. If it does, it'll come with the caveat that call caching will not work for any container using a floating tag

illusional:
I had a bit of troubles with getting this working, but wanted to report back and hopefully someone can learn from my misunderstandings. Just skip to the results for the tl;dr.

**Goal**: Find a suitable replacement (preferably drop in) to Docker.

The two main contenders were Singularity and udocker.

When you run a docker container through Cromwell, it runs a special submit script that gets the [container id](https://docs.docker.com/engine/reference/commandline/run/#capture-container-id---cidfile) through `--dockercid expected/path/to/dockercid`, _waits_ for the docker container to finish and then manually moves the `rc` file to the expected location.

### My misunderstanding

This docker script erroneously led me to believe that Cromwell needs job identifier and it uses some mechanism to continually check the status of that job. But in fact Cromwell doesn't really poll the workload manager, but my understanding from #1499 is that it actively polls the filesystem, looking  for the `rc` file within the execution directory (potentially `stdout` too if its looking for the job id). This is also logically verified by looking at the `script` file that Cromwell generates, the way it collects the return code and places it in the expected directory.

- Based on my initial incorrect understanding, I then believed if I couldn't get the containerId, then how would Cromwell know that the job has actually finished.

### My additional errors

These tended to redirect me away from the actual misunderstanding. 

1. In the udocker config I [posted before](https://github.com/broadinstitute/cromwell/pull/4039#issuecomment-454569364) (corrected with comment), I was accidentally providing the `--entry-point` param, which was causing an interactive shell to open, stopping the rest of the script from executing.

2. I failed to notice that `${out}` and `${err}` change between `submit` and `submit-docker`. When I would check the job that Cromwell schedules through SLURM, it would always fail. But I'm fairly sure that the job was failing to start because it was trying to write stdout to `/cromwell-executions/.../execution/stdout`, this is what led me to #1499.

3. An easy fix, but if your backend doesn't export a job-id, you need to set `run-in-background = true` in that backend's config.

### The results

The following execution strings can be inserted into the two container configs:
- Singularity: `singularity exec --bind ${cwd}:${docker_cwd} docker://${docker} ${job_shell} ${script}`
- udocker: `udocker run ${"--user " + docker_user} --rm -v ${cwd}:${docker_cwd} ${docker} ${script}`

My _container_ config template for no workflow manager:
```HOCON

# uncomment if using udocker
# docker.hash-lookup.enabled = false

backend {
    default: singularity
    providers: {
        singularity {
            # The backend custom configuration.

            config {
                run-in-background = true
                # The list of possible runtime custom attributes.
                runtime-attributes = """
                String? docker
                String? docker_user
                """
                # Submit string when there is a "docker" runtime attribute.
                submit-docker = """
                    ## PLACE THE CORRECT CONTAINER COMMAND HERE ##
                """
            }
        }
    }
}
```

And applied for something like SLURM:
```HOCON

# uncomment if using udocker
# docker.hash-lookup.enabled = false

backend {
    default: SLURM
    providers: {
        SLURM {
          config {
            runtime-attributes = """
            Int runtime_minutes = 600
            Int cpus = 2
            Int requested_memory_mb_per_core = 8000
            String queue?
            String? docker
            String? docker_user
            """
            # you should have a submit script as well, 
            submit-docker = """
                sbatch -J ${job_name} -D ${cwd} -o ${cwd}/execution/stdout -e ${cwd}/execution/stderr ${"-p " + queue} \
                    -t ${runtime_minutes} ${"-c " + cpus} --mem-per-cpu=${requested_memory_mb_per_core} \
                    --wrap "## PLACE THE CORRECT CONTAINER COMMAND HERE ##"
            """
            kill = "scancel ${job_id}"
          }
        }
    }
}
```

Thanks everyone for the comments above.

TMiguelT:
Thanks @illusional, I've come to a very similar configuration. I ended up settling on this:

```
        submit-docker = """
            export SINGULARITY_CACHEDIR=/some/cache/dir
            IMAGE=/path/to/image/storage/${docker}
            singularity build --sandbox $IMAGE docker://${docker}
            sbatch -J ${job_name} -D ${cwd} -o ${cwd}/execution/stdout -e ${cwd}/execution/stderr -t ${runtime_minutes} \
            ${"-c " + cpus} \
            --mem-per-cpu=${requested_memory_mb_per_core} \
            --wrap "singularity exec --userns -B ${cwd}:${docker_cwd} $IMAGE ${job_shell} ${script}"
        """
```

Just two things I'd like to discuss.

Firstly, because you are pulling the docker image inside the sbatch script, this depends on the cluster you're working on allowing network access for the workers. While that is possible on our local cluster, my discussion with some sysadmins made me realise that this wasn't necessarily commonplace, and even on our cluster they strongly discouraged me from relying too heavily on it. This made me look for a solution that was even more generalizable. This is why I `singularity build` the image before I submit it, using the head node. This ensures that all network-requiring work is done on the head node, where network access is guaranteed. I also make sure to set a cache directory, so we don't download the same docker image multiple times in the case of a scatter job etc. Of course, if you do have network access for your workers and the admins have no issue with you using it, pulling the image from the worker is probably a better option to avoid hogging the head node.

The second main difference in my config is that the singularity binary I was using did not have `setuid` permissions, meaning that I had to use the sandbox format, and run the image using `--userns`. This is obviously only required if your sysadmins don't trust `singularity`, but I think it's important to demonstrate a way of running containers without *any* privileges at all.

@geoffjentry all this discussion is obviously going way beyond this original PR. Once we've settled on our recommendations, how do you think we should share this information with the Cromwell community? Is an example config in the Cromwell repo the best way (like this PR), or would it serve better to have a new page in the Cromwell documentation? I'm sure that I (and @illusional if he is able) would be happy to write this up.

vsoch:
Based on the need for this use case, I strongly support merging this (or some) basic configuration, and then having documentation with a writeup about these different use cases. @TMiguelT you are spot on about creating the image before hand, and don't forget that in addition to build (which most users might not be able to do on their cluster) you can also just pull from docker:

```bash
singularity pull docker://<container>
```
and then create a binary (singularity image) that can be given directly to the run/exec command.

TMiguelT:
Would you mind explaining the difference between `pull` and `build`? The reason I did build is because I needed to know where the output image ended up, so I could run it directly. If I `singularity pull docker://ubuntu`'d the image, and then `singularity run docker://ubuntu` from the worker, it would still try to pull the image a second time, and then hang forever because it didn't have network access. 

Also, isn't the ability to build a binary image something that `build` can do, not `pull`?. The only reason I built a sandbox instead is simply because my admins wouldn't set the `setuid` bit, so it wouldn't work. Am I missing something here?

vsoch:
Good question! In the context of the docker://uri, pull and build are almost identical. Both will retrieve layers from Docker Hub (or a registry you target), tar.gz files, and dump them into a binary image. So these two commands do the same thing:

```bash
singularity build docker://<container>
singularity pull docker://<container>
```
Build is different if you do it in an environment where you can use sudo, because you can target a build recipe instead, e.g.,

```bash
sudo singularity build <container>.sif Singularity
```
and pull is different too if you target some non-docker pull source:

```bash
singularity pull shub://vsoch/singularity-images
```
You can look at the codebase to confirm this - a pull of a docker uri [comes down to calling build](https://github.com/sylabs/singularity/blob/03072a88e108966d50fd61b0e6a51e6dbc62ff20/internal/pkg/libexec/pull.go#L42)

```bash
func PullOciImage(path, uri string, opts types.Options) {
	b, err := build.NewBuild(uri, path, "sif", "", "", opts)
	if err != nil {
		sylog.Fatalf("Unable to pull %v: %v", uri, err)
	}

	if err := b.Full(); err != nil {
		sylog.Fatalf("Unable to pull %v: %v", uri, err)
	}
}
```
so choose whichever word you like better :) In the context of docker, without sudo, they do the same thing.

vsoch:
@TMiguelT you could just use --name to determine where the outcome of a pull ends up, combined with what you are using for `SINGLARITY_CACHEDIR`.

geoffjentry:
@illusional @vsoch @TMiguelT This is amazing, thank you all so much. I'm about to head out of town for a week but agree with Vanessa that documenting this makes sense, and will do so (no later than) when I get back.

vsoch:
Bon voyage, @geoffjentry !

TMiguelT:
Thanks for that explanation, @vsoch. Based on that, I still can't find a compelling reason to use `pull`, since it's still important that I'm able to run the generated image file directly, and this seems a tad easier with `build`.

vsoch:
giphy

TMiguelT:
I'm writing up a documentation PR, but I realised that the ideal spot for it ([here](https://cromwell.readthedocs.io/en/develop/tutorials/HPCIntro/)) is targeted towards PBS/TORQUE. Are there any torque users here that might be able to test out a config for us before I submit my PR?

vsoch:
@TMiguelT I would argue that Singularity has use cases beyond HPC, and that it would be much cleaner (and better long term) to have a "Getting started with Singularity" tab. I would bet there are more users that would be interesting in using Singularity that don't care about PBS / TORQUE, and so it doesn't make sense to tangle the two.

illusional:
Why not “Using containers with Cromwell”? Have a section for singularity (and undocker), and have a subsection of that for singularity with job schedulers?

There are other Cromwell github threads that have the user goal as practically: I can’t use docker, what do I do? That’s definitely a part of my use case, and reproducibility is a big point as well.

vsoch:
I like that idea @illusional, and I say go for it! Let me know if you want me to look over a draft or the like.

illusional:
I've started a new PR #4635  for some documentation, and @TMiguelT will hopefully be able to contribute to that. If I've got questions, I'll ping you on there to keep it a little more isolated.

geoffjentry:
@vsoch I believe this work is fully superseded by all of the hard work in #4635 so I'm closing this. If you disagree, just let me know and we'll figure out what needs to be done.

vsoch:
I disagree. #4635 is just documentation. This was a full (code) example in cromwell.examples.conf. They are different.

geoffjentry:
@vsoch That's a fair point. I was thinking that as there were viable config blocks in the new documentation that perhaps it's not necessarily, but I think we want **something** in `cromwell.examples.conf`. Thinking about it a little bit tho it winds up coming back to an issue I kept punting where I didn't want to add a bunch of variants for the different common situations. 

Would it make sense to you if there was a comment block in there pointing people to the tutorial for examples? If so I'm happy to do that.

If you think it's best to leave the concrete example(s), could you please confirm that they're consistent with the config files you all hashed out in #4635?

vsoch:
Is the cromwell.examples.conf biased to not include full examples for its clients? Having lots of experience reading documentation vs. getting code, my general preference is to find a complete config, copy paste, and customize it. Pointing the user to the docs is okay as long as the full configuration is there (not broken into sections.). On the other hand, I agree it's bad to confuse the user with too many options. My preference would be for consistency. If you have the examples file, you should provide all examples there, as this would be the expectation. If you decide to link to docs, this should be standard for other types, as the two could get disconnected (i.e. different versions across locations).

vsoch:
To be totally honest, [this file](https://github.com/broadinstitute/cromwell/blob/develop/cromwell.examples.conf) is a total config from my darkest of nightmares. If a user opens that and needs to figure out what to do? I'd guess they would be overwhelmed. What about a folder of example configuration files instead? Eg.

```
examples/
    cromwell.singularity.conf
    cromwell.docker.conf
    cromwell.slurm.singularity.conf
    cromwell.slurm.conf
 ....
```
At least that way it's not config overload. Thoughts?

TMiguelT:
I think I agree with @vsoch on this. For now, I think our PR (#4635) should probably also have a change to `cromwell.examples.conf` so that all the types of config are in the same place. In the long term, they should probably be split out into separate files in a directory as she suggested.

That said, I would also ideally like a file or document that explains *every* possible config option that Cromwell understands. At the moment `cromwell.examples.conf` serves this role (in a confusing way), but we would need a replacement if it's split out into separate files.

TMiguelT:
Oh, actually now that that PR has been merged I suppose it's too late to edit the example config file. It can be fixed in whichever PR reworks this config file, I guess

vsoch:
@geoffjentry I'll open an issue for this, and please assign me to it.

geoffjentry:
@vsoch if you've not already done that, we can always reopen this PR.

You all are right, I'm wrong.

The reason why I was leaning towards avoiding `cromwell.examples.conf` blurbs is that (as @TMiguelT points out) it's a bit of a mess right now due to having too much stuff in there. TBH work needs to be done to start making that more organized. I sometimes can lean towards throwing the baby out with the bathwater in circumstances like that.

I think it's fine to put a number of configurations into `cromwell.examples.conf` as long as the full block is fairly well self contained, and well documented. IOW something which would be easy to peel out into a separate file if/when we get there.

vsoch:
No worries about this PR, I think it's more correct to address the larger issue (and I'm already working on it). I'll have something to get us started soon!

vsoch:
Everyone in this conversation please see -> https://github.com/broadinstitute/cromwell/pull/4697/files We will need feedback / review of the various files to make sure they are good.

yinbinqiu:
> 我们的团队为 SGE 提供了类似的 Singularity 和/ENCODE [。](https://github.com/ENCODE-DCC/atac-seq-pipeline/blob/master/backends/backend.conf)

Hello,this link is invalid

PR #4047, opened 2018-08-28
Centaur AWS tests

ruchim:

PR #4053, opened 2018-08-30
Enable CI scripts to run only included centaur tests.

kshakir:


kshakir:
Might be helpful for #4047: include just a few tests instead of excluding dozens.

centaur/test_cromwell.sh
```diff
@@ -26,10 +26,11 @@ Arguments:
     -c    If supplied, the config file to pass to Cromwell
     -n    If supplied, the config file to pass to Centaur
     -t    If supplied, the timeout for request-plus-response from Centaur to Cromwell
+    -i    If supplied, will include tests with this tag
     -e    If supplied, will exclude tests with this tag
     -s    If supplied, will run only the specified suite
     -p    If supplied, number of tests to be run in parallel. 16 is the default
-    -i    If supplied, will run the tests in this directory instead of the standard tests
+    -d    If supplied, will run the tests in this directory instead of the standard tests
```
mcovarr: Are all users of this script aware of this change?

ruchim: Do you mean people outside of Batch? afaik, we're the only ones who use this script, is that right?

mcovarr: I thought greenies used some of this stuff but maybe I'm wrong or at least not this...

ruchim: Asked in the Emerald Empire slack -- looks all clear

Issue #4051, opened 2018-08-29
Tabs whtitespace does not work


ffinfo:
When I try to use tabs as whitespace in cromwell I'm getting a error:
```
[2018-08-29 09:25:20,10] [error] WorkflowManagerActor Workflow 1ed0e19c-fa18-4241-8f6b-0b72e181f59a failed (during MaterializingWorkflowDescriptorState): cromwell.engine.workflow.lifecycle.materialization.MaterializeWorkflowDescriptorActor$$anon$1: Workflow input processing failed:
Boxed Error
```

The specs say this should be possible, 0x9 == tab:
https://github.com/openwdl/wdl/blob/master/versions/1.0/SPEC.md#whitespace-strings-identifiers-constants

I have tested cromwell 34 and the current develop branch (ce27a93).

wdl file that I used (replaced '\t' with '\<tab\>')
```
version 1.0

workflow Test {
<tab>input {
<tab><tab>
<tab>}

<tab>call Echo as echo {
<tab>input:
<tab>}

<tab>output {
<tab>}
}

task Echo {
<tab>input {
<tab>}

<tab>command {
<tab><tab>kill -9 $$
<tab><tab>echo test
<tab>}

<tab>output {
<tab>}
}
```

Full stacktrace:
```
[2018-08-29 09:25:20,10] [error] WorkflowManagerActor Workflow 1ed0e19c-fa18-4241-8f6b-0b72e181f59a failed (during MaterializingWorkflowDescriptorState): cromwell.engine.workflow.lifecycle.materialization.MaterializeWorkflowDescriptorActor$$anon$1: Workflow input processing failed:
Boxed Error
scala.concurrent.impl.Promise$.resolver(Promise.scala:83)
scala.concurrent.impl.Promise$.scala$concurrent$impl$Promise$$resolveTry(Promise.scala:75)
scala.concurrent.impl.Promise$DefaultPromise.tryComplete(Promise.scala:280)
scala.concurrent.Promise.complete(Promise.scala:49)
scala.concurrent.Promise.complete$(Promise.scala:48)
scala.concurrent.impl.Promise$DefaultPromise.complete(Promise.scala:183)
scala.concurrent.Promise.failure(Promise.scala:100)
scala.concurrent.Promise.failure$(Promise.scala:100)
scala.concurrent.impl.Promise$DefaultPromise.failure(Promise.scala:183)
cats.effect.IO.$anonfun$unsafeToFuture$2(IO.scala:341)
[...]

PR #4059, opened 2018-08-30
Add tabs to one Centaur case

aednichols:
Related to (but does not close) https://github.com/broadinstitute/cromwell/issues/4051

In the issue above, doubt was raised whether we correctly support tabs. I wanted to test whether we do and add a test, and this change does both.


aednichols:
(We definitely do have other workflows with tabs, but seemingly none at version 1.0)

geoffjentry:
👍

PR #4071, opened 2018-09-05
Count subworkflows against the max concurrent workflows limit.

mcovarr:


geoffjentry:
How is this being tested?

mcovarr:
Per discussion with Ruchi going to try to do this as a unit test, probably in more than one piece. The other possibilities are 🤕

geoffjentry:
As a separate PR, in some future work, or is this just an attempt to draw out the shorebirds?

mcovarr:
I'm in the "thinking about it" stage now, not sure where it will go.

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -135,9 +136,14 @@ class WorkflowManagerActor(params: WorkflowManagerActorParams)
   override def preStart(): Unit = {
     // Starts the workflow polling cycle
     timers.startSingleTimer(RetrieveNewWorkflowsKey, RetrieveNewWorkflows, Duration.Zero)
+    // Listen on subworkflow start and end events to inform our decision whether to pick up new root workflows
+    // from the workflow store.
+    context.system.eventStream.subscribe(self, classOf[SubWorkflowStart])
+    context.system.eventStream.subscribe(self, classOf[SubWorkflowEnd])
```
Horneth: I wish we had used this `eventStream` stuff from the beginning for things like the service registry actor instead of wiring it through every single Cromwell actor -_-

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/SubWorkflowExecutionActor.scala
```diff
@@ -142,6 +142,7 @@ class SubWorkflowExecutionActor(key: SubWorkflowKey,
         case Some(id) =>
           pushWorkflowEnd(id)
           pushExecutionEventsToMetadataService(key, eventList)
+          context.system.eventStream.publish(SubWorkflowEnd(id))
```
Horneth: This seems to be the happy path but if for some reason this actor blows up we might not end up here which would forever mess up the sub workflow count. Maybe the WMA could watch the SWEA when it receives the Start event, and decrement then ?
If you do end up going the unit test route, killing this actor on purpose would be something worth testing too I think

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -239,8 +244,14 @@ class WorkflowManagerActor(params: WorkflowManagerActorParams)
       timers.cancel(RetrieveNewWorkflowsKey)
       sender() ! akka.Done
       goto(RunningAndNotStartingNewWorkflows)
+    case Event(SubWorkflowStart(actorRef), data) =>
+      // Watch for this subworkflow to expire to remove it from the Set of running subworkflows.
+      context.watch(actorRef)
+      stay using data.copy(subWorkflows = data.subWorkflows + actorRef)
+    case Event(Terminated(actorRef), data) =>
+      stay using data.copy(subWorkflows = data.subWorkflows - actorRef)
```
Horneth: I don't _think_ it matters that we go here even if the terminated actor is not a subworkflow actor (as removing it will do nothing), but it could deserve a comment explicitly stating it, as the goal here is really to count dead subworkflow actors but we're doing it for any dead actor that we're watching.

server/src/test/scala/cromwell/engine/WorkflowManagerActorSpec.scala
```diff
@@ -18,5 +25,89 @@ class WorkflowManagerActorSpec extends CromwellTestKitWordSpec with WorkflowDesc
...
+      val firstWorkflowId = rootActor.underlyingActor.submitWorkflow(firstSources)
+      waitForState(firstWorkflowId, WorkflowRunning)
+      // Give the subworkflows plenty of time to start running.
+      Thread.sleep(10 * 1000L)
+
+      // Submit a second workflow.
+      val secondSources = SubWorkflows(naptime = 0 seconds).asWorkflowSources()
+      val secondWorkflowId = rootActor.underlyingActor.submitWorkflow(secondSources)
+      // Wait a long time so that there would have been many root workflow pickup sweeps.
+      Thread.sleep(30 * 1000L)
```
Horneth: ToL: I think we could also check that the workflow store actor is receiving a `FetchRunnableWorkflows` message with a `0` value. But now that I'm writing this I think your way is more black box-y which is probably better.

PR #4073, opened 2018-09-06
Count I/O retries

Horneth:


core/src/main/scala/cromwell/core/io/IoAck.scala
```diff
@@ -18,6 +18,7 @@ sealed trait IoAck[T] {
 case class IoSuccess[T](command: IoCommand[T], result: T) extends IoAck[T] {
   override def toTry = Success(result)
 }
+
```
kshakir: Revert?

core/src/main/scala/cromwell/core/io/IoCommand.scala
```diff
@@ -28,12 +28,12 @@ trait IoCommand[+T] {
     * Completes the command successfully
     * @return a message to be sent back to the sender, if needed
     */
-  def success[S >: T](value: S): IoSuccess[S] = IoSuccess(this, value)
+  def success[TT >: T](value: TT): IoSuccess[TT] = IoSuccess(this, value)
```
kshakir: Curious, why the anti-`S` bias in this file? IMHO if we're making changes to this file, the trait should switch to `IoCommand[+A]`, and this should should go to `B >: A`.

core/src/main/scala/cromwell/core/retry/IORetry.scala
```diff
@@ -0,0 +1,52 @@
...
+                      isFatal: Throwable => Boolean = throwableToFalse,
+                      onRetry: (Throwable, S) => S = noOpOnRetry)
+                     (implicit timer: Timer[IO], statefulIoException: StatefulIoError[S]): IO[A] = {
+    lazy val delay = backoff.backoffMillis.millis
+
+    def fail(throwable: Throwable) = IO.raiseError(statefulIoException.toThrowable(state, throwable))
+
+    io handleErrorWith {
+      case throwable if isFatal(throwable) => fail(throwable)
+      case throwable =>
```
kshakir: Stop catching `java.lang.Error`.

```scala
io handleErrorWith {
      case exception: Exception if isFatal(exception) => fail(exception)
      case exception: Exception => ...
      case throwable => throw throwable
}
```

We should never be trying to do extra biz logic on fatal JVM errors such as `java.lang.OutOfMemoryError` etc. Also change the contracts/signatures to use `java.lang.Exception`, for example `isFatal: Exception => Boolean`.

core/src/main/scala/cromwell/core/retry/IORetry.scala
```diff
@@ -0,0 +1,52 @@
...
+
+  /**
+    * Same as Retry.withRetry using cats.effect.IO
+    * The state is being passed down upon each retry. It can be use to accumulate information as IO operations
+    * fail and are being retried.
+    */
+  def withRetry[A, S](io: IO[A],
+                      state: S,
+                      maxRetries: Option[Int],
+                      backoff: Backoff = SimpleExponentialBackoff(5.seconds, 10.seconds, 1.1D),
```
kshakir: TIL: [`backoff` is actually one word not two](https://en.wikipedia.org/wiki/Exponential_backoff) and is [misspelled in google's libs](https://github.com/google/google-http-java-client/blob/1.23.0/google-http-client/src/main/java/com/google/api/client/util/BackOffUtils.java).

core/src/main/scala/cromwell/core/io/IoCommand.scala
Horneth: Ah yes, that was a leftover from an earlier version. I'll revert

core/src/main/scala/cromwell/core/retry/IORetry.scala
kshakir: For posterity:

There's a bit of debate on the internet. Different camps agree that one shouldn't catch all `java.lang.Throwable`. However a scala-based camp prefers the use of `NonFatal` and a java-based camp is all about checked-vs-unchecked `Exception` handling.

I don't know what's "right" at the moment, but will accept any patches that don't catch OOME among others.

engine/src/main/scala/cromwell/engine/io/IoState.scala
```diff
@@ -0,0 +1,34 @@
...
+      EnhancedCromwellIoException(state, throwable)
+    }
+  }
+
+  val updateState: (Throwable, IoState) => IoState = (throwable, state) => {
+    state.copy(attempt = state.attempt + 1, throwables = state.throwables :+ throwable)
+  }
+}
+
+case class IoState(attempt: Int, throwables: List[Throwable] = List.empty)
```
danbills: ToL the name could probably be something more descriptive, e.g. `Attempts` or `PreviousAttempts` or similar?

PR #4075, opened 2018-09-07
Prevent anonymous expression nodes from teleporting values between scopes

### LINKED ISSUE #3999: input definition for one call affects input definition for others
Defining inputs in a call overwrites/affects inputs to other calls when these inputs have the same name. This happens in cromwell 34, as well as the develop version (9bee537). It happens for WDL version 1.0 and Biscayne, but not for draft-2.

example:
```
version 1.0
workflow test {
    String out = "hello"
    call echo1 { #Should run `echo hello1`, but runs `echo21` if run second
        input:
            out = out + "1"
    }
    call echo2 { #should run `echo hello2`, but runs `echo 12` if run second
        input:
            out = out + "2"
    }
}
task echo1 {
    input {
        String out
    }
    command {
        echo ~{out}
    }
}
task echo2 {
    input {
        String out
    }
    command {
        echo ~{out}
    }
}
```
I added the echo task twice to check if it might be caused by running the same task multiple times, but this also happens when it's two different tasks with equally named inputs.

Defining one or both inputs as variables before passing them to the call seems works as expected:
```
workflow test {
    String out = "hello"
    call echo1 { # runs `echo hello1`
        input:
            out = out + "1"
    }
    String out2 = out + "2"
    call echo2 { # runs `echo hello2`
        input:
            out = out2
    }
}
```

aednichols:
Closes #3999 

The first and second commits each fix the user-reported issue on their own.


mcovarr:
I dunno, quantum entanglement of expression nodes sounds pretty cool actually.

centaur/src/main/resources/standardTestCases/draft3_same_named_inputs_priority.test
```diff
@@ -0,0 +1,17 @@
+name: draft3_same_named_inputs_priority
+testFormat: workflowsuccess
+backends: [Local]
+tags: [localdockertest]
```
cjllanwarne: Since this is really a WOMification test you could probably tie this in to the `WdlFileToWomSpec` rather than resorting to centaur/local-only.

That would also allow us to have `draft-2`, `1.0` and `biscayne` editions without having to make separate centaur tests for each version

cjllanwarne: Since this is really a WOMification test you could probably tie this in to the `WdlFileToWomSpec` rather than resorting to centaur/local-only.

That would also allow us to have `draft-2`, `1.0` and `biscayne` editions without having to make separate centaur tests for each version

wdl/transforms/new-base/src/main/scala/wdl/transforms/base/wdlom2wom/WorkflowDefinitionElementToWomWorkflowDefinition.scala
```diff
@@ -87,10 +88,12 @@ object WorkflowDefinitionElementToWomWorkflowDefinition {
 
       currentValidation flatMap { currentList =>
         val availableValues: Map[String, OutputPort] = (for {
+          // Anonymous expression nodes are one-time-use and do not exist in the universe of available linking candidates (#3999)
+          node <- currentList.filterNot(_.isInstanceOf[AnonymousExpressionNode])
           node <- currentList
           port <- node.outputPorts
         } yield outputName(node, port) -> port).toMap
-        val nextGraphNodeValidation = WorkflowGraphElementToGraphNode.convert(GraphNodeMakerInputs(next, linkedGraph.consumedValueLookup, availableValues, linkedGraph.typeAliases, workflowName, insideAScatter, callables))
+        val nextGraphNodeValidation: ErrorOr[Set[GraphNode]] = WorkflowGraphElementToGraphNode.convert(GraphNodeMakerInputs(next, linkedGraph.consumedValueLookup, availableValues, linkedGraph.typeAliases, workflowName, insideAScatter, callables))
```
cjllanwarne: This made me notice that the name is a little confusing... maybe while you're here we could update it to something less obtuse, eg `generatedGraphNodesValidation`?

wom/src/main/scala/wom/graph/expression/AnonymousExpressionNode.scala
```diff
@@ -14,7 +14,15 @@ object AnonymousExpressionNode {
                                                      expression: WomExpression,
                                                      inputMapping: Map[String, OutputPort],
                                                      constructor: AnonymousExpressionConstructor[T]): ErrorOr[T] = {
-    ExpressionNode.buildFromConstructor(constructor)(identifier, expression, inputMapping)
+    // Anonymous expression nodes are created and then immediately consumed.
+    // Name mangling prevents lookups from finding them and using them after their expiration date.
+    // Naturally, leading underscore is an illegal identifier so would never happen IRL.
```
cjllanwarne: Well... that's true in WDL but I don't think WOM makes any such claim.

OTOH I don't think CWL uses anonymous expressions like this and our n=2 right now, so we should be fine 🤞

wom/src/main/scala/wom/graph/expression/AnonymousExpressionNode.scala
```diff
@@ -14,7 +14,15 @@ object AnonymousExpressionNode {
...
+    val mangledIdentifier = identifier.copy(
```
cjllanwarne: maybe `unfindable` rather than... `mangled`?

centaur/src/main/resources/standardTestCases/draft3_same_named_inputs_priority.test
aednichols: Hmm, I wanted to be able to read the metadata to verify the output values are as expected

But it seems likely that if the WOM is correct, the rest is already covered by other tests

wdl/transforms/new-base/src/main/scala/wdl/transforms/base/wdlom2wom/WorkflowDefinitionElementToWomWorkflowDefinition.scala
aednichols: Yes, that is exactly why I added the annotation!

I agree with pluralizing the name as well

wom/src/main/scala/wom/graph/expression/AnonymousExpressionNode.scala
aednichols: Is the trouble with "mangled" that it sounds unpleasant, or that it is inaccurate?

I think it would be a fairly standard term for what's going on here - see e.g. https://en.wikipedia.org/wiki/Name_mangling#Creating_unique_names_for_inner_and_anonymous_classes

cjllanwarne: Huh, TIL...

In my mental models "mangle" would involve changing it in a way that makes it look totally unrecognizable (rather than just adding a small prefix, for example), but I agree - according to that link, `mangle` is indeed the correct term for what's happening here

aednichols:
I backed out the name-mangling change because it was redundant in fixing the actual bug and had far-reaching consequences.
- The upgrade script was very broken because it makes extensive use of anonymous node names to come up with real names for what to put in the WDL
- String concatenation and string comparison feel like gross tools to use when we have types at our disposal... i.e. evaluating `.isInstanceOf[AnonymousExpressionNode]`. I can imagine a future where we have a `canLinkWith` function that evaluates name and type to return a boolean

wdl/transforms/new-base/src/main/scala/wdl/transforms/base/wdlom2wom/WorkflowDefinitionElementToWomWorkflowDefinition.scala
```diff
@@ -87,11 +88,15 @@ object WorkflowDefinitionElementToWomWorkflowDefinition {
 
       currentValidation flatMap { currentList =>
         val availableValues: Map[String, OutputPort] = (for {
-          node <- currentList
+          // Anonymous expression nodes are one-time-use and do not exist in the universe of available linking candidates (#3999)
+          node <- currentList.filterNot(_.isInstanceOf[AnonymousExpressionNode])
```
cjllanwarne: I'm pretty sure in a for comprehension that's equivalent to 
```
node <- currentList if !node.isInstanceOf[AnonymousExpressionNode])
```

wdl/transforms/draft3/src/test/scala/wdl/draft3/transforms/wdlom2wom/WdlFileToWomSpec.scala
```diff
@@ -134,4 +136,21 @@ class WdlFileToWomSpec extends FlatSpec with Matchers {
...
+
+    val exposedExpressionNode = nodes.filter(_.isInstanceOf[ExposedExpressionNode]).head
+    val callInputs = nodes.filter(_.isInstanceOf[TaskCallInputExpressionNode]).toList
+
+    callInputs.size shouldBe 4
+
+    // Usually, we use node names to check whether in-memory graphs are linked together correctly.
+    // Obviously, in this regression test for a same-named-node bug that won't work - instead,
+    // we use identity (the hash code) to differentiate multiple nodes with the same name.
+    callInputs(0).inputPorts.head.upstream eq exposedExpressionNode.outputPorts.head shouldBe true
```
cjllanwarne: I'm surprised/disappointed that there's not a scalatest primitive for reference equality checks

aednichols: Upgraded to `theSameInstanceAs`, just took some digging

PR #4076, opened 2018-09-07
using fs2 retry impl & cats effect

danbills:
wanted to get rid of 

* thread.sleep
* exponential backoff retry logic


supportedBackends/aws/src/main/scala/cromwell/backend/impl/aws/AwsBatchJob.scala
```diff
@@ -142,27 +138,19 @@ final case class AwsBatchJob(jobDescriptor: BackendJobDescriptor,           // W
...
-      } catch {
-        case e: ClientException if e.statusCode() == 404 && inx < 6 => {
-          // RegisterJobDefinition is eventually consistent, so it may not be there
-          Log.warn(s"Job Definition does not yet exist - retrying (${inx}/5)")
-          Thread.sleep(100L * (inx ^ 2L))
-        }
-      }
-    }
-    job
+    val definitionArn: F[String] = createDefinition[F](s"""${workflow.callable.name}-${jobDescriptor.taskCall.callable.name}""", uniquePath)
```
danbills: I should pass in the `ScheduledExecutorService` I already have here

supportedBackends/aws/src/main/scala/cromwell/backend/impl/aws/AwsBatchJob.scala
```diff
@@ -195,32 +187,21 @@ final case class AwsBatchJob(jobDescriptor: BackendJobDescriptor,           // W
...
-            .describeJobDefinitions(DescribeJobDefinitionsRequest.builder().jobDefinitionName(sanitize(name)).build())
-            .jobDefinitions()
-          // We need latest revision
-          defs.get(defs.size() - 1).jobDefinitionArn()
-        }
-      }
+    val submit = async.delay(client.registerJobDefinition(definitionRequest).jobDefinitionArn)
+
+    val retry: F[String] = Scheduler.fromScheduledExecutorService(ses).retry(submit, 0.millis, _ * 2, 6, {
+      case e: ClientException => e.statusCode() == 404
```
Horneth: Could you add back the existing comment explaining why this might happen (eventual consistency)

supportedBackends/aws/src/main/scala/cromwell/backend/impl/aws/AwsBatchJob.scala
```diff
@@ -195,32 +187,21 @@ final case class AwsBatchJob(jobDescriptor: BackendJobDescriptor,           // W
...
-          val defs = client
-            .describeJobDefinitions(DescribeJobDefinitionsRequest.builder().jobDefinitionName(sanitize(name)).build())
-            .jobDefinitions()
-          // We need latest revision
-          defs.get(defs.size() - 1).jobDefinitionArn()
-        }
-      }
+    val submit = async.delay(client.registerJobDefinition(definitionRequest).jobDefinitionArn)
+
+    val retry: F[String] = Scheduler.fromScheduledExecutorService(ses).retry(submit, 0.millis, _ * 2, 6, {
```
Horneth: ToL: Even if it's a little more code I personally prefer pulling out those "magic" numbers and functions into declared variables and non-anonymous function for clarity

supportedBackends/aws/src/main/scala/cromwell/backend/impl/aws/AwsBatchJob.scala
```diff
@@ -195,32 +187,21 @@ final case class AwsBatchJob(jobDescriptor: BackendJobDescriptor,           // W
...
+      case e: ClientException => e.statusCode() == 404
+      case _ => false
+    }).compile.last.map(_.get)
+
+    async.recoverWith(retry) {
+      case e: ClientException if e.statusCode() == 409 =>
+        Log.warn("Job definition already exists. Performing describe and retrieving latest revision")
```
Horneth: Is it something that's expected to happen frequently ? For instance if we scatter 10000 wide would we try to create the same job definition 10000 times and so log this message 9999 time ? If so I'd advocate for not logging this (or not at warn level at least)

danbills: It's a permissions thing, per comments above:
```scala
          // This could be a problem here, as we might have registerJobDefinition
          // but not describeJobDefinitions permissions. We've changed this to a
          // warning as a result of this potential
```
Thus we eitehr get 10,000 of them or we get 0.  I think `WARN` is ok IMO

danbills: True... I don't feel strongly about this issue as it is an edge case.  I'll put it in as `INFO` unless I hear a vote in favor of `DEBUG`

danbills:
@mcovarr yes.  We don't have the attempt index w/ fs2 but figured this was a reasonable approach.

supportedBackends/aws/src/main/scala/cromwell/backend/impl/aws/AwsBatchJob.scala
cjllanwarne: If this really is a "normal part of operation that will happen every time we scatter" I'd say don't log it at all (at least not separate messages for *every* *single* *shard*).

supportedBackends/aws/src/main/scala/cromwell/backend/impl/aws/AwsBatchJob.scala
```diff
@@ -190,51 +198,47 @@ final case class AwsBatchJob(jobDescriptor: BackendJobDescriptor,           // W
...
+    *
+    */
+  private def sanitize(name: String): String Refined AwsStringRefinement =
     // Up to 128 letters (uppercase and lowercase), numbers, hyphens, and underscores are allowed.
     // We'll replace all invalid characters with an underscore
-    name
-      .replaceAll("[^A-Za-z0-9_\\-]", "_")
-      .slice(0,128)
+    refineV[AwsStringRefinement](name.replaceAll("[^A-Za-z0-9_\\-]", "_").slice(0,128)).
+      getOrElse(throw new RuntimeException("this exception is impossible to be thrown as the previous line removed all offenders"))
```
cjllanwarne: I like to at least say `"Programmer Error! Please report this as..."` since otherwise it causes a ton of FUD when it eventually shows up (and as Jeff can attest... eventually, these always show up!)

danbills: Sure thing

supportedBackends/aws/src/main/scala/cromwell/backend/impl/aws/AwsBatchJob.scala
danbills: Per the original author's comment (I've since added back), it's clear that there is a potential issue w/ this logic and warrants a `WARN`.  I'm going to honor these intentions and leave it as a `WARN`.

PR #4083, opened 2018-09-11
 EJEA transition timings to statsD

Horneth:
@danbills I was looking at the EJEA and on second thoughts it seemed a lot easier to just have it push the state transition to statsD directly.
I was thinking we could still use the ammonite script to aggregate transition time statistics about a single test run (which we can't really easily do in grafana) in a SQL table like you suggested before.
This way we could compare test by test how the measurements evolve, while grafana would show us "real time" how the timings are moving (which may even be useful in prod). Let me know what you think.

Example from a test run (Y axis is ms):

screen shot 2018-09-11 at 9 27 48 am

~~Unrelated but I also took the opportunity to make execution events be pushed as they happen (the EJEA ones, not the backend ones - yet). This will hopefully give users more visibility on what's going on with their job in the timing diagram~~


services/src/main/scala/cromwell/services/metadata/MetadataQuery.scala
```diff
@@ -63,8 +63,10 @@ object MetadataValue {
       case _: Double | Float | _: java.lang.Double | _: java.lang.Float => new MetadataValue(value.toString, MetadataNumber)
       case _: Boolean | _: java.lang.Boolean => new MetadataValue(value.toString, MetadataBoolean)
       case other => new MetadataValue(other.toString, MetadataString)
-    }
+    } getOrElse Null
   }
+
+  val Null = new MetadataValue("", MetadataNull)
```
mcovarr: what's the difference in behavior with these changes?

Horneth: The only difference in behavior is that if you pass `null` to the apply of `MetadataValue` it will create a `new MetadataValue("", MetadataNull)` which will translate to a `null` in the JSON, whereas before you'd get an empty string.

mcovarr: Are the consumers of Cromwell metadata going to be okay with that change?

Horneth: Good question, I think it makes more sense this way but it'll probably be less painful to everyone to revert it in this case and pass a `MatadataValue.Null` explicitly for the execution event case.
Either way it will mean that the `endTime` of execution events can now be null which I don't think anyone will care about but it should definitely be made clear

geoffjentry: Do not inject a backwards incompatible change into the API plz

Horneth: Mokay I have a feeling this was a mistake 😄 I'll revert this whole thing

PR #4085, opened 2018-09-11
Caching actor: Don't emit fatal timeout messages that must be ignored

cjllanwarne:
@Horneth I think this was originally added by you (but git blame might be lying...), could you sanity check this?

### DIFF
--- a/backend/src/main/scala/cromwell/backend/standard/callcaching/StandardCacheHitCopyingActor.scala
+++ b/backend/src/main/scala/cromwell/backend/standard/callcaching/StandardCacheHitCopyingActor.scala
@@ -178,8 +178,12 @@ abstract class StandardCacheHitCopyingActor(val standardParams: StandardCacheHit
           stay() using Option(newData)
       }
     case Event(IoFailure(command: IoCommand[_], failure), Some(data)) =>
-      // any failure is fatal
-      context.parent ! JobFailedNonRetryableResponse(jobDescriptor.key, failure, None)
+      // We can get extra I/O timeout messages from previous cache copy attempt. This is due to the fact that if one file fails to copy,
+      // we immediately cache miss and try the next potential hit, but don't cancel previous copy requests.
+      // (any other failure is fatal)
+      if (!failure.isInstanceOf[TimeoutException]) {
+        context.parent ! JobFailedNonRetryableResponse(jobDescriptor.key, failure, None)
+      }
 
       val (newData, commandState) = data.commandComplete(command)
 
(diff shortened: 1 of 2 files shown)


backend/src/main/scala/cromwell/backend/standard/callcaching/StandardCacheHitCopyingActor.scala
```diff
@@ -178,8 +178,12 @@ abstract class StandardCacheHitCopyingActor(val standardParams: StandardCacheHit
           stay() using Option(newData)
       }
     case Event(IoFailure(command: IoCommand[_], failure), Some(data)) =>
-      // any failure is fatal
-      context.parent ! JobFailedNonRetryableResponse(jobDescriptor.key, failure, None)
+      // We can get extra I/O timeout messages from previous cache copy attempt. This is due to the fact that if one file fails to copy,
+      // we immediately cache miss and try the next potential hit, but don't cancel previous copy requests.
+      // (any other failure is fatal)
```
mcovarr: what about a permissions failure trying to copy one cache hit?

backend/src/main/scala/cromwell/backend/standard/callcaching/StandardCacheHitCopyingActor.scala
```diff
@@ -178,8 +178,12 @@ abstract class StandardCacheHitCopyingActor(val standardParams: StandardCacheHit
...
+      if (!failure.isInstanceOf[TimeoutException]) {
```
Horneth: Why the check on `TimeoutException` ? If it is a timeout then aren't we going to get the EJEA stuck waiting for a response ?

cjllanwarne: this is the message that the EJEA was dropping on the floor previously

Horneth:
Yeah I'm not sure I understand how this is the problem, it looks like as soon as we get an `IoFailure` we change state and get to `Failing`, where we just wait for the rest of the IO responses but don't send anything else to the EJEA. Since it looks like we create one of this per cache hit I don't see how this actor can keep sending "JobFailed" message multiple times

backend/src/main/scala/cromwell/backend/standard/callcaching/StandardCacheHitCopyingActor.scala
mcovarr: Per discussion in chat:

Since this actor is only responsible for copying a single cache hit, I don't think it should be able to say that all cache-hit copying attempts should cease and the job should run.

Horneth: I think we're just re-using `JobFailedNonRetryableResponse` and changing it's meaning when we really want something like a `JobCacheHitCopyingFailed`

PR #4087, opened 2018-09-12
Configurable call caching file batch size

Horneth:


geoffjentry:
How was the performance impact of this tested?

Horneth:
@geoffjentry It wasn't yet, AFAIK you need a branch to test stuff on the perf env so I thought I'd make a PR of it already and test it in the meantime. We could also leave the current value of 100 and merge this which would have no change in behavior and then tune the value if necessary with benchmarks

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheHashingJobActor.scala
```diff
@@ -237,15 +239,14 @@ object CallCacheHashingJobActor {
   }
 
   object CallCacheHashingJobActorData {
-    // Slick will eventually build a prepared statement with that many parameters. Don't set this too high or it will stackoverflow.
-    val BatchSize = 100
-
-    def apply(fileHashRequestsRemaining: List[SingleFileHashRequest], callCacheReadingJobActor: Option[ActorRef]) = {
-      new CallCacheHashingJobActorData(fileHashRequestsRemaining.grouped(BatchSize).toList, List.empty, callCacheReadingJobActor)
+    // Slick will eventually build a prepared statement with that many parameters. Don't set batchSize this too high or it will stackoverflow.
```
cjllanwarne: Perhaps this comment belongs in the config file now?

engine/src/main/scala/cromwell/server/CromwellRootActor.scala
```diff
@@ -56,7 +56,10 @@ abstract class CromwellRootActor(gracefulShutdown: Boolean, abortJobsOnTerminate
   import CromwellRootActor._
   
   // Make sure the filesystems are initialized at startup
-  val _ = CromwellFileSystems.instance
+  locally(CromwellFileSystems.instance)
+  // Fail immediately if call caching configuration is invalid
+  // TODO: Find a better way to do this
+  locally(cromwell.engine.workflow.lifecycle.execution.callcaching.FileBatchSize)
```
cjllanwarne: I'm curious what made this single option required here?

Horneth: I wanted Cromwell to not even start if this configuration is invalid. I imagine we do this differently for other configuration but this is the simplest place I could find where to fail so that we stop

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheHashingJobActor.scala
```diff
@@ -274,7 +275,7 @@ object CallCacheHashingJobActor {
           // If the current batch is empty, we got a partial result, take the first BatchSize of the list
           if (updatedBatch.isEmpty) {
             // hashResult + fileHashResults.take(BatchSize - 1) -> BatchSize elements
-            val partialHashes = NonEmptyList.of[HashResult](hashResult, fileHashResults.take(BatchSize - 1): _*)
+            val partialHashes = NonEmptyList.of[HashResult](hashResult, fileHashResults.take(batchSize - 1): _*)
```
Horneth: @cjllanwarne If you're talking about this `- 1`, it's not to make room for the base hash, it's just because this function is called when we have a new hash (`hashResult`). So to create a bucket of size `batchSize` we take this new hash plus the last `batchSize - 1` hashes

Horneth: Although adding the base hash as a filter on top of the bucket could be interesting to try out

PR #4091, opened 2018-09-12
Cache output copy failures

cjllanwarne:
Part II of the "unexpected `JobFailedNonRetryable` during cache output copying" saga.

- Stop using the confusing `JobFailedNonRetryable` response when an output copy fails (it's not like `JobSucceeded` which correctly tells us that the job has already succeeded)
- Update the tests to reflect this
- Add some sanity checks to the EJEA's handlers (specifically - did we copy the right outputs, and did we fetch the right outputs from the database)

### DIFF
--- a/common/src/test/scala/common/assertion/CaseClassAssertions.scala
+++ b/common/src/test/scala/common/assertion/CaseClassAssertions.scala
@@ -1,0 +1,14 @@
+package common.assertion
+
+import org.scalatest.Matchers
+
+object CaseClassAssertions extends Matchers {
+  implicit class ComparableCaseClass[A <: Product](actualA: A) {
+    // Assumes that expectedA and actualA are the same type. If we don't subtype case classes, that should hold...
+    def shouldEqualFieldwise(expectedA: A): Unit = {
+      (0 until actualA.productArity) foreach { i =>
+        actualA.productElement(i) should be(expectedA.productElement(i))
+      }
+    }
+  }
+}
--- a/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/job/EngineJobExecutionActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/job/EngineJobExecutionActor.scala
@@ -1,7 +1,5 @@
 package cromwell.engine.workflow.lifecycle.execution.job
 
-import java.util.concurrent.TimeoutException
-
 import akka.actor.SupervisorStrategy.{Escalate, Stop}
 import akka.actor.{ActorInitializationException, ActorRef, LoggingFSM, OneForOneStrategy, Props}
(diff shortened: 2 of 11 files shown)


engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/job/EngineJobExecutionActor.scala
```diff
@@ -208,18 +206,21 @@ class EngineJobExecutionActor(replyTo: ActorRef,
...
-        callCachingHitResultMetadataKey -> true,
-        callCachingReadResultMetadataKey -> s"Cache Hit: $cacheHitDetails"))
-      log.debug("Cache hit for {}! Fetching cached result {}", jobTag, cacheResultId)
-      makeBackendCopyCacheHit(womValueSimpletons, jobDetritus, returnCode, data, cacheResultId) using data.withCacheDetails(cacheHitDetails)
+    case Event(CachedOutputLookupSucceeded(womValueSimpletons, jobDetritus, returnCode, cacheResultId, cacheHitDetails), data @ ResponsePendingData(_, _, _, _, Some(ejeaCacheHit), _)) =>
+      if (cacheResultId != ejeaCacheHit.hit.cacheResultId) {
+        // Sanity check: was this the right set of results (a false here is a BAD thing!):
+        log.error(s"Received incorrect call cache results from FetchCachedResultsActor. Expected ${ejeaCacheHit.hit} but got $cacheResultId. Running job")
+        // Treat this like the "CachedOutputLookupFailed" event:
+        runJob(data)
```
cjllanwarne: Re "Treat this like the "CachedOutputLookupFailed" event", I wonder whether in both cases ` invalidateCacheHitAndTransition(cacheHit, data, throwable)` would be a more appropriate response than `runJob(data)`?

common/src/test/scala/common/assertion/CaseClassAssertions.scala
```diff
@@ -0,0 +1,14 @@
+package common.assertion
+
+import org.scalatest.Matchers
+
+object CaseClassAssertions extends Matchers {
+  implicit class ComparableCaseClass[A <: Product](actualA: A) {
+    // Assumes that expectedA and actualA are the same type. If we don't subtype case classes, that should hold...
```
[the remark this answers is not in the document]
cjllanwarne: Beyond what the `A` constraint is type-checking?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/job/EngineJobExecutionActor.scala
cjllanwarne: Mainly because I noticed that we can rather than because we've noticed this going awry. If you want I can change this to be "write the error but don't change behavior from what it was before (ie use the results regardless)"?

PR #4095, opened 2018-09-13
Allow member access when input name and input value have the same name

### LINKED ISSUE #3811: Valid WDL not compiling
The valid WDL file below fails to compile.

```wdl
workflow x {
    call cram
    call y { input:
        cram = cram.cram
    }
}

task cram {
    command {
        echo "."
    }
    output {
        String cram = "."
    }
}

task y {
    String cram
    command {
        echo "."
    }
}
```

aednichols:
Closes #3811 

Use a hint to the resolver to ignore the local scope of a `WdlTaskCall` and start the member access search in the next higher scope.

---

It would be really nice to use `relativeTo` to include only elements of `children` that are lexically before the line where the lookup is happening, but it seems that individual lines do not have their own `Scope`.

That would very theoretically look like this:
```
call.resolveVariable(memberAccess.lhsString, relativeTo = memberAccess) // Does not compile!
```


cjllanwarne:
With the tests, looks good to me

aednichols:
@kshakir I like the spirit of your comment but don't understand the exact intent. Do you have specific test coverage in mind beyond the four variants of `shouldBeProblematic` we have now?

PR #4108, opened 2018-09-17
Hog limits for job tokens

cjllanwarne:


engine/src/main/scala/cromwell/engine/workflow/tokens/JobExecutionTokenDispenserActor.scala
```diff
@@ -74,11 +75,14 @@ class JobExecutionTokenDispenserActor(override val serviceRegistryActor: ActorRe
         tokenAssignments = tokenAssignments + (actor -> lease)
         incrementJob("Started")
         actor ! JobExecutionTokenDispensed
-      // Only one token per actor, if you've already got one, you don't get this token !
-      case LeasedActor(_, lease) => lease.release()
+      // Only one token per actor, so if you've already got one, we don't need to use this new one:
+      case LeasedActor(actor, lease) =>
+        actor ! JobExecutionTokenDispensed
+        lease.release()
```
mcovarr: what's this change about?

cjllanwarne: Before: When an actor which already has a lease asks for another one, remove it from the queue and leave it hanging.
Now: When an actor which already has a lease asks for another one, remove it from the queue and resend the original `JobExecutionTokenDispensed` message.

Since this is almost certainly a sign of programmer error, I could probably add an error message alongside this

backend/src/main/scala/cromwell/backend/backend.scala
```diff
@@ -84,6 +84,18 @@ case class BackendWorkflowDescriptor(id: WorkflowId,
 
   val rootWorkflow = breadCrumbs.headOption.map(_.callable).getOrElse(callable)
   val rootWorkflowId = breadCrumbs.headOption.map(_.id).getOrElse(id)
+  lazy val hogGroup = {
+    val config = ConfigFactory.load
```
danbills: please take `hogGroupField` as `Option[String]` instead of statically loading Config

cjllanwarne: I'm interpreting this as "it would be nice to make this better, but now isn't necessarily the time". Let me know if that's wasn't your intention.

PR #4109, opened 2018-09-18
aws backend: support globbed outputs

delocalizer:


delocalizer:
I'm very happy to create a test case if someone gives me a pointer on how I might do that.

Horneth:
Thanks for the contribution @delocalizer !

Issue #4050, opened 2018-08-29
No timeout on rc file


ffinfo:
Related maybe: https://github.com/broadinstitute/cromwell/issues/3905

When my job is killed by our cluster (SGE or Slurm) the rc file will not appear anymore. Cromwell then stay forever waiting on this file.
When manual writing this file (aka `echo "2" > rc`) cromwell does see this as a failed job.

I have tested this also on a Local backend but here the problem does not exists. I think this is because cromwell stays connected to the process.

I have tested cromwell 34 and the current develop branch (ce27a93).

For testing I have used this workflow:
```
version 1.0

workflow Test {
  input {
  }

  call Echo as echo {
    input:
  }

  output {
  }
}

task Echo {
  input {
  }

  command {
    # killing bash process
    kill -9 $$
    echo test
  }

  output {
  }
}
```

ruchim:
@kshakir is there a configurable timeout for job completion in the SFS backends?

ffinfo:
@ruchim and @kshakir: Any update on this? This is for our piplines a real blocker to bring our pipelines to production. See also https://github.com/biowdl

ffinfo:
Fixed in https://github.com/broadinstitute/cromwell/pull/4112

PR #4112, opened 2018-09-18
Restructure sfs job status

ffinfo:
Fixing https://github.com/broadinstitute/cromwell/issues/4050

While doing this I notice that this check alive command is not used at all. First did a restructure of the statuses and now there is also a status Running.

- [x] Add timeout on `WaitingForReturnCode` step
- [x] Make timeout a config value

Meanwhile please give already feed on this of course ;)


ffinfo:
This seems to work for now. Anyone can take a look at this? The failing test is as far i can see not related what I did.

cjllanwarne:
Hi @ffinfo - thanks for contributing this!

My only concern as-is is that this has the potential to overload HPC clusters (see the discussions in https://github.com/broadinstitute/cromwell/issues/1499).

In your case, since you're aware of the dangers and willing to go ahead regardless so I don't see why we shouldn't let you - but I don't think this should be the default behavior for unsuspecting users. 

Would you be willing to make this a configurable option in the backend config (something like `check-alive-all-jobs`)?

Longer term if you start hitting HPC limits perhaps we can circle back round to the batching solutions hinted at in #1499

geoffjentry:
Turns out @cjllanwarne swooped in and said what I was going to say, so what he said. Also he said it much better than I ever could have, as usual.

@ffinfo you're obviously pretty familiar w/ this problem (I've traced it all the way back to an older PR of yours, #1346) and the main sticking point has been pretty consistent (i.e. the scaling Q), so I think @cjllanwarne is right and as long as it's configurable and defaulting in a manner which doesn't change current behavior I think this path is AOK.

ffinfo:
@geoffjentry and @cjllanwarne:
Thank you for you response. I do understand your concerns. In the future I we going to move away from traditional HPC systems (i hope). For now we have sadly to deal with them.

I already have this config value in place with a default. I can remove this default and using this behaviour when `exit-code-timeout` is set. This easy to do.

ffinfo:
@cjllanwarne and @geoffjentry: See now also ebb556e

geoffjentry:
Thanks @ffinfo we'll try to get this reviewed tomorrow

cjllanwarne:
@ffinfo my concern was more that the `isAlive` should be opt-in, not that that timeout should be opt in. 

if I'm reading this right:

- The job enters the `Running` state
- The first time we poll for it, we *always* check whether it's alive
  - Assuming it is, we enter the `WaitingForReturnCode`
- After a configurable timeout, we abandon it and declare it failed
  - If no timeout is configured, we keep waiting forever

I think this shouldn't be too much of a refactor:

- The job enters the `WaitingForReturnCode` state
- We immediately schedule an `CheckAlive` message to the actor at the configurable time
  - If the cadence is not set, we never send that message (this would be the default)
- When that CheckAlive arrives we can run `isAlive` and remember the result (and if we're still alive, schedule another `CheckAlive` again after the same delay)
- If the `isAlive` failed, the next poll would return `Failed`

ffinfo:
@cjllanwarne 
I'm not sure I fully understand what you mean here. `checkalive` does run/schedule when `isAlive` is called.
Al tough I think what I use know is almost as it already works. Only difference is that I first do isAlive instead of checking the exitcode. In f30c2be I did switch this. (first exitcode, then isAlive)
Before I did had a default timeout of 120 seconds but I did remove that because of earlier comments. I can bring that back in if you want?

cjllanwarne:
@ffinfo As I read it:

- If state is `None`, go to `Running` state
- If state is `Running`, the first thing we do is run `isAlive` (which I don't want to ever do!)

That means that I have no way to opt-out of ever running `isAlive` (which is the thing I want before approving this PR).

---

What I was suggesting is (but there are many other ways):
- If I set `pollForAliveness: "1 minute"` in the config file:
  - Use `context.system.scheduler.scheduleOnce` to run an `isAlive` 1 minute in the future (completely separate from `pollStatus`).
    - If that `isAlive` is true, schedule again another 1 minute in the future
    - If not, record the time at which the job was not alive
  - `pollStatus` continues on a different schedule:
    - If the job is no longer alive, the `pollStatus` switches to `WaitingForReturnCode`
    - If a time limit is set for the `WaitingForReturnCode` state, honor it
- If I set `pollForAliveness: false` in the config file:
  - Go straight to `WaitingForReturnCode`
  - No time limits for `WaitingForReturnCode`

cjllanwarne:
In other words, I'm looking for the *default* configuration of Cromwell to *never* run `isAlive` (except on server restarts, like today). I should have to change my configuration to make it happen.

geoffjentry:
Right, I think @cjllanwarne is correct here (although he might have biased me). The request is to make sure `isAlive` is only ever called during the polling if the user has configured Cromwell to allow it

ffinfo:
@cjllanwarne and @geoffjentry 
I did make a change that when exit-code-timeout is not set `isAlive` is not used anymore.

One note I need to make about this. This can give the wrong impression to HPC users. isAlive is there the only way to ensure if a job is still running while the rc file is not stable.
This is a setting that should be default for all dispatch systems that does force kill the jobs.

cpavanrun:
This is a problem we have ran into as well. Our solution for now was to patch cromwell to trap SIGUSR1/2 signals to modify the rc file. These are a pre-signal fired by SGE to let the job know it's about to be killed because of requested-resource limits.

However, patching Cromwell every release is a bit of a bother and your solution looks a lot more promissing. Looking at the different recurring discussions on this topic I would strongly urge you to to make the timeout a recurring poll instead of a one-time-fire; I am not well versed in scala so it might be that this is already what you do =).

supportedBackends/sfs/src/main/scala/cromwell/backend/impl/sfs/config/ConfigAsyncJobExecutionActor.scala
```diff
@@ -246,4 +250,50 @@ class DispatchedConfigAsyncJobExecutionActor(override val standardParams: Standa
...
+    handle.previousStatus match {
+      case None =>
+        // Is not set yet the status will be set default to running
+        SharedFileSystemRunStatus("Running")
+      case Some(s) if (s.status == "Running" || s.status == "WaitingForReturnCode") && jobPaths.returnCode.exists =>
+        // If exitcode file does exists status will be set to Done always
+        SharedFileSystemRunStatus("Done")
+      case Some(s) if s.status == "Running" =>
+        // Exitcode file does not exist at this point, checking is jobs is still alive
+        if (exitCodeTimeout.isEmpty) s
```
cjllanwarne: I would move this into the `case None` so that we jump straight to `WaitingForReturnCode ` if the timeout is not set.

supportedBackends/sfs/src/main/scala/cromwell/backend/impl/sfs/config/ConfigAsyncJobExecutionActor.scala
```diff
@@ -246,4 +250,50 @@ class DispatchedConfigAsyncJobExecutionActor(override val standardParams: Standa
...
+        else if (isAlive(handle.pendingJob).getOrElse(true)) s
```
cjllanwarne: Are you happy that this `getOrElse(true)` is safe?

cjllanwarne:
Thanks @ffinfo - this looks like it should work. I had a few comments but they're more "comments to think about" than "requirements to change". I'll leave it up to you.

One last thing - I agree that people should be aware that this is possible. Could you add:
- something to `cromwell-examples.conf` to demonstrate this 
- something to `CHANGELOG.MD`
- something to `docs/backends/SGE`/`LSF`/etc to talk about this new option

Thanks again!

supportedBackends/sfs/src/main/scala/cromwell/backend/impl/sfs/config/ConfigAsyncJobExecutionActor.scala
ffinfo: I tried this first but then I did hit a boxed error, this solution did work. Also don't know why that happened.

supportedBackends/sfs/src/main/scala/cromwell/backend/impl/sfs/config/ConfigAsyncJobExecutionActor.scala
ffinfo: no not really but I don't want to raise an exception because sometimes this kind of calls just fail. SGE is not so stable ;)

supportedBackends/sfs/src/main/scala/cromwell/backend/impl/sfs/config/ConfigAsyncJobExecutionActor.scala
ffinfo: btw, if you know a why this happens I'm happy to change this ofcourse

geoffjentry:
@cpavanrun If I understood you correctly I think this change will give you what you want, albeit you'll need to opt in to that behavior via configuration

ffinfo:
@cjllanwarne now with docs.
Can you check if those markdown links does work correctly? I think they do but with markdown you never know ;)

ffinfo:
@cpavanrun If this is enabled in the config like this it does retry:
https://cromwell.readthedocs.io/en/develop/RuntimeAttributes/#maxretries

Also the timeout will start counting when isAlive returns false for the first time. After that isAlive will not be called anymore.

cpavanrun:
Ah, I see now how you intent for it to work.

I don't think this will be very practical on any kind of shared SGE HPC without a seperate poll rate. 

As mentioned elsewhere; the call rate of the `pollStatus` is geared towards a low-impact filesytem check and not a high-impact call to the SGE queque master (i.e. `check-alive` uses `qstat`). Enabling the `exit-code-timeout` with a somewhat large numbers of tasks will quickly cripple any average SGE HPC. 

If you enable this and are an HPC admin; make sure to keep taps on your submission services.

ffinfo:
@cpavanrun Your right here, but this can be improved. What can be done here is lower the number of parallel jobs submitted by cromwell. This depends really on the cluster. In our case I did a stress test with 10000 parallel jobs and it still acts fine. Only downside is that the log is getting spammed a bit but it still works like it should.

Still in the past (on older hardware) the headnode could not deal with this number of jobs. If this is the case limiting the parallel jobs could be a fix.

cjllanwarne:
@cpavanrun I certainly think there's room to iterate on how we poll for terminated jobs on HPC clusters, so if you try this out and think "we need a small tweak to make this work" - 100% please make those issues (or PRs 😄)!

cjllanwarne:
Note - you'll probably need to rebase onto the most recent `develop` commit to have the tests pass.

geoffjentry:
Hi @cpavanrun - I see what you mean now. You're channeling thoughts from one of our devs (@kshakir ) when this has come up in the past. Our take in this pass is that @ffinfo is making it opt-in and this can allow folks who are using SGE to get some sense of what does/doesn't work here and potentially iterate on it, e.g. the stress tests he mentioned.

ffinfo:
@cjllanwarne Thank you for your comments, fixed the text issues :)
The branch is should already be up to date with develop currently

ffinfo:
main while cromwell 35 is released, that's to bad that this change is not in there yet ;)

cjllanwarne:
Thanks @ffinfo - looks good to me: 👍 

ps don't worry too much about missing Cromwell 35, we're hoping to get better at releasing more regularly (even if we don't have world-changing features in every release) - so it should only be a few weeks before Cromwell 36 gets released, including this change.

kshakir:
NOTE: If anyone wants to follow up on #1499 and put in [less aggressive polling](https://github.com/broadinstitute/cromwell/issues/1499#issuecomment-328984482), here is the [heart of the batch requesting](https://github.com/broadinstitute/cromwell/blob/de95456159c27a5145612b6acc95d4e713a5e9f8/supportedBackends/google/pipelines/common/src/main/scala/cromwell/backend/google/pipelines/common/api/PipelinesApiBatchHandler.scala) wired into the Google/PAPI backend that could be generalized for all backends.

supportedBackends/sfs/src/main/scala/cromwell/backend/impl/sfs/config/ConfigAsyncJobExecutionActor.scala
```diff
@@ -246,4 +250,50 @@ class DispatchedConfigAsyncJobExecutionActor(override val standardParams: Standa
   override def killArgs(job: StandardAsyncJob): SharedFileSystemCommand = {
     jobScriptArgs(job, "kill", KillTask)
   }
+
+  protected lazy val exitCodeTimeout: Option[Int] =
+    configurationDescriptor.backendConfig.as[Option[Int]](ExitCodeTimeoutConfig).map(math.abs)
```
mcovarr: Why `math.abs`? In that vein it might be good to have minimum acceptable values. 🙂

cromwell.examples.conf
```diff
@@ -452,6 +452,14 @@ backend {
     #    # Limits the number of concurrent jobs
     #    concurrent-job-limit = 5
     #
+    #    # If an 'exit-code-timeout' value is specified:
+    #    # - When a job has not been alive for longer than this timeout (in seconds)
```
mcovarr: Having 'seconds' be part of the variable name would help make this self-documenting.

supportedBackends/sfs/src/main/scala/cromwell/backend/impl/sfs/config/ConfigAsyncJobExecutionActor.scala
cjllanwarne: I did mean *not* set - on reflection I suspect there are many *many* ways to structure this... TBH I don't think we'd gain much with my suggestion so I'm happy to leave it as-is for now

cjllanwarne:
We've worked out why the CI is failing - hopefully once we merge https://github.com/broadinstitute/cromwell/pull/4196 you can rebase to fix the broken test 🤞

cjllanwarne:
@ffinfo sorry for the slog that this PR has turned into. To give you some hope: once you've satisfied @mcovarr and he gives you a 👍 (and once you've rebased to let the tests pass) we should be good to merge.

Thanks again for the PR!

supportedBackends/sfs/src/main/scala/cromwell/backend/impl/sfs/config/ConfigAsyncJobExecutionActor.scala
ffinfo: agree, will raise an exception if this is below 0

cromwell.examples.conf
ffinfo: Good one, will fix this

supportedBackends/sfs/src/main/scala/cromwell/backend/impl/sfs/config/ConfigAsyncJobExecutionActor.scala
ffinfo: ye that should be fine, does not happen that often.

ffinfo:
@cjllanwarne I think it's finished now.

Meanwhile I detected a bug but did solve this with an other opt-in config value to be able to retry jobs that are aborted.

Aborted here means an exitcode above 128 if i read the code correctly. Maybe rename this to ExternalKill instead?

cjllanwarne:
Hi @ffinfo would you might rolling back the `RetryAbortedJobs` changes and submitting them again as a separate PR? 

I suspect that you're conflating `abort` as something external  vs `abort` as something that Cromwell does itself (eg from a REST request or while it's shutting itself down) - and we need to be careful to get all of those interactions right - especially if this affects other backends.

In any case, I think it's worth having it properly reviewed as its own change (rather than having it delay an otherwise approved PR 😄).

PR #4121, opened 2018-09-19
Remove exists query for CC lookup

Horneth:
Perf testing has shown that removing this query improves CC time and reduces DB load (see last row in CC google doc)
Unclear if it's worth keeping it as a configurable thing ?
This keeps storing the individual hashes, it just stops using them for "fast" cache miss detection.

### DIFF
--- a/database/sql/src/main/scala/cromwell/database/slick/CallCachingSlickDatabase.scala
+++ b/database/sql/src/main/scala/cromwell/database/slick/CallCachingSlickDatabase.scala
@@ -1,6 +1,5 @@
 package cromwell.database.slick
 
-import cats.data.NonEmptyList
 import cats.instances.list._
 import cats.instances.tuple._
 import cats.syntax.apply._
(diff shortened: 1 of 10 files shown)


geoffjentry:
@Horneth What, if anything, are the downsides of this change?

IOW is this creating a tradeoff where we're deciding this improvement is overall better than the negative change elsewhere, or is this purely positive?

Horneth:
The only theoretical downside is that it could take longer before we realize it's a cache miss and fall back to running the job. In practice though, this query which is suppose to make us "fail faster" is under performing so badly that it's effectively slowing us down. So it's a net positive in call caching time and resources spent. The alternative would be to replace it with something more performant, which we can always do later if the need arises.

cjllanwarne:
@Horneth I assume for now we're still *recording* all the individual hashes even though we don't currently ever look at them? 

So if we did need to back out of this we could just reinstate this and be good to go?

Horneth:
@cjllanwarne correct

PR #4127, opened 2018-09-20
[34 HOTFIX] Switch from using JSON pointers to JQ filters.

kshakir:


kshakir:
The DOS schema changed to a pattern not queryable by JSON Paths-- specifically returning a nested array-of-objects where the array is unordered but only a certain object-value with a prefix should be returned...

Found that someone had made the more flexible JQ filter syntax available, so switched to that library. I've still seen no evidence that the DOS schema won't change again, but hopefully the JQ filters will allow flexibly plucking out the right field as the schema hardens.

The other alternative was submitting another PR if/when the schema changes again.

kshakir:
Same [tests are failing](https://travis-ci.org/broadinstitute/cromwell/builds/427793746) in the previous `34_hotfix`. Merging.

PR #4133, opened 2018-09-21
Slim logs to get them back under Travis' 4mb limit.

kshakir:
- Truncate stack traces for logged-but-expected test exceptions.
- Add logback xml for all test artifacts that don't import core's copy.
- Make sure akka is routing logs through slf4j.
- Make sure log4j is routing logs through slf4j.
- Only print sbt warnings/errors when publishing/pushing artifacts.


engine/src/test/scala/cromwell/engine/workflow/lifecycle/execution/callcaching/CallCacheHashingJobActorSpec.scala
```diff
@@ -19,6 +19,8 @@ import wom.core.LocallyQualifiedName
 import wom.graph.WomIdentifier
 import wom.values.{WomInteger, WomSingleFile, WomString, WomValue}
 
+import scala.util.control.NoStackTrace
```
Horneth: TIL 👍

PR #4135, opened 2018-09-21
There is no escape

### LINKED ISSUE #3990: Regression: Regex replacement in sub command broken in WDL 1.0
```WDL
version 1.0

workflow test_regex {
    call echo {
      input:
    }

}

task echo {
  input {
      String myFile = "dir/dir/bladiebla.fq.gz"
    }
  # Chop of .gz
  String name = sub(basename(myFile), "\\.gz$","")
   
  command {
    echo ${myFile}
    echo ${basename(myFile)}
    echo ${name}  
  }
  output {
    Array[String] result = read_lines(stdout())  
  }   
}
```

Run with cromwell version 34. 
I expect `name` to be `bladiebla.fq` but unfortunately it is `bladiebla.fq.gz`
~~The regex used work in earlier versions of cromwell (31).~~
We  noticed this when changing to WDL 1.0

aednichols:
Closes #3990

~Awaiting accompanying OpenWDL PR with the underlying grammar change (after lunch).~

OpenWDL PR: https://github.com/openwdl/wdl/pull/253

---

Reviewers: is it worth a changelog note to tell people `version draft-3` went away?

I think the change itself is OK because it seems like anyone who signed up to use a draft version is pretty breakage-tolerant.


aednichols:
OpenWDL PR is merged.

cjllanwarne:
NB this is now out of date with the WDL spec which expects `version development`

PR #4156, opened 2018-09-27
Don't delete rendered creds via sbt clean.

kshakir:


kshakir:
Hotfix version: #4157

PR #4157, opened 2018-09-27
Don't delete rendered creds via sbt clean.

kshakir:


kshakir:
Convention/Prudence. Most "how to run tests" tell you to "clean first".

PR #4165, opened 2018-09-27
Paranoia: handle exceptions from 'take'

cjllanwarne:
I haven't seen this happen live but in theory 'take' can throw an exception if the underlying queue changes between checking `available` and running `dequeue`.

That might happen if an unluckily timed `abort` removes an actor from the queue between those checks.

### DIFF
--- a/engine/src/main/scala/cromwell/engine/workflow/tokens/JobExecutionTokenDispenserActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/tokens/JobExecutionTokenDispenserActor.scala
@@ -70,7 +70,7 @@ class JobExecutionTokenDispenserActor(override val serviceRegistryActor: ActorRe
   private def distribute(n: Int) = if (tokenQueues.nonEmpty) {
     val iterator = new RoundRobinQueueIterator(tokenQueues.values.toList, currentTokenQueuePointer)
 
-    // In rare cases, an abort might empty an inner queue between "available" and "dequeue", which causes an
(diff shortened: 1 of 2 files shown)


cjllanwarne:
Note: I labelled this "paranoia" because even without the `Try`, the test passed consistently.

- Maybe this is already protected against elsewhere
- Maybe the test isn't testing the right thing
- Maybe I just got (un)lucky when I ran it

Either way, it doesn't seem like a *bad* thing to be safe in case `take` throws an exception.

engine/src/main/scala/cromwell/engine/workflow/tokens/JobExecutionTokenDispenserActor.scala
```diff
@@ -68,7 +70,15 @@ class JobExecutionTokenDispenserActor(override val serviceRegistryActor: ActorRe
   private def distribute(n: Int) = if (tokenQueues.nonEmpty) {
```
aednichols: [ToL] Would it not be a clearer design (though not any more correct) to synchronize on `tokenQueues`?

cjllanwarne: Hopefully the actor `receive` is synchronizing access to the token queues for us - this is in case that turns out to not be true

cjllanwarne:
Overriding codecov this time - it's ok that the "just in case" path is not run

PR #4167, opened 2018-09-27
Satisfy intellij with newline in Backends.md

cjllanwarne:


cjllanwarne:
Actually, this looks like it got bundled in with another PR. This is now a no-op.

PR #4172, opened 2018-09-28
[34_hotfix] Publish metadata matching PAPI v1 standard stream filenames.

mcovarr:
The PAPI v1 standard stream files that are periodically delocalized are named by the PAPI v1 controller and don't have the regular `stdout` and `stderr` names. So while Cromwell was teeing this data to the standard streams correctly and the PAPI v1 controller was delocalizing correctly, the paths Cromwell published to metadata corresponded to the `stdout` and `stderr` files that are only delocalized when the job completes. `stdout` and `stderr` are arguably more correct than the files periodically delocalized by PAPI v1 since they won't contain output from the wrapper script, but the expectations for FC are that these paths in metadata should point to the files that are periodically delocalized.

### DIFF
--- a/supportedBackends/google/pipelines/common/src/main/scala/cromwell/backend/google/pipelines/common/PipelinesApiInitializationActor.scala
+++ b/supportedBackends/google/pipelines/common/src/main/scala/cromwell/backend/google/pipelines/common/PipelinesApiInitializationActor.scala
@@ -105,7 +105,8 @@ class PipelinesApiInitializationActor(pipelinesParams: PipelinesApiInitializatio
     gcsCred <- gcsCredentials
     genomicsCred <- genomicsCredentials
     validatedPathBuilders <- pathBuilders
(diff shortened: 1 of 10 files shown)


supportedBackends/google/pipelines/common/src/main/scala/cromwell/backend/google/pipelines/common/PipelinesApiInitializationActor.scala
```diff
@@ -128,5 +129,13 @@ class PipelinesApiInitializationActor(pipelinesParams: PipelinesApiInitializatio
     } yield Option(data)
   }
 
+  def standardStreamNameToFileNameMetadataMapper(pipelinesApiJobPaths: PipelinesApiJobPaths, streamName: String): String =
+    PipelinesApiInitializationActor.defaultStandardStreamNameToFileNameMetadataMapper(pipelinesApiJobPaths, streamName)
```
Horneth: `defaultStandardStreamNameToFileNameMetadataMapper` new record ? 😄

mcovarr: 🏆

PR #4174, opened 2018-09-28
Disallow empty File constants

### LINKED ISSUE #4158: Unclear error when dealing with File evaluates to empty string
More details about recreating the error here: https://gatkforums.broadinstitute.org/firecloud/discussion/10740/error-the-local-copy-message-must-have-path-set

Essentially, if a task looks like 
```
task t {
	File x = ""
	
	command {
        ...
	}
	runtime {
        ...
	}
}
```
The job fails with: 
```
BackendJobDescriptorKey_CommandCallNode_w.t:-1:1/CCHashingJobActor-b12fef61-w.t:NA:1] Failed to hash 
cromwell.core.path.PathParsingException: java.lang.IllegalArgumentException: Either  exists on a filesystem not supported by this instance of Cromwell, or a failure occurred while building an actionable path from it. Supported filesystems are: Google Cloud Storage. Failures: Google Cloud Storage:  does not have a gcs scheme (IllegalArgumentException) Please refer to the documentation for more information on how to configure filesystems: http://cromwell.readthedocs.io/en/develop/backends/HPC/#filesystems
	at cromwell.core.path.PathFactory$.$anonfun$buildPath$4(PathFactory.scala:64)
	at scala.Option.getOrElse(Option.scala:121)
	at cromwell.core.path.PathFactory$.buildPath(PathFactory.scala:58)
	at cromwell.core.path.PathFactory.buildPath(PathFactory.scala:30)
```

AC: 
1. In case that a user has set the value of a required File as an empty string --this error message should instead accommodate for this special case and point out that an empty string isn't valid input for a File object. 
   1b. If easy, it would also be nice to remove the link to the HPC config docs and the rest of the info about supported filesystems. 
[Optional] 2. If this doesn't hurt performance somehow, it would be nice if the error message also included the name of the input that failed to hash, not just the name of the call/value of the file.

cjllanwarne:
- Closes #4158 
- [x] @ruchim could you confirm that [this error message](https://github.com/broadinstitute/cromwell/pull/4174/files#diff-aade89887d9abbfe15dc3bf8b809aec5R31) meets your AC?

### DIFF
--- a/backend/src/main/scala/cromwell/backend/backend.scala
+++ b/backend/src/main/scala/cromwell/backend/backend.scala
@@ -39,11 +39,6 @@ case class BackendJobDescriptor(workflowDescriptor: BackendWorkflowDescriptor,
                                 maybeCallCachingEligible: MaybeCallCachingEligible,
                                 prefetchedKvStoreEntries: Map[String, KvResponse]) {
 
-  val overrideme = () => throw new Exception("oopsie")
-  if (1 == (0 + 1)) {
-    overrideme.apply()
-  }
-
   val fullyQualifiedInputs: Map[String, WomValue] = evaluatedTaskInputs map { case (declaration, value) =>
     key.call.identifier.combine(declaration.name).fullyQualifiedName.value -> value
   }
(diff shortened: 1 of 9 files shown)


ruchim:
👍

PR #4178, opened 2018-09-29
Added engine upgrade test for call caching.

kshakir:
Related:
- Allow centaur tests to disable retries.
- Add the `test_timestamp` to bigquery reports.


cjllanwarne:
I assume there's a reason we can't just add the `engine_upgrade` tag to the existing capoeira?

kshakir:
The existing test doesn't have a sleep method long enough for a restart. The cha-cha and capoeira could be perhaps be merged in the future.

The cha-cha in this PR is actually a blend of the capoeira_jes + capoeira_local + capoeira_tes scripts, minus some local docker checks, plus extra sleeps for engine-upgrade restarts.

PR #4194, opened 2018-10-01
Generate and compare perf metrics

salonishah11:
This PR adds a new sub-project to Cromwell. It takes in 2 metadata json files, parses them, generates their metrics and compares them. Comparison right now is simply checking if the new metadata metric has not increased by more than 10% as compared to it's old metric. 

This PR is still WIP.


perf/src/main/scala/perf/CompareMetadata.scala
```diff
@@ -0,0 +1,112 @@
...
+    println(s"Workflow Running time: ${metadata.workflowRunningTime}")
+    println(s"Total jobs per root workflow: ${metadata.totalJobsPerRootWf}")
+    println(s"Avg cache copy retries: ${metadata.avgCacheRetries}")
+    println(s"Avg time job spent in Call Caching state: ${metadata.avgTimeInCallCachingState}")
+    println(s"Avg time job spent in Job Preparation state: ${metadata.avgTimeInJobPreparation}")
+    println(s"Avg time job spent in fetching and copying cache hit(s) state: ${metadata.avgTimeForFetchingAndCopyingCacheHit}")
+  }
+
+  def compareDurationMetrics(metadataOld: Metadata, metadataNew: Metadata, metricFunc: Metadata => Duration, metricName: String): ErrorOr[String] = {
+    if(metricFunc(metadataNew).compareTo(metricFunc(metadataOld).multipliedBy(1.1.toLong)) > 0){
```
kshakir: Re: `1.1.toLong` This is the same as `1L`, so it doesn't actually multiply the amount by anything.

Instead, you should convert each duration `.toMillis`, multiply the `Old` by `1.1`, and then do the comparison.

project/Dependencies.scala
```diff
@@ -518,4 +518,8 @@ object Dependencies {
 
   // Version of the swagger UI to write into config files
   val swaggerUiVersion = swaggerUiV
+
+  val perfDependencies = List(
+    "io.circe" %% "circe-java8" % circeV
```
kshakir: TIL 👍

perf/src/main/scala/perf/Call.scala
```diff
@@ -0,0 +1,69 @@
+package perf
```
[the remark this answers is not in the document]
salonishah11: Good point! Will do 👍

perf/src/main/scala/perf/CompareMetadata.scala
```diff
@@ -0,0 +1,112 @@
...
+  def parseMetadataFromFile(filePath: String): Either[circe.Error, Metadata] = {
+    val metadataFile = File(filePath)
+    val metadataFileContent = metadataFile.contentAsString
+
+    decode[Metadata](metadataFileContent)
+  }
+
+
+  def displayComputedMetrics(metadata: Metadata, displayMsg: String): Unit = {
+    println(displayMsg)
```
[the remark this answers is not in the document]
salonishah11: Yes, logging! It definitely is better than printlns! I don't know why I forgot about it! Thank you for pointing it out :)

perf/src/main/scala/perf/CompareMetadata.scala
salonishah11: Will make the change 👍Thank you :)

perf/src/main/scala/cromwell/perf/Call.scala
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  /**
+    * @return Time (in Duration) job spent in Call Caching states
+    */
+  def timeInCallCachingState: Duration = {
+    val eventsRelatedToCC = executionEvents.filter(event => callCachingEventStates.exists(state => state.equalsIgnoreCase(event.description)))
+
+    if (eventsRelatedToCC.nonEmpty) {
+      val eventsSortedByStartTime = eventsRelatedToCC.sortBy(x => x.startTime)
```
mcovarr: `_.startTime`

perf/src/main/scala/cromwell/perf/Call.scala
```diff
@@ -0,0 +1,69 @@
...
+  val jobPreparationEventStates = List("Pending", "RequestingExecutionToken", "WaitingForValueStore", "PreparingJob", "CheckingJobStore")
+  val cacheCopyingEventStates = List("FetchingCachedOutputsFromDatabase", "BackendIsCopyingCachedOutputs")
+
+  /**
+    * @return Cache copy retries before getting a successful hit or running the job
+    */
+  def cacheCopyRetries: Int = {
+    val hitFailuresMap = callCaching.map(callCachingObject => callCachingObject.hitFailures.getOrElse(Map.empty))
+
+    if (hitFailuresMap.isDefined) hitFailuresMap.get.size else 0
```
mcovarr: TOL

```
    val numFailures: Option[Int] = for {
      caching <- callCaching
      failures <- caching.hitFailures
    } yield failures.size
    
    numFailures.getOrElse(0)
```

perf/src/main/scala/cromwell/perf/Call.scala
```diff
@@ -0,0 +1,69 @@
...
+    else Duration.ZERO
+  }
+
+  /**
+    * @return Time (in Duration) job spent in job preparation state i.e time between job submission and running it
+    *         (this doesn't consider the time it spent in call caching states
+    */
+  def timeInJobPreparation: Duration = {
+    val durationOfEventsInPreparationState = executionEvents.filter(event => jobPreparationEventStates.exists(state => state.equalsIgnoreCase(event.description)))
+      .map(event => Duration.between(event.startTime, event.endTime))
```
mcovarr: TOL `collect` can often do the work of `filter` and `map` in one pass

```
    val durationOfEventsInPreparationState = executionEvents.collect { 
      case e if jobPreparationEventStates.exists(_.equalsIgnoreCase(e.description)) => Duration.between(e.startTime, e.endTime) }
```

perf/src/main/scala/cromwell/perf/CompareMetadata.scala
```diff
@@ -0,0 +1,111 @@
...
+import better.files.File
+import cats.data.Validated.{Invalid, Valid}
+import cats.implicits._
+import common.validation.ErrorOr._
+import io.circe
+import io.circe.generic.auto._
+import io.circe.parser._
+
+// Do not remove this unused import
+// If removed, circe parser fails to find implicit decoder for OffsetDateTime, and parsing fails
```
mcovarr: ugh Spray has this problem too, nice to have the warning here. 🙂

perf/src/main/scala/cromwell/perf/CompareMetadata.scala
```diff
@@ -0,0 +1,111 @@
...
+    println(s"Workflow Running time: ${metadata.workflowRunningTime}")
+    println(s"Total jobs per root workflow: ${metadata.totalJobsPerRootWf}")
+    println(s"Avg cache copy retries: ${metadata.avgCacheRetries}")
+    println(s"Avg time job spent in Call Caching state: ${metadata.avgTimeInCallCachingState}")
+    println(s"Avg time job spent in Job Preparation state: ${metadata.avgTimeInJobPreparation}")
+    println(s"Avg time job spent in fetching and copying cache hit(s) state: ${metadata.avgTimeForFetchingAndCopyingCacheHit}")
+  }
+
+  def compareDurationMetrics(metadataOld: Metadata, metadataNew: Metadata, metricFunc: Metadata => Duration, metricName: String): ErrorOr[String] = {
+    if(metricFunc(metadataNew).toMillis > 1.1 * metricFunc(metadataOld).toMillis){
```
mcovarr: can we pull this 1.1 out to a constant in case we ever decide the magic difference is not 10%?

perf/src/main/scala/cromwell/perf/Metadata.scala
```diff
@@ -0,0 +1,70 @@
...
+    * @return Time between the submission and start of workflow
+    */
+  def workflowStartedAfter: Duration = Duration.between(submission, start)
+
+  def workflowRunningTime: Duration = Duration.between(start, end)
+
+  def totalJobsPerRootWf: Int = sumElementsInOptionSeq(calls.map(taskMap => taskMap.map(callsPerTask => callsPerTask._2.size)), addInt, 0)
+
+  def avgCacheRetries: Int = {
+    val totalJobs = totalJobsPerRootWf
```
mcovarr: why this variable?

perf/src/main/scala/cromwell/perf/Call.scala
salonishah11: This looks more cleaner! Will change it. Thanks :)

perf/src/main/scala/cromwell/perf/Call.scala
salonishah11: It does in this case. Changed it. Thanks!

perf/src/main/scala/cromwell/perf/CompareMetadata.scala
salonishah11: Good point

project/Dependencies.scala
```diff
@@ -518,4 +518,11 @@ object Dependencies {
 
   // Version of the swagger UI to write into config files
   val swaggerUiVersion = swaggerUiV
+
+  val perfDependencies = List(
+    "io.circe" %% "circe-java8" % circeV,
+    "com.typesafe.scala-logging" %% "scala-logging" % scalaLoggingV,
```
kshakir: This is OK, but I think it's time this were just added to `val commonDependencies` and the other copies of this line removed.

perf/src/main/scala/cromwell/perf/Metadata.scala
salonishah11: Removing it 👍

project/Dependencies.scala
salonishah11: I will try and do that 👍

PR #4195, opened 2018-10-02
Ensure that logs write to and close the root wf log.

kshakir:


core/src/main/scala/cromwell/core/logging/WorkflowLogger.scala
```diff
@@ -18,8 +18,14 @@ import scala.util.Try
...
 trait WorkflowLogging extends ActorLogging { this: Actor =>
   def workflowIdForLogging: WorkflowId
-
-  lazy val workflowLogger = new WorkflowLogger(self.path.name, workflowIdForLogging, Option(log))
+  def rootWorkflowIdForLogging: WorkflowId
+
+  lazy val workflowLogger = new WorkflowLogger(
+    self.path.name,
+    workflowIdForLogging,
+    rootWorkflowIdForLogging,
```
mcovarr: TOL use keyword args to make sure the consecutive same-typed workflow ID params are being passed correctly?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/initialization/WorkflowInitializationActor.scala
```diff
@@ -45,18 +45,27 @@ object WorkflowInitializationActor {
...
+  def props(workflowIdForLogging: WorkflowId,
+            rootWorkflowIdForLogging: WorkflowId,
             workflowDescriptor: EngineWorkflowDescriptor,
             ioActor: ActorRef,
             serviceRegistryActor: ActorRef,
             restarting: Boolean): Props = {
-    Props(new WorkflowInitializationActor(workflowId, workflowDescriptor, ioActor, serviceRegistryActor, restarting)).withDispatcher(EngineDispatcher)
+    Props(new WorkflowInitializationActor(
+      workflowIdForLogging,
+      rootWorkflowIdForLogging,
```
mcovarr: TOL hmm now I'm almost wondering if they should be wrapped in distinct "root" and "possibly not root" types. Up to you...

kshakir:
> I'm almost wondering if they should be wrapped in distinct "root" and "possibly not root" types

Done, plus a squash and rebase. Will merge after tests pass unless the new types aren't what you were imagining.

PR #4198, opened 2018-10-02
[Release WDL] Split draft from publish to confirm creds early

cjllanwarne:
By making the draft release before building Cromwell, we will find out early if there's a problem with the auth token

### DIFF
--- a/release/release_workflow.wdl
+++ b/release/release_workflow.wdl
@@ -124,25 +124,19 @@ task versionPrep {
     }
 }
 
-task makeGithubRelease {
+task draftGithubRelease {
     input {
         String githubToken
         String organization
-        File cromwellJar
-        File womtoolJar
         String oldVersion
         String newVersion
     }
-
-    String cromwellJarName = basename(cromwellJar)
-    String womtoolJarName = basename(womtoolJar)
-
     command <<<
         set -e
         set -x
 
         # download changelog from master
-        curl https://raw.githubusercontent.com/~{organization}/cromwell/master/CHANGELOG.md -o CHANGELOG.md
+        curl --fail https://raw.githubusercontent.com/~{organization}/cromwell/master/CHANGELOG.md -o CHANGELOG.md
 
         # Extract the latest piece of the changelog corresponding to this release
         # head remove the last line, next sed escapes all ", and last sed/tr replaces all new lines with \n so it can be used as a JSON string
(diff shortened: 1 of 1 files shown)


release/release_workflow.wdl
```diff
@@ -124,25 +124,19 @@ task versionPrep {
...
-    String cromwellJarName = basename(cromwellJar)
-    String womtoolJarName = basename(womtoolJar)
-
     command <<<
         set -e
         set -x
 
         # download changelog from master
-        curl https://raw.githubusercontent.com/~{organization}/cromwell/master/CHANGELOG.md -o CHANGELOG.md
+        curl --fail https://raw.githubusercontent.com/~{organization}/cromwell/master/CHANGELOG.md -o CHANGELOG.md
```
aednichols: [ToL] It looks like `--fail` has the side effect of suppressing error output - could make things harder to debug (last time, we figured out the problem from the output).

But `--fail` >> nothing so 🤷‍♂️

release/release_workflow.wdl
```diff
@@ -308,46 +334,57 @@ workflow release_cromwell {
...
-    }
+  if (draftGithubRelease.draft_complete) {
+      if (majorRelease) {
+        call do_major_release { input:
+                organization = organization,
+                releaseVersion = cromwellVersion
+        }
+      }
+
+      if (!majorRelease) {
```
Horneth: When is `else` coming ?! 😄

mcovarr: 🔥

release/release_workflow.wdl
```diff
@@ -113,8 +106,10 @@ task versionPrep {
     }
 
     command <<<
+      set -e
+
       which jq || brew install jq
-      curl -s https://api.github.com/repos/~{organization}/cromwell/releases/latest | jq --raw-output '.tag_name' > version
+      curl --fail -v -s https://api.github.com/repos/~{organization}/cromwell/releases/latest | jq --raw-output '.tag_name' > version
```
mcovarr: what's the net effect of `--fail` (fail silently) and `-v` (be verbose)?

cjllanwarne: - `--fail` forces curl to change its exit code if something goes wrong
- `-v` means that we don't lose all context if that happens

PR #4205, opened 2018-10-04
Add retry on abort option

ffinfo:
Depends on: https://github.com/broadinstitute/cromwell/pull/4112


geoffjentry:
@ffinfo Just to double check, are all of those commits real or does this require a rebase or something like that?

ffinfo:
@geoffjentry this is because I did branch from a other PR, see also https://github.com/broadinstitute/cromwell/pull/4112

Some commits are already merged but because you guys doing squash merge the commit history is lost so this PR thinks those commits are new ;)

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -1075,7 +1075,10 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
               val executionHandle = Future.successful(FailedNonRetryableExecutionHandle(StderrNonEmpty(jobDescriptor.key.tag, stderrSize, stderrAsOption), Option(returnCodeAsInt)))
               retryElseFail(status, executionHandle)
             case Success(returnCodeAsInt) if isAbort(returnCodeAsInt) =>
-              Future.successful(AbortedExecutionHandle)
+              val executionHandle = FailedNonRetryableExecutionHandle(StderrNonEmpty(jobDescriptor.key.tag, stderrSize, stderrAsOption), Option(returnCodeAsInt))
+              if (jobDescriptor.workflowDescriptor.workflowOptions.retryAbortedJobs)
+                retryElseFail(status, Future.successful(executionHandle))
```
Horneth: This means that if (and only if) `retryAbortedJobs` is enabled and the `maxRetries` runtime attribute is set to > 0, any aborted job (even user triggered) will be retried, which is not desired.
I think this should differentiate between "the backend unexpectedly killed this job and we want to retry it" and "the user aborted the workflow".
I'm wondering if the config backend should not send back a `FailedRetryableExecutionHandle` instead of writing a made up return code and returning "Failed"

ruchim: @ffinfo just pinging you to see you've had a chance to see the comment above and get your thoughts on it. Thank you!

gemmalam:
Closing due to stale PR. Please reopen if you would like to add more work to this.

PR #4206, opened 2018-10-04
Enable WB health monitor to report status on PAPIV1 + PAPIV2

kshakir:


centaur/src/main/resources/standardTestCases/workbench_health_monitor_check/workbench_health_monitor_check.wdl
```diff
@@ -0,0 +1,17 @@
...
+    call run
+    output {
+        String out = run.out
+    }
+}
+
+task run {
+    command {
+        set -euo pipefail
+        curl http://localhost:8008/engine/v1/status | jq -crMS .
```
Horneth: 😄 Does the Travis VM have `jq` installed ?

kshakir: jq is [_everywhere_](https://docs.travis-ci.com/user/build-environment-updates/2014-12-09/#additions)!!!

kshakir:
FYI the new test should pass on V2 after https://issuetracker.google.com/issues/117213564 is resolved.

src/ci/resources/papi_application.inc.conf.ctmpl
```diff
@@ -43,3 +43,16 @@ google {
...
 }
 {{end}}
+
+services {
+  HealthMonitor {
+    class = "cromwell.services.healthmonitor.impl.workbench.WorkbenchHealthMonitorServiceActor"
+	config {
+	  papi-backend-name = "Papi"
+	  papi-v1-backend-name = "Papi"
+	  papi-v2-backend-name = "Papi"
```
cjllanwarne: `Papi2`, surely?

centaur/src/main/resources/standardTestCases/workbench_health_monitor_check.test
```diff
@@ -0,0 +1,13 @@
...
+backends: [Papi]
+
+files {
+  workflow: workbench_health_monitor_check/workbench_health_monitor_check.wdl
+}
+
+metadata {
+  workflowName: workbench_health_monitor_check
+  status: Succeeded
+  "outputs.workbench_health_monitor_check.out": """{"DockerHub":{"ok":true},"Engine Database":{"ok":true},"GCS":{"ok":true},"PAPI":{"ok":true},"PAPIV1":{"ok":true},"PAPIV2":{"ok":true}}"""
```
cjllanwarne: TBH I think this test is a little misleading?

It's checking that PAPIV2 responds `true` if the PAPI2 checker runs against `PAPI1`, on a Cromwell that only has the PAPI1 backend?

src/ci/resources/papi_application.inc.conf.ctmpl
kshakir: See below

centaur/src/main/resources/standardTestCases/workbench_health_monitor_check.test
kshakir: Yep.

This config is shared between both papi centaur setups. There is no current centaur that runs both Papi's currently. That would be yet-another-centaur matrix entry of PapiV1 **&** PapiV2.

The test is confirming that one can mix and match health monitor _names_ any way one likes.

These are each just names returned by the WHMSA, and can connect to any Papi backend:
`PAPI` + `PAPIV2`
`PAPI` + `PAPIV1`
`PAPIV1` + `PAPIV2`
`PAPI` + `PAPIV1` + `PAPIv2` (ultimate in backwards compatibility)

ToL: Ultimately all this "Workbench" specific code should probably be refactored so that others can use it anyway. I suspect more flexibility will come once AWS, possibly Dockstore, and other services need to be checked in a configurable way. I don't think it's going to be with the hardcoded list as it is now, where for every change in the engine we also update this class. There's also CaaS on the horizon too. There's also talk of soft warnings too-- example: "PAPI is not responding for the moment, but cromwell is up and will keep checking" vs. "The database is down and cromwell cannot accept submissions"

Issue #4204, opened 2018-10-04
missing config header


notestaff:
Does the config file https://github.com/broadinstitute/cromwell/blob/develop/scripts/docker-compose-mysql/compose/cromwell/app-config/application.conf#L1
need the header
include required(classpath("application"))
at the top?

kshakir:
Yes.

This pulls in the cromwell "server" [application.conf](https://github.com/broadinstitute/cromwell/blob/1dc56871a958c3c2cfdf1a59c6ccc2bec3968300/server/src/main/resources/application.conf) with a few required overrides of akka's various reference.conf files, and reports an error if for some reason the file is not found.

https://github.com/broadinstitute/cromwell/blob/1dc56871a958c3c2cfdf1a59c6ccc2bec3968300/server/src/main/resources/application.conf#L1-L23

PR #4207, opened 2018-10-04
Add missing config header

cjllanwarne:
Per #4204 (thanks for spotting, @notestaff!)


kshakir:
This is great. To make sure it works (and stays working) I'm going to add it to the [existing `dockerScripts`](https://github.com/broadinstitute/cromwell/blob/541636734705b7d93321a31ac5817f96b275eb0f/src/ci/bin/testDockerScripts.sh#L29-L39) tests this weekend.

PR #4208, opened 2018-10-04
Don't really glob when evaluating files

Horneth:
The `evaluateFiles` method is used to figure out what files are going to be produced or are referenced by an expression, before the task has been run.
In this context it doesn't make sense to do the work of evaluating `glob` which will fail since the task hasn't run.
Instead, fail immediately. When that happens, we fallback to a second algorithm to evaluate the files ([draft-2](https://github.com/broadinstitute/cromwell/blob/541636734705b7d93321a31ac5817f96b275eb0f/wdl/model/draft2/src/main/scala/wdl/draft2/model/expression/FileEvaluator.scala#L52), [draft-3](https://github.com/broadinstitute/cromwell/blob/541636734705b7d93321a31ac5817f96b275eb0f/wdl/model/draft3/src/main/scala/wdl/model/draft3/graph/expression/FileEvaluator.scala#L25))

This should fix the `dontglobinputs` transient centaur failures. With the caveat that the real underlying issue is this: https://github.com/broadinstitute/cromwell/issues/4209

### DIFF
--- a/backend/src/main/scala/cromwell/backend/io/DirectoryFunctions.scala
+++ b/backend/src/main/scala/cromwell/backend/io/DirectoryFunctions.scala
@@ -10,8 +10,8 @@ import cromwell.backend.BackendJobDescriptor
 import cromwell.backend.io.DirectoryFunctions.listFiles
 import cromwell.core.io.AsyncIoFunctions
 import cromwell.core.path.{Path, PathFactory}
-import wom.expression.{IoFunctionSet, NoIoFunctionSet}
 import wom.expression.IoFunctionSet.{IoDirectory, IoElement, IoFile}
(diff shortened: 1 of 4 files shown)


cjllanwarne:
This seems to be fixing the symptom rather than the cause?

Is it not the file evaluators themselves which should be evaluating things that look like `glob` into `WomGlobFile`s immediately rather than ever calling down into the `IoFunction`s?

Horneth:
@cjllanwarne That would be another way to do it. I don't know if the net result would be different though ?

cjllanwarne:
@Horneth Oh, I forgot that we always try to `evaluate` first, then predict second.

PR #4214, opened 2018-10-05
Fix lowestCommonSubtype order within complex types

### LINKED ISSUE #4131: Issue coercing Int to String inside of a Map
I'm getting an error from Cromwell `Failed to coerce one or more keys or values for creating a Map[String, Int?]` when defining an object as an input to a task that consists of a string, file and int. I would expect the Int to get converted into a string, but the error makes it seem as if Cromwell is trying to convert all of the values into integers.

Here is the part of the WDL that's throwing the error: https://github.com/HumanCellAtlas/pipeline-tools/blob/c949cb5ffa2df8f2a7fc7d7a4c34478e8eadbf34/adapter_pipelines/cellranger/adapter.wdl#L222

cjllanwarne:
Superficially fixes #4131 since there was a genuine problem, but see the comment on the issue about preferring `object { }` when creating objects (especially when they have non-homogeneous fields).

### DIFF
--- a/wom/src/main/scala/wom/types/WomType.scala
+++ b/wom/src/main/scala/wom/types/WomType.scala
@@ -115,13 +115,19 @@ object WomType {
         firstCommonPrimitive(types)
       } else None
     }
+
+    val coercePriority = List(
+      WomStringType, WomSingleFileType, WomUnlistedDirectoryType, WomFloatType, WomIntegerType, WomBooleanType, WomObjectType
+    )
+
     private def firstCommonPrimitive(types: Iterable[WomType]): Option[WomType] = {
-      // See if any of the types are a good candidate:
-      val options = types.filter(t => types.forall(_.isCoerceableFrom(t)))
-      womTypeCoercionOrder.collectFirst {
-        // Primitives:
-        case p: WomPrimitiveType if options.toList.contains(p) => p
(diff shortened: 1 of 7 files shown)


wom/src/main/scala/wom/types/WomType.scala
```diff
@@ -98,10 +98,112 @@ object WomType {
...
+        val leftType = lowestCommonSubtype(pairs.map(_.leftType))
+        val rightType = lowestCommonSubtype(pairs.map(_.rightType))
+        Some(WomPairType(leftType, rightType))
+      } else None
+    }
+  }
+
+  private object ListOfOptionals {
+    def unapply(types: Iterable[WomType]): Option[WomOptionalType] = {
+      val atLeastOneOptional = types.exists {
```
aednichols: This could be shorter and (imo) clearer as `types.exists(_.isInstanceOf[WomOptionalType])`

cjllanwarne: Yes, in this case... the reason for doing this at all is that some types have custom `unapply` methods (eg `WomArray` which also scoops up `WomMap`s as arrays of pairs).

But perhaps it's best to make that explicit rather than forcing everything else through the same weird pattern

PR #4220, opened 2018-10-10
Only execute isAlive once per timeout

ffinfo:
Follow up on https://github.com/broadinstitute/cromwell/pull/4112

This will reduce the load on the JVM a lot

 I did indeed a stress test on our system with 50.000 async qsub/qstat jobs but this was outside the jvm. Inside the jvm this ends up in blocking threads to cromwell.

When the timeout is set to 120 seconds, `isAlive` will only run once each 120 seconds.

### DIFF
--- a/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
+++ b/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
@@ -1006,8 +1006,7 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
       // the state names.
       // This logging and metadata publishing assumes that StandardAsyncRunState subtypes `toString` nicely to state names.
       val prevStatusName = previousState.map(_.toString).getOrElse("-")
-      if (prevStatusName == state.toString) jobLogger.debug(s"Status change from $prevStatusName to $state")
-      else jobLogger.info(s"Status change from $prevStatusName to $state")
+      jobLogger.info(s"Status change from $prevStatusName to $state")
       tellMetadata(Map(CallMetadataKeys.BackendStatus -> state))
     }
 
--- a/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
+++ b/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
@@ -1002,7 +1002,8 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
       the state names.
        */
       val prevStateName = previousStatus.map(_.toString).getOrElse("-")
-      jobLogger.info(s"Status change from $prevStateName to $status")
+      if (prevStateName == status.toString) jobLogger.debug(s"Status change from $prevStateName to $status")
+      else jobLogger.info(s"Status change from $prevStateName to $status")
       tellMetadata(Map(CallMetadataKeys.BackendStatus -> status))
     }
 
--- a/supportedBackends/sfs/src/main/scala/cromwell/backend/impl/sfs/config/ConfigAsyncJobExecutionActor.scala
+++ b/supportedBackends/sfs/src/main/scala/cromwell/backend/impl/sfs/config/ConfigAsyncJobExecutionActor.scala
@@ -250,8 +250,8 @@ class DispatchedConfigAsyncJobExecutionActor(override val standardParams: Standa
     jobScriptArgs(job, "kill", KillTask)
   }
 
-  protected lazy val exitCodeTimeout: Option[Int] = {
-    val timeout = configurationDescriptor.backendConfig.as[Option[Int]](ExitCodeTimeoutConfig)
+  protected lazy val exitCodeTimeout: Option[Long] = {
+    val timeout = configurationDescriptor.backendConfig.as[Option[Long]](ExitCodeTimeoutConfig)
     timeout.foreach{x => if (x < 0) throw new IllegalArgumentException(s"config value '$ExitCodeTimeoutConfig' must be 0 or higher")}
     timeout
   }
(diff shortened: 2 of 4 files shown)


ffinfo:
oke this also not fixing the cpu load, I'm trying t debug this but can't see where the load comes from. Maybe the calander call?

Right now I'm trying to create a extra poll queue just for isAlive. Altough with the the status is not passed anymore to the normal polling. I think this is because the handle stays in memory till the jobActor is finished, starting a new queue mean a copy of the handle.

ffinfo:
@geoffjentry We came to the conclusion that my earlier change did not cause the performance issue. I will make a new issue for this performence

Still this code I made here is useful to merge and is ready to review/merge.

supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystemAsyncJobExecutionActor.scala
```diff
@@ -17,6 +17,12 @@ import scala.util.{Failure, Success, Try}
 
 case class SharedFileSystemRunStatus(status: String, date: Calendar) {
   override def toString: String = status
+
+  def experired(timeoutSeconds: Int): Boolean = {
+    val currentDate = Calendar.getInstance()
```
Horneth: I think something like `LocalDateTime` would make this a bit more readable:

```scala
def expired(timeoutInSeconds: Int) = {
    LocalDateTime
      .ofInstant(date.toInstant, ZoneId.systemDefault())
      .plusSeconds(timeoutInSeconds)
      .isBefore(LocalDateTime.now())
  }
```

supportedBackends/sfs/src/main/scala/cromwell/backend/impl/sfs/config/ConfigAsyncJobExecutionActor.scala
```diff
@@ -268,22 +266,24 @@ class DispatchedConfigAsyncJobExecutionActor(override val standardParams: Standa
...
         // Exitcode file does not exist at this point, checking is jobs is still alive
-        if (exitCodeTimeout.isEmpty) s
-        else if (isAlive(handle.pendingJob).fold({ e =>
-          log.error(e, s"Running '${checkAliveArgs(handle.pendingJob).argv.mkString(" ")}' did fail")
-          true
-        }, x => x)) s
-        else SharedFileSystemRunStatus("WaitingForReturnCode")
+        exitCodeTimeout match {
+          case Some(timeout) =>
+            if (s.experired(timeout)) s
```
Horneth: I'm not sure I understand the logic, this seems to say "if we're past the timeout, do nothing, otherwise check if the job is still alive". Isn't that the opposite of the intent ?

ffinfo:
@Horneth Sorry for the spam because of the github issue ;)

I did process the comments, could you look at it again?

cpavanrun:
Looking forward to this feature. When enabled the `check-alive` command call via `exit-code-timeout-seconds` currently polls on average once every 10 seconds per running job (under minimum load).

backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -1006,7 +1006,8 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
       // the state names.
       // This logging and metadata publishing assumes that StandardAsyncRunState subtypes `toString` nicely to state names.
       val prevStatusName = previousState.map(_.toString).getOrElse("-")
-      jobLogger.info(s"Status change from $prevStatusName to $state")
+      if (prevStatusName == state.toString) jobLogger.debug(s"Status change from $prevStatusName to $state")
```
Horneth: I don't think this change is needed, the `if (!(previousState exists statusEquivalentTo(state)))` on line 1004 already makes sure we log this at the right time

ffinfo: Was not aware this was changed. All I did is fixing the merge conflicts ;)
Did remove this change again, seems to work fine indeed.

ffinfo:
@Horneth @cpavanrun 
Should be good to merge now. Only still added something to the changelog.

supportedBackends/sfs/src/main/scala/cromwell/backend/impl/sfs/config/ConfigAsyncJobExecutionActor.scala
```diff
@@ -252,54 +250,45 @@ class DispatchedConfigAsyncJobExecutionActor(override val standardParams: Standa
...
-              val writer = new PrintWriter(jobPaths.returnCode.toFile)
-              // 137 does mean a external kill -9, this is a assumption but easy workaround for now
-              writer.println(9)
-              writer.close()
-              SharedFileSystemRunState("Failed")
-            }
-          case _ => s
-        }
-      case Some(s) if s.status == "Done" => s // Nothing to be done here
+        jobLogger.error(s"Return file not found after ${exitCodeTimeout.getOrElse("-")} seconds, assuming external kill")
```
ruchim: Just thinking out loud -- at this point, the exitCodeTimeout has to exist, is that right? We can't enter this case unless the job has exceeded the timeout? So theoretically one should never see an error that looks like: `Return file not found after - seconds, assuming external kill"`

ruchim:
👍 

Something to consider for the future-- a user may want to see how often their job failed due to timeouts, so it might be interesting for this to be marked as a new state other than Failed, but it works perfectly well for the goal of this PR.

supportedBackends/sfs/src/main/scala/cromwell/backend/impl/sfs/config/ConfigAsyncJobExecutionActor.scala
ffinfo: This is true when `exit-code-timeout-seconds` is not yet. Once this is set the value will become Some(true) if the timeout is passed, and so enter this case.
Under the right conditions and a job will never be lost this case will be never touched. Sadly not all system are that reliable ;)

PR #4228, opened 2018-10-10
Un-hardcode AWS Region

danbills:
Use the config option where appropriate

### DIFF
--- a/cloudSupport/src/main/scala/cromwell/cloudsupport/aws/auth/AwsAuthMode.scala
+++ b/cloudSupport/src/main/scala/cromwell/cloudsupport/aws/auth/AwsAuthMode.scala
@@ -74,7 +74,12 @@ sealed trait AwsAuthMode {
    private[auth] var credentialValidation: ((AwsCredentials, Option[String]) => Unit) =
      (credentials: AwsCredentials, region: Option[String]) => {
        val builder = STSClient.builder
-       region.foreach(str => builder.region(Region.of(str)))
+
+       //If the region argument exists in config, set it in the builder.
+       //Otherwise it is left unset and the AwsCredential builder will look in various places to supply,
+       //ultimately using US-EAST-1 if none is found
+       region.map(Region.of).foreach(builder.region)
+
        builder.credentialsProvider(StaticCredentialsProvider.create(credentials))
          .build
          .getCallerIdentity(GetCallerIdentityRequest.builder.build)
--- a/cloudSupport/src/main/scala/cromwell/cloudsupport/aws/auth/AwsAuthMode.scala
+++ b/cloudSupport/src/main/scala/cromwell/cloudsupport/aws/auth/AwsAuthMode.scala
@@ -71,18 +71,17 @@ sealed trait AwsAuthMode {
     * All traits in this file are sealed, all classes final, meaning things
     * like Mockito or other java/scala overrides cannot work.
     */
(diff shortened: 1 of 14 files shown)


wleepang:
Agree with @elerch .

Also, should occurances of `Region.AP_NORTHEAST_1` actually be `Region.US_EAST_1` for this backend?

danbills:
@elerch 
1) changed to `region`
2) agree w/ approach. I'm going to make the config be an Option[Region], then if none will omit the `region()` builder call

elerch:
@danbills - changes lgtm!

cloudSupport/src/main/scala/cromwell/cloudsupport/aws/auth/AwsAuthMode.scala
```diff
@@ -71,18 +71,17 @@ sealed trait AwsAuthMode {
...
-       STSClient
-          .builder
-          .region(Region.of(region))
-          .credentialsProvider(StaticCredentialsProvider.create(credentials))
-          .build
-          .getCallerIdentity(GetCallerIdentityRequest.builder.build)
+   private[auth] var credentialValidation: ((AwsCredentials, Option[String]) => Unit) =
+     (credentials: AwsCredentials, region: Option[String]) => {
+       val builder = STSClient.builder
+       region.foreach(str => builder.region(Region.of(str)))
```
mcovarr: I have to admit it took me a bit to grok that this is conditionally setting the region into the statefully built builder... maybe a comment?

filesystems/s3/src/test/scala/cromwell/filesystems/s3/S3PathBuilderSpec.scala
```diff
@@ -49,7 +50,8 @@ class S3PathBuilderSpec extends TestKitSuite with FlatSpecLike with Matchers wit
     S3PathBuilder.fromCredentials(
       AnonymousCredentialsProvider.create.getCredentials,
       S3Storage.s3AdvancedConfiguration(),
-      wfOptionsWithProject
+      wfOptionsWithProject,
+      Some(Region.US_EAST_1)
```
mcovarr: `Option` here and below

supportedBackends/aws/src/main/scala/cromwell/backend/impl/aws/AwsBatchJob.scala
```diff
@@ -62,33 +62,36 @@ import scala.util.Try
...
-    //TODO: setting this to get past batch job unit test, should be configured externally
-    .region(Region.US_EAST_1)
-                   // .credentialsProvider(...)
-                   // .endpointOverride(...)
-                   .build
+  lazy val client = {
+   val builder = BatchClient.builder()
+   configRegion.foreach(builder.region)
+      // .credentialsProvider(...)
+      // .endpointOverride(...)
```
mcovarr: ✖️

supportedBackends/aws/src/main/scala/cromwell/backend/impl/aws/AwsBatchJob.scala
```diff
@@ -62,33 +62,36 @@ import scala.util.Try
...
+   builder.build
+  }
+  lazy val logsclient = {
+    val builder = CloudWatchLogsClient.builder()
+    configRegion.foreach(builder.region)
+      // .credentialsProvider(...)
+      // .endpointOverride(...)
```
mcovarr: ✖️

PR #4231, opened 2018-10-11
Add Span Scale Factor to fix BatchActorSpec flaky test

### LINKED ISSUE #4223: Fix flakey test from QA testbed
```
The code passed to eventually never returned normally. Attempted 14 times over 5.123310162 seconds. Last failure message: Vector("hola", "hello", "bonjour") was not equal to Vector("hola", "hello", "bonjour", "aurevoir").
```

```
tc: BatchingDbWriter should process again when previous processing finished and we're still over batch size (each time with same test case)
```

https://fc-jenkins.dsp-techops.broadinstitute.org/job/cromwell-test-runner/161/testReport/cromwell.core.actor/BatchActorSpec/BatchingDbWriter_should_process_again_when_previous_processing_finished_and_we_re_still_over_batch_size/

If one needs more info, please talk to @ndbolliger

salonishah11:
Closes #4223

### DIFF
--- a/src/ci/bin/test.inc.sh
+++ b/src/ci/bin/test.inc.sh
@@ -112,7 +112,7 @@ cromwell::private::create_build_variables() {
             CROMWELL_BUILD_MYSQL_PASSWORD=""
             CROMWELL_BUILD_MYSQL_SCHEMA="cromwell_test"
             CROMWELL_BUILD_GENERATE_COVERAGE=false
-            CROMWELL_BUILD_SPAN_SCALE_FACTOR="10"
+            CROMWELL_BUILD_SPAN_SCALE_FACTOR="12"
             ;;
         *)
             CROMWELL_BUILD_IS_CI=false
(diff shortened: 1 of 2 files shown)


src/ci/bin/test.inc.sh
```diff
@@ -112,6 +112,7 @@ cromwell::private::create_build_variables() {
             CROMWELL_BUILD_MYSQL_PASSWORD=""
             CROMWELL_BUILD_MYSQL_SCHEMA="cromwell_test"
             CROMWELL_BUILD_GENERATE_COVERAGE=false
+            CROMWELL_BUILD_SPAN_SCALE_FACTOR="12"
```
salonishah11: When I ran the test today, it failed for 1 out of 5 runs. So I have increased the factor from 10->12 (it hasn't failed for 12 yet)

cjllanwarne: Maybe make it 15 or even 20 to be safe?

- I believe these are backstops in case something goes wrong: 
  - In the "normal" case, even making this 10,000 shouldn't slow down the tests (though nb: *shouldn't*) 
  - What making this higher means is that if a test fails because some condition never gets met, we would wait longer before declaring that it will never happen

salonishah11: Sounds good. I will increase it to 15.

PR #4243, opened 2018-10-12
Insert a regulator between the Cromwell API and the spawning of MetadataBuilderActors.

mcovarr:


engine/src/main/scala/cromwell/webservice/CromwellApiService.scala
```diff
@@ -323,9 +323,10 @@ trait CromwellApiService extends HttpInstrumentation {
     }
   }
 
+  private lazy val metadataBuilderRegulatorActor = actorRefFactory.actorOf(MetadataBuilderRegulatorActor.props(serviceRegistryActor))
```
cjllanwarne: Would it make sense to push the regulator behind the service registry rather than having it awkward inserted in front?

engine/src/main/scala/cromwell/webservice/metadata/MetadataBuilderRegulatorActor.scala
```diff
@@ -0,0 +1,57 @@
...
+              requesters foreach { _ ! response}
+            case None =>
+              // unpossible: there had to have been a request that corresponded to this response
+              log.error(s"MetadataBuilderRegulatorActor unpossible error: no requesters found for action: $action")
+          }
+          builderRequests.remove(sndr)
+          ()
+        case None =>
+          // unpossible: this actor should know about all the child MetadataBuilderActors it has begotten
+          log.error(s"MetadataBuilderRegulatorActor unpossible error: unrecognized sender ${sender()}")
```
cjllanwarne: Should be `sndr`

engine/src/main/scala/cromwell/webservice/metadata/MetadataBuilderRegulatorActor.scala
```diff
@@ -0,0 +1,57 @@
...
+        metadataBuilderActor ! action
+      }
+    case response: MetadataBuilderActorResponse =>
+      val sndr = sender()
+      builderRequests.get(sndr) match {
+        case Some(action) =>
+          apiRequests.get(action) match {
+            case Some(requesters) =>
+              apiRequests.remove(action)
+              requesters foreach { _ ! response}
```
cjllanwarne: If the requesters die on timeout, will we end up with lots of dead letter errors here?

cjllanwarne: Metrics-wise, would we be able to record:
- The number of requests being responded to
- How long the oldest requested has been waiting

mcovarr: We can certainly open an enhancement issue for those.

mcovarr: I don't think any more than today.

engine/src/main/scala/cromwell/webservice/metadata/MetadataBuilderRegulatorActor.scala
mcovarr: yup

PR #4246, opened 2018-10-12
Refactor metadata facing routes into separate trait

geoffjentry:
This is mainly just moving stuff around, pulling the metadata based routes into a separate trait out of CromwellApiService. 

The **meat** of this was peeling out the inner bits of the metadata route logic into separate/composable functions so that I can access them via other incoming work. 

Because @mcovarr are actively poking at the same areas of the code I wanted to get this more basic refactor out ASAP and can follow up with further refactoring at some later point when he's done.


geoffjentry:
@Horneth yeah, it seemed too coincidental to not be directly related. otoh i finally got around to looking at it myself and of course i am not able to reproduce it locally. hmmm

geoffjentry:
it also doesn't make sense that anything changed, outside of the "they both involve metadata", since my changes are upstream of the tests AFAICT.  Let me poke around and see if anyone else has seen this, it's failing pretty consistently for me in travis.

PR #4256, opened 2018-10-15
Only compare the data in StandardAsyncRunStatus that represents the status.

mcovarr:
Addresses an issue with recent changes to `SharedFileSystemRunStatus` to include a timestamp. Since the timestamps on two different `SharedFileSystemRunStatus` objects are never the same, two different `SharedFileSystemRunStatus` objects are never the same. This led to Cromwell logging and publishing metadata events for every status poll even if the status hadn't changed.

This is not a great fix since (IMHO) the real problem is the modeling of a status type to include data that does not actually represent the status, but this fix is much narrower than reversing that design change.

### DIFF
--- a/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
+++ b/backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
@@ -85,10 +85,15 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
   type StandardAsyncRunInfo
 
   /** The type of the run status returned during each poll. */
-  type StandardAsyncRunStatus
+  type StandardAsyncRunState
+
+  /** Should return true if the status contained in `thiz` is equivalent to `that`, delta any other data that might be carried around
+    * in the state type.
+    */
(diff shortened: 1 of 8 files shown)


backend/src/main/scala/cromwell/backend/standard/StandardAsyncExecutionActor.scala
```diff
@@ -87,6 +87,11 @@ trait StandardAsyncExecutionActor extends AsyncBackendJobExecutionActor with Sta
   /** The type of the run status returned during each poll. */
   type StandardAsyncRunStatus
 
+  /** True if the status contained in `thiz` is equivalent to `that`, delta any other data that might be carried around
+    * in the status type but which doesn't actually represent the status.
+    */
+  def equivalentTo(thiz: StandardAsyncRunStatus)(that: StandardAsyncRunStatus): Boolean = thiz == that
```
aednichols: I think this comment is not accurate for the default implementation - it compares everything

mcovarr: Yes you're right, will fix.

PR #4260, opened 2018-10-16
Reduce number of logged "unhandled" messages

rsasch:
If someone wants to see these kinds messages, they can change the configuration of akka (see http://doc.akka.io/docs/akka/current/java/logging.html#Auxiliary_logging_options), instead.

- went through the `whenUnhandled` handlers and removed logging calls that had no or very little info
- removed one test that was associated with one of those logs

### DIFF
--- a/backend/src/main/scala/cromwell/backend/standard/callcaching/StandardCacheHitCopyingActor.scala
+++ b/backend/src/main/scala/cromwell/backend/standard/callcaching/StandardCacheHitCopyingActor.scala
@@ -217,9 +217,6 @@ abstract class StandardCacheHitCopyingActor(val standardParams: StandardCacheHit
   whenUnhandled {
     case Event(AbortJobCommand, _) =>
       abort()
-    case Event(unexpected, _) =>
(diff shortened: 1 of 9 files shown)


mcovarr:
Are these being processed by a default handler now or just gone?

rsasch:
From discussion with @geoffjentry, the purpose is to take this kind of logging out of the code; if someone wants to see these messages, they change the configuration of akka ( see http://doc.akka.io/docs/akka/current/java/logging.html#Auxiliary_logging_options).

geoffjentry:
I think perhaps it's worth sticking that in our default `application.conf` ... people can disable it if for some reason they're running `DEBUG` logging and find that small amount of extra noise not worth it.

rsasch:
@geoffjentry I added `akka.actor.debug.handled = on` to  the application.conf file (under server/src/main/resources); is that what you had in mind?

aednichols:
Looks like `SingleWorkflowRunnerActorSpec` is failing for a fairly legitimate reason - it never receives the warning it's checking for

PR #4261, opened 2018-10-16
Export Span Scale factor to sbt for Jenkins

### LINKED ISSUE #4223: Fix flakey test from QA testbed

salonishah11:
Closes #4223

### DIFF
--- a/src/ci/bin/test.inc.sh
+++ b/src/ci/bin/test.inc.sh
@@ -112,7 +112,7 @@ cromwell::private::create_build_variables() {
             CROMWELL_BUILD_MYSQL_PASSWORD=""
             CROMWELL_BUILD_MYSQL_SCHEMA="cromwell_test"
             CROMWELL_BUILD_GENERATE_COVERAGE=false
-            CROMWELL_BUILD_SPAN_SCALE_FACTOR="15"
+            CROMWELL_BUILD_SPAN_SCALE_FACTOR="20"
             ;;
         *)
             CROMWELL_BUILD_IS_CI=false
--- a/src/ci/bin/test.inc.sh
+++ b/src/ci/bin/test.inc.sh
@@ -112,7 +112,6 @@ cromwell::private::create_build_variables() {
             CROMWELL_BUILD_MYSQL_PASSWORD=""
             CROMWELL_BUILD_MYSQL_SCHEMA="cromwell_test"
             CROMWELL_BUILD_GENERATE_COVERAGE=false
-            CROMWELL_BUILD_SPAN_SCALE_FACTOR="20"
             ;;
         *)
             CROMWELL_BUILD_IS_CI=false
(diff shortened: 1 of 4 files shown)


src/ci/bin/test.inc.sh
```diff
@@ -112,7 +112,7 @@ cromwell::private::create_build_variables() {
             CROMWELL_BUILD_MYSQL_PASSWORD=""
             CROMWELL_BUILD_MYSQL_SCHEMA="cromwell_test"
             CROMWELL_BUILD_GENERATE_COVERAGE=false
-            CROMWELL_BUILD_SPAN_SCALE_FACTOR="15"
+            CROMWELL_BUILD_SPAN_SCALE_FACTOR="20"
```
mcovarr: 😦

kshakir:
I'm not 100% convinced that scaling is completely wired. Have you confirmed that this increase is also affecting your test timeouts? The linked error from #4223 says:

> Attempted 13 times over 5.126574189 seconds.

That's *way* too short for something that should have already been scaled 12x.

I'm in the process of [adding some println-debugging](https://github.com/broadinstitute/cromwell/commit/29d2f35662d6a81a4de383ad54df4ee0242611a4) on a similar issue. In my case I suspect one of the many, many timeouts wasn't scaled for an akka `.ask`, but will have to wait until the docker network issues are resolved to find out.

salonishah11:
@kshakir As per my understanding, `eventually` was the one that was timing out. When I tested it last week with several builds, the test did not fail in any of them. And I had observed that when I decreased the span scale factor, it started failing again. Maybe there are other timeouts failing the test as well. I will try and add println-debugging.

kshakir:
I'm still slowly investigating, but it looks like the scale factor environment variable isn't `export`ed, thus isn't available to `sbt`. Been playing around with the println's in a433a7f20a74faf70a1ec851545f0e9ec6836ce4 ([jenkins](https://fc-jenkins.dsp-techops.broadinstitute.org/job/cromwell-test-runner/555/consoleFull)) and 7fc56d3b73ce537b47aaecc6bf9cd0f1c020646f ([jenkins](https://fc-jenkins.dsp-techops.broadinstitute.org/job/cromwell-test-runner/557/consoleFull)).

PR #4262, opened 2018-10-16
Better message for bad path

aednichols:
It's a bit of a pet peeve of mine when error messages don't put quotes around raw values. This leads to  unnecessarily confusing scenarios like
```
Your input  did not meet our requirements.
```
for an input of `""`.

### DIFF
--- a/core/src/main/scala/cromwell/core/path/PathFactory.scala
+++ b/core/src/main/scala/cromwell/core/path/PathFactory.scala
@@ -58,7 +58,7 @@ object PathFactory {
     path.map(postMapping) getOrElse {
       val pathBuilderNames: String = pathBuilders map { _.name } mkString ", "
       throw PathParsingException(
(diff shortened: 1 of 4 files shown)


core/src/main/scala/cromwell/core/path/PathFactory.scala
```diff
@@ -58,7 +58,7 @@ object PathFactory {
     path.map(postMapping) getOrElse {
       val pathBuilderNames: String = pathBuilders map { _.name } mkString ", "
       throw PathParsingException(
-        s"Either $string exists on a filesystem not supported by this instance of Cromwell, or a failure occurred while building an actionable path from it." +
+        s"""Either the path "$string" exists on a filesystem not supported by this instance of Cromwell, or a failure occurred while building an actionable path from it.""" +
```
mcovarr: preexisting but what's an "actionable" path?

aednichols: I restrained myself from doing a comprehensive rewording, but now you're enabling me

PR #4270, opened 2018-10-17
Small changes to release WDL

rsasch:
I ran into issues in the last release related to the calls to `git clone` and the URLs used to create the names of the new `.jar` files, so
- changed github clones back to ssh
- changed jar paths

### DIFF
--- a/release/release_workflow.wdl
+++ b/release/release_workflow.wdl
@@ -18,7 +18,7 @@ task do_major_release {
      set -x 
 
      # Clone repo and checkout develop
-     git clone https://github.com/~{organization}/cromwell.git -b develop cromwell
+     git clone git@github.com:~{organization}/cromwell.git -b develop cromwell
      cd cromwell
 
      # Merge develop into master
@@ -73,7 +73,7 @@ task do_minor_release {
      set -x 
 
      # Clone repo and checkout hotfix branch
-     git clone https://github.com/~{organization}/cromwell.git -b ~{hotfixBranchName} cromwell
+     git clone git@github.com:~{organization}/cromwell.git -b ~{hotfixBranchName} cromwell
      cd cromwell
 
      # Make sure tests pass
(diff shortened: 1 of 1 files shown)


release/release_workflow.wdl
```diff
@@ -18,7 +18,7 @@ task do_major_release {
      set -x 
 
      # Clone repo and checkout develop
-     git clone https://github.com/~{organization}/cromwell.git -b develop cromwell
+     git clone git@github.com:~{organization}/cromwell.git -b develop cromwell
```
Horneth: FYI @cjllanwarne just had changed it after the last release [here](https://github.com/broadinstitute/cromwell/pull/4198) from ssh to https. I don't know what the right way to do it is but it would be nice to figure out a reliable way to make it work for everyone.

aednichols: We're not sure why we went from ssh -> https, but it failed because the script paused waiting for stdin for the user to type their github username & password

It worked fine when we changed back to ssh

release/release_workflow.wdl
```diff
@@ -176,8 +176,8 @@ task publishGithubRelease {
     String cromwellJarName = basename(cromwellJar)
     String womtoolJarName = basename(womtoolJar)
 
-    String cromwell_upload_url = "~{upload_url}?name=~{cromwellJarName}?label=~{cromwellJarName}"
-    String womtool_upload_url = "~{upload_url}?name=~{womtoolJarName}?label=~{womtoolJarName}"
+    String cromwell_upload_url = "~{upload_url}?name=~{cromwellJarName}&label=~{cromwellJarName}"
+    String womtool_upload_url = "~{upload_url}?name=~{womtoolJarName}&label=~{womtoolJarName}"
```
mcovarr: 🙈

release/release_workflow.wdl
aednichols: I've always preferred SSH as well. I don't want to re-type my credentials every time I do anything - and for 2FA users, you can't use your password at the prompt, you have to use a access token.

cjllanwarne: I changed this because the https version works out of the box for me whereas ssh fails... Sounds like you've had the opposite experience.

Perhaps I just set up the "auto authenticate with https" and everyone else has set up the "auto authenticate with ssh" 🤷‍♂️ 

FWIW the https version is what is copied to the clipboard from github when you want to clone a repo. But if I'm the minority here, we should go with ssh

kshakir: > FWIW the https version is what is copied to the clipboard from github when you want to clone a repo. 

FYI: 
screen shot 2018-10-18 at 2 58 52 pm

Switching between HTTPS and SSH is a toggle that gets saved in your account.

cjllanwarne: Ha, fair enough then!

PR #4271, opened 2018-10-17
Increase akka time factor in jenkins

Horneth:


CromIAM/src/test/scala/cromiam/webservice/QuerySupportSpec.scala
```diff
@@ -3,16 +3,19 @@ package cromiam.webservice
...
 import cromiam.webservice.QuerySupport.LabelContainsOrException
 import org.broadinstitute.dsde.workbench.model.WorkbenchUserId
 import org.scalatest.{FlatSpec, Matchers}
+import scala.concurrent.duration._
 
 class QuerySupportSpec extends FlatSpec with Matchers with ScalatestRouteTest with QuerySupport {
   override val cromwellClient = new MockCromwellClient()
   override val samClient = new MockSamClient()
   override val log: LoggingAdapter = NoLogging
+  implicit val timeout = RouteTestTimeout(10.seconds.dilated)
```
kshakir: Did you consider [activating the dilation](https://doc.akka.io/docs/akka/2.5/testing.html#accounting-for-slow-test-systems) of the existing `1.second` [akka-http testkit timeout](https://github.com/akka/akka-http/blob/v10.1.5/akka-http-testkit/src/main/scala/akka/http/scaladsl/testkit/RouteTestTimeout.scala#L14)? `testSbt.sh` could be modified to add `-Dakka.test.timefactor="${CROMWELL_SBT_TEST_SPAN_SCALE_FACTOR}"`. This would match up great with existing `10` times scaling for ScalaTest in #4261. Just be sure to set the `CROMWELL_SBT_TEST_SPAN_SCALE_FACTOR=1` for all the other `case` statements.

Horneth: Oh yep that does seem like a better way. Will change

kshakir: FYI #4261 is merged.

geoffjentry:
:+1:

PR #4276, opened 2018-10-18
report whole exception stack trace to BQ

danbills:
should fix nulls in BQ

### DIFF
--- a/centaur/src/it/scala/centaur/reporting/BigQueryReporter.scala
+++ b/centaur/src/it/scala/centaur/reporting/BigQueryReporter.scala
@@ -153,7 +153,7 @@ class BigQueryReporter(override val params: ErrorReporterParams) extends ErrorRe
       "test_attempt" -> Option(testEnvironment.attempt + 1),
       "test_message" -> Option(centaurTestException.message),
       "test_name" -> Option(testEnvironment.name),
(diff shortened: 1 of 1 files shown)


centaur/src/it/scala/centaur/reporting/BigQueryReporter.scala
```diff
@@ -153,7 +153,7 @@ class BigQueryReporter(override val params: ErrorReporterParams) extends ErrorRe
       "test_attempt" -> Option(testEnvironment.attempt + 1),
       "test_message" -> Option(centaurTestException.message),
       "test_name" -> Option(testEnvironment.name),
-      "test_stack_trace" -> centaurTestException.causeOption.map(ExceptionUtils.getStackTrace),
+      "test_stack_trace" -> ExceptionUtils.getStackTrace(centaurTestException),
```
mcovarr: Is this still going to print the cause when that is present?

kshakir: This should instead be:
```suggestion
      "test_stack_trace" -> Option(ExceptionUtils.getStackTrace(centaurTestException)),
```

Right now this is filtered out by the `collect` below`. In the corrected version, if the `getStackTrace` returns `null`, the stack trace won't be filled in.

PR #4280, opened 2018-10-19
Even more scaling of timespans.

kshakir:


kshakir:
Jenkins result here: https://fc-jenkins.dsp-techops.broadinstitute.org/job/cromwell-test-runner/619/

PR #4285, opened 2018-10-19
Womtool validates CWL (maybe)

aednichols:
This seems too simple to be correct, would appreciate a quick looking-over.

`1st-workflow.cwl` is an expected failure case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,18 @@
 # Cromwell Change Log
 
+## 37 Release Notes
+
+### Changing configuration options
+
+When the value `exit-code-timeout-seconds` is set, `check-alive` command is now only called once every timeout interval instead of each poll.
+
+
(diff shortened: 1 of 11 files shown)


cjllanwarne:
> This seems too simple to be correct

My usual response when I start thinking like this is "do I have any lingering doubts that can be documented?" And if so - "can I add any tests for to make myself feel more confident?"

aednichols:
- I have a failing test case, so that's a pretty good place to start looking!
- One assumption I need to validate is that every CWL file has exactly one `Callable`:
```
WomBundle(primaryCallable = Option(value), allCallables = Map.empty, typeAliases = Map.empty)
```
This is based on reading the CWL spec and the declaration
```
type Cwl = Workflow :+: CommandLineTool :+: ExpressionTool :+: CNil
```

PR #4295, opened 2018-10-23
Changed circe codecs from auto to semi-auto.

kshakir:
Also:
- Bumped circe dependency versions.
- Added the `perf` executable to the aggregated projects.


kshakir:
> Did you measure the speed gain ?

Non-scientifically, clean-compiles "felt" cut in half in both intellij + sbt.

Merging. Will be curious to hear from others if they notice a difference, plus-or-minus.

PR #4302, opened 2018-10-23
Increase max-content-length in centaur CI to fix CRON PAPIv2

Horneth:


src/ci/resources/centaur_application.conf
```diff
@@ -1,5 +1,7 @@
 include required(classpath("application.conf"))
 
+akka.http.client.parsing.max-content-length = 20030961
```
kshakir: what does this (magic?) number represent? is it approaching some max, some min, double-some-threshold, etc?

PR #4305, opened 2018-10-24
PAPI v2 Execution Event Grouping WIP

mcovarr:
~~This is simply a checkpoint of WIP and does not need to be reviewed unless it's going to be picked up by someone else in my absence. The grouping of execution events by correlation to labeled actions seems to basically be working. But from the current timing diagram it looks like there are a lot more actions that need labeling and/or correlation before this could possibly be called "Done".~~ All good now review it for realz!

### DIFF
--- a/engine/src/test/scala/cromwell/webservice/MetadataBuilderActorSpec.scala
+++ b/engine/src/test/scala/cromwell/webservice/MetadataBuilderActorSpec.scala
@@ -1,10 +1,10 @@
 package cromwell.webservice
 
-import java.time.OffsetDateTime
+import java.time.{OffsetDateTime, ZoneOffset}
 import java.util.UUID
 
-import akka.testkit._
 import akka.pattern.ask
+import akka.testkit._
 import akka.util.Timeout
 import cromwell.core._
 import cromwell.services.metadata.MetadataService._
@@ -18,6 +18,8 @@ import spray.json._
 
 import scala.concurrent.Future
 import scala.concurrent.duration._
+import cromwell.webservice.MetadataBuilderActorSpec._
+
 
 class MetadataBuilderActorSpec extends TestKitSuite("Metadata") with AsyncFlatSpecLike with Matchers with Mockito
   with TableDrivenPropertyChecks with ImplicitSender {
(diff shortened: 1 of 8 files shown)


engine/src/test/scala/cromwell/webservice/MetadataBuilderActorSpec.scala
```diff
@@ -535,4 +537,160 @@ class MetadataBuilderActorSpec extends TestKitSuite("Metadata") with AsyncFlatSp
...
+    }
+
+    val calls = List(Foo, Bar, Baz, Qux, Quux, Corge)
+    val actuals = calls map filterEventsByCall(actual)
+    val expecteds = calls map filterEventsByCall(expectations)
+
+    (actuals zip expecteds) foreach {
+      case (as, es) => (as.toList.map { _.toString } sorted) shouldBe (es.toList.map { _.toString } sorted)
+    }
+    1 shouldBe 1
```
Horneth: I'm pretty sure this one is true 😄

supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/PipelinesUtilityConversions.scala
```diff
@@ -22,7 +26,39 @@ trait PipelinesUtilityConversions {
...
+  def toExecutionEvent(actionIndexToEventType: Map[Int, String])(event: Event): Option[ExecutionEvent] = {
+    val groupingFromAction = for {
+      rawValue <- Option(event.getDetails.get("actionId"))
+      integerValue <- Try(Integer.valueOf(rawValue.toString)).toOption
+      group <- actionIndexToEventType.get(integerValue)
+    } yield group
+
+    // There are both "Started pulling" and "Stopped pulling" events but these are confusing for metadata, especially on the
+    // timing diagram. Create a single "Pulling <docker image>" grouping to absorb these events.
+    def groupingFromPull: Option[String] = List("Started", "Stopped") flatMap { k =>
```
Horneth: OOC any particular reason why this one is a `def` and the one above a `val` ?

Horneth: Also TOL: why is the grouping for this one done here instead of during the metadata building like for the other ones ?

engine/src/test/scala/cromwell/webservice/MetadataBuilderActorSpec.scala
mcovarr: 🙈

supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/PipelinesUtilityConversions.scala
mcovarr: The `val` vs `def` is because the first one always has to run but the second is the "else" of an `orElse`. The grouping logic here is different from the grouping that's currently in the metadata building, but I think this could still be moved over there to keep all the grouping logic in one place.

mcovarr: Oh I remember now...this is all PAPI v2 specific filtering and the MetadataBuilder is meant to be generic.

mcovarr: Welp those BCBIO changes will certainly necessitate some refactoring here. 😛 I'm inclined to have the "remove docker pull events during Delocalization" here in the PAPI v2 part of the world since the details of that really are specific to PAPI v2.

Horneth: Oh yeah sorry about the forced rebase -_- Yes I think keep that here is totally fine

PR #4307, opened 2018-10-24
Route Timing endpoint correctly

ruchim:


geoffjentry:
:+1:

cjllanwarne:
I suspect this still won't work because of the relative lookup to `/metadata` not getting the appropriate headers.

To do this right we might want to make the request to `metadata` *inside* the CromwellApi class and drop the string value returned as a static JSON object in the timing diagram `.html` page we return

geoffjentry:
@cjllanwarne IIRC this was solving “a”, not “the” problem :)

PR #4326, opened 2018-10-28
set aws default region in integration tests

RLuckom:
Noticed the integration tests were failing. 

After looking into it a little, I'm pretty sure that what happened is that when [this commit](https://github.com/broadinstitute/cromwell/commit/cc636457bf46d6fec001976ee830ad81eaed35b8) un-hardcoded the aws region, the integration tests started trying to submit batches to `batch.default.amazonaws.com` rather than `batch.us-east-1.amazonaws.com`. I think that whatever library is importing the credentials file for use with in AwsAuthMode.scala isn't defaulting to `us-east-1` as advertised. 

I believe this will fix the issue but you might prefer to solve the problem by correcting the default behavior instead; if so feel free to close this.

### DIFF
--- a/src/ci/resources/aws_application.conf
+++ b/src/ci/resources/aws_application.conf
@@ -12,8 +12,8 @@ aws {
     }
   ]
 
-  region = "default" // uses region from ~/.aws/config set by aws configure command,
-                     // or us-east-1 by default
+  region = "us-east-1"
+
 }
 
 engine {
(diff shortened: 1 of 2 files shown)


kshakir:
Merged in my travis-test-addition-PR into this branch.

I also manually launched a jenkins run to verify these changes and it already looks much happier: https://gotc-jenkins.dsp-techops.broadinstitute.org/job/cromwell-cron-aws/71/ 

@RLuckom Now that this PR has approvals, feel free to squash-and-merge (don't merge more than one commit please!), or give us a heads up and one of us can do it.

PR #4348, opened 2018-11-02
CromIAM check user authorization before forwarding request to Cromwell

### LINKED ISSUE #4284: CromIAM should check that the user has permission to add to a collection BEFORE submitting the workflow to Cromwell

salonishah11:
Closes #4284

### DIFF
--- a/CromIAM/src/main/scala/cromiam/webservice/QuerySupport.scala
+++ b/CromIAM/src/main/scala/cromiam/webservice/QuerySupport.scala
@@ -165,8 +165,6 @@ trait QuerySupport extends RequestSupport {
 }
 
 object QuerySupport {
-  def hasCollectionLabel(labels: Iterable[String]): Boolean = labels exists { _.startsWith(s"$CollectionLabelName:") }
-
   final case class InvalidQueryException(e: Throwable) extends
     Exception(s"Invalid JSON in query POST body: ${e.getMessage}", e)
 
--- a/CromIAM/src/test/scala/cromiam/webservice/QuerySupportSpec.scala
+++ b/CromIAM/src/test/scala/cromiam/webservice/QuerySupportSpec.scala
@@ -3,18 +3,13 @@ package cromiam.webservice
 import akka.event.{LoggingAdapter, NoLogging}
 import akka.http.scaladsl.model.headers.{Authorization, OAuth2BearerToken, RawHeader}
 import akka.http.scaladsl.model.{ContentTypes, HttpEntity, HttpHeader, StatusCodes}
-import akka.http.scaladsl.server.Route
-import akka.http.scaladsl.testkit.{RouteTestTimeout, ScalatestRouteTest}
+import akka.http.scaladsl.testkit.ScalatestRouteTest
 import cromiam.auth.{Collection, User}
 import cromiam.webservice.QuerySupport.LabelContainsOrException
 import org.broadinstitute.dsde.workbench.model.WorkbenchUserId
 import org.scalatest.{FlatSpec, Matchers}
 
-import scala.concurrent.duration._
-
 class QuerySupportSpec extends FlatSpec with Matchers with ScalatestRouteTest with QuerySupport {
-  implicit def default = RouteTestTimeout(5.seconds)
-
   override val cromwellClient = new MockCromwellClient()
   override val samClient = new MockSamClient()
   override val log: LoggingAdapter = NoLogging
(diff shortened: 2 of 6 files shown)


CromIAM/src/main/scala/cromiam/webservice/QuerySupport.scala
```diff
@@ -165,8 +165,6 @@ trait QuerySupport extends RequestSupport {
 }
 
 object QuerySupport {
-  def hasCollectionLabel(labels: Iterable[String]): Boolean = labels exists { _.startsWith(s"$CollectionLabelName:") }
```
salonishah11: Removed this function as it was not being used anymore

geoffjentry:
:+1:

CromIAM/src/test/scala/cromiam/webservice/SubmissionSupportSpec.scala
```diff
@@ -0,0 +1,41 @@
...
+  val helloWorldWdl = HelloWorld.workflowSource()
+  val helloWorldInputs = HelloWorld.rawInputs
+
+  val workflowSource = Multipart.FormData.BodyPart("workflowSource", HttpEntity(MediaTypes.`application/json`, helloWorldWdl))
+  val workflowInputs = Multipart.FormData.BodyPart("workflowInputs", HttpEntity(MediaTypes.`application/json`, helloWorldInputs.toJson.toString))
+  val formData = Multipart.FormData(workflowSource, workflowInputs).toEntity()
+
+  "Submit query" should "forward the request to Cromwell for authorized SAM user" in {
+    Post(submitPath).withHeaders(goodAuthHeaders).withEntity(formData) ~> submitRoute ~> check {
+      status shouldEqual StatusCodes.InternalServerError
```
Horneth: TOL: Looking at the mock code I was able to figure out that this is the expected response because the mock Cromwell client always returns 500, but it's not very obvious here IMO that a test for "user is authorized" should return 500.
I think it would be nice if possible to have the mock cromwell client respond differently based on the request (like you did for SAM). It could send back 200 for authorized and maybe throw an exception otherwise.
This would also have the added benefit to fail any test where Cromwell receives a request that wasn't authorized by CromIAM. For instance in the test below, it would make sure that on top of CromIAM responding with 400, it really didn't send the request to Cromwell.

CromIAM/src/test/scala/cromiam/webservice/QuerySupportSpec.scala
```diff
@@ -19,7 +19,7 @@ class QuerySupportSpec extends FlatSpec with Matchers with ScalatestRouteTest wi
 
   val queryPath = "/api/workflows/v1/query"
   val getQuery = s"$queryPath?status=Submitted&label=foo:bar&label=foo:baz"
-  val badGetQuery = s"$queryPath?status=Submitted?labelor=foo:bar&label=foo:baz"
+  val badGetQuery = s"$queryPath?status=Submitted&labelor=foo:bar&label=foo:baz"
```
mcovarr: ugh wrong kind of bad 😬

CromIAM/src/test/scala/cromiam/webservice/SubmissionSupportSpec.scala
```diff
@@ -0,0 +1,41 @@
...
+  val authorization = Authorization(OAuth2BearerToken("my-token"))
+  val badAuthHeaders: List[HttpHeader] = List(authorization, RawHeader("OIDC_CLAIM_user_id", samClient.unauthorizedUserCollectionStr))
+  val goodAuthHeaders: List[HttpHeader] = List(authorization, RawHeader("OIDC_CLAIM_user_id", samClient.authorizedUserCollectionStr))
+
+  val submitPath: String = "/api/workflows/v1"
+
+  val helloWorldWdl = HelloWorld.workflowSource()
+  val helloWorldInputs = HelloWorld.rawInputs
+
+  val workflowSource = Multipart.FormData.BodyPart("workflowSource", HttpEntity(MediaTypes.`application/json`, helloWorldWdl))
```
mcovarr: WDL is not JSON...

CromIAM/src/test/scala/cromiam/webservice/SubmissionSupportSpec.scala
mcovarr: I haven't dug into the mocks here but a 500 seems strange for a "success" expectation. 🙂

salonishah11: Good point. I will try that. Thanks!

PR #4358, opened 2018-11-05
BCBIO improvements part 1

Horneth:
- Allow a `0` value for CWL `outDirMin` and `tmpDirMin` resource attributes
- Adds an optional section to the language factory to define a command to run after the user's action that will return output files that can only be known at runtime
  - Only defined for CWL for now, which will remove unnecessary pull of jq for WDL tasks on PAPI2
  - Docker image and command can both be changed in the configuration
- The PAPI2 logic that handles delocalization of those file strips away some redundant pieces in the delocalized paths to reduce the overall length of the path

### DIFF
--- a/backend/src/main/scala/cromwell/backend/validation/InformationValidation.scala
+++ b/backend/src/main/scala/cromwell/backend/validation/InformationValidation.scala
@@ -25,16 +25,16 @@ import scala.util.{Failure, Success}
   * `withDefault` can be used to create a validation that defaults to a particular size.
   */
 object InformationValidation {
-  def instance(attributeName: String = RuntimeAttributesKeys.MemoryKey, defaultUnit: MemoryUnit): RuntimeAttributesValidation[MemorySize] =
-    new InformationValidation(attributeName, defaultUnit)
-  def optional(attributeName: String = RuntimeAttributesKeys.MemoryKey, defaultUnit: MemoryUnit): OptionalRuntimeAttributesValidation[MemorySize] =
-    instance(attributeName, defaultUnit).optional
-  def configDefaultString(attributeName: String = RuntimeAttributesKeys.MemoryKey, config: Option[Config], defaultUnit: MemoryUnit): Option[String] =
-    instance(attributeName, defaultUnit).configDefaultValue(config)
-  def withDefaultMemory(attributeName: String = RuntimeAttributesKeys.MemoryKey, memorySize: String, defaultUnit: MemoryUnit): RuntimeAttributesValidation[MemorySize] = {
+  def instance(attributeName: String = RuntimeAttributesKeys.MemoryKey, defaultUnit: MemoryUnit, allowZero: Boolean = false): RuntimeAttributesValidation[MemorySize] =
+    new InformationValidation(attributeName, defaultUnit, allowZero)
+  def optional(attributeName: String = RuntimeAttributesKeys.MemoryKey, defaultUnit: MemoryUnit, allowZero: Boolean = false): OptionalRuntimeAttributesValidation[MemorySize] =
+    instance(attributeName, defaultUnit, allowZero).optional
+  def configDefaultString(attributeName: String = RuntimeAttributesKeys.MemoryKey, config: Option[Config], defaultUnit: MemoryUnit, allowZero: Boolean = false): Option[String] =
+    instance(attributeName, defaultUnit, allowZero).configDefaultValue(config)
+  def withDefaultMemory(attributeName: String = RuntimeAttributesKeys.MemoryKey, memorySize: String, defaultUnit: MemoryUnit, allowZero: Boolean = false): RuntimeAttributesValidation[MemorySize] = {
     MemorySize.parse(memorySize) match {
-      case Success(memory) => instance(attributeName, defaultUnit).withDefault(WomInteger(memory.bytes.toInt))
-      case Failure(_) => instance(attributeName, defaultUnit).withDefault(BadDefaultAttribute(WomString(memorySize.toString)))
+      case Success(memory) => instance(attributeName, defaultUnit, allowZero).withDefault(WomInteger(memory.bytes.toInt))
+      case Failure(_) => instance(attributeName, defaultUnit, allowZero).withDefault(BadDefaultAttribute(WomString(memorySize.toString)))
     }
   }
 
@@ -44,12 +44,12 @@ object InformationValidation {
     "Expecting %s runtime attribute to be an Integer or String with format '8 GB'." +
       " Exception: %s"
 
-  private[validation] def validateString(attributeName: String, wdlString: WomString): ErrorOr[MemorySize] =
-    validateString(attributeName, wdlString.value)
+  private[validation] def validateString(attributeName: String, wdlString: WomString, allowZero: Boolean): ErrorOr[MemorySize] =
+    validateString(attributeName, wdlString.value, allowZero)
 
-  private[validation] def validateString(attributeName: String, value: String): ErrorOr[MemorySize] = {
+  private[validation] def validateString(attributeName: String, value: String, allowZero: Boolean): ErrorOr[MemorySize] = {
     MemorySize.parse(value) match {
-      case scala.util.Success(memorySize: MemorySize) if memorySize.amount > 0 =>
+      case scala.util.Success(memorySize: MemorySize) if memorySize.amount > 0 || (memorySize.amount == 0 && allowZero) =>
         memorySize.to(MemoryUnit.GB).validNel
       case scala.util.Success(memorySize: MemorySize) =>
         wrongAmountFormat.format(attributeName, memorySize.amount).invalidNel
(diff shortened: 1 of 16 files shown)


core/src/main/resources/reference.conf
```diff
@@ -391,6 +391,17 @@ languages {
         config {
           strict-validation: false
           enabled: true
+
+          # Optional section to define a command returning paths to output files that
+          # can only be known after the user's command has been run
+          output-runtime-extractor {
+            # Optional docker image on which the command should be run - Must have bash if provided
+            docker-image = "stedolan/jq@sha256:a61ed0bca213081b64be94c5e1b402ea58bc549f457c2682a86704dd55231e09"
```
kshakir: I'm sure The Emerald Empire will thank you for this one. 👍

supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/RuntimeOutputMapping.scala
```diff
@@ -0,0 +1,32 @@
...
+
+  /**
+    * List of prefixes to be stripped away from runtime output paths before
+    * appending them to the cloud call root to generate the delocalization path.
+    * The goal is to reduce unnecessary long paths which keep repeating the workflow root
+    * as the workflow progresses
+    * 
+    * For instance:
+    * 
+    * file:////cromwell_root/bucket/workflow_name/6d777414-5ee7-4c60-8b9e-a02ec44c398e/call-A/file.txt
```
kshakir: Is `file:////cromwell_root/` really a prefix? I'd guess it would only be three slashes, two after the scheme plus one before the root.

supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/api/Delocalization.scala
```diff
@@ -47,45 +50,72 @@ trait Delocalization {
...
-         * sh -c 'gsutil cp % $(echo % | sed -e "s/^\///")'
-         */
-        s""" | xargs -I % sh -c '${recoverRequesterPaysError(cloudCallRoot)(gsutilCommand)}'"""
-
-    ActionBuilder.cloudSdkShellAction(command)(mounts)
+    /*
+     * Delocalize all the files returned by the runtime output extractor
+     */
+    s"""
+       |#! /bin/bash
```
kshakir: ToL-ish: I'm not used to seeing a space between the [shebang](https://en.wikipedia.org/wiki/Shebang_(Unix)) and the executable.

supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/api/Delocalization.scala
```diff
@@ -47,45 +50,72 @@ trait Delocalization {
...
+       |set -x
+       |
+       |if [ -f $fofnPath ]; then
+       |  while IFS= read line
+       |  do
+       |    gcs_path=$$(echo $$line | $prefixFilters)
+       |    ${recoverRequesterPaysError(cloudCallRoot)(gsutilCommand)}
+       |  done  <$fofnPath
+       |fi
+     """.stripMargin
```
kshakir: I believe this IntelliJ copypasta leaves some extra whitespace at the beginning of the file, so the shebang won't be on the first line. Either start your string with `s"""|#!/bin/bash`, or change this to `.stripMargin.trim`.

supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/api/Delocalization.scala
```diff
@@ -47,45 +50,72 @@ trait Delocalization {
...
-        s""" | xargs -I % sh -c '${recoverRequesterPaysError(cloudCallRoot)(gsutilCommand)}'"""
-
-    ActionBuilder.cloudSdkShellAction(command)(mounts)
+    /*
+     * Delocalize all the files returned by the runtime output extractor
+     */
+    s"""
+       |#! /bin/bash
+       |
+       |set -x
```
kshakir: ToL: Have you considered `set -e`? There are also a few pipes, so I've started using `set -o pipefail`, and for unset variables I've been toying with `set -u`.

supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/RuntimeOutputMapping.scala
Horneth: Yeah good catch

supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/api/Delocalization.scala
Horneth: For the `set -e` I guess it would depend on if we want to try to delocalize everything we possibly can (which is what we do otherwise for normal output files) or fail at the first one. I'm actually not totally sure why we try to delocalize all files in general rather than stopping at the first failure. I think it might have come from the fact that in v1 things like the `rc` file were no different than an output file as far as PAPI was concerned and we definitely wanted the rc file delocalized regardless. But in v2 we have more control so we could stop delocalization at the first failure, and still delocalize the rc file...

Horneth: @ruchim @geoffjentry Do you have an opinion on this ? ^^

engine/src/main/scala/cromwell/engine/workflow/lifecycle/materialization/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -245,8 +246,8 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
         case (None, None) => "Either workflow source or url has to be supplied".invalidNelCheck
       }
     }
-
-    def buildValidatedNamespace(workflowSource: WorkflowSource, importResolvers: List[ImportResolver]): EitherT[IO, NonEmptyList[String], ValidatedWomNamespace] = {
+
```
mcovarr: 🔙

engine/src/main/scala/cromwell/engine/workflow/lifecycle/materialization/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -267,16 +268,20 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
         workflowLogger.info(s"Parsing workflow as ${validFactory.languageName} ${validFactory.languageVersionName}")
         pushLanguageToMetadata(validFactory.languageName, validFactory.languageVersionName)
       }
+      
+      factory
```
mcovarr: preexisting, but the "there is no valid factory" path seems even more interesting to log 🙂

backend/src/main/scala/cromwell/backend/validation/InformationValidation.scala
```diff
@@ -58,25 +58,25 @@ object InformationValidation {
...
 
-  private[validation] def validateInteger(attributeName: String, wdlInteger: WomInteger, defaultUnit: MemoryUnit): ErrorOr[MemorySize] =
-    validateInteger(attributeName, wdlInteger.value, defaultUnit)
+  private[validation] def validateInteger(attributeName: String, wdlInteger: WomInteger, defaultUnit: MemoryUnit, allowZero: Boolean): ErrorOr[MemorySize] =
+    validateInteger(attributeName, wdlInteger.value, defaultUnit, allowZero)
 
-  private[validation] def validateInteger(attributeName: String, value: Int, defaultUnit: MemoryUnit): ErrorOr[MemorySize] = {
-    if (value <= 0)
+  private[validation] def validateInteger(attributeName: String, value: Int, defaultUnit: MemoryUnit, allowZero: Boolean): ErrorOr[MemorySize] = {
+    if (value <= 0 && !allowZero)
```
mcovarr: less than zero ok with `allowZero`?

supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/PipelinesApiAsyncBackendJobExecutionActor.scala
```diff
@@ -73,7 +73,13 @@ class PipelinesApiAsyncBackendJobExecutionActor(standardParams: StandardAsyncExe
...
         GcsPathBuilder.validateGcsPath(path) match {
           case _: ValidFullGcsPath => path
-          case _ => (callRootPath / path.stripPrefix("file://").stripPrefix("/")).pathAsString
+          case _ => (callRootPath / 
+            RuntimeOutputMapping
+                .prefixFilters(workflowPaths.workflowRoot)
+                .foldLeft(path)({
+                  case (newPath, prefix) => newPath.stripPrefix(prefix)
+                })
+            ).pathAsString
```
mcovarr: What is this doing?

common/src/main/scala/common/util/StringUtil.scala
```diff
@@ -2,12 +2,47 @@ package common.util
...
 
+    /**
+      * Ensure string does not end with a /
+      */
     def ensureUnslashed: String = string.ensureSlashed |> StringUtils.chop
+
+    /**
+      * Ensure string does not start with a /
+      */
+    def ensureNoSlashPrefix: String = string.stripPrefix("/")
```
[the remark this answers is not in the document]
Horneth: Do you mean it should be `ensureUnSlashPrefix` ?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/materialization/MaterializeWorkflowDescriptorActor.scala
Horneth: It looks like there's a default factory that we always fall back to. Also `factory` here is an `ErrorOr` and the errors will be logged at some point later if it ends up as an `Invalid`

backend/src/main/scala/cromwell/backend/validation/InformationValidation.scala
Horneth: good catch :)

supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/PipelinesApiAsyncBackendJobExecutionActor.scala
Horneth: It's stripping the prefixes in `RuntimeOutputMapping.prefixFilters` from the path, one at a time.
For instance
`file:///cromwell_root/bucket/workflow_name/6d777414-5ee7-4c60-8b9e-a02ec44c398e/call-A/file.txt` will progressively become

`/cromwell_root/bucket/workflow_name/6d777414-5ee7-4c60-8b9e-a02ec44c398e/call-A/file.txt`
`bucket/workflow_name/6d777414-5ee7-4c60-8b9e-a02ec44c398e/call-A/file.txt`
`call-A/file.txt`

This code is called as part of a path mapper that will be applied to the WOMified `cwl.output.json`. The `cwl.output.json` when it's being read by Cromwell from the bucket still contains local paths (as they were created by the cwl tool). In order to keep things working we need to map those local paths to where they were actually delocalized, which is determined in [the other place](https://github.com/broadinstitute/cromwell/pull/4358/files#diff-9947ea7b949b621b47735a33d7009a7fR82) where `RuntimeOutputMapping.prefixFilters` is used.

mcovarr: commentify plz?

common/src/main/scala/common/util/StringUtil.scala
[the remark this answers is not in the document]
mcovarr: Maybe "ensureLeadingSlash"/"ensureNoLeadingSlash" and "ensureTrailingSlash"?

supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/PipelinesApiAsyncBackendJobExecutionActor.scala
Horneth: done

PR #4368, opened 2018-11-08
 Support for aborting On Hold workflows.

kshakir:
Separated service/biz objects (enums) and logic from data/sql objects.
Removed unused methods.


database/sql/src/main/scala/cromwell/database/slick/WorkflowStoreSlickDatabase.scala
```diff
@@ -87,40 +92,52 @@ trait WorkflowStoreSlickDatabase extends WorkflowStoreSqlDatabase {
...
-                                                  (implicit ec: ExecutionContext): DBIO[Unit] = {
+  private def updateForFetched(cromwellId: String,
+                               heartbeatTimestampTo: Timestamp,
+                               workflowStateFrom: String,
+                               workflowStateTo: String)
+                              (workflowStoreEntry: WorkflowStoreEntry)
+                              (implicit ec: ExecutionContext): DBIO[Unit] = {
     val workflowExecutionUuid = workflowStoreEntry.workflowExecutionUuid
     val updateState = workflowStoreEntry.workflowState match {
         // Submitted workflows become running when fetched
```
mcovarr: comment needs updating

database/sql/src/main/scala/cromwell/database/slick/WorkflowStoreSlickDatabase.scala
```diff
@@ -1,55 +1,49 @@
...
+    * @param workflowExecutionUuid  Id to update or delete.
+    * @param workflowStateToDelete  Delete rows with this state.
+    * @param workflowStateForUpdate Update other rows to this state.
+    * @return Number of rows deleted, plus Some(heartbeatTimestampIsEmpty) if the workflow exists in the store, where
+    *         heartbeatTimestampIsEmpty is true if the workflow's heartbeat timestamp is empty, and None if the workflow
+    *         does not exist.
     */
-  override def setToAborting(workflowId: String)
-                            (implicit ec: ExecutionContext): Future[Option[Boolean]] = {
+  override def deleteOrUpdateWorkflowToState(workflowExecutionUuid: String,
```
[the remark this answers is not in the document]
kshakir: Each `database/sql` method is a transactional boundary. Unless we significantly refactor the way we do transaction encapsulation, OR decide that this operation doesn't belong in the same transaction, then I recommend this stay as is for now. Very open to other ideas / review

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreEngineActor.scala
```diff
@@ -101,12 +101,20 @@ final case class WorkflowStoreEngineActor private(store: WorkflowStore,
...
+        store.aborting(id) map { workflowStoreAbortResponse =>
+          log.info(s"Abort requested for workflow $id.")
+          workflowStoreAbortResponse
+        } map {
+          case WorkflowStoreAbortResponse.AbortedOnHold =>
+            pushCurrentStateToMetadataService(id, WorkflowAborted)
+            sndr ! WorkflowAbortedResponse(id)
+          case WorkflowStoreAbortResponse.AbortingHeartbeatTimestampIsEmpty =>
             pushCurrentStateToMetadataService(id, WorkflowAborting)
-            log.info(s"Abort requested for workflow $id.")
```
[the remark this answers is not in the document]
kshakir: Happy to move this logging statement around, but this PR's current layout we'll log every time someone requests an abort, not just the first time we acknowledge it.

engine/src/main/scala/cromwell/engine/workflow/workflowstore/SqlWorkflowStore.scala
```diff
@@ -30,18 +52,33 @@ case class SqlWorkflowStore(sqlDatabase: WorkflowStoreSqlDatabase) extends Workf
...
+      WorkflowStoreState.Aborting.toString
+    ) map {
+      case (deleteOnHoldCount, _) if deleteOnHoldCount > 0 =>
+        WorkflowStoreAbortResponse.AbortedOnHold
+      case (_, None) =>
+        WorkflowStoreAbortResponse.NotFound
+      case (_, Some(true)) =>
+        WorkflowStoreAbortResponse.AbortingHeartbeatTimestampIsEmpty
+      case (_, Some(false)) =>
+        WorkflowStoreAbortResponse.AbortingHeartbeatTimestampNonEmpty
```
mcovarr: Perhaps an ADT rather than this tuple type coming back from `deleteOrUpdateWorkflowToState`?

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreSubmitActor.scala
```diff
@@ -66,11 +69,14 @@ final case class WorkflowStoreSubmitActor(store: WorkflowStore, serviceRegistryA
       } yield submissionResponses
 
       futureResponses onComplete {
-        case Success(futureResponses) =>
-          log.info("Workflows {} submitted.", futureResponses.toList.map(res => res.id).mkString(", "))
+        case Success(worklfowSubmissionResponses) =>
```
mcovarr: worklfows => workflows

database/sql/src/main/scala/cromwell/database/slick/WorkflowStoreSlickDatabase.scala
```diff
@@ -1,55 +1,49 @@
...
+                                            (implicit ec: ExecutionContext): Future[(Int, Option[Boolean])] = {
     val action =  for {
-      restarted <- dataAccess.heartbeatClearedForWorkflowId(workflowId).result.headOption
-      _ <- dataAccess.workflowStateForId(workflowId).update(WorkflowStoreState.Aborting)
-    } yield restarted
+      restarted <- dataAccess.heartbeatClearedForWorkflowId(workflowExecutionUuid).result.headOption
+      deleted <- dataAccess
+        .workflowStateForWorkflowExecutionUUidAndWorkflowState((workflowExecutionUuid, workflowStateToDelete))
+        .delete
+      _ <- dataAccess.workflowStateForWorkflowExecutionUUid(workflowExecutionUuid).update(workflowStateForUpdate)
```
Horneth: ToL:
Given that a (non quantum) workflow can only be in one state at a time I think we could avoid doing the update if the delete returns `1`, which would save us a little in DB requests.
Also I'm wondering (and really don't know the answer) if we expect most aborts to be coming for on hold workflows or `submitted/running`. If the latter we might want to reverse those 2 requests and try the update first.
This also makes me think, why do we not delete the entry when the status is `Submitted` too ?

kshakir: Reversing the rows is possible, but would require slightly more complex SQL.

Right now the flow looks like:

1. Delete the selected wfid with status 'On Hold' or 'Submitted'
2. (if deleted == 0) Update the selected wfid to status 'Aborting'

Reversing these and getting the same results would require a bit more complexity. Doable, but I'm not convinced it's worth the complexity, yet.

Horneth: That's fine by me, I'm 👍 with the current order if it's not worth the hassle

geoffjentry: On this point I think that @Horneth 's intuition that aborting normal workflows will happen **far** more frequently than on hold, so if there's a real performance hit that'd be not so great. 

It's at least worth considering from that perspective, but OTOH this could be more akin to my old "reorder the routes by frequency" performance enhancement which was never really going to matter

mcovarr: I wouldn't worry about any of this since the update is going to be removed for DB-driven abort. 🙂

engine/src/main/scala/cromwell/engine/workflow/workflowstore/SqlWorkflowStore.scala
kshakir: Thought about the ADT, but couldn't come up with a clean SQL-like abstraction for the `sql` layer. (I try to keep the `sql` layer as close to stored-procedures as possible, keeping all non-db-logic in the services layer.) Here, in the middle/services layer, the multiple out parameters are immediately converted into an enum.

PR #4371, opened 2018-11-08
Support InputResourceRequirement

Horneth:


src/ci/bin/testSbt.sh
```diff
@@ -13,6 +13,7 @@ CROMWELL_AKKA_TEST_TIME_FACTOR=1
 case "${CROMWELL_BUILD_PROVIDER}" in
     "${CROMWELL_BUILD_PROVIDER_TRAVIS}")
         CROMWELL_SBT_TEST_EXCLUDE_TAGS="AwsTest,CromwellIntegrationTest,GcsIntegrationTest"
+        CROMWELL_SBT_TEST_SPAN_SCALE_FACTOR=2
```
Horneth: This is unrelated, I just had the swagger spec timeout twice on travis so decided to add that

cwl/src/main/scala/cwl/requirement/RequirementToAttributeMap.scala
```diff
@@ -1,10 +1,10 @@
 package cwl.requirement
 
-import cwl._
+import cwl.{InputResourceRequirement, _}
```
mcovarr: 🔙

cwl/src/main/scala/cwl/model.scala
```diff
@@ -273,6 +275,16 @@ case class ResourceRequirement(
   def effectiveOutdirMax = outdirMax.orElse(outdirMin)
 }
 
+/**
+  * This promotes InputResourceRequirement to a first class citizen requirement, which it really isn't.
+  * Since it's the only one for now it's not a big deal but if more of these pop up we might want to treat custom requirements
+  * in a different way
+  */
```
mcovarr: Hmm so this is really a hint that Cromwell now supports?

geoffjentry: Exactly

Horneth: Yes, the reason behind this comment is that I saw this other comment: https://github.com/broadinstitute/cromwell/blob/913efc91927083788dec8f87e70c1ab22d800b00/cwl/src/main/scala/cwl/Tool.scala#L81
Which led me to think that hints/requirements that are not defined in the CWL Spec were meant to be handled differently than treating them as a "real" `Requirement`

src/ci/bin/testSbt.sh
kshakir: Can you merge the uses of the two `CROMWELL_SBT_TEST_SPAN_SCALE_FACTOR` and `CROMWELL_AKKA_TEST_TIME_FACTOR` into say a single `CROMWELL_SBT_TEST_SPAN_SCALE_FACTOR`? IMHO it's confusing when they have separate scaling factors.

cwl/src/main/scala/cwl/model.scala
mcovarr: Oh that may be an unclear comment. IIRC hints can be of any type, I think the comment is trying to say that if they're of a type other than `Requirement` they're going to be dropped on the floor. But `Requirement` hints should be handled just fine.

PR #4375, opened 2018-11-09
Handle invalid inputs parsing error

### LINKED ISSUE #4086: submitting empty json file returns Internal server error
Submitting an empty file as a `workflowInputs` causes cromwell to return an unhelpful 
`There was an internal server error.%`

It happens on both PAPIv1 and PAPIv2 (tested on cromwell 33 and 34) 

running the following:

` ``
curl -s -v -F workflowSource=empty -F workflowInputs=empty https://cromwell-v34.dsde-methods.broadinstitute.org/api/workflows/v1
```
results in:
```
There was an internal server error.%
```

The file `empty` is an empty file.

It would be nice if it returned a more useful error message to tell you what the problem was.

### PR DESCRIPTION
**Pull request by salonishah11**
This PR fixes [this](https://github.com/broadinstitute/cromwell/issues/4086) bug. It now handles the cases when empty input file or invalid json (but valid yaml) inputs is passed to Cromwell, and returns better error messages for them.

Closes #4086

### DIFF
--- a/engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
+++ b/engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
@@ -268,7 +268,7 @@ object PartialWorkflowSources {
       case Some(input: String) =>  {
         Try(input.parseJson).toErrorOrWithContext(s"parse input: '$input', which is not a valid json. Please check for syntactical errors.") flatMap {
           case JsObject(inputMap) => inputMap.validNel
-          case j: JsValue => s"Submitted input '$input' belongs to ${j.getClass}, which is not a valid json object.".invalidNel
+          case j: JsValue => s"Submitted input '$input' of type ${j.getClass.getSimpleName} is not a JSON object.".invalidNel
         }
       }
       case None => Map.empty[String, JsValue].validNel
--- a/engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
+++ b/engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
@@ -172,13 +172,13 @@ object PartialWorkflowSources {
       yaml.parser.parse(data) match {
         // If it's an array, treat each element as an individual input object, otherwise simply toString the whole thing
         case Right(json) => json.asArray.map(_.map(_.toString())).getOrElse(Vector(json.pretty(Printer.noSpaces))).validNel
-        case Left(error) => s"Input file is not a valid yaml or json. Error: ${ExceptionUtils.getMessage(error)}.".invalidNel
+        case Left(error) => s"Input file is not a valid yaml or json. Inputs data: '$data'. Error: ${ExceptionUtils.getMessage(error)}.".invalidNel
       }
     }
 
     parseInputsTry match {
       case Success(v) => v
-      case Failure(error) => s"Input file is not a valid yaml or json. Error: ${ExceptionUtils.getMessage(error)}.".invalidNel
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala [hunk 1]
```diff
@@ -249,15 +263,15 @@ object PartialWorkflowSources {
...
+  private def toMap(someInput: Option[String]): ErrorOr[Map[String, JsValue]] = {
     someInput match {
-      case Some(inputs: String) => inputs.parseJson match {
-        case JsObject(inputMap) => inputMap
-        case _ =>
-          throw new RuntimeException(s"Submitted inputs couldn't be processed, please check for syntactical errors")
+      case Some(input: String) =>  {
+        Try(input.parseJson).toErrorOrWithContext(s"parse input: '$input', which is not a valid json. Please check for syntactical errors.") flatMap {
+          case JsObject(inputMap) => inputMap.validNel
+          case j: JsValue => s"Submitted input '$input' belongs to ${j.getClass}, which is not a valid json object.".invalidNel
```
Horneth: ToL: Maybe something more along the lines of s"Submitted input '$input' of type ${j.getClass.getSimpleName} is not a JSON object"

engine/src/test/scala/cromwell/webservice/PartialWorkflowSourcesSpec.scala
```diff
@@ -1,17 +1,49 @@
...
+    val validInput = "{\"key\":\"value\"}"
+    val invalidAuxJsonInput = "invalidInput"
+    val mergedMapsErrorOr = PartialWorkflowSources.mergeMaps(Seq(Option(validInput), Option(invalidAuxJsonInput)))
+
+    mergedMapsErrorOr.isValid shouldBe false
 
-    allInputs.fields.keys should contain allOf("wf.a1", "wf.a2")
-    allInputs.fields("wf.a2") should be(JsString("universe"))
+    mergedMapsErrorOr match {
+      case Valid(_) => "This is unexpected! This test is designed to fail!"
```
Horneth: wrap this around a `fail`: `fail("This is unexpected! This test is designed to fail!")` to make the test actually fail. As it is now the test would still pass if it got here

engine/src/test/scala/cromwell/webservice/PartialWorkflowSourcesSpec.scala
```diff
@@ -1,17 +1,49 @@
...
+  }
+
+  it should "return error when workflow input is not a valid json object" in {
+    val invalidJsonInput = "\"invalidInput\""
+    val mergedMapsErrorOr = PartialWorkflowSources.mergeMaps(Seq(Option(invalidJsonInput)))
+
+    mergedMapsErrorOr.isValid shouldBe false
+
+    mergedMapsErrorOr match {
+      case Valid(_) => "This is unexpected! This test is designed to fail!"
```
Horneth: Add a `fail` here (see explanation in comment below)

engine/src/main/scala/cromwell/webservice/PartialWorkflowSources.scala
salonishah11: 👍

engine/src/test/scala/cromwell/webservice/PartialWorkflowSourcesSpec.scala
salonishah11: TIL. Thank you!

PR #4376, opened 2018-11-09
Match on IoFailAck and not specifically IoFailure.

mcovarr:
The code was previously pattern matching on `IoFailure` when it also needed to be able to handle `IoReadForbiddenFailure`. This adds an `unapply` to a trait implemented by both of those case classes to pattern match both types.

### DIFF
--- a/engine/src/main/scala/cromwell/engine/io/IoActor.scala
+++ b/engine/src/main/scala/cromwell/engine/io/IoActor.scala
@@ -146,7 +146,7 @@ trait IoCommandContext[T] extends StreamContext {
   def request: IoCommand[T]
   def replyTo: ActorRef
   def fail(failure: Throwable): IoResult = (request.fail(failure), this)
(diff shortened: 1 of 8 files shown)


engine/src/main/scala/cromwell/engine/io/IoActor.scala
```diff
@@ -146,7 +146,7 @@ trait IoCommandContext[T] extends StreamContext {
   def request: IoCommand[T]
   def replyTo: ActorRef
   def fail(failure: Throwable): IoResult = (request.fail(failure), this)
-  def failForbidden(failure: Throwable, forbiddenPath: String): IoResult = (request.failReadForbidden(failure, forbiddenPath), this)
+  def failReadForbidden(failure: Throwable, forbiddenPath: String): IoResult = (request.failReadForbidden(failure, forbiddenPath), this)
```
mcovarr: unrelated method name fix

PR #4377, opened 2018-11-10
add AES256 encryption to copyToS3()

wleepang:
Adds the AES256 encryption flag to the function that copies RC and STDOUT/ERR logs to S3


geoffjentry:
👍

Issue #3889, opened 2018-07-12
Feature request: Allow for setting the root (cromwell-executions) folder more easily


rhpvorderman:
For our CI testing we want to be able to set the location of the `cromwell-executions` folder. This can already be done by messing with the config file, but this is not ideal. The `root` setting is nested within `backends-> providers -> the-used-backend -> root`. This is very hard to set using the command-line with the `-D` option. 

Is there a possibility to set an option that is agnostic of background? `-Dbackends.default_root` for example? Or just on option to set it on the command-line `run --root my_preferred_root my_workflow.wdl`?

huangzhibo:
When use the server mode, set "backend.providers.Local.config.root" in options.json is unusefule. 
It's a question.

PR #4379, opened 2018-11-10
Allow for setting the root folder in workflowOptions

huangzhibo:
Allow for setting the root (cromwell-executions) folder in workflowOptions(option.json)
#3889

### DIFF
--- a/backend/src/main/scala/cromwell/backend/io/WorkflowPaths.scala
+++ b/backend/src/main/scala/cromwell/backend/io/WorkflowPaths.scala
@@ -19,8 +19,10 @@ trait WorkflowPaths extends PathFactory {
 
   /**
     * Path (as a String) of the root directory Cromwell should use for ALL workflows.
+    * Allow for setting the root (cromwell-executions) folder in workflowOptions(option.json)
     */
-  protected lazy val executionRootString: String = config.as[Option[String]]("root").getOrElse(WorkflowPaths.DefaultExecutionRootString)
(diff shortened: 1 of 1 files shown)


cjllanwarne:
Hi @huangzhibo, thanks for your contribution! I have two comments:

---

1. I think this feature is OK if the cromwell owner wants to allow it, but not in all cases. 
    - Eg if I'm hosting a Cromwell instance I might not want users to be able to specify an arbitrary location on my filesystem to write their workflow contents. 
    - I want to allow others to comment on this rather than making any solid statements... personally I think at the very least we'd want to require an opt-in to the configuration to enable this - something like `"allow_workflow_specified_execution_root"`?

---

2. Could you consider adding a centaur test for this? This should help get you started:
    - Have a look in `centaur/src/main/resources`
    - Make a new test directory for the new test
    - Add a workflow that can determine the workflow root - off the top of my head, maybe running `pwd` from within the `command` and selecting the workflow root from the output?
    - Add a new workflow options file alongside the workflow file and use it to specify an execution root.
    - Create a new `.test` file next to the others, and use it to specify your workflow and your workflow options as inputs

mcovarr:
I agree with what Chris said above about making this "opt-in" and adding a Centaur test. It also looks like those Travis `sbt` failures are real so those tests would need to be fixed as well.

gemmalam:
Closing PR unless you would like to reopen to add work to it.

PR #4380, opened 2018-11-11
docs noAddress attribute

ruchim:
This has come up a few times as a question on the gatk forums quite a few times.

### DIFF
--- a/docs/RuntimeAttributes.md
+++ b/docs/RuntimeAttributes.md
@@ -325,7 +325,7 @@ runtime {
 
 The task below will fail for two reasons:
  1. The command is accessing an external service, in this case GitHub.
- 2. The docker image is available in DockerHub and not the Google Container Registry. b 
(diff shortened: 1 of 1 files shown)


geoffjentry:
👍

docs/RuntimeAttributes.md
```diff
@@ -305,6 +305,41 @@ runtime {
...
+
+runtime {
+  docker: "gcr.io/gcp-runtimes/ubuntu_16_0_4:latest"
+  noAddress: true
+}
+```
+
+The task below will fail for two reasons:
+ 1. The command is accessing an external service, in this case GitHub.
+ 2. The docker image is available in DockerHub and not the Google Container Registry. b
```
Horneth: extra `b` at the end

cjllanwarne: To `b` or not to `b`, that is the question here.

PR #4384, opened 2018-11-13
Regression test for comments before 'version'

cjllanwarne:
Allows closing of https://github.com/openwdl/wdl/pull/251 by proving that it's possible to have comments before the `version` statement in WDL files.


cjllanwarne:
@mcovarr right - anything in here gets picked up by the `WomtoolValidateSpec` and run through `womtool validate`

PR #4385, opened 2018-11-13
[Test Reliability] More time for SubmissionSupportSpec

cjllanwarne:
I noticed that I got a few failures using the default 1 second timeout here. 

So this PR ups that to 10 seconds... 🤞

### DIFF
--- a/CromIAM/src/test/scala/cromiam/webservice/SubmissionSupportSpec.scala
+++ b/CromIAM/src/test/scala/cromiam/webservice/SubmissionSupportSpec.scala
@@ -4,13 +4,16 @@ import akka.event.{LoggingAdapter, NoLogging}
 import akka.http.scaladsl.model._
 import akka.http.scaladsl.model.headers.{Authorization, OAuth2BearerToken, RawHeader}
 import akka.http.scaladsl.server.AuthorizationFailedRejection
-import akka.http.scaladsl.testkit.ScalatestRouteTest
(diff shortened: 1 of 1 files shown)


CromIAM/src/test/scala/cromiam/webservice/SubmissionSupportSpec.scala
```diff
@@ -4,14 +4,17 @@ import akka.event.{LoggingAdapter, NoLogging}
...
+import akka.http.scaladsl.testkit.{RouteTestTimeout, ScalatestRouteTest}
 import org.scalatest.{FlatSpec, Matchers}
+import scala.concurrent.duration._
 
 class SubmissionSupportSpec extends FlatSpec with Matchers with ScalatestRouteTest with SubmissionSupport {
   override val cromwellClient = new MockCromwellClient()
   override val samClient = new MockSamClient()
   override val log: LoggingAdapter = NoLogging
 
+  implicit val routeTestTimeout = new RouteTestTimeout(10.seconds)
```
mcovarr: This isn't amenable to the scaling magic we've been using successfully elsewhere?

kshakir: If this is problematically slow, this should be `10.seconds.dilated` available via the magic of `import akka.testkit._`.

Issue #4117, opened 2018-09-18
Imports missing from workflow metadata


samanehsan:
While scale testing a pipeline last week, we saw this failure message in Cromwell-as-a-Service on 3/999 workflows:

`Failed to import workflow SmartSeq2SingleSample.wdl.:\nBad import SmartSeq2SingleSample.wdl: Failed to resolve 'SmartSeq2SingleSample.wdl' using resolver: 'http importer (no 'relative-to' origin)' (reason 1 of 1): Relative path"`

All 999 workflows are the same and were started programmatically, but the three that have this import error are missing "imports" in the Cromwell metadata. We are providing a zipped directory containing all of the workflow dependencies instead of using http imports. It's unclear whether there was an issue submitting these workflows to Cromwell, or if the zip file was received by Cromwell but was not passed on to the workflow. 

Here are the workflow IDs for the failed workflows:
- 8eb0f933-5e9c-477c-9de8-6ef03049aafa
- 96ca1d93-78c6-4eae-ba03-b540728353ca
- c548deda-e518-4a15-8191-97e73b68cd4b

cjllanwarne:
A few more questions occurred to me - 
- was the imports zip the same in each workflow?
- were the workflows all the same?
- do you have any logs from the sender to check that a zip was indeed sent?
- were they submitted as a series of 999 separate submits or as a single batch submit POST?

samanehsan:
> Was the imports zip the same in each workflow?
Yes, we were running the same workflow so the imports zip is the same. We get those import files by downloading them from github and adding them to a cache so that we don't have to download them for each workflow.

> Were the workflows all the same?
Yes, all of the workflows were same 

> Do you have any logs from the sender to check that a zip was indeed sent?
No, we just have logs that a workflow was submitted to Cromwell 😞

> Were they submitted as a series of 999 separate submits or as a single batch submit POST? They were submitted as a series of 999 individual requests in the "On Hold" status, and a separate process started those on-hold workflows every 10 seconds.

dshiga:
Another example from this morning:
d5ea4ae4-1644-4e71-98fb-368465c67a2c

kbergin:
Of note, in the 'run this morning' we observed this 9 out of 2539 workflows submitted in the same way that Saman highlighted above. @cjllanwarne

dshiga:
[...]

PR #4389, opened 2018-11-14
Extra logging around imports zip

aednichols:
In support of #4117.

I picked two strategic places to dump some info about a workflow's imports zip (if present):
- Right as the workflow is submitted
  - Did we receive the zip and correctly write it to the workflow store table?
- Immediately before we attempt to instantiate the workflow
  - Did we unzip the zip and create a viable `DirectoryResolver` from it?
  - Based on this log output, we can SSH into the server and look at the actual directory created, in the pattern `/tmp/99104084175675660.zip2922990665093481334`

No tests specifically for this, but you can see the new log statements in the CI output. (That's where you'll find this pudding's proof.)

### DIFF
--- a/engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
@@ -294,6 +294,13 @@ class WorkflowManagerActor(params: WorkflowManagerActorParams)
       logger.info(s"$tag Starting workflow UUID($workflowId)")
     }
 
+    // Debugging #4117 - if there are imports present at this point, they were received by Cromwell and written
+    // into the workflow store. (AEN 2018-11-14)
+    workflow.sources.importsZipFileOption match {
+      case Some(zip) => logger.info(s"$tag Found imports zip on workflow $workflowId, size ${zip.length} bytes.")
+      case None => logger.info(s"$tag No imports zip on workflow $workflowId.")
+    }
+
     val fileHashCacheActor: Option[ActorRef] =
       if (fileHashCacheEnabled) Option(context.system.actorOf(RootWorkflowFileHashCacheActor.props(params.ioActor))) else None
 
--- a/engine/src/main/scala/cromwell/engine/workflow/lifecycle/materialization/MaterializeWorkflowDescriptorActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/lifecycle/materialization/MaterializeWorkflowDescriptorActor.scala
@@ -288,7 +288,7 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
       if (importLocalFilesystem) DirectoryResolver.localFilesystemResolvers(None)
       else List.empty
 
-    val zippedResolverCheck: Parse[Option[ImportResolver]] = fromEither[IO](sourceFiles.importsZipFileOption match {
+    val zippedResolverCheck: Parse[Option[DirectoryResolver]] = fromEither[IO](sourceFiles.importsZipFileOption match {
       case None => None.validNelCheck
       case Some(zipContent) => zippedImportResolver(zipContent, workflowId).toEither.map(Option.apply)
     })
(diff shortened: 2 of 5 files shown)


engine/src/main/scala/cromwell/engine/workflow/lifecycle/materialization/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -306,7 +306,16 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
       validatedNamespace <- buildValidatedNamespace(factory, sourceAndResolvers._1, sourceAndResolvers._2)
       _ = pushNamespaceMetadata(validatedNamespace)
       ewd <- fromEither[IO](buildWorkflowDescriptor(id, validatedNamespace, workflowOptions, labels, conf, pathBuilders, outputRuntimeExtractor).toEither)
-    } yield ewd
+    } yield {
+
+      // Debugging #4117 - make sure we are creating an importable directory out of the zip (AEN 2018-11-14)
+      zippedImportResolver match {
```
cjllanwarne: I would move this up right below `zippedImportResolver <- zippedResolverCheck`.

Otherwise if something subsequently fails building `ewd` this `yield` block won't run and this log will be lost. And I believe that that's precisely the case when we want to get this logging?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/materialization/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -306,7 +306,16 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
...
+        case Some(zir) => workflowLogger.info(s"Using zip import resolver '${zir.name}'")
```
cjllanwarne: Is it possible to log either here (or at unzip time) how many items were actually unzipped? One hypothesis might be "the unzip somehow incorrectly contained zero (or wrong) files"?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/materialization/MaterializeWorkflowDescriptorActor.scala
aednichols: That is... an excellent point

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -294,6 +294,13 @@ class WorkflowManagerActor(params: WorkflowManagerActorParams)
       logger.info(s"$tag Starting workflow UUID($workflowId)")
     }
 
+    // Debugging #4117 - if there are imports present at this point, they were received by Cromwell and written
+    // into the workflow store. (AEN 2018-11-14)
+    workflow.sources.importsZipFileOption match {
+      case Some(zip) => logger.info(s"$tag Found imports zip on workflow $workflowId, size ${zip.length} bytes.")
+      case None => logger.info(s"$tag No imports zip on workflow $workflowId.")
+    }
```
mcovarr: assuming this will be removed before merge?

engine/src/main/scala/cromwell/engine/workflow/lifecycle/materialization/MaterializeWorkflowDescriptorActor.scala
mcovarr: not sure we want to log that kind of detail in all cases...

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
cjllanwarne: I believe the new logging is the whole point of this PR

mcovarr: Discussed with faces, thumb remains up. 🙂

engine/src/main/scala/cromwell/engine/workflow/lifecycle/materialization/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -288,16 +288,23 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
...
 
     val labels = convertJsonToLabels(sourceFiles.labelsJson)
 
     for {
       _ <- publishLabelsToMetadata(id, labels)
       zippedImportResolver <- zippedResolverCheck
+      // Debugging #4117 - make sure we are creating an importable directory out of the zip (AEN 2018-11-14)
+      _ = zippedImportResolver match {
+        case Some(dr: DirectoryResolver) =>
+          workflowLogger.info(s"Importing from unzip directory '${dr.name}' with files [${dr.directory.list.map(_.toFile.getName).mkString(", ")}]")
```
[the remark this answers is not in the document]
aednichols: Heard! The format of this PR (into develop) absolutely does not make clear my intentions for this code.

I'm planning to give Mint a special build of 35 with this code hotfixed onto it, and then open up a second PR into develop with just the deletions.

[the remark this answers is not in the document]
aednichols: I particularly don't want to go through the whole hotfix ordeal again on the off-chance we didn't have enough logging the first time, so I'm being conservative... by being liberal with logging

PR #4390, opened 2018-11-15
Synchronize access to Liquibase APIs.

kshakir:


database/migration/src/main/scala/cromwell/database/migration/liquibase/LiquibaseUtils.scala
```diff
@@ -11,6 +11,7 @@ import liquibase.{Contexts, LabelExpression, Liquibase}
 import org.hsqldb.persist.HsqlDatabaseProperties
 
 object LiquibaseUtils {
+  private val mutex = new Object
```
Horneth: OOC what is the difference between this and synchronizing on `LiquibaseUtils` itself ?

database/migration/src/main/scala/cromwell/database/migration/liquibase/LiquibaseUtils.scala
```diff
@@ -21,16 +22,18 @@ object LiquibaseUtils {
     * @param jdbcConnection A jdbc connection to the database.
     */
   def updateSchema(settings: LiquibaseSettings)(jdbcConnection: Connection): Unit = {
-    val liquibaseConnection = newConnection(jdbcConnection)
-    try {
-      val database = DatabaseFactory.getInstance.findCorrectDatabaseImplementation(liquibaseConnection)
-      database.setDatabaseChangeLogLockTableName(settings.databaseChangeLogLockTableName.toUpperCase)
-      database.setDatabaseChangeLogTableName(settings.databaseChangeLogTableName.toUpperCase)
+    mutex.synchronized {
```
aednichols: Do you really need `mutex`?

It seems like `this.synchronized { ... }` would accomplish the same thing (just as there is only ever one `mutex`, there is only ever one `LiquibaseUtils`)

kshakir:
It will take me a while to dig up an online reference (googling java thread safety returns a ton of results to sort through). But creating one's own private lock is an extra level of paranoia, kind of like marking all java variables as `final`, or reducing the scope of classes to `private`.

If one uses `this` as a mutex, then others can actually steal your lock, by locking **you**.

```scala
object LiquibaseUtils {
  def echoQuick = {
    this.synchronized {
      println("hello");
    }
  }
}

object ForeignImplementation {
  LiquibaseUtils.synchronized {
    // I have your lock!
    Thread.sleep(1.day.toMillis)
  }
}
```

If however the synchronization is done on a private variable, it can never be shared by outside participants.

```scala
object LiquibaseUtils {
  private val cantTouchThis = new Object
  def echoQuick = {
    cantTouchThis.synchronized {
      println("hello");
    }
  }
}

object ForeignImplementation {
  LiquibaseUtils.synchronized {
    // Doesn't affect echoQuick
    Thread.sleep(1.day.toMillis)
  }
}
```

database/migration/src/main/scala/cromwell/database/migration/liquibase/LiquibaseUtils.scala
kshakir: See comment below.

database/migration/src/main/scala/cromwell/database/migration/liquibase/LiquibaseUtils.scala
kshakir: See comment below.

kshakir:
Found a good discussion. Will put it in a comment.

https://stackoverflow.com/questions/442564/avoid-synchronizedthis-in-java

aednichols:
Cool, TIL

PR #4401, opened 2018-11-17
Database-driven abort.

mcovarr:
Will add docs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## 37 Release Notes
 
+### Instance independent abort
+
+For multi-instance Cromwell deployments sharing a single database, earlier versions of Cromwell required abort
+requests to be sent specifically to the instance that was running the targeted workflow. Cromwell 37 now
+allows abort commands to be sent to any Cromwell instance in the shared-database deployment. Configuration details
+[here](https://cromwell.readthedocs.io/en/develop/Configuring/abort-configuration).
+
 ### Call cache blacklisting
 
 The Google Pipelines API (PAPI) version 1 and 2 backends now offer the option of call cache blacklisting on a per-bucket basis.
--- a/cromwell.examples.conf
+++ b/cromwell.examples.conf
@@ -74,6 +74,26 @@ system {
 
     #object = 128000
   }
+
+  abort {
+    # These are the default values in Cromwell, in most circumstances there should not be a need to change them.
+
+    # How frequently Cromwell should scan for aborts.
+    scan-frequency: 30 seconds
+
+    # The cache of in-progress aborts. Cromwell will add entries to this cache once a WorkflowActor has been messaged to abort.
+    # If on the next scan an 'Aborting' status is found for a workflow that has an entry in this cache, Cromwell will not ask
+    # the associated WorkflowActor to abort again.
+    cache {
+      enabled: true
+      # Guava cache concurrency.
+      concurrency: 1
+      # How long entries in the cache should live from the time they are added to the cache.
+      ttl: 20 minutes
+      # Maximum number of entries in the cache.
+      size: 100000
+    }
+  }
 }
 
 workflow-options {
--- a/database/sql/src/main/scala/cromwell/database/slick/WorkflowStoreSlickDatabase.scala
+++ b/database/sql/src/main/scala/cromwell/database/slick/WorkflowStoreSlickDatabase.scala
@@ -6,6 +6,7 @@ import cats.instances.future._
 import cats.syntax.functor._
 import cromwell.database.sql.WorkflowStoreSqlDatabase
 import cromwell.database.sql.tables.WorkflowStoreEntry
+import mouse.all._
 
 import scala.concurrent.{ExecutionContext, Future}
 
@@ -27,25 +28,24 @@ trait WorkflowStoreSlickDatabase extends WorkflowStoreSqlDatabase {
                                              workflowStateToDelete1: String,
                                              workflowStateToDelete2: String,
                                              workflowStateForUpdate: String)
-                                            (implicit ec: ExecutionContext): Future[(Int, Option[Boolean])] = {
-    val action =  for {
-      restarted <- dataAccess.heartbeatClearedForWorkflowId(workflowExecutionUuid).result.headOption
-      // First, delete all rows in either of our states to be deleted
+                                            (implicit ec: ExecutionContext): Future[Option[Boolean]] = {
+    val action = for {
+      // First, delete all rows in either of our states to be deleted.
       deleted <- dataAccess
         .workflowStoreEntryForWorkflowExecutionUUidAndWorkflowStates(
           (workflowExecutionUuid, workflowStateToDelete1, workflowStateToDelete2)
         )
         .delete
-      // Second, for any single row still present, update its state
-      _ <-
+      // Second, for any single row still present update its state.
+      updated <-
         deleted match {
           case 0 =>
             dataAccess.workflowStateForWorkflowExecutionUUid(workflowExecutionUuid).update(workflowStateForUpdate)
           case _ => assertUpdateCount("deleteOrUpdateWorkflowToState", deleted, 1)
         }
-    } yield (deleted, restarted)
+    } yield (deleted, updated)
     
-    runTransaction(action)
+    runTransaction(action) map { case (deleted, updated) => if (deleted == 0 && updated == 0) None else Option(deleted > 0) }
   }
 
   override def addWorkflowStoreEntries(workflowStoreEntries: Iterable[WorkflowStoreEntry])
@@ -138,5 +138,9 @@ trait WorkflowStoreSlickDatabase extends WorkflowStoreSqlDatabase {
     } yield updated
 
     runTransaction(action)
+  }
+
+  override def findWorkflowsWithAbortRequested(cromwellId: String)(implicit ec: ExecutionContext): Future[Iterable[String]] = {
+    dataAccess.findWorkflowsWithAbortRequested(cromwellId).result |> runTransaction
   }
 }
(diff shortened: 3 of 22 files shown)


engine/src/main/scala/cromwell/engine/workflow/workflowstore/AbortRequestScanningActor.scala
```diff
@@ -0,0 +1,68 @@
...
+
+import scala.collection.JavaConverters._
+import scala.concurrent.duration._
+import scala.language.postfixOps
+
+class AbortRequestScanningActor(workflowStoreActor: ActorRef,
+                                workflowManagerActor: ActorRef,
+                                abortConfig: AbortConfig,
+                                workflowHeartbeatConfig: WorkflowHeartbeatConfig) extends Actor with Timers with ActorLogging {
+  private val cache = CacheBuilder.newBuilder()
```
mcovarr: oops wire in config

database/sql/src/main/scala/cromwell/database/slick/WorkflowStoreSlickDatabase.scala
```diff
@@ -27,25 +28,24 @@ trait WorkflowStoreSlickDatabase extends WorkflowStoreSqlDatabase {
                                              workflowStateToDelete1: String,
                                              workflowStateToDelete2: String,
                                              workflowStateForUpdate: String)
-                                            (implicit ec: ExecutionContext): Future[(Int, Option[Boolean])] = {
-    val action =  for {
-      restarted <- dataAccess.heartbeatClearedForWorkflowId(workflowExecutionUuid).result.headOption
```
kshakir: This seems to be the only time this `heartbeatClearedForWorkflowId` was used. If so, also kill off the declaration in the Component.

database/sql/src/main/scala/cromwell/database/sql/WorkflowStoreSqlDatabase.scala
```diff
@@ -48,7 +48,7 @@ ____    __    ____  ______   .______       __  ___  _______  __        ______
                                     workflowStateToDelete1: String,
                                     workflowStateToDelete2: String,
                                     workflowStateForUpdate: String)
-                                   (implicit ec: ExecutionContext): Future[(Int, Option[Boolean])]
+                                   (implicit ec: ExecutionContext): Future[Option[Boolean]]
```
kshakir: The scaladoc `@return` above needs updating too.

engine/src/main/scala/cromwell/engine/workflow/workflowstore/AbortRequestScanningActor.scala
```diff
@@ -0,0 +1,68 @@
...
+  override def receive: Receive = {
+    case RunScan =>
+      workflowStoreActor ! FindWorkflowsWithAbortRequested(cromwellId = workflowHeartbeatConfig.cromwellId)
+    case FindWorkflowsWithAbortRequestedSuccess(ids) =>
+      val oldIds = cache.getAllPresent(ids.asJava).keySet().asScala
+      // Diff this list to the cache.
+      val newIds = ids.toSet -- oldIds
+      if (newIds.nonEmpty) {
+        // If there are new IDs then query the WorkflowManagerActor for the corresponding ActorRefs.
+        workflowManagerActor ! WorkflowManagerActor.QueryRootWorkflowActorsRequest(newIds)
```
Horneth: ToL: Just curious about the "ask WMA for ActorRef corresponding to WF Id and then send Abort directly to Workflow Actor" compared to "send an `AbortCommand` with this WF Id to the WMA and let it do what it was doing before"

mcovarr: hmm let me try that.

cjllanwarne:
Is this the default behavior? Might be worth updating the swagger description from `Cromwell will make a best effort attempt to immediately halt any currently running jobs from this workflow` to make the claims less... immediate

engine/src/main/scala/cromwell/engine/workflow/workflowstore/InMemoryWorkflowStore.scala
```diff
@@ -76,6 +76,8 @@ class InMemoryWorkflowStore extends WorkflowStore {
     Future.successful(workflowIds.size)
 
   override def switchOnHoldToSubmitted(id: WorkflowId)(implicit ec: ExecutionContext): Future[Unit] = Future.successful(())
+
+  override def findWorkflowsWithAbortRequested(cromwellId: String)(implicit ec: ExecutionContext): Future[Iterable[WorkflowId]] = Future.successful(List.empty)
```
cjllanwarne: I assume this is ok because this `InMemory` version is only used in tests... but in that case could this class be put in `src/test` rather than `src/main`?

CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 ## 37 Release Notes
 
+### Instance independent abort
+
+For multi-instance Cromwell deployments sharing a single database, earlier versions of Cromwell required abort
+requests to be sent specifically to the instance that was running the targeted workflow. Cromwell 37 now
```
cjllanwarne: IMO I'd cut the context and just say:

> Aborts can be sent to any Cromwell instance within the shared-database deployment (see `Configuration Details`...)

cromwell.examples.conf
```diff
@@ -74,6 +74,26 @@ system {
...
   }
+
+  abort {
+    # These are the default values in Cromwell, in most circumstances there should not be a need to change them.
+
+    # How frequently Cromwell should scan for aborts.
+    scan-frequency: 30 seconds
+
+    # The cache of in-progress aborts. Cromwell will add entries to this cache once a WorkflowActor has been messaged to abort.
+    # If on the next scan an 'Aborting' status is found for a workflow that has an entry in this cache, Cromwell will not ask
```
cjllanwarne: This feature kind of confuses me...

This description makes it sound like you really want to remember everything that's in the set, rather than cache the top n results.

What if an abort takes more than 20 minutes? Or what if we have 100,001 aborts going concurrenty? Is the `WorkflowActor` abort logic good enough to handle multiple abort messages? Is this just here because we want to avoid multiple messages as much as possible?

database/sql/src/main/scala/cromwell/database/sql/WorkflowStoreSqlDatabase.scala
```diff
@@ -40,15 +40,13 @@ ____    __    ____  ______   .______       __  ___  _______  __        ______
     * @param workflowStateToDelete2 Also delete rows with this state.
     * @param workflowStateForUpdate Update other rows to this state.
     * @param ec                     The execution context where to run asynchronous operations.
-    * @return Number of rows deleted, plus Some(heartbeatTimestampIsEmpty) if the workflow exists in the store, where
-    *         heartbeatTimestampIsEmpty is true if the workflow's heartbeat timestamp is empty, and None if the workflow
-    *         does not exist.
+    * @return None if no rows were matched, Some(true) if a row was deleted or Some(false) if a row was updated.
```
cjllanwarne: It sort of feels to me like a set of three well-named `case object`s would be better than this complicated encoding of Options and Booleans?

docs/Configuring.md
```diff
@@ -148,6 +148,32 @@ On every poll, Cromwell will take at limited number of new submissions, provided
 system.max-workflow-launch-count = 50
 ```
 
+***Abort configuration***
+
+Cromwell will scan for abort requests using default configuration values equivalent to those below. In most circumstances
+there shouldn't be a need to override these defaults.
```
cjllanwarne: Would it be better to link to the appropriate section in `reference.conf` - just in case this changes over time?

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -42,10 +42,12 @@ object WorkflowManagerActor {
   sealed trait WorkflowManagerActorCommand extends WorkflowManagerActorMessage
   case object RetrieveNewWorkflowsKey
   case object RetrieveNewWorkflows extends WorkflowManagerActorCommand
-  final case class AbortWorkflowCommand(id: WorkflowId, restarted: Boolean) extends WorkflowManagerActorCommand
   case object AbortAllWorkflowsCommand extends WorkflowManagerActorCommand
   final case class SubscribeToWorkflowCommand(id: WorkflowId) extends WorkflowManagerActorCommand
   case object EngineStatsCommand extends WorkflowManagerActorCommand
+  case class AbortWorkflowsCommand(ids: Set[WorkflowId]) extends WorkflowManagerActorCommand
+
+  sealed trait WorkflowManagerActorResponse
```
cjllanwarne: Is this ever used?

engine/src/main/scala/cromwell/engine/workflow/WorkflowManagerActor.scala
```diff
@@ -243,6 +223,12 @@ class WorkflowManagerActor(params: WorkflowManagerActorParams)
   when (Done) { FSM.NullFunction }
 
   whenUnhandled {
+    case Event(AbortWorkflowsCommand(ids), stateData) =>
+      for {
+        id <- ids
+        actor <- stateData.workflows.get(id)
+      } yield actor ! WorkflowActor.AbortWorkflowCommand
+      stay()
```
Horneth: @mcovarr is this logic enough now ? It seems simpler compared to what was there before but was deleted https://github.com/broadinstitute/cromwell/pull/4401/files#diff-916acc6cffb855ee13688ea50412095fL172

cromwell.examples.conf
mcovarr: Ex post mergeo: the previous system did not try to shield WAs from duplicate abort requests, this is just a nicety. 🙂

PR #4405, opened 2018-11-19
Add new hash path strategy with last modified time

rhpvorderman:
Dear developers,

During testing I ran into the problem that the `HashPathStrategy` does not include the last modified date of the file. It assumes: "if the path is there, it is the same file". This is not necessarily the case. Files can be modified or replaced.Therefore the current `HashPathStrategy` is a big liability when trying to get reproducible results.

By adding a "last modified date" to the `HashPathStrategy` this will ensure that nothing has  happened to the file from the user or system side. This of course is not as safe as the `HashFileStrategy` since it does not protect against filesystem or hardware errors, but it provides a lot more safety compared to the current `HashPathStrategy`.  
This is also how Snakemake checks if files are the same and it works quite well.

Alternatively there could be an option in the Configfile that allows you to set this behaviour. Please let me know what you think of this.

### DIFF
--- a/docs/Configuring.md
+++ b/docs/Configuring.md
@@ -318,18 +318,15 @@ When running a job on the Config (Shared Filesystem) backend, Cromwell provides 
 
 ```HOCON
       config {
-        #...
         filesystems {
-          #...
           local {
-            #...
             caching {
               # When copying a cached result, what type of file duplication should occur. Attempted in the order listed below:
               duplication-strategy: [
                 "hard-link", "soft-link", "copy"
               ]
 
-              # Possible values: file, path
+              # Possible values: file, path, path+modtime
               # "file" will compute an md5 hash of the file content.
               # "path" will compute an md5 hash of the file path. This strategy will only be effective if the duplication-strategy (above) is set to "soft-link",
               # in order to allow for the original file path to be hashed.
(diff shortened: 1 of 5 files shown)


geoffjentry:
Hi - the vulnerability is known and viewed as a trade off for the user in the. And of speed. I’d recommend something like this be a new strategy altogether

rhpvorderman:
@geoffjentry 
Thanks for your input. Making it a new strategy would indeed be better. I will see if I can modify the pull request.

rhpvorderman:
@geoffjentry Something like this?

rhpvorderman:
Added testing as well. @cjllanwarne could you remove the back with originator label? Thanks!

docs/Configuring.md
```diff
@@ -316,13 +316,13 @@ Cromwell also accepts [Workflow Options](wf_options/Overview#call-caching-option
 
 When running a job on the Config (Shared Filesystem) backend, Cromwell provides some additional options in the backend's config section:
 
-```
+```HOCON
       config {
-        ...
+        #...
```
cjllanwarne: I actually think removing these `...`s would be fine given that this is HOCON and is extending an existing config file anyway

docs/Configuring.md
```diff
@@ -333,6 +333,7 @@ When running a job on the Config (Shared Filesystem) backend, Cromwell provides
               # "file" will compute an md5 hash of the file content.
               # "path" will compute an md5 hash of the file path. This strategy will only be effective if the duplication-strategy (above) is set to "soft-link",
               # in order to allow for the original file path to be hashed.
+              # "path+modtime" will compute an md5 hash of the file path and the last modified time. The same conditions as for "path" apply here.
```
cjllanwarne: Unfortunately, I can't add a comment to line 332, but could you add this to the `# possible values: file, path` list too

rhpvorderman:
@cjllanwarne . Thanks for the review. Made the minor changes.

PR #4407, opened 2018-11-19
Backport zip import logging to 35_hotfix

aednichols:
This extra logging should not merge into mainline `35_hotfix`; rather I'm using the PR as a mechanism for review and approval.

Once approved, I'll work with Mint to deploy `35_hotfix_mint_logging` to their Cromwell.

---

Mint is running `cromwell-35-5f86a05-SNAP.jar`, which refers to [commit `5f86a05`](https://github.com/broadinstitute/cromwell/commit/5f86a055a03bacbe842181783e3ea02bf778d6a6).

As far as I can tell, [`5f86a05`](https://github.com/broadinstitute/cromwell/commit/5f86a055a03bacbe842181783e3ea02bf778d6a6) is the same as [`b05e204`](https://github.com/broadinstitute/cromwell/pull/4091/commits/b05e2045eff34b179deb889f6c492dfd3c73b8b7), which is definitely present in `35_hotfix` (navigate to the [commit list](https://github.com/broadinstitute/cromwell/commits/35_hotfix?after=f3267d30aab008b00d04aeb6fdfcc35800b27c5e+34) and hit "Older" once).

They'll get a few extra commits, but nothing too crazy. I felt this was a more understandable strategy than applying my changes to the (seemingly random) commit they're running.

### DIFF
--- a/engine/src/main/scala/cromwell/engine/workflow/lifecycle/materialization/MaterializeWorkflowDescriptorActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/lifecycle/materialization/MaterializeWorkflowDescriptorActor.scala
@@ -272,7 +272,7 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
       if (importLocalFilesystem) DirectoryResolver.localFilesystemResolvers(None)
       else List.empty
 
-    val zippedResolverCheck: Parse[Option[ImportResolver]] = fromEither[IO](sourceFiles.importsZipFileOption match {
+    val zippedResolverCheck: Parse[Option[DirectoryResolver]] = fromEither[IO](sourceFiles.importsZipFileOption match {
       case None => None.validNelCheck
       case Some(zipContent) => zippedImportResolver(zipContent).toEither.map(Option.apply)
     })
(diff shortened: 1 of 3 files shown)


engine/src/main/scala/cromwell/engine/workflow/lifecycle/materialization/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -282,6 +282,13 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
     for {
       _ <- publishLabelsToMetadata(id, labels)
       zippedImportResolver <- zippedResolverCheck
+      // Debugging #4117 - make sure we are creating an importable directory out of the zip (AEN 2018-11-14)
+      _ = zippedImportResolver match {
+        case Some(dr: DirectoryResolver) =>
```
mcovarr: Why the ascription here? It appears to be redundant. If it turns out not to be redundant then there's a non-exhaustive match which would be a bigger problem. 🙂

aednichols:
We're pausing this investigation for now so it doesn't make sense to deploy an enhanced-logging version of Cromwell if we're not going to look at the logs for a while

PR #4408, opened 2018-11-19
Self contained Timing endpoint

### LINKED ISSUE #4309: Make the /timing response self-contained
There's a whole host of times when `/timing`'s redirect to `/metadata' causes problems.

One simple workaround (while we wait for `/timing` to "graduate" out of the Cromwell API) would be to make the call to metadata from *within* `CromwellApi`, eg:

Today:
-
- I call `/timing`
- I get a static HTML page back, which contains a request to `./metadata`
- That brings in the JSON object which contains timing information
- That gets used to build the diagram

Proposed
-
- I call `timing`
- Inside of `CromwellAPI`, the endpoint handler makes a call to the handler of `metadata()`
- It takes the JSON result from that and includes it as a static JSON inside the HTML response
- When I get the `HTML` page back, it's self contained and doesn't need to redirect out to any other endpoints to display the diagram

salonishah11:
Closes #4309

### DIFF
--- a/engine/src/main/scala/cromwell/webservice/routes/CromwellApiService.scala
+++ b/engine/src/main/scala/cromwell/webservice/routes/CromwellApiService.scala
@@ -105,8 +105,8 @@ trait CromwellApiService extends HttpInstrumentation with MetadataRouteSupport {
       instrumentRequest {
         onComplete(validateWorkflowId(possibleWorkflowId, serviceRegistryActor)) {
           case Success(workflowId) => completeTimingRouteResponse(metadataLookupForTimingRoute(workflowId))
-          case Failure(e: UnrecognizedWorkflowException) => e.errorRequest(StatusCodes.NotFound)
-          case Failure(e: InvalidWorkflowException) => e.errorRequest(StatusCodes.BadRequest)
+          case Failure(e: UnrecognizedWorkflowException) => e.failRequest(StatusCodes.NotFound)
+          case Failure(e: InvalidWorkflowException) => e.failRequest(StatusCodes.BadRequest)
           case Failure(e) => e.failRequest(StatusCodes.InternalServerError)
         }
       }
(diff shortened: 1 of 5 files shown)


engine/src/main/scala/cromwell/webservice/routes/CromwellApiService.scala
```diff
@@ -175,6 +179,29 @@ trait CromwellApiService extends HttpInstrumentation with MetadataRouteSupport {
...
+    val includeKeys = NonEmptyList.of("start", "end", "executionStatus", "executionEvents", "subWorkflowMetadata")
+    val readMetadataRequest = (w: WorkflowId) => GetSingleWorkflowMetadataAction(w, Option(includeKeys), None, expandSubWorkflows = true)
+
+    metadataBuilderRegulatorActor.ask(readMetadataRequest(workflowId)).mapTo[MetadataBuilderActorResponse]
+  }
+
+  private def completeTimingRouteResponse(metadataResponse: Future[MetadataBuilderActorResponse]) = {
+    onComplete(metadataResponse) {
+      case Success(r: BuiltMetadataResponse) => {
+        Try(Source.fromResource("workflowTimings/workflowTimings.html").mkString) match {
```
kshakir: ToL: I think this HTML is small enough that it could be saved in a string (lazy?) `val`, instead of decompressed from disk on every request.

cjllanwarne: I thought about suggesting this but walked back - I'd hope that `Source.fromResource` is smart enough to handle that kind of caching for us, and hopefully in a smarter way than we would come up with ad-hoc

kshakir: Talked IRL. No reason to over optimize at this time. If measured performance happens to be a problem, we'll fix it then.

PR #4423, opened 2018-11-26
Remove CWL syncs

Horneth:
- Renames the existing `Parse` to `IOChecked` <- I changed the name because it is now used in places that have nothing to do with parsing but this is still not a great name so suggestions welcome
- Uses `IOChecked` in place of `ErrorOr` or `Checked` to keep the operation async as long as possible. It's not all the way async yet but at least pushes the sync call further up
- Provides a value for the `cats.Parallel` typeclass over `CheckedIO` using `CheckedIOPar`. This gives the ability to use `parTraverse` and `parSequence` on collections and have the IO execute in parallel over the elements of the collection.

### DIFF
--- a/cwl/src/main/scala/cwl/internal/EcmaScriptEncoder.scala
+++ b/cwl/src/main/scala/cwl/internal/EcmaScriptEncoder.scala
@@ -33,7 +33,7 @@ class EcmaScriptEncoder {
     */
   def encode(value: WomValue): ECMAScriptVariable = {
     value match {
-      case file: WomFile => encodeFileOrDirectory(file, withSize = true)
+      case file: WomFile => encodeFileOrDirectory(file)
       case WomOptionalValue(_, None) => ESPrimitive(null)
       case WomOptionalValue(_, Some(innerValue)) => encode(innerValue)
       case WomString(string) => string |> ESPrimitive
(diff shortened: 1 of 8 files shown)


cwl/src/main/scala/cwl/internal/EcmaScriptEncoder.scala
```diff
@@ -106,10 +102,7 @@ class EcmaScriptEncoder(ioFunctionSet: IoFunctionSet) {
       Option("nameroot" -> (File.nameroot(file.value) |> ESPrimitive)),
       Option("nameext" -> (File.nameext(file.value) |> ESPrimitive)),
       file.checksumOption.map("checksum" -> ESPrimitive(_)),
-      if (withSize) 
-        sync(file.withSize(ioFunctionSet)).toOption.flatMap(_.sizeOption).map(Long.box).map("size" -> ESPrimitive(_))
-      else
-        None,
+      file.sizeOption.map(Long.box).map("size" -> ESPrimitive(_)),
       Option("secondaryFiles" -> encodeFileOrDirectories(file.secondaryFiles, withSize = false)),
```
kshakir: `withSize` should be removed, as it does not appear to be used, only recursively-plumbed-then-ignored.

PR #4427, opened 2018-11-27
[WDL Biscayne] Move escape evaluation into Cromwell (and make it work)

cjllanwarne:
- [ ] NB will require approval of openWDL PR https://github.com/openwdl/wdl/pull/272
- [ ] On merge, notify openWDL that https://github.com/openwdl/wdl/pull/247 can be merged

Problems with the old method:

- Quotes were not being escaped
  - eg this would fail: `String s = "a\"b"`
- Hex codes were not being interpreted by `wdl_unescape`:
  - eg this would fail: `String hex_hello = "\x68\x65\x6C\x6c\x6F"`
- Double length unicode codepoints were not being parsed:
  - eg this would fail: `String unicode_hello = "\u0068\U00000065\u006C\U0000006C\u006F"`

New method:

- Parse escape sequences in hermes grammar
- Allow Cromwell to interpret the escape sequences as it needs to

The main benefit to this was not having to mess with the `wdl_unescape` method in hermes to get the unrespected escape types to work.


wdl/model/draft3/src/main/scala/wdl/model/draft3/elements/StringEscapeSequence.scala
```diff
@@ -0,0 +1,27 @@
...
+import common.validation.ErrorOr.ErrorOr
+import wdl.model.draft3.elements.ExpressionElement.{BackslashEscape, DoubleQuoteEscape, NewlineEscape, SingleQuoteEscape, StringEscapeSequence, TabEscape, UnicodeCharacterEscape}
+
+object StringEscapeSequence {
+  val Octal = "\\\\([0-7]{3})".r
+  val Hex = "\\\\x([0-9a-fA-F]{2})".r
+  val FourDigitUnicode = "\\\\u([0-9a-fA-F]{4})".r
+  val EightDigitUnicode = "\\\\U([0-9a-fA-F]{8})".r
+
+  def parseEscapeSequence(seq: String): ErrorOr[StringEscapeSequence] = seq match {
```
aednichols: We test a wider range of escape sequences in `string_escaping.wdl` - will these still work?

Of course, we are allowed to change the behavior, with consensus - I doubt they see much use.

cjllanwarne: I'm not touching the WDL 1.0 parser in this change, which means that any WDL 1.0 string literal runs through the hermes-encapsulated `wdl_unescape` function and becomes a plain old `:string`.

This code will only run over `:escape` tokens. 

And yes, it's deliberate that the set is smaller in Biscayne - per openwdl/wdl#272

aednichols: To be clear I referred to the 1.0 file because it's just a convenient list to point at

geoffjentry:
@cjllanwarne Please remind me when this is merged as IMO this, plus your new openwdl PR fully satisfy the implementation requirement for the underlying wdl spec PR

womtool/src/test/scala/womtool/WomtoolValidateSpec.scala
```diff
@@ -13,7 +13,6 @@ class WomtoolValidateSpec extends FlatSpec with Matchers {
 
   behavior of "womtool validate"
 
-
```
kshakir: Unless you have strong opinions on this line, I'd revert this to keep the file from showing up in the change set.

wdl/model/draft3/src/main/scala/wdl/model/draft3/elements/ExpressionElement.scala
```diff
@@ -17,6 +17,18 @@ object ExpressionElement {
     */
   final case class StringPlaceholder(expr: ExpressionElement) extends StringPiece
 
+  sealed trait StringEscapeSequence extends StringPiece {
+    def unescape: String
+  }
+  case object NewlineEscape extends StringEscapeSequence { override val unescape: String = System.lineSeparator }
```
kshakir: I would either change `parseEscapeSequence()` to also use something like `case System.lineSeparator`, or change this to be `override val unescape = "\\n"`.

Otherwise on some systems the values may not match. Even if we don't support those systems, I still prefer consistency.

cjllanwarne: They're actually subtly different - `parseEscapeSequence` is looking at what escape sequence was in the WDL (eg if my script has `String s = "\n"`) - this line is specifying what value to replace the escape sequence with in the resulting scala `String` value

womtool/src/test/scala/womtool/WomtoolValidateSpec.scala
cjllanwarne: Because IMO two-line-breaks are worthy of shrinking (but mainly to avoid having to re-run this PR through Travis) I'll leave this in unless you strongly object?

PR #4429, opened 2018-11-27
Add a transaction isolation parameter with default, override it in select workflow store spots.

mcovarr:


database/sql/src/main/scala/cromwell/database/slick/WorkflowStoreSlickDatabase.scala
```diff
@@ -71,7 +71,7 @@ trait WorkflowStoreSlickDatabase extends WorkflowStoreSqlDatabase {
       )
     } yield workflowStoreEntries
 
-    runTransaction(action)
+    runTransaction(action, TransactionIsolation.ReadCommitted)
```
aednichols: It would be nice to briefly call out why it's appropriate to override the default in these two cases

mcovarr: Three cases, but actually I don't know why this is okay. Allowing non-repeatable read isolation in the workflow pickup in particular seems kind of dicey. Calling for backup @kcibul

aednichols: Not a merge blocker by any means FWIW

mcovarr: Updates based on Reading The Friendly MySQL docs. I removed `ReadCommitted` from two of the call sites since I don't think it was having any effect there: the abort scan is read-only and therefore non-locking, while the heartbeat writes are index-driven and should not have been affected by the change from `RepeatableRead` since only the updated rows were being locked anyway. The remaining transaction should be affected by the `ReadCommitted` change, hopefully my argument for its safety is plausible.

database/sql/src/main/scala/cromwell/database/slick/WorkflowStoreSlickDatabase.scala
```diff
@@ -62,6 +62,32 @@ trait WorkflowStoreSlickDatabase extends WorkflowStoreSqlDatabase {
...
+        updateCount <- dataAccess
+          .workflowStoreFieldsForPickup(workflowExecutionUuid)
+          .update((updateState, Option(cromwellId), Option(heartbeatTimestampTo)))
+        _ <- assertUpdateCount(
+          s"Update $workflowExecutionUuid to $updateState, heartbeat timestamp to $heartbeatTimestampTo",
+          updateCount,
+          1
+        )
+      } yield ()
+    }
```
mcovarr: No real changes here. This was really just a helper function in disguise before and is now scoped as one.

PR #4431, opened 2018-11-28
old mock is old and is exercising a dead code path. rejuvinate

geoffjentry:


engine/src/test/scala/cromwell/webservice/routes/CromwellApiServiceSpec.scala
```diff
@@ -101,30 +101,26 @@ class CromwellApiServiceSpec extends AsyncFlatSpec with ScalatestRouteTest with
         }
     }
 
-    it should "return 403 for abort of a workflow in a terminal state" in {
+    it should "return 200 Aborted for abort of a workflow which was already aborted" in {
```
mcovarr: The description is perhaps a bit unclear. "Aborted" will be returned when abort is requested for a workflow that was found in on-hold or submitted status, in which case Cromwell just blasts it out of the workflow store and that's the entirety of the abort process.

geoffjentry:
@cjllanwarne The issue was that reality had diverged from the mock, so the unit tests were testing fantasy land in the first place. The code was already doing something completely different in that situation, the test now just reflects that.

engine/src/test/scala/cromwell/webservice/routes/CromwellApiServiceSpec.scala
geoffjentry: Good point. I might as well make the mock even more lifelike to cover both of those situations

PR #4432, opened 2018-11-28
First draft of a swagger format for Workflow-Description-As-A-Service

cjllanwarne:
In case you don't want to fire up a server on this branch just to see the rendered example in the swagger UI...:

```json
{
  "valid": true,
  "errors": [
    "The 'errors' field will be filled if 'valid' is false",
    "We might also provide warnings to a 'valid' workflow here",
    "Otherwise, 'errors' will be empty or unspecified"
  ],
  "name": "my_wf",
  "inputs": [
    {
      "name": "my_wf.string_input",
      "valueType": {
        "typeName": "String"
      },
      "optional": false
    },
    {
      "name": "my_wf.array_input",
      "valueType": {
        "typeName": "Array",
        "arrayType": {
          "typeName": "String"
        }
      },
      "optional": false
    },
    {
      "name": "my_wf.optional_input",
      "valueType": {
        "typeName": "Optional",
        "optionalType": {
          "typeName": "String"
        }
      },
      "optional": true,
      "default": "hello"
    },
    {
      "name": "my_wf.map_input",
      "valueType": {
        "typeName": "Map",
        "mapType": {
          "keyType": {
            "typeName": "String"
          },
          "valueType": {
            "typeName": "Int"
          }
        }
      },
      "optional": false
    },
    {
      "name": "my_wf.object_input",
      "valueType": {
        "typeName": "Object",
        "objectFieldTypes": [
          {
            "fieldName": "int_field",
            "fieldType": {
              "typeName": "Int"
            }
          },
          {
            "fieldName": "int_array_field",
            "fieldType": {
              "typeName": "Array",
              "arrayType": {
                "typeName": "Int"
              }
            }
          }
        ]
      },
      "optional": false
    },
    {
      "name": "my_wf.int_string_pair_input",
      "valueType": {
        "typeName": "Tuple",
        "tupleTypes": [
          {
            "typeName": "Int"
          },
          {
            "typeName": "String"
          }
        ]
      },
      "optional": false
    }
  ],
  "images": [
    [
      "quay.io/seqware/seqware_full/1.1",
      "ubuntu:latest",
      "EXPRESSION_BASED_IMAGE"
    ]
  ],
  "submitted_descriptor_type": {
    "descriptorType": "WDL",
    "descriptorTypeVersion": "1.0"
  },
  "imported_descriptor_types": [
    {
      "descriptorType": "WDL",
      "descriptorTypeVersion": "1.0"
    }
  ],
  "meta": {
    "author": "Batman",
    "copyright": "BSD 3-clause"
  }
}
```

### DIFF
--- a/engine/src/main/resources/swagger/cromwell.yaml
+++ b/engine/src/main/resources/swagger/cromwell.yaml
@@ -765,13 +765,67 @@ definitions:
         description: Whether the workflow and inputs as submitted passes validation
       errors:
         type: array
-        items: string
+        items:
+          type: string
+        example:
+          - "Your workflow is bad because ..."
         description: If valid is false, a set of validation failure messages
+      name:
+        type: string
+        description: The name of the workflow in the description file
+        example: my_wf
       inputs:
         description: A list of inputs for this tool
         type: array
         items:
           $ref: '#/definitions/ToolParameter'
+        example:
+          - name: my_wf.string_input
+            valueType:
+              typeName: String
+            optional: false
+          - name: my_wf.array_input
+            valueType:
+              typeName: Array
+              arrayType:
+                typeName: String
+            optional: false
+          - name: my_wf.optional_input
+            valueType:
+              typeName: Optional
+              optionalType:
+                typeName: String
+            optional: true
+            default: "hello"
+          - name: my_wf.map_input
+            valueType:
+              typeName: Map
+              mapType:
+                keyType:
+                  typeName: String
+                valueType:
+                  typeName: Int
+            optional: false
+          - name: my_wf.object_input
+            valueType:
+              typeName: Object
+              objectFieldTypes:
+                - fieldName: "int_field"
+                  fieldType:
+                    typeName: Int
+                - fieldName: "int_array_field"
+                  fieldType:
+                    typeName: Array
+                    arrayType:
+                      typeName: Int
+            optional: false
+          - name: my_wf.int_string_pair_input
+            valueType:
+              typeName: Tuple
+              tupleTypes:
+                - typeName: Int
+                - typeName: String
+            optional: false
       images:
         type: array
         items:
@@ -788,20 +842,24 @@ definitions:
         type: array
         description: The type (or types) of descriptors available.
         items:
-          type: array
-          items:
-            type: object
-            required:
-              - name
-              - descriptorType
-              - descriptorTypeVersion
-            properties:
-              name:
-                type: string
-              descriptorType:
-                $ref: '#/definitions/DescriptorType'
-              descriptorTypeVersion:
-                type: string
+          type: object
+          example:
+            descriptorType: WDL
+            descriptorTypeVersion: "1.0"
+          required:
+            - descriptorType
+            - descriptorTypeVersion
+          properties:
+            descriptorType:
+              $ref: '#/definitions/DescriptorType'
+            descriptorTypeVersion:
+              type: string
+      meta:
+        type: object
+        description: An open set of key/value pairs echoing any workflow-level metadata found in the description file.
+        example:
+          author: "Batman"
+          copyright: "BSD 3-clause"
   ToolParameter:
     type: object
     description: >-
@@ -826,6 +884,12 @@ definitions:
       documentation:
         type: string
         description: Documentation for this parameter provided in the tool descriptor file.
+      parameter_meta:
+        type: object
+        description: An open set of key/value pairs echoing any custom parameter information in the description file.
+        example:
+          description: "This input provides the BAM file for processing"
+          localization_optional: true
   ValueType:
     description: The type expected for a given value.
     type: object
@@ -833,7 +897,7 @@ definitions:
     - valueType
     maxProperties: 2
     properties:
-      valueType:
+      typeName:
         type: string
         description: The type of this value
         enum:
@@ -865,7 +929,7 @@ definitions:
           properties:
             fieldName:
               type: string
-            fieldDescription:
+            fieldType:
               $ref: '#/definitions/ValueType'
   MapValueType:
     type: object
--- a/engine/src/main/resources/swagger/cromwell.yaml
+++ b/engine/src/main/resources/swagger/cromwell.yaml
@@ -130,7 +130,7 @@ paths:
         '200':
           description: Successful Request
           schema:
-            $ref: '#/definitions/WorkflowIdAndStatus'
+            $ref: '#/definitions/WorkflowDescription'
         '400':
           $ref: '#/responses/InvalidSubmission'
         '500':
--- a/engine/src/main/resources/swagger/cromwell.yaml
+++ b/engine/src/main/resources/swagger/cromwell.yaml
@@ -768,8 +768,10 @@ definitions:
         items:
           type: string
         example:
-          - "Your workflow is bad because ..."
-        description: If valid is false, a set of validation failure messages
+          - "The 'errors' field will be filled if 'valid' is false"
+          - "We might also provide warnings to a 'valid' workflow here"
+          - "Otherwise, 'errors' will be empty or unspecified"
+        description: The set of validation failure messages
       name:
         type: string
         description: The name of the workflow in the description file
@@ -838,22 +840,14 @@ definitions:
           All known docker images (and versions/hashes) used by this tool.
           If the tool will have to evaluate one or more docker image strings at runtime, the placeholder
           "EXPRESSION_BASED_IMAGE" will be used.
-      descriptor_types:
+      submitted_descriptor_type:
+        $ref: '#/definitions/DescriptorTypeAndVersion'
+        description: The type of descriptor submitted via the 'workflowSource' or 'workflowUrl' field.
+      imported_descriptor_types:
         type: array
-        description: The type (or types) of descriptors available.
         items:
-          type: object
-          example:
-            descriptorType: WDL
-            descriptorTypeVersion: "1.0"
-          required:
-            - descriptorType
-            - descriptorTypeVersion
-          properties:
-            descriptorType:
-              $ref: '#/definitions/DescriptorType'
-            descriptorTypeVersion:
-              type: string
+          $ref: '#/definitions/DescriptorTypeAndVersion'
+        description: The set of all types imported by the submitted workflow.
       meta:
         type: object
         description: An open set of key/value pairs echoing any workflow-level metadata found in the description file.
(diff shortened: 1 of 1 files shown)


cjllanwarne:
@denis-yuen I'd be really interested in your thoughts on this result format for "describe"ing a workflow as a service - and whether I could make small changes now to make it more useful as a way to feed descriptions into systems that subsequently host them (ala https://github.com/ga4gh/tool-registry-service-schemas).

engine/src/main/resources/swagger/cromwell.yaml
```diff
@@ -716,6 +756,209 @@ definitions:
...
+      errors:
+        type: array
+        items:
+          type: string
+        example:
+          - "The 'errors' field will be filled if 'valid' is false"
+          - "We might also provide warnings to a 'valid' workflow here"
+          - "Otherwise, 'errors' will be empty or unspecified"
+        description: The set of validation failure messages
+      name:
```
Horneth: Is this an optional or nullable field ? I'm thinking of the case where the workflow is so broken that it doesn't have a name (or a valid one)

cjllanwarne: Correct - only the `valid` field is listed at the top in the `required` section because it's the only field that's always required. 

As a side comment, OpenAPI 3 lets you be more expressive with `oneOf` format options where we could have separate "if valid" and "if invalid" formats and say that the result has to be "one of" the two..., but since we're still using Swagger 2, that kind of "name will only appear if `valid` is `true`" logic has to be informal

aednichols:
This looks good to me.

The only thing I have on my branch, that would be nice to see here, is placing the "HTTP calls to Womtool" endpoints in their own section.

My reasoning is that all the endpoints in the `Workflows` section somehow read or write to the state of workflows in Cromwell, while the Womtool-style endpoints will always be stateless.

Maintaining this distinction will probably help in future versions of WaaS when we have WaaS-only servers or cloud functions.

denis-yuen:
@cjllanwarne Hi, looks interesting. 
I'm probably missing some context to understand what this is for. 
To start, I see input parameters for a run of of a workflow (if I understand correctly) and the Docker images used by a workflow, but how are the WDL/CWL documents referenced in this?   Are one of the input types ("my_wf.string_input", "my_wf.array_input", etc.) containing the workflow files as opposed to parameters for this run of a workflow?

FYI, we could probably use your opinion on the (swagger 2.0 part of) https://github.com/ga4gh/tool-registry-service-schemas/pull/59

davidangb:
```
  "errors": [
    "The 'errors' field will be filled if 'valid' is false",
    "We might also provide warnings to a 'valid' workflow here",
    "Otherwise, 'errors' will be empty or unspecified"
  ],
```
could there be separate `errors` and `warnings` fields?

davidangb:
is the `Optional` type really necessary?

Here's a description of a required String input:
```
    {
      "name": "my_wf.string_input",
      "valueType": {
        "typeName": "String"
      },
      "optional": false
    },
```

could an optional String input simply be:
```
    {
      "name": "my_wf.optional_input",
      "valueType": {
        "typeName": "String"
      },
      "optional": true,
      "default": "hello"
    },
```

engine/src/main/resources/swagger/cromwell.yaml
```diff
@@ -95,6 +95,46 @@ paths:
...
+      - name: workflowSource
+        description: The workflow source file to submit for execution. Either workflow source or workflow url is required.
+        required: false
+        type: file
+        in: formData
+      - name: workflowUrl
+        description: URL which points to the workflow. Either workflow source or workflow url is required.
+        required: false
+        type: string
+        in: formData
```
davidangb: it looks like I can send either a file or a URL as input. Can I send raw WDL (or CWL, etc)? That's the Workbench use case.

engine/src/main/resources/swagger/cromwell.yaml
```diff
@@ -716,6 +756,208 @@ definitions:
...
+        type: array
+        items:
+          $ref: '#/definitions/DescriptorTypeAndVersion'
+        description: The set of all types imported by the submitted workflow.
+      meta:
+        type: object
+        description: An open set of key/value pairs echoing any workflow-level metadata found in the description file.
+        example:
+          author: "Batman"
+          copyright: "BSD 3-clause"
```
davidangb: where are the workflow `outputs`?

aednichols:
>is the Optional type really necessary?

@davidangb I think the current format makes a fair bit of sense because it's using the same representation for the type nested inside `Option` as the rest of the type system.

`Array[String]`:
```
"valueType": {
  "typeName": "Array",
  "arrayType": {
    "typeName": "String"
  }
}
```

`Option[String]` (not WDL syntax but same idea as `String?`):
```
"valueType": {
  "typeName": "Option",
  "optionalType": {
    "typeName": "String"
  }
}
```

If the concern is duplication of information in the response between the type and the `"optional"` field, I _think_ `"optional"` can be considered informational to save the frontend some parsing (perhaps @cjllanwarne can comment).

I do feel like perhaps the distinct `"arrayType"` and `"optionalType"` keys could be replaced by e.g. `"innerType"`.

engine/src/main/resources/swagger/cromwell.yaml
cjllanwarne: I think that should just be the file case

cjllanwarne:
Hi @denis-yuen thanks for having a look. And sorry, a little context would indeed have been useful for you!

This is a endpoint spec for a service that's answering questions like:
```
1. 
User: "Is <this> workflow valid"?
Service: yes/no

2.
User: what are the input parameters to <this> workflow?
Service: <names, types, and other info about the workflow's input parameters>

3.
User: what docker images does <this> workflow use?
Service: <list of images used in the WDL/CWL file>
```

As it turned out, it's the same amount of effort for us to answer question 1 as it is to answer 2/3 so we combined these questions into a single `describe` endpoint. The workflow source is an input to the API, the `inputs` in the swagger refer to Cromwell's response format for the input parameter definitions.

Does that help explain what's going on at all better?

engine/src/main/resources/swagger/cromwell.yaml
cjllanwarne: Good point! Is that something that you'll need? If so, we'll have to add something for `outputs`, presumably equivalent to `inputs`?

aednichols: Yes, Terra shows the outputs of a workflow on its "method configuration" screen

See for example https://github.com/DataBiosphere/saturn-ui/issues/423

cjllanwarne:
@davidangb re "do we need both types of optionals", I think @aednichols put it exactly right - 

* You need "optional" attached to the type - not just the input as a whole - to be able to say "array of optional `Int`s"
* It's presumably more convenient, since we already know it, for us to tell you an input as a whole is optional
* It's possible to have an optional input with an optional *and* an optional input with a non-optional type. 
  * For example in the workflow below, both `x` and `y` are optional in the inputs file, but `x` can be set to `"x": null` whereas `y`, if specified, must be a full, real, `Int`. So we need both the optional on the input value *and* the optional on the `Int` type to distinguish between these cases.
```wdl
workflow foo {
  input {
    Int? x = 5
    Int y = 5
  }
}
```

engine/src/main/resources/swagger/cromwell.yaml
davidangb: agreed - yes, we need the outputs. I imagine it would be equivalent to the inputs section.

davidangb:
gotcha on the need for an `optional` type. I see your points here. At a high level, my concern is making the response from this API easily parseable - and the large variety of key names (optionalType, arrayType, valueType, etc.) adds some complexity. I like @aednichols ' suggestion of using e.g. `innerType` where possible to reduce the set of distinct key names.

engine/src/main/resources/swagger/cromwell.yaml
davidangb: as a calling client, there's a structural difference between building a legal/valid `multipart/form-data` payload (more complex) and simply specifying a text clob (simpler). I'd really like to not jump through the extra hoops here. BUT ... after looking, I realize this is parallel to what Cromwell supports for submitting workflows. I prefer the consistency between the two APIs over making one API easier.

aednichols: Yeah, my WIP implementation routes requests through pretty much the same code as submitted workflows

( The WIP implementation is progressing on a separate branch to allow the Swagger spec to be discussed on its own here. )

engine/src/main/resources/swagger/cromwell.yaml
```diff
@@ -716,6 +756,208 @@ definitions:
...
+                  fieldType:
+                    typeName: Array
+                    arrayType:
+                      typeName: Int
+            optional: false
+          - name: my_wf.int_string_pair_input
+            valueType:
+              typeName: Tuple
+              tupleTypes:
+                - typeName: Int
```
ruchim: Is it that the order in which `typeName` is provided differentiates between the left and right components of the tuple?

engine/src/main/resources/swagger/cromwell.yaml
ruchim: In my mind, the only thing that may or may not apply to "outputs" is the field `images`. My mental abstraction is such that I think of both docker & inputs to be the left side of the equation and outputs on the right...
docker + inputs => outputs

engine/src/main/resources/swagger/cromwell.yaml
```diff
@@ -716,6 +756,208 @@ definitions:
...
+      valueType:
+        $ref: '#/definitions/ValueType'
+      optional:
+        type: boolean
+        description: Whether the tool allows this value to not be specified
+      default:
+        type: string
+        description: The in-language expression used to evaluate a default value for this parameter, if
+          none is supplied.
+      documentation:
```
ruchim: Is documentation essentially a first-class key of parameter_meta?

ruchim: ToL: There are certain types of workflows that we would want to feature and I would think we would expect all non-optional inputs to those workflows to require documentation. I'm thinking the groundwork laid here will help us head towards a distant future where one can "validate" a workflow such that it requires documentation.

engine/src/main/resources/swagger/cromwell.yaml
```diff
@@ -716,6 +756,208 @@ definitions:
...
+        description: An open set of key/value pairs echoing any custom parameter information in the description file.
+        example:
+          description: "This input provides the BAM file for processing"
+          localization_optional: true
+  ValueType:
+    description: The type expected for a given value.
+    type: object
+    required:
+    - valueType
+    maxProperties: 2
```
ruchim: This max is set to accommodate the tuple case?

ruchim: This might be a dumb question, but when using Objects to make structs, it could have more than 2 value types?

engine/src/main/resources/swagger/cromwell.yaml
```diff
@@ -716,6 +756,208 @@ definitions:
...
+        $ref: '#/definitions/ValueType'
+      arrayType:
+        $ref: '#/definitions/ValueType'
+      mapType:
+        $ref: '#/definitions/MapValueType'
+      tupleTypes:
+        type: array
+        items:
+          $ref: '#/definitions/ValueType'
+      objectFieldTypes:
```
ruchim: Oh is this covering structs?! if yes, ignore my comment above.

aednichols:
Should we include `parameter_meta`? I could see it appearing in a UI as a sort of optional description blurb.

PR #4447, opened 2018-12-01
Only assemble required subprojects for CWL conformance builds

mcovarr:
May help to unexplode conformance CI, at the very least it should run faster.

### DIFF
--- a/src/ci/bin/test.inc.sh
+++ b/src/ci/bin/test.inc.sh
@@ -525,7 +525,9 @@ cromwell::private::exists_cromwell_jar() {
 }
 
 cromwell::private::assemble_jars() {
-    CROMWELL_SBT_ASSEMBLY_LOG_LEVEL=error sbt coverage assembly -error
+    # CROMWELL_SBT_ASSEMBLY_COMMAND allows for an override of the default `assembly` command for assembly.
+    # This can be useful to reduce time and memory that might otherwise be spent assembling unused subprojects.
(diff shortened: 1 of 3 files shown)


src/ci/bin/test.inc.sh
```diff
@@ -525,7 +525,7 @@ cromwell::private::exists_cromwell_jar() {
 }
 
 cromwell::private::assemble_jars() {
-    CROMWELL_SBT_ASSEMBLY_LOG_LEVEL=error sbt coverage assembly -error
+    CROMWELL_SBT_ASSEMBLY_LOG_LEVEL=error sbt coverage ${CROMWELL_SBT_ASSEMBLY_COMMAND:-assembly} -error
```
kshakir: Favor: Can you leave a small comment explaining that this isn't necessary for our CI? SBT has a habit of changing around commands now and then, and it's possible we may have to re-simplify this in the future. I don't want someone fighting with maintaining backwards compatibility for this statement when it isn't required for our builds. FYI- The most recent case of this was removing the scala-cross-version-building support busted in SBT 1.0.x.

From what I understand this change is a recommendation for an external build environment that does not use this command here (yet?). True this change probably shave off 2 mins in Travis, but IMHO 1) this extra complexity isn't technically required in Travis and 2) doesn't shave off too much for a 60+ minute CI.

geoffjentry:
I can confirm that these changes do indeed work, but not as-is as i had to manually fold them into the cwl jenkins script. I'll let folks decide what to do w/ that information in regards to this PR (I'd seen @kshakir at some point request a comment regarding our own CI server)

PR #4448, opened 2018-12-02
BCBIO improvements

Horneth:
- Do not localize inputs if `inputDirMin` == 0
- Add a test for it and fix the existing WDL test for `localization_optional` on papiv2 which was passing but not really testing the feature
- Attempts to adjust the size of the working disk if the sum of the known sizes of input files is larger than what is asked for
"Attempts" because it only uses the size of files that have already been pre-fetched but does not take extra step to get the size of all of them.
In CWL we pre-fetch all the sizes to be able to evaluate expression.
In WDL this isn't needed so most files do not carry their size.


supportedBackends/google/pipelines/common/src/main/scala/cromwell/backend/google/pipelines/common/io/PipelinesApiAttachedDisk.scala
```diff
@@ -56,6 +58,14 @@ object PipelinesApiAttachedDisk {
       case _: IllegalArgumentException => s"$value not convertible to a Long".invalidNel
     }
   }
+  
+  implicit class EnhancedDisks(val disks: Seq[PipelinesApiAttachedDisk]) extends AnyVal {
+    def adjustWorkingDiskWithNewMin(minimum: MemorySize): Seq[PipelinesApiAttachedDisk] = disks map {
+      case disk: PipelinesApiWorkingDisk if disk == PipelinesApiWorkingDisk.Default && disk.sizeGb < minimum.to(MemoryUnit.GB).amount.toInt =>
+        disk.copy(sizeGb = minimum.to(MemoryUnit.GB).amount.toInt)
```
Horneth: Might be useful to log something when that happens ?

centaur/src/main/resources/standardTestCases/cwl_resources/cwl_inputdir_zero_doesnt_localize.cwl
```diff
@@ -0,0 +1,129 @@
...
+          glob: errors.txt
+          loadContents: true
+          outputEval: $(self[0].contents)
+      class: CommandLineTool
+      requirements:
+      - class: ShellCommandRequirement
+      - class: InlineJavascriptRequirement
+      arguments:
+      # Unfortunately we can't use the same trick as in draft3_nio_file_papi1 because in papi 2 the instance metadata doesn't seem to contain
+      # the PAPI operation ID. So instead simply check that the input files are not there. This is not as neat as the PAPI1 version
```
mcovarr: Maybe we should ask our friends to include this in the metadata.

centaur/src/main/resources/standardTestCases/wdl_draft3/draft3_nio_file/draft3_nio_file_papi2.wdl
```diff
@@ -0,0 +1,135 @@
...
+
+struct FileBox {
+  File val0
+  File val1
+}
+
+task mk_files {
+
+  parameter_meta {
+    salt: "Shakes things up a little bit!"
```
mcovarr: sigh

supportedBackends/google/pipelines/common/src/main/scala/cromwell/backend/google/pipelines/common/io/PipelinesApiAttachedDisk.scala
```diff
@@ -6,6 +6,8 @@ import cats.syntax.validated._
 import common.exception.MessageAggregation
 import common.validation.ErrorOr._
 import cromwell.core.path.{DefaultPathBuilder, Path}
+import wdl4s.parser.MemoryUnit
```
mcovarr: 👁 should we be importing from this package here?

cwl/src/main/scala/cwl/Tool.scala
```diff
@@ -115,6 +117,20 @@ trait Tool {
         _.select[SchemaDefRequirement].toList
       }.headOption.getOrElse(SchemaDefRequirement())
 
+      lazy val localizationOptionalMetaObject: MetaValueElement = MetaValueElementObject(
+        Map(
+          "localization_optional" -> MetaValueElementBoolean(true)
+        )
+      )
+
+      // If input dir min == 0 that's a signal not to localize the input files
```
kshakir: Give credit where credit's due. From what I understand this isn't in the official spec, so please list for what group this extra code is being added. If it does happen to be true spec, then please mention so.

geoffjentry: As it is part of support for a specific hint, if it's somehow feasible to codify that in a way other than a comment that'd be 💯

PR #4460, opened 2018-12-05
Allow for a 'name-for-call-caching-purposes' to override the backend name.

mcovarr:
Appears to work in local testing: I can rename my backend to any goofy name I want and call caching still works as long as `name-for-call-caching-purposes` is set to whatever the backend name used to be. Still working on the whole PAPI upgrade Centaur test thing and needs docs obvs.


ruchim:
Also connected to #4462

mcovarr:
@Horneth Yes I'll try that, I don't think it should be too bad...

cjllanwarne:
Could you add this to all the backends in the `reference.conf` and `cromwell.examples.conf` to demonstrate it in use?

mcovarr:
Temporarily closing as the centaur test is very much under construction.

Issue #4119, opened 2018-09-19
WOMtf happened to CWL in womtool?


aednichols:
```
java -jar /Users/anichols/Documents/Repos/cromwell/womtool/target/scala-2.12/womtool-35-fd560e9-SNAP.jar womgraph three_step.cwl
Failed to create wom Graph (reason 1 of 1): No getWomBundle method implemented in CWL v1
```

cf. [docs](https://cromwell.readthedocs.io/en/stable/WOMtool/), section Command Line Usage

aednichols:
The `CwlV1_0LanguageFactory` can `validateNamespace` (used in `MaterializeWorkflowDescriptorActor`) but lacks any concept of `WomBundle`, the output of `getWomBundle` and the input to `createExecutable`

dheiman:
If this ever does get fixed - the last version that didn't throw `No getWomBundle method implemented in CWL v1` (`31.1`) threw an error for me when I ran on https://github.com/NCI-GDC/gdc-dnaseq-cwl/blob/master/workflows/dnaseq/transform.cwl:

```
$ java -jar ~/bin/womtool-31.1.jar womgraph transform.cwl
Exception in thread "main" scala.MatchError: WomMaybePopulatedFileType (of class wom.types.WomMaybePopulatedFileType$)
	at womtool.graph.WomGraph$.fakeInput(WomGraph.scala:222)
	at womtool.graph.WomGraph$.$anonfun$womExecutableFromCwl$2(WomGraph.scala:205)
	at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:234)
	at scala.collection.immutable.HashMap$HashMap1.foreach(HashMap.scala:231)
	at scala.collection.immutable.HashMap$HashTrieMap.foreach(HashMap.scala:462)
	at scala.collection.TraversableLike.map(TraversableLike.scala:234)
	at scala.collection.TraversableLike.map$(TraversableLike.scala:227)
	at scala.collection.AbstractTraversable.map(Traversable.scala:104)
	at womtool.graph.WomGraph$.$anonfun$womExecutableFromCwl$1(WomGraph.scala:205)
	at scala.util.Either.map(Either.scala:350)
	at womtool.graph.WomGraph$.womExecutableFromCwl(WomGraph.scala:201)
	at womtool.graph.WomGraph$.fromFiles(WomGraph.scala:172)
	at womtool.Main$.$anonfun$womGraph$2(Main.scala:98)
[...]

PR #4467, opened 2018-12-07
/describe endpoint validates workflows

### LINKED ISSUE #4433: Implement the validate functionality for /describe
For the new /describe endpoint being added for the purposes of Womtool-as-a-Service, implement the ability to validate a given workflow source file.

AC: Given a workflow source file (WDL), return whether its a valid WDL workflow, and provide errors if its invalid.

{
  "valid": true,
  "errors": [
    "The 'errors' field will be filled if 'valid' is false",
    "We might also provide warnings to a 'valid' workflow here",
    "Otherwise, 'errors' will be empty or unspecified"
  ]
}

aednichols:
Closes #4433 

The changes to `MaterializeWorkflowDescriptorActor.scala` and `CromwellApiService.scala` are basically just refactoring out reusable code into a shared place.

It is a known issue (#4119) that Womtool cannot validate CWL; this limitation applies to the endpoint as well.


engine/src/main/scala/cromwell/webservice/routes/WomtoolRouteSupport.scala
```diff
@@ -0,0 +1,75 @@
...
+import cromwell.core.WorkflowSourceFilesCollection
+import cromwell.services.womtool.WomtoolServiceMessages.{DescribeFailure, DescribeRequest, DescribeResult, DescribeSuccess}
+import cromwell.services.womtool.WomtoolServiceMessages.JsonSupport.workflowDescriptionFormat
+import cromwell.webservice.WebServiceUtils
+import cromwell.webservice.WebServiceUtils.EnhancedThrowable
+
+import scala.concurrent.ExecutionContext
+import scala.util.{Failure, Success}
+
+trait WomtoolRouteSupport extends WebServiceUtils {
```
geoffjentry: Thanks for continuing the trend of not putting **everything** in CromwellApiService :) 💯

aednichols: Word, mega-classes are bad news

engine/src/main/scala/cromwell/engine/workflow/lifecycle/materialization/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -233,36 +231,9 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
...
-            case Some(other) => s"Unknown version '$other' for workflow language '$languageName'".invalidNel
-            case _ => chooseFactory(language.allVersions.values.toList).getOrElse(language.default).valid
-          }
-        case Some(other) => s"Unknown workflow type: $other".invalidNel[LanguageFactory]
-        case None =>
-          val allFactories = CromwellLanguages.instance.languages.values.flatMap(_.allVersions.values)
-          chooseFactory(allFactories.toList).getOrElse(CromwellLanguages.instance.default.default).validNel
-      }
+
+      val factory = LanguageFactoryUtil.chooseFactory(workflowSource, sourceFiles)
```
mcovarr: hooray

services/src/test/scala/cromwell/services/womtool/impl/WomtoolServiceInCromwellActorSpec.scala
```diff
@@ -0,0 +1,150 @@
...
+
+class WomtoolServiceInCromwellActorSpec extends ServicesSpec("Womtool") {
+
+  val womtoolActor = system.actorOf(WomtoolServiceInCromwellActor.props(ConfigFactory.empty(), ConfigFactory.empty(), TestProbe().ref))
+  CromwellLanguages.initLanguages(LanguageConfiguration.AllLanguageEntries)
+
+  object TestData {
+    val workflowUrlValid = "https://raw.githubusercontent.com/broadinstitute/cromwell/develop/womtool/src/test/resources/validate/wdl_draft3/valid/callable_imports/my_workflow.wdl"
+    val workflowUrlNotFound = "https://raw.githubusercontent.com/broadinstitute/cromwell/develop/my_workflow"
+    val workflowUrlBadHost = "https://zardoz.zardoz"
```
mcovarr: ugh you made me Google that... 😱

services/src/test/scala/cromwell/services/womtool/impl/WomtoolServiceInCromwellActorSpec.scala
```diff
@@ -0,0 +1,150 @@
...
+    val helloWorldInputs = """{"wf_hello.hello.addressee": "World"}"""
+    val bogusInputs = """{"foo.bar": "World"}"""
+    val wdlInvalid = "This is not a valid WDL."
+  }
+
+  "WomtoolServiceInCromwellActor" should {
+
+    "return valid for a valid workflow" in {
+
+      val wsfc = wsfcConjurer(workflowSource = Some(TestData.wdlValid))
```
mcovarr: `Option` all the things

engine/src/main/scala/cromwell/webservice/WebServiceUtils.scala
```diff
@@ -0,0 +1,61 @@
...
+
+trait WebServiceUtils {
+
+  type MaterializedFormData = Map[String, ByteString]
+
+  def materializeFormData(formData: Multipart.FormData)(implicit timeout: Timeout, materializer: Materializer, executionContext: ExecutionContext): Future[MaterializedFormData] = {
+    formData.parts.mapAsync[(String, ByteString)](1) {
+      bodyPart => bodyPart.toStrict(timeout.duration)(materializer).map(strict => bodyPart.name -> strict.entity.data)(executionContext)
+    }.runFold(Map.empty[String, ByteString])((map, tuple) => map + tuple)(materializer)
+  }
```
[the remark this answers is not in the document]
geoffjentry: It depends. If one is on a fairly strictly defined golden path there are easier ways to do this, but if you're off the path at all it's this monstrosity. I'll grab @aednichols w/ faces to see which this is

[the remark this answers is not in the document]
aednichols: FWIW this is just refactored out of `CromwellApiService.scala` with no changes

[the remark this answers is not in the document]
mcovarr: ok @geoffjentry may need to grab someone else's face

engine/src/main/scala/cromwell/webservice/WebServiceUtils.scala
```diff
@@ -0,0 +1,61 @@
...
+import akka.http.scaladsl.model.{Multipart, StatusCode}
+import akka.http.scaladsl.server.Route
+import akka.http.scaladsl.marshalling.ToEntityMarshaller
+import akka.http.scaladsl.model.headers.RawHeader
+import akka.http.scaladsl.server.Directives.complete
+import akka.stream.Materializer
+import akka.util.{ByteString, Timeout}
+import cromwell.webservice.routes.CromwellApiService._
+import cromwell.webservice.WorkflowJsonSupport.errorResponse
+import spray.json._
```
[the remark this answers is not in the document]
aednichols: This dependency moved over from `CromwellApiService` in concert with `EnhancedThrowable`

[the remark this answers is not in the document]
aednichols: It looks like it should be just a slight amount of work to swap over so I'll see how it goes

services/src/main/scala/cromwell/services/womtool/impl/WomtoolServiceInCromwellActor.scala
```diff
@@ -0,0 +1,81 @@
...
+import cromwell.services.womtool.WomtoolServiceMessages._
+import cromwell.util.GracefulShutdownHelper.ShutdownCommand
+import wom.core.WorkflowSource
+import wom.expression.NoIoFunctionSet
+
+class WomtoolServiceInCromwellActor(serviceConfig: Config, globalConfig: Config, serviceRegistryActor: ActorRef) extends WomtoolServiceActor {
+
+  override def receive: Receive = {
+    case DescribeRequest(filesCollection) =>
+      sender ! describeWorkflow(filesCollection)
```
cjllanwarne: Avoid blocking calls in receive blocks. IMO it would be best to async this callout to `describeWorkflow` even if that just means stuffing it into an in-memory queue that gets processed one at a time.

Otherwise if we get more than `mailboxSize` requests within the time it takes to make a few http lookups, we'll lose requests

services/src/main/scala/cromwell/services/womtool/impl/WomtoolServiceInCromwellActor.scala
```diff
@@ -0,0 +1,81 @@
...
+    case ShutdownCommand =>
+      // This service doesn't do any work on shutdown but the service registry pattern requires it (see #2575)
+      context.stop(self)
+  }
+
+  def describeWorkflow(wsfc: WorkflowSourceFilesCollection): DescribeResult = {
+
+    // The HTTP resolver is used to pull down workflows submitted by URL
```
cjllanwarne: Would it make sense to always have `HttpResolver` built into the `findWorkflowSource` method?

engine/src/main/scala/cromwell/webservice/WebServiceUtils.scala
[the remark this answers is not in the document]
cjllanwarne: >...Stranger thing...
> ...grab someone else's face...

what a thread...

Stranger Things

services/src/test/scala/cromwell/services/womtool/impl/WomtoolServiceInCromwellActorSpec.scala
cjllanwarne: Some might say Sean Connery's finest hour

engine/src/test/scala/cromwell/webservice/routes/WomtoolRouteSupportSpec.scala
```diff
@@ -0,0 +1,140 @@
...
+
+  val akkaHttpService = new MockWomtoolRouteSupport()
+  val version = "v1"
+
+  implicit def default = RouteTestTimeout(5.seconds)
+
+  behavior of "/describe endpoint"
+
+  object BodyParts {
+    val workflowSource  = Multipart.FormData.BodyPart("workflowSource", HttpEntity(MediaTypes.`application/json`, "This is not a WDL, but that's OK for this test of request routing."))
```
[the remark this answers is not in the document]
aednichols: TBH I hadn't considered that, that's just what we seem to use for `workflowSource` throughout the codebase (e.g. `CromwellApiServiceSpec`)

Using `text/plain(UTF-8)` works just fine though and seems more sane

services/src/main/scala/cromwell/services/womtool/WomtoolServiceMessages.scala
```diff
@@ -0,0 +1,26 @@
...
+    override def serviceName: String = WomtoolServiceName
+  }
+
+  case class DescribeRequest(filesCollection: WorkflowSourceFilesCollection) extends WomtoolServiceMessage
+
+  sealed trait DescribeResult extends WomtoolServiceMessage
+  case class DescribeSuccess(description: WorkflowDescription) extends DescribeResult
+  case class DescribeFailure(reason: String) extends DescribeResult
+
+  case class WorkflowDescription(valid: Boolean, errors: List[String])
```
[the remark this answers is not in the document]
aednichols: I see you state
>"Otherwise, 'errors' will be empty or unspecified"

in the spec PR but I think having the same set of keys every time (with an empty array) will make life on the frontend easier because one does not have to remember that some fields (but not all!) require a null-check before they're used

engine/src/main/scala/cromwell/webservice/WebServiceUtils.scala
```diff
@@ -0,0 +1,61 @@
...
+
+import scala.concurrent.{ExecutionContext, Future}
+
+trait WebServiceUtils {
```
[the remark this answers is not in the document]
aednichols: With a `trait` `WomtoolRouteSupport` can do
```
onComplete(materializeFormData(formData))
```
instead of
```
onComplete(WebServiceUtils.materializeFormData(formData))
``` 
which is shorter and makes it seem more like materializing form data is a fundamental capability of any class that's accepting forms (which I intend)

services/src/test/scala/cromwell/services/womtool/impl/WomtoolServiceInCromwellActorSpec.scala
Horneth: You mean Option might say...

aednichols: You can thank my CS program's functional programming class, e.g. https://www.cs.cornell.edu/courses/cs3110/2013sp/hw/ps1/ps1.html

I believe there is an annual Zardoz screening after the final, though I did not go

services/src/main/scala/cromwell/services/womtool/impl/WomtoolServiceInCromwellActor.scala
aednichols: >stuffing it into an in-memory queue that gets processed one at a time
Surely with all this advanced tooling lying around we should easily be able to do better? Is creating one `WomtoolServiceInCromwellActor` per request and making them run in parallel an option?

geoffjentry: There's a balance in there. Having it per request could easily flood the system. There are multiple ways to get some controlled parallelism, e.g. the more typical akka way would be to have a worker pool of a defined size pulling from the queue (but then you have to deal with the question of "what if the queue gets too big"). We handle similar situations with token counting as well. You could also have a specific thread pool and go with your plan (less good as too easy to go awry).

In general I've found that this is the sort of thing which rapidly winds up in premature optimization land (and frequently run afoul of it). One of the nice things about the actor model is that it makes it pretty easy to change the implementation to be more parallel as need be - e.g. replace the current actor w/ a router which is in front of a worker pool and the interface stays the same.

Which is all a long winded way of saying "I wouldn't worry about it just yet" :)

Another thing to consider here is that streaming models (akka or fs2) might be applicable here.

aednichols: Thanks, this is helpful. I’m happy to put exactly as much or as little sophistication into this as we feel is appropriate, just need some guidance to know how

engine/src/test/scala/cromwell/webservice/routes/CromwellApiServiceSpec.scala
```diff
@@ -558,6 +559,25 @@ object CromwellApiServiceSpec {
...
+            val readBack = List(
+              "this is fake data from the mock SR actor",
+              s"workflow hashcode: ${sourceFiles.workflowSource.map(_.hashCode)}",
+              s"workflow url: ${sourceFiles.workflowUrl}",
+              s"inputs: ${sourceFiles.inputsJson}",
+              s"type: ${sourceFiles.workflowType}",
+              s"version: ${sourceFiles.workflowTypeVersion}"
+            )
+
+            sender ! DescribeSuccess(description = WorkflowDescription(valid = true, readBack))
```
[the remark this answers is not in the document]
cjllanwarne: Gotcha. Would you be amenable to making them look more obviously like made-up errors? (eg `List("test error 1", "test error 2", ...)`)

services/src/main/scala/cromwell/services/womtool/WomtoolServiceMessages.scala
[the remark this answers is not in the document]
cjllanwarne: Agreed.

services/src/main/scala/cromwell/services/womtool/impl/WomtoolServiceInCromwellActor.scala
cjllanwarne: Right - there definitely are a whole host of "better" options! All I really care about for now is removing the blocking call in the `receive` method, further optimizations can be done as we find we need them (which may be very soon if we get a lot of timeouts replying to rawls... or it may be never)

services/src/main/scala/cromwell/services/womtool/WomtoolServiceActor.scala
```diff
@@ -0,0 +1,5 @@
+package cromwell.services.womtool
+
+import akka.actor.Actor
+
+trait WomtoolServiceActor extends Actor
```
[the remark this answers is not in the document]
aednichols: It used to do something, but no longer, so I deleted it

server/src/main/resources/application.conf
```diff
@@ -53,4 +53,11 @@ services {
       # See cromwell.examples.conf for details on settings one can use here
     }
   }
+  Womtool {
+    # Default - run within the Cromwell JVM
+    class = "cromwell.services.womtool.impl.WomtoolServiceInCromwellActor"
+    config {
+      # See cromwell.examples.conf for details on settings one can use here
```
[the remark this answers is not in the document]
aednichols: Nope, in fact the whole `config` block is useless, deleted

services/src/main/scala/cromwell/services/womtool/impl/WomtoolServiceInCromwellActor.scala
```diff
@@ -0,0 +1,94 @@
...
+  implicit val ec: ExecutionContext = context.dispatcher
+
+  override def receive: Receive = {
+    case DescribeRequest(filesCollection) =>
+
+      // We are consciously wrapping a Future around the Await.result way down in the HTTP import resolver until we can update the whole call hierarchy to async
+      // https://doc.akka.io/docs/akka/2.5.16/actors.html?language=scala#ask-send-and-receive-future
+      Future {
+        describeWorkflow(filesCollection)
+      } pipeTo sender()
```
Horneth: I know you've discussed it already. But just as a "ToL possibly going wrong scenario", since you run this actor on the service dispatcher (https://github.com/broadinstitute/cromwell/pull/4467/files#diff-b58faeefe405210df3e6d6c858cf66a9R93) and this schedules the `describeWorkflow` work on that dispatcher and immediately proceeds to process the next message in the mailbox, you could potentially flood the dispatcher (and its underlying execution context) with `describeWorkflow` work, which would impact all other services, including metadata and potentially slow everything down especially if `describeWorkflow(filesCollection)` is somewhat resource consuming or blocking for a non trivial amount of time.
If you've thought about this already then ignore this comment :)

aednichols: You're absolutely right about the potential dangers, I've added an extra warning to Swagger (in addition to changelog)
>This endpoint is undergoing active development. Its performance has not been validated and it **should not be called on a production instance of Cromwell**.

PR #4472, opened 2018-12-10
Boot disk auto-size in Papi v2

Horneth:
TL;DR: Boot disk size is increased automatically in Papi v2 if the (approximated) docker image size is > than what is requested (or the default one). Only works for Dockerhub and GCR at this moment because quay doesn't support yet manifest v2.

- Refactor of the DockerHashActor: got rid of akka stream, replaced with fs2 streams and cats.effect.IO. Easier to read/write, modify and maintain
- Instead of just getting the digest from the manifest header response, parse the content and get the size of the layer. The sum of the sizes == compressed size of the image
- The size of the layers is only available in the version 2 of the manifest schema https://docs.docker.com/registry/spec/manifest-v2-2/
GCR and Dockerhub support it, quay.io not yet (soon according to their support)
- Approximate the uncompressed size using a configurable compression factor. If the approximation is higher than the requested boot disk size, use that instead (only for Papi v2)
- The refactoring allows for: no need for explicitly differentiation gcr zones, which means `gcr.io`, `us.gcr.io`, `eu.gcr.io` and `asia.gcr.io` all work
- The refactoring also isolate each registry from each other with their own thread pool. Which means if say dockerhub is having very long response time all of sudden, gcr and quay aren't impacted


dockerHashing/src/main/scala/cromwell/docker/registryv2/DockerRegistryV2Abstract.scala
```diff
@@ -0,0 +1,285 @@
...
+  val ManifestV2MediaType = "application/vnd.docker.distribution.manifest.v2+json"
+  val ManifestListV2MediaType = "application/vnd.docker.distribution.manifest.list.v2+json"
+
+  // If one of those fails it means someone changed one of the strings above to an invalid one.
+  val ManifestV2MediaRange = MediaRange.parse(ManifestV2MediaType)
+    .unsafe("Cannot parse invalid manifest v2 content type. Please report this error.")
+  val ManifestListV2MediaRange = MediaRange.parse(ManifestListV2MediaType)
+    .unsafe("Cannot parse invalid manifest list v2 content type. Please report this error.")
+  val AcceptManifestV2Header = Accept.parse(ManifestV2MediaType)
+    .unsafe("Cannot parse invalid manifest v2 Accept header. Please report this error.")
```
kshakir: ToL: I don't know this particular API yet. Does it also log the actual header that failed to parse? Or does the user just get the string above with no other context? I also don't know the headers and if they contain sensitive info that _should not_ be logged.

ToL assuming this code was existing and just moved.

supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/GenomicsFactory.scala
```diff
@@ -83,11 +86,26 @@ case class GenomicsFactory(applicationName: String, authMode: GoogleAuthMode, en
         .gpuResource.map(toAccelerator).toList.asJava
       
       /*
-       * Adjust for the additional docker images Cromwell uses for (de)localization
-       * At the moment, google/cloud-sdk:slim (173MB) and stedolan/jq:latest (182MB)
-       * Round it up to 1GB
+       * Adjust using docker images used by Cromwell as well as
```
kshakir: Incomplete setence

PR #4476, opened 2018-12-10
Don't request more mem than most IntelliJ's need for sbt-import.

kshakir:


kshakir:
The removed `-Xms2g` was saying "Never run sbt, ever, without less than 2g of memory".

Meanwhile, Intellij has its own "[Maximum Heap Size](https://www.jetbrains.com/help/idea/sbt.html#82b10b37)" configuration value for the amount of memory required to import an sbt project. The IDE doesn't use this for running tests, so it does not need to be as large as the sbt options one uses for `sbt test` from the command line. The net effect of having the Intellij maximum less than 2g and the sbt opts _minimum_ at 2g caused  a cryptic error of: 

```
Error while importing sbt project:

Error occurred during initialization of VM
Initial heap size set to a larger value than the maximum heap size
```

This PR still leaves .sbtopts maximum amount of memory for running `sbt test` at 4g. It just no longer states that the JVM should start at 2g of memory.

kshakir:
Thx for the pointer to the outdated docs. Removed that section since that error should no longer occur in IntelliJ.

Issue #4235, opened 2018-10-11
Support submit by URL for CWL workflows


cjllanwarne:

cjllanwarne:
This appears to already work (see, eg https://github.com/broadinstitute/cromwell/blob/develop/centaur/src/main/resources/standardTestCases/cwl_glob_sort_with_workflow_url.test)

geoffjentry:
Let’s talk in person because I think the answer is it works sometimes

geoffjentry:
I just stood up a server locally.

Submitted this workflow: https://raw.githubusercontent.com/bcbio/bcbio_validation_workflows/master/NA12878-chr20/NA12878-platinum-chr20-workflow/main-NA12878-platinum-chr20.cwl

Used this as inputs json
```
{
    "checkerWorkflow.expectedMd5sum": "00579a00e3e7fa0674428ac7049423e2",
    "checkerWorkflow.inputFile": "md5sum.input"
}
```

Received this output:

```
2018-10-12 16:33:43,542  INFO  - Pre-Processing /var/folders/bs/wc6g67396rg8qnfj9qhvbvfsg7jhsw/T/cwl_temp_dir_822121598177295457/cwl_temp_file_7193cfe7-2342-48cc-8d7c-bf7d3434c57f.cwl
2018-10-12 16:33:58,870 cromwell-system-akka.dispatchers.engine-dispatcher-43 ERROR - WorkflowManagerActor Workflow 7193cfe7-2342-48cc-8d7c-bf7d3434c57f failed (during MaterializingWorkflowDescriptorState): cromwell.engine.workflow.lifecycle.materialization.MaterializeWorkflowDescriptorActor$$anon$1: Workflow input processing failed:
running cwltool on file /var/folders/bs/wc6g67396rg8qnfj9qhvbvfsg7jhsw/T/cwl_temp_dir_822121598177295457/cwl_temp_file_7193cfe7-2342-48cc-8d7c-bf7d3434c57f.cwl failed with Traceback (most recent call last):
  File "/Users/jgentry/projects/cromwell/server/target/scala-2.12/cromwell-36-82fd70f-SNAP.jar/Lib/heterodon/__init__.py", line 24, in apply
  File "<string>", line 1, in <module>
  File "<string>", line 12, in cwltool_salad
[...]

PR #4477, opened 2018-12-10
Update prealign to latest version

Horneth:
Update BCBIO prealign with the latest fix. Will fix the PAPIv2 CRON job.
At this point we should be able to run the workflows without modification. However there seems to be a bug preventing referencing the workflow by http directly, see https://github.com/broadinstitute/cromwell/issues/4235#issuecomment-445825939
If that worked we could point to it directly in the centaur test.
Otherwise we could modify the cron scripts to pull the git repos and submit using the local file.


Horneth:
@kshakir good question, they're the latest commits on master from here https://github.com/bcbio/test_bcbio_cwl

Issue #4458, opened 2018-12-05
Flakey Tests: CromwellTestKitSpec is a PITA


geoffjentry:
Multiple tickets have a root cause of CromwellTestKitSpec and are leading to frequent enough failures in QA Triage to warrant some eyeballs. These tickets include (but are not limited to):

#4457 
#4454 
#4453 
#4418 
#4469
#4470 
#4521

--

https://fc-jenkins.dsp-techops.broadinstitute.org/view/Testing/view/Test%20Runners/job/cromwell-test-runner/2438

PR #4486, opened 2018-12-11
Detect stalled workflows

cjllanwarne:
Resolves (or at least provides resolution to) one class of "my workflow never completes" problems.

We will *hopefully* only ever see this error during development, but it seems like a useful backstop condition to enforce in case it ever does happen to live workflows.


geoffjentry:
Could this be behind #4458 ?

cjllanwarne:
@geoffjentry I *think* this is coincidence (I'm trying to debug presumably unrelated `after` problems), but it's quite possible that those tests hit my "stalled" condition too since they appear to be showing the same symptoms.

In which case we might get some more useful debug clues out of this

geoffjentry:
@cjllanwarne got it, thanks. I did note that you said "one class" before commenting :) The belief **is** that those WFs are getting stuck and it's not just a dilation issue, so it'll be interesting to see if the failures dry up after this merges

PR #4415, opened 2018-11-21
Add test to recreate MySQL heartbeat deadlock.

kshakir:

Issue #4414, opened 2018-11-21
Recreate deadlock


ruchim:
AC: We require a unit test that recreates a deadlock occurs when workflow heartbeats are de-serialized (#4239). This is important because this is a problem we've seen before in production (prior to the serialization of heartbeats) and that's the *main* scenario we have to solve for as a requirement to be able to scale Cromwell horizontally.

Since we've only really seen deadlock behavior in production, it seems running a unit test at high scale could be one possible way to reproduce the deadlocking error. For example:

Start 10K workflows (that just sleep) and configure the heartbeat-interval rate to be the shortest duration possible (or whatever is reasonable). One would have to check the Cromwell server logs to see if there's a SQL exception

PR #4488, opened 2018-12-12
Create deadlocks in a multi-cromwell.

kshakir:


cjllanwarne:
> Create deadlocks in a multi-cromwell

... wait, what?

kshakir:
See #4414 for more info. We needed a reproducible test for deadlocks.

This adds PR adds a deadlock-creating-test at the high-level using multiple cromwells in a Docker Compose.

FYI- an earlier PR #4415 added very-very-low level deadlocks at the SQL/Database interface level.

database/sql/src/main/scala/cromwell/database/slick/SlickDatabase.scala
```diff
@@ -159,10 +160,30 @@ abstract class SlickDatabase(override val originalDatabaseConfig: Config) extend
...
   }
 
+  private val debugExitConfigPath = "danger.debug.only.exit-on-rollback-exception-with-status-code"
+  private val debugExitStatusCodeOption = ConfigFactory.load.getAs[Int](debugExitConfigPath)
+
   protected[this] def runTransaction[R](action: DBIO[R], isolationLevel: TransactionIsolation = TransactionIsolation.RepeatableRead): Future[R] = {
-    //database.run(action.transactionally) <-- See comment above private val actionThreadPool
-    val dbio = action.transactionally.withTransactionIsolation(isolationLevel)
-    Future(Await.result(database.run(dbio), Duration.Inf))(actionExecutionContext)
+    runAutoCommit(action.transactionally.withTransactionIsolation(isolationLevel))
```
mcovarr: Discussed with faces, this seemed contradictory but is actually just a naming issue.

database/sql/src/main/scala/cromwell/database/slick/SlickDatabase.scala
```diff
@@ -159,10 +160,30 @@ abstract class SlickDatabase(override val originalDatabaseConfig: Config) extend
...
+  }
+
+  protected[this] def runAutoCommit[R](action: DBIO[R]): Future[R] = {
+    //database.run(action) <-- See comment above private val actionThreadPool
+    Future {
+      try {
+        Await.result(database.run(action), Duration.Inf)
```
mcovarr: Is this functionally different from a Future with a recover for `MySQLTransactionRollbackException`?

src/ci/bin/testDockerDeadlock.sh
```diff
@@ -0,0 +1,53 @@
...
+docker image ls -q broadinstitute/cromwell:"${TEST_CROMWELL_TAG}" | grep . || \
+CROMWELL_SBT_DOCKER_TAGS="${TEST_CROMWELL_TAG}" sbt server/docker
+
+# Turn off exit-on-error temporarily, as we expect an error
+set +o errexit
+CROMWELL_TAG="${TEST_CROMWELL_TAG}" \
+docker-compose \
+  -f scripts/docker-compose-mysql/docker-compose-deadlock.yml \
+  up \
+  --scale cromwell=3 \
```
salonishah11: Just out of curiosity, why scale to only 3?

kshakir: Shrug. 3 was what was in Thibault's original example, and worked well enough.

database/sql/src/main/scala/cromwell/database/slick/SlickDatabase.scala
kshakir: Yes, it's _slightly_ different.

`try/catch` doesn't create a new `Future`, while using [`recover`](https://www.scala-lang.org/api/2.12.6/scala/concurrent/Future.html) would. `recover` requires plumbing through an execution context to create its new `Future`. That non-database EC is currently not plumbed through (but easily could be). Plus IMHO this java-esque hack is working well enough around the already existing hacks.

PR #4489, opened 2018-12-12
Power of 2s !

Horneth:


cjllanwarne:
Given that:
- The WDL spec today is un-opinionated on the size of an `Int`
- The WDL `Int` is defined as 64 bit in the upcoming https://github.com/openwdl/wdl/pull/250
  - NB: swap the red and green diffs in your head because Jeff merged it too soon and this is a placeholder "if we for some reason needed to revert" PR
- And assuming that people are not currently writing workflows specifically relying on an explosion if their Ints are above 2^16

I suggest you just default to making anything WDL produce`WomLong`s in all new situations, since we'll probably sweep through and update everything else at some point soon

supportedBackends/sfs/src/main/scala/cromwell/backend/impl/sfs/config/MemoryDeclarationValidation.scala
```diff
@@ -83,9 +85,13 @@ class MemoryDeclarationValidation(declaration: Declaration, attributeName: Strin
   }
 
   private def coerceMemorySize(womType: WomType)(value: MemorySize): WomValue = {
+    val amount = value.to(declarationMemoryUnit).amount
     womType match {
-      case WomIntegerType => WomInteger(value.to(declarationMemoryUnit).amount.toInt)
-      case WomFloatType => WomFloat(value.to(declarationMemoryUnit).amount)
+      case WomIntegerType if amount.inIntRange => WomInteger(amount.toInt)
+      // Should we fail here instead ?
```
kshakir: ToL: Can this be answered definitively before merging? Or, assuming WDL Int is going to be expanded, this comment could mention a github ticket/issue for followup? It looks like the previous version just failed.

supportedBackends/sfs/src/test/scala/cromwell/backend/impl/sfs/config/MemoryDeclarationValidationSpec.scala
```diff
@@ -96,7 +96,7 @@ class MemoryDeclarationValidationSpec extends FlatSpec with Matchers with TableD
       val extracted = memoryDeclarationValidation.extractWdlValueOption(validatedRuntimeAttributes)
 
       val expectedDefault = expectedDefaultAmount
-        .map(amount => WomInteger(MemorySize(amount.toDouble, MemoryUnit.GB).bytes.toInt))
+        .map(amount => WomLong(MemorySize(amount.toDouble, MemoryUnit.GB).bytes.toLong))
```
kshakir: Wild! I had no idea this WomLong's worked internally with WDL-parsing.

wom/src/main/java/wdl4s/parser/MemoryUnit.java
```diff
@@ -2,14 +2,10 @@
...
-    GiB(1L << 30, "GiB", "Gi"),
-    TiB(1L << 40, "TiB", "Ti"),
-    KB(1000, "KB", "K"),
-    MB(KB.bytes * 1000, "MB", "M"),
-    GB(MB.bytes * 1000, "GB", "G"),
-    TB(GB.bytes * 1000, "TB", "T");
+    KB(1L << 10, "KB", "K", "KiB", "Ki"),
+    MB(1L << 20, "MB", "M", "MiB", "Mi"),
+    GB(1L << 30, "GB", "G", "GiB", "Gi"),
+    TB(1L << 40, "TB", "T", "TiB", "Ti");
```
kshakir: 🙌 (Though I'm sure someone somewhere is going to disagree with this over-simplicity)

backend/src/test/scala/cromwell/backend/MemorySizeSpec.scala
```diff
@@ -65,26 +53,32 @@ class MemorySizeSpec extends FlatSpec with Matchers with TryValues {
...
     )
 
     forAll(memTable) { (memorySize, newUnit, result) =>
       memorySize.to(newUnit) shouldEqual result
     }
   }
+  
+  it should "round trip" in {
+    List(
+      "2 GB"
```
cjllanwarne: Seems like this is a test setup which might be expecting more than one test case?

centaur/src/main/resources/standardTestCases/cwl_resources_papiv2.test
```diff
@@ -8,7 +8,7 @@ files {
 
 metadata {
   status: Succeeded
-  "outputs.machineTypeTool.machine_type": "projects/1005074806481/machineTypes/custom-2-6912"
+  "outputs.machineTypeTool.machine_type": "projects/1005074806481/machineTypes/custom-2-7168"
```
cjllanwarne: [ToL] I hadn't seen this test before, it seems somewhat brittle if ever google decides to re-select their machine sizes? Oh well.... 🤷‍♂️

wom/src/main/java/wdl4s/parser/MemoryUnit.java
cjllanwarne: Hmm, this looks like a bit shift, but why is that in units of 10? I'm sure it makes sense but 😕

wom/src/main/scala/wom/types/WomIntegerLike.scala
```diff
@@ -5,4 +5,7 @@ object WomIntegerLike {
   implicit class EnhancedLong(val long: Long) extends AnyVal {
     def inIntRange: Boolean = long >= Int.MinValue && long <= Int.MaxValue
   }
+  implicit class EnhancedDouble(val double: Double) extends AnyVal {
+    def inIntRange: Boolean = double >= Int.MinValue && double <= Int.MaxValue
```
cjllanwarne: As well as not seeing this used anywhere else, I'm a little skeptical of this as something which is comparing `Double`s and `Int`s?

wom/src/main/java/wdl4s/parser/MemoryUnit.java
Horneth: Hmm to be honest I just kept what was already there for the `xiB` values. There are tests that verify that 1KB now is 1024 bytes so it seems right ?

PR #4490, opened 2018-12-13
`name-for-call-caching-purposes` and a couple of tests to prove it works

mcovarr:
Two new centaur tests (test types actually):

* Simulates the transition from PAPI v1 to PAPI v2. Start with an FC-like config where PAPI v1 is the default `Papi` backend and PAPI v2 is available in a `Papiv2` backend. Launch a workflow into this system with two sequential calls. Shut down Cromwell once the second call starts. Tweak the Cromwell config to add `name-for-call-caching-purposes = "Papi"` to the `Papiv2` backend. Bring Cromwell back up and let the previously-running workflow finish. Then launch a second workflow identical to the first except that its options request `{ "backend": "Papiv2"}`. Let this workflow finish and confirm it call cached both calls from the first workflow despite running on different versions of PAPI.

* Simulates the new steady-state after the v2 transition when new workflows are submitted with `{"backend": "Papiv2"}`. Run a workflow, shut down Cromwell in the middle of its execution, then bring Cromwell back up. Confirm this workflow completes successfully. Launch a second copy of the workflow also requesting `{"backend": "Papiv2"}` and confirm it completes successfully and caches to all the calls of the first run and that both executed on PAPI v2.


src/ci/bin/test.inc.sh
```diff
@@ -709,7 +716,12 @@ cromwell::build::setup_centaur_environment() {
     cromwell::private::create_centaur_variables
     if [ "${CROMWELL_BUILD_CENTAUR_TYPE}" = "${CROMWELL_BUILD_CENTAUR_TYPE_ENGINE_UPGRADE}" ]; then
         cromwell::private::setup_prior_version_resources
-    fi
+    elif [ "${CROMWELL_BUILD_CENTAUR_TYPE}" = "${CROMWELL_BUILD_CENTAUR_TYPE_PAPI_UPGRADE}" ]; then
+        cromwell::private::setup_prior_version_resources
+        export CROMWELL_BUILD_CROMWELL_CONFIG="${CROMWELL_BUILD_RESOURCES_DIRECTORY}/papi_v1_37_papi_v2_upgrade.application.conf"
+    elif [ "${CROMWELL_BUILD_CENTAUR_TYPE}" = "${CROMWELL_BUILD_CENTAUR_TYPE_PAPI_UPGRADE_NEW_WORKFLOWS}" ]; then
+        export CROMWELL_BUILD_CROMWELL_CONFIG="${CROMWELL_BUILD_RESOURCES_DIRECTORY}/papi_v1_37_papi_v2_upgrade.application.conf"
```
kshakir: 🤔 As `37` gets released, then `develop` moves to `38`, `39`, etc. -- will this still work as expected? Aka is this variable value "hard-coded" `_37_` still correct after the next release? There are other configs files with version numbers in their names. But there is bash code in `setup_prior_version_resources` than dynamically generates the current & previous version numbers, only detecting those config files for the correct version of the tests in `develop`.

CHANGELOG.md
```diff
@@ -32,6 +32,13 @@ More info [here](http://cromwell.readthedocs.io/en/develop/CallCaching/#call-cac
 - All memory units in WDL are now treated as base-2.
 For instance `1 KB == 1 KiB == 1024 Bytes`.
 
+### Backend name for call caching purposes
+
+Previous versions of Cromwell incorporated the name of the backend on which a call was run into the call cache hashes generated for that call.
+Unfortunately this made it impossible to change the name of a backend without losing all previously run calls as potential cache hits.
+Cromwell 37 introduces the `name-for-call-caching-purposes` backend configuration option as a means of decoupling the backend name from the
```
aednichols: [ToL] seems a bit verbose, could we say `call-caching-identifier`?

src/ci/bin/test.inc.sh
mcovarr: Discussed in person, leaving this as is for now and ticketing a decision on the future of this code. Perhaps all of this can just be deleted once v2 goes to production.

PR #4492, opened 2018-12-14
Fix BCBIO prealign in CRON

Horneth:
Also adds an env variable to filter by tag in Jenkins which makes it easier to only run a test or set of tests.


centaur/src/main/resources/integrationTestCases/cwl/bcbio/prealign-workflow/main-prealign-samples.json
```diff
@@ -287,6 +305,28 @@
             ]
         }
     ],
+    "genome_resources__variation__gnomad_exome": [
+        {
+            "class": "File",
+            "path": "gs://bcbiodata/test_bcbio_cwl/testdata/genomes/hg19/variation/gnomad_exome.vcf.gz",
```
Horneth: These file (and the `.tbi` are not in the `bcbiodata` bucket yet. I've emailed Brad to see if he can add them there and then this should work

geoffjentry:
I believe the necessary files are in place. If someone other than thibault can take the lead of taking it from there that’d be 💯

PR #4502, opened 2018-12-18
Ignore refresh token tests

aednichols:
I am under the impression that this feature is EOL and we should not bother fixing the tests.

~Next stop, delete the code and tests, but that's more than I can take on today.~

Let the 👍/👎 's decide!


aednichols:
If approved I will create an issue for Ruchi to triage as to whether we fix these or nerf the feature.

geoffjentry:
do it, rockapella

👍

aednichols:
Created https://github.com/broadinstitute/cromwell/issues/4504

Issue #4504, opened 2018-12-19
Refresh token functionality: fix tests or remove feature


aednichols:
In https://github.com/broadinstitute/cromwell/pull/4502 I ignored three tests related to the refresh token functionality because they were blocking all other merges and I had no idea how to fix them. My rationale for doing so is my impression the feature is unused/end-of-life.

@ruchim to decide whether we
1. Fix the tests
or
2. Delete the feature

kshakir:
If the feature is removed, also cleanse it from [the docs](https://cromwell.readthedocs.io/en/stable/search.html?q=refresh_token).

geoffjentry:
The feature is to be removed. I wouldn't put any effort into fixing the tests. Whether or not you want to put the effort in to excise the feature now is up to you for now :)

PR #4508, opened 2018-12-29
Horizontal Cromwell Deadlock Fix

kcibul:
Things Changed
- added a cromwell-master, which refreshes status metadata.  There can be only one.
- created a pool of cromwell_norefresh, which has status refresh turned off, these can scale
- found race condition when multiple cromwells try to create the liquibase lock table at once, configured to have master go first
- updated scripts/compose to handle the above two kinds of cromwell
- increased to 100 batches of 10 workflow each
- changed timeout script to show number of completed workflows and break when done
- delete database at start of run, so the above works
- ran heartbeats in auto-commit mode rather than in a single transaction
- dump out logs at end of run for debugging

Things I'd like to share
 - Lock Ordering in SELECT...FOR UPDATE no es bueno, there are great feature in MYSQL v8 (SKIP LOCKED) but we can't use those yet
 - how to configure mysql for query logging, and what it shows
 - heartbeat batches were never a batched update, just a big transaction
 - slick terminology can give give the wrong intuition
 - impact of cleaning db before each run
 - No deadlocks!

### DIFF
--- a/database/sql/src/main/scala/cromwell/database/slick/WorkflowStoreSlickDatabase.scala
+++ b/database/sql/src/main/scala/cromwell/database/slick/WorkflowStoreSlickDatabase.scala
@@ -116,7 +116,8 @@ trait WorkflowStoreSlickDatabase extends WorkflowStoreSqlDatabase {
         dataAccess.heartbeatForWorkflowStoreEntry(workflowExecutionUuid).update(heartbeatTimestampOption)
       })
     } yield counts.sum
-    runTransaction(action)
+    // Auto-commit mode, so each statement is individually committed to avoid deadlocks
+    runAction(action)
   }
 
   override def releaseWorkflowStoreEntries(cromwellId: String)(implicit ec: ExecutionContext): Future[Unit] = {
--- a/scripts/docker-compose-mysql/compose/cromwell/bin/test-deadlock.sh
+++ b/scripts/docker-compose-mysql/compose/cromwell/bin/test-deadlock.sh
@@ -22,10 +22,10 @@ done
 # Submit batches of workflows to the load balancer.
 BATCHES_TO_RUN=100
 WF_PER_BATCH=10
-TOTAL_WFS=$(( $BATCHES_TO_RUN * $WF_PER_BATCH ))
+TOTAL_WFS=$(( BATCHES_TO_RUN * WF_PER_BATCH ))
 for _ in $(seq $BATCHES_TO_RUN); do
-  echo `date`
-	curl \
+  echo $(date)
+  curl \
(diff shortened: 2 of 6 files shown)


scripts/docker-compose-mysql/compose/cromwell/bin/test-deadlock.sh
```diff
@@ -20,17 +20,30 @@ for host in "${cromwell_hosts[@]}"; do
 done
 
 # Submit batches of workflows to the load balancer.
-for _ in $(seq 50); do
+BATCHES_TO_RUN=100
+WF_PER_BATCH=10
```
mcovarr: TOL it's not immediately clear that the `BATCHES_TO_RUN` variable controls the number of batches to run but `WF_PER_BATCH` only describes the number of empty inputs currently hardcoded into `workflowInputs`.  These could be synced with something like

```shell
inputs=$(for ((i=1; i <= $WF_PER_BATCH; i++)); do echo -n '{},'; done)
.
.
.
-F "workflowInputs=[${inputs%?}]"
```

database/sql/src/main/scala/cromwell/database/slick/WorkflowStoreSlickDatabase.scala
```diff
@@ -116,7 +116,8 @@ trait WorkflowStoreSlickDatabase extends WorkflowStoreSqlDatabase {
         dataAccess.heartbeatForWorkflowStoreEntry(workflowExecutionUuid).update(heartbeatTimestampOption)
       })
     } yield counts.sum
-    runTransaction(action)
+    // Auto-commit mode, so each statement is individually committed to avoid deadlocks
+    runAction(action)
```
mcovarr: heck yeah

aednichols:
Yes, those tests broke and we have issue https://github.com/broadinstitute/cromwell/issues/4504 to address them.

For now I have merged PR https://github.com/broadinstitute/cromwell/pull/4502 which ignores them. I think merging `develop` into your branch will fix it (that's what happens for the "PR" build and it passeD)

aednichols:
Were you able to test whether the move from `runTransaction` to `runAction` affects performance?

Apologies if you already shared this at yesterday's refinement, I was still on vacation.

aednichols:
I dug up my source for my statement about autocommit:

>If autocommit mode is enabled, each SQL statement forms a single transaction on its own.

https://dev.mysql.com/doc/refman/5.7/en/innodb-autocommit-commit-rollback.html

danbills:
for posterity, @mcovarr commented "Heck Yeah!" and I accidentally deleted it...

PR #4510, opened 2019-01-02
Task monitoring action for PAPIv2

dinvlad:
Hi all,

As discussed with @TimothyTickle, @ruchim, @benjamincarlin, @gsaksena, @abaumann, @kshakir, @geoffjentry, and others at the Broad retreat and DSP holiday hackathon, we're putting a proposal for a new feature that reports task call resource utilization metrics to Stackdriver Monitoring API.

This serves 2 important goals:

1) Users can easily plot real-time resource usage statistics across all tasks in a workflow, or for a single task call across many workflow runs, etc.

    This can be very powerful to quickly determine outlier tasks that could use optimization, without the need for any configuration or code (or any changes to the workflow). It's also much easier than the current state-of-the-art, i.e. parsing task-level monitoring logs.

2) Scripts can easily get aggregate statistics on resource utilization and could produce suggestions based on those.

    This could provide a path towards automatic runtime configuration based on the models trained with historical data. One could also detect situations like out-of-memory calls and automatically adjust resources according to those.

It would also be pretty easy to add logic for estimation of task call-level cost based on the pricing of associated resources. This could provide a long-sought feature of real-time cost monitoring/control (thanks to @TimothyTickle for the suggestion).

Monitoring is done using the new "monitoring action" for PAPIv2, which currently uses the hard-coded [quay.io/broadinstitute/cromwell-monitor](https://quay.io/repository/broadinstitute/cromwell-monitor) image, built from https://github.com/broadinstitute/cromwell-monitor (I wasn't sure if that code belonged here or in a separate repo). This is advantageous to just using it as a _monitoring_script_, because it removes all assumptions on the "user" Docker image (for the task itself). For example, we don't have to assume a particular distribution or presence of Python and its libraries. So it should work exactly the same for any task.

Per @geoffjentry's suggestion, we've [consulted](https://groups.google.com/forum/#!topic/google-genomics-discuss/caYM7oHbfx0) with the Google Genomics team, and they don't see any apparent issues with the concept.

We could expose this as a workflow option like `monitoring_image`, and allow configuring it at the Cromwell level, so e.g. any user of Terra (or any other hosted Cromwell with PAPIv2 backend) could get usage reports without having to configure anything. The metrics are reported in their GCP project, so a user gets automatic access to them as long as they're a viewer.

We could also easily expose a link to workflow- and task-level reports in Job Manager UI, so they will be literally point-and-click away.

Each timepoint is designed to be self-sufficient, as it is labeled with:
- Cromwell-specific values, such as workflow ID, task call name, index and attempt.
- GCP instance values such as instance name, zone, number of CPU cores, total memory and disk size.

Here's an example graph of cpu/memory/disk utilization for one of our production workflows, as it is running right now - one can already see we could probably save ~40% of the cost:
screen shot 2019-01-02 at 4 43 20 pm

Reporting itself costs very little if anything at all, because Stackdriver provides a generous free tier worth ~65K instance-hours each month, and ~$0.0006 per instance-hour after that (at the current rate of 5 metric points reported each minute).

@kshakir suggested using a "vendor-neutral" reporting library such as [Micrometer.io](http://micrometer.io/), although I have reservations around that - mostly because that may require additional setup and we want this to "just work"; but also because the implementation is currently PAPIv2-specific anyway, so it is already non-vendor agnostic. Likewise, one could export metrics from Stackdriver monitoring if they wanted to. But we're open to the idea.

Finally, I haven't added any tests yet, as it's unclear in which shape or form (if at all) you'd like to integrate this code.

Thanks in advance for any feedback!
~[@broadinstitute/wintergreen](https://github.com/orgs/broadinstitute/teams/wintergreen) team.

### DIFF
--- a/docs/wf_options/Google.md
+++ b/docs/wf_options/Google.md
@@ -21,6 +21,6 @@ Keys | Possible Values | Description
   "google_compute_service_account": " my-new-svcacct@my-google-project.iam.gserviceaccount.com"
   "auth_bucket": "gs://my-auth-bucket/private",
   "monitoring_script": "gs://bucket/script.sh",
-  "monitoring_image": "gcr.io/my_google_project/my-monitoring-image"
+  "monitoring_image": "quay.io/broadinstitute/cromwell-monitor"
 }
 ```
--- a/docs/wf_options/Google.md
+++ b/docs/wf_options/Google.md
@@ -10,6 +10,7 @@ Keys | Possible Values | Description
 `refresh_token` |`string` |   Only used if `localizeWithRefreshToken` is specified in the [Configuration](../Configuring).
 `auth_bucket` |`string` |     A GCS URL that only Cromwell can write to.  The Cromwell account is determined by the `google.authScheme` (and the corresponding `google.userAuth` and `google.serviceAuth`). Defaults to the the value in [jes_gcs_root](#jes_gcs_root).
 `monitoring_script` |`string` |   Specifies a GCS URL to a script that will be invoked prior to the user command being run.  For example, if the value for monitoring_script is `"gs://bucket/script.sh"`, it will be invoked as `./script.sh > monitoring.log &`.  The value `monitoring.log` file will be automatically de-localized.
+`monitoring_image` |`string` |   Specifies a Docker image to monitor the task. This image will run concurrently with the task container, and provides an alternative mechanism to `monitoring_script` (the latter runs *inside* the task container). For example, one can use `quay.io/broadinstitute/cromwell-monitor`, which reports cpu/memory/disk utilization metrics to [Stackdriver](https://cloud.google.com/monitoring/).
 
 # Example
 ```json
@@ -19,6 +20,7 @@ Keys | Possible Values | Description
   "refresh_token": "1/Fjf8gfJr5fdfNf9dk26fdn23FDm4x",
   "google_compute_service_account": " my-new-svcacct@my-google-project.iam.gserviceaccount.com"
   "auth_bucket": "gs://my-auth-bucket/private",
-  "monitoring_script": "gs://bucket/script.sh"
+  "monitoring_script": "gs://bucket/script.sh",
+  "monitoring_image": "gcr.io/my_google_project/my-monitoring-image"
 }
 ```
--- a/supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/api/ActionBuilder.scala
+++ b/supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/api/ActionBuilder.scala
@@ -64,12 +64,35 @@ object ActionBuilder {
   def withImage(image: String) = new Action()
     .setImageUri(image)
 
+  /**
+    * Define a shared PID namespace for the monitoring container and its termination controller
+    */
+  private val monitoringPidNamespace = "monitoring"
+
   def monitoringAction(image: String, environment: Map[String, String], mounts: List[Mount] = List.empty): Action =
     new Action()
       .setImageUri(image)
       .withFlags(List(ActionFlag.RunInBackground))
       .withMounts(mounts)
       .setEnvironment(environment.asJava)
+      .setPidNamespace(monitoringPidNamespace)
+
+  /**
+    * monitoringTerminationAction is needed to gracefully terminate monitoring action,
+    * because PAPIv2 currently sends SIGKILL to terminate background actions.
+    *
+    * A fixed timeout is used to avoid hunting for monitoring PID.
+    */
+
+  private val monitoringTerminationImage = "alpine"
+  private val monitoringTerminationGraceTime = 10
+
+  def monitoringTerminationAction(): Action =
+    new Action()
+      .setImageUri(monitoringTerminationImage)
+      .withCommand(s"/bin/sh", "-c", s"kill -TERM -1 && sleep $monitoringTerminationGraceTime")
+      .withFlags(List(ActionFlag.AlwaysRun))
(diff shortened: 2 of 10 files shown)


supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/api/ActionBuilder.scala
```diff
@@ -64,6 +64,23 @@ object ActionBuilder {
...
+  /**
+    * The Docker image to report runtime metrics to Stackdriver Monitoring API.
+    * It is currently hard-coded, with the assumption that if folks use PAPI v2,
+    * they would want to get monitoring over a Google service as well.
+    *
+    * However, it could be made into a configuration option overridable
+    * at the workflow and/or Cromwell config levels, at the cost
+    * of additional complexity in configuration.
+    */
+  val monitoringImage = "quay.io/broadinstitute/cromwell-monitor"
```
mcovarr: Perhaps this should be in GCR if it's going to be pulled frequently from GCE VMs.

dinvlad: Yeah, that'd make sense!

dinvlad: Although come to think of it, we won't have to pay if a GCR bucket is multi-regional and accessed through PAPI only.

danbills:
Is this run on all jobs?  If so Is it potentially something a user would want to turn off?

danbills:
Also did you produce that graph manually? Is there a way to generate it easily for a workflow?

danbills:
I'm concerned that the awesomeness of this will be lost if there's not a dead-simple way to get the graph.  But I can accept that as a feature request on this already nice PR.

Put the code in `supportedBackends/google/pipelines/v2alpha1/src/main/resources/cromwell-monitor/`, please and thank you!

danbills:
Re: the option for turning it off:
Does stackdriver API need to be enabled for this to work?

dinvlad:
Nope as far as I know, it's turned on by default (and is used by GCP to monitor most resources): https://cloud.google.com/service-usage/docs/enabled-service

dinvlad:
We may need to verify the same applies to projects created via an API call though.

geoffjentry:
Please make this behavior configurable and defaulting to off

dinvlad:
Alright, would you prefer to expose it as a workflow (or Cromwell) option?

dinvlad:
Hmm, another moment is the Cromwell SA used by the GCE instances will need to have `Monitoring Metric Writer` role to be able to write metrics. I'll verify what happens when they don't. This is not an issue if the SA is Default Compute Engine account though.

geoffjentry:
that's a good question. I'd say to wire it in the same way as the current monitoring script option (don't have that answer easily available to me atm). 

re why default off, i've learned to be conservative w/ these sorts of things

dinvlad:
Makes sense! I'll add it as a workflow option for now.

dinvlad:
Another caveat is we can't measure network I/O this way. PAPIv2 doesn't (yet?) support specifying a network namespace for each action, so they all run in isolation.

dinvlad:
I exposed this as `monitoring_image` workflow option, and included the Dockerfile & script in `supportedBackends/google/pipelines/v2alpha1/src/main/resources/cromwell-monitor/`.

Should we also add a CI script that builds the image?

dinvlad:
Per @TimothyTickle's request, we're also adding a metric for cumulative cost of each task (as part of the same Python script). It will report the running cost of each task since its start. Once can also display cumulative cost of the entire workflow using secondary aggregation in Stackdriver UI.

geoffjentry:
@dinvlad I wouldn't bother doing that as there's already a workbench team embarking on cost management/reporting

dinvlad:
Does that include task-level cost reporting? Also we can do it here in pseudo-real time, as opposed to waiting. And this will apply to any Google project, not just workbench-managed one (and we're primarily non-workbench users ourselves). So I'd still like to include it.

geoffjentry:
@dinvlad I don't know all of the details but I'd be :-1: in adding that here until more shakes out.

dinvlad:
Alright, that's understandable. Thanks.

docs/wf_options/Google.md
```diff
@@ -19,6 +20,7 @@ Keys | Possible Values | Description
   "refresh_token": "1/Fjf8gfJr5fdfNf9dk26fdn23FDm4x",
   "google_compute_service_account": " my-new-svcacct@my-google-project.iam.gserviceaccount.com"
   "auth_bucket": "gs://my-auth-bucket/private",
-  "monitoring_script": "gs://bucket/script.sh"
+  "monitoring_script": "gs://bucket/script.sh",
+  "monitoring_image": "gcr.io/my_google_project/my-monitoring-image"
```
danbills: should use `quay.io/broadinstitute/cromwell-monitor` as you suggest above

dinvlad: Ok! If this is going to be the "official" url, we will need to 1) hand over control over that Quay repo to your team; 2) configure automatic builds in Quay triggered by the GitHub Cromwell repo (to build off of the monitoring Dockerfile in the Cromwell repo).

Alternatively, we could add a CI script to the Cromwell repo (though I'm not sure where and in which form).

Also, is everyone ok with the name (`cromwell-monitor`), or should it be more specific, like `cromwell-task-stackdriver-monitor` or similar?

Thanks!

dinvlad: Perfect, thanks. Submitting the fix for the doc.

PR #4516, opened 2019-01-05
Centaurify (de-CTKS) scatter test.

mcovarr:
Another CTKS test down.


core/src/test/scala/cromwell/util/SampleWdl.scala
```diff
@@ -479,64 +479,6 @@ object SampleWdl {
     )
   }
 
-  trait ZeroOrMorePostfixQuantifier extends SampleWdl {
```
mcovarr: oops, this block of red should have been deleted in the last CTKS PR.

mcovarr:
@rebrown1395 unassigning myself as there's currently nothing for me to do here: there are no reviewers and no feedback.

PR #4517, opened 2019-01-05
Relative import doc update

### LINKED ISSUE #4515: Failure to find imported WDLs using relative paths
When using relative path-based imports in a WDL, and imported WDL which also imports a WDL does not look in the right places for relative imports.
For example, given the directory structure:
```
import_issue
|   Issue.wdl
|
a___
|   |___A.wdl
|
|___b
    |___B.wdl
    |
    |___c
        |___C.wdl
```
where `Issue.wdl` imports `"a/A.wdl" as WorkflowA` and `"b/B.wdl" as WorkflowB` and `B.wdl` imports `import "b/c/C.wdl" as CTasks`, Cromwell is able to process and run the WDL in v34, but not in v36. It fails both submission to a Cromwell server and validation in Womtool.

Attached is a zip file containing this toy example.
[import_issue.zip](https://github.com/broadinstitute/cromwell/files/2728020/import_issue.zip)

aednichols:
Closes #4515

### DIFF
--- a/docs/Imports.md
+++ b/docs/Imports.md
@@ -12,11 +12,16 @@ There are two types of resources that are supported in imports: *http(s)* and *f
 import "http://mywdlrepository/my.wdl" as http_import1
 import "https://github.com/broadinstitute/cromwell/blob/master/engine/src/main/resources/3step.wdl" as http_import2
 ```
-To use a file-based import resource, you must provide a ZIP bundle of your resources and then use a path relative to that ZIP in your import statement. For example:
+To use a file-based import resource, provide a ZIP bundle of your resources and then use a path relative to that ZIP in your import statement. For example:
 
 ```wdl
 import "my-wdl-in-the-root-directory.wdl" as file_import1
 import "my/wdl/sub/directory/example.wdl" as file_import2
+```
+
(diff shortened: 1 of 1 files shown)


docs/Imports.md
```diff
@@ -12,13 +12,18 @@ There are two types of resources that are supported in imports: *http(s)* and *f
...
 ```
-To use a file-based import resource, you must provide a ZIP bundle of your resources and then use a path relative to that ZIP in your import statement. For example:
+To use a file-based import resource, provide a ZIP bundle of your resources and then use a path relative to that ZIP in your import statement. For example:
 
 ```wdl
 import "my-wdl-in-the-root-directory.wdl" as file_import1
 import "my/wdl/sub/directory/example.wdl" as file_import2
 ```
 
+If there exists a `my/wdl/sub/directory/imports/importing_an_import.wdl`, `my/wdl/sub/directory/example.wdl` could import it relatively like so:
```
cjllanwarne: hmm, I found this a little bit hard to understand. I might provide a bit of context, eg 

"imports from your submitted workflow are evaluated relative to the base of the zip file. In other case, import paths are relative to the file you are currently importing from, for example <<your example>>"

PR #4522, opened 2019-01-07
DAG view

cjllanwarne:
From the hackathon:

Does a few potentially positive things:
- Adds a `dot` format graph to metadata
- Implements `wom graph` for WOM (ie so it should work equally for all languages - CWL, WDL 1.0, et al)
- Provides a live status DAG endpoint, showing the current status of each step of a workflow as nodes within the graph of other dependencies.

And a potentially negative one:
- Changes up the `wom graph` output for draft-2 in potentially backwards incompatible ways (and is not tested so might be outright broken)


geoffjentry:
"From the" ..... ?

geoffjentry:
One off the cuff thought - should we add the dot to metadata? It's potentially quite large and the original WF is already part of the WF so the dot could always be recomputed

centaur/src/main/resources/standardTestCases/forkjoin/forkjoin.wdl
```diff
@@ -5,6 +5,7 @@
 
 task mkFile {
   command {
+    sleep 20
```
cjllanwarne: I doubt this was deliberate

cjllanwarne:
I'm going to close this PR as clutter, but I will leave the branch intact as a guide for if (hopefully when!) we circle back to this functionality

PR #4524, opened 2019-01-07
WorkflowFailSlow(andFast)Spec de-CTKS

mcovarr:


centaur/src/main/resources/standardTestCases/failing_fast_and_slow/fail.wdl
```diff
@@ -0,0 +1,78 @@
...
+    }
+}
+
+task failMeSlowly {
+    input {
+      Int a
+    }
+    command {
+        echo "The number was: ~{a}"
+        # Less than 2 so this should finish first
```
cjllanwarne: This seems like an incredibly brave assumption, especially in PAPI-land

mcovarr: Whilst Cromwell is incredibly brave you may have a point here.

PR #4525, opened 2019-01-07
Increase connection pool timeout for unit tests

mcovarr:
Makes the existing comment even more true.

### DIFF
--- a/core/src/test/resources/application.conf
+++ b/core/src/test/resources/application.conf
@@ -17,7 +17,7 @@ backend {
 }
 
 # Bumped up for tests
-database.db.connectionTimeout = 3000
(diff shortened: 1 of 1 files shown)


cjllanwarne:
It really takes 30 seconds for unit tests to get a connection? 🤢

mcovarr:
It really takes more than 3. 🙂

mcovarr:
@geoffjentry yes, this is turning the knob higher and hoping for the best. The downstream tests succeeded except for the one that depended on cross-talking with the failed test...

geoffjentry:
Discussed a bit w/ faces.

👍

PR #4534, opened 2019-01-09
[Biscayne] Struct value assignment directly from calls

cjllanwarne:
Part 2 of implementing the "Remove Object" [PR in OpenWDL](https://github.com/openwdl/wdl/pull/228).

Allows us to treat calls as structs of their outputs, for purposes of evaluating downstream expressions, eg:
```wdl
workflow blah { 
  call foo
  FooOutputStruct foo_outputs = foo
}

task foo {
  # ...

  output {
    Int i = ...
    String s = ...
  }
}

struct FooOutputStruct {
  Int i
  String s
}
```


cjllanwarne:
Needs some changes. Closing for now...

PR #4552, opened 2019-01-15
de-CTKS engine IO function testing

mcovarr:


centaur/src/main/resources/standardTestCases/workflow_engine_functions/workflow_engine_functions.wdl
```diff
@@ -0,0 +1,24 @@
...
+  String lines2 = "world"
+  String map1 = "foo\tbar"
+  String map2 = "baz\tqux"
+  Array[String] array_lines = [lines1, lines2]
+  Array[String] array_map = [map1, map2]
+  File array_file = write_lines(array_lines)
+  File map_file = write_lines(array_map)
+
+  Array[String] out_array = read_lines(array_file)
+  Map[String, String] out_map = read_map(map_file)
```
cjllanwarne: These two variations seem to be testing more the underlying code in `read_map` rather than the ability to call it from a workflow context?

Admittedly I'm assuming that they're ultimately calling into the same IO function so to me they look the same... Or are they genuinely implemented differently under the hood and worth checking separately?

centaur/src/main/resources/standardTestCases/workflow_engine_functions.test
```diff
@@ -0,0 +1,13 @@
...
+
+files {
+  workflow: workflow_engine_functions/workflow_engine_functions.wdl
+}
+
+metadata {
+  "outputs.w.hello": "hello"
+  "outputs.w.bar": "bar"
+  "outputs.w.world": "world"
+  "outputs.w.qux": "qux"
```
cjllanwarne: Could these be written in the same order that they would be produced? ie `hello`, `world`, `bar`, `qux`?

mcovarr: sure

centaur/src/main/resources/standardTestCases/workflow_engine_functions/workflow_engine_functions.wdl
mcovarr: My goal here was only to create a rough equivalent to the CTKS test I deleted, though I agree it would be a good idea to have more rigorous testing around workflow-context functions.

PR #4559, opened 2019-01-16
Reallow custom ImportResolvers

cjllanwarne:
There's no reason we should be precluding callers from bringing their own import resolvers


orodeh:
Thanks! This will work in v37. 

Is there a way to get the source code in cromwell v36?

cjllanwarne:
@orodeh I don't think we'll do a hotfix for this back into 36, but the better news is that 37 isn't too far away, so it shouldn't be too long before this gets released.

PR #4564, opened 2019-01-18
Auto-commit metadata inserts.

kshakir:


database/sql/src/main/scala/cromwell/database/slick/SlickDatabase.scala
```diff
@@ -164,10 +164,14 @@ abstract class SlickDatabase(override val originalDatabaseConfig: Config) extend
   private val debugExitStatusCodeOption = ConfigFactory.load.getAs[Int](debugExitConfigPath)
 
   protected[this] def runTransaction[R](action: DBIO[R], isolationLevel: TransactionIsolation = TransactionIsolation.RepeatableRead): Future[R] = {
-    runAction(action.transactionally.withTransactionIsolation(isolationLevel))
+    runActionInternal(action.transactionally.withTransactionIsolation(isolationLevel))
   }
 
   protected[this] def runAction[R](action: DBIO[R]): Future[R] = {
+    runActionInternal(action.withPinnedSession)
```
mcovarr: OOC did you find this to be necessary or just an optimization / keeping it more in line with `transactionally` behavior minus the transaction?

kshakir: The latter: "keeping it more in line with transactionally behavior minus the transaction"

PR #4566, opened 2019-01-18
Fix PAPI v2 USA auth handling in config

mcovarr:
The previous code was needlessly trying to create a Google Credential at Cromwell startup time. For user service accounts that's nonsensical and unpossible since USAs are specified per-workflow in workflow options.

### DIFF
--- a/src/ci/resources/papi_v2_34_application.conf
+++ b/src/ci/resources/papi_v2_34_application.conf
@@ -39,3 +39,15 @@ backend.providers.Papi-Caching-No-Copy.config.filesystems.http {
   project = "broad-dsde-cromwell-dev"
   auth = "service_account"
 }
+backend.providers.Papiv2USADockerhub.config.filesystems.http {
+  project = "broad-dsde-cromwell-dev"
+  auth = "service_account"
+}
+
+# The default v2 config sets this to "user_service_account" but that value isn't handled correctly in
+# versions prior to 37 so override this to be `service_account`.
+backend.providers.Papi.config.dockerhub.auth = "service_account"
+backend.providers.Papiv2.config.dockerhub.auth = "service_account"
+backend.providers.Papiv2RequesterPays.config.dockerhub.auth = "service_account"
+backend.providers.Papi-Caching-No-Copy.config.dockerhub.auth = "service_account"
+backend.providers.Papiv2USADockerhub.config.dockerhub.auth = "service_account"
--- a/supportedBackends/google/pipelines/common/src/main/scala/cromwell/backend/google/pipelines/common/authentication/PipelinesApiVMAuthentication.scala
+++ b/supportedBackends/google/pipelines/common/src/main/scala/cromwell/backend/google/pipelines/common/authentication/PipelinesApiVMAuthentication.scala
@@ -1,12 +1,13 @@
 package cromwell.backend.google.pipelines.common.authentication
 
 import cats.data.Validated.{Invalid, Valid}
-import com.google.api.client.googleapis.auth.oauth2.GoogleCredential
+import cats.syntax.validated._
+import common.validation.ErrorOr._
+import common.validation.Validation._
 import cromwell.cloudsupport.gcp.GoogleConfiguration
 import cromwell.cloudsupport.gcp.auth.ClientSecrets
 import cromwell.core.DockerCredentials
 import spray.json.{JsString, JsValue}
-
 /**
  * Interface for Authentication information that can be included as a json object in the file uploaded to GCS
  * upon workflow creation and used in the VM.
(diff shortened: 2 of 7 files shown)


cjllanwarne:
🇺🇸

src/ci/resources/papi_v2_34_application.conf
```diff
@@ -39,3 +39,15 @@ backend.providers.Papi-Caching-No-Copy.config.filesystems.http {
...
+  auth = "service_account"
+}
+
+# The default v2 config sets this to "user_service_account" but that value isn't handled correctly in
+# versions prior to 37 so override this to be `service_account`.
+backend.providers.Papi.config.dockerhub.auth = "service_account"
+backend.providers.Papiv2.config.dockerhub.auth = "service_account"
+backend.providers.Papiv2RequesterPays.config.dockerhub.auth = "service_account"
+backend.providers.Papi-Caching-No-Copy.config.dockerhub.auth = "service_account"
+backend.providers.Papiv2USADockerhub.config.dockerhub.auth = "service_account"
```
kshakir: Did you use this file manually?

ToL / Conflicting thoughts:

- This orphaned conf can be deleted. It's no longer automatically used by the CI tests in `develop`.

but

- There are still GOTC folks using 34, so one may need to prove to them that cromwell can do a multi-version jump.

supportedBackends/google/pipelines/common/src/main/scala/cromwell/backend/google/pipelines/common/authentication/PipelinesApiVMAuthentication.scala
```diff
@@ -34,19 +35,24 @@ object PipelinesApiDockerCredentials {
...
     }
 
-    new PipelinesApiDockerCredentials(
-      token = dockerCredentials.token,
-      keyName = dockerCredentials.keyName,
-      authName = dockerCredentials.authName,
-      credential = credential)
+    authValidation match {
+      case Invalid(errors) =>
+        throw new RuntimeException(errors.toList.mkString(", "))
```
danbills: I know this is embiggening the scope but we should return `Checked[PipelinesApiDockerCredentials]` rather than throw here.

supportedBackends/google/pipelines/common/src/main/scala/cromwell/backend/google/pipelines/common/authentication/PipelinesApiVMAuthentication.scala
```diff
@@ -34,19 +35,24 @@ object PipelinesApiDockerCredentials {
...
+      case Valid(_) =>
+        new PipelinesApiDockerCredentials(
```
danbills: If we return Checked we can move this code up to L39 and skip this second pattern match

supportedBackends/google/pipelines/common/src/main/scala/cromwell/backend/google/pipelines/common/authentication/PipelinesApiVMAuthentication.scala
```diff
@@ -34,19 +35,24 @@ object PipelinesApiDockerCredentials {
 
   def apply(dockerCredentials: DockerCredentials, googleConfig: GoogleConfiguration): PipelinesApiDockerCredentials = {
     // If there's an encryption key defined there must be a valid auth defined to encrypt it.
-    val credential = dockerCredentials.keyName flatMap { _ =>
-      val authName = dockerCredentials.authName.getOrElse(throw new RuntimeException("KMS Encryption key defined for private Docker but no auth specified"))
-      googleConfig.auth(authName) match {
-        case Invalid(e) => throw new RuntimeException("Error creating GoogleAuthMode: " + e.toList.mkString(", "))
-        case Valid(auth) => auth.apiClientGoogleCredential(identity)
-      }
+    val authValidation = dockerCredentials.keyName match {
```
danbills: ToL I hear the cool kids are using `fold` instead of pattern matching.  Something about less code

geoffjentry: OTOH Odersky has stated his disapproval with the cool kids so YMMV ;)

mcovarr: No danger of me being mistaken for a cool kid. Anyway I thought we collectively decided to generally avoid `fold`ing `Option`s due to the impact on readability?

supportedBackends/google/pipelines/common/src/main/scala/cromwell/backend/google/pipelines/common/authentication/PipelinesApiVMAuthentication.scala
mcovarr: Hmm unfortunately too many ripple effects for a stowaway change to a bugfix. That could always be made an enhancement request (to which I would tack on my own stowaway request for a multi-error capturing type that doesn't throw away stack traces 😛).

src/ci/resources/papi_v2_34_application.conf
kshakir: I'm fine either way. Under the "they who last touched it" principle be prepared to answer questions about this file though. 😉

PR #4578, opened 2019-01-24
Diet jq

danbills:
alpine-based image is ~ 4 MB, whereas the one we use is ~ 1 GB.

### DIFF
--- a/centaur/src/main/resources/standardTestCases/cwl_input_json/cwl_input_json.yaml
+++ b/centaur/src/main/resources/standardTestCases/cwl_input_json/cwl_input_json.yaml
@@ -44,7 +44,7 @@ $graph:
   class: CommandLineTool
   hints:
   - class: DockerRequirement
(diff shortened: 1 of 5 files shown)


cjllanwarne:
Seems like positive progress... 👍

mcovarr:
Not sure about switching from the canonical repo to an unknown source for something like this. Also it might be nice to upgrade from jq 1.5 to the 1.6 that was released a couple of months ago if we're going to change this.

danbills:
Is there something about 1.6 that is desirable?  

If we are very concerned I don't mind publishing our own

cjllanwarne:
@mcovarr I had the same thought but then `"stedolan/jq:latest"` didn't seem any more official to me than what we replaced it with

danbills:
closing this and opening a new one when I get time to figure out the best way to publish a docker image from our end

PR #4580, opened 2019-01-24
Add config for disabling sam submit whitelist.

kshakir:


CromIAM/src/main/scala/cromiam/sam/SamClient.scala
```diff
@@ -23,13 +23,22 @@ import scala.concurrent.{ExecutionContextExecutor, Future}
...
+                checkSubmitWhitelist: Boolean,
+                log: LoggingAdapter,
+                serviceRegistryActorRef: ActorRef)
                (implicit system: ActorSystem, ece: ExecutionContextExecutor, materializer: ActorMaterializer) extends StatusCheckedSubsystem with CromIamInstrumentation {
 
   override val statusUri = uri"$samBaseUri/status"
   override val serviceRegistryActor: ActorRef = serviceRegistryActorRef
 
   def isSubmitWhitelisted(user: User, cromIamRequest: HttpRequest): Future[Boolean] = {
+    if (!checkSubmitWhitelist) {
```
danbills: can we extract the impl below into another function?  Then can `fold` on this boolean value; could be a little tidyer

CromIAM/src/main/scala/cromiam/server/config/CromIamServerConfig.scala
```diff
@@ -54,19 +55,33 @@ final case class CromIamConfig(http: ServiceConfig, serverSettings: ServerSettin
 
 object CromIamConfig {
 
-  private def getValidatedServerSettings: ErrorOr[ServerSettings] = Try(ServerSettings(ConfigFactory.load())) match {
-    case Success(serverSettings) => serverSettings.validNel
-    case Failure(e) => s"Unable to generate server settings from configuration file: ${e.getMessage}".invalidNel
+  private def getValidatedServerSettings(conf: Config): ErrorOr[ServerSettings] = {
```
danbills: :100: for removing static call

CromIAM/src/main/scala/cromiam/sam/SamClient.scala
```diff
@@ -23,13 +23,22 @@ import scala.concurrent.{ExecutionContextExecutor, Future}
...
+      return Future.successful(true)
+    }
+
```
mcovarr: Any reason not to `if/else` this logic to avoid the `return`?

CromIAM/src/test/scala/cromiam/webservice/MockClients.scala
```diff
@@ -95,6 +102,10 @@ class MockSamClient()(implicit system: ActorSystem,
   }
 
   override def isSubmitWhitelisted(user: User, cromIamRequest: HttpRequest): Future[Boolean] = {
+    if (!checkSubmitWhitelist) {
+      return super.isSubmitWhitelisted(user, cromIamRequest)
```
mcovarr: same request here

CromIAM/src/main/scala/cromiam/sam/SamClient.scala
kshakir: Reason, sure? But not the strongest one. Using `else` means indenting the rest of the function. Which then increases the size of the diff.

PR #4581, opened 2019-01-24
Clean up unzipped directories

### LINKED ISSUE #4406: Cleanup routine for unzipped imports directory
When [creating the zipped import resolver](https://github.com/broadinstitute/cromwell/blob/develop/languageFactories/language-factory-core/src/main/scala/cromwell/languages/util/ImportResolver.scala#L133), we create a zip file on disk with format
```
/tmp/imports_workflow_a5756f37-a2dc-4e57-be4e-0d0531812fef_5414558548666056879.zip
```
and then unzip it into a directory like
```
/tmp/imports_workflow_a5756f37-a2dc-4e57-be4e-0d0531812fef_5414558548666056879.zip3360767550785143230
```
We immediately clean up the temporary file, but the temporary directory sticks around forever.

Import resolvers are instantiated from scratch during restart, so it should be safe to clean up the directory as soon as the workflow stops.

---

For extra credit:

Writing temporary items to disk has two very undesirable features:
- Global scope
- Infinite lifetime

Find and adopt a solution that lets us completely isolate workflows from each other's filesystem side-effects, and lets the JVM do the cleanup for us with garbage collection as soon as the workflow stops. I think we could pull this off with NIO, but I have not done my homework on this.

aednichols:
Closes #4406 

Might help #4117

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Cromwell Change Log
 
+## 38 Release Notes
+
+### Bug fixes
+- Fixed an issue that could cause Cromwell to consume disk space unnecessarily when using zipped dependencies
+
 ## 37 Release Notes
 
 ### Docker
@@ -76,7 +81,6 @@ The endpoint is still undergoing heavy development and should not be used in pro
 ### Bug fixes
 
 - Fixed a regression in Cromwell 36 that could cause operations on empty arrays to fail with a spurious type error (closes [#4318](https://github.com/broadinstitute/cromwell/issues/4318))
-- Fixed an issue that could cause Cromwell to consume disk space unnecessarily when using zipped dependencies
 
 #### Abort On Hold Workflows
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,7 @@ The endpoint is still undergoing heavy development and should not be used in pro
 ### Bug fixes
 
 - Fixed a regression in Cromwell 36 that could cause operations on empty arrays to fail with a spurious type error (closes [#4318](https://github.com/broadinstitute/cromwell/issues/4318))
+- Fixed an issue that could cause Cromwell to consume disk space unnecessarily when using zipped dependencies
 
 #### Abort On Hold Workflows
 
--- a/engine/src/main/scala/cromwell/engine/workflow/lifecycle/materialization/MaterializeWorkflowDescriptorActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/lifecycle/materialization/MaterializeWorkflowDescriptorActor.scala
@@ -275,6 +275,7 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
       factory <- findFactory(sourceAndResolvers._1).toIOChecked
       outputRuntimeExtractor <- factory.womOutputRuntimeExtractor.toValidated.toIOChecked
       validatedNamespace <- buildValidatedNamespace(factory, sourceAndResolvers._1, sourceAndResolvers._2)
+      _ = sourceAndResolvers._2.foreach(_.close())
       _ = pushNamespaceMetadata(validatedNamespace)
       ewd <- buildWorkflowDescriptor(id, validatedNamespace, workflowOptions, labels, conf, pathBuilders, outputRuntimeExtractor).toIOChecked
     } yield ewd
--- a/engine/src/main/scala/cromwell/engine/workflow/lifecycle/materialization/MaterializeWorkflowDescriptorActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/lifecycle/materialization/MaterializeWorkflowDescriptorActor.scala
@@ -277,7 +277,7 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
       factory <- findFactory(sourceAndResolvers._1).toIOChecked
       outputRuntimeExtractor <- factory.womOutputRuntimeExtractor.toValidated.toIOChecked
       validatedNamespace <- buildValidatedNamespace(factory, sourceAndResolvers._1, sourceAndResolvers._2)
-      closeResult = sourceAndResolvers._2.traverse(_.close())
+      closeResult = sourceAndResolvers._2.traverse(_.cleanupIfNecessary())
       _ = pushNamespaceMetadata(validatedNamespace)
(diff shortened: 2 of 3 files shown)


aednichols:
I think this solution is failing tests because it also deletes directories used by `DirectoryResolver` that _did not_ come from unzipping imports

aednichols:
Fixed the above issue.

CHANGELOG.md
```diff
@@ -76,6 +76,7 @@ The endpoint is still undergoing heavy development and should not be used in pro
 ### Bug fixes
 
 - Fixed a regression in Cromwell 36 that could cause operations on empty arrays to fail with a spurious type error (closes [#4318](https://github.com/broadinstitute/cromwell/issues/4318))
+- Fixed an issue that could cause Cromwell to consume disk space unnecessarily when using zipped dependencies
```
cjllanwarne: This should now move into version 38's bug fixes

engine/src/main/scala/cromwell/engine/workflow/lifecycle/materialization/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -275,6 +275,7 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
       factory <- findFactory(sourceAndResolvers._1).toIOChecked
       outputRuntimeExtractor <- factory.womOutputRuntimeExtractor.toValidated.toIOChecked
       validatedNamespace <- buildValidatedNamespace(factory, sourceAndResolvers._1, sourceAndResolvers._2)
+      _ = sourceAndResolvers._2.foreach(_.close())
```
cjllanwarne: This means that any failure in `close()` will  be ignored. Which might be what you intended, but in that case it feels awkward that we're returning `Try`s from them in any case.

As an exercise in "wouldn't it be nice"ing, perhaps `close` could return an ErrorOr and we could traverse the `close()`s here to get a single "did anything fail" result. And then separately we could decide whether to log, ignore, or fail the workflow as a result?

languageFactories/language-factory-core/src/main/scala/cromwell/languages/util/ImportResolver.scala
```diff
@@ -66,7 +67,8 @@ object ImportResolver {
 
   case class DirectoryResolver(directory: Path,
                                dontEscapeFrom: Option[String] = None,
-                               customName: Option[String]) extends ImportResolver {
+                               customName: Option[String],
+                               deleteOnClose: Boolean = false) extends ImportResolver {
```
cjllanwarne: Just OOI, where is this not true (it seems like this was a deliberate choice)?

aednichols: Because my impression is that most directory resolvers are instantiated around a directory of WDLs a user created themselves, and we don't want to delete those

engine/src/main/scala/cromwell/engine/workflow/lifecycle/materialization/MaterializeWorkflowDescriptorActor.scala
aednichols: I chose `Try` somewhat arbitrarily and fully support adopting the latest and greatest `ErrorOr`

That said there's nothing we can do about a failure ** for this particular case **

cjllanwarne: I would think a log message at least is appropriate?

languageFactories/language-factory-core/src/main/scala/cromwell/languages/util/ImportResolver.scala
cjllanwarne: Right.... very good point!

engine/src/main/scala/cromwell/engine/workflow/lifecycle/materialization/MaterializeWorkflowDescriptorActor.scala
aednichols: Agreed, I missed that suggestion

ruchim:
@mcovarr would you mind re-reviewing plz? you're the best.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Cromwell Change Log
 
+## 38 Release Notes
+
+### Bug fixes
+- Fixed an issue that could cause Cromwell to consume disk space unnecessarily when using zipped dependencies
```
cjllanwarne: "`could`" seems optimistic here

engine/src/main/scala/cromwell/engine/workflow/lifecycle/materialization/MaterializeWorkflowDescriptorActor.scala
```diff
@@ -275,9 +277,18 @@ class MaterializeWorkflowDescriptorActor(serviceRegistryActor: ActorRef,
       factory <- findFactory(sourceAndResolvers._1).toIOChecked
       outputRuntimeExtractor <- factory.womOutputRuntimeExtractor.toValidated.toIOChecked
       validatedNamespace <- buildValidatedNamespace(factory, sourceAndResolvers._1, sourceAndResolvers._2)
+      closeResult = sourceAndResolvers._2.traverse(_.close())
```
cjllanwarne: TOL: I just realized that this calls `close()` on our http importer too. I assume this is OK because the HTTP importer `close()` doesn't actually do anything, although that in itself might be confusing.

I wonder whether we should rename this method to `removeTempDirectories` to capture than nuance a little clearer?

aednichols: Does `removeTempDirectories` make sense on an HTTP resolver that has no concept of directories?

CHANGELOG.md
aednichols: Deliberate wording; if someone is running Cromwell on a Linux distro that e.g. cleans up `/tmp` every 24 hours rather than on every reboot, they will have never seen this bug

engine/src/main/scala/cromwell/engine/workflow/lifecycle/materialization/MaterializeWorkflowDescriptorActor.scala
aednichols: How about `cleanupIfNecessary`

PR #4583, opened 2019-01-25
Temporary patch: Use an encrypted variable for the vault token.

kshakir:


kshakir:
I've manually confirmed that various Travis builds are already reporting `vault auth success` and rendering the secrets.

As this diff otherwise shouldn't affect the build, merging early to unblock other PRs.

PR #4585, opened 2019-01-25
Revert: Temporary patch: Use an encrypted variable for the vault token.

kshakir:


kshakir:
We received a workaround from Travis support and I was able to rotate the Vault access token.

Follow details over at https://support.travis-ci.com/hc/en-us/requests/3231 but the TL;DR is to use the travis rest-api **V3**, not the travis web-ui nor the travis cli.

kshakir:
Like https://github.com/broadinstitute/cromwell/pull/4583, I've manually confirmed that various Travis builds are already reporting vault auth success and rendering the secrets.

As this diff otherwise shouldn't affect the build, merging early to so that other PRs don't have to do as much rebasing later.

PR #4588, opened 2019-01-25
Job rate control for AWS CI

mcovarr:
I confirmed this does throttle the token dispenser from the log output, though it may not get the AWS build passing until everyone rebases.


ruchim:
smaht 👍

Issue #4608, opened 2019-02-01
Add minCpuPlatform runtime setting for PAPI V2 workflows


pshapiro4broad:
When testing our workflow, we are seeing variations in some calculations based on the cpu architecture chosen for a VM. In PAPI V1, we worked around this issue by restricting the test to only run in a single region. This caused the same cpu to be used for all of the affected tests, and we were able to easily validate the workflow output.

Now that we are testing the workflow under PAPI V2, we are running into the same issue, and setting the region no longer fixes the issue. It looks like the `minCpuPlatform` pipelines API setting would be sufficient for this. See [Specifying a Minimum CPU Platform for VM Instances](https://cloud.google.com/compute/docs/instances/specify-min-cpu-platform).

Does this sound like a reasonable feature to add to cromwell's PAPI V2 support?

PR #4609, opened 2019-02-01
Proposed addition of cpuPlatform for PAPIv2 VM creation

pshapiro4broad:
This is my attempt at addressing issue #4608 

- add `cpuPlatform` as a new runtime attribute
- contents are not validated; they're passed on directly to PAPI V2

### DIFF
--- a/supportedBackends/google/pipelines/common/src/test/scala/cromwell/backend/google/pipelines/common/PipelinesApiRuntimeAttributesSpec.scala
+++ b/supportedBackends/google/pipelines/common/src/test/scala/cromwell/backend/google/pipelines/common/PipelinesApiRuntimeAttributesSpec.scala
@@ -25,7 +25,7 @@ class PipelinesApiRuntimeAttributesSpec extends WordSpecLike with Matchers with 
     )))
   }
 
-  val expectedDefaults = new PipelinesApiRuntimeAttributes(refineMV(1), None, Vector("us-central1-b", "us-central1-a"), 0, 10,
+  val expectedDefaults = new PipelinesApiRuntimeAttributes(refineMV(1), None, None, Vector("us-central1-b", "us-central1-a"), 0, 10,
     MemorySize(2, MemoryUnit.GB), Vector(PipelinesApiWorkingDisk(DiskType.SSD, 10)), "ubuntu:latest", false,
     ContinueOnReturnCodeSet(Set(0)), false)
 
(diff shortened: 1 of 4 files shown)


pshapiro4broad:
> looks good to me. why the do not merge label?

Because I tested it yet with PAPI v2, and am not sure how to do so. But if a cromwell expert is confident it looks good, that's OK with me.

danbills:
@pshapiro4broad you flatter me.  But yeah we should probably come up w/ a centaur test that makes sure it is actually working as intended.

The good news is that our travis PR tests conduct a whole battery of papi v2 tests and those pass so this didn't break anything :smile:

mcovarr:
The following could be used as the command body of a Centaur WDL to test this feature (copy-paste-edited from Thibault's [`docker_size_gcr.wdl`](https://github.com/broadinstitute/cromwell/blob/develop/centaur/src/main/resources/standardTestCases/docker_size/docker_size_gcr.wdl)):
```
apt-get install --assume-yes jq > /dev/null
NAME=`curl -s -H "Metadata-Flavor: Google" http://metadata.google.internal/computeMetadata/v1/instance/name`
ZONE=`basename \`curl -s -H "Metadata-Flavor: Google" http://metadata.google.internal/computeMetadata/v1/instance/zone\``
PROJECT=`curl -s -H "Metadata-Flavor: Google" http://metadata.google.internal/computeMetadata/v1/project/project-id`
curl -s -H "Authorization: Bearer `gcloud auth print-access-token`" "https://www.googleapis.com/compute/v1/projects/$PROJECT/zones/$ZONE/instances/$NAME?fields=cpuPlatform" | jq -r '.cpuPlatform'
```
Run on one of Jeff's legion GCE VMs this produces

```
Intel Haswell
```

supportedBackends/google/pipelines/common/src/test/scala/cromwell/backend/google/pipelines/common/PipelinesApiRuntimeAttributesSpec.scala
```diff
@@ -279,6 +279,18 @@ class PipelinesApiRuntimeAttributesSpec extends WordSpecLike with Matchers with
...
 
+  "should parse cpuPlatform correctly" in {
+    val runtimeAttributes = Map("docker" -> WomString("ubuntu:latest"))
+    val workflowOptionsJson =
+      """{
+        |  "default_runtime_attributes": { "cpuPlatform": "the platform" }
+        |}
+      """.stripMargin.parseJson.asInstanceOf[JsObject]
+    val workflowOptions = WorkflowOptions.fromJsonObject(workflowOptionsJson).get
+    val expectedRuntimeAttributes = expectedDefaults.copy(cpuPlatform = Some("the platform"))
```
mcovarr: We generally `Option` rather than `Some` to construct option instances and handle nulls in the expected way (generating `None`s rather than `Some(null)`s. Obviously for a literal that's not a concern but probably a good habit anyway. 🙂

cjllanwarne:
@mcovarr is there a reason why your test WDL is installing `jq` in the command rather than picking a docker image which already has it installed?

mcovarr:
@cjllanwarne uh because "my" WDL is 99.9% copy-pasted from Thibault's and that's what he did? 😛 I think the WDL could do what you're suggesting assuming the image also contains `curl` and `basename`.

ruchim:
@pshapiro4broad just want to be sure you'll be adding the Centaur tests?

pshapiro4broad:
> @pshapiro4broad just want to be sure you'll be adding the Centaur tests?

It turned out that we (green) didn't need this feature after all, and I've been a bit busy to run the required test. Maybe a red person could take this over, as I think it may be generally useful for other users?

cjllanwarne:
@pshapiro4broad would you be able to provide a WDL which could, in theory, test that this is working? Maybe with a short "and then you'd want to check this value to make sure"piece of commentary?

 If so, I bet we could wire the test into our testing framework on your behalf and merge this

mcovarr:
I've started working on this, the bit I pasted above a while back should hopefully be a good basis for the Centaur test.

mcovarr:
With a rebase, cherry pick of commit 82eb41502e208e590d2a6ef65adf29e30b119d3e from [this branch](https://github.com/broadinstitute/cromwell/commits/mlc_ps-set-gcs-min-cpu) and two thumbs hopefully we can get this merged.

mcovarr:
Oh we should probably document and CHANGELOG this too, I'll do that tomorrow.

mcovarr:
Closing in favor of #4704

Issue #4561, opened 2019-01-17
Query includeSubworkflow=false Speed


cjllanwarne:

danbills:
@cjllanwarne can you elaborate on the technical findings you and @mcovarr have discussed?  I.e. the procedure Cromwell is undertaking and why it is slow?  

Would like @kcibul to weigh in on our slow MySQL queries when we get to that level of detail.

mcovarr:
#4598 is related (not great performance in JMUI use cases) but not that same.

mcovarr:
Can you describe in more detail the conditions for this slow query? I'm curious if the changes in #4610 might have improved the situation here.

PR #4610, opened 2019-02-02
Labels query performance improvement

mcovarr:
One representative pair of timings on a stone cold prod clone (restarted between queries) for a single submission ID label with two ORed collection labels:

develop: 31.20 sec
this branch: 0.08 sec

No index changes required. MySQL will use a k/v index on `CUSTOM_LABEL_ENTRY` if it's available but its presence doesn't seem to actually make much difference in execution time.

Exclude labels have been folded into the new system, I still need to measure the performance impact of those changes.

### DIFF
--- a/database/sql/src/main/scala/cromwell/database/slick/tables/WorkflowMetadataSummaryEntryComponent.scala
+++ b/database/sql/src/main/scala/cromwell/database/slick/tables/WorkflowMetadataSummaryEntryComponent.scala
@@ -2,7 +2,7 @@ package cromwell.database.slick.tables
 
 import java.sql.Timestamp
 
-import cromwell.database.sql.tables.WorkflowMetadataSummaryEntry
+import cromwell.database.sql.tables.{CustomLabelEntry, WorkflowMetadataSummaryEntry}
 
 trait WorkflowMetadataSummaryEntryComponent {
 
@@ -60,10 +60,6 @@ trait WorkflowMetadataSummaryEntryComponent {
                                            workflowStatuses: Set[String],
                                            workflowNames: Set[String],
                                            workflowExecutionUuids: Set[String],
-                                           labelAndKeyValues: Set[(String, String)],
-                                           labelOrKeyValues: Set[(String, String)],
-                                           excludeLabelAndValues: Set[(String,String)],
-                                           excludeLabelOrValues: Set[(String,String)],
                                            submissionTimestampOption: Option[Timestamp],
                                            startTimestampOption: Option[Timestamp],
                                            endTimestampOption: Option[Timestamp],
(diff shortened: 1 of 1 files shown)


database/sql/src/main/scala/cromwell/database/slick/tables/WorkflowMetadataSummaryEntryComponent.scala
```diff
@@ -123,20 +111,79 @@ trait WorkflowMetadataSummaryEntryComponent {
...
+      endTimestampOption,
+      includeSubworkflows
+    )
+
+    // First do the summary table filtering, then some performance optimizations below for labels.
+    val baseQuery = workflowMetadataSummaryEntries.filter(filter)
+
+    // Label ANDs and ORs can be expressed as `join`s which at least on MySQL appear to execute more efficiently than
+    // equivalent `exists` formulations.
+    type LabelsQuery = Query[CustomLabelEntries, CustomLabelEntry, Seq]
```
cjllanwarne: Why `Seq` here and `Set` below?

danbills:
:+1: but curious what excludes perf looks like.

database/sql/src/main/scala/cromwell/database/slick/tables/WorkflowMetadataSummaryEntryComponent.scala
mcovarr: The label kv inputs are `Set`s and the `Query`s collections parallel those. The `Seq`s come from the Slick API.

database/sql/src/main/scala/cromwell/database/slick/tables/WorkflowMetadataSummaryEntryComponent.scala
```diff
@@ -94,22 +90,14 @@ trait WorkflowMetadataSummaryEntryComponent {
       val workflowStatusFilter = workflowStatuses.
         map(workflowStatus => workflowMetadataSummaryEntry.workflowStatus.fold(ifEmpty = exclude)(_ === workflowStatus)).
         reduceLeftOption(_ || _)
-      val labelsAndFilter = existsWorkflowLabels(workflowMetadataSummaryEntry, labelAndKeyValues, _ && _)
-      val labelsOrFilter = existsWorkflowLabels(workflowMetadataSummaryEntry, labelOrKeyValues, _ || _)
-      val excludeLabelsAndFilter = existsWorkflowLabels(workflowMetadataSummaryEntry, excludeLabelAndValues, _ && _).map(v => !v)
-      val excludeLabelsOrFilter = existsWorkflowLabels(workflowMetadataSummaryEntry, excludeLabelOrValues, _ || _).map(v => !v)
       val notASubworkflowFilter: Option[Rep[Boolean]] =
-        if (includeSubworkflows) None else Some(!metadataEntryExistsForWorkflowExecutionUuid(workflowMetadataSummaryEntry.workflowExecutionUuid, parentIdWorkflowMetadataKey))
+        if (includeSubworkflows) None else Option(!metadataEntryExistsForWorkflowExecutionUuid(workflowMetadataSummaryEntry.workflowExecutionUuid, parentIdWorkflowMetadataKey))
```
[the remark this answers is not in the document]
mcovarr: For the separate #4561 issue, yes

PR #4615, opened 2019-02-06
Adjust CromIAM timeout to match Cromwell

aednichols:
Searched the codebase for `request-timeout`, found that we seem to use 40 seconds not the 55 previously discussed.

Copied the config stanza from `cromwell/server/src/main/resources/application.conf` to CromIAM.

### DIFF
--- a/CromIAM/src/main/resources/application.conf
+++ b/CromIAM/src/main/resources/application.conf
@@ -34,4 +34,12 @@ swagger_oauth {
 akka {
   log-dead-letters = "off"
   loggers = ["akka.event.slf4j.Slf4jLogger"]
+
+  http {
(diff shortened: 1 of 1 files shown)


danbills:
Thanks! 

https://github.com/broadinstitute/firecloud-develop/blob/18ebf8ff504bd21618515e1d748d2d0f5fd38e32/base-configs/cromwell/cromwell.conf.ctmpl#L36'

This should similarly go into `firecloud-develop`

danbills:
https://github.com/broadinstitute/firecloud-develop/pull/1532

danbills:
Sorry for premature closing.  I do think this should be in default config

PR #4617, opened 2019-02-07
Workflow validation for labels PATCH goes to summary not metadata.

mcovarr:
To prevent unfortunate things from happening to our beloved summarizer.

### DIFF
--- a/database/sql/src/main/scala/cromwell/database/slick/MetadataSlickDatabase.scala
+++ b/database/sql/src/main/scala/cromwell/database/slick/MetadataSlickDatabase.scala
@@ -43,6 +43,11 @@ class MetadataSlickDatabase(originalDatabaseConfig: Config)
     runTransaction(action)
   }
 
+  override def metadataSummaryEntryExists(workflowExecutionUuid: String)(implicit ec: ExecutionContext): Future[Boolean] = {
+    val action = dataAccess.workflowMetadataSummaryEntryExistsForWorkflowExecutionUuid(workflowExecutionUuid).result
+    runTransaction(action)
+  }
+
   override def queryMetadataEntries(workflowExecutionUuid: String)
                                    (implicit ec: ExecutionContext): Future[Seq[MetadataEntry]] = {
     val action = dataAccess.metadataEntriesForWorkflowExecutionUuid(workflowExecutionUuid).result
--- a/database/sql/src/main/scala/cromwell/database/slick/tables/WorkflowMetadataSummaryEntryComponent.scala
+++ b/database/sql/src/main/scala/cromwell/database/slick/tables/WorkflowMetadataSummaryEntryComponent.scala
@@ -49,6 +49,13 @@ trait WorkflowMetadataSummaryEntryComponent {
       if workflowMetadataSummaryEntry.workflowExecutionUuid === workflowExecutionUuid
     } yield workflowMetadataSummaryEntry)
 
+  val workflowMetadataSummaryEntryExistsForWorkflowExecutionUuid = Compiled(
+    (workflowExecutionUuid: Rep[String]) => (for {
+      summaryEntry <- workflowMetadataSummaryEntries
+      if summaryEntry.workflowExecutionUuid === workflowExecutionUuid
+    } yield summaryEntry).exists
+  )
+
   val workflowStatusesForWorkflowExecutionUuid = Compiled(
     (workflowExecutionUuid: Rep[String]) => for {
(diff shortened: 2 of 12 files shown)


database/sql/src/main/scala/cromwell/database/slick/tables/WorkflowMetadataSummaryEntryComponent.scala
```diff
@@ -49,6 +49,13 @@ trait WorkflowMetadataSummaryEntryComponent {
       if workflowMetadataSummaryEntry.workflowExecutionUuid === workflowExecutionUuid
     } yield workflowMetadataSummaryEntry)
 
+  val workflowMetadataSummaryEntryExistsForWorkflowExecutionUuid = Compiled(
+    (workflowExecutionUuid: Rep[String]) => (for {
+      summaryEntry <- workflowMetadataSummaryEntries
+      if summaryEntry.workflowExecutionUuid === workflowExecutionUuid
+    } yield summaryEntry).exists
```
aednichols: Doesn't `exists` potentially generate quite slow SQL? A count might be faster.

(Or, we might not care about performance here, which is perfectly valid.)

database/sql/src/main/scala/cromwell/database/slick/MetadataSlickDatabase.scala
```diff
@@ -43,6 +43,11 @@ class MetadataSlickDatabase(originalDatabaseConfig: Config)
     runTransaction(action)
   }
 
+  override def metadataSummaryEntryExists(workflowExecutionUuid: String)(implicit ec: ExecutionContext): Future[Boolean] = {
+    val action = dataAccess.workflowMetadataSummaryEntryExistsForWorkflowExecutionUuid(workflowExecutionUuid).result
```
cjllanwarne: I think you could drop the `ForWorkflowExecutionUuid` since that’s already the name of the parameter

database/sql/src/main/scala/cromwell/database/slick/tables/WorkflowMetadataSummaryEntryComponent.scala
cjllanwarne: Why is `.exists` returning a `Rep[String]`?

database/sql/src/main/scala/cromwell/database/slick/tables/WorkflowMetadataSummaryEntryComponent.scala
```diff
@@ -49,6 +49,13 @@ trait WorkflowMetadataSummaryEntryComponent {
       if workflowMetadataSummaryEntry.workflowExecutionUuid === workflowExecutionUuid
     } yield workflowMetadataSummaryEntry)
 
+  val workflowMetadataSummaryEntryExistsForWorkflowExecutionUuid = Compiled(
```
cjllanwarne: This looks like it's just `val workflowMetadataSummaryEntryExistsForWorkflowExecutionUuid = workflowMetadataSummaryEntriesForWorkflowExecutionUuid.exists`

services/src/main/scala/cromwell/services/metadata/impl/MetadataDatabaseAccess.scala
```diff
@@ -167,10 +167,14 @@ trait MetadataDatabaseAccess {
     metadataDatabaseInterface.getWorkflowLabels(id.toString)
   }
 
-  def workflowExistsWithId(possibleWorkflowId: String)(implicit ec: ExecutionContext): Future[Boolean] = {
+  def workflowWithIdExistsInMetadata(possibleWorkflowId: String)(implicit ec: ExecutionContext): Future[Boolean] = {
```
danbills: ToL Remove this and `workflowWithIdExistsInMetadataSummaries` as they don't encapsulate any logic.

database/sql/src/main/scala/cromwell/database/slick/tables/WorkflowMetadataSummaryEntryComponent.scala
mcovarr: It's not. Why do you think `.exists` is returning a `Rep[String]`?

mcovarr:
This is intended to be a very narrow change to move the "does this workflow exist" check from metadata to metadata summaries so the release can go forward. This is deliberately intended to follow existing naming conventions, coding patterns and API structure as much as possible. If anyone feels strongly that these things should be improved there can certainly be enhancement requests.

Issue #4626, opened 2019-02-11
hello world runs without error on cromwell 36. on cromwell 37 if tinishes but throws several exceptions


wresch:
Running on a Local backend with `java -jar $CROMWELL_JAR run -i input.json wf.wdl`

```
task hello {
    String outfilename
    String ? name

    command {
        echo "Hello ${default='world' name}" > ${outfilename}
    }
    output {
        File out = "${outfilename}"
    }
}

workflow test1 {
    String name

    call hello {
        input: outfilename="${name}.txt", name = "${name}"
    }
}
```

Using default configuration.

Output:
```
[2020-02-11 10:13:03,33] [info] Running with database db.url = jdbc:hsqldb:mem:89c116e0-5bca-4467-aaff-ae492c2ebbaf;shutdown=false;hsqldb.tx=mvcc
[2019-02-11 10:13:14,71] [info] Running migration RenameWorkflowOptionsInMetadata with a read batch size of 100000 and a write batch size of 100000
[2019-02-11 10:13:14,75] [info] [RenameWorkflowOptionsInMetadata] 100%
[2019-02-11 10:13:15,05] [info] Running with database db.url = jdbc:hsqldb:mem:6b5d8035-4932-4680-b912-34885765f705;shutdown=false;hsqldb.tx=mvcc
[2019-02-11 10:13:15,63] [info] Slf4jLogger started
[2019-02-11 10:13:16,02] [info] Workflow heartbeat configuration:
{
  "cromwellId" : "cromid-1ddecb5",
  "heartbeatInterval" : "2 minutes",
  "ttl" : "10 minutes",
  "writeBatchSize" : 10000,
  "writeThreshold" : 10000
}
[2019-02-11 10:13:16,08] [info] Metadata summary refreshing every 2 seconds.
[2019-02-11 10:13:16,20] [info] KvWriteActor configured to flush with batch size 200 and process rate 5 seconds.
[2019-02-11 10:13:16,23] [info] WriteMetadataActor configured to flush with batch size 200 and process rate 5 seconds.
[2019-02-11 10:13:16,25] [warn] 'docker.hash-lookup.gcr-api-queries-per-100-seconds' is being deprecated, use 'docker.hash-lookup.gcr.throttle' instead (see reference.conf)
[2019-02-11 10:13:16,26] [info] CallCacheWriteActor configured to flush with batch size 100 and process rate 3 seconds.
[...]

PR #4632, opened 2019-02-11
noop health check by default

mcovarr:
For discussion. The current default `StandardHealthMonitorServiceActor` checks both Docker and the Engine database. Depending on deployment the first check may be completely unnecessary or undesirable (see #4626) and the second check is only really useful if there are queries to Cromwell's status endpoint to read the status.

If the basic idea is accepted there would need to be additional documentation announcing the change in default and how to restore the previous default behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,13 @@ The `ConfigBackendLifecycleActorFactory` path variables `script`, `out` and `err
 and without docker. Similarly, when killing a docker task the `kill-docker` configuration key is now used instead of
 `kill`. For more information see the [online documentation](https://cromwell.readthedocs.io/en/stable/backends/SGE/).
 
+### No-op Health Monitor is now the default health monitor
+
+Previous versions of Cromwell defaulted to using a health monitor service that checked Docker Hub and engine database status.
+Neither check was useful if the `status` endpoint was never consulted as is likely the case in most deployments. Cromwell 38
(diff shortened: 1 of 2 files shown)


cjllanwarne:
I'd vote 👍 on using noop as the default out-of-the-box option so long as:
- it's documented as a config change
- we do something sensible in centaur
- we make sure the specs are still working as expected

danbills:
:shipit: :+1: 

Noting that a sister fc-develop PR is now needed

danbills:
@ruchim might opine?

mcovarr:
The Travis builds for this are a total trash fire, investigating.

mcovarr:
@cjllanwarne [rebasing this PR](https://github.com/broadinstitute/cromwell/pull/4631/files) so the `NoopHealthMonitorServiceActor` actually became usable.

mcovarr:
@danbills no `fc-develop` PR required, FC is already specifying their own health check impl.

mcovarr:
@danbills @cjllanwarne this requires two approving reviews. I think it may have been created back in the time of pullapprove when thumbs counted as approvals, but those days are no more.

cjllanwarne:
Do I need to do more here? I think you're just waiting for @danbills ?

screen shot 2019-03-05 at 2 38 00 pm

mcovarr:
@cjllanwarne you're good, thanks 🙂 The wording on the GitHub message made it sound like two reviews were missing but it was actually only one.

PR #4633, opened 2019-02-11
Add note about bucket permissions in PAPI v2

ruchim:


CHANGELOG.md
```diff
@@ -235,6 +235,12 @@ In addition, the following changes are to be expected:
 * Error messages for failed jobs might differ from V1
 * The Pipelines API log file content might differ from V1
 
+**Permissions***:
```
cjllanwarne: I don't like the idea of this being "documented" by adding it onto in a months (years?) old entry in the CHANGELOG. IMO this belongs in user-facing documentation which the CHANGELOG could link out to if it chose to.

... TL;DR: Can we put this in the readthedocs PAPIv2 page?

cjllanwarne: @ruchim this doc is talking about PAPIv2, and seems like an appropriate place for this content: https://cromwell.readthedocs.io/en/develop/backends/Google/

Then the CHANGELOG could link out to that page (or even the `.MD` file in the `docs/` directory which feeds it?)

gemmalam:
@ruchim could you take a look at this?

ruchim:
@cjllanwarne Just need to update the piece around private dockerhub support --otherwise approved 👍

PR #4635, opened 2019-02-11
Create documentation for Cromwell and containers

illusional:
Following on from the conversation at #4039, I've started a "Getting started with Containers" on the Cromwell docs. My thoughts are it might be a good place to put thoughts about containers, how they can be used and any caveats about them.

This tutorial follows the [user goal](https://github.com/broadinstitute/cromwell/issues/2177#issuecomment-330341279) (from #2177):
> As a **user with images in Singularity**, I want **Cromwell to support using Singularity images (either via Singularity Hub and the command line, or connecting via API)**, so that I can **use Singularity images and not have to duplicate them in Docker**.

However, without any code changes, just purely through configuration. 

I think it's worth touching on using these container technologies with job schedulers, so I've included the ConfigurationFile and the HPCIntro tutorials as prerequisites.

### DIFF
--- a/docs/tutorials/Containers.md
+++ b/docs/tutorials/Containers.md
@@ -356,6 +356,9 @@ With a job queue like SLURM, you just need to wrap this script in an `sbatch` su
 
 ```
 submit-docker = """
+    # Pull the image using the head node, in case our workers don't have network access
+    udocker pull ${docker}
+    
     sbatch \
       -J ${job_name} \
       -D ${cwd} \
--- a/docs/tutorials/Containers.md
+++ b/docs/tutorials/Containers.md
@@ -75,13 +75,13 @@ ___
 
 #### Docker on a Local Backend
 
-On a single machine (laptop or server), no extra configuration is needed to allow docker to run provided Docker is installed.
+On a single machine (laptop or server), no extra configuration is needed to allow docker to run, provided Docker is installed.
 
 You can install Docker for Linux, Mac or Windows from [Docker Hub](https://hub.docker.com/search/?type=edition&offering=community)
 
 #### Docker on Cloud
 
-It is strongly advised that you provider Docker tags to tasks that will run on Cloud backends, and in fact most Cloud providers require it.
+It is strongly advised that you provide Docker tags to tasks that will run on Cloud backends, and in fact most Cloud providers require it.
 
 It might be possible to use an alternative container engine, but this is not recommended if Docker is supported.
 
@@ -434,7 +434,7 @@ Within the two configurations above:
 
 By enabling Cromwell's run-in-background mode, you remove the necessity for the `kill`, `check-alive` and `job-id-regex` blocks, which disables some safety checks when running workflows:
 
-- If there is an error starting the container or executing the script, Cromwell may not recognise this error and hang.
+- If there is an error starting the container or executing the script, Cromwell may not recognise this error and hang. For example, this may occur if the container attempts to exceed its allocated resources (runs out of memory); the container daemon may terminate the container without completing the script.
 - If you abort the workflow (by attempting to close Cromwell or issuing an abort command), Cromwell does not have a reference to the job and will not be able to execute the container.
 
 This is only necessary in local environments where there is no job manager to control this, however if your container technology can emit an identifier to stdout, then you are able to remove the run-in-background flag. Note that the check-alive block is not called to check a job-status, rather Cromwell looks for the presence of an `rc` file (see the previous section for more information).
--- a/docs/tutorials/Containers.md
+++ b/docs/tutorials/Containers.md
@@ -103,7 +103,7 @@ You can install Docker for Linux, Mac or Windows from [Docker Hub](https://hub.d
 
 #### Docker on Cloud
 
-It is strongly advised that you provide Docker tags to tasks that will run on Cloud backends, and in fact most Cloud providers require it.
+It is strongly advised that you provide a Docker image to tasks that will run on Cloud backends, and in fact most Cloud providers require it.
 
 It might be possible to use an alternative container engine, but this is not recommended if Docker is supported.
 
@@ -464,7 +464,6 @@ By enabling Cromwell's run-in-background mode, you remove the necessity for the 
 - If you abort the workflow (by attempting to close Cromwell or issuing an abort command), Cromwell does not have a reference to the job and will not be able to execute the container.
 
 This is only necessary in local environments where there is no job manager to control this, however if your container technology can emit an identifier to stdout, then you are able to remove the run-in-background flag. Note that the check-alive block is not called to check a job-status, rather Cromwell looks for the presence of an `rc` file (see the previous section for more information).
-
 
 ### Next Steps
 
--- a/docs/tutorials/Containers.md
+++ b/docs/tutorials/Containers.md
@@ -1,7 +1,7 @@
 ## Containers
 
 Containers are self-contained software archives, that hold everything from the tool you want to run, to libraries and runtimes it requires, and the operating system it runs on.
- 
+
 Best Practise WDL and CWL define containers for their tasks to run in, to ensure reproducibility and portability - that running the same task on a different system will run the exact same software. 
 
 Docker images are the most common container format, but it is not advisable for certain systems to run Docker itself, and for this reason Cromwell can be configured to support a number of alternatives.
@@ -23,6 +23,7 @@ Docker images are the most common container format, but it is not advisable for 
 * [udocker](#udocker)
     * [Installation](#installation-1)
     * [Configuration](#configuration)
+    * [Caching](#caching)
 * [Configuration in Detail](#configuration-in-detail)
     * [Enforcing container requirements](#enforcing-container-requirements)
     * [Docker Digests](#docker-digests)
@@ -64,7 +65,6 @@ task hello_world {
         docker: 'ubuntu:latest'
     }
 }
-
 workflow hello {
     call hello_world
 }
@@ -84,7 +84,6 @@ inputs:
           prefix: "Hello, "
 outputs:
     out: stdout
-
 requirements:
     DockerRequirement:
         dockerPull: "ubuntu:latest"
@@ -146,14 +145,12 @@ As the Singularity container does not emit a job-id, we must include the `run-in
 Putting this together, we have an example base configuration for a local environment:
 ```hocon
 include required(classpath("application"))
-
 backend {
     default: singularity
     providers: {
         singularity {
             # The backend custom configuration.
             actor-factory = "cromwell.backend.impl.sfs.config.ConfigBackendLifecycleActorFactory"
-
             config {
                 run-in-background = true
                 runtime-attributes = """
@@ -215,7 +212,6 @@ Putting this all together, a complete SLURM + Singularity config might look like
 ```
 backend {
   default = slurm
-
   providers {
     slurm {
       actor-factory = "cromwell.backend.impl.sfs.config.ConfigBackendLifecycleActorFactory"                                                                                     
@@ -226,7 +222,6 @@ backend {
         Int requested_memory_mb_per_core = 8000
         String? docker
         """
-
         submit = """
             sbatch \
               --wait \
@@ -239,7 +234,6 @@ backend {
               --mem-per-cpu=${requested_memory_mb_per_core} \
               --wrap "/bin/bash ${script}"
         """
-
         submit-docker = """
             # Ensure singularity is loaded if it's installed as a module
             module load Singularity/3.0.1
@@ -247,7 +241,6 @@ backend {
             # Build the Docker image into a singularity image
             IMAGE=${cwd}/${docker}.sif
             singularity build $IMAGE docker://${docker}
-
             # Submit the script to SLURM
             sbatch \
               --wait \
@@ -260,7 +253,6 @@ backend {
               --mem-per-cpu=${requested_memory_mb_per_core} \
               --wrap "singularity exec --bind ${cwd}:${docker_cwd} $IMAGE ${job_shell} ${script}"
         """
-
         kill = "scancel ${job_id}"
         check-alive = "squeue -j ${job_id}"
         job-id-regex = "Submitted batch job (\\d+).*"
@@ -277,12 +269,10 @@ In addition, if you or your sysadmins were not able to give `setuid` permissions
 ```
 submit-docker = """
     [...]
-
     # Build the Docker image into a singularity image
     # We don't add the .sif file extension because sandbox images are directories, not files
     IMAGE=${cwd}/${docker}
     singularity build --sandbox $IMAGE docker://${docker}
-
     # Now submit the job
     # Note the use of --userns here
     sbatch \
@@ -409,13 +399,10 @@ docker {
   hash-lookup {
     # Set this to match your available quota against the Google Container Engine API
     #gcr-api-queries-per-100-seconds = 1000
-
     # Time in minutes before an entry expires from the docker hashes cache and needs to be fetched again
     #cache-entry-ttl = "20 minutes"
-
     # Maximum number of elements to be kept in the cache. If the limit is reached, old elements will be removed from the cache
     #cache-size = 200
-
     # How should docker hashes be looked up. Possible values are "local" and "remote"
     # "local": Lookup hashes on the local docker daemon using the cli
     # "remote": Lookup hashes on docker hub and gcr
@@ -445,7 +432,6 @@ runtime {
 ```
 
 You can find the `sha256` of an image using `docker images --digests`
- 
  
 ### Notes
 
--- a/docs/tutorials/Containers.md
+++ b/docs/tutorials/Containers.md
@@ -1,7 +1,7 @@
 ## Containers
 
 Containers are self-contained software archives, that hold everything from the tool you want to run, to libraries and runtimes it requires, and the operating system it runs on.
-
+ 
 Best Practise WDL and CWL define containers for their tasks to run in, to ensure reproducibility and portability - that running the same task on a different system will run the exact same software. 
 
 Docker images are the most common container format, but it is not advisable for certain systems to run Docker itself, and for this reason Cromwell can be configured to support a number of alternatives.
@@ -65,6 +65,7 @@ task hello_world {
         docker: 'ubuntu:latest'
     }
 }
+
 workflow hello {
     call hello_world
 }
@@ -84,6 +85,7 @@ inputs:
           prefix: "Hello, "
 outputs:
     out: stdout
+
 requirements:
     DockerRequirement:
         dockerPull: "ubuntu:latest"
@@ -145,12 +147,14 @@ As the Singularity container does not emit a job-id, we must include the `run-in
 Putting this together, we have an example base configuration for a local environment:
 ```hocon
 include required(classpath("application"))
+
 backend {
     default: singularity
     providers: {
         singularity {
             # The backend custom configuration.
             actor-factory = "cromwell.backend.impl.sfs.config.ConfigBackendLifecycleActorFactory"
+
             config {
                 run-in-background = true
                 runtime-attributes = """
@@ -212,6 +216,7 @@ Putting this all together, a complete SLURM + Singularity config might look like
 ```
 backend {
   default = slurm
+
   providers {
     slurm {
       actor-factory = "cromwell.backend.impl.sfs.config.ConfigBackendLifecycleActorFactory"                                                                                     
@@ -222,6 +227,7 @@ backend {
         Int requested_memory_mb_per_core = 8000
         String? docker
         """
+
         submit = """
             sbatch \
               --wait \
@@ -234,6 +240,7 @@ backend {
               --mem-per-cpu=${requested_memory_mb_per_core} \
               --wrap "/bin/bash ${script}"
         """
+
         submit-docker = """
             # Ensure singularity is loaded if it's installed as a module
             module load Singularity/3.0.1
@@ -241,6 +248,7 @@ backend {
             # Build the Docker image into a singularity image
             IMAGE=${cwd}/${docker}.sif
             singularity build $IMAGE docker://${docker}
+
             # Submit the script to SLURM
             sbatch \
               --wait \
@@ -253,6 +261,7 @@ backend {
               --mem-per-cpu=${requested_memory_mb_per_core} \
               --wrap "singularity exec --bind ${cwd}:${docker_cwd} $IMAGE ${job_shell} ${script}"
         """
+
         kill = "scancel ${job_id}"
         check-alive = "squeue -j ${job_id}"
         job-id-regex = "Submitted batch job (\\d+).*"
@@ -269,10 +278,12 @@ In addition, if you or your sysadmins were not able to give `setuid` permissions
 ```
 submit-docker = """
     [...]
+
     # Build the Docker image into a singularity image
     # We don't add the .sif file extension because sandbox images are directories, not files
     IMAGE=${cwd}/${docker}
     singularity build --sandbox $IMAGE docker://${docker}
+
     # Now submit the job
     # Note the use of --userns here
     sbatch \
@@ -362,7 +373,7 @@ For instance, when you run a normal Docker image with `docker run image`, it wil
 However, by default Cromwell requests and runs images using their `sha` hash, rather than using tags.
 This strategy is actually preferable, because it ensures every execution of the task or workflow will use the exact same version of the image, but some engines such as `udocker` don't support this feature.
 
-If you are using `udocker` or want to enable the use of hash-based image references, you can set the following config option:
+If you are using `udocker` or want to disable the use of hash-based image references, you can set the following config option:
 ```
 docker.hash-lookup.enabled = false
 ```
@@ -399,10 +410,13 @@ docker {
   hash-lookup {
     # Set this to match your available quota against the Google Container Engine API
     #gcr-api-queries-per-100-seconds = 1000
+
     # Time in minutes before an entry expires from the docker hashes cache and needs to be fetched again
     #cache-entry-ttl = "20 minutes"
+
     # Maximum number of elements to be kept in the cache. If the limit is reached, old elements will be removed from the cache
     #cache-size = 200
+
     # How should docker hashes be looked up. Possible values are "local" and "remote"
     # "local": Lookup hashes on the local docker daemon using the cli
     # "remote": Lookup hashes on docker hub and gcr
@@ -432,6 +446,7 @@ runtime {
 ```
 
 You can find the `sha256` of an image using `docker images --digests`
+ 
  
 ### Notes
 
--- a/docs/tutorials/Containers.md
+++ b/docs/tutorials/Containers.md
@@ -328,7 +328,7 @@ To configure `udocker` to work in a local environment, you must tag the provider
 ```
 run-in-background = true
 submit-docker = """
-    udocker run --rm -v ${cwd}:${docker_cwd} ${docker} ${script}
+    udocker run --rm -v ${cwd}:${docker_cwd} ${docker} ${job_shell} ${script}
 """
 ```
 
@@ -347,7 +347,7 @@ submit-docker = """
       -t ${runtime_minutes} \
       ${"-c " + cpus} \
       --mem-per-cpu=${requested_memory_mb_per_core} \
-      --wrap "udocker run --rm -v ${cwd}:${docker_cwd} ${docker} ${script}"
+      --wrap "udocker run --rm -v ${cwd}:${docker_cwd} ${docker} ${job_shell} ${script}"
 """
 ```
 
--- a/docs/tutorials/Containers.md
+++ b/docs/tutorials/Containers.md
@@ -370,7 +370,7 @@ submit-docker = """
 ```
 
 In addition, to run `udocker`, you'll have to make sure `hash-lookup` is disabled.
-Refer to [this section](#docker_digests) for more detail.
+Refer to [this section](#docker-digests) for more detail.
 
 ### Configuration in Detail
 The behaviour of Cromwell with containers can be modified using a few other options.
--- a/docs/tutorials/Containers.md
+++ b/docs/tutorials/Containers.md
@@ -1,15 +1,10 @@
 ## Containers
 
-Containers are technologies that describe an environment's software requirements. Broadly, there are two primary types of containers:
+Containers are self-contained software archives, that hold everything from the tool you want to run, to libraries and runtimes it requires, and the operating system it runs on.
+ 
+Best Practise WDL and CWL define containers for their tasks to run in, to ensure reproducibility and portability - that running the same task on a different system will run the exact same software. 
 
-* Images - Software requirements that describe a sandbox environment that you.
-* Runtime - Specifies a series of processes that run inside a container image.
-
-Containers aim to be content-agnostic, infrastructure-agnostic, designed for automation and promote reproducibility.
-
-There are organisations such as the [Open Container Initiative](https://www.opencontainers.org) that are aiming to standardise the interface that container technologies can implement.
-
-The motivation for this tutorial was driven by: "I want to use containers, but can't use Docker".
+Docker images are the most common container format, but is is not advisable for certain systems to run Docker itself, and for this reason Cromwell supports a number of alternatives
 
 ### Prerequisites
 This tutorial page relies on completing the previous tutorials:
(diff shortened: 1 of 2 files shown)


cjllanwarne:
Hi @illusional thank you so much for contributing to our docs - this is awesome!

Would you like to me make this branch a hosted readthedocs branch (so you can preview your changes live?)

Also - let us know when you're done with the "WIP" and we can do a review for you. In the meantime I'll add the "Back with Originator" label so that we know that you're still working on it.

Thanks again!

cjllanwarne:
Well, some bad news... I just tried to add the branch to readthedocs but they won't let us host a forked repo's branch. So, I guess you'll have to host your changes locally if you want to preview the changes - sorry!

illusional:
Hey @vsoch, in our submit script, we're opting to pull and build our docker images on the head node as @TMiguelT suggested. If the build exists, singularity will ask to overwrite the existing build. Just wondering if there's an elegant way to skip the build if it exists (and it's up to date) or whether we should be forcing a rebuild every time.

I could skip it like this, but it's not as friendly./
```
echo 'n' | singularity build --sandbox $IMAGE docker://${docker}
```

vsoch:
How are you building with a sandbox without sudo?

vsoch:
This would be a better use case for pull, which will check the cache first and not ask for the confirmation.

TMiguelT:
I'm confused. Sandbox mode doesn't/shouldn't require sudo. We included it because it requires the *least* permissions to use it. 

As for `pull`, it seems to force rebuild the image every time, exactly the same as `build --force`. What we want ideally is a way to build the image if it doesn't exist or needs to be updated, but if neither is the case, then do nothing.

vsoch:
Yes that's correct, sorry I didn't see you are building from a docker uri. 

I don't think there exists the exact functionality you want, but we are getting there. The current cache support is for docker layers, which will save you download time, but you still would rebuild from them each time. I think it would be worth the effort to ask for what you need. Take a look at the [latest release](https://github.com/sylabs/singularity/releases) that has some support for caching (for library images). Then I would open an issue and say something about it - you want the same caching but for docker pulled containers. This particular flow to pull (or build) and honor a cached image if it exists is very common and reasonable, and should be supported.

TMiguelT:
Thanks, @vsoch, I will look into the new caching functionality once 3.1.0 is fully released, and I'll open an issue soon. The reason this matters is because if we run a scatter job in Cromwell where we have multiple jobs running the same image, they will all simultaneously try to rebuild the image file, which I imagine will catastrophically fail. But now that I think about it, if the image does need updating this will still cause it to fail. I might have to think about some kind of locking mechanism to use here.

TMiguelT:
In the meantime, I think my solution is to build the image into each job's working directory, so they don't all try to write to the same file.

vsoch:
I don't think the pull/build should be built in to the command that is run at scale - the image should be pulled /built once, and then the direct path passed into the script / pipeline that is run at scale.

TMiguelT:
You're right, it shouldn't. But because Cromwell only gives us one single hook to customise running containers) the `submit-docker`, a script that has to be able to run at scale. Ideally Cromwell would have a `pull-docker` hook that's run each time a new image needs to be pulled or updated. But since that's not an option, I have to build the image separately for each job.

TMiguelT:
Okay, I've pushed some config changes so that each `singularity build` happens in its own working directory. And I've added `sbatch --wait`, because then if the job fails, Cromwell will know about it. I'm now happy for this to reviewed, I think.

docs/tutorials/Containers.md
```diff
@@ -65,30 +67,32 @@ requirements:
...
+___
 ### Docker
 
 [Docker](https://www.docker.com) is a popular container technology that is natively supported by Cromwell and WDL.
 
 
 #### Docker on a Local Backend
 
-On a single machine (laptop or server), no extra configuration is needed to allow docker to run, provided Docker is installed.
+On a single machine (laptop or server), no extra configuration is needed to allow docker to run provided Docker is installed.
```
TMiguelT: This sentence doesn't make sense without the comma

docs/tutorials/Containers.md
```diff
@@ -65,30 +67,32 @@ requirements:
...
 
 You can install Docker for Linux, Mac or Windows from [Docker Hub](https://hub.docker.com/search/?type=edition&offering=community)
 
 #### Docker on Cloud
 
-It is strongly advised that you use Docker images on Cloud backends, because most cloud backends require Docker images to run.
+It is strongly advised that you provider Docker tags to tasks that will run on Cloud backends, and in fact most Cloud providers require it.
```
TMiguelT: "provider" should be "provide". Also, This section is talking about technologies in a broad sense, I don't think tags need to be discussed here.

illusional:
Hey @TMiguelT, I made a few small changes to udocker, added the notes we discussed and a next steps section to follow the general template of the other tutorials. I wanted to add a small section about the caching of udocker images but don't know udocker well enough to really assert this:

> #### Caching in udocker
> udocker caches images within the install or user directory, thus reducing the need to pull and build the docker containers at every stage. Clarification is required on whether udocker will concurrently write to the same cache directory for largely scattered workflows.

So I've just left it out. I also think it might be worth saying more explicitly that Singularity is technically user-installable (just without `setuid`, as I didn't realise until our conversation.

If you're happy with what's there now, I'll remove the WIP and put it up for review again. If there's anyone out there reading, we'd love to get your feedback or clarification on any points.

docs/tutorials/Containers.md
```diff
@@ -293,12 +297,15 @@ In essence, udocker provides a command line interface that mimics `docker`, and
...
+udocker can be installed without any kind of root permissions. Refer to udocker's installation documentation [here](https://github.com/indigo-dc/udocker/blob/master/doc/installation_manual.md) for more information.
 
 #### Configuration
 
-`udocker` mimics the `docker` command-line interface, so all you need to do to configure it is set the `submit-docker` configuration to something like this:
+(As of [2019-02-18](https://github.com/indigo-dc/udocker/issues/112)) udocker does not support looking up docker container by digests, hence you'll have to make ensure `hash-lookup` is disabled. Refer to [this section](#docker-digests) for more detail.
+
+To configure `udocker` to work in a local environment, you must tag the provider's configuration to `run-in-background` and update the `submit-docker` to use udocker:
 ```
+run-in-background = true
```
TMiguelT: Good catch

docs/tutorials/Containers.md
```diff
@@ -413,4 +419,31 @@ runtime {
...
 
 You can find the `sha256` of an image using `docker images --digests`
- 
\ No newline at end of file
+ 
+ 
+### Notes
+
+#### How does Cromwell know when a job or container has completed?
+Cromwell uses the presence of the `rc` (returncode) file to determine whether a task has succeeded or failed. This `rc` file is generated as part of the `script` within the execution directory, where the script is assembled at runtime. This is important as if the script executes successfully but the container doesn't terminate, Cromwell will continue the execution of the workflow and the container will persist hogging system resources.
```
TMiguelT: Can you add a situation where the container doesn't terminate, as an example?

docs/tutorials/Containers.md
```diff
@@ -413,4 +419,31 @@ runtime {
...
+
+Within the two configurations above:
```
TMiguelT: You explain how these technologies run, but not how they relate to the generation of the rc file. Does `exec` running in the foreground mean that the container will exit properly?

illusional: Yeah, I looked for specific wording in the documentation for this but couldn't find it. I tested it on the cluster and it runs in the foreground and terminates when complete, so have sort of asserted this. I think the point is important, and it would be great to have some more solid backing from the docs.

TMiguelT:
So the gist of how udocker caches things is that it uses a directory similar to Docker, which defaults to `~/.udocker`, but you can override that with a config file [described here](https://github.com/indigo-dc/udocker/blob/master/doc/installation_manual.md#9-configuration), to set a custom location, e.g. `reposdir = "/path/to/cache"`. Within that directory it caches the different layers and images in different subdirectories. I'll write that up into the document.

TMiguelT:
@illusional, I've added Singularity installation docs, and udocker cache docs

illusional:
Hey @cjllanwarne, I think we're good to be reviewed! I've removed the WIP status from this merge request and read through a couple of times (hopefully I've caught everything). Is there a process I can follow to make it easy for you to review?

Thanks!
___

Sidenote for the future: It might be good to reference the [open container initiative](https://www.opencontainers.org) at some point. It's not important for the tutorial, but I think it'll be good once their ([runtime](
http://www.github.com/opencontainers/runtime-spec) | [image](
http://www.github.com/opencontainers/image-spec)) specs become more widely adopted.

vsoch:
Is there some issue with your commit history? We shouldn't be seeing all those commits from 2015, it should be a clean addition of those that you have added. Could you perhaps try cloning again, creating a new branch from master, and then recreating the changes? I've been in squibbles like this before, and usually when I do it again the extraneous commits go away.

It looks like it's primarily new files, so it shouldn't be too hard to do! I'd be happy to review after that.

illusional:
@vsoch My guess is there's commits in master that aren't in develop, because I accidentally merged from master and that's where those problems started. I also made a mistake by using my develop branch to make the changes, so I took my develop back to d6038cff64b0160559466f498a006687179b1d0d (current broad/cromwell) and cherry-picked all the commits back in. 

This should be all good now to merge.

vsoch:
Hmm, this still makes me very uncomfortable - if the PR is from your develop to develop, you shouldn't have those extra commits. You should not have merged master into your develop.

To fix this, you can either start clean from a develop (not merged with master) or if you've tried that and had trouble, I'll offer to help you and open a separate PR. It's messy merges like this with force pushes that lead to a lot of headaches, and it should be avoided if possible. It's important to do this right.

It's after 11pm here and I'm not in work mode, but I can help with this tomorrow.

TMiguelT:
Yes, the merge was not great, but he's gone back and effectively rebased it (via cherry picking). I don't think there should be any issues as it is

vsoch:
Okay, if you say so, it does looks like the extraneous commits are cleaned up! :D 

Comments coming soon!

docs/tutorials/Containers.md
```diff
@@ -0,0 +1,475 @@
+## Containers
+
+Containers are self-contained software archives, that hold everything from the tool you want to run, to libraries and runtimes it requires, and the operating system it runs on.
+
```
vsoch: It doesn't quite hit the nail to call it a self-contained software archive - it's really more of an isolated set of system processes, and another layer of abstraction. But for the purposes of understanding, I see what you are getting at. What about:

> Containers are encapsulated environments that include an operating system, libraries, and software. If you have a host machine running Centos, you can run an isolated container with Ubuntu 18.04. At a high level, it's useful to think of a container as a program or binary.

docs/tutorials/Containers.md
```diff
@@ -0,0 +1,475 @@
+## Containers
+
+Containers are self-contained software archives, that hold everything from the tool you want to run, to libraries and runtimes it requires, and the operating system it runs on.
+ 
+Best Practise WDL and CWL define containers for their tasks to run in, to ensure reproducibility and portability - that running the same task on a different system will run the exact same software. 
+
```
vsoch: > Best practice

I'm not sure I agree that you can state about "best practice" here - there are inordinate numbers of workflow managers, and sometimes a workflow manager, period, is overly complex for the need (using containers). I would remove the bias and just say:

> To ensure reproducibility and portability, WDL and CWL can define containers for their tasks to run in. Using containers means that you can be confident about running the same software across different systems.

docs/tutorials/Containers.md
```diff
@@ -0,0 +1,475 @@
...
+* Recommended: [Getting started on HPC clusters](HPCIntro.md)
+
+### Goals
+
+At the end of this tutorial, you'll become familiar with container technologies and how to configure Cromwell to use these independently, or with job schedulers.
+
+
+### Specifying Containers in your Workflow
+
+Containers are specified on a per-task level, this can be achieved in WDL by specifying a [`docker`](https://github.com/openwdl/wdl/blob/master/versions/1.0/SPEC.md#docker) tag in the `runtime` section. For example, specifying that the following WDL script should use the container `ubuntu:latest` can be achieved by:
```
vsoch: > level, *and* this can be achieved

docs/tutorials/Containers.md
```diff
@@ -0,0 +1,475 @@
...
+It is strongly advised that you provide a Docker image to tasks that will run on Cloud backends, and in fact most Cloud providers require it.
+
+It might be possible to use an alternative container engine, but this is not recommended if Docker is supported.
+
+#### Docker on HPC
+
+Docker can allow running users to gain superuser privileges, called the [Docker daemon attack surface](https://docs.docker.com/engine/security/security/#docker-daemon-attack-surface). In HPC and multi-user environments, Docker recommends that "only trusted users should be allowed to control your Docker Daemon".
+
+For this reason it's worth exploring other technologies that will support the reproducibility and simplicity of running a workflow that uses docker containers; Singularity or udocker.
+
```
vsoch: > Singularity or udocker, discussed next.

I think we should add that, because my first instinct was "Where is the link to learn more about this?" and I almost suggested adding it. But if the reader knows it's coming next, no need! It's coming :)

docs/tutorials/Containers.md
```diff
@@ -0,0 +1,475 @@
...
+___
+
+### Singularity
+
+Singularity is a container engine designed for use on HPC systems in particular, while ensuring an appropriate level of security that Docker cannot provide.
+
```
vsoch: I would just call it a "container technology." Container engine sounds funny :)

> Singularity is a container technology that was designed to provide an appropriate level of security for shared HPC resources.

docs/tutorials/Containers.md
```diff
@@ -0,0 +1,475 @@
...
+#### Installation
+Before you can configure Cromwell on your HPC system, you (or your sysadmin) will have to install Singularity, which is documented [here](https://www.sylabs.io/guides/3.0/admin-guide/admin_quickstart.html#installation).
+Installation of Singularity doesn't require root, but to access the full features of Singularity, such as the use of its preferred image format, the Singularity binary must be owned by root, with the `setuid` bit enabled.
+This is [documented here](https://www.sylabs.io/guides/2.6/admin-guide/security.html#how-does-singularity-do-it).
```
vsoch: I would encourage the reader to first go to their resource admin and ask for a systems install. It's a reasonable suggestion, it's also likely that it's already installed, and telling the user it's doable to install in user space is more likely to lead to many issues that can be avoided.

docs/tutorials/Containers.md
```diff
@@ -0,0 +1,475 @@
...
+If you do not give Singularity these privileges, you are restricted to using sandbox images, which come with a number of downsides.
```
vsoch: I wouldn't encourage using Sandbox images either. The entire point is to ensure reproducibility, and with suggestions like this, we are undoing that. We should encourage the user to build read only squash fs binaries.

docs/tutorials/Containers.md
```diff
@@ -0,0 +1,475 @@
...
+Because this is not an ideal way to run Singularity, you might consider forwarding [this letter](https://www.sylabs.io/guides/3.0/user-guide/installation.html#singularity-on-a-shared-resource) to your admins.
+
```
vsoch: Yes! This last bit should come first, something like:

> Singularity is best installed on the system level. if your resource doesn't have it, (here is a letter) that you can forward to your admins. If they are unable or unwilling to install the software, then your fallback is (some link here), although this approach is more error prone, less reproducible, and generally not recommended.

docs/tutorials/Containers.md
```diff
@@ -0,0 +1,475 @@
...
+#### Configuring Cromwell for Singularity
+
+Once Singularity is installed, the main configuration block you'll need to modify is the `config` block inside `backend.providers` in your Cromwell configuration.
+In particular, this block contains a key called `submit-docker`, which will contain a script that is run whenever a job needs to run that uses a Docker image.
+If the job does not specify a Docker image, then the regular `submit` block will be used instead.
+
```
vsoch: My expectation is to see the example config here.

docs/tutorials/Containers.md
```diff
@@ -0,0 +1,475 @@
...
+    -J ${job_name} \
+    -D ${cwd} \
+    -o ${cwd}/execution/stdout \
+    -e ${cwd}/execution/stderr \
+    -t ${runtime_minutes} \
+    ${"-c " + cpus} \
+    --mem-per-cpu=${requested_memory_mb_per_core} \
+    --wrap "singularity exec --bind ${cwd}:${docker_cwd} docker://${docker} ${job_shell} ${script}"
+  """
+```
```
vsoch: I'm just curious - can you provide an example use case (somewhere in these docs) of when it *is* appropriate to use Singularity on top of Cromwell, and when it isn't? For many users that just need to launch Singularity containers via slurm, Cromwell is overkill. I'm finding that users don't know this - and think they are required to have a workflow manager.

docs/tutorials/Containers.md
```diff
@@ -0,0 +1,475 @@
...
+    # Build the Docker image into a singularity image, using the head node
+    IMAGE=${cwd}/${docker}.sif
+    singularity build $IMAGE docker://${docker}
+  
+    # Submit the script to SLURM
+    sbatch \
+      [...]
+      --wrap "singularity exec --bind ${cwd}:${docker_cwd} $IMAGE ${job_shell} ${script}"
+  """
+```
```
vsoch: I would follow this up by saying "if worker nodes have access to a shared filesystem, it's best to build the container beforehand, and then provide a direct path to it"

(then put example of build here, followed by config with path). 

If the above is true, we are hugely helpful to advise the user to not run a thing that is going to try downloading the same layers to 1000 different worker nodes when it doesn't have to.

docs/tutorials/Containers.md
```diff
@@ -0,0 +1,475 @@
...
+      --wrap "singularity exec --userns --bind ${cwd}:${docker_cwd} $IMAGE ${job_shell} ${script}"
+"""
+```
+
+
+#### Singularity Cache
+By default, Singularity will cache the Docker images you pull in `~/.singularity`, in your home directory.
+
+However, if you are sharing your Docker images with other users, you probably want this cache to exist somewhere that is accessible to everyone.
+
```
vsoch: It's also important the variable is exported (not just defined) in the .bashrc or similar otherwise it won't take effect. I'd advise users to do this immediately because even if they *don't* need to share layers, their $HOME is going to fill up almost immediately, they will get locked out, and have to ask their admin for help. Best to avoid that.

docs/tutorials/Containers.md
```diff
@@ -0,0 +1,475 @@
...
+    docker: 'ubuntu:latest'
+}
+```
+
+But what you should do is this:
+```wdl
+runtime {
+    docker: 'ubuntu@sha256:7a47ccc3bbe8a451b500d2b53104868b46d60ee8f5b35a24b41a86077c650210'
+}
+```
```
vsoch: Amen!

vsoch:
Bravo! This is seriously great work, and so badly needed.

geoffjentry:
@illusional I wanted to echo the comments from @vsoch. Thanks to all of you who worked to get this going.

docs/tutorials/Containers.md
geoffjentry: That's an interesting point.

IMO I'd still declare it to be "best practice" for both WDL & CWL, but that's because IMO having one's workflow be portable is **also** "best practice". One might choose not to care for all sorts of reasons, but I think as a community we should strive to encourage that direction.

Regardless, I agree on `s/practise/practice` as we try to aim for 🇺🇸 spellings (much to the chagrin of @cjllanwarne )

docs/tutorials/Containers.md
```diff
@@ -0,0 +1,475 @@
...
+
+#### Cromwell: Run-in-background
+
+By enabling Cromwell's run-in-background mode, you remove the necessity for the `kill`, `check-alive` and `job-id-regex` blocks, which disables some safety checks when running workflows:
+
+- If there is an error starting the container or executing the script, Cromwell may not recognise this error and hang. For example, this may occur if the container attempts to exceed its allocated resources (runs out of memory); the container daemon may terminate the container without completing the script.
+- If you abort the workflow (by attempting to close Cromwell or issuing an abort command), Cromwell does not have a reference to the job and will not be able to execute the container.
+
+This is only necessary in local environments where there is no job manager to control this, however if your container technology can emit an identifier to stdout, then you are able to remove the run-in-background flag. Note that the check-alive block is not called to check a job-status, rather Cromwell looks for the presence of an `rc` file (see the previous section for more information).
```
illusional: I'm going to remove:
> Note that the check-alive block is not called to check a job-status, rather Cromwell looks for the presence of an `rc` file (see the previous section for more information). 

This seems to be changed in [cromwell-37](https://github.com/broadinstitute/cromwell/issues/4560), so I don't want to make any assertions.

docs/tutorials/Containers.md
illusional: I changed it to your suggestion :)

docs/tutorials/Containers.md
illusional: I reordered the sentence to mix both of your ideas.
> To promote reproducibility and portability, it's considered best practice to define containers for a WDL and CWL task to run in - this ensures that running the same task on a different system will run the exact same software.

docs/tutorials/Containers.md
illusional: I changed it to lead better towards the next sections, I'm pretty happy to change it back and add the "discussed next" if you don't think this change adds value.
> For this reason, this tutorial will also explore other technologies that support the reproducibility and simplicity of running a workflow that use docker containers; Singularity and udocker.

docs/tutorials/Containers.md
illusional: I reordered these paragraphs to more strongly imply you should install with root, and add you can technically do it without later:
> Before you can configure Cromwell on your HPC system, you (or your sysadmin) will have to install Singularity, which is documented here. It's recommended to perform an installation with root access to gain access to the full set of features in Singularity. You might consider forwarding this letter to your admins.
>
> Although the installation of Singularity doesn't strictly require root, to access the full features of Singularity, such as the use of its preferred image format, the Singularity binary must be owned by root with the setuid bit enabled (documented here).

docs/tutorials/Containers.md
illusional: I did a bit of restructuring in this section to take advantage of some of the ideas you raised through your comments below to. I moved the caching stuff up to the top, reworded a few things to better lead into the configs and added some stronger motivations for the choices in the config. I'll reply to your comments below, but they may appear out of order on this PR.

docs/tutorials/Containers.md
illusional: I've added these three dot-points to give motivation (and reasoning) behind the build and execs, and removed one of the simpler configs as it's kind of not great practi_s_e given you probably don't want people to accidentally download and build the layers, and then consider the file system on top of that.

> - Make sure Singularity is loaded (and in our path), for example you can call module load Singularity/3.0.1.
> - We should treat worker nodes as if they do not have stable access to the internet or build access, so we will pull the container before execution of the task.
> - It's a good idea to ask Singularity to build the image into the execution directory of the task as an artifact, and to save rendering time on the worker node.

IMO, I think giving them the full context and config, and they want to strip it back to be _simpler_, then they can do that.

docs/tutorials/Containers.md
illusional: My personal use case, I've got workflows with 15-30 tasks each which have a different piece of software (GATK4, bwamem, samtools, etc):
- Each have / I can easily find a docker container for each
- I'm running the workflow on multiple HPCs and cloud environments
- For the life of me I can't figure out how to `module load` through CWL.
- I don't have access to install software, or it's impractical to install the exact version I want
- Reproducibility is a key performance indicator for the workflows I'm running.

Docker has a well established container repository that we can use, but unfortunately we can't run Docker on our HPC. My motivation and seemingly [others](https://github.com/broadinstitute/cromwell/issues/2177#issuecomment-330341279) is, I kind of want a (moderately) drop-in replacement for docker.

___

When it isn't is you don't have a software requirement, you're happy to always module load or always have the modules loaded. eg, the way I'm understanding bcbio, the CWL generated by bcbio talks to the bcbio client, so you don't need to run software in Singularity.

docs/tutorials/Containers.md
illusional: I've added a more explicit requirement to export the variable, more comments here: https://github.com/broadinstitute/cromwell/pull/4635#discussion_r258272667,

illusional:
Thanks for the comments! I've (hopefully) replied to them all and covered all of your points. @TMiguelT  do you mind reading through once more and make sure I haven't removed anything essential.

docs/tutorials/Containers.md
TMiguelT: I don't think I'm trying to encourage the use of sandboxes here, it's just that it's the only option without `setuid` on `singularity`

docs/tutorials/Containers.md
TMiguelT: I've tweaked this section and added a tad more text

TMiguelT:
I notice the Singularity Cache is now the second topic within Singularity. Do you mind if I put it back at the end? I don't think there's any need to confuse users with it immediately. @illusional

docs/tutorials/Containers.md
TMiguelT: I've now clarified this

docs/tutorials/Containers.md
TMiguelT: I think @vsoch is asking about when you might use purely Singularity on HPC, rather than Cromwell + Singularity. You've mostly been talking about using pure Cromwell.

TMiguelT: But to answer that question, I think the Cromwell documentation makes it very clear that Cromwell is a WDL + CWL workflow engine (and that you don't need Cromwell unless you have these workflows). I don't think people will find this tutorial without already having decided that they want Cromwell to orchestrate their tasks.

geoffjentry: I think that makes sense. It seems safe to assume that the reader is already in the Cromwell space. We can keep an eye out for the situation @vsoch is describing and edit as necessary

docs/tutorials/Containers.md
illusional: I missed replying to this comment, but new text is:
> For example, the following script should run in the `ubuntu:latest` container:

illusional:
@TMiguelT I've moved the singularity cache section back down. Hopefully this is all the comments actioned.

Sorry @vsoch, I realised I've been resolving your comments as I actioned them, but I probably should've left them open as they're from your review. Also sorry for what was probably a lot of email spam over the past few hours.

docs/tutorials/Containers.md
TMiguelT: This topic is a tricky one. Building the images once beforehand is obviously a better solution because it saves having to rebuild the image. But the point of Cromwell is that we want to run the exact same workflow on HPC as on Cloud. And if you have to manually build images and then edit the workflow in order to find that image file, then we're not facilitating that. At very least, Singularity will cache the Docker layers. As I've said before, if Cromwell had a `build-docker` or `pull-docker` hook that let us build once before running, that would solve the problem. But in the meantime I actually don't want to add the recommendation to build the container once beforehand, because it harms portability, increases the amount of work to run a portable workflow, and confuses this already confusing topic for readers.

vsoch: It's more than just saving time to build the images, for some clusters the concurrent downloads could hurt the performance of the entire system, and not just for one user. Hiding this information because Cromwell doesn't have a command line flag doesn't help anyone. The users want to run the job right, and it's our job to provide complete information. It's wrong for us to hide information from them for any reason. It's wrong for us to make decisions on their behalf that hide important details. It's reasonable to say "if you are able to do A, then do A" and "if you are not able to do A, then do B."

illusional: I'll second this for udocker as well, udocker will (re)build the image on each run, there's a way to pull and build (_create_), but there doesn't seem to be an intuitive way to get the built image, or build otherwise. Every time you `udocker create` it will rebuild the image, and there's no easy way to look up by the floating tag.

TMiguelT: Okay, well then I'll write a section explaining the downside to this approach, and explain the alternative of pre-building images. But how would we actually implement this with Cromwell config? If they go through the whole workflow and build a bunch of `.sif` files into a directory, how do we now tell Cromwell to use them? Do we then have to also edit every task command line to point to these image files?

vsoch: @geoffjentry is this possible?

TMiguelT: I want to get this documentation in ASAP, so I hope we can find a workaround. However at the same time I wonder what long term changes need to be made to fix this problem properly. Ideally we want Singularity to work like Docker, which I think can handle multiple simultaneous pull commands. I've opened a Cromwell issue [here](https://github.com/broadinstitute/cromwell/issues/4673) that would solve it from one angle, but maybe this problem is better solved by better image caching in Singularity? What do you all think?

vsoch: From the [release notes](https://github.com/sylabs/singularity/releases) it looks like more work is being done for caching,  but we have to outline your use case. What Singularity already does is cache docker layers, and then assemble them at runtime. This applies to docker pulls. For shub (or image binary address) pulls, that's when you get an image file. So the strategy to get singularity support for caching isn't to use docker, but actual (entire) image binaries, either via some https address or their library.

TMiguelT: We still need the caching to work given a docker URL, because that's how all WDL/CWL workflows will be written. But we want Singularity to cache the whole image that it builds from that, not just the layers.

vsoch: I don't think that need is served, so you should open an issue and request it.

danbills:
Thanks for all the hard work @illusional and @vsoch !

illusional:
Hey @TMiguelT @vsoch, we noticed that on a system without `mksquashfs` in its path, the `singularity exec` from Dockerhub fails. This seems to be backed up here: http://singularity.lbl.gov/install-linux

> Note that when you configure, squashfs-tools is not required, however it is required for full functionality. You will see this message after the configuration:
> `mksquashfs from squash-tools is required for full functionality`
> If you choose not to install squashfs-tools, you will hit an error when you try a pull from Docker Hub, for example.

I get slightly conflicting information from the Singularity 3 docs which just says: 
> Note that squashfs-tools is an image build dependency only and is not required for Singularity build and run commands.
(https://www.sylabs.io/guides/3.0/user-guide/quick_start.html?highlight=squashfs)

We did install `squashfs` and it's in our `$PATH`, but it seems Singularity is only looking at:
- `/bin/mksquashfs`
- `/usr/bin/mksquashfs`
- `/sbin/mksquashfs`
- `/usr/sbin/mksquashfs`
- `/usr/local/bin/mksquashfs`
- `/usr/local/sbin/mksquashfs`

Any thoughts here, as you are almost always required to pull from docker hub (it's kind of the default).

___

I also noticed with some testing that in the udocker submit, if you exclude the `--rm` it will run a bit quicker. @danbills, am I able to make changes since the review?

geoffjentry:
@illusional please do make the edits

vsoch:
@illusional this discrepancy is something you need to open an issue for on the sylabs board.

TMiguelT:
@illusional I'm sure it's true that squashfs is a build requirement. The admins installed it on our cluster worker nodes for that reason. So yes we should probably mention that in the Singularity installation section. Is that what you're asking?

vsoch:
Yes it's a requirement - what you might want to clarify with Sylabs is if there is different functionality (or different limitations given you don't have it) between the two versions.

TMiguelT:
Although having said that, Singularity 3 doesn't use squashfs for its images, and I don't think the SIF format is based on squashfs, so maybe it's not required? It would indeed be helpful to have someone from Singularity to comment. I might ask on the Slack channel?

vsoch:
That's not entirely correct - SIF can [support multiple types](https://github.com/sylabs/sif/blob/838fcf45c7628a0d057cc218db20b45eefdd33e5/pkg/sif/sif.go#L143) but is generally thought of as is a derivative of squashfs. The squashfs-tools are required.

vsoch:
Yes, please ask on the slack channel for details! :)

illusional:
Thanks @geoffjentry, I've removed the `--rm` flag. I think my other point is Singularity and not relevant to this discussion, so I'm happy with this now. Is there anything I should be doing to finalise this and get it merged?

geoffjentry:
@illusional I think that as long as @vsoch and @TMiguelT are :+1: on the current state that we're good to go. I'll wait to hear from them just to be on the safe side.

docs/tutorials/Containers.md
vsoch: LGTM

vsoch:
Looks good to me. :)

PR #4639, opened 2019-02-12
[Biscayne] WDL 'None' literal

cjllanwarne:
This PR is:
- 90% examples
- 9% Cromwell code
- 1% WDL grammar update

Once merged, OpenWDL can merge as "implemented":
  - [ ] https://github.com/openwdl/wdl/pull/263
  - [ ] https://github.com/openwdl/wdl/pull/284


wom/src/main/scala/wom/types/WomType.scala
```diff
@@ -40,6 +40,7 @@ trait WomType {
 
   final def isCoerceableFrom(otherType: WomType): Boolean = otherType match {
     case WomAnyType => true
+    case WomNothingType => true
```
aednichols: Should this be false?

It seems strange that we would let `WomString("asdf")` coerce to the same underlying value as the `None` literal.

cjllanwarne: Note that this is `isCoerceableFrom` - and `Nothing` can coerce to everyting.

`Nothing` (aka the [bottom type](https://en.wikipedia.org/wiki/Bottom_type)) is "the inner type of an empty array" - or in this case, an empty `Optional`.

In case its proximity to `Any` is confusing here, a real `Any` type should probably be the ultimate supertype, and we should have `case WomAnyType => false`... but in fact "WomAnyType" is really closer to some sort of "WomDeterminedAtRuntimeType".

NB: *perhaps* `Any` and `Nothing` are actually the same thing in Wom and the distinction is meaningless... but I haven't tested out what happens if we try to reduce these two down to just a single type

PR #4644, opened 2019-02-13
Token logging sanity log

cjllanwarne:
* Makes the logging start after 5 rather than 10 seconds
* Logs at least once, even if just to say that there will be no further logging
* Runs in centaur too (so that we can see the logs, and the code gets exercised more)

### DIFF
--- a/engine/src/main/scala/cromwell/engine/workflow/tokens/JobExecutionTokenDispenserActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/tokens/JobExecutionTokenDispenserActor.scala
@@ -55,7 +55,7 @@ class JobExecutionTokenDispenserActor(override val serviceRegistryActor: ActorRe
         }(context.dispatcher)
         ()
       case None =>
-        log.info(s"Not triggering log of token queue status. Effective log interval = $effectiveLogInterval")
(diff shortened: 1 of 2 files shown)


engine/src/main/scala/cromwell/engine/workflow/tokens/JobExecutionTokenDispenserActor.scala
```diff
@@ -38,20 +38,25 @@ class JobExecutionTokenDispenserActor(override val serviceRegistryActor: ActorRe
...
+  context.system.scheduler.scheduleOnce(5.seconds) {
+    effectiveLogInterval match {
+      case Some(someInterval) =>
+        log.info(s"Triggering log of token queue status. Effective log interval = $someInterval")
+        context.system.scheduler.scheduleOnce(someInterval) {
+          self ! LogJobExecutionTokenAllocation(someInterval)
+        }(context.dispatcher)
+        ()
+      case None =>
+        log.info(s"Not triggering log of token queue status. Effective log interval = $effectiveLogInterval")
```
aednichols: Won't `effectiveLogInterval` just be trivially `None` in this case? Can we just omit that part?

PR #4645, opened 2019-02-13
Log metadata summarization totals

cjllanwarne:
A useful refactor of the summarizer FSM, and a log of where the summarizer has reached
- Don't be afraid to vote 👎 on the logging if you think it's going to be too noisy

Sample log message: `2019-02-13 12:52:19,609 cromwell-system-akka.dispatchers.service-dispatcher-97 INFO  - Metadata summarizer has now reached: 52837`


danbills:
Have you considered `debug` level logging?

cjllanwarne:
I think this was decided to be "too noisy". Removing.

PR #4654, opened 2019-02-14
Refactor CAJEA pollStatus logic

### LINKED ISSUE #4651: error running cromwell on SLURM: "Fatal exception polling for status"
I think I am experiencing a bug in cromwell version 37.  The problem occurs when I submit a job to SLURM.  The job gets submitted but cromwell crashes without waiting for the job to finish.  Cromwell works fine when run locally or when I use version 36.  

## my command 
 java is v1.8
```
java -Dconfig.file=cori.conf -jar cromwell-37.jar run test.wdl 
```

## wdl 
```
workflow jgi_dap_leo {

    call doSomething { }

}

task doSomething {
    runtime {
        mem: "8G"
        cpu: 1
        time: "0:60:0"
        backend: "SLURM"
    }
    command {
        free
    }
}
```

## config 
```

system {
    job-rate-control {
        jobs = 1
        per = 1 second
    }
}

backend {
    default="Local"
    providers {
        SLURM {
            config {
                runtime-attributes = """
                String time
                Int cpu
                String mem
                """

                submit = """
                sbatch -J leo_dap -t ${time} -c ${cpu} --mem=${mem} -C haswell -q regular -A m342 --wrap "/bin/bash ${script}"
                """
                kill = "scancel ${job_id}"
            }
        }
    }
}
```

## system I'm running on 

NERSC's cori machines:
Cray XC40, comprised of Intel Xeon "Haswell" processor nodes

## cromwell logs
[cromwellError.txt](https://github.com/broadinstitute/cromwell/files/2866540/cromwellError.txt)

cjllanwarne:
To fix the "unmatched case" error in #4651 :

* Make states a strict ADT rather than string-based
* Hopefully made the `pollStatus` logic a little easier to follow

Closes #4651

### DIFF
--- a/supportedBackends/sfs/src/main/scala/cromwell/backend/impl/sfs/config/ConfigAsyncJobExecutionActor.scala
+++ b/supportedBackends/sfs/src/main/scala/cromwell/backend/impl/sfs/config/ConfigAsyncJobExecutionActor.scala
@@ -1,6 +1,7 @@
 package cromwell.backend.impl.sfs.config
 
 import java.io.PrintWriter
+import java.time.LocalDateTime
 
 import common.validation.Validation._
 import cromwell.backend.RuntimeEnvironmentBuilder
@@ -17,6 +18,8 @@ import wom.callable.Callable.{InputDefinition, OptionalInputDefinition}
 import wom.expression.NoIoFunctionSet
 import wom.transforms.WomCommandTaskDefinitionMaker.ops._
 import wom.values.{WomEvaluatedCallInputs, WomOptionalValue, WomString, WomValue}
+
+import scala.util.{Failure, Success}
 
 /**
   * Base ConfigAsyncJobExecutionActor that reads the config and generates an outer script to submit an inner script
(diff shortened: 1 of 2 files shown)


cjllanwarne:
Waiting behind #4666 for the SLURM tests

supportedBackends/sfs/src/main/scala/cromwell/backend/sfs/SharedFileSystemAsyncJobExecutionActor.scala
```diff
@@ -16,19 +15,34 @@ import scala.concurrent.Future
 import scala.concurrent.duration._
 import scala.util.{Failure, Success, Try}
 
-case class SharedFileSystemRunState(status: String, date: Calendar) {
+sealed trait SharedFileSystemRunState {
+  def status: String
+  def terminal: Boolean = false
```
aednichols: [ToL] if I were writing this, I would not use a default so that someone extending this trait has to explicitly consider whether a newly-added state is terminal or not

supportedBackends/sfs/src/main/scala/cromwell/backend/impl/sfs/config/ConfigAsyncJobExecutionActor.scala
```diff
@@ -248,44 +251,84 @@ class DispatchedConfigAsyncJobExecutionActor(override val standardParams: Standa
...
+            returnCodeTemp.moveTo(jobPaths.returnCode)
+            SharedFileSystemJobFailed
+          } catch {
+            case _: FileAlreadyExistsException =>
+              // If the process has already completed, there will be an existing RC:
+              // Essentially, this covers the rare race condition whereby the task completes between starting to write the
+              // fake RC, and trying to copy it:
+              returnCodeTemp.delete(true)
+              // Looks like a new RC file has appeared - let's go back out of this loop and wait for it:
+              waiting
```
aednichols: If the task completed, why are we waiting?

Is it that Cromwell needs to hold on for another clock tick before it picks up the real RC?

cjllanwarne: It's more like "Cromwell should pick this up on the next poll anyway, so there's no need to reimplement the same logic again here". Sort of like "recurse back into the same function again", albeit on a delay

I guess I should add a warning log in case it gets wedged here somehow.

cjllanwarne:
@aednichols I would like to point out that just because
> someone extending this trait has to explicitly consider whether a newly-added state is terminal or not

... that doesn't mean they'll necessarily get it right...  

cf [[1]](https://github.com/broadinstitute/cromwell/pull/4654/commits/ff6790e7242dfd1e25eea2568d3bc42b649714f7), [[2]](https://github.com/broadinstitute/cromwell/pull/4654/commits/accfb5bc5d144d61e79b0e6058f4efa144ff424f) 😳

chapmanb:
Chris and Adam;
Thanks so much for this fix. I really appreciate having an updated release with this working plus all the goodness since release 36.

PR #4657, opened 2019-02-14
Revert "Auto-commit metadata inserts."

mcovarr:
Revert problematic metadata autocommit changes.
This reverts commit c3f56e2c8775625e4c36303d61f75a75974a2ef4.

### DIFF
--- a/database/sql/src/main/scala/cromwell/database/slick/MetadataSlickDatabase.scala
+++ b/database/sql/src/main/scala/cromwell/database/slick/MetadataSlickDatabase.scala
@@ -35,7 +35,7 @@ class MetadataSlickDatabase(originalDatabaseConfig: Config)
   override def addMetadataEntries(metadataEntries: Iterable[MetadataEntry])
                                  (implicit ec: ExecutionContext): Future[Unit] = {
     val action = DBIO.seq(metadataEntries.grouped(insertBatchSize).map(dataAccess.metadataEntries ++= _).toSeq:_*)
-    runAction(action)
+    runTransaction(action).map(_ => ())
   }
 
   override def metadataEntryExists(workflowExecutionUuid: String)(implicit ec: ExecutionContext): Future[Boolean] = {
--- a/database/sql/src/main/scala/cromwell/database/slick/SlickDatabase.scala
+++ b/database/sql/src/main/scala/cromwell/database/slick/SlickDatabase.scala
@@ -164,14 +164,10 @@ abstract class SlickDatabase(override val originalDatabaseConfig: Config) extend
   private val debugExitStatusCodeOption = ConfigFactory.load.getAs[Int](debugExitConfigPath)
 
   protected[this] def runTransaction[R](action: DBIO[R], isolationLevel: TransactionIsolation = TransactionIsolation.RepeatableRead): Future[R] = {
-    runActionInternal(action.transactionally.withTransactionIsolation(isolationLevel))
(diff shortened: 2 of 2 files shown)


ruchim:
👍

database/sql/src/main/scala/cromwell/database/slick/MetadataSlickDatabase.scala
```diff
@@ -35,7 +35,7 @@ class MetadataSlickDatabase(originalDatabaseConfig: Config)
   override def addMetadataEntries(metadataEntries: Iterable[MetadataEntry])
                                  (implicit ec: ExecutionContext): Future[Unit] = {
     val action = DBIO.seq(metadataEntries.grouped(insertBatchSize).map(dataAccess.metadataEntries ++= _).toSeq:_*)
-    runAction(action)
+    runTransaction(action).map(_ => ())
```
kshakir:
```suggestion
    runTransaction(action)
```

The `map` doesn't do anything.

database/sql/src/main/scala/cromwell/database/slick/SlickDatabase.scala
```diff
@@ -164,14 +164,10 @@ abstract class SlickDatabase(override val originalDatabaseConfig: Config) extend
   private val debugExitStatusCodeOption = ConfigFactory.load.getAs[Int](debugExitConfigPath)
 
   protected[this] def runTransaction[R](action: DBIO[R], isolationLevel: TransactionIsolation = TransactionIsolation.RepeatableRead): Future[R] = {
-    runActionInternal(action.transactionally.withTransactionIsolation(isolationLevel))
+    runAction(action.transactionally.withTransactionIsolation(isolationLevel))
   }
 
   protected[this] def runAction[R](action: DBIO[R]): Future[R] = {
-    runActionInternal(action.withPinnedSession)
```
kshakir: I don't think this file needs-to or should-be reverted. Changing this line effectively changes `runAction` to spread composite database actions across as many database connections as possible. Instead, I believe callers would expect composite commands to run in the same connection.

danbills: I was under impression that this line is the source of the bug, as this is the code called by metadata inserts. 

I'm all for improvements but for the sake of stability I think we revert and go from there.

kshakir: Revert away, but this should be re-fixed in the future for `writeWorkflowHeartbeats`.

> I was under impression that this line is the source of the bug

Switching from `runAction` to `runTransaction` will effectively run `addMetadataEntries` in a [single session anyway, due to the call to `action.transactionally`](http://slick.lightbend.com/doc/3.3.0/dbio.html#transactions).

There are two reverts currently in this PR:
1. `addMetadataEntries` now calls `runTransaction` instead of `runAction`
2. `runAction` now calls `database.run(action)` instead of `database.run(action.withPinnedSession)`

If 1. is reverted, then `addMetadataEntries` won't be affected by changes to `runAction`. Meanwhile, reverting 2. ends up changing the `writeWorkflowHeartbeats` that still calls `runAction`, removing the session isolation.

https://github.com/broadinstitute/cromwell/blob/6e76a0e8d8e7a1c70ed18c2dff30307832866453/database/sql/src/main/scala/cromwell/database/slick/WorkflowStoreSlickDatabase.scala#L119-L120

cjllanwarne:
Adding "DO NOT MERGE" until we determine that this is really the problem

PR #4660, opened 2019-02-15
Expect some progress in centaur tests

cjllanwarne:
Adds an expectation before all `pollUntilStatus` tests that we see at least *some* progress within a minute


src/ci/resources/centaur_application.conf
```diff
@@ -11,7 +11,7 @@ centaur {
   standardTestCasePath: ${CROMWELL_BUILD_CENTAUR_TEST_DIRECTORY}
 
   error-reporter {
-    retry-attempts: 2
+    retry-attempts: 0
```
salonishah11: Will this increase the occurrences of flaky tests in Travis?

centaur/src/main/scala/centaur/test/Test.scala
```diff
@@ -171,10 +171,50 @@ object Operations {
   }
 
   /**
-    * Polls until a specific status is reached. If a terminal status which wasn't expected is returned, the polling
-    * stops with a failure.
+    * Polls until a valid status is reached.
+    * If an unexpected status returned, the polling stops with a failure.
```
mcovarr: doesn't quite sentence

src/ci/resources/centaur_application.conf
mcovarr: did you mean to commit this?

centaur/src/main/scala/centaur/test/Test.scala
```diff
@@ -171,10 +171,50 @@ object Operations {
...
-  def pollUntilStatus(workflow: SubmittedWorkflow, testDefinition: Workflow, expectedStatus: WorkflowStatus): Test[SubmittedWorkflow] = {
+  def expectSomeProgress(workflow: SubmittedWorkflow,
+                         testDefinition: Workflow,
+                         expectedStatuses: Set[WorkflowStatus],
+                         timeout: FiniteDuration): Test[SubmittedWorkflow] = {
+    new Test[SubmittedWorkflow] {
+      def status(remainingTimeout: FiniteDuration): IO[SubmittedWorkflow] = {
+        for {
+          workflowStatus <- CentaurCromwellClient.status(workflow)
+          mappedStatus <- workflowStatus match {
```
aednichols: Any reason you don't do `mappedStatus = workflowStatus` and get rid of all the non-IO `IO`'s in the `match`?

aednichols: Is it because the `IO.sleep` is captured in the IO and deferred?

Issue #4555, opened 2019-01-15
Cannot construct WomMapType(WomStringType,WomAnyType) with mixed types


antonkulaga:
Totally valid workflow fails in the last stage in the latest development version of cromwell because of:
```
Cannot construct WomMapType(WomStringType,WomAnyType) with mixed types
```
error.
I enclose both wdl and input. 
[quantification.zip](https://github.com/broadinstitute/cromwell/files/2761544/quantification.zip)

The error is the following:

```json
Workflow failed

[...]

Issue #4512, opened 2019-01-03
Certain heterogeneous maps validate in Womtool, fail in Cromwell


aednichols:
Based on a report from @curoli 

I can't immediately reason out what the correct behavior is (post-vacation fuzziness...) so this is just a  a log of stuff that looked suspicious to me.

---

A workflow like
```
version 1.0

workflow test {
  
  Map[String, String] m = {"a": "a", "b": "b"}
  String s = "string"

  output {
    File write_attempt = write_json({"m": m, "s": s})
  }
}
```
validates in Womtool but fails at runtime with error
```
WorkflowManagerActor Workflow 2a3db889-e126-467b-be60-6abb815ea46e failed (during ExecutingWorkflowState):
  java.lang.UnsupportedOperationException:
    Cannot construct WomMapType(WomStringType,WomAnyType) with mixed types:
      Map(
        WomString(m) -> WomMap(WomMapType(WomStringType,WomStringType),Map(WomString(a) -> WomString(a), WomString(b) -> WomString(b))),
        WomString(s) -> WomString(string)
      )
```

---

The problem is not so simple as heterogeneous types in the map values; the workflow
```
version 1.0

workflow test {
  
  String s = "string"
  Float f = 0.1
  File file = "asdf"

  output {
    File write_attempt = write_json({"s": s, "f": f, "file": file})
  }

}
```
works just fine:
```
{"s":"string","f":"0.1","file":"asdf"}
```

---

Interestingly, if we take out `String s = "string"` we do get an error in Womtool, but it's a confusing one - why would we say a map value has to be an `Object` when we clearly used `String`, `Float`, and `File` right above?
```
version 1.0

workflow test {
  
  Map[String, String] m = {"a": "a", "b": "b"}

  output {
    File write_attempt = write_json({"m": m})
  }
}
```
yields
```
womtool validate any_map.wdl 
Failed to process workflow definition 'test' (reason 1 of 1):
  Failed to process declaration 'File write_attempt = write_json({ "m": m })' (reason 1 of 1):
    Failed to process expression 'write_json({ "m": m })' (reason 1 of 1):
      Invalid parameter 'MapLiteral(Map(StringLiteral(m) -> IdentifierLookup(m)))'. Expected 'Object' but got 'Map[String, Map[String, String]]'
```

curoli:
[...]

PR #4661, opened 2019-02-15
Enhance map errors

aednichols:
In response to:
1. the difficulty of debugging https://github.com/broadinstitute/cromwell/issues/4555 and https://github.com/broadinstitute/cromwell/issues/4512 due to excessive, unformatted output
2. the lack of tests on the error message generation

### DIFF
--- a/wom/src/main/scala/wom/values/WomMap.scala
+++ b/wom/src/main/scala/wom/values/WomMap.scala
@@ -52,18 +52,18 @@ final case class WomMap private(womType: WomMapType, value: Map[WomValue, WomVal
   val typesUsedInKey = value.map { case (k, _) => k.womType }.toSet
 
   if (typesUsedInKey.size == 1 && typesUsedInKey.head != womType.keyType)
-    throw new UnsupportedOperationException(s"Could not construct a $womType with this value: $value")
+    throw new UnsupportedOperationException(s"Could not construct a $womType with map keys of unexpected type: [${value.keys.mkString(", ")}]")
 
   if (typesUsedInKey.size > 1)
-    throw new UnsupportedOperationException(s"Cannot construct $womType with mixed types: $value")
+    throw new UnsupportedOperationException(s"Cannot construct $womType with mixed types in map keys: [${value.keys.mkString(", ")}]")
 
   val typesUsedInValue = value.map { case (_, v) => v.womType }.toSet
 
   if (typesUsedInValue.size == 1 && typesUsedInValue.head != womType.valueType)
-    throw new UnsupportedOperationException(s"Could not construct a $womType as this value: $value")
+    throw new UnsupportedOperationException(s"Could not construct a $womType with map values of unexpected type: [${value.values.mkString(", ")}]")
 
   if (typesUsedInValue.size > 1)
-    throw new UnsupportedOperationException(s"Cannot construct $womType with mixed types: $value")
+    throw new UnsupportedOperationException(s"Cannot construct $womType with mixed types in map values: [${value.values.mkString(", ")}]")
 
   override def toWomString: String =
     "{" + value.map {case (k,v) => s"${k.toWomString}: ${v.toWomString}"}.mkString(", ") + "}"
--- a/wom/src/test/scala/wom/types/WomMapTypeSpec.scala
+++ b/wom/src/test/scala/wom/types/WomMapTypeSpec.scala
@@ -57,14 +57,47 @@ class WomMapTypeSpec extends FlatSpec with Matchers  {
       case Failure(_) => fail("Failed to coerce a map to an object")
     }
   }
-  it should "detect invalid map construction if there are mixed types" in {
+
+  it should "detect invalid map construction if there are mixed keys" in {
     try {
-      WomMap(WomMapType(WomStringType, WomStringType), Map(WomInteger(0) -> WomString("foo"), WomString("x") -> WomInteger(2)))
+      WomMap(WomMapType(WomStringType, WomStringType), Map(WomInteger(0) -> WomString("foo"), WomString("x") -> WomString("y")))
       fail("Map initialization should have failed")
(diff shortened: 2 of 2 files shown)


wom/src/main/scala/wom/values/WomMap.scala
```diff
@@ -52,18 +52,18 @@ final case class WomMap private(womType: WomMapType, value: Map[WomValue, WomVal
...
 
   val typesUsedInValue = value.map { case (_, v) => v.womType }.toSet
 
   if (typesUsedInValue.size == 1 && typesUsedInValue.head != womType.valueType)
-    throw new UnsupportedOperationException(s"Could not construct a $womType as this value: $value")
+    throw new UnsupportedOperationException(s"Could not construct a $womType with map values of unexpected type: [${value.values.mkString(", ")}]")
 
   if (typesUsedInValue.size > 1)
-    throw new UnsupportedOperationException(s"Cannot construct $womType with mixed types: $value")
+    throw new UnsupportedOperationException(s"Cannot construct $womType with mixed types in map values: [${value.values.mkString(", ")}]")
```
cjllanwarne: Maybe even just report the types rather than values?

wom/src/test/scala/wom/types/WomMapTypeSpec.scala
```diff
@@ -57,14 +57,47 @@ class WomMapTypeSpec extends FlatSpec with Matchers  {
...
+
+  it should "detect invalid map construction if there are mixed keys" in {
     try {
-      WomMap(WomMapType(WomStringType, WomStringType), Map(WomInteger(0) -> WomString("foo"), WomString("x") -> WomInteger(2)))
+      WomMap(WomMapType(WomStringType, WomStringType), Map(WomInteger(0) -> WomString("foo"), WomString("x") -> WomString("y")))
       fail("Map initialization should have failed")
     } catch {
-      case _: UnsupportedOperationException => // expected
+      case e: UnsupportedOperationException =>
+        e.getMessage shouldEqual "Cannot construct WomMapType(WomStringType,WomStringType) with mixed types in map keys: [WomInteger(0), WomString(x)]"
```
cjllanwarne: Should map these to the value Strings to avoid leaking these `WomX()` wrappers

wom/src/test/scala/wom/types/WomMapTypeSpec.scala
```diff
@@ -57,14 +57,47 @@ class WomMapTypeSpec extends FlatSpec with Matchers  {
...
     }
   }
+
+  it should "detect invalid map construction if there are mixed values" in {
+    try {
+      WomMap(WomMapType(WomStringType, WomStringType), Map(WomString("bar") -> WomString("foo"), WomString("x") -> WomInteger(2)))
+      fail("Map initialization should have failed")
+    } catch {
+      case e: UnsupportedOperationException =>
+        e.getMessage shouldEqual "Cannot construct WomMapType(WomStringType,WomStringType) with mixed types in map values: [WomString(foo), WomInteger(2)]"
```
cjllanwarne: This definitely *feels* like a bug...  I would hope that we'd have no problem doing that coercion from Int to String

wom/src/test/scala/wom/types/WomMapTypeSpec.scala
aednichols: I think the string coercion might strip out crucial information, such as eliminating the visible difference between `WomSingleFile("/foo/bar.wdl")` and `WomString("/foo/bar.wdl")`

wom/src/main/scala/wom/values/WomMap.scala
aednichols: I thought about this, but I think it's useful to clue the user in to which values have which type, to avoid them having to work backwards to figure out which value introduced the offending type incompatibility

wom/src/test/scala/wom/types/WomMapTypeSpec.scala
aednichols: Very possible, but feels out of scope for this PR

antonkulaga:
@aednichols you probably misunderstood my issue that you closed. It was not about having a mixed map, but about Cromwell coercing well defined the types to Map[String, Any] at runtime while everything is validated by both Cromwell and womtool at compiletime.
I opened another issue, where the same problem emerges with the struct, where Cromwell tries to consider struct output as Map[String, Any] at runtime. https://github.com/broadinstitute/cromwell/issues/4663

aednichols:
@antonkulaga this enhancement is intended as a general improvement, not a fix for any particular issue.

Rest assured we will get to the bottom of your struct issue!

wom/src/test/scala/wom/types/WomMapTypeSpec.scala
aednichols: We should discuss this one

wom/src/main/scala/wom/values/WomMap.scala
cjllanwarne: I think that's normally OK, but if this is a list of 10,000 file names then the value can quickly become unwieldy

cjllanwarne: OTOH in the spirit of "making improvements" and "biasing for action", this is definitely a step forward from what was here previously.

wom/src/test/scala/wom/types/WomMapTypeSpec.scala
cjllanwarne: I do think we ought to this an issue with this exact example as the error case? I'd add labels `Bug`/`Redwood`.

FWIW I wanted to also suggest `Comply with language spec` but the spec is _probably_ silent on the exact point of what is acceptable if the types are mixed in `Map` construction but would be coercible, it might be more a case of "Update language spec comply with what is intuitively right"

PR #4672, opened 2019-02-20
Double summarization to reduce skipping rows

cjllanwarne:
Obviously not a *good*  fallback position, but so long as we are skipping metadata entries for summarization, reducing the probability seems like an idea worth considering (with the obvious intention to turn this off as soon as we work out the problems at the database layer).

NB: the actual change here is the third commit - the other two are just ride-alongs to make the FSM nicer and provide status reporting


cjllanwarne:
I think my work here is done...

Issue #4689, opened 2019-02-28
Database communications link failures in prod


aednichols:
![screenshot-2019-2-28 kibana](https://user-images.githubusercontent.com/1087943/53603858-d5c7bb00-3b80-11e9-9330-a9ac9f9032dc.png)

[Kibana link](https://kibana.logit.io/app/kibana#/discover?_g=(refreshInterval:(display:Off,pause:!f,value:0),time:(from:now-90d,mode:quick,to:now))&_a=(columns:!(_source),index:'*-*',interval:auto,query:(query_string:(analyze_wildcard:!t,query:'host:%22gce-cromwell-prod601%22%20AND%20%22Communications%20link%20failure%22')),sort:!('@timestamp',desc)))

The same error message showed up in #4360, #3387, and #2519 but in those the "last packet" time was short and more or less random, while here it's repeatedly 929,284 milliseconds - or precisely 15 minutes, 30 seconds.

aednichols:
We haven't had one of these in a month (knock on wood)

![Screen Shot 2019-03-27 at 4 03 33 PM](https://user-images.githubusercontent.com/1087943/55108399-e9940d80-50a9-11e9-9597-2bbd7eebb51b.png)

PR #4690, opened 2019-03-01
Upgrade mysql connector

aednichols:
According to [the docs](https://dev.mysql.com/downloads/connector/j/) it is "highly recommended" that we upgrade to the latest version, even if our MySQL is behind - `5.6.36-google-log` in our case.

Reason I started looking into this was https://github.com/broadinstitute/cromwell/issues/4689

### DIFF
--- a/database/sql/src/main/scala/cromwell/database/slick/SlickDatabase.scala
+++ b/database/sql/src/main/scala/cromwell/database/slick/SlickDatabase.scala
@@ -3,7 +3,7 @@ package cromwell.database.slick
 import java.sql.{Connection, PreparedStatement, Statement}
 import java.util.concurrent.{ExecutorService, Executors}
 
-import com.mysql.jdbc.exceptions.jdbc4.MySQLTransactionRollbackException
+import com.mysql.cj.jdbc.exceptions.MySQLTransactionRollbackException
 import com.typesafe.config.{Config, ConfigFactory}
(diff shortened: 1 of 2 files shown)


aednichols:
Note: I probably need to do something about the comments Khalid left in the `dbmsDependencies` section, gonna see if CI explodes then circle back

antonkulaga:
Guys, you must update it! Most of my workflow failures were because of mysql connection loss

danbills:
this is affecting prod, we should merge it and figure out maria from there.

database/sql/src/main/scala/cromwell/database/slick/SlickDatabase.scala
```diff
@@ -3,7 +3,7 @@ package cromwell.database.slick
 import java.sql.{Connection, PreparedStatement, Statement}
 import java.util.concurrent.{ExecutorService, Executors}
 
-import com.mysql.jdbc.exceptions.jdbc4.MySQLTransactionRollbackException
+import com.mysql.cj.jdbc.exceptions.MySQLTransactionRollbackException
```
cjllanwarne: Who's CJ?

aednichols: "MySQL Connector/J" (wonder if it works on OS/2)

PR #4655, opened 2019-02-14
Revert auto-commit metadata inserts [37 hotfix]

mcovarr:
Revert problematic metadata autocommit changes.
This reverts commit c3f56e2c8775625e4c36303d61f75a75974a2ef4.

PR #4693, opened 2019-03-01
Revert "Revert "Auto-commit metadata inserts." (#4655)"

mcovarr:
This reverts commit 2ce5910f6102fe4fefe415c5bdc9180153c2b9b0.


kshakir:
Looks like this PR will either need to also add [this commit](https://github.com/broadinstitute/cromwell/pull/4669), or retry the PapiV2 tests around a dozen times until it gets just enough memory.

I'm also fine if #4669 is submitted as a separate 37_hotfix-PR first.

mcovarr:
Thanks for the tip and the fix @kshakir !

PR #4695, opened 2019-03-02
Add key/value index for custom labels.

mcovarr:


database/sql/src/main/scala/cromwell/database/slick/tables/WorkflowMetadataSummaryEntryComponent.scala
```diff
@@ -110,11 +110,8 @@ trait WorkflowMetadataSummaryEntryComponent {
         submissionTimestampFilter,
         notASubworkflowFilter
       )
-      // Unwrap the optional filters.  If any of these filters are not defined, replace with `include` to include all
-      // rows which might otherwise have been filtered.
-      val filters = optionalFilters.map(_.getOrElse(include))
-      // AND together the filters.  If there are no filters at all return `include`.
-      filters.reduceLeftOption(_ && _).getOrElse(include)
+      // `flatten` to remove empty filters, AND together the remaining filters.  If there are no filters at all return `include`.
+      optionalFilters.flatten.reduceLeftOption(_ && _).getOrElse(include)
```
mcovarr: I can revert this bit if it's going to be superseded by the custom SQL work @cjllanwarne is doing (and would needlessly complicated his rebase)

PR #4697, opened 2019-03-02
adding cromwell example backends folder

vsoch:
This pull request will add a folder of example backend providers to cromwell, along with a README.md to (verbosely) describe them. Importantly:

 - links to documentation are included in the examples
 - the user understands the context of where to provide it under backend -> providers
 - the user doesn't have to uncomment a million things

I couldn't find the backend page for AWS - the closest I found was https://cromwell.readthedocs.io/en/stable/tutorials/AwsBatch101/. Is there a backend page, proper?

I've done my best to add notes and documentation, and maybe it would be useful to bring in others with expertise for the various providers? This is a first shot, and I expect we will want feedback from others to get it in good shape!

### DIFF
--- a/cromwell.example.backends/LocalExample.conf
+++ b/cromwell.example.backends/LocalExample.conf
@@ -1,0 +1,117 @@
+# This is an example of how you can use the LocalExample backend to define
+# a new backend provider. *This is not a complete configuration file!* The
+# content here should be copy pasted into the backend -> providers section
+# of the cromwell.examples.conf in the root of the repository. You should
+# uncomment lines that you want to define, and read carefully to customize
+# the file. If you have any questions, please open an issue at
+# https://www.github.com/broadinstitute/cromwell/issues
+
+# Documentation
+# https://cromwell.readthedocs.io/en/stable/backends/Local/
+
+    # Define a new backend provider.
+
+    LocalExample {
+
+      # The actor that runs the backend. In this case, it's the Shared File System (SFS) ConfigBackend.
+      actor-factory = "cromwell.backend.impl.sfs.config.ConfigBackendLifecycleActorFactory"
+
+      # The backend custom configuration.
+      config {
+
+        # Optional limits on the number of concurrent jobs
+        #concurrent-job-limit = 5
+
+        # If true submits scripts to the bash background using "&". Only usefull for dispatchers that do NOT submit
+        # the job and then immediately return a scheduled job id.
+        run-in-background = true
+
+        # `temporary-directory` creates the temporary directory for commands.
+        #
+        # If this value is not set explicitly, the default value creates a unique temporary directory, equivalent to:
+        # temporary-directory = "$(mktemp -d \"$PWD\"/tmp.XXXXXX)"
+        #
+        # The expression is run from the execution directory for the script. The expression must create the directory
+        # if it does not exist, and then return the full path to the directory.
+        #
+        # To create and return a non-random temporary directory, use something like:
+        # temporary-directory = "$(mkdir -p /tmp/mydir && echo /tmp/mydir)"
+
+        # `script-epilogue` configures a shell command to run after the execution of every command block.
+        #
+        # If this value is not set explicitly, the default value is `sync`, equivalent to:
+        # script-epilogue = "sync"
+        #
+        # To turn off the default `sync` behavior set this value to an empty string:
+        # script-epilogue = ""
+
+        # The list of possible runtime custom attributes.
+        runtime-attributes = """
+        String? docker
+        String? docker_user
+        """
+
+        # Submit string when there is no "docker" runtime attribute.
+        submit = "/usr/bin/env bash ${script}"
+
+        # Submit string when there is a "docker" runtime attribute.
+        submit-docker = """
+        docker run \
+          --rm -i \
+          ${"--user " + docker_user} \
+          --entrypoint ${job_shell} \
+          -v ${cwd}:${docker_cwd} \
+          ${docker} ${script}
+        """
+
+        # Root directory where Cromwell writes job results.  This directory must be
+        # visible and writeable by the Cromwell process as well as the jobs that Cromwell
+        # launches.
+        root = "cromwell-executions"
+
+        # Root directory where Cromwell writes job results in the container. This value
+        # can be used to specify where the execution folder is mounted in the container.
+        # it is used for the construction of the docker_cwd string in the submit-docker
+        # value above.
+        dockerRoot = "/cromwell-executions"
+
+        # File system configuration.
+        filesystems {
+
+          # For SFS backends, the "local" configuration specifies how files are handled.
+          local {
+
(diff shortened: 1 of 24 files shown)


cromwell.example.backends/LocalExample.conf
```diff
@@ -0,0 +1,117 @@
+# This is an example of how you can use the LocalExample backend to define
+# a new backend provider. *This is not a complete configuration file!* The
+# content here should be copy pasted into the backend -> providers section
```
cjllanwarne: @kshakir may need to correct me, but you may be able to add a static `include` to the top of these files to get the defaults automatically.

FWIW you *always* get the `reference.conf` automatically, so even with just this you should have a working Cromwell system

cjllanwarne:
Just a comment to say that this is *awesome* and thank you for adding these!

cromwell.example.backends/LocalExample.conf
vsoch: Oh interesting! Can you give an example of how include would work? I'll add a note about the reference.conf.

vsoch: The idea (or what I was going for) isn't to include these separate files, but just to cleanly set out examples for the user to choose (and then copy paste into their final configuration).

vsoch:
hey @kshakir ! I'd be happy to make that change for you (the first one). For the second one, I've linked the example in the folder to the online docs, and I'd suggest that the update of the docs themselves be a separate PR. I've noticed with Cromwell (and other software, generally) that it's much cleaner / better to have smaller PRs that have scoped changes.

vsoch:
@kshakir does this apply also to udocker?

vsoch:
Note that there is no such thing as "docker and HPC" but there *is* udocker and HPC.

vsoch:
@kshakir please see my changes in https://github.com/broadinstitute/cromwell/pull/4697/commits/2c44e5eae1457453854b230adc1d109d8057ea01 to remove out and err, and replace "script" with "docker_script"

cromwell.example.backends/LocalExample.conf
kshakir: > you may be able to add a static include to the top of these files to get the defaults automatically

Yep! Stanzas are nice for copy-composing configs. For example grab different stanzas to build a conf that includes SLURM + mysql + call-caching. However for a "complete configuration file" technically all one needs is to prepend `include required(classpath("application.conf"))` above the `backend {}` stanza.

For example, a fully testable SLURM conf (based on [the conf used in CI](https://github.com/broadinstitute/cromwell/blob/3c3a3f85ba1229738265b11c3573ccb538b719c2/src/ci/resources/slurm_application.conf#L47)):

```hocon
include required(classpath("application.conf"))

backend {
  default = "SlurmNoDocker"
  providers {
    SlurmNoDocker {
      config {
        submit = """sbatch -J ${job_name} -D ${cwd} --wrap "/bin/bash ${script}""""
        kill = "scancel ${job_id}"
      }
    }
  }
}
```

Something like that could get someone started. If there isn't going to be the include at the top, I **do** like the comment stating that this is a part of a full config, and "not a complete configuration file!"

vsoch: This might be a personal preference, but I would advise people to try to minimize dependencies (if they want to share their configuration, for example) and try to put all the config information into one file. The include is a neat trick, but it overcomplicates it. This isn't like writing a script where you want to import / include to have a cleaner application, it's just setting up a configuration file that is actually more optimal to just have one file. Does that make sense? The above could, if desired, go into some "advanced usage" part of the docs as to not confuse the new user.

vsoch: It would also potentially lead to bugs if there are equivalently defined objects in both the main config and the included files (e.g., duplicates of backends -> providers). It's much simpler to say "here is an example, take what you want from it and put it into your configuration file"

vsoch:
Woohoo! Thanks @kshakir. When this one goes through I'd be happy to talk more about your vision for the docs, and open up another PR to fix that up.

cromwell.example.backends/LocalExample.conf
kshakir: Makes sense. I'm fine leaving the existing comment instead of the include.

kshakir:
Haven't used udocker yet myself, but I'm pretty sure `${docker_script}` should be used there also instead of `${script}`.

As for HPC and docker, there are some renegades out there running nodes with docker. The very few I've come across are not on-prem, spin up private larger machines on cloud resources, add HPC+docker, and then run whatever scripts on top of that. This includes our CI... until we get one of the non-root LXC implementations going.

Related side-note: I would love one day for our CI to also install-then-test udocker, [rootless docker](https://github.com/moby/moby/blob/fc01c2b481097a6057bec3cd1ab2d7b4488c50c4/docs/rootless.md), singularity, etc. If anyone comes across this and knows the magic spell to fully-install-and-configure any of these container tools, or other HPC like HTCondor, GridEngine, etc., (especially on Travis!), please drop a gist or a PR. For example, here's the installation others helped me scrape together for SLURM:
- [Cromwell SLURM CI Installation Script](https://github.com/broadinstitute/cromwell/blob/3c3a3f85ba1229738265b11c3573ccb538b719c2/src/ci/bin/test_slurm.inc.sh)
- [Cromwell SLURM CI Conf](https://github.com/broadinstitute/cromwell/blob/3c3a3f85ba1229738265b11c3573ccb538b719c2/src/ci/resources/slurm_application.conf) with commands to run/kill/check jobs
- [Cromwell SLURM CI Test Runner](https://github.com/broadinstitute/cromwell/blob/3c3a3f85ba1229738265b11c3573ccb538b719c2/src/ci/bin/testCentaurSlurm.sh)

vsoch:
> As for HPC and docker, there are some renegades out there running nodes with docker. The very few I've come across are not on-prem, spin up private larger machines on cloud resources, add HPC+docker, and then run whatever scripts on top of that. This includes our CI... until we get one of the non-root LXC implementations going.

Interesting! If it's a cloud setup, I don't know if we would call that HPC. When I say HPC I typically mean SLURM, SGE, etc., where there are many thousands of untrusted users. Putting Docker on those systems - no admin thinking clearly would do it. For a really small number of users that you absolutely trust? That could be a thing, haha. :)

vsoch:
Is the one failed run in travis related to these changes? It seems like a random TravisCI issue, but it might be my inexperience with the tests.

aednichols:
This is likely a bogus failure, I went ahead and restarted the job

vsoch:
Poor TravisCI, I have had the same, and wound up switching a ton of testing suites over to CircleCI. I hope TravisCI survives after the major layoffs, now things aren't looking so great! :/ 

You can do it Travis!!

vsoch:
Sure thing. https://github.com/broadinstitute/cromwell/pull/4697/commits/d994821808bf9aa9b11ea54271b2702c50c4b713

danbills:
Thanks @vsoch !!

vsoch:
This is great! Thank you!

PR #4701, opened 2019-03-04
Revert upgrade mysql connector (we have a better theory now)

aednichols:
Better theory is https://github.com/broadinstitute/firecloud-develop/pull/1556

IMO the risk/reward for the upgrade no longer checks out, especially since we are in a bit of a crunch mode and relatively ill-equipped to deal with unexpected problems.


mcovarr:
background?

aednichols:
When looking into https://github.com/broadinstitute/cromwell/issues/4689 there was a period of time when I thought an obsolete MySQL connector could be the most likely cause in the absence of other theories, so we went ahead and pushed this through.

Soon after, we found the connection pool clue, which is rather more likely. I take a conservative view of upgrading such essential libraries, so just being on the latest wasn’t reason enough to keep it (IMO).

danbills:
@kshakir is this the library that we downgraded the last time we tried to upgrade all of our libraries?

PR #4706, opened 2019-03-05
Clean up MySQL connector deprecation warning

aednichols:
Also checked whether any of the comments about MySQL were actionable and cleaned them up.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -313,9 +313,6 @@ object Dependencies {
   private val dbmsDependencies = List(
     "org.hsqldb" % "hsqldb" % hsqldbV,
     /*
-    When going to 6.0.x, will need to change the jdbc driver to com.mysql.cj.jdbc.Driver
-    - https://dev.mysql.com/doc/connector-j/6.0/en/connector-j-api-changes.html
-
     The url may also need the parameters:
     - serverTimezone=UTC via http://stackoverflow.com/a/36793896/3320205
     - nullNamePatternMatchesAll=true via https://liquibase.jira.com/browse/CORE-2723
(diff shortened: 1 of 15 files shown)


project/Dependencies.scala
```diff
@@ -312,14 +312,6 @@ object Dependencies {
...
   private val dbmsDependencies = List(
     "org.hsqldb" % "hsqldb" % hsqldbV,
-    /*
-    When going to 6.0.x, will need to change the jdbc driver to com.mysql.cj.jdbc.Driver
-    - https://dev.mysql.com/doc/connector-j/6.0/en/connector-j-api-changes.html
-
-    The url may also need the parameters:
-    - serverTimezone=UTC via http://stackoverflow.com/a/36793896/3320205
-    - nullNamePatternMatchesAll=true via https://liquibase.jira.com/browse/CORE-2723
-     */
```
cjllanwarne: I guess we didn't need these parameters after all?

aednichols: As far as I could tell from my research (Travis might prove me wrong)

kshakir:
You will likely need two `v37` confs.  During the upgrade tests, new confs are needed to run against the cromwell-37 jar.

Something like these should work:

#### local_37_application.conf
```hocon
include "local_application.conf"
database.db.driver = "com.mysql.jdbc.Driver"
```

#### papi_v2_37_application.conf
```hocon
include "papi_v2_application.conf"
database.db.driver = "com.mysql.jdbc.Driver"
```

aednichols:
Thanks @kshakir, you really saved my bacon on this one. Would have taken me a while to figure out I actually need to add files, instead of just editing existing ones.

PR #4719, opened 2019-03-07
Engine/PAPI upgrade testing script hotfix support.

mcovarr:


src/ci/bin/test.inc.sh
```diff
@@ -532,7 +532,14 @@ cromwell::private::setup_prior_version_resources() {
         grep 'val cromwellVersion' "${CROMWELL_BUILD_ROOT_DIRECTORY}/project/Version.scala" \
         | awk -F \" '{print $2}' \
         )"
-    prior_version=$((current_version - 1))
+
+    # This function should only ever run on Travis PR builds where TRAVIS_PULL_REQUEST_BRANCH is set.
+    # https://stackoverflow.com/questions/229551/how-to-check-if-a-string-contains-a-substring-in-bash
+    if [ -z "${TRAVIS_PULL_REQUEST_BRANCH##*_hotfix}" ]; then
```
mcovarr: perhaps this should error if the variable is in fact not set

cjllanwarne:
Recurring bash script changes question: 

> Did you run `shellcheck` over the generated files?

src/ci/bin/test.inc.sh
cjllanwarne: Yes please!

cjllanwarne: I personally think an `egrep` with a real regex for `[0-9]+_hotfix` would be 
- (a) better (eg don't hit branches like `test_hotfix_logic`) and 
- (b) maybe even easier to read (I'd never seen this bash-ism for substrings before)

src/ci/bin/testCentaurEngineUpgradeLocal.sh
```diff
@@ -6,11 +6,15 @@ export CROMWELL_BUILD_OPTIONAL_SECURE=true
 # shellcheck source=/dev/null
 source "${BASH_SOURCE%/*}/test.inc.sh" || source test.inc.sh
 
-cromwell::build::setup_common_environment
+if [ "${CROMWELL_BUILD_PROVIDER}" = "${CROMWELL_BUILD_PROVIDER_TRAVIS}" -a  -n "${TRAVIS_PULL_REQUEST_BRANCH}" ]; then
```
cjllanwarne: This might be the right way to do things, but it seems odd to me to see `=` in the same expression as `-a` and `-n`?

cjllanwarne: Should we sanity-check that these magic environment variables also exist here too?

src/ci/bin/test.inc.sh
mcovarr: discussed facially, when in Rome and all that

src/ci/bin/testCentaurEngineUpgradeLocal.sh
mcovarr: These other variables are much more fundamental to the operation of the script, if they're not set I think that fact would be impossible to ignore. 🙂

src/ci/bin/test.inc.sh
mcovarr: Actually I think I'll implement that suggestion after all. The looseness of the substring match was bugging me but Bash regex matching is too quirky for my tastes.

src/ci/bin/testCentaurEngineUpgradeLocal.sh
mcovarr: Hopefully more palatable now as a result of your earlier Shellcheck suggestion.

PR #4726, opened 2019-03-11
Fix path in run_tests_parallel.sh script

mcovarr:
Wondering if this has been broken since the Great Monorepoing? 😬


geoffjentry:
:+1:

Issue #4740, opened 2019-03-13
S3 permissions errors using Cromwell 37+


kshakir:
### Possible Workaround

Try using the `default` AWS auth scheme along with the [AWS default credentials](https://docs.aws.amazon.com/sdk-for-java/v2/developer-guide/credentials.html#credentials-default) setup on the system / environment.

### Background

Starting in cromwell 37, the AWS S3 SDK was upgraded from `2.0.0-preview-9` to `2.3.9`. Along with this upgrade several cromwell calls to the S3 SDK were updated to match changes in the library.

Post upgrade, the existing Cromwell AWS CI tests continue to pass, however there have been reports of permissions problems or other errors.
- https://gatkforums.broadinstitute.org/firecloud/discussion/comment/56245/#Comment_56245
- https://github.com/broadinstitute/cromwell/issues/4541
- https://github.com/broadinstitute/cromwell/issues/4686
- https://github.com/broadinstitute/cromwell/issues/4731

Because the CI tests are passing, and they use default credentials, it is possible that each of these issues may be also be worked around by using the [AWS default credentials](https://docs.aws.amazon.com/sdk-for-java/v2/developer-guide/credentials.html#credentials-default) scheme.

It is unclear how the very similar SDK calls worked with `2.0.0-preview-9` and not `2.3.9`. However the fix might include passing in credentials via the `Map env` / `Properties props`.

| Type                                         | Cromwell copy targeting `2.3.9`                                                                                                                                                      | "Original" targeting `2.0.0-preview-9`                                                                                                                                               |
|----------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| (Attempted) setting of key/secret from props | [cromwell-37/AmazonS3Factory.java](https://github.com/broadinstitute/cromwell/blob/37/filesystems/s3/src/main/java/org/lerch/s3fs/AmazonS3Factory.java#L81)          | [elerch-sdk2/AmazonS3Factory.java](https://github.com/elerch/Amazon-S3-FileSystem-NIO2/blob/sdk2/src/main/java/org/lerch/s3fs/AmazonS3Factory.java#L82)               |
[...]

PR #4743, opened 2019-03-13
Update AWS Batch example conf to reflect our docs

geoffjentry:
Addresses(-ish) #4740 

Our docs were telling users to do one thing, but our copy/paste example was telling users to do something else


cromwell.example.backends/AWS.conf
```diff
@@ -9,16 +9,59 @@
 # Documentation
 # https://cromwell.readthedocs.io/en/stable/tutorials/AwsBatch101/
 
-backend {
-  default = AWS
+include required(classpath("application"))
```
geoffjentry: remove this

PR #4745, opened 2019-03-14
Fix womtool --version

kshakir:


aednichols:
Since I always build `womtool` myself instead of downloading, I did notice this problem but assumed there was some magic that made the version set correctly for release builds.

Issue #4725, opened 2019-03-11
Failing Centaur tests found during horicromtal testing


mcovarr:
With a local MySQL config, Local backend and call caching enabled, the following Centaur tests consistently fail for me regardless of horicromtal or unicromtal:

- [x] draft3_call_cache_capoeira_local
- [x] call_cache_capoeira_local
- [ ] cwl_dynamic_initial_workdir
- [x] long_cmd

mcovarr:
The `long_cmd` is failing because the default MySQL packet size is too small for the gigantic metadata being generated, though I'm not sure if the problem is on the client or server or both. I'm not sure why the other tests are failing.

salonishah11:
The `cwl_dynamic_initial_workdir` fails with below stack trace:
```
2019-03-13 12:29:04,388 cromwell-system-akka.dispatchers.backend-dispatcher-88 ERROR - BackgroundConfigAsyncJobExecutionActor [UUID(c9194073)main:NA:1]: Error attempting to Execute
java.lang.Exception: Failed command instantiation
	at cromwell.backend.standard.StandardAsyncExecutionActor.instantiatedCommand(StandardAsyncExecutionActor.scala:581)
	at cromwell.backend.standard.StandardAsyncExecutionActor.instantiatedCommand$(StandardAsyncExecutionActor.scala:515)
	at cromwell.backend.impl.sfs.config.BackgroundConfigAsyncJobExecutionActor.instantiatedCommand$lzycompute(ConfigAsyncJobExecutionActor.scala:200)
	at cromwell.backend.impl.sfs.config.BackgroundConfigAsyncJobExecutionActor.instantiatedCommand(ConfigAsyncJobExecutionActor.scala:200)
	at cromwell.backend.standard.StandardAsyncExecutionActor.commandScriptContents(StandardAsyncExecutionActor.scala:317)
	at cromwell.backend.standard.StandardAsyncExecutionActor.commandScriptContents$(StandardAsyncExecutionActor.scala:316)
	at cromwell.backend.impl.sfs.config.BackgroundConfigAsyncJobExecutionActor.commandScriptContents(ConfigAsyncJobExecutionActor.scala:200)
	at cromwell.backend.sfs.SharedFileSystemAsyncJobExecutionActor.writeScriptContents(SharedFileSystemAsyncJobExecutionActor.scala:175)
[...]

PR #4747, opened 2019-03-14
Remove leading slash from S3Path's pathWithoutSchema

geoffjentry:
Fixes three centaur tests: input_mirror, sizeenginefunction, if_then_…else_expressions

### DIFF
--- a/filesystems/s3/src/main/scala/cromwell/filesystems/s3/batch/S3BatchIoCommand.scala
+++ b/filesystems/s3/src/main/scala/cromwell/filesystems/s3/batch/S3BatchIoCommand.scala
@@ -32,14 +32,7 @@ package cromwell.filesystems.s3.batch
 
 import software.amazon.awssdk.core.exception.SdkException
 
-import software.amazon.awssdk.services.s3.model.{HeadObjectRequest,
-                                                  HeadObjectResponse,
-                                                  CopyObjectRequest,
-                                                  CopyObjectResponse,
-                                                  DeleteObjectRequest,
-                                                  NoSuchKeyException,
-                                                  S3Response
-                                                }
+import software.amazon.awssdk.services.s3.model.{HeadObjectResponse, CopyObjectResponse, NoSuchKeyException}
 import cromwell.core.io.{IoCommand,
                          IoDeleteCommand,
                          IoSizeCommand,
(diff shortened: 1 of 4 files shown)


geoffjentry:
@kshakir huh, TIL. It currently passes **most** of the centaur tests, maybe I can square that away in a separate PR as we should be running as many as possible.

What's weird is that `long_cmd` fails consistently for me but looks to be on the good list.

geoffjentry:
actually i'll add the 3 for now

kshakir:
> What's weird is that long_cmd fails consistently for me

Are you using the CI config files, or your own config file? `long_cmd` generates an _extremely_ long command line that approximates one of the longer gatk best practice commands. However the test line is so long that an [optional abbreviation](https://github.com/broadinstitute/cromwell/blob/develop/src/ci/resources/build_application.inc.conf#L23) setting was added and enabled-in-CI so that the test could run.

With [a bit of setup](https://github.com/broadinstitute/cromwell/issues/4725#issuecomment-472915580) one can run `src/ci/bin/testCentaurAws.sh` and it will attempt to use the CI configs locally.

geoffjentry:
@kshakir just a standard config file. I hadn't actually looked at the error yet, but will make a note to come back to this discussion :)

filesystems/s3/src/main/scala/cromwell/filesystems/s3/batch/S3BatchIoCommand.scala
```diff
@@ -56,11 +49,6 @@ import cromwell.filesystems.s3.S3Path
   * @tparam U Return type of the response
   */
 sealed trait S3BatchIoCommand[T, U] extends IoCommand[T] {
-  /**
-    * S3 operation to be executed by this command
-    */
-  def operation: S3Response
```
geoffjentry: The `operation` deletions are only semi-related. They weren't being used, which I discovered while tracing through some of the code

mcovarr:
@geoffjentry I ran into the same problem with `long_cmd` when I made the same mistake of using a non-CI config during horicromtal testing. Some lessons learned [here](https://github.com/broadinstitute/cromwell/pull/4748/files).

geoffjentry:
@danbills ;)

Screen Shot 2019-03-19 at 9 25 40 AM

Issue #4354, opened 2018-11-02
Task doesn't fail if the output file can't be found on AWS


TimurIs:
Hi,

The task doesn't fail when the output file can't be found, hence, cromwell tries to run the next job.

Here is a small sample workflow:

```
workflow should_fail
{
  call Task1
  {
    input:
     out_base = "foo"
  }
}

task Task1
{
  String out_base

  command
  <<<
    #typo simulation, one extra 't'
    echo "Hello World!" > ${out_base}.ttxt
  >>>

  runtime
  {
    cpu: 1
    memory: "128 MB"
    docker: "ubuntu:latest"
  }

  output
  {
    File out = "${out_base}.txt"
  }
}
```

command used to run the workflow on the aws: 
`java -Dconfig.file=aws.conf -jar cromwell-36.jar run test.wdl`

Thank you,
Timur

rebrown1395:
@ruchim to set acceptance criteria

geoffjentry:
It looks like at least one of the `bad_output_task` and `bad_file_string` centaur tests fail (well, succeed, but you know ...) for the same reason

geoffjentry:
See if this is fixed by pushing the new proxy

geoffjentry:
Closing. This should now be fixed. Please reopen if you still encounter this.

Issue #4710, opened 2019-03-06
Cromwell retry file download


adamvonp:
Hi All, 

I am currently running Cromwell 36 in AWS batch (both Server and run mode) in one account, lets say ACCT1, and all of my resource files are cross account in another account (ACCT2). Intermittently jobs will fail due to a failure to download a reference file.

Error seen through the Cromwell proxy logs:

> aws s3   cp --sse AES256 --no-progress   s3://s4-somaticgenomicsrd-benchmark-gatk4/database/1.0/ucsc.hg19.fasta.bwt   /cromwell_root//s4-somaticgenomicsrd-benchmark-gatk4/database/1.0/ucsc.hg19.fasta.bwt          download failed: s3://s4-somaticgenomicsrd-benchmark-gatk4/database/1.0/ucsc.hg19.fasta.bwt   to   ../cromwell_root/s4-somaticgenomicsrd-benchmark-gatk4/database/1.0/ucsc.hg19.fasta.bwt   ("Connection broken: error(104, 'Connection reset by peer')",   error(104, 'Connection reset by `peer'))

I'm wondering if it is in the scope of Cromwell on AWS to have a file download retry, or check at least, rather than continuing the task with an empty file which may persist silently causing troubleshooting to be quite difficult.

Best,
Adam

geoffjentry:
I believe this is resolved in #4749 but it requires the proxy image to be pushed & verified that this is resolved

geoffjentry:
Closing as this should now be resolved. Please reopen if you stil encounter this

PR #4749, opened 2019-03-15
AWS backend: sleep 60 secs and retry 5 times if download or upload fails

hurchu:
When download input from S3 to instance, it may fail. But the workflow will continue to run.
Retry and exit after 5 times.

### DIFF
--- a/supportedBackends/aws/src/main/resources/ecs-proxy/proxy
+++ b/supportedBackends/aws/src/main/resources/ecs-proxy/proxy
@@ -40,6 +40,27 @@
 # AWS_CROMWELL_INPUTS
 # AWS_CROMWELL_OUTPUTS
 
+
+awscp() {
+  initExitCode=1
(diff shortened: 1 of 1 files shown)


geoffjentry:
I think this might close #4354 and #4710, as well as fix the `bad_output_task` and `bad_file_string` centaur tests. Will need to double check when the new proxy is pushed

geoffjentry:
Thank you for the contribution @hurchu !

hurchu:
Hi @geoffjentry , I don't know whether this is the right place to ask the question.
I'm currently using cromwell 44. It seems the issue came back again. It failed without retry.
Could you please tell me when the proxy image is pulled? Is it pulled everytime when a new ec2 instance is running? And it pulls from latest docker image?

hurchu:
I figured it. I need to rebuid my AMI to get the latest verion of proxy docker image.

PR #4754, opened 2019-03-18
Fix womtool endpoint to understand nonempty arrays

aednichols:


services/src/test/resources/describe/wdl_1_0/arrays/workflow.wdl
```diff
@@ -7,6 +7,8 @@ workflow arrays {
     Array[Int]? optional_array
     Array[Int?] array_of_optionals
     Array[Int?]? optional_array_of_optionals
+    Array[Int]+ nonempty_array
+    Array[Int]+? optional_nonempty_array
```
salonishah11: What is `optional non-empty array`? Is it that if you have an array it should be non-empty, else no array at all? Similar to `Option[Array[Int]+]`?

aednichols: Yes, that's right - an array that may or may not be defined - but if it is, then it must be nonempty.

services/src/test/resources/describe/wdl_1_0/arrays/workflow.wdl
```diff
@@ -17,6 +19,8 @@ workflow arrays {
     Array[Int]? output_optional_array = []
     Array[Int?] output_array_of_optionals = []
     Array[Int?]? output_optional_array_of_optionals = []
+    Array[Int]+ output_nonempty_array = [5]
```
salonishah11: Can you also explain `output non-empty array`?

services/src/test/resources/describe/wdl_1_0/arrays/workflow.wdl
salonishah11: Sounds good. Thank you for the explanation!

services/src/test/resources/describe/wdl_1_0/arrays/workflow.wdl
aednichols: I stuck an "output" on just to distinguish it from the input variables with the same names, which is helpful when looking at the JSON

salonishah11: Thank you for the explanation @aednichols! :)

PR #4765, opened 2019-03-21
Add TMiguelT's long file name test to Centaur

geoffjentry:


src/ci/bin/testCentaurAws.sh
```diff
@@ -28,5 +28,6 @@ cromwell::build::run_centaur \
     -i inline_files \
     -i sizeenginefunction \
     -i if_then_else_expressions \
+    -e file_name_too_long
```
kshakir: I don't think this does anything in this PR. The `-i` entries above only include those tags/tests. The addition of a `-e` attempts to exclude `file_name_too_long`, but it wasn't included in the first place.

geoffjentry: This is a placeholder to cause a merge conflict with #4766 as the raison d'etre of this centaur test is that it fails on the AWS backend at the moment

PR #4766, opened 2019-03-21
Embiggen the Centaur test suite for AWS backend

geoffjentry:


geoffjentry:
@kshakir I'm hoping that most of them go away in the next month or so

Issue #4772, opened 2019-03-26
Spike: Investigate NPE causing workflow failure


ruchim:
Version: Cromwell v38
Backend: PAPI v2

Running a hello world workflow on Pipelines API v2 against inputs originating from a Requester Pays bucket.

The job finishes running, generates a 0 return code and detritus logs (stdout/stderr). However the workflow failures with an error message (full message attached).
[failed-workflow-metadata.txt](https://github.com/broadinstitute/cromwell/files/3008597/failed-workflow-metadata.txt)

```Unable to complete JES Api Request. Caught NPE while processing operation projects/dvoet-prod-test-20190305-3/operations/11506961050484846699```

Searched the code and found the error here:
https://github.com/broadinstitute/cromwell/blob/2d8ff3b0962bbe84828445fd3ac77d2379e499c2/supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/api/request/GetRequestHandler.scala#L49-L83

Operation metadata attached as well.
[operation-metadata.yaml.txt](https://github.com/broadinstitute/cromwell/files/3008623/operation-metadata.yaml.txt)

ruchim:
This is a regression-- we should have a centaur test case that goes through this scenario, so this failure is surprising and worth investigating.

rexwangcc:
We have also found 2000/2000 workflows failed due to this issue `Unable to complete JES Api Request. Caught NPE while processing...` today. We are using cass-prod `"cromwell": "36-fde91e6"`, and the Cromwell hasn’t been changed for at least ~2 weeks

aednichols:
Response from Google:
>We are aware of a change that might be causing this issue and we are actively working on fixing it.

lmtani:
My workflows are running without errors now.

ruchim:
Learned that a change in Pipelines API omitted the "preemptible" key from the operations metadata, and that change introduced a null pointer in the Cromwell code. 

AC: As a way to address this, it would be great if we could modify the Cromwell code so that when its parsing operation metadata, that if certain keys are missing (such as Preemptible) -- we use the defaults where possible, else fail gracefully with an error that states which information couldn't be parsed, and that caused the workflow to fail.

ruchim:
Closing this as the real fix is filed here: https://github.com/broadinstitute/cromwell/issues/4774

PR #4773, opened 2019-03-26
PAPIv2 defensive programming

aednichols:
`pipeline.getResources.getVirtualMachine.getPreemptible` is intended to be `null` if there was no preemption.

Aaron says,
>Client code should be updated to properly access these fields even if they are absent.

Closes https://github.com/broadinstitute/cromwell/issues/4772

### DIFF
--- a/supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/api/request/GetRequestHandler.scala
+++ b/supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/api/request/GetRequestHandler.scala
@@ -80,7 +80,7 @@ trait GetRequestHandler extends LazyLogging { this: RequestHandler =>
       }
     } catch {
       case npe: NullPointerException =>
-        throw new RuntimeException(s"Caught NPE while interpreting operation ${operation.getName}: ${ExceptionUtils.getStackTrace(npe)} ")
+        throw new RuntimeException(s"Caught NPE while interpreting operation ${operation.getName}: ${ExceptionUtils.getStackTrace(npe)}. JSON was $operation")
     }
   }
 
--- a/supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/api/request/GetRequestHandler.scala
+++ b/supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/api/request/GetRequestHandler.scala
@@ -78,7 +78,7 @@ trait GetRequestHandler { this: RequestHandler =>
       }
     } catch {
       case npe: NullPointerException =>
-        throw new RuntimeException(s"Caught NPE while processing operation ${operation.getName}: ${npe.toString}", npe)
+        throw npe
     }
   }
 
--- a/supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/api/request/GetRequestHandler.scala
+++ b/supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/api/request/GetRequestHandler.scala
@@ -6,7 +6,6 @@ import akka.actor.ActorRef
 import com.google.api.client.googleapis.batch.BatchRequest
 import com.google.api.client.googleapis.json.GoogleJsonError
 import com.google.api.services.genomics.v2alpha1.model._
-import com.typesafe.scalalogging.LazyLogging
 import common.validation.Validation._
 import cromwell.backend.google.pipelines.common.api.PipelinesApiRequestManager._
 import cromwell.backend.google.pipelines.common.api.RunStatus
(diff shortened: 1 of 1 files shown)


supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/api/request/GetRequestHandler.scala
```diff
@@ -78,7 +79,9 @@ trait GetRequestHandler { this: RequestHandler =>
       }
     } catch {
       case npe: NullPointerException =>
-        throw new RuntimeException(s"Caught NPE while processing operation ${operation.getName}: $operation", npe)
+        logger.error(s"Caught NPE while interpreting operation ${operation.getName} from the following JSON response: $operation")
+
+        throw npe
```
mcovarr: TOL logging *and* throwing is usually considered an antipattern as it usually results in multiple copies of the same error message, though it sounds like we were suffering from the opposite problem here. Though it looks like this may be intentionally omitting the NPE from `logger.error(...)` to mitigate that issue somewhat?

aednichols: We want both the JSON and the stack trace to make it into the log.

Unfortunately, I could not find a way to feed the stack trace of `npe` into `new RuntimeException()` such that it would get printed out, so we still need `throw npe`.

Meanwhile, there is no way to add a custom message to `npe`, hence the log statement.

I agree it's a bit goofy but I did not want to invest too much time in code that is a candidate for rewriting.

supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/api/request/GetRequestHandler.scala
```diff
@@ -60,7 +62,7 @@ trait GetRequestHandler { this: RequestHandler =>
         // preemptible is only used if the job fails, as a heuristic to guess if the VM was preempted.
         // If we can't get the value of preempted we still need to return something, returning false will not make the failure count
         // as a preemption which seems better than saying that it was preemptible when we really don't know
-        val preemptible = pipeline.exists(_.getResources.getVirtualMachine.getPreemptible.booleanValue())
+        val preemptible = pipeline.exists(pipeline => Option(pipeline.getResources.getVirtualMachine.getPreemptible).exists(_.booleanValue()))
```
danbills: ToL the nesting could be reduced if the first `exists` were a `flatMap` as in 

```scala
        val preemptible = pipeline.flatMap(pipeline => Option(pipeline.getResources.getVirtualMachine.getPreemptible)).exists(_.booleanValue())
```

supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/api/request/GetRequestHandler.scala
```diff
@@ -78,7 +80,7 @@ trait GetRequestHandler { this: RequestHandler =>
       }
     } catch {
       case npe: NullPointerException =>
-        throw new RuntimeException(s"Caught NPE while processing operation ${operation.getName}: $operation", npe)
+        throw new RuntimeException(s"Caught NPE while interpreting operation ${operation.getName}: ${ExceptionUtils.getStackTrace(npe)} ")
```
danbills: Maybe I'm missing something but you could just pass in the npe to the runtime exception:

```scala 
        throw new RuntimeException(s"Caught NPE while interpreting operation ${operation.getName}", npe)
```

supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/api/request/GetRequestHandler.scala
```diff
@@ -16,13 +17,14 @@ import cromwell.backend.google.pipelines.v2alpha1.api.Deserialization._
...
 import cromwell.core.ExecutionEvent
+import org.apache.commons.lang3.exception.ExceptionUtils
 
 import scala.collection.JavaConverters._
 import scala.concurrent.{ExecutionContext, Future}
 import scala.language.postfixOps
 import scala.util.{Failure, Try, Success => TrySuccess}
 
-trait GetRequestHandler { this: RequestHandler =>
+trait GetRequestHandler extends LazyLogging { this: RequestHandler =>
```
danbills: don't think you need this anymore

supportedBackends/google/pipelines/v2alpha1/src/main/scala/cromwell/backend/google/pipelines/v2alpha1/api/request/GetRequestHandler.scala
aednichols: That is what the old code did and the result was not satisfactory: no stack trace was being printed

Example output: https://travis-ci.com/broadinstitute/cromwell/jobs/187652847

danbills: I see.  I think that is an unfortunate logging situation that is unrelated to this particular issue.

Issue #2519, opened 2017-08-04
Cromwell can fail to respond to metadata requests on large workflows


droazen:
When busy running a large workflow (like the current 20k gnomAD callset), cromwell can fail to respond to metadata requests. This makes it very hard to see what's going on with the workflow.

droazen:
In this situation, metadata requests (via, eg., `curl` or similar) either hang (seemingly) indefinitely or time out with a message like:

```
status": "error",
  "message": "Communications link failure\n\nThe last packet successfully received from the server was 3 milliseconds ago.  The last packet sent successfully to the server was 173,470 milliseconds ago."
}
```

LeeTL1220:
👍 👍 

This makes debugging failures so difficult.  Even with includeKey and excludeKey.

katevoss:
@Horneth is this because of the Slick database exceptions?

Horneth:
The slick exceptions could technically cause this but no this is more generally due to the metadata building process being very expensive therefore causing all kinds of timeouts / errors.

mcovarr:
There's a good forum post on this topic here, including not only `includeKey` and `excludeKey` but also increasing Akka HTTP timeouts.

We might want to change those timeout defaults in `reference.conf` if users are seeing this routinely even with those filters.

https://gatkforums.broadinstitute.org/wdl/discussion/10209/retrieving-metadata-for-large-workflows

aednichols:
This will be fixed with the new metadata implementation.

AlexMTX:
Hi @aednichols ! Are there any news on the new metadata implementation?

aednichols:
Yes, the Cromwell 51 release completed the metadata archiving implementation and it is now operating in production at Broad Institute. Archiving is targeted at our internal production instance only so it's not an officially documented or supported feature, but an enterprising Cromwell admin could surely get it working.

PR #4776, opened 2019-03-28
Add example akka config to avoid metadata timeout issue

EvanTheB:
Add akka http request-timeout idle-timeout examples in the config
To allow users to get metadata results from large workflows

Also delete the now duplicated cromwell.examples.conf file
And delete the backends section which has been split into
separate files

https://gatkforums.broadinstitute.org/wdl/discussion/10209/retrieving-metadata-for-large-workflows

https://github.com/broadinstitute/cromwell/issues/2519


EvanTheB:
The file mentions:
`and you can see it defined in references.conf.`

What is references.conf?

EvanTheB:
Also is there any way to actually enumerate all the available settings? Probably this would be too many, and modifying some of them would break the system. Still it is nice to have, and I have tried searching for this 'inbuilt' defaults file in vain. I guess it is scattered across many java files.

salonishah11:
@EvanTheB `reference.conf` is the configuration file used by Cromwell.
> Also is there any way to actually enumerate all the available settings?

I am not sure how we can do that. But you are right, it might be a lot of effort to restructure/modify those files to be able to enumerate it.

cromwell.example.backends/cromwell.examples.conf
```diff
@@ -12,6 +19,16 @@ webservice {
 
 akka {
   # Optionally set / override any akka settings
+  http {
+    server {
+      # Increasing these timeouts allow rest api responses for very large jobs
+      # to be returned to the user. When the timeout is reached the server would respond
+      # `The server was not able to produce a timely response to your request.`
+      # https://gatkforums.broadinstitute.org/wdl/discussion/10209/retrieving-metadata-for-large-workflows
+      # request-timeout = 300s
```
danbills: I would prefer to list the default, which I believe is 20s

```suggestion
      # request-timeout = 20s
```

danbills:
@EvanTheB references.conf should be "reference.conf` and it is the default settings for all these options, [seen here](https://github.com/broadinstitute/cromwell/blob/1ebf35417c879cf7ed83fd122564613d9d4d86c8/core/src/main/resources/reference.conf).

There are likely ways to enumerate the settings we are overriding, but the default settings that things like the DB connection Pool are legion and I don't know how to list them.

EvanTheB:
thanks for the reviews

PR #4785, opened 2019-04-02
Add config option to shutdown cromwell when unable to write heartbeats.

kshakir:
- Additional wiring for the cromwell terminator.
- Reduced duplicate calls to ConfigFactory.load().
- Increased ability to pass around test configs.
- Provide names for more actors.
- Instrument failures in batch actors.

### DIFF
--- a/engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreHeartbeatWriteActor.scala
+++ b/engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreHeartbeatWriteActor.scala
@@ -1,19 +1,25 @@
 package cromwell.engine.workflow.workflowstore
 
-import java.time.OffsetDateTime
+import java.time.{OffsetDateTime, Duration => JDuration}
+import java.util.concurrent.TimeUnit
 
-import akka.actor.{ActorRef, Props}
+import akka.actor.{ActorRef, CoordinatedShutdown, Props}
 import cats.data.{NonEmptyList, NonEmptyVector}
 import cromwell.core.Dispatcher.EngineDispatcher
 import cromwell.core.WorkflowId
 import cromwell.core.instrumentation.InstrumentationPrefixes
+import cromwell.engine.CromwellTerminator
 import cromwell.engine.workflow.workflowstore.WorkflowStoreActor.WorkflowStoreWriteHeartbeatCommand
 import cromwell.services.EnhancedBatchActor
+import mouse.all._
 
 import scala.concurrent.Future
+import scala.concurrent.duration.FiniteDuration
+import scala.util.{Failure, Success, Try}
 
 case class WorkflowStoreHeartbeatWriteActor(workflowStoreAccess: WorkflowStoreAccess,
                                             workflowHeartbeatConfig: WorkflowHeartbeatConfig,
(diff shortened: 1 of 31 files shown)


engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreHeartbeatWriteActor.scala
```diff
@@ -38,18 +52,60 @@ case class WorkflowStoreHeartbeatWriteActor(workflowStoreAccess: WorkflowStoreAc
...
+          firstFailureOption = Option(startTime)
+        }
+
+        for {
+          failureShutdownDuration <- failureShutdownDurationOption
+          firstFailure <- firstFailureOption
+        } {
+          val now = OffsetDateTime.now()
+          val failureDuration = JDuration.between(firstFailure, now)
+          if (failureDuration.toNanos > failureShutdownDuration.toNanos) {
```
danbills: As I understand it the logic for shutting down:

1. we are waiting for multiple failures
1. a duration must have passed between the first and the last failure, and all results in between must have been failures.

If that is true, I see a couple issues:

* We should fail on the first error
* Even if we decide to wait and see if errors are transient (worth discussing), the current logic won't execute if the `failureShutdownDurationOption` isn't set.  Thus this setting shouldn't be optional.

engine/src/main/scala/cromwell/engine/workflow/workflowstore/WorkflowStoreHeartbeatWriteActor.scala
```diff
@@ -38,18 +61,59 @@ case class WorkflowStoreHeartbeatWriteActor(workflowStoreAccess: WorkflowStoreAc
...
+  We are expecting the underlying FSM to ensure that the call to this method does NOT occur in parallel, waiting for
+  the call to `process` to complete.
+   */
+  private def trackRepeatedFailures(heartbeatDateTime: OffsetDateTime, workflowCount: Int)(processTry: Try[Int]): Unit = {
+    processTry match {
+      case Success(_) =>
+        lastSuccessOption = Option(heartbeatDateTime)
+      case Failure(_: Exception) =>
+        val lastSuccess = lastSuccessOption.getOrElse(
+          sys.error("The lastSuccessOption was somehow still None when calling trackRepeatedFailures")
```
aednichols: This doesn't seem too crazy... what if there's something wrong with Cromwell or the DB that causes all heartbeat writes to fail from the very beginning?

engine/src/test/scala/cromwell/engine/workflow/workflowstore/WorkflowHeartbeatConfigSpec.scala
```diff
@@ -0,0 +1,163 @@
...
+      "when the heartbeat interval is equal to the ttl",
+      """|system.workflow-heartbeats {
+         |  ttl = 2 minutes
+         |  write-failure-shutdown-duration = 0 seconds
+         |}
+         |""".stripMargin,
+      AggregatedMessageException(
+        "Errors parsing WorkflowHeartbeatConfig",
+        List(
+          "The system.workflow-heartbeats.heartbeat-interval (2 minutes)" +
```
aednichols: Nice!

PR #4790, opened 2019-04-03
Enable call caching with Cromwell's AWS backend. Closes #4412

### LINKED ISSUE #4412: Call Caching seems no longer work when aws batch is enabled.
When using aws batch feature, the call caching seems no longer work. 

I am using cromwell-35.jar, the aws.conf is the following

aws {
  application-name = "cromwell"
  auths = [{
      name = "default"
      scheme = "default"
  }]
  region = "default"
}

engine { filesystems { s3 { auth = "default" } } }

backend {
  default = "AWSBATCH"
  providers {
    AWSBATCH {
      actor-factory = "cromwell.backend.impl.aws.AwsBatchBackendLifecycleActorFactory"
      config {
        numSubmitAttempts = 3
        numCreateDefinitionAttempts = 3
        root = "s3://mybucket/cromwell-execution"
	auth = "default"
        concurrent-job-limit = 16
        default-runtime-attributes { queueArn = "arn:aws:batch:us-east-1:<account-id>:job-queue/MyHighPriorityQue-ae4256f76f07d96" }
        filesystems { s3 { auth = "default" } }
      }
    }
  }
}

call-caching {

  enabled = true

  # In a multi-user environment this should be false so unauthorized users don't invalidate results for authorized users.
  invalidate-bad-cache-results = false

  blacklist-cache {
    # The call caching blacklist cache is off by default. This is used to blacklist cache hit paths based on the
    # prefixes of cache hit paths that Cromwell previously failed to copy for authorization reasons.
    enabled: false
    # Guava cache concurrency.
    concurrency: 10000
    # How long entries in the cache should live from the time of their last access.
    ttl: 20 minutes
    # Maximum number of entries in the cache.
    size: 1000
  }
}

database {
  profile = "slick.jdbc.MySQLProfile$"
  db {
    driver = "com.mysql.jdbc.Driver"
    url = "jdbc:mysql://cromwell-db-rdscluster.cluster.us-east-1.rds.amazonaws.com/cromwell"
    user = "myuser"
    password = "my password"
    connectionTimeout = 5000
  }
}

my hello.wdl is:
task hello {
  String name
  command {
    echo 'Hello ${name}!' > "hello${name}.txt"
  }
  output {
    File response = "hello${name}.txt"
  }
  runtime {
    docker: "ubuntu:latest"
  }
}

workflow test {
  call hello
}

My hello_inputs.json is:
{
  "test.hello.name": "World",
}

I ran java -Dconfig.file=aws.callcache.conf -jar ~/cromwell-35.jar run -i hello_inputs.json hello.wdl several times, each time there is a new job submitted to aws batch. Should it submit aws batch only once  when first time ran it?  

Your assistance will be highly appreciated.

Thanks
Jing

### LINKED ISSUE #4563: call caching consistently disabled (AWS Batch back end)
Backend: AWS Batch

Workflow: https://github.com/FredHutch/reproducible-workflows/blob/master/WDL/unpaired-panel-consensus-variants-human/frankenstein.wdl

Input file: https://github.com/FredHutch/reproducible-workflows/blob/master/WDL/unpaired-panel-consensus-variants-human/map-variantcall-hg38.json

Possibly related to #4412 but not sure as I don't see the same error message.

When submitting a workflow via the cromwell server we **consistently** see a failure to hash some items in S3 resulting in call caching being disabled for the run. We have seen this for a number of workflows, here we are including just one. 

Call caching is a **hugely** important feature for us and if it is not available we may would have to reconsider using Cromwell.

I think I have discussed with @ruchim the fact that all objects in S3 have a hash already computed (the ETag header) so there should not be timeouts in computing these hashes as they are available with a head request (you don't need to download the whole object). 

Error message (extract from `/metadata` output):

```
     "callCaching": {
          "hashFailures": [
            {
              "causedBy": [],
              "message": "Hashing request timed out for: s3://bucketname/cromwell-tests/Panel_BWA_GATK4_Samtools_Var_Annotate/162c863f-c22a-4b7c-bb37-f5195b329b36/call-ApplyBQSR/shard-0/smallTestData.hg38.recal.bam"
            }
          ],
          "allowResultReuse": false,
          "hit": false,
          "result": "Cache Miss",
          "effectiveCallCachingMode": "CallCachingOff"
        },
```

Config file:

```


call-caching {
    enabled = true
    invalidate-bad-cache-results = true
}


database {
  # Store metadata in a file on disk that can grow much larger than RAM limits.
    profile = "slick.jdbc.HsqldbProfile$"
    db {
      driver = "org.hsqldb.jdbcDriver"
      url = "jdbc:hsqldb:file:aws-database;shutdown=false;hsqldb.tx=mvcc"
      connectionTimeout = 3000
    }
}



aws {
  application-name = "cromwell"
  auths = [
    {
      name = "default"
      scheme = "default"
    }
    {
        name = "assume-role-based-on-another"
        scheme = "assume_role"
        base-auth = "default"
        role-arn = "arn:aws:iam::xxx:role/fbucketname"
    }
  ]
  // diff 1:
  # region = "us-west-2" // uses region from ~/.aws/config set by aws configure command,
  #                    // or us-east-1 by default
}
engine {
  filesystems {
    s3 {
      auth = "assume-role-based-on-another"
    }
  }
}
backend {
  default = "AWSBATCH"
  providers {
    AWSBATCH {
      actor-factory = "cromwell.backend.impl.aws.AwsBatchBackendLifecycleActorFactory"
      config {
        // Base bucket for workflow executions
        root = "s3://bucketname/cromwell-tests"
        // A reference to an auth defined in the `aws` stanza at the top.  This auth is used to create
        // Jobs and manipulate auth JSONs.
        auth = "default"
        // diff 2:
        numSubmitAttempts = 1
        // diff 3:
        numCreateDefinitionAttempts = 1
        default-runtime-attributes {
          queueArn: "arn:aws:batch:us-west-2:xxx:job-queue/GenomicsHighPriorityQue-xxx"
        }
        filesystems {
          s3 {
            // A reference to a potentially different auth for manipulating files via engine functions.
            auth = "default"
          }
        }
      }
    }
  }
}
```

geoffjentry:
I believe this also resolves #4563 but haven't verified that yet.


supportedBackends/aws/src/main/scala/cromwell/backend/impl/aws/AwsBatchJobPaths.scala
```diff
@@ -55,19 +53,4 @@ final case class AwsBatchJobPaths(override val workflowPaths: AwsBatchWorkflowPa
   // at the time that code runs.
   override def defaultStdoutFilename: String = s"$logBasename-stdout.log"
   override def defaultStderrFilename: String = s"$logBasename-stderr.log"
-
```
geoffjentry: This was all copypasta from the PAPI backends. that `foo.log` file is the PAPI log. What'd happen is that this would be inserted into the custom metadata but not exist when files would get copied over

cjllanwarne:
Are we at a point where we can add a centaur test for call caching in AWS?

geoffjentry:
@cjllanwarne did you not see the set of call caching centaur tests which were enabled by this PR?

cjllanwarne:
I did not, but now that I do.... 🙌

geoffjentry:
@cjllanwarne I admittedly didn't turn on the ones which would require a specialized test (i.e. involves S3) even though I believe they should now work. I was going to follow that up later on, mainly being lazy about getting the files in appropriate places

src/ci/bin/testCentaurAws.sh
```diff
@@ -30,12 +30,8 @@ cromwell::build::run_centaur \
     -e non_root_default_user \
     -e draft3_glob_access \
     -e bad_output_task \
-    -e cachewithinwf \
```
cjllanwarne: TOL I never really understood why these files existed rather than listing the allowable (or excluded) backends in the `.test` files (which is also possible AFAIK?)

supportedBackends/aws/src/main/scala/cromwell/backend/impl/aws/AwsBatchJobCachingActorHelper.scala
```diff
@@ -56,13 +56,6 @@ trait AwsBatchJobCachingActorHelper extends StandardCachingActorHelper {
   lazy val callRootPath: Path = callPaths.callExecutionRoot
   lazy val returnCodeFilename: String = callPaths.returnCodeFilename
   // lazy val returnCodePath: Path = callPaths.returnCode
```
cjllanwarne: More `//`s to remove?

engine/src/main/scala/cromwell/engine/io/nio/NioFlow.scala
```diff
@@ -110,7 +111,9 @@ class NioFlow(parallelism: Int,
     hash.file match {
       case gcsPath: GcsPath => IO { gcsPath.cloudStorage.get(gcsPath.blob).getCrc32c }
       case drsPath: DrsPath => getFileHashForDrsPath(drsPath)
-      case path => IO.fromEither(
+      case s3Path: S3Path => IO { s3Path.eTag }
+      case path =>
```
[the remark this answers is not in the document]
geoffjentry: FWIW all that happens now is it throws an error and defaults to running the WF so I'm not sure that'd be gaining anything in a practical sense. OTOH I **am** the person who repeatedly made the claim that type safety was its own reward (ignoring that `asInstanceOf` in this PR!)

Up to you if you think it's worth having

filesystems/s3/src/main/scala/cromwell/filesystems/s3/S3PathBuilder.scala
```diff
@@ -167,6 +169,10 @@ case class S3Path private[s3](nioPath: NioPath,
 
   def key: String = safeAbsolutePath
 
+  lazy val s3Utils = new S3Utils()
+  lazy val s3Path = nioPath.asInstanceOf[org.lerch.s3fs.S3Path]
```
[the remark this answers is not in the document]
geoffjentry: By the time it gets here it's known to be a valid transformation. I'm not a huge fan of this and I **think** it can be alleviated directly in the underlying S3 code but this was a quicker win. I'm working out what can be done on the latter side

src/ci/bin/testCentaurAws.sh
geoffjentry: 🤷‍♂️ This is what I was shown to edit in a previous aws/centaur PR. TBH I like this better as it puts it easily visible in one place - e.g. if I ask the question "what's the set of known to be failing centaur tests", well now I know.

supportedBackends/aws/src/main/scala/cromwell/backend/impl/aws/AwsBatchJobCachingActorHelper.scala
geoffjentry: Sure. But I'll ask you don't go looking for more as that's a PR in and of itself 😛

geoffjentry: Oh wait, i intentionally left that one. I can 💯 guarantee that it'll be uncommented in the future

src/ci/bin/testCentaurAws.sh
kshakir: > rather than listing the allowable (or excluded) backends

AFAIK, there is not a way using only the `.test` files to say "run this test on PapiV2, Local, Slurm, but NOT AWS, PapiV1, Alibaba, TES". Thus we ended up with each `testCentaur*.sh` either opting-in or opting-out a set of tests.

engine/src/main/scala/cromwell/engine/io/nio/NioFlow.scala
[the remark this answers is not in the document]
kshakir: Yeah, not a sealed trait, so no compile time errors. But it could be a clearer runtime error by doing Chris' suggestion.

alfre_and_cromwell_small

Someday...

filesystems/s3/src/main/scala/cromwell/filesystems/s3/S3PathBuilder.scala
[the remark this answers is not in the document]
kshakir: Class cast issues are only exceptions, not instances of `java.lang.Error`, so it shouldn't threaten the VM. Depending on if the Filesystem was used in the call or the workflow level I'd expect either to fail, respectively.

Btw, same copypasta code, longer and slightly more scala-esque:

https://github.com/broadinstitute/cromwell/blob/f1cce2cd2723b849c4d8f285510f30913ec188a0/filesystems/gcs/src/main/scala/cromwell/filesystems/gcs/GcsPathBuilder.scala#L221-L224

https://github.com/broadinstitute/cromwell/blob/f1cce2cd2723b849c4d8f285510f30913ec188a0/filesystems/oss/src/main/scala/cromwell/filesystems/oss/OssPathBuilder.scala#L131-L134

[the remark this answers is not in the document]
geoffjentry: So by nature of being the `nioPath` of a `cromwell.filesystems.s3.S3Path` it's already been determined that it's a valid `org.lerch.s3fs.S3Path`. 

There's a lot of redundancy in this whole relationship (in the "duplicated code" sense, not the "better safe than sorry" sense) and I'm working out the best feasible path. But since that potentially requires a PR against something other than Cromwell I want to get the less than ideal fix out now, as yes - saying things like "I know it can't happen, see the code 20 lines above" isn't something I like saying.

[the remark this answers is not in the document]
geoffjentry: That was my way of saying "I'd really prefer to leave it as is for now"

engine/src/main/scala/cromwell/engine/io/nio/NioFlow.scala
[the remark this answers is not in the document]
geoffjentry: Fair 'nuff.

supportedBackends/aws/src/main/scala/cromwell/backend/impl/aws/AwsBatchJobCachingActorHelper.scala
geoffjentry: And now that I'm not running to catch a bus - this is going to be uncommented when whomever implements `continueOnReturnCode` implements `continueOnReturnCode`. I think it's better to leave those sorts of breadcrumbs here as it makes it easier for dopes like myself to follow the trail instead of figure it out from scratch :)

cjllanwarne: Personally every time I've left something like that for future-Chris, future-Chris ended up reimplementing the same line elsewhere anyway (or sometimes even right next to the comment without noticing).

But 🤷‍♂️ I don't want to hold up the PR over a preexisting commented line.

PR #4797, opened 2019-04-03
Ensure in containers tutorial, Singularity container name is valid

illusional:
When Cromwell returns the name for the docker container, there are often non compliant characters for the unix filesystem ('/', '@', ':'). When pulling and rebuilding the container, replace non alphabetical, numerical or some symbols (`[A-Za-z0-9._-]`) with an underscore (`_`).

And ensure that Singularity isn't constantly rebuilding containers with no reason. Even if this doesn't get merged, I think it could potentially be valuable for someone.

Basically, I just added these three lines to the `docker-submit` script for SLURM (applicable for all batch systems). It might be valuable to change `-f` to `-x`, to check if a file is executable by you (`help test`).

```bash
DOCKER_NAME=$(sed -e 's/[^A-Za-z0-9._-]/_/g' <<< ${docker})
IMAGE=${cwd}/$DOCKER_NAME.sif
if [ ! -f $IMAGE ]; then
    singularity pull $IMAGE docker://${docker}
fi
```

### DIFF
--- a/docs/tutorials/Containers.md
+++ b/docs/tutorials/Containers.md
@@ -181,14 +181,19 @@ When constructing this block, there are a few things to keep in mind:
 - Make sure Singularity is loaded (and in our path), for example you can call `module load Singularity/3.0.1`.
 - We should treat worker nodes as if they do not have stable access to the internet or build access, so we will pull the container before execution of the task.
 - It's a good idea to ask Singularity to build the image into the execution directory of the task as an artifact, and to save rendering time on the worker node.
+- We'll use the docker name (potentially including the docker hash) to generate a compliant filename that can be looked up later.
+- If the container exists, there is no need to rebuild the container.
 
 ```
 submit-docker = """
     module load Singularity/3.0.1
   
     # Build the Docker image into a singularity image, using the head node
-    IMAGE=${cwd}/${docker}.sif
-    singularity build $IMAGE docker://${docker}
+    DOCKER_NAME=$(sed -e 's/[^A-Za-z0-9._-]/_/g' <<< ${docker})
+    IMAGE=${cwd}/$DOCKER_NAME.sif
+    if [ ! -f $IMAGE ]; then
+        singularity pull $IMAGE docker://${docker}
+    fi
   
     # Submit the script to SLURM
     sbatch \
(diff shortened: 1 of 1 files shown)


cjllanwarne:
I ran the script through [shellcheck](https://www.shellcheck.net/) - if you could address these warnings that'd be great.

Screen Shot 2019-04-04 at 9 36 33 AM

NB: If they're not valid, maybe state why it's ok to ignore them? eg I assume `docker` is created further up the file so we hopefully can ignore that warning... but even better would be to generate an example of the final rendered script file and see what shellcheck has to say about the script as a whole? Thanks!

illusional:
@cjllanwarne woah I didn't know about this! I've fixed the script per the spellcheck output:

```bash
docker_subbed=$(sed -e 's/[^A-Za-z0-9._-]/_/g' <<< ${docker})
singularity_image=${cwd}/$docker_subbed.sif

if [ ! -f "$singularity_image" ]; then
    singularity pull "$singularity_image" docker://${docker}
fi
```

image

Addressing each point separately:
- [SC2001](https://github.com/koalaman/shellcheck/wiki/SC2001): I can't use the `${variable//search/replace}` syntax as Cromwell parses this expression and rejects the config as it would try to execute this when generating the block.
- [SC2154](https://github.com/koalaman/shellcheck/wiki/SC2154): `docker` and `cwd` are substituted by Cromwell.
- [SC2086](https://github.com/koalaman/shellcheck/wiki/SC2086): Similar to above, this is substituted into a string, so not needed at runtime.


Which results in something like this:
```bash
# Build the Docker image into a singularity image
docker_subbed=$(sed -e 's/[^A-Za-z0-9._-]/_/g' <<< quay.io/biocontainers/cutadapt@sha256:8799129dfef6de4e0503e8f2c20acafdf261793cc392f312d84df8016bfeef24)
singularity_image=/path/to/cromwell-executions/whole_genome_germline/47077714-9e90-49a6-91b6-902297b443ce/call-s1_alignsortedbam/shard-0/alignsortedbam/a8d7227b-20cf-4997-adf5-c1401d2eeb1f/call-cutadapt/$docker_subbed.sif

if [ ! -f "$singularity_image" ]; then
    singularity pull "$singularity_image" docker://quay.io/biocontainers/cutadapt@sha256:8799129dfef6de4e0503e8f2c20acafdf261793cc392f312d84df8016bfeef24
fi
```

Which leaves `$singularity_image` with the value 
```
/path/to/cromwell-executions/my_wf/$exec_id/etc/cutadapt/quay.io_biocontainers_cutadapt_sha256_8799129dfef6de4e0503e8f2c20acafdf261793cc392f312d84df8016bfeef24.sif
```

PR #4817, opened 2019-04-09
Tactical fix for centaur AWS 429s

cjllanwarne:
~~Might not make much difference until *all* our PRs are rebased to include this throttle~~

Actually this PR is more about allowing the AWS backend to hook into the existing poll retry logic to allow more resilience in the face of "429 / Too Many Request" exceptions during status polling.


wleepang:
Is it possible that this will work without system wide job rate throttling?  Also, is there a back-off rate for retries if hitting a 429?

PR #135, opened 2015-08-10
Updating Cromwell for the new WDL spec changes

scottfrazer:
Updated Language Spec: https://github.com/broadinstitute/wdl/blob/spec/SPEC.md#workflow-description-language

What you'll find in here:

* Updates to the parser for the various syntax changes we decided on.  Tests have also been updated and all WDL files I could find have been updated for the new syntax.
* `EngineFunctions` -> `WdlStandardLibraryFunctions` with stubs for everything that's in the specification.
* Implemented the following on `LocalEngineFunctions`: `write_map()`, `read_map()`, `write_lines()`.  Tests also added for these
* Refactoring of parts of the backend to be able to accommodate the implementation of the functions described above.
* Syntax Highlighter updated for new syntax
* Map literals now allowed in WDL source.  Tests included.
* Slightly better coercion semantics:  Added an identity coercion 
* Remove / refactor a lot of tech debt.
* Removed the `Command` class, now consumed into `Task` class.  Since the spec change, Commands needed to know more about stuff defined at the task level (specifically, Declarations).
* Exposure of workflow outputs section in the `Workflow` class, they are not honored further than that though.

PR #4819, opened 2019-04-09
CWL conformance: always set basename in outputs

aednichols:
This change is necessary, but not sufficient, for us to pass conformance #135, i.e. packed CWL.

We do not yet run #135 because it was added after our pinned commit. Once the other other necessary change, https://github.com/common-workflow-language/common-workflow-language/pull/849, lands I will make a PR to bump the pinned test cases.

I should disclose that I am aware of [an old comment](https://github.com/broadinstitute/cromwell/pull/3468#discussion_r178306868) explaining why the conditional is necessary, but conformance passes on the branch and as we say the proof of the pudding is in the eating.

### DIFF
--- a/centaurCwlRunner/src/main/scala/centaur/cwl/OutputManipulator.scala
+++ b/centaurCwlRunner/src/main/scala/centaur/cwl/OutputManipulator.scala
@@ -42,14 +42,7 @@ object OutputManipulator extends Poly1 {
     }
   }
 
-  /**
-    * @param isInsideDirectory Conformance tests expect "basename" to be filled in for files iff the file is listed
-    *                          in a directory. This field lets us know whether or not that's the case and if we should
-    *                          add the "basename" field to the output JSON.
-    * @param schemaOption      Optional schema.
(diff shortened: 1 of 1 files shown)


centaurCwlRunner/src/main/scala/centaur/cwl/OutputManipulator.scala
```diff
@@ -134,9 +127,8 @@ object OutputManipulator extends Poly1 {
       val defaultSize = sizeContent.orElse(sizeFile(path)).map(Json.fromLong).getOrElse(Json.Null)
       val size = valueOrNull("size", defaultSize)
 
-      val basename: Option[Json] = if (isInsideDirectory) {
+      val basename: Option[Json] =
```
mcovarr: seems like this is no longer `Option`al?

aednichols: It is still optional because of `valueOrNull`

PR #4821, opened 2019-04-10
Womtool endpoint - sleep well at night tests

aednichols:
Just some insurance above and beyond "that should definitely work"

### DIFF
--- a/services/src/test/scala/cromwell/services/womtool/impl/WomtoolServiceInCromwellActorSpec.scala
+++ b/services/src/test/scala/cromwell/services/womtool/impl/WomtoolServiceInCromwellActorSpec.scala
@@ -81,6 +81,19 @@ class WomtoolServiceInCromwellActorSpec extends ServicesSpec("Womtool") {
          |}
        """.stripMargin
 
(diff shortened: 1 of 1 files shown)


services/src/test/scala/cromwell/services/womtool/impl/WomtoolServiceInCromwellActorSpec.scala
```diff
@@ -132,6 +148,48 @@ class WomtoolServiceInCromwellActorSpec extends ServicesSpec("Womtool") {
...
+            ),
+            importedDescriptorTypes = List.empty,
+            meta = Map.empty,
+            parameterMeta = Map.empty
+          )
+        )
+      )
+    }
+
+    "return invalid for a valid no-inputs workflow with extraneous inputs" in {
```
cjllanwarne: Technically, I think this is only asserted in 1.0 (there are some cases where people use made up inputs as "comments" in their inputs files.

aednichols: Thank you and TIL

PR #4822, opened 2019-04-10
Cromwell ID in instrumentation path

### LINKED ISSUE #4787: Standardize a way to support different dashboards for each Cromwell
Determine whether our metrics can support labels or include the host in metric path.

A/C: Different instances of same dashboard for each Cromwell instance

salonishah11:
Closes #4787

### DIFF
--- a/docs/developers/Instrumentation.md
+++ b/docs/developers/Instrumentation.md
@@ -11,12 +11,12 @@ ignores these metrics, but Cromwell includes an alternate implementation that ca
 [StatsD](https://github.com/etsy/statsd) server.
 To specify this implementation in your configuration file:
 
-```bash
+```hocon
 services.Instrumentation.class = "cromwell.services.instrumentation.impl.statsd.StatsDInstrumentationServiceActor"
 ```
 Make sure to configure your StatsD service:
 
-```bash
+```hocon
 services.Instrumentation.config.statsd {
     hostname = "localhost" # Replace with your host
     port = 8125 # Replace with your port
@@ -27,10 +27,18 @@ services.Instrumentation.config.statsd {
 
 There is also an additional configuration value that can be set: 
 
-```bash
+```hocon
 # Rate at which Cromwell updates its gauge values (number of workflows running, queued, etc...)
 system.instrumentation-rate = 5 seconds
 ```
+
+If you have multiple Cromwell instance, and would like to separate the metrics path for each instance, set the `system.cromwell_id` with the unique identifier for your each Cromwell instance configuration. For example,
+
+```hocon
+system.cromwell_id = "cromwell-instance-1"
+```
+will prepend all the metrics with path `cromwell.cromwell-instance-1...` for the instance.
+
 
 **Metrics**
 
--- a/docs/developers/Instrumentation.md
+++ b/docs/developers/Instrumentation.md
@@ -32,7 +32,7 @@ There is also an additional configuration value that can be set:
 system.instrumentation-rate = 5 seconds
 ```
 
-If you have multiple Cromwell instance, and would like to separate the metrics path for each instance, set the `system.cromwell_id` with the unique identifier for your each Cromwell instance configuration. For example,
+If you have multiple Cromwell instances, and would like to separate the instrumentation path for each instance, set the `system.cromwell_id` with the unique identifier for your each Cromwell instance configuration. For example,
 
 ```hocon
 system.cromwell_id = "cromwell-instance-1"
--- a/docs/developers/Instrumentation.md
+++ b/docs/developers/Instrumentation.md
@@ -33,7 +33,6 @@ system.instrumentation-rate = 5 seconds
 ```
 
 If you have multiple Cromwell instances, and would like to separate the instrumentation path for each instance, set the `system.cromwell_id` with the unique identifier for your each Cromwell configuration. For example,
-
 ```hocon
 system.cromwell_id = "cromwell-instance-1"
 ```
--- a/docs/developers/Instrumentation.md
+++ b/docs/developers/Instrumentation.md
@@ -32,12 +32,12 @@ There is also an additional configuration value that can be set:
 system.instrumentation-rate = 5 seconds
 ```
 
-If you have multiple Cromwell instances, and would like to separate the instrumentation path for each instance, set the `system.cromwell_id` with the unique identifier for your each Cromwell instance configuration. For example,
+If you have multiple Cromwell instances, and would like to separate the instrumentation path for each instance, set the `system.cromwell_id` with the unique identifier for your each Cromwell configuration. For example,
 
 ```hocon
 system.cromwell_id = "cromwell-instance-1"
 ```
-will prepend all the metrics with path `cromwell.cromwell-instance-1...` for the instance.
+will prepend all the metrics with path `cromwell.cromwell-instance-1...` for that instance.
 
 
 **Metrics**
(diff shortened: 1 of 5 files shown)


services/src/test/scala/cromwell/services/instrumentation/impl/statsd/StatsDInstrumentationServiceActorSpec.scala
```diff
@@ -70,11 +80,37 @@ class StatsDInstrumentationServiceActorSpec extends TestKitSuite with FlatSpecLi
...
+      Set("prefix_value.cromwell.cromwell-101.test_prefix.test.metric.bucket.timing.m1_rate"),
+      prependCromwellInstancePrefix = true
     )
   ) foreach {
-    case StatsDTestBit(description, metric, expectedExactPackets, expectedFuzzyPackets) =>
+    case StatsDTestBit(description, metric, expectedExactPackets, expectedFuzzyPackets, prependCromwellInstancePrefix) =>
       it should description in {
-        val instrumentationActor = TestActorRef(new StatsDInstrumentationServiceActor(config, ConfigFactory.load(), registryProbe))
+
+        val globalConfig = if (prependCromwellInstancePrefix) cromwellInstanceConfig else ConfigFactory.load()
```
kshakir: This is so weird... and existing code! I haven't looked to closely, but the fact that a config with _only_ the `system.cromwell_id` key is passing tests implies that this `ConfigFactory.load()` could be simplified to `ConfigFactory.empty()`.

Anyway, if you want to add your `system.cromwell_id` to the `ConfigFactory.load()` instead of replacing the whole thing, use something like:

```scala
ConfigFactory.load().withValue("system.cromwell_id", ConfigValueFactory.fromAnyRef("cromwell-101"))
```

So, either both of values in the `if` should be full-ish configs, or both should be empty-ish.

services/src/main/scala/cromwell/services/instrumentation/impl/statsd/StatsDInstrumentationServiceActor.scala
```diff
@@ -41,9 +42,13 @@ object StatsDInstrumentationServiceActor {
   * by making use of downsampling and / or multi metrics packets: https://github.com/etsy/statsd/blob/master/docs/metric_types.md
   */
 class StatsDInstrumentationServiceActor(serviceConfig: Config, globalConfig: Config, serviceRegistryActor: ActorRef) extends Actor with DefaultInstrumented {
+  val CROMWELL_METRIC_PREFIX: String = "cromwell"
+
   val statsDConfig = StatsDConfig(serviceConfig)
+  val cromwellInstanceIdOption: Option[String] = globalConfig.as[Option[String]]("system.cromwell_id")
+
+  override lazy val metricBaseName = MetricName{cromwellInstanceIdOption.fold(CROMWELL_METRIC_PREFIX)( p => s"$CROMWELL_METRIC_PREFIX.$p")}
```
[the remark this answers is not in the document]
salonishah11: Ah this is so much better! TIL. Thank you!

services/src/main/scala/cromwell/services/instrumentation/impl/statsd/StatsDInstrumentationServiceActor.scala
```diff
@@ -41,9 +42,13 @@ object StatsDInstrumentationServiceActor {
   * by making use of downsampling and / or multi metrics packets: https://github.com/etsy/statsd/blob/master/docs/metric_types.md
   */
 class StatsDInstrumentationServiceActor(serviceConfig: Config, globalConfig: Config, serviceRegistryActor: ActorRef) extends Actor with DefaultInstrumented {
+  val CROMWELL_METRIC_PREFIX: String = "cromwell"
+
   val statsDConfig = StatsDConfig(serviceConfig)
+  val cromwellInstanceIdOption: Option[String] = globalConfig.as[Option[String]]("system.cromwell_id")
```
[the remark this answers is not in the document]
salonishah11: TIL

services/src/main/scala/cromwell/services/instrumentation/impl/statsd/StatsDInstrumentationServiceActor.scala
```diff
@@ -41,9 +42,13 @@ object StatsDInstrumentationServiceActor {
   * by making use of downsampling and / or multi metrics packets: https://github.com/etsy/statsd/blob/master/docs/metric_types.md
   */
 class StatsDInstrumentationServiceActor(serviceConfig: Config, globalConfig: Config, serviceRegistryActor: ActorRef) extends Actor with DefaultInstrumented {
+  val CROMWELL_METRIC_PREFIX: String = "cromwell"
```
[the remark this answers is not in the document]
salonishah11: 👍

services/src/test/scala/cromwell/services/instrumentation/impl/statsd/StatsDInstrumentationServiceActorSpec.scala
```diff
@@ -26,6 +26,12 @@ class StatsDInstrumentationServiceActorSpec extends TestKitSuite with FlatSpecLi
     """.stripMargin
   )
 
+  val cromwellInstanceConfig = ConfigFactory.parseString(
+    """
+      |system.cromwell_id = "cromwell-101"
+    """.stripMargin
+  )
```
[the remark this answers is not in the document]
salonishah11: 👍

services/src/test/scala/cromwell/services/instrumentation/impl/statsd/StatsDInstrumentationServiceActorSpec.scala
salonishah11: I am not 100% sure, but the `StatsDInstrumentationServiceActor` only requires `globalConfig` to extract the `system.cromwell_id` now. Before I don't think it was used at all. So I guess we can replace `ConfigFactory.load()` with `ConfigFactory.empty()` no? Or should we still load the default config?

kshakir: If tests pass, I would just use `empty()`.

Otherwise, here's another alternative to the `withValue()` above:

```scala
val cromIdConfig = if (prependCromwellInstancePrefix) cromwellInstanceConfig else ConfigFactory.empty()
val globalConfig = cromIdConfig.withFallBack(ConfigFactory.load())
```

docs/developers/Instrumentation.md
```diff
@@ -27,11 +27,18 @@ services.Instrumentation.config.statsd {
 
 There is also an additional configuration value that can be set: 
 
-```bash
+```hocon
 # Rate at which Cromwell updates its gauge values (number of workflows running, queued, etc...)
 system.instrumentation-rate = 5 seconds
 ```
 
+If you have multiple Cromwell instances, and would like to separate the instrumentation path for each instance, set the `system.cromwell_id` with the unique identifier for your each Cromwell configuration. For example,
```
mcovarr: suggestion: "set `system.cromwell_id` with the unique identifier for each Cromwell instance."

services/src/test/scala/cromwell/services/instrumentation/impl/statsd/StatsDInstrumentationServiceActorSpec.scala
salonishah11: Yes, the tests do pass with `ConfigFactory.empty()`. So I will go ahead with it 👍

docs/developers/Instrumentation.md
salonishah11: 👍

Issue #4103, opened 2018-09-14
enable parsing of packed cwl containing `SchemaDefRequirement` with multiple `$import:`


jeremiahsavage:
I am attempting to run a packed CWL workflow.

The output from the latest cromwell is:
```
(p3cwl) [jeremiah@localhost ~]$ git clone https://github.com/NCI-GDC/gdc-dnaseq-cwl.git
(p3cwl) [jeremiah@localhost ~]$ java -jar cromwell/server/target/scala-2.12/cromwell-35-fd560e9-SNAP.jar run gdc-dnaseq-cwl/workflows/bamfastq_align/transform_pack.cwl --inputs gdc-dnaseq-cwl/workflows/bamfastq_align/transform_c9bab596-0b03-4066-b428-d9cd36fb1c.yml --type CWL --workflow-root main
[2018-09-14 13:19:10,55] [info] Running with database db.url = jdbc:hsqldb:mem:d07a09a8-8d20-4095-967b-c6f375a3f309;shutdown=false;hsqldb.tx=mvcc
[2018-09-14 13:19:19,53] [info] Running migration RenameWorkflowOptionsInMetadata with a read batch size of 100000 and a write batch size of 100000
[2018-09-14 13:19:19,55] [info] [RenameWorkflowOptionsInMetadata] 100%
[2018-09-14 13:19:19,67] [info] Running with database db.url = jdbc:hsqldb:mem:e41fe9de-508c-4f49-aeaa-ce7474d7c1e2;shutdown=false;hsqldb.tx=mvcc
[2018-09-14 13:19:20,18] [info] Slf4jLogger started
[2018-09-14 13:19:20,25] [info] Pre Processing Workflow...
[2018-09-14 13:19:20,65] [info] Pre-Processing /home/jeremiah/gdc-dnaseq-cwl/workflows/bamfastq_align/transform_pack.cwl
[2018-09-14 13:19:54,70] [info] Pre Processing Inputs...
[2018-09-14 13:19:54,94] [info] Workflow heartbeat configuration:
{
  "cromwellId" : "cromid-89ab52b",
  "heartbeatInterval" : "2 minutes",
  "ttl" : "10 minutes",
  "writeBatchSize" : 10000,
  "writeThreshold" : 10000
}
[2018-09-14 13:19:55,04] [info] Metadata summary refreshing every 2 seconds.
[...]

PR #4823, opened 2019-04-11
Adopt newest CWL conformance

aednichols:
Follow-on to #4103


aednichols:
Just in time for thumbs to come in, I'm realizing that I rather blindly added `55` to expected failures when in fact there is a chance I broke it. Investigating...

aednichols:
Figured out that I need to update the contents of `gs://centaur-cwl-conformance` to match the latest commit

PR #4825, opened 2019-04-11
Custom google labels

### LINKED ISSUE #3900: Feature Request: Ability to add custom labels to Google VMs
[WDL Forums discussion](https://gatkforums.broadinstitute.org/wdl/discussion/12295/how-to-submit-labels-to-gcs-backend)

In #3233 the ability to customize GCS VM labels was removed. My understanding was that this was to lift Google's restrictions on VM labels from the workflow labels. However, I would benefit greatly from the ability to optionally supply labels to each VM. 

Specifically, I'm operating my own fork of the [cromwell_on_google](https://github.com/openwdl/wdl/tree/master/runners/cromwell_on_google/wdl_runner) tool which I use to spin up large batches of VMs to execute workflows. However, it's hard to pick out individual VMs from the compute api console, especially when running multiple unrelated submissions.

My tool creates one Cromwell VM for each submission, which then (via the Google JES backend) starts new VMs for each workflow. To distinguish and filter the VMs in the console, I want to attach two labels to every vm:

* `execution-role` : A label to distinguish between Cromwell VMs and worker VMs
* `submission-id` : A label with submission IDs generated by my tool to distinguish between VMs working on different submissions

The `cromwell-workflow-id` label is helpful, but not convenient if I wish to operate on all VMs working on a single submission _en masse_.

The Google genomics API allows me to set `execution-role=Cromwell` and `submission-id={ID}` labels on the Cromwell VMs. However, without Cromwell support, I can't set `execution-role=Worker` and `submission-id={ID}` labels on the workers at creation time. 

I'm currently pursuing a workaround to manually set these labels after creation, but having control via Cromwell would be preferable. 

Thanks!

cjllanwarne:
Re-add the ability to provide custom google labels. These can now be provided via the `google-labels` option in `workflowoptions.json`.

Closes #3900.

### DIFF
--- a/centaur/src/main/resources/standardTestCases/google_labels/bad_options.json
+++ b/centaur/src/main/resources/standardTestCases/google_labels/bad_options.json
@@ -1,5 +1,5 @@
 {
   "google_labels": {
-    "custom-label-label-label-label-label-label-label": "0custom-value"
+    "custom-label-label-label-label-label-label-label-label-label-label": "0custom-value"
   }
 }
--- a/centaur/src/main/resources/standardTestCases/google_labels/bad_options.json
+++ b/centaur/src/main/resources/standardTestCases/google_labels/bad_options.json
@@ -1,0 +1,5 @@
+{
+  "google_labels": {
+    "custom-label-label-label-label-label-label-label": "0custom-value"
+  }
+}
--- a/centaur/src/main/resources/standardTestCases/google_labels/google_labels.wdl
+++ b/centaur/src/main/resources/standardTestCases/google_labels/google_labels.wdl
@@ -25,6 +25,10 @@ workflow google_labels {
     ] }
   }
 
+  output {
+    Boolean done = true
+  }
+
 }
 
 task check_labels {
--- a/centaur/src/main/resources/standardTestCases/google_labels/google_labels.wdl
+++ b/centaur/src/main/resources/standardTestCases/google_labels/google_labels.wdl
@@ -1,13 +1,36 @@
 version 1.0
 
 workflow google_labels {
-  call check_labels
+
+  meta {
+    description: "Confirms that the google_labels workflow option is propogated correctly to tasks"
+  }
+
+  input {
+    Array[Pair[String, String]] expected_kvps = [ ("wdl-task-name", "check-labels") ]
+    Boolean check_aliases = true
+  }
+
+  # Build an array of checker lines to confirm the existence of each KVP:
+  scatter (p in expected_kvps) {
+    String checker_lines = "echo $INSTANCE_INFO | grep -q '\"~{p.left}\": \"~{p.right}\"' || echo \"Couldn't find correct '~{p.left}' label in instance info: $INSTANCE_INFO\" 1>&2"
+  }
+
+  call check_labels { input: label_checker_lines = checker_lines }
+
+  if (check_aliases) {
+    call check_labels as check_label_alias { input: label_checker_lines = [
+      "echo $INSTANCE_INFO | grep -q '\"wdl-task-name\": \"check-labels\"' || echo \"Couldn't find correct 'wdl-task-name' label in instance info: $INSTANCE_INFO\" 1>&2",
+      "echo $INSTANCE_INFO | grep -q '\"wdl-call-alias\": \"check-label-alias\"' || echo \"Couldn't find correct 'wdl-call-alias' label in instance info: $INSTANCE_INFO\" 1>&2"
+    ] }
+  }
+
 }
 
 task check_labels {
 
-  meta {
-    description: "Confirms that the google_labels workflow option is propogated correctly to tasks"
+  input {
+    Array[String] label_checker_lines
   }
 
   command {
@@ -19,7 +42,7 @@ task check_labels {
       -H 'Accept: application/json')
 
     # Check for the custom label in the instance info:
-    echo $INSTANCE_INFO | grep -q '"custom-label": "custom-value"' || echo "Couldn't find custom-label in instance info: $INSTANCE_INFO" 1>&2
+    ~{sep="\n" label_checker_lines}
   }
   runtime {
     docker: "google/cloud-sdk:slim"
--- a/centaur/src/main/resources/standardTestCases/google_labels/google_labels.wdl
+++ b/centaur/src/main/resources/standardTestCases/google_labels/google_labels.wdl
@@ -1,0 +1,30 @@
+version 1.0
+
+workflow google_labels {
+  call check_labels
+}
+
+task check_labels {
+
+  meta {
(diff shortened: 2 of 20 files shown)


centaur/src/main/resources/standardTestCases/google_labels/bad_options.json
```diff
@@ -0,0 +1,5 @@
+{
+  "google_labels": {
+    "custom-label-label-label-label-label-label-label": "0custom-value"
```
aednichols: Is this key intended to run afoul of a length limit? There does not appear to be an assertion for it, i.e. the test that uses this file fails for a different reason

centaur/src/main/resources/standardTestCases/google_labels/google_labels.wdl
```diff
@@ -0,0 +1,57 @@
+version 1.0
+
+workflow google_labels {
+
+  meta {
+    description: "Confirms that the google_labels workflow option is propogated correctly to tasks"
```
mcovarr: not propagating the correct spelling of propagated

centaur/src/main/resources/standardTestCases/google_labels/google_labels.wdl
```diff
@@ -0,0 +1,57 @@
...
+    INSTANCE_INFO=$(curl -s "https://www.googleapis.com/compute/v1/projects/broad-dsde-cromwell-dev/zones/us-central1-c/instances/$INSTANCE" \
+      -H "Authorization: Bearer $TOKEN" \
+      -H 'Accept: application/json')
+
+    # Check for the custom label in the instance info:
+    ~{sep="\n" label_checker_lines}
+  }
+  runtime {
+    docker: "google/cloud-sdk:slim"
+    zones: ["us-central1-c"]
```
mcovarr: why a specific zone?

docs/wf_options/Google.md
```diff
@@ -11,6 +11,7 @@ Keys | Possible Values | Description
...
 `monitoring_script` |`string` |   Specifies a GCS URL to a script that will be invoked prior to the user command being run.  For example, if the value for monitoring_script is `"gs://bucket/script.sh"`, it will be invoked as `./script.sh > monitoring.log &`.  The value `monitoring.log` file will be automatically de-localized.
 `monitoring_image` |`string` |   Specifies a Docker image to monitor the task. This image will run concurrently with the task container, and provides an alternative mechanism to `monitoring_script` (the latter runs *inside* the task container). For example, one can use `quay.io/broadinstitute/cromwell-monitor`, which reports cpu/memory/disk utilization metrics to [Stackdriver](https://cloud.google.com/monitoring/).
+`google_labels` | `object` | An object containing only string values. Represent custom google labels to send with PAPI job requests. Per the PAPI specification, each key and value must conform to the regex `a-z]([-a-z0-9]*[a-z0-9])?`.
```
mcovarr: that regex is broken, I think just missing an opening square brace

mcovarr: Also Google not google

supportedBackends/google/pipelines/common/src/main/scala/cromwell/backend/google/pipelines/common/PipelinesApiInitializationActor.scala
```diff
@@ -125,6 +125,7 @@ class PipelinesApiInitializationActor(pipelinesParams: PipelinesApiInitializatio
     for {
       paths <- workflowPaths
       _ = publishWorkflowRoot(paths.workflowRoot.pathAsString)
+      _ <- Future.fromTry(GoogleLabels.fromWorkflowOptions(workflowOptions))
```
aednichols: I am not too sure what this line does... `fromWorkflowOptions` does not seem to side-effect anything and then we throw away its result.

If we are checking that it completes successfully but don't care about the result perhaps a comment would be appropriate so people don't think it's useless and delete it

centaur/src/main/resources/standardTestCases/google_labels/bad_options.json
cjllanwarne: Now asserted

supportedBackends/google/pipelines/common/src/main/scala/cromwell/backend/google/pipelines/common/PipelinesApiInitializationActor.scala
cjllanwarne: Agreed. Added a comment.

centaur/src/main/resources/standardTestCases/google_labels/google_labels.wdl
cjllanwarne: Turns out I'm needing to bring this back after all. It *was* needed after all because the compute query via `curl` is using the zone in building the URL.

I'll add a comment to that effect...

agraubert:
👍 👍 👍 
Thanks so much! I've just updated my system with the latest Cromwell and this is perfect

blawney:
I need this ability to label Google VMs for resource tracking, but have been thus far unable to have a VM labelled correctly.  The jobs submit and run, but the labels do not show up.  

From the documentation here (https://cromwell.readthedocs.io/en/develop/wf_options/Google/), it's not clear where the google-specific options are added, so I tried the following:  
```
{
    "default_runtime_attributes":{
        "zones":"us-east1-b", 
        "google_labels": {"custom-label":"custom-value"}
    }
}
```
I submit (Cromwell v42) with:
```
curl -X POST "<CROMWELL URL>/api/workflows/v1" \
    -H  "accept: application/json" \
    -H  "Content-Type: multipart/form-data" \
    -F "workflowSource=@main.wdl" 
    -F "workflowInputs=@inputs.json" \
    -F "workflowOptions=@options.json" \
    -F "workflowType=WDL" \
    -F "workflowTypeVersion=draft-2"
```

That submits/runs fine, but when I check the VM that spins up, I only see the two labels of `cromwell-workflow-id` and `wdl-task-name`.  If I change the options JSON to anything else, e.g.
```
{
    "default_runtime_attributes":{"zones":"us-east1-b"},
    "google_labels": {"custom-label":"custom-value"}
}
```
then it fails to submit, returning:
```
{
  "status": "fail",
  "message": "Error(s): Invalid workflow options provided: Unsupported key/value pair in WorkflowOptions: google_labels -> {\"custom-label\":\"custom-value\"}"
}
```

pbilling:
I have the same issue with that page. Can someone please clarify where/in what file these options should be specified? Changing the example file from a snippet to an entire file would be helpful as well.

EvaLangu:
@blawney and @pbilling, I have the same exact issue; did you manage to get it working and if so would you mind sharing how you fixed it? Thank you

PR #4830, opened 2019-04-12
update metadata summary defaults

mcovarr:


cjllanwarne:
We should probably also update the "look like default" commented out values in `examples.conf`.

PR #4831, opened 2019-04-12
Expose new /query fields

### LINKED ISSUE #4667: if workflow is a subworkflow give its parent id
As you know I am developer of a cromwell Client application that visualizes cromwell results https://github.com/antonkulaga/cromwell-client In any more or less complex workflow you usually have several subworkflows, visualizing such workflow in a flat way makes the user output messy. 
I will be very grateful if you can provide parent_id for each subworkflow so I can setup a hierarcy in the UI.

cjllanwarne:
- Shows `parentWorkflowId` in `/query` results without having to opt-in
- Additionally shows `rootWorkflowId` in the results
- Corrects documentation of `additionalQueryParameters` to be less... overpromise-ey

Closes #4667


common/src/main/scala/common/collections/WeightedQueue.scala
```diff
@@ -14,9 +14,9 @@ object WeightedQueue {
   * In addition to the queue, a weight function is provided that provides the weight W of an element T.
   * The total weight of the queue is accessible, as well as a method to take the head of the queue based on a max weight value.
   */
-case class WeightedQueue[T, W](innerQueue: Queue[T],
-                               private val weightFunction: T => W,
-                               weight: W)(implicit n: Numeric[W]) {
+final case class WeightedQueue[T, W](innerQueue: Queue[T],
+                                     private val weightFunction: T => W,
+                                     weight: W)(implicit n: Numeric[W]) {
```
cjllanwarne: Oops - this is a stowaway change which doesn't have anything to do with the rest of the PR.

cjllanwarne:
Requesting reviews from:
- @kshakir because you created the values being exposed
- @rsasch to make sure adding new fields to `/query` won't play badly with Job Manager

PR #4836, opened 2019-04-16
Exclude failing Centaur tests

kshakir:


src/ci/bin/testCentaurHoricromtalPapiV2.sh
```diff
@@ -40,6 +40,9 @@ cromwell::build::run_centaur \
     -p 100 \
     -e localdockertest \
     "${CROMWELL_BUILD_CENTAUR_TEST_ADDITIONAL_PARAMETERS:-""}" \
-    -d "${CROMWELL_BUILD_CENTAUR_TEST_DIRECTORY}"
+    -d "${CROMWELL_BUILD_CENTAUR_TEST_DIRECTORY}" \
+    -e relative_output_paths \
+    -e relative_output_paths_colliding \
+    -e standard_output_paths_colliding_prevented \
```
cjllanwarne: Could these move up to join the existing `-e localdockertest` (or vice versa move that one down)?

src/ci/bin/testCentaurPapiV1.sh
```diff
@@ -26,6 +26,9 @@ cromwell::build::run_centaur \
     -p 100 \
     -e localdockertest \
     -e gpu_on_papi \
-    -d "${CROMWELL_BUILD_CENTAUR_TEST_DIRECTORY}"
+    -d "${CROMWELL_BUILD_CENTAUR_TEST_DIRECTORY}" \
+    -e relative_output_paths \
+    -e relative_output_paths_colliding \
+    -e standard_output_paths_colliding_prevented \
```
cjllanwarne: Same comment about keeping the `-e`s together

src/ci/bin/testCentaurPapiV2.sh
```diff
@@ -29,6 +29,9 @@ cromwell::build::run_centaur \
     -p 100 \
     -e localdockertest \
     "${CROMWELL_BUILD_CENTAUR_TEST_ADDITIONAL_PARAMETERS:-""}" \
-    -d "${CROMWELL_BUILD_CENTAUR_TEST_DIRECTORY}"
+    -d "${CROMWELL_BUILD_CENTAUR_TEST_DIRECTORY}" \
+    -e relative_output_paths \
+    -e relative_output_paths_colliding \
+    -e standard_output_paths_colliding_prevented \
```
cjllanwarne: I wonder whether the trailing `\` here is what's causing problems?

src/ci/bin/testCentaurAws.sh
```diff
@@ -55,6 +55,9 @@ cromwell::build::run_centaur \
     -e continue_on_return_code \
     -e globbingscatter \
     -e inline_file_custom_entryname \
-    -e draft3_globs   
+    -e draft3_globs \
+    -e relative_output_paths \
+    -e relative_output_paths_colliding \
+    -e standard_output_paths_colliding_prevented \
```
cjllanwarne:
```suggestion
    -e standard_output_paths_colliding_prevented 
```

kshakir: AWS passed both PR & push, so the trailing slashes by themselves weren't the problem... 🤔

mcovarr: AWS is also the only PR with all the `-e`s grouped, but I can't see from the code why not grouping `-e`s would be a problem

kshakir: Latest force-push gives up on cleanup / ordering. I'll figure out post release why the generated  `-l` options disappear when the `-e` options are passed in a different order.

PR #4837, opened 2019-04-16
Log workflow IDs and counts if metadata fails to be written

cjllanwarne:
I'm still hoping to add some metrics around this, but "log if metadata is dropped" seems like it could be a valuable "really quick win"?

### DIFF
--- a/services/src/main/scala/cromwell/services/metadata/impl/WriteMetadataActor.scala
+++ b/services/src/main/scala/cromwell/services/metadata/impl/WriteMetadataActor.scala
@@ -38,6 +38,9 @@ class WriteMetadataActor(override val batchSize: Int,
       case Success(_) =>
         putWithResponse foreach { case (ev, replyTo) => replyTo ! MetadataWriteSuccess(ev) }
       case Failure(regerts) =>
+        val workflowMetadataFailureCounts = e.toVector.flatMap(_.events).groupBy(x => x.key.workflowId).map { case (wfid, list) => s"$wfid: ${list.size}" }
+        log.error("Count of metadata events missed for the following workflows: " + workflowMetadataFailureCounts.mkString(","))
+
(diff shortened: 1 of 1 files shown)


services/src/main/scala/cromwell/services/metadata/impl/WriteMetadataActor.scala
```diff
@@ -38,6 +38,9 @@ class WriteMetadataActor(override val batchSize: Int,
       case Success(_) =>
         putWithResponse foreach { case (ev, replyTo) => replyTo ! MetadataWriteSuccess(ev) }
       case Failure(regerts) =>
+        val workflowMetadataFailureCounts = e.toVector.flatMap(_.events).groupBy(x => x.key.workflowId).map { case (wfid, list) => s"$wfid: ${list.size}" }
+        log.error("Metadata events permanently dropped for the following workflows: " + workflowMetadataFailureCounts.mkString(","))
```
aednichols: I feel like "dropped" doesn't quite do justice to the situation because sometimes things get dropped deliberately
```suggestion
        log.error("Metadata events irretrievably lost for the following workflows: " + workflowMetadataFailureCounts.mkString(","))
```

danbills: > sometimes things get dropped deliberately

Under what circumstances?

aednichols: One can deliberately drop a DB table during a migration - this was, perhaps confusingly, a comment on the word "drop" rather than on dropping metadata

danbills: got it, I was indeed confused

aednichols: @cjllanwarne you have my blessing to completely ignore this comment so as to avoid running tests again

PR #4839, opened 2019-04-17
Change "push" tests to be opt-in

cjllanwarne:
See the various commit test timings to see whether tests did or did not run.

Idea:
* One green light is as good as two since in any real situation the 'push' tests would be restarted if the 'PR' tests are green.
* We can still test on push if we really want to, using `FORCETEST` in the commit message
* We won't have so many "background" tests running at any given time, hopefully reducing the queue-time for PRs

In other words:
* Should significantly *increase* cycle time for PRs and *reduce* developer time shepherding PRs through two sets of identical tests

### DIFF
--- a/src/ci/bin/test.inc.sh
+++ b/src/ci/bin/test.inc.sh
@@ -103,7 +103,6 @@ cromwell::private::create_build_variables() {
             else
               CROMWELL_RUN_TESTS=false
             fi
-
             ;;
         "${CROMWELL_BUILD_PROVIDER_JENKINS}")
             # External variables must be passed through in the ENVIRONMENT of src/ci/docker-compose/docker-compose.yml
--- a/src/ci/bin/test.inc.sh
+++ b/src/ci/bin/test.inc.sh
@@ -103,6 +103,7 @@ cromwell::private::create_build_variables() {
             else
               CROMWELL_RUN_TESTS=false
             fi
+
             ;;
         "${CROMWELL_BUILD_PROVIDER_JENKINS}")
             # External variables must be passed through in the ENVIRONMENT of src/ci/docker-compose/docker-compose.yml
--- a/src/ci/bin/test.inc.sh
+++ b/src/ci/bin/test.inc.sh
@@ -78,6 +78,8 @@ cromwell::private::create_build_variables() {
       CROMWELL_BUILD_IS_VIRTUAL_ENV=false
     fi
 
+    CROMWELL_RUN_TESTS=true
+
     case "${CROMWELL_BUILD_PROVIDER}" in
         "${CROMWELL_BUILD_PROVIDER_TRAVIS}")
             CROMWELL_BUILD_IS_CI=true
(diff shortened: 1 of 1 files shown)


cjllanwarne:
> More helpful FORCEPUSH docs when exiting early

🤦‍♂️ that's not `FORCETEST`

src/ci/bin/test.inc.sh
```diff
@@ -138,6 +144,7 @@ cromwell::private::create_build_variables() {
             CROMWELL_BUILD_MYSQL_PASSWORD="${CROMWELL_BUILD_MYSQL_PASSWORD-}"
             CROMWELL_BUILD_MYSQL_SCHEMA="${CROMWELL_BUILD_MYSQL_SCHEMA-cromwell_test}"
             CROMWELL_BUILD_GENERATE_COVERAGE=true
+            CROMWELL_RUN_TESTS=true
```
mcovarr: could you DRY out everything by defaulting `CROMWELL_RUN_TESTS=true` and only setting it to `false` under the right circumstances?

cjllanwarne:
@kshakir I just made the following changes:

- SBT is now run on pushes, to make sure that the artifact publishing still happens. Since it's ~30 minutes, not dependent on external services, and less flaky than the other tests I still think this is an improvement over today
- I switched the syntax to `[force ci]`. See the most recent commit message and it triggering the tests to run.

src/ci/bin/test.inc.sh
```diff
@@ -747,6 +756,10 @@ cromwell::private::kill_tree() {
 
 cromwell::build::exec_test_script() {
     cromwell::private::create_build_variables
+    if [[ "${CROMWELL_RUN_TESTS}" == "false" ]]; then
+      echo "Use 'FORCETEST' in commit message to run tests on 'push'"
```
kshakir:
```suggestion
      echo "Use '[force ci]' in commit message to run tests on 'push'"
```

src/ci/bin/test.inc.sh
```diff
@@ -79,6 +78,8 @@ cromwell::private::create_build_variables() {
       CROMWELL_BUILD_IS_VIRTUAL_ENV=false
     fi
 
+    CROMWELL_RUN_TESTS=true
```
kshakir:
```suggestion
    CROMWELL_BUILD_RUN_TESTS=true
```

All variables for the build are prefixed as `CROMWELL_BUILD_` to distinguish that the variable is affecting the **cromwell repository build**, versus just building the **cromwell server executable**.

src/ci/bin/test.inc.sh
```diff
@@ -48,7 +48,6 @@ cromwell::private::create_build_variables() {
     else
         CROMWELL_BUILD_PROVIDER="${CROMWELL_BUILD_PROVIDER_UNKNOWN}"
     fi
-
```
kshakir: Revert?

src/ci/bin/test.inc.sh
```diff
@@ -870,6 +883,8 @@ cromwell::build::publish_artifacts() {
     fi
 }
 
+
+
```
kshakir: Revert?

PR #4840, opened 2019-04-17
Add test to verify aliased tasks run in separate directories

geoffjentry:


centaur/src/main/resources/standardTestCases/alias_separation/alias.wdl
```diff
@@ -0,0 +1,29 @@
...
+  command {
+    echo ${str} >> outfile
+  }
+
+  output {
+    Array[String] out = read_lines("outfile")
+  }
+
+  runtime {
+    docker: "amazonlinux:latest"
```
cjllanwarne: This seems like a weird choice for image

geoffjentry:
If you’d like to submit such a test go for it 🙂

centaur/src/main/resources/standardTestCases/alias_separation/alias.wdl
geoffjentry: It was from the exemplar in the ticket, tho the ticket was in the vein of "this doesn't work in amazon" so yeah i'll change it

geoffjentry:
@cjllanwarne updated the container image. I'm not inclined to add a glob test as I don't think it's really adding any value. As I said previously, if you'd like to add it knock yourself out

mcovarr:
TOL / pipe dream: inspired by the lovely system we have for CWL conformance testing, it would be nice to give Centaur distinct concepts for `-e` (exclude a test that is conceptually inappropriate for this configuration) and `-s` (really should work on this configuration but right now doesn't). Centaur could try to run "shoulda" tests with the sense of pass and fail reversed.

cjllanwarne:
> I'm not inclined to add a glob test as I don't think it's really adding any value

The value it would add is guaranteeing that the aliased task directories really are guaranteed to be distinct. Eg this particular test case would pass if the tasks were ran in serial (cf. our "concurrent job start limit" option for AWS tests)

geoffjentry:
@cjllanwarne I thought the issue is that they're literally running in the same directory, and thus it shouldn't matter regardless of when they run. I'll admit that this interpretation is just from the bug reports and could be entirely incorrect

cjllanwarne:
Sure... well in the spirit of moving the ball forward, there's no reason not to merge this. LGTM

geoffjentry:
As per hackathon - this test doesn't reliably fail in the intended circumstances. Closing in lieu of #4848

PR #4864, opened 2019-04-18
[cromwell_aws_hackathon] Ecr support

mcovarr:


geoffjentry:
if #4865 has already merged before this one, update that table

mcovarr:
On hold for now. Using regular auth this appears to work manually but not in code. e.g. doing something like [this](https://docs.aws.amazon.com/AmazonECR/latest/userguide/Registries.html#registry_auth):

```
TOKEN=$(aws ecr get-authorization-token --output text --query 'authorizationData[].authorizationToken')
curl -i -H "Authorization: Basic $TOKEN" https://952500931424.dkr.ecr.us-east-1.amazonaws.com/v2/broadinstitute/cromwell/manfests/latest
```

returns a blob of JSON containing all sorts of manifesty-looking data. However this doesn't seem to match up exactly with what the current code is doing: 

* The current code sets `Bearer` auth, but that `curl` command works with `Basic` and not `Bearer`.
* The current code expects to find the digest in the returned headers. However the digest is in the body and not the headers.

mcovarr:
https://github.com/keel-hq/keel/issues/173

wleepang:
The `ListImages` API call to ECR should return an image digest in the response body.  Can this be used instead?

mcovarr:
Doesn't that API list all the images in a repository? https://docs.aws.amazon.com/AmazonECR/latest/APIReference/API_ListImages.html

mcovarr:
I have some config-parsing code that appears to work but is too gross to push, I'll see if I can clean it up.

wleepang:
@mcovarr - yes.  Alteratively, you could use `DescribeImages` and supply an `ImageIds` argument to filter down the results by either `imageTag` or `imageDigest`.

mcovarr:
To attempt to accommodate ECR I changed Cromwell's Docker Registry V2 code to use the digest provided in the body of the response rather than the headers. Unfortunately this causes the tests on all the other registries (Docker Hub, GCR and Quay) to fail. Tabling for now.

TMiguelT:
I'm confused, why can't you try to get the digest from the header and fallback to the body if that doesn't work? That's what the keel PR does: https://github.com/rusenask/docker-registry-client/pull/1/files

mcovarr:
The comment in the keel PR asserts that the hashes returned in `Docker-Content-Digest` and within the body are the same, but in my limited testing on Docker Hub, Quay and GCR that did not seem to be the case. If the body value actually represents something other than the image digest it may cause downstream issues for Cromwell to treat as such. 

Basically this issue needs more investigation. Since ECR support is apparently not as a high a priority as originally thought this issue has been deprioritized for now.

TMiguelT:
I don't think it implies the header and body will **both** have the same hash for any given request. It sounds more like it will either be in the header or the body, hence they check one and then the other. Can you point me to a docker registry that doesn't follow this rule?

mcovarr:
Per @ruchim, closing the PR for now, the branch can live on as the basis for future work (maybe).

PR #4865, opened 2019-04-18
docker compatibility for AWS Batch

ruchim:
Update what docker repos are compatible with AWS Batch


geoffjentry:
Merging as this is all green and there were no unresolved comments

PR #4868, opened 2019-04-19
Fix syntax of docker build command

dtenenba:
The command as written was incorrect. Works now.


geoffjentry:
So I had thought that I had just changed this recently but it turns out that it was [in a different file](https://github.com/broadinstitute/cromwell/tree/develop/scripts/docker-develop). 

I think that these docs are different enough that duplication is fine, but for other Cromwellians reviewing this I figured I'd point it out. @dtenenba NB that I'm not asking you to change anything

geoffjentry:
Thanks @dtenenba !

PR #4881, opened 2019-04-23
[40_hotfix] Ensure HealthMonitorServiceActor can initialize without google config

cjllanwarne:


cjllanwarne:
Ignoring codecov hiccup. See https://github.com/broadinstitute/cromwell/pull/4876 for proof that testing does indeed cover the diff

PR #4886, opened 2019-04-24
Don't pass on Sam Conflict [develop]

kshakir:


CromIAM/src/main/scala/cromiam/sam/SamClient.scala
```diff
@@ -136,15 +136,17 @@ class SamClient(scheme: String,
     val createCollection = registerCreation(user, collection, cromIamRequest)
 
     createCollection flatMap {
-      case r if r.status == StatusCodes.Conflict => requestAuth(CollectionAuthorizationRequest(user, collection, "add"), cromIamRequest)
       case r if r.status == StatusCodes.NoContent => Monad[FailureResponseOrT].unit
       case r => FailureResponseOrT[IO, HttpResponse, Unit](IO.raiseError(SamRegisterCollectionException(r.status)))
+    } recoverWith {
+      case r if r.status == StatusCodes.Conflict => requestAuth(CollectionAuthorizationRequest(user, collection, "add"), cromIamRequest)
```
cjllanwarne: Is this creating an infinite retry loop? Do we want a break-clause in here?

CromIAM/src/test/scala/cromiam/sam/SamClientSpec.scala
```diff
@@ -46,28 +49,37 @@ class SamClientSpec extends AsyncFlatSpec with Matchers with BeforeAndAfterAll
...
   it should "eventually return the collection(s) of user" in {
+    val samClient = new MockSamClient()
     samClient.collectionsForUser(authorizedUserWithCollection, emptyHttpRequest).map(collectionList =>
-      assert(collectionList == samClient.userCollectionList)
+      assert(collectionList == MockSamClient.UserCollectionList)
     ).asIo.unsafeToFuture()
   }
 
   it should "fail if user doesn't have any collections" in {
+    val samClient = new MockSamClient()
```
cjllanwarne: ToL (⚠️ "Semantics of 'mock' discussion" alert ⚠️): 

Personally I'd go with `TestableSamClient` rather than `MockSamClient` since this isn't really mocking the sam client, it's extending in such a way as to make it testable.

CromIAM/src/test/scala/cromiam/sam/SamClientSpec.scala
```diff
@@ -76,19 +88,14 @@ class SamClientSpec extends AsyncFlatSpec with Matchers with BeforeAndAfterAll
     )
   }
 
-  it should "return sam errors while checking collections" in {
```
cjllanwarne: By deleting this test are you saying it should *never* return sam errors?

CromIAM/src/test/scala/cromiam/sam/SamClientSpec.scala
```diff
@@ -117,10 +120,80 @@ class SamClientSpec extends AsyncFlatSpec with Matchers with BeforeAndAfterAll
...
+      }
+
+      override def requestAuth(authorizationRequest: CollectionAuthorizationRequest,
+                               cromiamRequest: HttpRequest): FailureResponseOrT[Unit] = {
+        Monad[FailureResponseOrT].unit
+      }
+    }
+    samClient
+      .requestSubmission(unauthorizedUserWithNoCollection, unauthorizedCollection, emptyHttpRequest)
+      .map(_ => succeed)
```
cjllanwarne: But what happens if this never gets called? Does the test finish running and never fail, thus succeed by default?

Could you make this update a "I did what you expected" variable to true, and assert that it eventually happens with something like this?
```
eventually {
  didWhatIExpected should be(true)
}
```

CromIAM/src/test/scala/cromiam/sam/SamClientSpec.scala
```diff
@@ -117,10 +120,80 @@ class SamClientSpec extends AsyncFlatSpec with Matchers with BeforeAndAfterAll
...
+                                              cromiamRequest: HttpRequest): FailureResponseOrT[HttpResponse] = {
+        val unexpectedFailureResponse = HttpResponse(StatusCodes.ImATeapot, entity = HttpEntity("elided error message"))
+        returnResponse(unexpectedFailureResponse)
+      }
+    }
+    recoverToExceptionIf[SamRegisterCollectionException] {
+      samClient.requestSubmission(unauthorizedUserWithNoCollection, unauthorizedCollection, emptyHttpRequest)
+        .asIo.unsafeToFuture()
+    } map { exception =>
+      exception.getMessage should be("Can't register collection with Sam. Status code: 418 I'm a teapot")
```
cjllanwarne: 🍵

CromIAM/src/test/scala/cromiam/webservice/MockClients.scala
```diff
@@ -88,73 +90,73 @@ class MockCromwellClient()(implicit system: ActorSystem,
...
-  val unauthorizedUserCollectionStr: String = "987654321"
-
-  val notWhitelistedUser: String = "ABC123"
+  )(system, ece, materializer)
 
-  val userCollectionList: List[Collection] = List(Collection("col1"), Collection("col2"))
+/**
+  * Extends the base mock client with overriden methods.
+  */
+class MockSamClient(checkSubmitWhitelist: Boolean = true)
```
cjllanwarne: I'd make this more specific, not less (eg `UserValidatingSamClient`)

CromIAM/src/main/scala/cromiam/sam/SamClient.scala
kshakir: Unless I've made a cut-paste errors, I believe these are still two different calls, where the latter never calls the former, thus no looping:

- The outer call is still `requestSubmission`, aka CREATE
- This inner call is still `requestAuth`, aka UPDATE

https://github.com/broadinstitute/cromwell/blob/852a330a6a3ad3b06e839112b752bdbfd0efe2a8/CromIAM/src/main/scala/cromiam/sam/SamClient.scala#L125-L129

CromIAM/src/test/scala/cromiam/sam/SamClientSpec.scala
kshakir: Happy to rename this class, however the `Mock` prefix is an existing pattern throughout Cromiam specs though, and I'm not sure where to stop w/ the cleanup...

The `MockSamClient` is in a file called `MockClients`. In `MockClients` there's also a `MockCromwellClient`. Cromiam actually has two of these classes though, 
`cromiam.webservice.MockCromwellClient` and `cromiam.cromwell.CromwellClientSpec.MockCromwellClient`. And there's `MockCromwellApiClient` and `MockStatusService`.

In general, I'm more a fan of consistency (leave them all prefixed `Mock`, or change them _all_) but in this case I'm open to incremental progress?

CromIAM/src/test/scala/cromiam/sam/SamClientSpec.scala
kshakir: Related to the not-really-a-Mock discussion, I believe this Spec was testing the "Mock" method, and not actually reaching any part of the main-code.

CromIAM/src/test/scala/cromiam/sam/SamClientSpec.scala
kshakir: I'm assuming the "never gets called" refers to the `requestSubmission`/`requestAuth` loop-vs-create/update discussion. If there's no loop the code shouldn't need an eventually.

CromIAM/src/test/scala/cromiam/webservice/MockClients.scala
kshakir: Names are hard... I we go forward with cleaning up the names in this PR I'll change this one too.

CromIAM/src/test/scala/cromiam/sam/SamClientSpec.scala
cjllanwarne: Not a strong opinion, just a TOL. Since it's not a new addition, I'd be happy to leave them as-is

CromIAM/src/test/scala/cromiam/sam/SamClientSpec.scala
cjllanwarne: I mean that even without this `.map(_ => succeed)`, or even if the operation fails, the test wouldn't this test still "pass"... because that's the default and nothing is ever asserting otherwise?

CromIAM/src/test/scala/cromiam/webservice/MockClients.scala
cjllanwarne: Another case of TOL. I'll leave it up to you and won't hold back approval if you don't

CromIAM/src/test/scala/cromiam/sam/SamClientSpec.scala
kshakir: This class is an [Asynchronous ScalaTest Spec](http://www.scalatest.org/user_guide/async_testing), required to return a `Future[Assertion]`<sup>1</sup>. That's the only reason why the `map(_ => succeed)` is even there, because the api/compiler won't allow you to even accidentally return anything else, not even `Unit` nor `Future[Unit]`.

So if the operation fails, the result will be a failed-Future, which will fail test.

<hr/>
<sup>1</sup> There is built-in ScalaTest sugar that allows one to also end an Async-Spec with just an `Assertion` like `succeed` that will implicitly get converted to `Future[Assertion]`.

cjllanwarne: Aha! That's actually a cool pattern - and which I have probably reinvented at various points in the past. The more you know...

kshakir:
Leaving re-naming for another day. Merging.

PR #4898, opened 2019-04-24
Don't let awkward scheduled actions wreck test cases

cjllanwarne:
Or, "unflakify `WriteMetadataSpec`"

This test's assertions that metadata write batching and retrying would happen in the way it was expecting them to was upset if a scheduled metadata flush occurred before the first batch was full.

This was leading to a non-trivial rate of sbt tests failing with messages like "expected 50 database writes but got 52"

### DIFF
--- a/core/src/main/scala/cromwell/core/actor/BatchActor.scala
+++ b/core/src/main/scala/cromwell/core/actor/BatchActor.scala
@@ -1,5 +1,7 @@
 package cromwell.core.actor
 
+import java.time.OffsetDateTime
+
 import akka.actor.{ActorRef, FSM, Timers}
 import akka.dispatch.ControlMessage
 import cats.data.NonEmptyVector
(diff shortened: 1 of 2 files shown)


core/src/main/scala/cromwell/core/actor/BatchActor.scala
```diff
@@ -72,14 +74,28 @@ abstract class BatchActor[C](val flushRate: FiniteDuration,
 
   def commandToData(snd: ActorRef): PartialFunction[Any, C]
 
+  // If work has arrived 'recently' - ie within this duration - ignore any periodic 'ScheduledProcessAction' messages
+  val recentArrivalThreshold: Option[FiniteDuration] = None
```
aednichols: This val can never change and stores no information. Does that mean line 89 is a `None foreach { ... }`? What does that even do?

cjllanwarne:
@mcovarr @aednichols you both asked the same question in different directions... I'll try to answer both in one go:

* This *does not* alter how the class operates in production because it's a `None foreach { ... }`.
* I override the `None` in the spec so that I can guarantee the events land as a unit and aren't interrupted by the occasional flushing action.

FWIW I did try to move *all* of this logic into the test class to avoid cluttering the main, but got tangled up trying to override the FSM actions with a `receive` in the test class and it ended up not working as I'd hoped.

cjllanwarne:
FWIW 2: it's also possible to "fix" the tests by pushing the flush interval back and thus reduce the chance that the probe messages occur at the same time as a flush message. The downside to that was that the tests were taking 20 seconds each, which didn't feel great (and even though unlikely, there was still a small chance of an accident happening)

PR #4899, opened 2019-04-24
fail fast if output is never going to be there, plus shellcheck

geoffjentry:


supportedBackends/aws/src/main/resources/ecs-proxy/proxy
```diff
@@ -94,7 +94,13 @@ copyfile() {
     if [[ "$localname" == glob-*/"*" ]]; then
       awscp sync "${localdirectory}/${name}" "${s3location}"
     else
-      awscp cp "${localdirectory}/${localname}" "${s3location}"
+      localfilepath="${localdirectory}/${localname}"
+      if [ -e  "$localfilepath" ] ; then
+        awscp cp "${localfilepath}" "${s3location}"
+      else
+        echo "No such local file: ${localfilepath}"
```
cjllanwarne: Pipe this to stderr?

geoffjentry: we chatted via slack earlier but for posterity - doing so won't accomplish anything (and **potentially** could have a deleterious effect) due to how/why this script is run and other candidates for stderr'ing are also going to stdout. I can explore that particular set of belts & suspenders when i return

supportedBackends/aws/src/main/resources/ecs-proxy/proxy
```diff
@@ -145,19 +151,20 @@ fi
...
-fi
 echo "Copying rc file"
 copyToS3 "${AWS_CROMWELL_RC_FILE}" "${AWS_CROMWELL_CALL_ROOT}/"
 echo "Copying stdout file"
 copyToS3 "${AWS_CROMWELL_STDOUT_FILE}" "${AWS_CROMWELL_CALL_ROOT}/"
 echo "Copying stderr file"
 copyToS3 "${AWS_CROMWELL_STDERR_FILE}" "${AWS_CROMWELL_CALL_ROOT}/"
 
+# Wait until copying stdout/stderr/rc before copying outputs to ensure they're always there in case of
+# transfer error. Check first that there are outputs prior to trying to copy the outputs
```
mcovarr: not sure why that's better but whatevs

